PR #1578, opened 2020-09-01
[FIXED] Error when importing an account results in an error, retry later

matthiashanel:
This failure can be caused if the imported account can't be fetched or
is not up to date yet.

This change introduces a new account flag.
If this flag is set, perform the update and do NOT return ErrAccountResolverSameClaims.
Added two unit tests for both situations.
Shorten one unit test that was able to use the default operator.

### DIFF
--- a/server/accounts.go
+++ b/server/accounts.go
@@ -2351,6 +2351,14 @@ func (a *Account) isClaimAccount() bool {
 // This will replace any exports or imports previously defined.
 // Lock MUST NOT be held upon entry.
 func (s *Server) UpdateAccountClaims(a *Account, ac *jwt.AccountClaims) {
+	s.updateAccountClaimsWithRefresh(a, ac, true)
+}
+
+// updateAccountClaimsWithRefresh will update an existing account with new claims.
+// If refreshImportingAccounts is true it will also update incomplete dependent accounts
+// This will replace any exports or imports previously defined.
+// Lock MUST NOT be held upon entry.
+func (s *Server) updateAccountClaimsWithRefresh(a *Account, ac *jwt.AccountClaims, refreshImportingAccounts bool) {
 	if a == nil {
 		return
 	}
@@ -2633,7 +2641,7 @@ func (s *Server) UpdateAccountClaims(a *Account, ac *jwt.AccountClaims) {
 		}
 	}
 
-	if _, ok := s.incompleteAccExporterMap.Load(old.Name); ok {
+	if _, ok := s.incompleteAccExporterMap.Load(old.Name); ok && refreshImportingAccounts {
 		s.incompleteAccExporterMap.Delete(old.Name)
 		s.accounts.Range(func(key, value interface{}) bool {
 			acc := value.(*Account)
@@ -2646,7 +2654,9 @@ func (s *Server) UpdateAccountClaims(a *Account, ac *jwt.AccountClaims) {
 			acc.mu.RUnlock()
 			if incomplete && name != old.Name {
 				if accClaims, _, err := s.verifyAccountClaims(claimJWT); err == nil {
-					s.UpdateAccountClaims(acc, accClaims)
+					// Since claimJWT has not changed, acc can become complete
+					// but it won't alter incomplete for it's dependents accounts.
+					s.updateAccountClaimsWithRefresh(acc, accClaims, false)
 					// old.Name was deleted before ranging over accounts
 					// If it exists again, UpdateAccountClaims set it for failed imports of acc.
 					// So there was one import of acc that imported this account and failed again.
--- a/server/accounts.go
+++ b/server/accounts.go
@@ -2449,7 +2449,7 @@ func (s *Server) UpdateAccountClaims(a *Account, ac *jwt.AccountClaims) {
 			a.mu.Unlock()
 		}
 	}
-	incomplete := false
+	var incompleteImports []*jwt.Import
 	for _, i := range ac.Imports {
 		// check tmpAccounts with priority
 		var acc *Account
@@ -2461,8 +2461,7 @@ func (s *Server) UpdateAccountClaims(a *Account, ac *jwt.AccountClaims) {
 		}
 		if acc == nil || err != nil {
 			s.Errorf("Can't locate account [%s] for import of [%v] %s (err=%v)", i.Account, i.Subject, i.Type, err)
-			incomplete = true
-			s.incompleteAccExporterMap.Store(i.Account, struct{}{})
+			incompleteImports = append(incompleteImports, i)
 			continue
 		}
 		switch i.Type {
@@ -2470,16 +2469,14 @@ func (s *Server) UpdateAccountClaims(a *Account, ac *jwt.AccountClaims) {
 			s.Debugf("Adding stream import %s:%q for %s:%q", acc.Name, i.Subject, a.Name, i.To)
 			if err := a.AddStreamImportWithClaim(acc, string(i.Subject), string(i.To), i); err != nil {
 				s.Debugf("Error adding stream import to account [%s]: %v", a.Name, err.Error())
-				incomplete = true
-				s.incompleteAccExporterMap.Store(i.Account, struct{}{})
+				incompleteImports = append(incompleteImports, i)
 			}
 		case jwt.Service:
 			// FIXME(dlc) - need to add in respThresh here eventually.
 			s.Debugf("Adding service import %s:%q for %s:%q", acc.Name, i.Subject, a.Name, i.To)
 			if err := a.AddServiceImportWithClaim(acc, string(i.Subject), string(i.To), i); err != nil {
 				s.Debugf("Error adding service import to account [%s]: %v", a.Name, err.Error())
-				incomplete = true
-				s.incompleteAccExporterMap.Store(i.Account, struct{}{})
+				incompleteImports = append(incompleteImports, i)
 			}
 		}
 	}
@@ -2582,7 +2579,10 @@ func (s *Server) UpdateAccountClaims(a *Account, ac *jwt.AccountClaims) {
 		}
 	}
 	a.defaultPerms = buildPermissionsFromJwt(&ac.DefaultPermissions)
-	a.incomplete = incomplete
+	a.incomplete = len(incompleteImports) != 0
+	for _, i := range incompleteImports {
+		s.incompleteAccExporterMap.Store(i.Account, struct{}{})
+	}
 	a.mu.Unlock()
 
 	clients := gatherClients()
@@ -2647,6 +2647,10 @@ func (s *Server) UpdateAccountClaims(a *Account, ac *jwt.AccountClaims) {
 			if incomplete && name != old.Name {
 				if accClaims, _, err := s.verifyAccountClaims(claimJWT); err == nil {
 					s.UpdateAccountClaims(acc, accClaims)
+					// old.Name was deleted before ranging over accounts
+					// If it exists again, UpdateAccountClaims set it for failed imports of acc.
+					// So there was one import of acc that imported this account and failed again.
+					// Since this account just got updated, the import itself may be in error. So trace that.
 					if _, ok := s.incompleteAccExporterMap.Load(old.Name); ok {
 						s.incompleteAccExporterMap.Delete(old.Name)
 						s.Errorf("Account %s has issues importing account %s", name, old.Name)
--- a/server/accounts.go
+++ b/server/accounts.go
@@ -71,6 +71,7 @@ type Account struct {
 	jsLimits     *JetStreamAccountLimits
 	limits
 	expired      bool
+	incomplete   bool
 	signingKeys  []string
 	srv          *Server // server this account is registered with (possibly nil)
 	lds          string  // loop detection subject for leaf nodes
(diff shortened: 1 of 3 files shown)


matthiashanel:
@kozlovic  This would probably go hand in hand with disconnecting clients and evicting the account from memory
>   We could always add an "incomplete failure count" in the account and bump this in reloadAccount() and when reaching a certain threshold stop trying (maybe disconnect clients in that case?)

server/accounts.go
```diff
@@ -71,6 +71,7 @@ type Account struct {
 	jsLimits     *JetStreamAccountLimits
 	limits
 	expired      bool
+	incomplete   bool
```
derekcollison, line 74: Not for now but should turn these into flags, possibly with own mutex.

server/accounts.go
```diff
@@ -2623,6 +2629,35 @@ func (s *Server) UpdateAccountClaims(a *Account, ac *jwt.AccountClaims) {
 			}
 		}
 	}
+
+	// if we encountered issues during this process, retry in a bit
```
derekcollison, line 2633: This feels like it could possibly spin no?

server/accounts.go
```diff
@@ -2623,6 +2629,35 @@ func (s *Server) UpdateAccountClaims(a *Account, ac *jwt.AccountClaims) {
...
+		})
+	}
+}
+
+func (s *Server) reloadAccount(accName string) {
+	if !s.isRunning() {
+		return
+	}
+	if acc, ok := s.accounts.Load(accName); ok {
+		acc := acc.(*Account)
```
derekcollison, line 2647: Usually like to use diff variable name vs aliasing.

server/accounts.go
kozlovic: Like I mentioned, we could always keep track of the number of times we tried and give up after a certain amount.

server/accounts.go
matthiashanel: changed first acc to a

server/accounts.go
derekcollison: If there is an import that fails I think that means the import is invalid but does not mean anything about the account itself. Should we be trying to reload?

matthiashanel: or it means the account exporting has not been updated (to add the export) yet.

matthiashanel: If this becomes an issue we can add a limit as @kozlovic said. When we do that we probably should also evict the account. (smth yet to be added)

derekcollison: Can't add an import without the export. This is more the import required a token and the token expired or was revoked, or the exporting account was updated to remove or change the export.

derekcollison: I don't think if an import fails the account should be evicted at all. That would mean the user can not connect at all. We would want to make it easy for the importing account to debug though.

matthiashanel: If you modify two accounts (add an export and corresponding import) and push all, because the server has to process one of them first you have a high likelihood of that happening

derekcollison: Hmm, good to think about the race condition for sure. My hope would be the export account update gets pushed first but I hear what you are saying.

matthiashanel: > I don't think if an import fails the account should be evicted at all. That would mean the user can not connect at all. We would want to make it easy for the importing account to debug though.

The problem I see with that is noticing that smth is wrong in the first place. 
This particular error manifests itself as missing messages. 
Events associated with disconnect/reconnect are better observable than the lack of messages.

derekcollison: Are we comfortable though with an account being suspended due to a bad import until said import is removed or fixed and the account re-pushed? That sounds very dangerous to me and violates path of least surprise.

matthiashanel: We don't expect this condition to happen a lot, so as of now we keep trying. 
Do you think this is sufficient for now?


> Are we comfortable though with an account being suspended due to a bad import until said import is removed or  fixed and the account re-pushed? That sounds very dangerous to me and violates path of least surprise.

The account wouldn't be suspended. Clients would be able to do everything while they are connected. 
Every once in a while, they would be disconnected with a close reason account import broken.
Perhaps we could only close clients with subscriptions for the import, as well as the ones publishing to it.
However, this seems to be a discussion to be had when we get there.

derekcollison: I don't think we should disconnect clients at all. We could keep trying to resolve but feels like we should be smarter. For instance we should know if an account has a broken import on it, and when we get an export update go try to reload the import account that was broke again then.

matthiashanel: I take this as don't use timers then

derekcollison: I would not auto try to resolve TBH, would pick concrete event that would cause us to try to resolve TBH.

matthiashanel: We went down this route as we do not expect adding imports/exports happens often.
So automatically retrying once a second per initial failure appeared more straight forward than adding and maintaining data structures to track the issue.
An edge case to consider for an event driven approach is the update of the exporting account never happening. (say the import was wrong and gets removed)

derekcollison: Do we have tests that show we get an import account before and export account update and produce those failures?

I don't think we should keep trying to reload or disconnect existing clients due to import failure. Just my $0.02.

kozlovic: I will let Matthias comment, but he has been investigating an issue with a user that shows interest drop that has to do with import/export and accounts push. He did add tests to this PR. I do think the PR is fine as it is, the only thing we could do is stop trying after some time, but just giving up without closing clients seem wrong because then users would have the interest-loss issue which is difficult to debug as opposed to get the client disconnected.

derekcollison: If I import a service and someone revokes my token, I don't think that should have all my connections cut off. I agree we need more tools for end users to be able to identify the issue/

matthiashanel: I have been contacted about the same symptom multiple times.
Because there was such a delay between dropping the subscriptions and investigating, logs turned out to be of no use.
I realized what might be going on based on narrowing down the situation to a single server + accounts with imports and then reading code.
This PR contains unit tests for fetch failing as well as unfortunate update reordering.
I'm very interested in this becoming part of the next release just so I know if I'm on the right track. 
After-all this still may be another issue altogether.

I just pushed two commits.
First one keeps track of the account names experiencing the issue.
After this was done (passed all tests) I realized that ranging over all accounts is smth that is done, and could slim down the necessary state.

So I'd say have a look at the last commit. 
Only if that does not fit the bill, have a look at the previous one.

derekcollison: ok I think we are closer. Could you squash these into 1 and ping me and will take another look all together. Thx!

server/accounts.go
```diff
@@ -2623,6 +2632,29 @@ func (s *Server) UpdateAccountClaims(a *Account, ac *jwt.AccountClaims) {
...
+			acc := value.(*Account)
+			acc.mu.RLock()
+			incomplete := acc.incomplete
+			// Must use jwt in account or risk failing on fetch
+			// This jwt may not be the same that caused exportingAcc to be in incompleteAccExporterMap
+			claimJWT := acc.claimJWT
+			acc.mu.RUnlock()
+			if incomplete {
+				if accClaims, _, err := s.verifyAccountClaims(claimJWT); err == nil {
+					s.UpdateAccountClaims(acc, accClaims)
```
kozlovic, line 2648: This is the function from which we are inside. So we will possibly recurse. Do you see any problem with that?

kozlovic (changes requested):
some little comment rephrasing and wondering about recursive calls.

server/server.go
```diff
@@ -219,8 +219,8 @@ type Server struct {
 	// Websocket structure
 	websocket srvWebsocket
 
-	// exporting account name -> (incomplete) importing account
-	incompleteAccMap map[string]map[string]struct{}
+	// exporting account name the importer experienced issues
```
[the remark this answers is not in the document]
matthiashanel: fixed

server/accounts.go
matthiashanel: Not really.
sync.Map has no problem with it. 
We are also only recursive for cases where the account is incomplete. 
This happens at the tail end of UpdateAccountClaims, so I'm not worried about not being fully set up.
In the recursive situation, an incomplete account can turn complete and thus not be considered any longer.
At any given time I believe the number of incomplete accounts to be small.
#inclomplete accounts * # accounts.

**I guess this comes back to: is ranging over all accounts appropriate at all?**
If there is one persistently bad account every time we update the exporter we will visit all accounts.
Say the token expired, because we don't get rid of the clients or the account, we will range over all accounts.

kozlovic: Valid concern and - although we do range over accounts, for instance on reload - if we plan to have many, many accounts, that could become an issue. But in order to have an impact (that is sync.map range being slow) I am guessing that it would require a LOT of accounts, no?

matthiashanel: I would say we need a great many accounts. 
And then, this probably happens not that often as we remove the account from incompleteAccExporterMap.

I also want to point out that this scheme only works as long as we can continue to tolerate this situation.
update account A, fail to fetch import on account B. 
While account B has no clients, account A won't be updated.

server/accounts.go
matthiashanel: Done

server/accounts.go
```diff
@@ -2574,6 +2582,7 @@ func (s *Server) UpdateAccountClaims(a *Account, ac *jwt.AccountClaims) {
 		}
 	}
 	a.defaultPerms = buildPermissionsFromJwt(&ac.DefaultPermissions)
+	a.incomplete = incomplete
 	a.mu.Unlock()
```
derekcollison, line 2586: nit but maybe here check incomplete and do the ExporterMap assignment once here.

server/accounts.go
```diff
@@ -2623,6 +2632,30 @@ func (s *Server) UpdateAccountClaims(a *Account, ac *jwt.AccountClaims) {
 			}
 		}
 	}
+
+	if _, ok := s.incompleteAccExporterMap.Load(old.Name); ok {
+		s.incompleteAccExporterMap.Delete(old.Name)
+		s.accounts.Range(func(key, value interface{}) bool {
```
derekcollison, line 2638: Hard to follow really what is going on here, comment might be good.

derekcollison (approved):
Some minor comments but LGTM.

server/accounts.go
matthiashanel: Not sure I follow. When we iterated over all imports and did NOT run into an error, it does not matter what value a.incomplete had. At this point it is complete.

Since each import can refer to a different export there is not really such a thing as once.
I assume assigning exportermap once is about setting it together with a.incomplete. (will change that)

server/accounts.go
matthiashanel: added

server/accounts.go
```diff
@@ -2647,6 +2647,10 @@ func (s *Server) UpdateAccountClaims(a *Account, ac *jwt.AccountClaims) {
 			if incomplete && name != old.Name {
 				if accClaims, _, err := s.verifyAccountClaims(claimJWT); err == nil {
 					s.UpdateAccountClaims(acc, accClaims)
+					// old.Name was deleted before ranging over accounts
```
kozlovic, line 2650: I am uneasy with this approach of doing it in place (as opposed to timer based). Say that we had A import from B that import from C, and all this chain is broken (meaning processing of A import failed, and B import failed).
Say that we are processing A at the moment (so old.Name == A). There is no guarantee of ordering doing the accounts.Range(). So say that the first we get is "B" in the Range() and it is incomplete (and name != A) then we recursively call this function. Now we are back at line 2636 and we have "B" marked as incomplete. We range and now say that the first in the Range() is "A". It is still incomplete and we now recursively call this function again, and we happen to going from "A" being given first in Range(), to "B", would that not end-up panic'ing with a stack overflow?

matthiashanel: permanently broken circular imports did the trick. fixed

server/jwt_test.go
```diff
@@ -1911,6 +1911,110 @@ func TestAccountURLResolverFetchFailurePushReorder(t *testing.T) {
...
+	ncB := natsConnect(t, sA.ClientURL(), createUserCreds(t, nil, expkp))
+	defer ncB.Close()
+	sysc := natsConnect(t, sA.ClientURL(), createUserCreds(t, nil, syskp))
+	defer sysc.Close()
+	// push accounts
+	natsPub(t, sysc, fmt.Sprintf(accUpdateEventSubj, imppub), []byte(impjwt))
+	natsPub(t, sysc, fmt.Sprintf(accUpdateEventSubj, exppub), []byte(expjwt))
+	sysc.Flush()
+	importErrCnt := 0
+	for {
```
kozlovic, line 2002: This is going to be long test (2 or 4 seconds, etc..). Replace with this:
```
get_errors_loop:
	for {
		select {
		case line := <-l.dbgCh:
			if strings.HasPrefix(line, "Error adding stream import to account") {
				fmt.Printf("@@IK: here: %q\n", line)
				importErrCnt++
				if importErrCnt == 6 {
					break get_errors_loop
				}
			}
		case <-time.After(2 * time.Second):
			t.Fatal("Did not get all 6 debug traces")
		}
	}
	// Make sure we don't get more..
	for {
		select {
		case line := <-l.dbgCh:
			if strings.HasPrefix(line, "Error adding stream import to account") {
				t.Fatalf("Got more error lines than expected: %q", line)
			}
		case <-time.After(250 * time.Millisecond):
			// OK
			return
		}
	}
```

kozlovic: or, as discussed, make sure you create the timer outside of for loop and reduce to 500ms.

matthiashanel: done

Issue #1580, opened 2020-09-02
JetStream: durable restart resends last stored message


kozlovic:
#### Versions of `nats-server` and affected client libraries used:
- From main, rev: d7946169454a724c7aab47d4aa301a5983811ad2

#### OS/Container environment:
- All

#### Steps or code to reproduce the issue:
- Create a stream:
```
&StreamConfig{
   Name:      streamName,
   Storage:   FileStorage, // same if it was mem storage
   Subjects:  []string{"foo"},
   Retention: InterestPolicy,
}
```
- Add a durable consumer with this config:
```
&ConsumerConfig{
   DeliverSubject: inbox,
   DeliverPolicy:  DeliverNew,
   AckPolicy:      AckNone,
}
```
- While the consumer is running send 2 messages M1 and M2, they are received. 
- Stop the consumer, which means delete the subscription on the given inbox (but do not delete the durable JS consumer).
- Send 2 more messages M3 and M4
- Restart the durable with a new inbox attached
- It will receive M2, M3 and M4.
- You can now stop and restart it, it will keep receiving M4.

#### Expected result:
- Receive only M3 and M4 after the first restart, nothing after the second.

#### Actual result:
- The last stored message is always redelivered.

wallyqs:
Fixed via #1581

PR #1581, opened 2020-09-03
JetStream fixes for InterestPolicy retention

derekcollison:
This fixes both #1580 and https://github.com/nats-io/jetstream/issues/314

/cc @nats-io/core

### DIFF
--- a/server/consumer.go
+++ b/server/consumer.go
@@ -1735,7 +1735,25 @@ func (o *Consumer) stop(dflag, doSignal, advisory bool) error {
 	mset.unsubscribe(ackSub)
 	mset.unsubscribe(reqSub)
 	delete(mset.consumers, o.name)
+	rp := mset.config.Retention
 	mset.mu.Unlock()
+
+	// We need to optionally remove all messages since we are interest based retention.
+	if dflag && rp == InterestPolicy {
+		var seqs []uint64
+		o.mu.Lock()
+		for seq := range o.pending {
+			seqs = append(seqs, seq)
+		}
+		o.mu.Unlock()
+		// Sort just to keep pending sparse array state small.
+		sort.Slice(seqs, func(i, j int) bool { return seqs[i] < seqs[j] })
+		for _, seq := range seqs {
+			if !mset.checkInterest(seq, o) {
+				mset.store.RemoveMsg(seq)
+			}
+		}
+	}
 
 	// Make sure we stamp our update state
 	if !dflag {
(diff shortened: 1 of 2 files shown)


ripienaar:
Maybe not a direct result of this PR but this is a bit weird,

```json
  "state": {
    "messages": 0,
    "bytes": 0,
    "first_seq": 3,
    "first_ts": "1970-01-01T00:00:00Z",
    "last_seq": 2,
    "last_ts": "2020-09-03T11:57:53.7182588Z",
    "consumer_count": 0
  }
```

we have 3 messages (now deleted) and `first_ts` is 1970, I get why this is but it seems weird while last_ts is the last received message.  I know we spoke about these recently too, hard to know what the right behaviour is here

Otherwise this PR works for me

kozlovic (approved):
LGTM and verified that it would resolve issue that I saw.

kozlovic:
@derekcollison Unrelated to this PR obviously, but following flapper:
```
=== RUN   TestLeafNodeDaisyChain
    leafnode_test.go:3330: Expected a response
--- FAIL: TestLeafNodeDaisyChain (1.10s)
```
I believe could be resolved with adding:
```
checkSubInterest(t, s3, "$G", "ngs.usage", time.Second)
```
to `test/leafnode_test.go:3323`.

derekcollison:
Will fix leafnode test.

derekcollison:
R.I. regarding last ts, understand but I feel we really need state to hold last message received at all times, especially when the stream has been emptied vie expiration or purge.

kozlovic (approved):
LGTM

ripienaar:
I agree on the last, but here first behaviour seems at odds then

ripienaar:
I also just noticed that if I add a consumer messages are kept - but if I delete the only consumer messages remain

```json
{
  "config": {
    "name": "INTEREST",
    "subjects": [
      "js.in.interest"
    ],
    "retention": "interest",
    "max_consumers": -1,
    "max_msgs": -1,
    "max_bytes": -1,
    "max_age": 0,
    "max_msg_size": -1,
    "storage": "file",
    "discard": "old",
    "num_replicas": 1,
    "duplicate_window": 120000000000
  },
  "state": {
    "messages": 1,
    "bytes": 45,
    "first_seq": 2,
    "first_ts": "2020-09-03T14:57:14.966587902Z",
    "last_seq": 2,
    "last_ts": "2020-09-03T14:57:14.966587902Z",
    "consumer_count": 0
  }
```

derekcollison:
ok let me look into that real quick and add a test for that. Removing the consumer removes interest so should have the messages removed.

kozlovic (approved):
LGTM

ripienaar:
LGTM but I can only properly test tomorrow

PR #1587, opened 2020-09-04
Updating jetstream account settings from jwt

matthiashanel:

I'm setting the account to incomplete so it is set for when we start to report things like that.
Have to set srv in update as on connect update is called before register. JS depends on that pointer existing.

### DIFF
--- a/server/jwt_test.go
+++ b/server/jwt_test.go
@@ -3946,10 +3946,9 @@ func TestJWTJetStreamLimits(t *testing.T) {
 		system_account: %s
     `, ojwt, dir, sysPub)))
 	defer os.Remove(conf)
-	opts := LoadConfig(conf)
-	s := RunServer(opts)
+	s, opts := RunServerWithConfig(conf)
 	defer s.Shutdown()
-
+	port := opts.Port
 	updateJwt(s.ClientURL(), sysCreds, sysPub, sysJwt)
 	sys := natsConnect(t, s.ClientURL(), nats.UserCredentials(sysCreds))
 	expect_InfoError(sys)
@@ -3985,7 +3984,6 @@ func TestJWTJetStreamLimits(t *testing.T) {
 	validate_limits(c, limits2)
 	updateJwt(s.ClientURL(), sysCreds, aPub, aJwtLimitsExceeded)
 	validate_limits(c, limits2)
-
 	s.Shutdown()
 	conf = createConfFile(t, []byte(fmt.Sprintf(`
 		listen: %d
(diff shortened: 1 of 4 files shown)


server/jwt_test.go
```diff
@@ -3836,3 +3836,194 @@ func TestJWTNoOperatorMode(t *testing.T) {
...
+		operator: %s
+		resolver: {
+			type: full
+			dir: %s
+		}
+		system_account: %s
+    `, opts.Port, ojwt, dir, sysPub)))
+	defer os.Remove(conf)
+	opts = LoadConfig(conf)
+	opts.NoLog = false
```
kozlovic, line 4019: These are causing logs to appear, probably was meant to debug the test but should be removed now?
If so, go back to RunServerWithConfig()

server/jwt_test.go
```diff
@@ -3836,3 +3836,194 @@ func TestJWTNoOperatorMode(t *testing.T) {
...
+	opts.Debug = true
+	opts.Trace = true
+	s = RunServer(opts)
+	c.Flush() // force client to discover the disconnect
```
kozlovic, line 4023: forgot to put a `defer s.Shutdown()` here, which will cause some go routines to be left running..

server/jwt_test.go
```diff
@@ -3836,3 +3836,194 @@ func TestJWTNoOperatorMode(t *testing.T) {
...
+		jetstream: {max_mem_store: 10Mb, max_file_store: 10Mb}
+		operator: %s
+		resolver: {
+			type: full
+			dir: %s
+		}
+		system_account: %s
+    `, ojwt, dir, sysPub)))
+	defer os.Remove(conf)
+	opts := LoadConfig(conf)
```
kozlovic, line 3949: Could replace these 2 calls (load+run) with:
```
s, opts := RunServerWithConfig(conf)
```

server/jwt_test.go
```diff
@@ -3836,3 +3836,194 @@ func TestJWTNoOperatorMode(t *testing.T) {
...
+		jetstream: {max_mem_store: 20Mb, max_file_store: 20Mb}
+		operator: %s
+		resolver: {
+			type: full
+			dir: %s
+		}
+		system_account: %s
+    `, opts.Port, ojwt, dir, sysPub)))
+	defer os.Remove(conf)
+	s = RunServer(LoadConfig(conf))
```
kozlovic, line 4001: Same than above. And if you don't need options, you can always do `s, _ = RunServerWithConfig(conf)`

kozlovic (changes requested):
Forgot a defer in the test and could simplify some calls.

server/jwt_test.go
matthiashanel: will change

server/jwt_test.go
matthiashanel: will change

server/jwt_test.go
matthiashanel: the defer is still around from the original start, since I explicitly shut that one down.

server/jwt_test.go
matthiashanel: yes, will remove the logs.

server/jwt_test.go
kozlovic: no, because the original defer as captured the pointer, so it applies to the pointer value of that first server.

kozlovic (approved):
LGTM

derekcollison (approved):
LGTM

PR #1590, opened 2020-09-08
[ADDED] Support for route hostname resolution

### LINKED ISSUE #1586: Support DNS discovery of DNS names on routes
## Feature Request
Support resolving DNS names for routes and using resolved ips, excluding self ip.

#### Use Case:
Boostraping a NATS cluster using Consul 

#### Proposed Change:
Resolve DNS names instead of directly calling net.DialTCP() for routes

Environments that use DNS as service discovery, including Consul clusters.

kozlovic:
We previously simply called DialTimeout() on a route's url when
soliciting. If it resolved to the IP of the host, it would create
a route to self, which server detects, but then would not try again
with other IPs that would have allowed to form a cluster with
other servers running on the other IPs.

This PR keeps track of local IPs + cluster port and exclude them
from the list of IPs returned by LookupHost API. This even prevent
solicitation of routes to self. Only non-local IPs will be tried.

Resolves #1586

### DIFF
--- a/server/client.go
+++ b/server/client.go
@@ -4207,7 +4207,7 @@ func (c *client) reconnect() {
 			srv.Debugf("Not attempting reconnect for solicited route, already connected to \"%s\"", rid)
 			return
 		} else if rid == srv.info.ID {
-			srv.Debugf("Detected route to self, ignoring \"%s\"", rurl)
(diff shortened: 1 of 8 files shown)


derekcollison (approved):
LGTM

PR #1592, opened 2020-09-09
Fix tests for Go 1.15

wallyqs:
Starting at Go 1.15 now the TLS connections from Go clients will fail with an error such as:

```
x509: certificate relies on legacy Common Name field, use SANs or temporarily enable Common Name matching with GODEBUG=x509ignoreCN=0
```

when the certificates use the Common Name field.  As a workaround to enable running tests with Go 1.15 (and also to demonstrate the workaround for possibly impacted nats.go users in a commit), we disable the client verification for the TLS map permissions.

More info on this at: https://github.com/golang/go/issues/39568

/cc @nats-io/core

### DIFF
--- a/.github/workflows/go-test.yaml
+++ b/.github/workflows/go-test.yaml
@@ -5,7 +5,7 @@ jobs:
   test:
     strategy:
       matrix:
-        go: [1.13, 1.14]
+        go: [1.14, 1.15]
 
     env:
       GOPATH: /home/runner/work/nats-server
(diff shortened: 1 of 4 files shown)


wallyqs:
Closing and instead will update the certs since the `VerifyConnection` API which is used for this workaround got introduced as well in Go 1.15 (https://github.com/golang/go/issues/36736)

wallyqs:
Closing, for reference, green build with Go 1.15 here: https://travis-ci.org/github/nats-io/nats-server/jobs/725449002

PR #1593, opened 2020-09-09
Fixes for Go1.15 

wallyqs:
Generating new TLS certs that have the proper SAN fields filled in since now the Go client is more strict about the certs it uses. Starting at Go 1.15 now the TLS connections from Go clients will with an error such as:

```
x509: certificate relies on legacy Common Name field, use SANs or temporarily enable Common Name matching with GODEBUG=x509ignoreCN=0
```

when the certificates use the `Common Name` field. More info on that here: https://github.com/golang/go/issues/39568

```sh
# To generate the certs
openssl req -newkey rsa:2048 -nodes -keyout client-$CLIENT_ID.key -subj "/C=US/ST=CA/L=Los Angeles/O=NATS/OU=NATS/CN=localhost/DC=foo1/DC=foo2" -out client-$CLIENT_ID.csr -extensions EXT -config <( \
    printf "[dn]\nDC=foo2\nDC=foo1\nCN=localhost\nOU=NATS\nO=NATS\nL=Los Angeles\nST=CA\nC=US\n[req]\ndistinguished_name = dn\n[EXT]\nsubjectAltName=DNS:localhost\nkeyUsage=digitalSignature\nextendedKeyUsage=serverAuth")
openssl x509 -req -extfile <(printf "subjectAltName=DNS:localhost,DNS:example.com,DNS:www.example.com") -days 365 -in client-$CLIENT_ID.csr -CA ca.pem -CAkey ca.key -CAcreateserial -out client-$CLIENT_ID.pem
```

 - [X] Tests added


server/auth.go
```diff
@@ -682,6 +682,7 @@ func checkClientTLSCertSubject(c *client, fn tlsMapAuthFn) bool {
 				return true
 			}
 		}
+		fallthrough
```
wallyqs, line 685: This bugfix also included, in case a subjectAlternativeName was present in the cert, then URIs (e.g SVID SPIFFE auth) would not have been attempted.

server/config_check_test.go
```diff
@@ -1110,7 +1110,7 @@ func TestConfigCheck(t *testing.T) {
 			config: `
 				lame_duck_grace_period: abc
 			`,
-			err:       errors.New(`error parsing lame_duck_grace_period: time: invalid duration abc`),
+			err:       errors.New(`error parsing lame_duck_grace_period: time: invalid duration`),
```
wallyqs, line 1113: In Go 1.15 time.Parse errors are now within quotes, so changed to match a substring only

test/configs/certs/rdns/ca.key
```diff
@@ -1,27 +1,27 @@
 -----BEGIN RSA PRIVATE KEY-----
```
wallyqs, line 1: Gist on how to generate these certs here: https://gist.github.com/wallyqs/18b241cf3eb7ae785dd09a9fbd40301c

kozlovic (approved):
LGTM

wallyqs:
For reference, TL;DR; about Go 1.15 CN and SAN changes: other than the tests there are no impacts for the server from this, but Go 1.15 users of the NATS client may run into issues when using their self signed certs. If that is the case, then can follow the instructions from this comment:
https://github.com/golang/go/issues/39568#issuecomment-671424481

derekcollison:
Do we need to update Go client as well?

ripienaar:
Just to repeat something from slack - we need to be careful that performing these updates and releasing the server would not result in people with existing certificates suddenly failing.  So a go client user with a cert without SANs would fail validation, this is not something we can really do in a non major change I think :( The go 1.15 change is really annoying

philpennock (approved):
I agree with what RI said.  This change switches us to building deploys with Go 1.14 instead of 1.13 but we need to sort out how we communicate to customers the need for fixes before we switch to official releases built with Go 1.15.

We might need to write a small test tool which folks can try, to check things work for them?

philpennock:
Within Synadia, I've created a calendar reminder for the 1st Tuesday in September to cycle these.  It might not be enough if we have to stick to 365 days, not, say, 372.  It will do for 2021.  If we cycle before then, we can update the reminder.  I think one spam per year if we forget about the reminder when regenerating is fine, and I'd rather get the reminder than not.

wallyqs:
Thanks @philpennock good point... Maybe should recreate the certs so that they are for 5 years?

philpennock:
I think that's fighting a trend for shorter certs in the broader ecosystem.  Browsers are about to cap lifetime at 1 year and I don't want us to be screaming at Go 1.17 doing something like capping the max lifetime at 1 year and needing a lot of tuning to override it.

Perhaps we should consider having a cert build tool/script invoked at the start of the test run to generate certificates for the test run, so that we don't need to check keys and certs into the code-base and can rely upon certs always being current for a given run?  That would also let us more easily add tests around freshly-expiring certs, etc.  I think that should be a follow-up PR to this one, if we go that route, rather than blocking this PR.

wallyqs:
Thanks, ok so I will leave the certs as is then and we could have an action that makes a PR to regenerate the certs at some point. 
Just to clarify, there are no impacts to clients by merging this PR since it is only fixing the tests and building the server with Go 1.15 does not affect clients either.  The errors come from the `nats.go` client when calling `VerifyConnection`, so for `nats.go` client users I will create a separate doc page to troubleshoot in case they may run into issues.

PR #1594, opened 2020-09-09
Added an allowed connection type filter for users

kozlovic:
Users and NKey users will now have the option to specify a list
of allowed connection types.

This will allow for instance a certain user to be allowed to
connect as a standard NATS client, but not as Websocket, or
vice-versa.

This also fixes the websocket auth override. Indeed, with
the original behavior, the websocket users would have been bound
to $G, which would not work when there are accounts defined, since
when that is the case, no app can connect/bind to $G account.


derekcollison (approved):
LGTM

kozlovic:
@derekcollison I have completely changed the approach with a list of connection types allowed for given user/nkey. We updated the JWT library to provide this new field in the JWT and some constants.

derekcollison (approved):
LGTM

matthiashanel (commented):
The change looks good. 
I wonder if we should have a test in jwt_test.go that tests using a new type in a jwt and make sure the server rejects it.

kozlovic:
@matthiashanel Thanks for that because it made me realize that in the case of JWT coming with an unknown type, we should not automatically fail the connection. See the big comment that I added in auth.go and let me know if you agree or not.

matthiashanel (commented):
LGTM

matthiashanel (approved):
LGTM

kozlovic:
@matthiashanel Sorry, I added yet another commit: do validation of options for users that embed NATS Server in their app and would configure Users/NKeys directly. (I will squash all those commits at time of merge)

matthiashanel (approved):
LGTM

PR #1595, opened 2020-09-11
Fix dirstore code and speed up some tests

kozlovic:
When using Unix() time, since it is number of seconds, it is better
to round up the time before adding a ttl. Trying to shorten some
of the tests showed that in some cases a file was removed too early.
This was because the computed expiration with ttl fell in the same
second, so the file was removed prematurely.

So anywhere where we used to do: time.Now().Addd(ttl).Unix(), I
changed to time.Now().Round(time.Second).Add(ttl).Unix().

I was able to reduce the time of TestTTL from 21 seconds down to
less than 5. TestExpiration was also shorten.

### DIFF
--- a/server/dirstore.go
+++ b/server/dirstore.go
@@ -566,7 +566,7 @@ func (pq *expirationTracker) updateTrack(publicKey string) {
 		i := e.Value.(*jwtItem)
 		if pq.ttl != 0 {
 			// only update expiration when set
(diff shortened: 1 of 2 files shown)


derekcollison (approved):
LGTM

matthiashanel (approved):
LGTM

PR #1598, opened 2020-09-16
Tie account in system event subject to data and clean up subject names

matthiashanel:
Lookup is a request.
STATSZ was duplicated so it falls into this scheme: 
$SYS.REQ.SERVER.<ID>.<OPERATION> and $SYS.REQ.SERVER.<ID>.PING.<OPERATION>.
The original SYS.REQ.SERVER.<ID>.PING is kept for backwards compatibility. (reason for change in events_test.go)

As for update I didn't go all in on avoiding parsing the jwt twice as it' not a frequent operation and avoids code changes.

There are more changes to come in that area but these where the straight forward ones.

### DIFF
--- a/server/accounts.go
+++ b/server/accounts.go
@@ -2937,7 +2937,8 @@ func (dr *DirAccResolver) Start(s *Server) error {
 		}
 	}
 	const accountPackRequest = "$SYS.ACCOUNT.CLAIMS.PACK"
-	const accountLookupRequest = "$SYS.ACCOUNT.*.CLAIMS.LOOKUP"
+	const accountLookupRequest = "$SYS.REQ.ACCOUNT.*.CLAIMS.LOOKUP"
+	const accountLookupTokens = 6
 	packRespIb := s.newRespInbox()
 	// subscribe to account jwt update requests
 	if _, err := s.sysSubscribe(fmt.Sprintf(accUpdateEventSubj, "*"), func(_ *subscription, _ *client, subj, resp string, msg []byte) {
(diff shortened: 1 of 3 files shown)


server/accounts.go
```diff
@@ -2937,7 +2937,8 @@ func (dr *DirAccResolver) Start(s *Server) error {
 		}
 	}
 	const accountPackRequest = "$SYS.ACCOUNT.CLAIMS.PACK"
-	const accountLookupRequest = "$SYS.ACCOUNT.*.CLAIMS.LOOKUP"
+	const accountLookupRequest = "$SYS.REQ.ACCOUNT.*.CLAIMS.LOOKUP"
```
kozlovic, line 2940: Originally this was used only in this function, but now I wonder if we should move then to events.go (where you have defined number of tokens and index of account).
Also, we could make it a subject with the `%s` token, like `const accountLookupRequest = "$SYS.REQ.ACCOUNT.%s.CLAIMS.LOOKUP"`.
Here, to subscribe you would sprintf with this constant and pass "*" for the "%s".

server/accounts.go
```diff
@@ -3070,7 +3071,7 @@ func (dr *CacheDirAccResolver) Fetch(name string) (string, error) {
 		return "", ErrNoAccountResolver
 	}
 	respC := make(chan []byte, 1)
-	accountLookupRequest := fmt.Sprintf("$SYS.ACCOUNT.%s.CLAIMS.LOOKUP", name)
+	accountLookupRequest := fmt.Sprintf("$SYS.REQ.ACCOUNT.%s.CLAIMS.LOOKUP", name)
```
kozlovic, line 3074: Here, I would use a different variable name and sprintf with the constant referenced earlier.

kozlovic (changes requested):
Would make this a constant subject defined in events.go and make use of it to create the subscription on "*" or lookup on specific account.

server/accounts.go
matthiashanel: will make these changes

matthiashanel:
@kozlovic , incorporated your suggestions. also renamed claims.pack as it is a request as well. Moved the constant for it into system events just so we have one place where these can be found.

kozlovic (approved):
LGTM

PR #1599, opened 2020-09-17
Adding new subscription for account conns

matthiashanel:
This fits better with similar events
New subject is $SYS.ACCOUNT.%s.SERVER.CONNS
Old subject remains for backwards compatibility

For backwards compatibility 2 subjects are used.
The utility to check if interest exists comes in handy here.

Going forward I'd add those to other system events as well, just thought this is a good first show case.

### DIFF
--- a/server/events.go
+++ b/server/events.go
@@ -571,11 +571,16 @@ func (s *Server) initEventTracking() {
 	s.sys.inboxPre = subject
 	// This is for remote updates for connection accounting.
 
-	for _, subj := range []string{accConnsEventSubjOld, accConnsEventSubjNew, connsRespSubj} {
+	for _, subj := range []string{accConnsEventSubjOld, accConnsEventSubjNew} {
 		subject = fmt.Sprintf(subj, "*")
 		if _, err := s.sysSubscribe(subject, s.remoteConnsUpdate); err != nil {
 			s.Errorf("Error setting up internal tracking for %s: %v", subject, err)
 		}
+	}
+	// This will be for responses for account info that we send out.
+	subject = fmt.Sprintf(connsRespSubj, s.info.ID)
+	if _, err := s.sysSubscribe(subject, s.remoteConnsUpdate); err != nil {
+		s.Errorf("Error setting up internal tracking: %v", err)
 	}
 	// Listen for broad requests to respond with account info.
 	subject = fmt.Sprintf(accConnsReqSubj, "*")
--- a/server/events.go
+++ b/server/events.go
@@ -1070,26 +1070,16 @@ func (s *Server) sendAccConnsUpdate(a *Account, subj ...string) {
 	if !s.eventsEnabled() || a == nil {
 		return
 	}
-	// Build event with account name and number of local clients and leafnodes. (if there is any interest in receiving)
-	var m *AccountNumConns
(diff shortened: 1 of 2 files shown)


server/events.go
```diff
@@ -569,14 +570,12 @@ func (s *Server) initEventTracking() {
...
-	if _, err := s.sysSubscribe(subject, s.remoteConnsUpdate); err != nil {
-		s.Errorf("Error setting up internal tracking: %v", err)
-	}
-	// This will be for responses for account info that we send out.
-	subject = fmt.Sprintf(connsRespSubj, s.info.ID)
-	if _, err := s.sysSubscribe(subject, s.remoteConnsUpdate); err != nil {
-		s.Errorf("Error setting up internal tracking: %v", err)
+
+	for _, subj := range []string{accConnsEventSubjOld, accConnsEventSubjNew, connsRespSubj} {
+		subject = fmt.Sprintf(subj, "*")
```
kozlovic, line 575: You are changing the subscription subject here for `connsRespSub`. Used to be on `s.info.ID` but you all 3 are on `*`. I am not sure this is right to have this server listen to all responses, should be only for itself.

server/events.go
```diff
@@ -1067,23 +1066,32 @@ func (s *Server) sendLeafNodeConnectMsg(accName string) {
...
-	m := AccountNumConns{
-		Account:    a.Name,
-		Conns:      a.numLocalConnections(),
-		LeafNodes:  a.numLocalLeafNodes(),
-		TotalConns: a.numLocalConnections() + a.numLocalLeafNodes(),
+	// Build event with account name and number of local clients and leafnodes. (if there is any interest in receiving)
+	var m *AccountNumConns
+	for _, sub := range subj {
+		if s.sys != nil {
+			if sac := s.sys.account; sac != nil && !sac.SubscriptionInterest(sub) {
```
kozlovic, line 1077: Careful: SubscriptionInterest(), although public, was really added for tests. For instance, it does not support wildcards (see comments on that function). That being said, as far as I can tell, you are calling this function only with literals, right?

kozlovic: Other issue I see with suppressing interest, is that this is just looking for interest registered in the account. But we have situations (for instance case for gateways) where interest is not stored in an account, but instead of gw structures. Asking gw to send a message will then check if it has to be done or not (based on known no-interest, optimistic mode, etc..). 
So I am not sure, but you may be breaking things here?

kozlovic (commented):
Not sure about that one...

server/events.go
matthiashanel: @kozlovic dang, I did not consider gateways. Basically I won't be able to use SubscriptionInterest or anything like it unless gateways will change. I'll remove the check for interest and basically publish the same message twice.

kozlovic: Is that an issue to publish twice? (except for the "waste")

server/events.go
kozlovic: What about this concern?

server/events.go
matthiashanel: I don't see one.

server/events.go
matthiashanel: totally missed that one, thanks for catching that

kozlovic (approved):
LGTM

PR #1601, opened 2020-09-21
Do not ack to AckNxt

ripienaar:
This causes nil messages mid content stream and is confusing,
the test for this is amended to test the content received for
validity - where previously the test assumed any content is
good content but in fact it only received half the content.

Removing this behaviour until we can design this properly

/cc @nats-io/core

### DIFF
--- a/server/consumer.go
+++ b/server/consumer.go
@@ -698,12 +698,15 @@ func (o *Consumer) sendAckReply(subj string) {
 func (o *Consumer) processAck(_ *subscription, _ *client, subject, reply string, msg []byte) {
 	sseq, dseq, dcount, _ := o.ReplyInfo(subject)
 
+	var skipAckReply bool
+
 	switch {
 	case len(msg) == 0, bytes.Equal(msg, AckAck), bytes.Equal(msg, AckOK):
 		o.ackMsg(sseq, dseq, dcount)
 	case bytes.Equal(msg, AckNext):
 		o.ackMsg(sseq, dseq, dcount)
 		o.processNextMsgReq(nil, nil, subject, reply, nil)
+		skipAckReply = true
 	case bytes.Equal(msg, AckNak):
 		o.processNak(sseq, dseq)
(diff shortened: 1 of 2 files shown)


server/consumer.go
```diff
@@ -698,12 +698,15 @@ func (o *Consumer) sendAckReply(subj string) {
...
 
+	var skipAckReply bool
+
 	switch {
 	case len(msg) == 0, bytes.Equal(msg, AckAck), bytes.Equal(msg, AckOK):
 		o.ackMsg(sseq, dseq, dcount)
 	case bytes.Equal(msg, AckNext):
 		o.ackMsg(sseq, dseq, dcount)
 		o.processNextMsgReq(nil, nil, subject, reply, nil)
+		skipAckReply = true
```
derekcollison, line 709: I think we should send ack if no new messages can be sent, otherwise just send new message.

ripienaar: I strongly disagree, we should not send random no content messages into the payload stream.  If we need a control channel to advise about this kind of thing, you can see how awkward it makes the code and how easy it is to be thrown by this by your own test that silently passed when in fact it should have failed.

ripienaar: But whatever we choose, until we decide - we need to remove this bad behaviour

derekcollison: We need to decide how to resolve the conflicting behaviors. All of this will be handled properly by the helpers so not too worried about the raw behavior, we can solve for that.

ripienaar: There's no way to tell the difference between this nil reply and a potential real empty message in the stream, this behaviour is a bug that was introduced we really need to think about signaling more, but this as it is is not sufficient and will cause problems.  Can we please back it out until we think this through?

derekcollison: Fair enough, have not had enough time to think through a proper solution.

derekcollison (approved):
Will pull out for now.

PR #1602, opened 2020-09-21
Add $SYS.REQ.ACCOUNT.%s.CLAIMS.UPDATE

matthiashanel:
Old $SYS.ACCOUNT.%s.CLAIMS.UPDATE is kept for backwards compatibility.
The old name is in the same name space as events.
To be able to abuse this, an attacker needs to be in possession of the operator key as well.

In order to not introduce more version dependencies I decided to support old and new name for nats based account resolver as well.


kozlovic (approved):
LGTM

PR #1605, opened 2020-09-22
Expose SUBSZ and CONNZ monitoring endpoints for accounts

matthiashanel:

This is the low hanging fruit.
Take existing monitoring endpoints that allow to filter by account.
Expose these endpoints on an account specific subject. 
Extract account id from subject and set the account filter. (this will overwrite whatever other value may have been specified) 

I also moved CONNS into that setup loop.


kozlovic (approved):
LGTM

PR #1607, opened 2020-09-22
[FIXED] Prevent LeafNode loop detection on early reconnect

### LINKED ISSUE #1606: Possible Leafnode loop detected if soliciting server reconnects before accepting detects failure
## Defect

Latest from main branch: 1dd0c0666f1a910b461ca3b9dd33f2b9886cbb7e

N/A

Was observed by a user that had a router that would switch interfaces causing reconnects.
Otherwise, using a proxy that causes the soliciting side connection to be closed while the accepting side does not detect that failure right away.

#### Expected result:

The soliciting server should be able to reconnect as soon as it has detected that the previous connection was broken. The accepting server should not reject the reconnect as forming a loop.

#### Actual result:

The error "Loop detected" is reported and the soliciting server will be disconnected and attempt to reconnect after 30 seconds

kozlovic:
If the soliciting side detects the disconnect and attempts to
reconnect but the accepting side did not yet close the connection,
a "loop detected" error would be reported and the soliciting server
would not try to reconnect for 30 seconds.

Made a change so that the accepting server checks for existing
leafnode connection for the same server and same account, and if
it is found, close the "old" connection so it is replaced by
the "new" one.

Resolves #1606

### DIFF
--- a/server/leafnode.go
+++ b/server/leafnode.go
@@ -1035,6 +1035,9 @@ func (c *client) processLeafNodeConnect(s *Server, arg []byte, lang string) erro
 		return ErrWrongGateway
 	}
 
+	// Check for stale connection from same server/account
+	c.replaceOldLeafNodeConnIfNeeded(s, proto)
+
 	// Leaf Nodes do not do echo or verbose or pedantic.
 	c.opts.Verbose = false
 	c.opts.Echo = false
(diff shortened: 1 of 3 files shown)


server/leafnode.go
```diff
@@ -1068,6 +1071,42 @@ func (c *client) processLeafNodeConnect(s *Server, arg []byte, lang string) erro
...
+		}
+		ol.mu.Unlock()
+		if old != nil {
+			break
+		}
+	}
+	s.mu.Unlock()
+	if old != nil {
+		old.Warnf("Replacing connection from same server")
+		old.closeConnection(ReadError)
```
matthiashanel, line 1106: would StaleConnection be a more appropriate closedState?

derekcollison (approved):
LGTM

matthiashanel (commented):
Overall, change looks good to me. 
Only thing I'd change is the close reason. But I'll leave that up to you.

server/leafnode.go
kozlovic: I went from the principle that this is what would be reported in a normal disconnect, in the read loop, when getting an error on Read(). But I can change it if you think it makes more sense. (to me, stale is used when we close connection after sending max out pings without corresponding pongs).

matthiashanel: ok, that makes sense. LGTM

matthiashanel (approved):
LGTM

PR #1612, opened 2020-09-24
[Added] filtering by account to leafz and exposing this as per acc subject

matthiashanel:
On the monitoring endpoint /leafz specify ?acc=<account id>

### DIFF
--- a/server/monitor.go
+++ b/server/monitor.go
@@ -1765,8 +1765,13 @@ func (s *Server) Leafz(opts *LeafzOptions) (*Leafz, error) {
 	if len(s.leafs) > 0 {
 		lconns = make([]*client, 0, len(s.leafs))
 		for _, ln := range s.leafs {
-			if opts != nil && opts.Account != "" && ln.acc.Name != opts.Account {
-				continue
+			if opts != nil && opts.Account != "" {
+				ln.mu.Lock()
+				ok := ln.acc.Name == opts.Account
+				ln.mu.Unlock()
+				if !ok {
+					continue
+				}
 			}
 			lconns = append(lconns, ln)
 		}
--- a/server/monitor.go
+++ b/server/monitor.go
@@ -1739,7 +1739,8 @@ type Leafz struct {
 // LeafzOptions are options passed to Leafz
 type LeafzOptions struct {
 	// Subscriptions indicates that Leafz will return a leafnode's subscriptions
-	Subscriptions bool `json:"subscriptions"`
+	Subscriptions bool   `json:"subscriptions"`
+	Account       string `json:"account"`
 }
 
 // LeafInfo has detailed information on each remote leafnode connection.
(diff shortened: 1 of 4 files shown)


server/monitor.go
```diff
@@ -1764,6 +1765,9 @@ func (s *Server) Leafz(opts *LeafzOptions) (*Leafz, error) {
 	if len(s.leafs) > 0 {
 		lconns = make([]*client, 0, len(s.leafs))
 		for _, ln := range s.leafs {
+			if opts != nil && opts.Account != "" && ln.acc.Name != opts.Account {
```
kozlovic, line 1768: I think we should lock `ln` to access its `acc` no? (can it be a race with config reload?).
```
if opts != nil && opts.Account != "" {
    ln.mu.Lock()
    ok := ln.acc.Name == opts.Account
    ln.mu.Unlock()
    if !ok {
        continue
    }
}
```

server/monitor.go
```diff
@@ -1813,12 +1817,7 @@ func (s *Server) HandleLeafz(w http.ResponseWriter, r *http.Request) {
 	if err != nil {
 		return
 	}
-	var opts *LeafzOptions
-	if subs {
-		opts = &LeafzOptions{Subscriptions: true}
-	}
-
-	l, err := s.Leafz(opts)
+	l, err := s.Leafz(&LeafzOptions{subs, r.URL.Query().Get("acc")})
```
kozlovic, line 1820: Wonder if we should have a `decodeString()` since I see that you start to use that in few places.
We could return an error if not a valid option, but not sure we do that for say "subs".
Also, you could have:
```
acc, err := decodeString(w, r, "acc") // Or acc := r.URL.Query().Get("acc") if we don't want to check for "errors"
if err != nil {
    return
}
var opts *LeafzOptions
if subs || acc != "" {
    opts = &LeafzOptions{subs, acc}
}
...
```
I know, saving creating `LeafzOptions` is likely not going to make any difference, but..

kozlovic (changes requested):
Wonder if we should have locking when checking account, just to be on the safe side.

server/monitor.go
matthiashanel: Space wise this could even be better. (unless the compiler does tricks)
It's on the heap anyway. If you store the result of the lookup in acc, and later on in LeafzOptions, the pointer to the string could be on the heap twice. 
This avoids extra ifs at the cost of computing a pointer value. 

As for r.URL.Query().Get("acc"), we have decode functions in cases where we need to convert the string into some other type. The error then is about conversion issues. 
Here, we do not do a conversion. The returned value is the empty string, which means do not filter, or not, meaning only provide leaf nodes by this name.

kozlovic: I am ok with keeping this the way it is.

server/monitor.go
matthiashanel: agreed, will change that

matthiashanel: fixed

kozlovic (approved):
LGTM

PR #1614, opened 2020-09-24
[Fixed] race condition where account conns timer was disabled too soon

### LINKED ISSUE #1613: Race condition where account conns timer was disabled too soon
as a result remote servers did not observer the remote connection count of the server where the issue occurred go down to 0.

matthiashanel:
The connection count sent and the connection count used to determine if
the timer should be disabled could differ.

Also fixed issues in unit test triggering this behavior.
It did not check if remote connections where set to 0 prior to doing
more tests.

Fixes #1613 

When running with -race, tests kept failing until I increased the check time to 4 seconds.


kozlovic (approved):
LGTM. Good catch!

PR #1616, opened 2020-09-26
Adding fetch on missing jwt of full nats based resolver

matthiashanel:
Full nats based resolver sync within a cluster.
This functionality addresses syncing between cluster.

Fixing deadlock when more than one server responds to lookup.
Fixing 2 crashes when shutdown and pack happen at the same time.

As a side effect of also doing fetch in the full account resolver, functionality was moved from the caching account resolver.

### DIFF
--- a/server/accounts.go
+++ b/server/accounts.go
@@ -2892,11 +2892,16 @@ func (ur *URLAccResolver) Fetch(name string) (string, error) {
 // Resolver based on nats for synchronization and backing directory for storage.
 type DirAccResolver struct {
 	*DirJWTStore
+	*Server
 	syncInterval time.Duration
 }
 
 func (dr *DirAccResolver) IsTrackingUpdate() bool {
 	return true
+}
+
+func (dr *DirAccResolver) Reload() error {
+	return dr.DirJWTStore.Reload()
 }
 
 func respondToUpdate(s *Server, respSubj string, acc string, message string, err error) {
@@ -2929,6 +2934,7 @@ func respondToUpdate(s *Server, respSubj string, acc string, message string, err
 func (dr *DirAccResolver) Start(s *Server) error {
 	dr.Lock()
 	defer dr.Unlock()
+	dr.Server = s
 	dr.DirJWTStore.changed = func(pubKey string) {
 		if v, ok := s.accounts.Load(pubKey); !ok {
 		} else if jwt, err := dr.LoadAcc(pubKey); err != nil {
@@ -2998,7 +3004,7 @@ func (dr *DirAccResolver) Start(s *Server) error {
 				// let them timeout
 				s.Errorf("pack request error: %v", err)
 			} else {
-				s.Debugf("pack request hash %x - finished responding with hash %x")
+				s.Debugf("pack request hash %x - finished responding with hash %x", theirHash, ourHash)
 				s.sendInternalMsgLocked(reply, "", nil, []byte{})
 			}
 		}); err != nil {
@@ -3039,7 +3045,18 @@ func (dr *DirAccResolver) Start(s *Server) error {
 }
 
 func (dr *DirAccResolver) Fetch(name string) (string, error) {
-	return dr.LoadAcc(name)
+	if theJWT, err := dr.LoadAcc(name); theJWT != "" {
+		return theJWT, nil
+	} else {
+		dr.Lock()
+		srv := dr.Server
+		dr.Unlock()
+		if srv == nil {
+			return "", err
+		} else {
+			return srv.fetch(dr, name) // lookup from other server
+		}
+	}
 }
 
 func (dr *DirAccResolver) Store(name, jwt string) error {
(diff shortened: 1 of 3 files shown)


server/jwt_test.go
```diff
@@ -3196,23 +3195,68 @@ func TestAccountNATSResolverFetch(t *testing.T) {
...
+func writeJWT(t *testing.T, dir string, pub string, jwt string) {
+	t.Helper()
+	err := ioutil.WriteFile(filepath.Join(dir, pub+".jwt"), []byte(jwt), 0644)
+	require_NoError(t, err)
+}
+
+func TestAccountNATSResolverFetch(t *testing.T) {
+	origEventsHBInterval := eventsHBInterval
+	eventsHBInterval = 50 * time.Millisecond // speed up eventing
+	defer func() { eventsHBInterval = origEventsHBInterval }()
```
kozlovic, line 3259: Running locally I got this data race once. It means that `require_NoLocalOrRemoteConnections` was invoking `checkAccClientsCount` that failed, and this line was invoked while the server was still trying to send an account update msg. This is because sB and sC don't have a "defer shutdown" after you call RunServerWithConfig() because you expect to shut them down explicitly as part of the test. Had you had the defer shutdown you would not have gotten the DATA RACE, but a failed test nonetheless.
```
=== RUN   TestAccountNATSResolverFetch
    jwt_test.go:3488: Expected account "ABBFF6DDYLSTIZXJVXLAGJGPUHUPKWQMHMJT5NZ65E5I3EAHK62FYRHN" to have 0 clients, got 1
==================
WARNING: DATA RACE
Write at 0x0000024a5a48 by goroutine 8:
  github.com/nats-io/nats-server/server.TestAccountNATSResolverFetch.func1()
      /Users/ivan/dev/go/src/github.com/nats-io/nats-server/server/jwt_test.go:3259 +0x3a
  runtime.call32()
      /usr/local/go/src/runtime/asm_amd64.s:539 +0x3a
  testing.(*common).Fatal()
      /usr/local/go/src/testing/testing.go:766 +0x89
  github.com/nats-io/nats-server/server.checkFor()
      /Users/ivan/dev/go/src/github.com/nats-io/nats-server/server/server_test.go:52 +0x18f
  github.com/nats-io/nats-server/server.checkAccClientsCount()
      /Users/ivan/dev/go/src/github.com/nats-io/nats-server/server/client_test.go:172 +0x9a
  github.com/nats-io/nats-server/server.TestAccountNATSResolverFetch.func2()
      /Users/ivan/dev/go/src/github.com/nats-io/nats-server/server/jwt_test.go:3264 +0x108
  github.com/nats-io/nats-server/server.TestAccountNATSResolverFetch.func4()
      /Users/ivan/dev/go/src/github.com/nats-io/nats-server/server/jwt_test.go:3298 +0xcf
  github.com/nats-io/nats-server/server.TestAccountNATSResolverFetch()
      /Users/ivan/dev/go/src/github.com/nats-io/nats-server/server/jwt_test.go:3488 +0x32e8
  testing.tRunner()
      /usr/local/go/src/testing/testing.go:1039 +0x1eb

Previous read at 0x0000024a5a48 by goroutine 66:
  github.com/nats-io/nats-server/server.(*Server).sendAccConnsUpdate()
      /Users/ivan/dev/go/src/github.com/nats-io/nats-server/server/events.go:1184 +0x506
  github.com/nats-io/nats-server/server.(*Server).accConnsUpdate()
      /Users/ivan/dev/go/src/github.com/nats-io/nats-server/server/events.go:1198 +0x2b6
  github.com/nats-io/nats-server/server.(*Server).sendAccConnsUpdate.func1()
      /Users/ivan/dev/go/src/github.com/nats-io/nats-server/server/events.go:1182 +0x4a

Goroutine 8 (running) created at:
  testing.(*T).Run()
      /usr/local/go/src/testing/testing.go:1090 +0x700
  testing.runTests.func1()
      /usr/local/go/src/testing/testing.go:1334 +0xa6
  testing.tRunner()
      /usr/local/go/src/testing/testing.go:1039 +0x1eb
  testing.runTests()
      /usr/local/go/src/testing/testing.go:1332 +0x527
  testing.(*M).Run()
      /usr/local/go/src/testing/testing.go:1249 +0x43f
  github.com/nats-io/nats-server/server.TestMain()
      /Users/ivan/dev/go/src/github.com/nats-io/nats-server/server/sublist_test.go:1260 +0x379
  main.main()
      _testmain.go:1680 +0x223

Goroutine 66 (finished) created at:
  time.goFunc()
      /usr/local/go/src/time/sleep.go:168 +0x51
==================
    testing.go:954: race detected during execution of test
--- FAIL: TestAccountNATSResolverFetch (6.09s)
=== CONT  
    testing.go:954: race detected during execution of test
```

kozlovic (changes requested):
Let's add the defer s.Shutdown in the test and see if we can find the condition that still made that test fail.

server/jwt_test.go
matthiashanel: Does this really relate to shutdown or is this related to altering eventsHBInterval which is not really designed to be changed (no lock)?
sendAccConnsUpdate is executed on a timer and in different go routines. 
Not sure how calling shutdown manually would alter this.

Given my other change, what I could do on the last line is making sure the the timer is nil.

kozlovic: What I tried to explain is that in the above data race, the test had failed (checkAccClientsCount() ends-up calling t.Fatalf() which ends the test). But because some of the server did not have a `defer s.Shutdown()` (because you are explicitly shutting down in some part of the test), then this created the race between the `defer` that resets the HB event interval and some go routine of still running servers. Had you had the `defer s.Shutdown()` for each server that has been started, they would have executed before, preventing the data race on that last defer.
So as I said, we would not get the DATA RACE, but the test is still failing in some situations. Happy to chat about it if this is still not clear.

server/accounts.go
```diff
@@ -3039,7 +3045,18 @@ func (dr *DirAccResolver) Start(s *Server) error {
...
-	return dr.LoadAcc(name)
+	if theJWT, err := dr.LoadAcc(name); theJWT != "" {
+		return theJWT, nil
+	} else {
+		dr.Lock()
+		srv := dr.Server
+		dr.Unlock()
+		if srv == nil {
+			return "", err
+		} else {
```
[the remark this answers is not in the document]
matthiashanel: fixed

server/jwt_test.go
matthiashanel: fixed by waiting for the system account timer being nil.

matthiashanel:
@kozlovic fixed the issue by waiting for the system account timer being gone and rebased.

server/jwt_test.go
kozlovic: Could you do me a favor and still add a `defer s<x>.Shutdown()` at lines 3419, 3445 and 3505? It is always a good practice to put the defer shutdown even if you know you are going to explicitly shutdown. It helps if there is a failure, etc.. if the server is explicitly shutdown, the defer will be a no-op. Thanks!

kozlovic (changes requested):

server/jwt_test.go
matthiashanel: Sorry didn't see your first reply prior to making the change. Added defer s<x> Shutdown() where it was missing. 
Reason why I went with a check is that except for the first line, sendAccConnsUpdate and especially turning off the timer is not affected by Shutdown(). We speed up the timer so the odds for the timer firing prior to the go routine being finished are high. I check all clients to be gone. The only one that wasn't was the system account.

kozlovic (approved):
LGTM

PR #1617, opened 2020-09-26
Fix event ordering issue caused by re-locking

matthiashanel:
Noticed TestSystemAccountConnectionUpdatesStopAfterNoLocal failing du to the timer still being around.

I moved clearing of the timer, but still got failures where the last message does not contain conn count 0.
This hinted at one event from one client interfering with an event from another.

To avoid this holding the account lock during send seems the most appropriate solution.
sendInternalMsg essentially records the send queue and unlocks the server. 
This change just merges that with the code in sendAccConnsUpdate.

### DIFF
--- a/server/events.go
+++ b/server/events.go
@@ -1185,7 +1185,14 @@ func (s *Server) sendAccConnsUpdate(a *Account, subj ...string) {
 		}
 	}
 	for _, sub := range subj {
-		sendQ <- &pubMsg{nil, sub, _EMPTY_, &m.Server, &m, false}
+		msg := &pubMsg{nil, sub, _EMPTY_, &m.Server, &m, false}
+		select {
+		case sendQ <- msg:
+		default:
+			a.mu.Unlock()
+			sendQ <- msg
+			a.mu.Lock()
+		}
 	}
 	a.mu.Unlock()
 	s.mu.Lock()
--- a/server/events.go
+++ b/server/events.go
@@ -1159,8 +1159,13 @@ func (s *Server) sendAccConnsUpdate(a *Account, subj ...string) {
 	if !s.eventsEnabled() || a == nil {
 		return
 	}
+	sendQ := s.sys.sendq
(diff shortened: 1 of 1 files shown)


server/events.go
```diff
@@ -1184,7 +1184,11 @@ func (s *Server) sendAccConnsUpdate(a *Account, subj ...string) {
 			a.ctmr.Reset(eventsHBInterval)
 		}
 	}
+	for _, sub := range subj {
+		sendQ <- &pubMsg{nil, sub, _EMPTY_, &m.Server, &m, false}
```
kozlovic, line 1188: Yeah, but I could see how this whole thing could deadlock (may not be the only case, but...): suppose the send queue is full (because other go routines keep trying to add to the queue) and the internalSendLoop() is currently processing 1 message and as part of processInboundClientMsg() is locking this account. Then we have a problem.

kozlovic (changes requested):
Need to revisit locking...

server/events.go
matthiashanel: Agreed, was hoping the queue size would avoid that. 

The other alternative is to let go of the account lock earlier. 
This will improve the test as well (essentially moving the timer removal prior to send), but it will not rule out a test failure.  (just makes it more unlikely)

What do you think about unlocking if the send where to block?
```
select {
case sendQ <- &pubMsg{nil, sub, _EMPTY_, &m.Server, &m, false}:
default:
    a.mu.Unlock()
    sendQ <- &pubMsg{nil, sub, _EMPTY_, &m.Server, &m, false}:
    a.mu.Lock()
}
```

kozlovic: That's indeed better than blocking with the account lock held and risking a deadlock when queue is full.

matthiashanel: Will do that then. This will avoid event reordering until the queue gets full (which really it shouldn't)

kozlovic (approved):
LGTM

PR #1620, opened 2020-09-28
AddConsumer causes redelivery of first message

### LINKED ISSUE #1619: JetStream: AddConsumer() causes first message to be redelivered twice
- Create a durable consumer
- Send a message while the sub is running but don't ack it
- Stop the subscription
- Send another message while the sub is not running.
- Call AddConsumer() again with a different delivery subject
- The message number 1 is delivered twice, followed by message number 2.

kozlovic:
It seems that when updating the delivery subject, we use as the
first sequence the ack floor, but if no message was ever ack'ed
then it causes the first message to be redelivered twice.

Resolves #1619

### DIFF
--- a/server/consumer.go
+++ b/server/consumer.go
@@ -675,6 +675,14 @@ func (o *Consumer) updateDeliverSubject(newDeliver string) {
 	o.dseq = o.adflr
 	o.sseq = o.asflr
 
+	// If we never received an ack, set to 1.
+	if o.dseq == 0 {
+		o.dseq = 1
+	}
+	if o.sseq == 0 {
+		o.sseq = 1
+	}
+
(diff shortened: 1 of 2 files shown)


kozlovic:
@derekcollison This PR does not fix the issue, simply add a test that shows the problem. I just copied TestJetStreamDurableConsumerReconnect() but in that test send only 1 message before stopping the sub and that message is not ack'ed. Then send a second message while the sub is not running. On "restart" of the durable, I can see message 1 delivered twice, followed by message 2.

derekcollison:
ok your hunch was right, easy fix. Feel free to merge etc.

derekcollison (approved):
LGTM - I added the fix too

PR #1622, opened 2020-09-29
Fixed possible panic on consumer.Delete() during server Shutdown()

### LINKED ISSUE #1621: JetStream: possible panic on consumer delete during server shutdown
This would produce the following stack:
```
panic: close of closed channel

goroutine 2360 [running]:
github.com/nats-io/nats-server/v2/server.(*Consumer).stop(0xc0003c4e00, 0x10101, 0x0, 0x0)
	/Users/ivan/dev/go/src/github.com/nats-io/nats-server/server/consumer.go:1722 +0xf5
github.com/nats-io/nats-server/v2/server.(*Consumer).Delete(...)
	/Users/ivan/dev/go/src/github.com/nats-io/nats-server/server/consumer.go:1706
```

kozlovic:
The panic was caused by the closing of an already closed Go channel.
The Delete() relied on the consumer's mset being nil to consider
the consumer already closed. However, the consumer's mset is set
to nil after invoking sendDeleteAdvisoryLocked() which internally
invokes sendAdvisory() which releases/reacquires the consumer lock.
This left an open door for a race to occur and Delete() to be
invoked twice on the same consumer.

Moving setting the consumer's mset to nil too early would prevent
the sendAdvisory() to actually do its job. We could pass the mset
to sendAvisory(), but a simpler approach is to simply use a "closed"
boolean on the Consumer object that is set to true at the beginning
of the Delete() function.

Resolves #1621

### DIFF
--- a/server/consumer.go
+++ b/server/consumer.go
@@ -193,6 +193,7 @@ type Consumer struct {
 	ackEventT         string
 	deliveryExcEventT string
 	created           time.Time
+	closed            bool
 }
 
 const (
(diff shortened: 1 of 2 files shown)


derekcollison (approved):
LGTM

PR #1623, opened 2020-09-30
Moving TestQueueAutoUnsubscribe to norace_test.go

kozlovic:
This test has been found to cause TestAccountNATSResolverFetch to
fail on macOS. We did not find the exact reason yet, but it seem
that with `-race`, the queue auto-unsub test (that creates 2,000
queue subs and sends 1,000 messages) cause mem to grow to 256MB
(which we know -race is memory hungry) and that may be causing
interactions with the account resolver test.

For now, moving it to norace_test.go, which consumes much less
memory (25MB) and anyway is a better place since it would stress
better the "races" of having a queue sub being unsubscribed while
messages were inflight to this queue sub.

### DIFF
--- a/server/client_test.go
+++ b/server/client_test.go
@@ -1586,65 +1586,6 @@ func TestDynamicBuffers(t *testing.T) {
 	checkBuffers(minBufSize, minBufSize)
 }
 
-// Similar to the routed version. Make sure we receive all of the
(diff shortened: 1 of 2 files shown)


matthiashanel (approved):
LGTM

PR #1624, opened 2020-09-30
Updated jwt library and check (account & token) issuer prior to jwt Validate

matthiashanel:


kozlovic (approved):
LGTM

PR #1626, opened 2020-09-30
Add Store() method for URLAccResolver

masudur-rahman:
Signed-off-by: Masudur Rahman <masudjuly02@gmail.com>

- Reference : https://github.com/nats-io/nsc/blob/master/cmd/store/store.go#L355

### DIFF
--- a/server/accounts.go
+++ b/server/accounts.go
@@ -2815,6 +2815,10 @@ func (*resolverDefaultsOpsImpl) Reload() error {
 func (*resolverDefaultsOpsImpl) Close() {
 }
 
+func (*resolverDefaultsOpsImpl) Store(_, _ string) error {
+	return fmt.Errorf("Store operation not supported for the specified Account Resolver")
+}
+
 // MemAccResolver is a memory only resolver.
 // Mostly for testing.
 type MemAccResolver struct {
@@ -2889,13 +2893,15 @@ func (ur *URLAccResolver) Fetch(name string) (string, error) {
 // to the base url, appending the account name onto the end.
 func (ur *URLAccResolver) Store(name, jwt string) error {
 	url := ur.url + name
-	resp, err := http.Post(url, "application/jwt", strings.NewReader(jwt))
+	resp, err := ur.c.Post(url, "application/jwt", strings.NewReader(jwt))
 	if err != nil {
 		return err
-	} else if resp.StatusCode != http.StatusOK {
+	}
+
+	defer resp.Body.Close()
+	if resp.StatusCode != http.StatusOK {
 		return fmt.Errorf("could not post to <%q>: %v", url, resp.Status)
 	}
-	resp.Body.Close()
 
 	return err
 }
--- a/server/accounts.go
+++ b/server/accounts.go
@@ -2815,10 +2815,6 @@ func (*resolverDefaultsOpsImpl) Reload() error {
 func (*resolverDefaultsOpsImpl) Close() {
 }
 
-func (*resolverDefaultsOpsImpl) Store(_, _ string) error {
-	return fmt.Errorf("Store operation not supported for URL Resolver")
-}
-
 // MemAccResolver is a memory only resolver.
 // Mostly for testing.
 type MemAccResolver struct {
@@ -2887,6 +2883,21 @@ func (ur *URLAccResolver) Fetch(name string) (string, error) {
 		return _EMPTY_, err
 	}
 	return string(body), nil
+}
+
+// Store will store or update the jwt claim of account `name`
+// to the base url, appending the account name onto the end.
+func (ur *URLAccResolver) Store(name, jwt string) error {
+	url := ur.url + name
+	resp, err := http.Post(url, "application/jwt", strings.NewReader(jwt))
+	if err != nil {
+		return err
+	} else if resp.StatusCode != http.StatusOK {
+		return fmt.Errorf("could not post to <%q>: %v", url, resp.Status)
+	}
+	resp.Body.Close()
+
+	return err
 }
 
 // Resolver based on nats for synchronization and backing directory for storage.


server/accounts.go
```diff
@@ -2815,10 +2815,6 @@ func (*resolverDefaultsOpsImpl) Reload() error {
 func (*resolverDefaultsOpsImpl) Close() {
 }
 
-func (*resolverDefaultsOpsImpl) Store(_, _ string) error {
```
kozlovic, line 2818: Should not remove this one. This is the "base" implementation that has all interface's method implemented - even if returning errors. Other implementations that embed this implementation will either override the method they chose or fallback to the base implementation.
I agree that the error text could be changed though.

server/accounts.go
```diff
@@ -2889,6 +2885,21 @@ func (ur *URLAccResolver) Fetch(name string) (string, error) {
 	return string(body), nil
 }
 
+// Store will store or update the jwt claim of account `name`
+// to the base url, appending the account name onto the end.
+func (ur *URLAccResolver) Store(name, jwt string) error {
+	url := ur.url + name
+	resp, err := http.Post(url, "application/jwt", strings.NewReader(jwt))
```
kozlovic, line 2892: `ur` has a `c` http client that should be used here. So `ur.c.Post(` instead.

server/accounts.go
```diff
@@ -2889,6 +2885,21 @@ func (ur *URLAccResolver) Fetch(name string) (string, error) {
...
+	if err != nil {
+		return err
+	} else if resp.StatusCode != http.StatusOK {
```
kozlovic, line 2895: After the first `if err != nil` you should first do a `defer resp.Body.Close()`. Then have this `if resp.StatusCode != ..` test.

server/accounts.go
```diff
@@ -2889,6 +2885,21 @@ func (ur *URLAccResolver) Fetch(name string) (string, error) {
...
+		return fmt.Errorf("could not post to <%q>: %v", url, resp.Status)
```
kozlovic, line 2896: The directive `%q` is already surrounding the text with double quotes, so no need for `<>`.

kozlovic (changes requested):
Thank you for your PR. I have some code changes suggestions, but more importantly, curious to know what you decided to add that to the URL account resolver. Clearly the server does not use it, so just wondering what is your use case.

matthiashanel (commented):
I would like to learn more about your usecase. 
The server only calls Store when looking at resolver_preloads.

masudur-rahman:
@matthiashanel, when someone uses this `go client` to create server, s/he may need it. It will be useful if the the server provides a `Store` method for `URLAccResolver`.

server/accounts.go
masudur-rahman: I added the default `Store` method but changed a bit in the error text.

server/accounts.go
masudur-rahman: Thanks, I updated like you suggested.

server/accounts.go
masudur-rahman: done

server/accounts.go
masudur-rahman: The `Fetch` method has the same format. I wanted to have same for `Store` method.

https://github.com/masudur-rahman/nats-server/blob/6685cd2ece64172caa206c5c83f8efd7245e2918/server/accounts.go#L2877-L2882

kozlovic: Fair enough!

kozlovic (approved):
LGTM

kozlovic:
@masudur-rahman The code change looks good to me. I would like to check with @matthiashanel, based on your use-case, if this is going to be an issue or not. Thanks!

matthiashanel:
@masudur-rahman The interface was introduced because we added new resolver types. 
They are using nats instead of http in order to perform lookups.
To support resolver_preload, that up to this point was only used by the memory resolver, Store was added.
Accompanying Store is IsReadOnly. If IsReadOnly is true store will be called on startup or reload.

This means whenever a server will be started or reloaded, the webserver it is being pointed at will see the jwt be set to whatever is in the server config. 
If you have no other way to put the jwt in the webserver you should use the memory resolver outright and avoid the http lookup altogether. 
If you have another way to put jwt in the webserver (does not apply to account server, it checks for that) you may see your changes overwritten. Say you update a jwt manually then the server restarts and posts his (possibly outdated) jwt.

To make this more useful we could try calling Store form within accountClaimUpdate as well. 
In a cluster where every server is set up with this, the every server will post the same update. 
This is in part why nsc does the post to the account server, it generates the system event and notifies nats-server.

Another thing I'm somewhat weary of is the lack of auth for this.
If a nats-server can post a jwt, everyone can. 

@kozlovic what do you think?

kozlovic:
@matthiashanel This is why I asked your opinion, clearly I did not see the impact of this PR. I was just commenting on the code changes, which after some iterations looked good, but I did not have a grasp on the API itself.

With what you said, I am a bit confused. It looks like a resolver with IsReadOnly() causes server to invoke Store(). That does not sound intuitive to me. And the fact that the "default" resolver implementation set it to true means that the change to the URL resolver overriding the Store() function would cause the server to invoke it, although it sounds like we should not. I would have expected the default implementation to minimize risk of that to happen, meaning that any override should have been done in such a way that the coder knows that he/she needs to "enable" the Store() behavior. What I say may not be very clear, so happy to discuss it if it helps.

matthiashanel:
Sorry @kozlovic, this should have been isReadOnly false then Store will be called.

kozlovic:
@matthiashanel Ok, so in this case for instance, the addition of URL resolver's Store() will have no impact to the server. The user (if embedding NATS Server, etc..) can if desired invoke Store() directly on a resolver, as we do for some tests with the memory one. Correct?

matthiashanel:
You are right. Maybe embedding is the only use case for this AND your setup is not conducive to using memory resolver (meaning you are running a cluster of embedded server).

kozlovic:
@masudur-rahman So with what has been discussed, do you still think that adding this to the URL resolver is needed? Matthias mentioned to me that the new NATS resolver (that is available in the main branch) would even be better than using an URL resolver or account server.
We appreciate the contribution, but once we add something it has to be supported for a while, so we want to make sure that this addition is useful and not introduce any issue (which I don't think it does for the NATS Server). Thank you!

bruth:
@masudur-rahman Closing as stale, feel free to re-open if still relevant.

Issue #1627, opened 2020-10-01
JetStream: AckExplicit message removed while durable consumer "offline"


kozlovic:
Have 2 durable consumers on the same stream.
Both consumers filter on the same subject and use explicit ack.
Send 2 messages, each consumer receive and ack.
Stop the subscription on the 2nd durable.
Send 2 more messages.
First subscription receives and acks.
"Restart" 2nd subscription (update the 2nd durable with new delivery subject)
No message is received, the stream info shows that there are no message in the stream (message 3 and 4 have been removed).

derekcollison:
Let me know if you are doing any work here like writing a test, if not will take a look in a few.

kozlovic:
I am writing a test (I did sent you a note on Slack)

kozlovic:
Created the PR just with the test. See comment in the PR about other issues related to the removal of messages from the stream.

kozlovic:
Closed by #1628

PR #1628, opened 2020-10-01
AckExplicit removes message for "offline" durable

kozlovic:
As of now, this PR has only a new test that shows the issue.

It seems that Consumer.needAck() for AckExplicit should consider
that an Ack is needed if sseq > o.asflr and there is no pending
ack at all. However making this change would break the test
TestJetStreamInterestRetentionStream.

Also, running the new test with `-count 100` and by adding an
artificial delay in stream.ackMsg() (just before calling
mset.store.RemoveMsg(seq)) causes sometimes delete requests
for the same sequence to be processed twice, which causes
the new test to fail (even with an attempted fix as discussed
above). I think that the attempt to remove the same sequence twice
is messing up the state.

Resolves to #1627 

### DIFF
--- a/server/consumer.go
+++ b/server/consumer.go
@@ -664,28 +664,13 @@ func (o *Consumer) updateDeliverSubject(newDeliver string) {
 	o.mu.Lock()
 	defer o.mu.Unlock()
 
-	mset := o.mset
-	if mset == nil || o.isPullMode() {
+	if o.mset == nil || o.isPullMode() {
 		return
 	}
 
-	oldDeliver := o.config.DeliverSubject
-	o.dsubj = newDeliver
-	o.config.DeliverSubject = newDeliver
-	// FIXME(dlc) - check partitions, we may need offset.
-	o.dseq = o.adflr + 1
-	o.sseq = o.asflr + 1
-
-	// If we never received an ack, set to 1.
-	if o.dseq == 0 {
-		o.dseq = 1
-	}
-	if o.sseq == 0 {
-		o.sseq = 1
-	}
-
+	o.acc.sl.ClearNotification(o.dsubj, o.inch)
+	o.dsubj, o.config.DeliverSubject = newDeliver, newDeliver
 	// When we register new one it will deliver to update state loop.
-	o.acc.sl.ClearNotification(oldDeliver, o.inch)
 	o.acc.sl.RegisterNotification(newDeliver, o.inch)
 }
 
@@ -1037,7 +1022,7 @@ func (o *Consumer) needAck(sseq uint64) bool {
 		if sseq > o.asflr {
 			// Generally this means we need an ack, but just double check pending acks.
 			needAck = true
-			if len(o.pending) > 0 {
+			if len(o.pending) > 0 && sseq < o.sseq {
 				_, needAck = o.pending[sseq]
 			}
 		}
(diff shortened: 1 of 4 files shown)


server/consumer.go
```diff
@@ -664,28 +664,13 @@ func (o *Consumer) updateDeliverSubject(newDeliver string) {
 	o.mu.Lock()
 	defer o.mu.Unlock()
 
-	mset := o.mset
-	if mset == nil || o.isPullMode() {
+	if o.mset == nil || o.isPullMode() {
```
kozlovic, line 667: Could you use `o.closed` here instead of `o.mset`? I fixed recently the race on close (due to release of consumer's lock when sending advisory).

test/jetstream_test.go
```diff
@@ -7976,26 +7990,33 @@ func TestJetStreamAckExplicitMsgRemoval(t *testing.T) {
...
-			for i := 0; i < toSend; i++ {
+			for i := 1; i <= toSend; i++ {
 				m, err := sub2.NextMsg(time.Second)
 				if err != nil {
-					t.Fatalf("Error acking message: %v", err)
+					t.Fatalf("Error receiving message %d: %v", i, err)
 				}
 				m.Respond(nil)
+
+				sseq, dseq, dc, _ := o2.ReplyInfo(m.Reply)
```
kozlovic, line 8008: For all the following checks: if I add the artificial delay in stream.go:
```
		if !hasInterest {
			time.Sleep(100 * time.Millisecond)
			mset.store.RemoveMsg(seq)
		}
```
then these checks fail. What I mean is that is these checks not making too many assumptions on the timing of things? Is that going to be a flapper?
Would dseq not be possibly 3 and 4 (and not 4 and 5) if say the unsubscribe() is "fully processed" by the time new messages are sent. Adding a 250ms delay between unsubscribe and sending messages 3 and 4 show that it is the case.

To sum-up, I am not sure that this test should check the delivery/redelivery sequence but simply that message 3 and 4 are received (in any order, but not several times the same) .

server/consumer.go
derekcollison: yes will fix.

test/jetstream_test.go
derekcollison: Let me take a look and simulate your delay as well and see.

kozlovic: If you don't put the delay before RemoveMsg() but instead pause (like 250ms) between sub2.Unsubscribe() and sending messages 3 and 4, you will likely have message 3 and 4 delivered "normally". So the point of my comments was that these checks may make the test flap.

derekcollison: ok agree will remove check on consumer sequence.

test/jetstream_test.go
```diff
@@ -8005,10 +8005,7 @@ func TestJetStreamAckExplicitMsgRemoval(t *testing.T) {
 				}
 				m.Respond(nil)
 
-				sseq, dseq, dc, _ := o2.ReplyInfo(m.Reply)
-				if dseq != uint64(i+3) {
-					t.Fatalf("Expected consumer sequence of %d got %d", i+3, dseq)
-				}
+				sseq, _, dc, _ := o2.ReplyInfo(m.Reply)
```
kozlovic, line 8008: I think checking only that message 3 and 4 (and no duplicate) are received. Again, checking `sseq == 3 && dc == 1` being a failure is wrong if it happens that 3 and 4 are delivered (and not redelivered) should timing be such that there is no redelivery.

kozlovic:
LGTM

derekcollison (approved):
LGTM after discussion with @kozlovic and subsequent changes.

PR #1629, opened 2020-10-05
Avoid unnecessary CONNS subscription

matthiashanel:

Only one subscription is necessary. What is important for the user is that we publish on both, not that we also subscribe on both.

### DIFF
--- a/server/events.go
+++ b/server/events.go
@@ -575,12 +575,9 @@ func (s *Server) initEventTracking() {
 	}
 	s.sys.inboxPre = subject
 	// This is for remote updates for connection accounting.
-
-	for _, subj := range []string{accConnsEventSubjOld, accConnsEventSubjNew} {
-		subject = fmt.Sprintf(subj, "*")
-		if _, err := s.sysSubscribe(subject, s.remoteConnsUpdate); err != nil {
-			s.Errorf("Error setting up internal tracking for %s: %v", subject, err)
-		}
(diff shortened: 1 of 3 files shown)


server/events.go
```diff
@@ -575,12 +575,9 @@ func (s *Server) initEventTracking() {
 	}
 	s.sys.inboxPre = subject
 	// This is for remote updates for connection accounting.
-
-	for _, subj := range []string{accConnsEventSubjOld, accConnsEventSubjNew} {
-		subject = fmt.Sprintf(subj, "*")
-		if _, err := s.sysSubscribe(subject, s.remoteConnsUpdate); err != nil {
-			s.Errorf("Error setting up internal tracking for %s: %v", subject, err)
-		}
+	subject = fmt.Sprintf(accConnsEventSubjOld, "*")
```
kozlovic, line 578: But why introduce publish on `accConnsEventSubjNew` subject if we never subscribe on that subject. What is the plan then? Replace subscription on "old" with "new" at one point?
Not sure about all that..

kozlovic (commented):
Should we not publish on the new subject if we never subscribe to it?

server/events.go
kozlovic: Or is the new subject subscription for apps to listen to, not the server?

kozlovic: ok, got it.

kozlovic (approved):
LGTM

PR #1632, opened 2020-10-06
[Fixed] revocation check used current time instead of jwt issue time

matthiashanel:
Also empty revoked keys once account jwt has no revocations.

### DIFF
--- a/server/accounts.go
+++ b/server/accounts.go
@@ -2253,13 +2253,13 @@ func (a *Account) clearExpirationTimer() bool {
 }
 
 // checkUserRevoked will check if a user has been revoked.
-func (a *Account) checkUserRevoked(nkey string) bool {
+func (a *Account) checkUserRevoked(nkey string, issuedAt int64) bool {
 	a.mu.RLock()
 	defer a.mu.RUnlock()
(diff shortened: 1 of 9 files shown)


server/accounts.go
```diff
@@ -2650,17 +2651,16 @@ func (s *Server) updateAccountClaimsWithRefresh(a *Account, ac *jwt.AccountClaim
 		}
 		c.mu.Lock()
 		c.applyAccountLimits()
-		// Check for being revoked here. We use ac one to avoid
-		// the account lock.
-		var nkey string
-		if c.user != nil {
-			nkey = c.user.Nkey
-		}
+		// Check for being revoked here. We use ac one to avoid the account lock.
```
kozlovic, line 2654: I think this comment should be moved down to where we check for ac.Revocations, etc.. because it is confusing here :-)

kozlovic (changes requested):

server/accounts.go
matthiashanel: Moved 2 lines down and merged with the comment there

kozlovic (approved):
LGTM

PR #1633, opened 2020-10-08
[CHANGED] Default TLS and Auth timeout

kozlovic:
The default TLS timeout has been bumped to 2 seconds.
The default Auth timeout is now 2 seconds, but if TLS config
is present, it is equal to TLS timeout (possibly default value) + 1s.


wallyqs (approved):
LGTM!

derekcollison (approved):
LGTM

PR #1634, opened 2020-10-09
Accepting route warns if remote server has same name

kozlovic:

### DIFF
--- a/server/route.go
+++ b/server/route.go
@@ -491,6 +491,7 @@ func (c *client) processRouteInfo(info *Info) {
 
 	supportsHeaders := c.srv.supportsHeaders()
 	clusterName := c.srv.ClusterName()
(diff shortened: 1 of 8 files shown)


server/route.go
```diff
@@ -578,6 +579,12 @@ func (c *client) processRouteInfo(info *Info) {
 		return
 	}
 
+	// Check if remote has same server name than this server.
+	if !c.route.didSolicit && info.Name == srvName {
+		// For now simply report as a warning.
+		c.Warnf("Remote server has the same name: %q", info.Name)
```
derekcollison, line 585: has a duplicate name

derekcollison (approved):
LGTM - Minor suggestion on warning

server/route.go
kozlovic: done

PR #1635, opened 2020-10-09
Skip revocation check if the client has no jwt and is therefore internal

matthiashanel:
Related to #1632


kozlovic (approved):
LGTM

PR #1637, opened 2020-10-09
Update nightly image server binary path

variadico:
Currently, the NATS operator doesn't work with the nightly NATS Server images because binary is in a different location. This symlinks the binary to a few different known paths.

This is a backwards compatible change and shouldn't break existing users.

/cc @nats-io/core


wallyqs (approved):
LGTM

derekcollison (commented):
This feels like a bit of a kludge, is it possible to standardize on /bin/nats-server?

wallyqs:
We can revert the change, the other approach is to check in the nats operator that the image tag is `nightly` and use a different path instead.

ripienaar:
We can make them all the same, just move it to the same location is fine. Ditto config file locations

PR #1638, opened 2020-10-09
Revert "Update nightly image server binary path"

wallyqs:
Approach we'll take instead is to change to use for the path in the container to be `/bin/nats-server`

Reverts nats-io/nats-server#1637


derekcollison (approved):
LGTM

PR #1639, opened 2020-10-12
Adding list/delete/update operations for jwt stored by nats-resolver

matthiashanel:
Update already existed scoped by account, this exposes update without account.
List returns a list of all stored accounts.
Delete deletes accounts.
Fix a crash on startup with non existing directory.

This does NOT cause existing accounts to be evicted from memory and will be added in a separate PR.

### DIFF
--- a/server/accounts.go
+++ b/server/accounts.go
@@ -2905,9 +2905,17 @@ func (dr *DirAccResolver) Reload() error {
 
 func respondToUpdate(s *Server, respSubj string, acc string, message string, err error) {
 	if err == nil {
-		s.Debugf("%s - %s", message, acc)
+		if acc == "" {
+			s.Debugf("%s", message)
+		} else {
+			s.Debugf("%s - %s", message, acc)
+		}
 	} else {
-		s.Errorf("%s - %s - %s", message, acc, err)
+		if acc == "" {
+			s.Errorf("%s - %s", message, err)
+		} else {
+			s.Errorf("%s - %s - %s", message, acc, err)
+		}
 	}
 	if respSubj == "" {
 		return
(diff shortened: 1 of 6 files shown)


server/accounts.go
```diff
@@ -2917,19 +2925,65 @@ func respondToUpdate(s *Server, respSubj string, acc string, message string, err
 	if err == nil {
```
kozlovic, line 2925: Maybe you could have:
```
m := map[string]interface{}{}
if acc != _EMPTY_ {
    m["account"] = acc
}
if err == nil {
  m["code"] = http.StatusOK
  m["message"] = message
  response["data"] = m
} else {
 ...
```

server/accounts.go
```diff
@@ -2970,6 +3024,17 @@ func (dr *DirAccResolver) Start(s *Server) error {
 			return fmt.Errorf("error setting up update handling: %v", err)
 		}
 	}
+	if _, err := s.sysSubscribe(accClaimsReqSubj, func(_ *subscription, _ *client, subj, resp string, msg []byte) {
+		if claim, err := jwt.DecodeAccountClaims(string(msg)); err != nil {
```
kozlovic, line 3028: I know we already discussed this many times, and I can't force you to change your style, but I find those `if/else` statements sometimes difficult to read. I personally find this easier:
```
claim, err := jwt.DecodeAccountClaims(string(msg))
if err != nil {
    respondToUpdate(s, resp, "n/a", "jwt update resulted in error", err)
    return
}
if err := dr.save(claim.Subject, string(msg)); err != nil {
    respondToUpdate(s, resp, claim.Subject, "jwt update resulted in error", err)
    return
}
respondToUpdate(s, resp, claim.Subject, "jwt updated", nil)
```

server/accounts.go
```diff
@@ -3008,6 +3073,17 @@ func (dr *DirAccResolver) Start(s *Server) error {
 			}
 		}); err != nil {
 		return fmt.Errorf("error setting up pack request handling: %v", err)
+	} else if _, err = s.sysSubscribe(accListReqSubj,
```
kozlovic, line 3076: again, those `else` here make each line touch each other and very difficult to read.

server/accounts.go
```diff
@@ -3008,6 +3073,17 @@ func (dr *DirAccResolver) Start(s *Server) error {
...
+		// respond to list requests with one message containing all account ids
+		func(_ *subscription, _ *client, _, reply string, _ []byte) {
+			handleListRequest(dr.DirJWTStore, s, reply)
+		}); err != nil {
```
kozlovic, line 3080: that threw me off, had to have a second look to see that this was the `err` for a failed s.sysSubscribe()

server/accounts.go
```diff
@@ -3181,6 +3257,32 @@ func (dr *CacheDirAccResolver) Start(s *Server) error {
 			return fmt.Errorf("error setting up update handling: %v", err)
 		}
 	}
+	if _, err := s.sysSubscribe(accClaimsReqSubj, func(_ *subscription, _ *client, subj, resp string, msg []byte) {
+		if claim, err := jwt.DecodeAccountClaims(string(msg)); err != nil {
```
kozlovic, line 3261: won't say it again, but this is painful to read for me..

server/dirstore.go
```diff
@@ -420,6 +420,24 @@ func (store *DirJWTStore) write(path string, publicKey string, theJWT string) (b
...
+		return fmt.Errorf("store is read-only")
+	}
+	store.Lock()
+	defer store.Unlock()
+	if err := os.Remove(store.pathForKey(publicKey)); err != nil {
+		if _, ok := err.(*os.PathError); ok || err == os.ErrNotExist {
+			return nil
+		}
+		return err
+	} else {
```
kozlovic, line 434: again, if there was an error, we have returned already, so `else` is not necessary..

server/opts.go
```diff
@@ -872,11 +872,14 @@ func (o *Options) processConfigFileLine(k string, v interface{}, errors *[]error
...
 			}
 			if dir == "" {
-				*errors = append(*errors, &configErr{tk, "dir needs to point to a directory"})
+				*errors = append(*errors, &configErr{tk, "dir has no value and needs to point to a directory"})
 				return
 			}
-			if info, err := os.Stat(dir); err != nil || !info.IsDir() || info.Mode().Perm()&(1<<(uint(7))) == 0 {
-				info.IsDir()
+			if info, err := os.Stat(dir); err != nil {
+
```
kozlovic, line 879: Could we write this another way? Maybe like this?
```
if info, err := os.Stat(dir); err == nil && (!info.IsDir() || info.Mode().Perm()&(1<<(uint(7))) == 0) {
    *errors = append(*errors, &configErr{tk, "dir needs to point to an accessible directory"})
    return
}
```

kozlovic (commented):
Again, it is a matter of code style, but I wonder if you could make some of the changes.

server/opts.go
matthiashanel: done

server/dirstore.go
matthiashanel: done

server/accounts.go
matthiashanel: agreed

server/accounts.go
matthiashanel: done

server/accounts.go
matthiashanel: moved the comment ahead of if and the function on the sysSubscribe line

server/accounts.go
matthiashanel: done

kozlovic (approved):
LGTM

PR #1641, opened 2020-10-14
[FIXED] Stream's subscription propagation issue with gateways

kozlovic:
When creating shadow subscriptions for import streams, we were
not invoking code for gateway subscription accounting, which means
that when the account (for leafnodes) was switched to interest
only, those shadow subscriptions were not sent.


derekcollison (approved):
LGTM

PR #1643, opened 2020-10-15
fix npe for offsets larger than items

ripienaar:

/cc @nats-io/core


derekcollison (approved):
LGTM - Thanks!

PR #1644, opened 2020-10-15
Traffic shaping and full wildcard support for streams and services exports/imports

derekcollison:
This PR contains code to allow accounts to form subject mappings from one subject to another for both client inbound and service import invocations. This also allows weighted sets for the destinations.

```go
mappings = {
	foo: bar
	foo.*: [ { dest: bar.v1.$1, weight: 40% }, { destination: baz.v2.$1, weight: 20 } ]
	bar.*.*: RAB.$2.$1
}
```

Full wildcard transformations are possible, such that a user has full control over how subjects are matched and transformed.

This transform code was also utilized to enable full wildcard support for service and stream exports/imports.
For instance this is now possible.

```go
imports = [
	{ service: {account: "foo", subject:"request.*"}, to:"my.request.*"}
	{ stream:  {account: "foo", subject:"events.>"}, to:"foo.events.>"}
	{ stream:  {account: "foo", subject:"info.*.*.>"}, to:"foo.info.$2.$1.>"}
]
```

/cc @nats-io/core

### DIFF
--- a/server/accounts.go
+++ b/server/accounts.go
@@ -709,7 +709,7 @@ func (a *Account) selectMappedSubject(dest string) (string, bool) {
 	} else {
 		w := uint8(a.prand.Int31n(100))
 		for _, rm := range m.dests {
-			if w <= rm.weight {
+			if w < rm.weight {
 				d = rm
 				break
 			}
--- a/server/accounts.go
+++ b/server/accounts.go
@@ -525,6 +525,11 @@ func (a *Account) AddWeightedMappings(src string, dests ...*MapDest) error {
 	a.mu.Lock()
 	defer a.mu.Unlock()
 
+	// We use this for selecting between multiple weighted destinations.
+	if a.prand == nil {
+		a.prand = rand.New(rand.NewSource(time.Now().UnixNano()))
+	}
+
 	if !IsValidSubject(src) {
 		return ErrBadSubject
 	}
@@ -698,9 +703,6 @@ func (a *Account) selectMappedSubject(dest string) (string, bool) {
 	if len(m.dests) == 1 && m.dests[0].weight == 100 {
 		d = m.dests[0]
 	} else {
-		if a.prand == nil {
-			a.makeRand()
-		}
 		w := uint8(a.prand.Int31n(100))
 		for _, rm := range m.dests {
 			if w <= rm.weight {
@@ -720,16 +722,6 @@ func (a *Account) selectMappedSubject(dest string) (string, bool) {
 
 	a.mu.RUnlock()
 	return dest, true
-}
-
-// Small helper function.
-// Read lock assumed held.
-func (a *Account) makeRand() {
-	a.mu.RUnlock()
-	a.mu.Lock()
-	a.prand = rand.New(rand.NewSource(time.Now().UnixNano()))
-	a.mu.Unlock()
-	a.mu.RLock()
 }
 
 // SubscriptionInterest returns true if this account has a matching subscription
--- a/server/accounts.go
+++ b/server/accounts.go
@@ -562,8 +562,10 @@ func (a *Account) AddWeightedMappings(src string, dests ...*MapDest) error {
 		m.dests = append(m.dests, &destination{tr, d.Weight})
 	}
 
-	// Auto add in original at weight if all entries weight does not total to 100.
-	if tw != 100 {
+	// Auto add in original at weight difference if all entries weight does not total to 100.
+	// Iff the src was not already added in explicitly, meaning they want loss.
+	_, haveSrc := seen[src]
+	if tw != 100 && !haveSrc {
 		dest := src
 		if m.wc {
 			// We need to make the appropriate markers for the wildcards etc.
@@ -699,6 +701,7 @@ func (a *Account) selectMappedSubject(dest string) (string, bool) {
 
 	// The selected destination for the mapping.
 	var d *destination
+	var ndest string
 
 	// Optimize for single entry case.
 	if len(m.dests) == 1 && m.dests[0].weight == 100 {
@@ -715,14 +718,14 @@ func (a *Account) selectMappedSubject(dest string) (string, bool) {
 
 	if d != nil {
 		if len(d.tr.dtpi) == 0 {
-			dest = d.tr.dest
+			ndest = d.tr.dest
 		} else if nsubj, err := d.tr.transform(tts); err == nil {
-			dest = nsubj
+			ndest = nsubj
 		}
 	}
 
 	a.mu.RUnlock()
-	return dest, true
+	return ndest, true
 }
 
 // SubscriptionInterest returns true if this account has a matching subscription
--- a/server/accounts.go
+++ b/server/accounts.go
@@ -64,6 +64,7 @@ type Account struct {
 	lqws         map[string]int32
 	usersRevoked map[string]int64
 	actsRevoked  map[string]int64
+	mappings     []*mapping
 	lleafs       []*client
 	imports      importMap
 	exports      exportMap
@@ -100,8 +101,11 @@ type sconns struct {
 type streamImport struct {
 	acc     *Account
 	from    string
-	prefix  string
+	to      string
+	tr      *transform
+	rtr     *transform
 	claim   *jwt.Import
+	usePub  bool
 	invalid bool
 }
 
@@ -114,12 +118,13 @@ type serviceImport struct {
 	from        string
 	to          string
 	exsub       string
+	tr          *transform
 	ts          int64
 	rt          ServiceRespType
 	latency     *serviceLatency
 	m1          *ServiceLatency
 	rc          *client
-	hasWC       bool
+	usePub      bool
 	response    bool
 	invalid     bool
 	share       bool
@@ -209,7 +214,6 @@ func NewAccount(name string) *Account {
 		limits:   limits{-1, -1, -1, -1},
 		eventIds: nuid.New(),
 	}
-
 	return a
 }
 
@@ -488,6 +492,244 @@ func (a *Account) TotalSubs() int {
 	a.mu.RLock()
 	defer a.mu.RUnlock()
 	return int(a.sl.Count())
+}
+
+// MapDest is for mapping published subjects for clients.
+type MapDest struct {
+	Subject string
+	Weight  uint8
+}
+
+// destination is for internal representation for a weighted mapped destination.
+type destination struct {
+	tr     *transform
+	weight uint8
+}
+
+// mapping is an internal entry for mapping subjects.
+type mapping struct {
+	src   string
+	wc    bool
+	dests []*destination
+}
+
+// AddMapping adds in a simple route mapping from src subject to dest subject
+// for inbound client messages.
+func (a *Account) AddMapping(src, dest string) error {
+	return a.AddWeightedMappings(src, &MapDest{dest, 100})
+}
+
+// AddWeightedMapping will add in a weighted mappings for the destinations.
+// TODO(dlc) - Allow cluster filtering
+func (a *Account) AddWeightedMappings(src string, dests ...*MapDest) error {
+	a.mu.Lock()
+	defer a.mu.Unlock()
+
+	if !IsValidSubject(src) {
+		return ErrBadSubject
+	}
+
+	m := &mapping{src: src, wc: subjectHasWildcard(src), dests: make([]*destination, 0, len(dests)+1)}
+	seen := make(map[string]struct{})
+
+	var tw uint8
+	for _, d := range dests {
+		if _, ok := seen[d.Subject]; ok {
+			return fmt.Errorf("duplicate entry for %q", d.Subject)
+		}
+		seen[d.Subject] = struct{}{}
+		if d.Weight > 100 {
+			return fmt.Errorf("individual weights need to be <= 100")
+		}
+		tw += d.Weight
+		if tw > 100 {
+			return fmt.Errorf("total weight needs to be <= 100")
+		}
+		if !IsValidSubject(d.Subject) {
+			return ErrBadSubject
+		}
+		tr, err := newTransform(src, d.Subject)
+		if err != nil {
+			return err
+		}
+
+		m.dests = append(m.dests, &destination{tr, d.Weight})
+	}
+
+	// Auto add in original at weight if all entries weight does not total to 100.
+	if tw != 100 {
+		dest := src
+		if m.wc {
+			// We need to make the appropriate markers for the wildcards etc.
+			dest = transformTokenize(dest)
+		}
+		tr, err := newTransform(src, dest)
+		if err != nil {
+			return err
+		}
+		m.dests = append(m.dests, &destination{tr, 100 - tw})
+	}
+	sort.Slice(m.dests, func(i, j int) bool { return m.dests[i].weight < m.dests[j].weight })
+
+	var lw uint8
+	for _, d := range m.dests {
+		d.weight += lw
+		lw = d.weight
+	}
+	// Replace an old one if it exists.
+	for i, m := range a.mappings {
+		if m.src == src {
+			a.mappings[i] = m
+			return nil
+		}
+	}
+	// If we did not replace add to the end.
+	a.mappings = append(a.mappings, m)
+
+	return nil
+}
+
+// Helper function to tokenize subjects with partial wildcards into formal transform destinations.
+// e.g. foo.*.* -> foo.$1.$2
+func transformTokenize(subject string) string {
+	// We need to make the appropriate markers for the wildcards etc.
+	i := 1
+	var nda []string
+	for _, token := range strings.Split(subject, tsep) {
+		if token == "*" {
+			nda = append(nda, fmt.Sprintf("$%d", i))
+			i++
+		} else {
+			nda = append(nda, token)
+		}
+	}
+	return strings.Join(nda, tsep)
+}
+
+func transformUntokenize(subject string) (string, []string) {
+	var phs []string
+	var nda []string
+
+	for _, token := range strings.Split(subject, tsep) {
+		if len(token) > 1 && token[0] == '$' && token[1] >= '1' && token[1] <= '9' {
+			phs = append(phs, token)
+			nda = append(nda, "*")
+		} else {
+			nda = append(nda, token)
+		}
+	}
+	return strings.Join(nda, tsep), phs
+}
+
+// RemoveMapping will remove an existing mapping.
+func (a *Account) RemoveMapping(src string) bool {
+	a.mu.Lock()
+	defer a.mu.Unlock()
+	for i, m := range a.mappings {
+		if m.src == src {
+			// Swap last one into this spot. Its ok to change order.
+			a.mappings[i] = a.mappings[len(a.mappings)-1]
+			a.mappings[len(a.mappings)-1] = nil // gc
+			a.mappings = a.mappings[:len(a.mappings)-1]
+			return true
+		}
+	}
+	return false
+}
+
+// Indicates we have mapping entries.
+func (a *Account) hasMappings() bool {
+	if a == nil {
+		return false
+	}
+	a.mu.RLock()
+	n := len(a.mappings)
+	a.mu.RUnlock()
+	return n > 0
+}
+
+// This performs the logic to map to a new dest subject based on mappings.
+// Should only be called from processInboundClientMsg or service import processing.
+func (a *Account) selectMappedSubject(dest string) (string, bool) {
+	a.mu.RLock()
+	if len(a.mappings) == 0 {
+		a.mu.RUnlock()
+		return dest, false
+	}
+
+	// In case we have to tokenize for subset matching.
+	tsa := [32]string{}
+	tts := tsa[:0]
+
+	var m *mapping
+	for _, rm := range a.mappings {
+		if !rm.wc && rm.src == dest {
+			m = rm
+			break
+		} else {
+			// tokenize and reuse for subset matching.
+			if len(tts) == 0 {
+				start := 0
+				subject := dest
+				for i := 0; i < len(subject); i++ {
+					if subject[i] == btsep {
+						tts = append(tts, subject[start:i])
+						start = i + 1
+					}
+				}
+				tts = append(tts, subject[start:])
+			}
+			if isSubsetMatch(tts, rm.src) {
+				m = rm
+				break
+			}
+		}
+	}
+
+	if m == nil {
+		a.mu.RUnlock()
+		return dest, false
+	}
+
+	// The selected destination for the mapping.
+	var d *destination
+
+	// Optimize for single entry case.
+	if len(m.dests) == 1 && m.dests[0].weight == 100 {
+		d = m.dests[0]
+	} else {
+		if a.prand == nil {
+			a.makeRand()
+		}
+		w := uint8(a.prand.Int31n(100))
+		for _, rm := range m.dests {
+			if w <= rm.weight {
+				d = rm
+				break
+			}
+		}
+	}
+
+	if d != nil {
+		if len(d.tr.dtpi) == 0 {
+			dest = d.tr.dest
+		} else if nsubj, err := d.tr.transform(tts); err == nil {
+			dest = nsubj
+		}
+	}
+
+	a.mu.RUnlock()
+	return dest, true
+}
+
+// Small helper function.
+// Read lock assumed held.
+func (a *Account) makeRand() {
+	a.mu.RUnlock()
+	a.mu.Lock()
+	a.prand = rand.New(rand.NewSource(time.Now().UnixNano()))
+	a.mu.Unlock()
+	a.mu.RLock()
 }
 
 // SubscriptionInterest returns true if this account has a matching subscription
@@ -1040,15 +1282,14 @@ func (a *Account) AddServiceImportWithClaim(destination *Account, from, to strin
 	if destination == nil {
 		return ErrMissingAccount
 	}
-	// Empty means use from. Also means we can use wildcards since we are not doing remapping.
-	if !IsValidSubject(from) || (to != "" && (!IsValidLiteralSubject(from) || !IsValidLiteralSubject(to))) {
-		return ErrInvalidSubject
-	}
-
 	// Empty means use from.
 	if to == "" {
 		to = from
 	}
+	if !IsValidSubject(from) || !IsValidSubject(to) {
+		return ErrInvalidSubject
+	}
+
 	// First check to see if the account has authorized us to route to the "to" subject.
 	if !destination.checkServiceImportAuthorized(a, to, imClaim) {
 		return ErrServiceImportAuthorization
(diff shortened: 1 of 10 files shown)


server/accounts.go
```diff
@@ -490,6 +497,117 @@ func (a *Account) TotalSubs() int {
...
+	// Optimize for single entry case.
+	if len(rms) == 1 && rms[0].Weight == 100 {
+		dest = []byte(rms[0].Subject)
+	} else {
+		if a.prand == nil {
+			a.prand = rand.New(rand.NewSource(time.Now().UnixNano()))
+		}
+		w := uint8(a.prand.Int31n(100))
+		for _, rm := range rms {
+			if w <= rm.Weight {
```
kozlovic: So with the question I had above with the resulting `ndests`, when the user select a weight of `20, 30, 50` for 3 destinations, since it becomes `20, 50, 100`, the one with 50% weight would be selected *only* if the random number here is 100, which is 1 out of 100 chances? That does not seem right to me.

kozlovic (commented):
Review of 1st commit...

kozlovic:
@derekcollison @wallyqs @matthiashanel I started to review 1 commit at a time and this is a mistake since they are not independent (meaning that some code in first commit disappears in 2nd, etc..). So I guess better to do review of the whole thing at once. Some of my earlier comments may not make much sense if you don't check the 1st commit separately..

server/sublist.go
```diff
@@ -1010,6 +1010,32 @@ func IsValidSubject(subject string) bool {
 	return true
 }
 
+// Will share relevant info regarding the subject.
+// Returns valid, num tokens, num pwcs, has fwc.
```
kozlovic, line 1014: `tokens`, not `num tokens`

server/server.go
```diff
@@ -586,7 +586,17 @@ func (s *Server) configureAccounts() error {
 	// Check opts and walk through them. We need to copy them here
 	// so that we do not keep a real one sitting in the options.
 	for _, acc := range s.opts.Accounts {
-		a := acc.shallowCopy()
+		var a *Account
```
kozlovic, line 589: Would want @matthiashanel to have a look since if I recall he had to add the shallowCopy() to fix some bugs during reload. That being said, since I believe the $G account cannot referenced in configurations, this should not be a problem, but Matthias has looked at this in more details in the past.

server/parser.go
```diff
@@ -389,6 +389,17 @@ func (c *client) parse(buf []byte) error {
...
 					return err
 				}
+				// Check if we have and account mappings or tees or filters.
+				// FIXME(dlc) - Probably better way to do this.
+				// Could add in cache but will be tricky since results based on pub subject are dynamic
+				// due to wildcard matching and weight sets.
+				if c.kind == CLIENT && c.in.flags.isSet(hasMappings) {
+					old := c.pa.subject
+					c.selectMappedSubject()
+					if trace && !bytes.Equal(old, c.pa.subject) {
```
kozlovic, line 399: other way would be for selectMappedSubject() to return a boolean to indicate that an actual mapping took place.

server/client.go
```diff
@@ -3755,9 +3814,15 @@ func (c *client) processMsgResults(acc *Account, r *SublistResult, msg, deliver,
 			// Assume delivery subject is normal subject to this point.
 			dsubj = subj
 			// Check for stream import mapped subs. These apply to local subs only.
-			if sub.im != nil && sub.im.prefix != "" {
-				dsubj = append(_dsubj[:0], sub.im.prefix...)
-				dsubj = append(dsubj, subj...)
+			if sub.im != nil {
```
kozlovic, line 3817: Looks the same than the one just above. Could maybe make it a function (one inside this function if we want to make use of _dsubj), but it's ok to duplicate..

derekcollison:
Should I squash into 1 commit?

server/accounts.go
```diff
@@ -490,6 +497,117 @@ func (a *Account) TotalSubs() int {
...
+
+	if !IsValidSubject(src) {
+		return ErrBadSubject
+	}
+
+	seen := make(map[string]struct{})
+	ndests := make([]*RouteDest, 0, len(dests)+1)
+	var tw uint8
+	for _, d := range dests {
+		if _, ok := seen[string(d.Subject)]; ok {
```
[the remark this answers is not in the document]
derekcollison: I think fixed in subsequent commit.

server/accounts.go
```diff
@@ -490,6 +497,117 @@ func (a *Account) TotalSubs() int {
...
+			return fmt.Errorf("duplicate entry for %q", d.Subject)
+		}
+		seen[string(d.Subject)] = struct{}{}
+		if d.Weight > 100 {
```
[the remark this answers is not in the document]
derekcollison: Toss up, either way is fine by me.

server/accounts.go
```diff
@@ -490,6 +497,117 @@ func (a *Account) TotalSubs() int {
...
+	// Auto add in original at weight if entries weight does not total 100.
+	if tw != 100 {
+		ndests = append(ndests, &RouteDest{src, 100 - tw})
+	}
+	sort.Slice(ndests, func(i, j int) bool { return ndests[i].Weight < ndests[j].Weight })
+	var lw uint8
+	for _, d := range ndests {
+		d.Weight += lw
+		lw = d.Weight
+	}
```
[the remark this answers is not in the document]
derekcollison: When we do weight selection and select a random number makes it easier to check which entry to use.

server/client.go
```diff
@@ -2352,36 +2390,24 @@ func (c *client) addShadowSubscriptions(acc *Account, sub *subscription) error {
...
-	if len(ims) > 0 || len(froms) > 0 {
-		shadow = make([]*subscription, 0, len(ims)+len(froms))
+	if len(ims) > 0 {
+		shadow = make([]*subscription, 0, len(ims))
 	}
 
-	// Now walk through collected importMaps
-	for _, im := range ims {
+	// Now walk through collected stream imports that matched.
+	for _, ime := range ims {
```
kozlovic, line 2408: This is going to cause a copy of the `ime` struct. Is that intentional? If not, either convert ims to `[]*ime` or do a for loop with `i` so that you can do `ime := &ims[i]` to save a copy.

server/accounts.go
derekcollison: No last entry will be selected when random number > 50 and <= 100.

server/accounts.go
```diff
@@ -1800,12 +2049,52 @@ func (a *Account) AddStreamImportWithClaim(account *Account, from, prefix string
...
+		tr     *transform
+		err    error
+	)
+	if subjectHasWildcard(from) {
+		if to == from {
+			usePub = true
+		} else {
+			// Create a transform
+			if tr, err = newTransform(from, transformTokenize(to)); err != nil {
+				a.mu.Unlock()
```
kozlovic, line 2085: I think remove this: I don't see the corresponding `a.mu.Lock()` above?

server/accounts.go
```diff
@@ -1332,9 +1563,27 @@ func (a *Account) addServiceImport(dest *Account, from, to string, claim *jwt.Im
...
+		err    error
+	)
+	if subjectHasWildcard(to) {
+		// If to and from match, then we use the published subject.
+		if to == from {
+			usePub = true
+		} else {
+			// Create a transform
+			if tr, err = newTransform(from, transformTokenize(to)); err != nil {
+				a.mu.Unlock()
```
kozlovic, line 1579: Remove?

server/accounts.go
```diff
@@ -490,6 +494,234 @@ func (a *Account) TotalSubs() int {
...
+
+	// The selected destination for the mapping.
+	var d *destination
+
+	// Optimize for single entry case.
+	if len(m.dests) == 1 && m.dests[0].weight == 100 {
+		d = m.dests[0]
+	} else {
+		if a.prand == nil {
+			a.prand = rand.New(rand.NewSource(time.Now().UnixNano()))
```
kozlovic, line 702: If you set something in the account (a.prand) under the account RLock() you may get some report. Should we set this when account is created instead?

matthiashanel:
These are two separate functionalities, right? 
For my understanding on how to best review this, how are they entangled so to make a single PR necessary?

kozlovic (changes requested):

server/accounts.go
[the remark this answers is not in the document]
kozlovic: Leave it this way.

server/accounts.go
kozlovic: Of course... sorry about that.

server/sublist.go
derekcollison: Thx fixed.

server/parser.go
derekcollison: I like that idea. Will incorporate.

server/accounts.go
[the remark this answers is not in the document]
kozlovic: Actually no, still use `string(d.Subject)`

server/client.go
derekcollison: Yep saw that and just decided to dupl like before.

server/client.go
derekcollison: Yes that was intentional, trying to avoid heap..

server/accounts.go
derekcollison: Thx

server/accounts.go
derekcollison: Not sure we want to blind create, but I see the issue you raise. Will fix.

server/accounts.go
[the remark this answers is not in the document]
derekcollison: ok fixed.

server/accounts.go
```diff
@@ -490,6 +497,117 @@ func (a *Account) TotalSubs() int {
...
+		d.Weight += lw
+		lw = d.Weight
+	}
+
+	if a.routeMapping == nil {
+		a.routeMapping = make(map[string][]*RouteDest)
+	}
+	// TODO(dlc) - check wildcards
+
+	a.routeMapping[src] = ndests
```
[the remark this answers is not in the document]
derekcollison: All destinations map to src for weighted destinations when creating the transform.

server/accounts.go
kozlovic: Sorry, this one was needed indeed.

server/client.go
kozlovic: Sure, but the `for range` is going to return a copy of `ime`. So it would be better to:
```
for i := 0; i < len(ims); i++{
   ime := &ims[i]
   ...
```

server/accounts.go
```diff
@@ -490,6 +494,244 @@ func (a *Account) TotalSubs() int {
...
+
+	// The selected destination for the mapping.
+	var d *destination
+
+	// Optimize for single entry case.
+	if len(m.dests) == 1 && m.dests[0].weight == 100 {
+		d = m.dests[0]
+	} else {
+		if a.prand == nil {
+			a.makeRand()
```
kozlovic, line 702: No please, don't do that, releasing the RLock to get the Lock, then release then grab RLock again is a recipe for nasty bugs where the state before/after this call may have changed.
Either set the rand on create or at least the first time a mapping is added, under the write lock.

kozlovic (changes requested):
Please do not introduce another case where lock is released/reacquired in a function that assumes the lock is held the whole time.

server/client.go
derekcollison: True, was not worried about the copying for speed per se but was trying to avoid lots of garbage generation during subscriptions.

server/accounts.go
derekcollison: Agree add in to AddMap where proper lock is held. Will update.

kozlovic (approved):
LGTM

derekcollison:
This does not address the race accessing `c.acc` in the client readloop which in general does not hold client lock. Will get an update for that and will ping in a bit.

server/client.go
```diff
@@ -1031,7 +1031,7 @@ func (c *client) readLoop(pre []byte) {
 		// Check if the account has mappings and if so set the local readcache flag.
 		// We check here to make sure any changes such as config reload are reflected here.
 		if c.kind == CLIENT {
-			if c.acc.hasMappings() {
+			if c.Account().hasMappings() {
```
kozlovic, line 1034: This is fine, but if we are concerned about acquiring the client lock just to check that, and since we are not protecting the whole "check+parse(which processes the message" anyway, we could also capture `acc` at top of function, under c's lock, and refresh at bottom of this function when we reacquire lock.

kozlovic (approved):
LGTM, but proposing alternative.

server/client.go
derekcollison: That is a much better idea, did not remember the lock was re-acquired at the bottom, thanks!

derekcollison: Will update.

server/accounts.go
```diff
@@ -490,6 +494,236 @@ func (a *Account) TotalSubs() int {
...
+		}
+		tr, err := newTransform(src, d.Subject)
+		if err != nil {
+			return err
+		}
+
+		m.dests = append(m.dests, &destination{tr, d.Weight})
+	}
+
+	// Auto add in original at weight if all entries weight does not total to 100.
```
matthiashanel, line 564: I'd have a special one so users can easily inject loss for testing.
And if not, I'd force the sum to always be 100, be explicit

kozlovic (approved):
LGTM

server/accounts.go
```diff
@@ -3281,3 +3571,185 @@ func (dr *CacheDirAccResolver) Start(s *Server) error {
 	s.Noticef("Managing some jwt in exclusive directory %s", dr.directory)
 	return nil
 }
+
+func (dr *CacheDirAccResolver) Reload() error {
+	return dr.DirAccResolver.Reload()
+}
+
+// Transforms for arbitrarily mapping subjects from one to another for maps, tees and filters.
```
matthiashanel, line 3579: What is a tee and filter? (I have an idea but I don't find a test or description how that'd be configured)

is a tee a weighted mapping with 200%?
from one subject to more than one other?
or a weighted mapping with > 100%? (say send all traffic from foo to foo but route one percent to a special subject?)

server/accounts.go
```diff
@@ -3281,3 +3571,185 @@ func (dr *CacheDirAccResolver) Start(s *Server) error {
...
+			return nil, ErrBadSubject
+		}
+	}
+
+	return &transform{src: src, dest: dest, dtoks: dtokens, stoks: stokens, dtpi: dtpi}, nil
+}
+
+// match will take a literal published subject that is associated with a client and will match and transform
+// the subject if possible.
+// TODO(dlc) - We could add in client here to allow for things like foo -> foo.$ACCOUNT
```
matthiashanel, line 3648: not sure if $ACCOUNT can be trusted (not faked) especially when used cross account

server/client.go
derekcollison: Fixed.

matthiashanel (changes requested):

server/server.go
```diff
@@ -586,7 +586,19 @@ func (s *Server) configureAccounts() error {
...
+		if acc.Name == globalAccountName {
+			a = s.gacc
+		} else {
+			a = acc.shallowCopy()
+		}
+		if acc.hasMappings() {
+			// For now just move and wipe from opts.Accounts version.
+			a.mappings = acc.mappings
+			acc.mappings = nil
```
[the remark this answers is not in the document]
derekcollison: If there are no mappings at all this will not detect a change. However you are saying if they do exist but no changes where made a reload would force use to re-evaluate the account, correct?

server/parser.go
```diff
@@ -389,6 +389,17 @@ func (c *client) parse(buf []byte) error {
 				if err := c.processPub(arg); err != nil {
 					return err
 				}
+				// Check if we have and account mappings or tees or filters.
```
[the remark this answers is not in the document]
derekcollison: Can't move the mappings into shallow copy. We have condition where its global account but mappings were configured.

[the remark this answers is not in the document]
derekcollison: For the tracing block, we need to processPub as is first then perform mapping etc.

server/parser.go
```diff
@@ -389,6 +389,17 @@ func (c *client) parse(buf []byte) error {
...
+				// FIXME(dlc) - Probably better way to do this.
+				// Could add in cache but will be tricky since results based on pub subject are dynamic
+				// due to wildcard matching and weight sets.
+				if c.kind == CLIENT && c.in.flags.isSet(hasMappings) {
+					old := c.pa.subject
+					changed := c.selectMappedSubject()
+					if trace && changed {
```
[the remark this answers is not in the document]
derekcollison: `selectMappedSubject` always has to be invoked, not just in trace mode.

server/opts.go
```diff
@@ -603,6 +603,14 @@ func (o *Options) processConfigFileLine(k string, v interface{}, errors *[]error
 	case "logtime":
 		o.Logtime = v.(bool)
 		trackExplicitVal(o, &o.inConfig, "Logtime", o.Logtime)
+	case "mappings", "maps":
+		gacc := NewAccount(globalAccountName)
+		o.Accounts = append(o.Accounts, gacc)
```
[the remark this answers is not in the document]
derekcollison: No it shouldn't, we just use the opts.Accounts one to hold the mappings metadata. I will do a quick double check though..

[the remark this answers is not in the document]
derekcollison: It doesn't.

server/monitor.go
```diff
@@ -2035,16 +2035,12 @@ func (s *Server) accountInfo(accName string) (*AccountInfo, error) {
 	}
 	imports := []ExtImport{}
 	for _, v := range a.imports.streams {
-		to := ""
-		if v.prefix != "" {
-			to = v.prefix + "." + v.from
-		}
```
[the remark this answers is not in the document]
derekcollison: Not following that comment, apologies.

server/monitor.go
```diff
@@ -2035,16 +2035,12 @@ func (s *Server) accountInfo(accName string) (*AccountInfo, error) {
 	}
 	imports := []ExtImport{}
 	for _, v := range a.imports.streams {
-		to := ""
```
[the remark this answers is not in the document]
derekcollison: Ok will look into what is needed there.

server/accounts.go
derekcollison: Not following, I coded explicit, was not good always doing math especially if say you have a service and you just want some traffic to be diverted.

If you have a destination that no one subscribes to that will be loss ;)

server/accounts.go
derekcollison: Tees or Filters not impl in this PR. Prefer to keep them there but can remove the comment.

server/accounts.go
derekcollison: Server fills it in, so can't be faked.. and you would own the mapping construct for the export.

server/server.go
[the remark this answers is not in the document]
matthiashanel: because the mappings in opts are set to nill the deepequal in reload.go is bound to detect a change.
```
func (s *Server) diffOptions(newOpts *Options) ([]option, error) {
	...
	for i := 0; i < oldConfig.NumField(); i++ {
	...
		if changed := !reflect.DeepEqual(oldValue, newValue); !changed {
			continue
		}
```

server/monitor.go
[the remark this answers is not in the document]
matthiashanel: you added new values to imports, these should be exposed as well.
lin 1927 has 
```
type ExtImport struct {
	jwt.Import
	Invalid bool `json:"invalid"`
       Transform transform `json:"transform"`   // <-- probably add smth. like this and assing in the loop where we iterate over imports
}
```

server/accounts.go
matthiashanel: Say I have a subscriber foo.
For testing a user application behaving under loss.
I set the weighted transform to 90% or -10% and we drop 10 percent of messages. With -0
Thought that could be useful, which is why I brought it up

server/accounts.go
matthiashanel: was  curios as I read in a comment smth that I did not see in code.
For tee, would have expected a mapping to return more than one string.

server/accounts.go
matthiashanel: Care needs to be taken that if foo -> foo.$ACCOUNT exists, the user can't send foo.<not my account> themselves

server/parser.go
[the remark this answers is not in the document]
matthiashanel: now I get it, this is PUB and not the message yet. ok

server/accounts.go
```diff
@@ -1800,12 +2051,51 @@ func (a *Account) AddStreamImportWithClaim(account *Account, from, prefix string
 			prefix = prefix + string(btsep)
 		}
 	}
+	return a.AddMappedStreamImportWithClaim(account, from, prefix+from, imClaim)
```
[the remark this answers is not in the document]
matthiashanel: @derekcollison will you address this? (had quite a few comments, just making sure that one does not get lost)

server/server.go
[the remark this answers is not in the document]
derekcollison: But we nil out the sl and clients, and that was before this change. So that would mean we always no matter what reload accounts..

server/monitor.go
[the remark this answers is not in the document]
derekcollison: Transform should not be added to import, config is same, transform just enable wildcard mapping based configs.

server/accounts.go
derekcollison: Ah so do not force the sum to be 100 to allow for loss. ok I think I understand now.

derekcollison: Would `foo -> {foo:80}` work? Meaning that this would mean we want 20% loss?

derekcollison: If so we can add the behavior in later since the interface would be same.

matthiashanel: That would be the idea yes. 

I wouldn't wait too long, just so we can still control what the behavior should be if the values do not add up to 100%

derekcollison: ok let me add in the behavior now.

server/client.go
```diff
@@ -3216,6 +3260,15 @@ func (c *client) processInboundMsg(msg []byte) {
...
 
+// selectMappedSubject will chose the mapped subject based on the client's inbound subject.
+func (c *client) selectMappedSubject() bool {
+	nsubj, changed := c.acc.selectMappedSubject(string(c.pa.subject))
+	if changed {
+		c.pa.subject = []byte(nsubj)
+	}
+	return changed
+}
+
```
matthiashanel, line 3271: Note:
For an inbound tee functionality I would support lists of (weighted) mappings.
mappings = { foo: [bar, baz] } -> send all to bar and baz
mappings = { foo: [foo, baz] } -> send all to foo and baz
mappings = { foo: [bar, [{ dest: baz, weight: 1% }]} -> send all to bar and 1% of traffic to baz (I'd see sampling use cases here)

selectMappedSubject would return more than one subject.
when processing MSG_END_N I would just iterate over that list and assign 
for subj, _ := range c.pa.subjects {
c.pa.subject = subj
if trace {
    c.traceMsg(c.msgBuf)
}
c.processInboundMsg(c.msgBuf)
}

server/opts.go
```diff
@@ -2075,10 +2183,18 @@ func parseAccounts(v interface{}, opts *Options, errors *[]error, warnings *[]er
...
-			msg := fmt.Sprintf("Error adding stream import %q: %v", stream.sub, err)
-			*errors = append(*errors, &configErr{tk, msg})
-			continue
+		if stream.pre != "" {
+			if err := stream.acc.AddStreamImport(ta, stream.sub, stream.pre); err != nil {
+				msg := fmt.Sprintf("Error adding stream import %q: %v", stream.sub, err)
+				*errors = append(*errors, &configErr{tk, msg})
+				continue
+			}
+		} else {
```
matthiashanel, line 2192: Can we just change 
if err := stream.acc.AddStreamImport(ta, stream.sub, stream.pre); err != nil {
To
if err := stream.acc.AddMappedStreamImport(ta, stream.sub, stream.pre+stream.sub); err != nil {
Thus allowing us to delete AddStreamImport (It does not do anything extra), thus delete duplicate code and different function names that would suggest that smth. different happens.

But not a blocker as this is maintainance only

matthiashanel (approved):
Basically LGTM.

The additional comments are suggestions and not blocker.

matthiashanel (approved):
See above comment. (forgot to click approve)

server/accounts_test.go
```diff
@@ -2642,6 +2642,33 @@ func TestAccountRouteMappingsConfiguration(t *testing.T) {
 	}
 }
 
+func TestAccountRouteMappingsWithLossInjection(t *testing.T) {
+	cf := createConfFile(t, []byte(`
+	mappings = {
+		foo: [ { dest: foo, weight: 80% } ]
```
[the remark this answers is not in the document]
derekcollison: ok done.

server/client.go
derekcollison: I think we can reuse a bunch of this for `tee` and `filter` but will leave for now. I would probably make separate config section TBH.

`tee = {
    foo: bar
    baz: [ baz.1, baz.2]
}`

server/opts.go
derekcollison: I wanted to reuse normal addStreamImport and change prefix to `to`..  ;)

The prefix historically could have the token separator `.` or not, so I need to do a bit of pre-processing. I kept error handling in AddStreamImport to keep tests passing that look for specific errors.

server/client.go
derekcollison: FYI checked account reload, if any account section is defined in config we always reloadAuthorization which includes accounts, so no change in behavior with mappings being added.

server/opts.go
matthiashanel: ok

server/client.go
matthiashanel: > I think we can reuse a bunch of this for `tee` and `filter` but will leave for now. I would probably make separate config section TBH.
> 
> `tee = { foo: bar baz: [ baz.1, baz.2] }`

That will raise the question when is it applied in relation to mappings and will have to be explained.
If you only have mappings from one subject to one or more other subjects, some of which can induce loss that question won't come up.
I get that a [] is easily missed. How about an object instead? {tee: []} where the list consists of (weighted) mappings. 
mappings = { foo: {tee: [bar, { dest: baz, weight: 1% }}}

matthiashanel (approved):
LGTM

server/client.go
derekcollison: Possibly, we can punt tee (and filter) to another PR for sure.

server/accounts.go
```diff
@@ -703,12 +743,22 @@ func (a *Account) selectMappedSubject(dest string) (string, bool) {
 	var d *destination
 	var ndest string
 
+	dests := m.dests
+	if len(m.cdests) > 0 {
+		cn := a.srv.ClusterName()
```
matthiashanel, line 748: Lock hierarchie violation?
We hold a.mu, will lock s.mu
```
// clusterName returns our cluster name which could be dynamic.
func (s *Server) ClusterName() string {
	s.mu.Lock()
	cn := s.info.Cluster
	s.mu.Unlock()
	return cn
}
```

matthiashanel (changes requested):

server/accounts.go
derekcollison: It was that or unlock which @kozlovic is not a fan of.. And I can't hold it since its now dynamic (even without reload).

derekcollison: Could create its own lock..

derekcollison: ok fixed.

matthiashanel (approved):
LGTM

matthiashanel (approved):
LGTM

PR #1645, opened 2020-10-15
[Fixed] revocation check for activations used current time instead of jet issue time

matthiashanel:

moved code from checkUserRevoked into separate function to be used by both revocation checks.


kozlovic (approved):
LGTM

PR #1648, opened 2020-10-16
Add support for Causal-Id header

bruth:
 - [ ] Tests added

Resolves #

 -
 -
 -

/cc @nats-io/core

### DIFF
--- a/server/stream.go
+++ b/server/stream.go
@@ -86,6 +86,7 @@ type Stream struct {
 	ddarr     []*ddentry
 	ddindex   int
 	ddtmr     *time.Timer
+	lastMsgId string
 }
 
 // JSPubId is used for identifying published messages and performing de-duplication.
@@ -98,6 +99,10 @@ type ddentry struct {
 	seq uint64
 	ts  int64
 }
+
+// JSCausalId is used to perform an optimistic concurrency control check for a published message.
+const JSCausalId = "Causal-Id"
+const JSCausalAny = "*"
 
 // Replicas Range
 const (
@@ -187,6 +192,7 @@ func (a *Account) AddStreamWithStore(config *StreamConfig, fsConfig *FileStoreCo
 	mset.pubAck = append(mset.pubAck, fmt.Sprintf(" {\"stream\": %q, \"seq\": ", cfg.Name)...)
 
 	// Rebuild dedupe as needed.
+	mset.rebuildLastId()
 	mset.rebuildDedupe()
 
 	// Setup subscriptions
@@ -201,6 +207,23 @@ func (a *Account) AddStreamWithStore(config *StreamConfig, fsConfig *FileStoreCo
 	return mset, nil
 }
 
+// rebuildLastMsg will rebuild any dedupe structures needed after recovery of a stream.
+// Lock not needed, only called during initialization.
+func (mset *Stream) rebuildLastId() {
+	state := mset.store.State()
+	if state.Msgs == 0 {
+		return
+	}
+
+	// Set the last message id.
+	_, hdr, _, _, err := mset.store.LoadMsg(state.LastSeq)
+	if err == nil && len(hdr) > 0 {
+		if msgId := getMsgId(hdr); msgId != "" {
+			mset.lastMsgId = msgId
+		}
+	}
+}
+
 // rebuildDedupe will rebuild any dedupe structures needed after recovery of a stream.
 // Lock not needed, only called during initialization.
 // TODO(dlc) - Might be good to know if this should be checked at all for streams with no
@@ -210,6 +233,7 @@ func (mset *Stream) rebuildDedupe() {
 	if state.Msgs == 0 {
 		return
 	}
+
 	// We have some messages. Lookup starting sequence by duplicate time window.
 	sseq := mset.store.GetSeqFromTime(time.Now().Add(-mset.config.Duplicates))
 	if sseq == 0 {
@@ -529,6 +553,7 @@ func (mset *Stream) Purge() uint64 {
 	for _, o := range obs {
 		o.purge(stats.FirstSeq)
 	}
+	mset.lastMsgId = ""
 	return purged
 }
 
@@ -744,6 +769,15 @@ func getMsgId(hdr []byte) string {
 	return string(getHdrVal(JSPubId, hdr))
 }
 
+// Fast lookup of causalId
+func getCausalId(hdr []byte) (string, bool) {
+	index := bytes.Index(hdr, []byte(JSCausalId))
+	if index < 0 {
+		return "", false
+	}
+	return string(getHdrVal(JSCausalId, hdr)), true
+}
+
 // processInboundJetStreamMsg handles processing messages bound for a stream.
 func (mset *Stream) processInboundJetStreamMsg(_ *subscription, pc *client, subject, reply string, msg []byte) {
 	mset.mu.Lock()
@@ -762,10 +796,11 @@ func (mset *Stream) processInboundJetStreamMsg(_ *subscription, pc *client, subj
 	numConsumers := len(mset.consumers)
 	interestRetention := mset.config.Retention == InterestPolicy
 
-	// Process msgId if we have headers.
+	// Process msgId and causalId if we have headers.
 	var msgId string
 	if pc != nil && pc.pa.hdr > 0 {
 		msgId = getMsgId(msg[:pc.pa.hdr])
+
 		if dde := mset.checkMsgId(msgId); dde != nil {
 			if doAck && len(reply) > 0 {
 				response := append(pubAck, strconv.FormatUint(dde.seq, 10)...)
@@ -775,7 +810,20 @@ func (mset *Stream) processInboundJetStreamMsg(_ *subscription, pc *client, subj
 			mset.mu.Unlock()
 			return
 		}
+
+		// Check if the causal ID is present or not. If not, process
+		// without a check.
+		causalId, present := getCausalId(msg[:pc.pa.hdr])
+		if present && causalId != JSCausalAny && causalId != mset.lastMsgId {
+			if doAck && len(reply) > 0 {
+				response := []byte("-ERR 'causal id conflict'")
+				mset.sendq <- &jsPubMsg{reply, _EMPTY_, _EMPTY_, nil, response, nil, 0}
+			}
+			mset.mu.Unlock()
+			return
+		}
 	}
+
 	mset.mu.Unlock()
 
 	if c == nil {
@@ -820,6 +868,7 @@ func (mset *Stream) processInboundJetStreamMsg(_ *subscription, pc *client, subj
 			}
 			// If we have a msgId make sure to save.
 			if msgId != "" {
+				mset.lastMsgId = msgId
 				mset.storeMsgId(&ddentry{msgId, seq, ts})
 			}
 			// If we are interest based retention and have no consumers clean that up here.
--- a/test/jetstream_test.go
+++ b/test/jetstream_test.go
@@ -2583,6 +2583,133 @@ func TestJetStreamPublishDeDupe(t *testing.T) {
 	}
 }
 
+func TestJetStreamPublishCausalConflict(t *testing.T) {
+	s := RunBasicJetStreamServer()
+	defer s.Shutdown()
+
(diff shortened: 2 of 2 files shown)


bruth:
One caveat to using event sourcing with streams is that the granularity of these conditional checks would ideally by at a subject level. For example, if you has a stream `users` with subjects `users.*`, you would almost always want to do the conditional check at the `user.1` and `user.2`. This would allow events corresponding to those individual subjects to be handled concurrently without conflict (e.g. an update to one user's address does not impact and should not block another user).

Another pattern is to have subjects that represent cross-cutting concerns, for example, `users.email` which may be published to for any email changes across users to ensure uniqueness is enforced.

All that said, now that I have looked into the code, I understand that all messages regardless of subject are processed serially for a stream(?). If per-subject conditional checks were implemented, it could increase the memory overhead (and/or latency) to perform these conditional checks on the subject level. That said, conditional checks should be used sparingly in general, so it would only affect a handful of messages. Basically the client is trading off latency with this causality check.

So to reiterate, this PR demonstrates the check at the stream level, however I would be interested to test out what it would look like on the per-stream-subject level.

bruth:
> One caveat to using event sourcing with streams is that the granularity of these conditional checks would ideally be at a subject level.

The reason I say this is because a _stream_ in the event sourcing context represents a consistency boundary for some entity (which often corresponds to a NATS subject). A NATS stream is a bit broader scope and thus can incorporate many subjects. The incumbent database that implements this is [EventStore](https://eventstore.com) which effectively treats the whole database as one giant event stream (i.e. you can subscribe the `$all` stream), but it supports creating and appending to individual named streams (that are just indexed on top of the `$all` stream).

So in many ways with this additional functionality, each NATS stream would effectively model a single instance of EventStore.

bruth:
Another option to an arbitrary client string is to provide the expected sequence ID of the stream. That would be lighter weight and more predictable with memory consumption.

bruth:
Created a separate branch/PR for subject-level https://github.com/bruth/nats-server/pull/1

bruth:
@ripienaar I consolidated the comments and updated the PR description so its easier to review. I acknowledge there are improvements to the code, but I wanted to get a working example implemented to start the discussion.

server/stream.go
```diff
@@ -99,6 +101,12 @@ type ddentry struct {
 	ts  int64
 }
 
+// JSExpSeq is used to perform an optimistic concurrency control check
+// for a published message. The value of this header should be the current
+// sequence of the stream.
+const JSExpSeq = "Exp-Seq"
+const JSExpSeqStreamPrefix = "!"
```
ripienaar, line 108: would this be better as a stream level config?

ripienaar:
Yeah I like the feature, bit niche but seems that event sourcing as a pattern is picking up adoption speed so worth looking at adding these behaviours it relies on.  Will leave looking into the code details to @derekcollison as I know he's looking into per subject tracking of wildcard streams already and this might slot in there

bruth:
@ripienaar Thanks for reviewing. It is a bit niche, however it could be framed simply as "optimistic concurrency control" on the per stream/subject level so there may be other use cases served. Since there is memory/replay/tracking overhead of the latest sequence per subject, unless that gets optimized somehow, this could be a stream-level config option, e.g. `--enable-exp-seq` which would signal the server to keep track of these sequences, otherwise it would not (the default behavior).

derekcollison:
Yes I am getting into that code as we speak. If ok let's hold for a bit to see how that develops and how this may fit in.

bruth:
@derekcollison Sounds good, thank you

delucca:
👀

derekcollison:
Can you take a look at #1735 and see if this solves the issue?

bruth:
It solves the stream-level sequence check, but not the subject-level sequence check. I do think the subject-level check should be an opt-in thing since it has memory overhead. For example `orders.*` which could be a growing number of subjects over time so the memory issue would need to be solved in a different way so it doesn't simply grow linearly with the number of subjects.

In any case, #1735 is a good improvement and solves the first half of this PR. The subject-level sequence expect header may need more thought.

derekcollison:
Consumers, regardless of how they filter the stream, always have sequence numbers that start at 1 and increase monotonically. 

Which case for consumers are you considering?

bruth:
Ah, I am talking about publishers. The `Nats-Expected-Last-Sequence` header essentially forces multiple clients to serialize their message publishing. For example if I have two clients publishing to `order.1` and `order.2` for stream `ORDERS` matching `orders.*`, they would have to query for the last sequence number on each publish in case the _other_ client published in the meantime.

One way around this would be to have one stream per order, but there is a lot of overhead for that relative to how short-lived a given order might be. So the use case I note above is effectively multiplexing these mini logical streams onto the Nats stream. For cases where coordination between subjects is needed to satisfy some invariant (e.g. uniqueness) a subject like (flipping the domain for a second) `user.email` could be used. Publishers to that subject *should* coordinate so that adding/changing email addresses serialized to ensure uniqueness.

Just to reiterate the pattern I have in mind that this would support, command-query responsibility segregation (CQRS) decouples the write side from the read side. This is commonly paired with Event Sourcing and boils down to two functions. The first is in support of the read/query side and derives *states* from a stream of events.

```
(state, event) -> state'
```

Starting with the empty/nil state and the beginning of a stream of relevant events, apply the event to the state value and return the next version of the state. There can be many implementations each focused on a different subset of a stream or even across streams to satisfy different *read/view* use cases. An implementation of this and the state is produces is typically called a *projection*. This would be analogous to writing various `SELECT` queries in SQL.

The second function is used to actually *produce* the events, e.g. the *write* side. This takes some state, a command from a client/user and produces an event representing the state transition.

```go
(state, command) -> event
```

The scope of this *state* should be as small as possible to satisfy command processing, specifically asserting invariants. For example, if I wanted to maintain the email uniqueness as noted above, I might have a `user.email` subject that all `UserCreated` and `UserEmailChanged` events would be published to. A command handler for `CreateUser` or `ChangeUserEmail` could be implemented to process those commands against the state which could simply be a set/map of email addresses that are currently in use. The `CreateUser` command would fail if the email address is already in case.

The command handler side is where the expected sequence number would be used as part of the event publishing. Once the event is produced from this command handler, it would be published on the `user.email` subject to the `USERS` stream. Although the `Nats-Expected-Last-Sequence` could be used, the sequence number relative to the `user.email` subject is really only needs to be asserted. Other events for `user.1` are generally independent of `user.2`. Again this would be analogous to an OLTP database supports concurrent writes to individual rows in tables. The *table-level* lock comes in when *uniqueness* constraints or other conditions are defined.

Sorry for the long-winded response, but I just wanted to full explain in case you or anyone else reading was not familiar with the pattern or use case. I will say that having the stream level sequence check at least enables to build a layer of subject-level sequence checking on top, but it would not work at seamless and would probably not be efficient once clustering is introduced. Having the subject-level sequence check per stream baked into the server would be more optimal.

derekcollison:
So if I understand correctly, if we simply add on a subject pre-condition header in addition to MsgID and Sequence that should do the trick yes?

derekcollison:
You can combine them with "and" semantics. So you could set both subject and msgId for instance.

bruth:
Yes a subject-level pre-condition would do the trick. 

On November 24, 2020, GitHub Notifications <notifications@github.com>
wrote:
> You can combine them with "and" semantics. So you could set both
> subject and msgId for instance.
>

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/nats-io/nats-server/pull/1648#issuecomment-
733006766>, or unsubscribe
<https://github.com/notifications/unsubscribe-
auth/AAD5YRUV2LFQYDSFLVTYEWDSRO63NANCNFSM4STTSEEA>.

derekcollison:
ok will look more into that. 

So I am understanding clearly, if the following exists in a stream, from oldest to newest. Subj-Seq-MsgID

1-foo-A  2-bar-B   3-baz-C

You would want to be able to add a new message for foo to this stream iff last message with subject foo had msgId A yes?

bruth:
Yes MsgID of A or Seq of 1. MsgID may be more appropriate since that is

client driven but both could satisfy the server side check. 

On November 24, 2020, GitHub Notifications <notifications@github.com>
wrote:
> ok will look more into that.
>
> So I am understanding clearly, if the following exists in a stream,
> from oldest to newest. Subj-Seq-MsgID
>
> 1-foo-A 2-bar-B 3-baz-C
>
> You would want to be able to add a new message for foo to this stream

> iff last message with subject foo had msgId A yes?
>

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/nats-io/nats-server/pull/1648#issuecomment-
733029672>, or unsubscribe
<https://github.com/notifications/unsubscribe-
auth/AAD5YRQGCP3YOX5R6X2KUD3SRPDEVANCNFSM4STTSEEA>.

tpihl:
This is an interesting path that I’m not sure I understand fully 😟 

Are we trying to add atomicity/transactional behavior to an append-only log? While “having it all” is my goal as well, aren’t we trying to implement something that disagrees with NATS core value of prioritize speed/low latency.

The balancing issue with latency vs consistency is described quite good here https://en.wikipedia.org/wiki/PACELC_theorem

I would have assumed a solution not using an (even if optimistic) locking mechanism but rather a Command Queue (or a native nats req/reply) for the client that want to add something for an order, and some server component issue the actual event to the stream. The server component could use any of many distributed/concurrent synchronization pattern (raft, compensation, ...) to avoid putting that logic on the client. Or if the client is a server component, why solve this in jetstream (convenience only or is there other good reasons)?

If the locking was actually needed, this will trigger retries as new requests increasing load and risk of new failed requests, and if it’s solved server side, is there any good option for back-pressure or anything else safeguarding the whole service, specially in a clustered environment?

And if the lock wasn’t needed, we spent resources on something slowing every member of the cluster down for all messages. If we compare with compensation (buzzword saga pattern) a failure would not increase traffic/load for the critical component, in this case jetstream. 

All of the above may be completely irrelevant if the actual cluster implementation (that I know nothing about, but I have very good faith in the developers) solves the consensus/consistency with patterns that support locking concepts. 

/T

tpihl:
And if we can get this without loosing speed, I agree with the per subject in the stream approach 👍

bruth:
@tpihl Good questions and concerns!

FWIW the stream-level check was implemented in #1735. In terms of slow-down the stream sequence number is present, in-memory, for other operations. In addition, a mutex is already being used as well so the added latency for a check is negligible.

My also limited understanding of the clustering implementation is that Raft will be used for the algorithm. Since Raft is leader-based, this check would also only occur on the leader like it does now with a single node. So this would not impact the consensus rounds.

The per-subject check is certainly more challenging to implement and the latency profile for a given stream/application usage would vary quite a bit depending on the number of subjects mapped to a stream and how often the check is used. For example tracking the sequence/msg ID for a 10000 subjects is going to add more overhead than just for 10. But this is still a user decision in how streams are designed and used.

It is worth re-iterating that this is an _opt-in_ check so the happy path is still optimized as it was before. Depending on the implementation, the per-subject check would introduce some memory overhead to keep per-subject sequences available for the check, but it would proportional to the design of the stream.

I will also note that I did evaluate the options of implementing this client-side so it didn't need to be added into NATS. However, what ends up happening is that without the basic compare-and-set (or in this case check-and-append) primitive, you need to serialize your writes on the client side and, with clustering, you would need to re-implement the consensus yourself if you want to scale. So having NATS support this natively (since it has all the knowledge and necessary locks already), this brings a big relief to clients to certain kinds of workloads.

Even with the #1735 change, using NATS with streams as an lightweight event store is possible. Adding the per-subject check simply increases the observed concurrency/reduced contention than the single stream-level check.

tpihl:
@bruth I actually pre-agreed 😄 since what you say is handled by a lock-permitting cluster (elected leader is one of those). Should the cluster however be implemented with some CRDT or similar solution or something with majority confirm per write to allow horizontal scaling, this might be a concern. I am trying to keep an open mind and hope for a very scalable cluster backend to allow high speed persistence handling write io cost by multiple instances. But who knows, maybe it will be a single active jetstream instance only... (hope not)

bruth:
@tpihl For sure! A leader-less clustering approach would certainly make require a different approach.

PR #1651, opened 2020-10-19
add type hints num accounts message

ripienaar:

/cc @nats-io/core


ripienaar:
I want to show these in `nats event`

matthiashanel (approved):
LGTM

PR #1652, opened 2020-10-19
[FIXED] Possible panic when server accepts TLS leafnode connection

kozlovic:

### DIFF
--- a/server/leafnode.go
+++ b/server/leafnode.go
@@ -617,6 +617,17 @@ func (s *Server) createLeafNode(conn net.Conn, remote *leafNodeCfg) *client {
 	// Grab lock
 	c.mu.Lock()
 
+	// If client has been closed, this function will unlock and make sure
+	// the the connection is closed for proper clean-up.
+	isClosedUnlockAndReturn := func() bool {
+		if c.isClosed() {
+			c.mu.Unlock()
+			c.closeConnection(WriteError)
+			return true
+		}
+		return false
+	}
+
 	if solicited {
 		// We need to wait here for the info, but not for too long.
 		c.nc.SetReadDeadline(time.Now().Add(DEFAULT_LEAFNODE_INFO_WAIT))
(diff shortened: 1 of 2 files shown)


server/leafnode.go
```diff
@@ -649,6 +649,14 @@ func (s *Server) createLeafNode(conn net.Conn, remote *leafNodeCfg) *client {
 			return nil
 		}
 
+		// Not sure that can happen, but in case the connection was marked
+		// as closed during In case connection was marked as closed during parse...
```
derekcollison, line 653: In -> in

derekcollison (commented):
Does this solve for the panic though and the stack trace we saw in Slack?

server/leafnode.go
kozlovic: whole comment is wrong :-) will fix that.

philpennock (approved):

kozlovic:
@derekcollison The panic was for the non solicited. When calling `sendProtoNow()` this could have marked the connection as closed right there if there was a TCP write error.

kozlovic:
Pushing the change for the wrong comment..

server/leafnode.go
```diff
@@ -649,6 +649,14 @@ func (s *Server) createLeafNode(conn net.Conn, remote *leafNodeCfg) *client {
 			return nil
 		}
 
+		// Not sure that can happen, but in case the connection was marked
+		// as closed during the call to parse...
+		if c.isClosed() {
```
derekcollison, line 654: Let's make this a small anonymous function inside this function to clean up?

`ifClosedReturn := func() `

derekcollison (approved):
LGTM, maybe make that a small internal function inside the create function.

server/leafnode.go
kozlovic: I would still need to have something like:
```
if closedReturn() {
   return nil
}
```

derekcollison (approved):
LGTM

PR #1653, opened 2020-10-19
Fix nil deref when account is in bad/incomplete state

matthiashanel:

This endpoint is for debugging, so we shouldn't rely on things not being nil.
Also include validation results and missing mapping

### DIFF
--- a/server/monitor.go
+++ b/server/monitor.go
@@ -2024,25 +2024,16 @@ func (s *Server) accountInfo(accName string) (*AccountInfo, error) {
 	}
 	exports := []ExtExport{}
 	for k, v := range a.exports.services {
-		var e ExtExport
-		subj := jwt.Subject(k)
-		if v == nil {
-			e = ExtExport{
-				Export: jwt.Export{
-					Subject: subj,
-					Type:    jwt.Service,
-				},
-			}
-		} else {
-			e = ExtExport{
-				Export: jwt.Export{
-					Subject:      subj,
-					Type:         jwt.Service,
-					TokenReq:     v.tokenReq,
-					ResponseType: jwt.ResponseType(v.respType.String()),
-				},
-				ApprovedAccounts: []string{},
-			}
+		e := ExtExport{
+			Export: jwt.Export{
+				Subject: jwt.Subject(k),
+				Type:    jwt.Service,
+			},
+			ApprovedAccounts: []string{},
+		}
+		if v != nil {
+			e.TokenReq = v.tokenReq
+			e.ResponseType = jwt.ResponseType(v.respType.String())
 			for name := range v.approved {
 				e.ApprovedAccounts = append(e.ApprovedAccounts, name)
 			}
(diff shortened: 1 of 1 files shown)


server/monitor.go
```diff
@@ -2005,68 +2012,121 @@ func (s *Server) accountInfo(accName string) (*AccountInfo, error) {
...
-				Subject:      jwt.Subject(k),
-				Type:         jwt.Service,
-				TokenReq:     v.tokenReq,
-				ResponseType: jwt.ResponseType(v.respType.String()),
-			},
-			ApprovedAccounts: []string{},
-		}
-		for name := range v.approved {
-			e.ApprovedAccounts = append(e.ApprovedAccounts, name)
+		var e ExtExport
```
kozlovic, line 2027: Would this be easier to read?
```
e := ExtExport{
   Subject: jwt.Subject(k),
   Type: jwt.Service,
}
if v != nil {
  e.TokenReq = v.tokenReq
  e.ResponseType = ..
  ...
}
```
Basically create the export with the fields that don't depend on `v` being nil or not, then complete in the `if v != nil` section.

server/monitor.go
```diff
@@ -2005,68 +2012,121 @@ func (s *Server) accountInfo(accName string) (*AccountInfo, error) {
...
+				ApprovedAccounts: []string{},
+			}
+			for name := range v.approved {
+				e.ApprovedAccounts = append(e.ApprovedAccounts, name)
+			}
 		}
 		exports = append(exports, e)
 	}
 	for k, v := range a.exports.streams {
+		token := false
```
kozlovic, line 2053: Similar to previous comment. Would create the ExtExport{} with common stuff then fill `if v != nil`

server/monitor.go
```diff
@@ -2005,68 +2012,121 @@ func (s *Server) accountInfo(accName string) (*AccountInfo, error) {
...
+			}
 		}
 		exports = append(exports, e)
 	}
 	imports := []ExtImport{}
 	for _, v := range a.imports.streams {
-		imports = append(imports, ExtImport{
-			Import: jwt.Import{
+		invalid := true
+		var imp jwt.Import
```
kozlovic, line 2077: If v is nil, then imp will be an empty jwt.Import structure. Wajt doe this mean to add it to the list of imports? We will have it as `invalid` but nothing from the Import details.

server/monitor.go
```diff
@@ -2005,68 +2012,121 @@ func (s *Server) accountInfo(accName string) (*AccountInfo, error) {
...
+		}
+		imports = append(imports, ExtImport{
+			Import:  imp,
+			Invalid: invalid,
 		})
 	}
 	for _, v := range a.imports.services {
-		imports = append(imports, ExtImport{
-			Import: jwt.Import{
+		invalid := true
```
kozlovic, line 2092: Same comment than above. What does it mean to add an ExtImport with an empty jwt.Import structure?

kozlovic (commented):
Some code readability suggestions and really wonder what it means to add a ExtImport with an empty jwt.Import struct..

server/monitor.go
matthiashanel: fixed, essentially creating the empty export, overwriting in case of v != nil

server/monitor.go
matthiashanel: fixed, same as above

server/monitor.go
matthiashanel: the use case of /accountz is debugging, so right now I don't know what to expect.
The important bit would be to communicate that it's there but smth is wrong.
just that, smth is wrong invalid = true. no other details to be had.

server/monitor.go
matthiashanel: same as above

kozlovic (approved):
LGTM

PR #1654, opened 2020-10-19
Fixed account update issue where signing key (and thus issuer) changes

matthiashanel:
Fix error and remove print from test.

My new found jwt knowledge made me realize that the issuer could change.
What needs to remain constant is the subject.
At the time of test we know we trust the jwt.

Internally acc.Name is set from accClaims.Subject 
The condition is to prevent one valid jwt becoming another by modifying the subject.


kozlovic (approved):
LGTM

PR #1655, opened 2020-10-20
Adding support and an option for removal of jwt

matthiashanel:
To enable in full mode configure allow_delete: true
When enabled the file will be renamed to allow for manual restore.

In cache mode it will be enabled by default.
When enabled files will be deleted.

For now deletes need to be done using the operator identity key.
Going forward we can allow signing keys to remove accounts issued by them while the operator identity key can be used to delete any jwt.

### DIFF
--- a/server/accounts.go
+++ b/server/accounts.go
@@ -32,6 +32,7 @@ import (
 	"time"
 
 	"github.com/nats-io/jwt/v2"
+	"github.com/nats-io/nkeys"
 	"github.com/nats-io/nuid"
 )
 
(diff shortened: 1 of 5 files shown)


server/accounts.go
```diff
@@ -3301,14 +3302,48 @@ func handleListRequest(store *DirJWTStore, s *Server, reply string) {
...
+		} else if accIds, ok = list.([]interface{}); !ok {
+			err = fmt.Errorf("malformed request")
+		} else {
+			for _, entry := range accIds {
+				if acc, ok := entry.(string); !ok ||
+					acc == "" || !nkeys.IsValidPublicAccountKey(acc) {
+					err = fmt.Errorf("malformed request")
+					break
+				} else if acc == sysAccName {
+					err = fmt.Errorf("not allowed to delete system account")
```
kozlovic, line 3334: Should you not break?

server/dirstore.go
```diff
@@ -135,16 +145,17 @@ func NewDirJWTStore(dirPath string, shard bool, create bool) (*DirJWTStore, erro
 // evictOnLimit determines the behavior once limit is reached.
 //     true - Evict based on lru strategy
 //     false - return an error
-func NewExpiringDirJWTStore(dirPath string, shard bool, create bool, expireCheck time.Duration, limit int64,
+func NewExpiringDirJWTStore(dirPath string, shard bool, create bool, delete deleteType, expireCheck time.Duration, limit int64,
```
kozlovic, line 148: Since those are public (exported) functions, but not yet released, should we have an option object/array of functions here to be able to extend in the future?

kozlovic (changes requested):

server/dirstore.go
matthiashanel: Yes I can add that. array or map[string]interface{}?

server/accounts.go
matthiashanel: yes, fixed

server/dirstore.go
matthiashanel: went with  `_ ...interface{}`

kozlovic: No, I meant to pass an "Options" structure that we can add new options, or functions like we do for say nats.Connect(), etc..

server/accounts.go
```diff
@@ -3544,21 +3600,26 @@ func (s *Server) fetch(res AccountResolver, name string) (string, error) {
 	return theJWT, err
 }
 
-func NewCacheDirAccResolver(path string, limit int64, ttl time.Duration) (*CacheDirAccResolver, error) {
+func NewCacheDirAccResolver(path string, limit int64, ttl time.Duration, _ ...interface{}) (*CacheDirAccResolver, error) {
```
kozlovic, line 3603: No, this is not I had in mind. As explained on Slack, if you want to make it extensible in the future, you should replace all params with a public option structure, or a variadic of options. If you don't plan to add anything after the public release, then remove and do nothing.

server/dirstore.go
```diff
@@ -103,8 +105,8 @@ func newDir(dirPath string, create bool) (string, error) {
 
 // Creates a directory based jwt store.
 // Reads files only, does NOT watch directories and files.
-func NewImmutableDirJWTStore(dirPath string, shard bool) (*DirJWTStore, error) {
-	theStore, err := NewDirJWTStore(dirPath, shard, false)
+func NewImmutableDirJWTStore(dirPath string, shard bool, _ ...interface{}) (*DirJWTStore, error) {
```
kozlovic, line 108: No, remove this.

kozlovic (changes requested):
Don't add []interface{}.

server/dirstore.go
```diff
@@ -101,10 +103,13 @@ func newDir(dirPath string, create bool) (string, error) {
 	return fullPath, nil
 }
 
+// future proving in case new options will be added
```
kozlovic, line 106: would that be `future proofing`?

server/accounts.go
kozlovic: Should you remove from this one? Is this a jwt store? If not, then you should not be using that, otherwise, it should be `_ ...dirJWTStoreOption` with your current approach.

kozlovic (changes requested):
Not sure if you forgot to remove/change the use of `...interface{}`.
For the record, we discussed the approach of `...dirJWTStoreOption` and I am not sold on that. To me, a simple public jwt options struct would have been better (possibly removing lots of the current parameters especially for NewExpiringDirJWTStore() (and possibly having dedicated options if some of the options don't make sense from a type of store to the other).

kozlovic (approved):
LGTM

PR #1658, opened 2020-10-20
[JetStream] Upgrade for wait queues for pull based consumers

derekcollison:
Upgraded the wait queue to be more robust and process formal requests with options for batch size, no waiting on messages and expiration.

/cc @nats-io/core


server/consumer.go
```diff
@@ -1031,47 +1051,174 @@ func (o *Consumer) needAck(sseq uint64) bool {
...
+		// If our waiting queue is full return an empty response with the proper header.
+		// FIXME(dlc) - Should we do advisory here as well?
+		sendErr(500, "WaitQueue Exceeded")
+		return
+	}
+
+	// Check payload here to see if they sent in batch size or a formal request.
+	expires, batchSize, noWait, err := nextReqFromMsg(msg)
+	if err != nil {
+		sendErr(400, "Bad Request")
```
ripienaar, line 1192: perhaps an opportunity to make the reason for failure clearer here? Request JSON invalid or invalid batch size number or similar?

server/consumer.go
```diff
@@ -1031,47 +1051,174 @@ func (o *Consumer) needAck(sseq uint64) bool {
...
+		o.mu.Unlock()
+		hdr := []byte(fmt.Sprintf("NATS/1.0 %d %s\r\n\r\n", status, description))
+		pmsg := &jsPubMsg{reply, reply, _EMPTY_, hdr, nil, nil, 0}
+		sendq <- pmsg // Send message.
+	}
+
+	if o.waiting.isFull() {
+		// If our waiting queue is full return an empty response with the proper header.
+		// FIXME(dlc) - Should we do advisory here as well?
+		sendErr(500, "WaitQueue Exceeded")
```
ripienaar, line 1185: So when a user gets to this situation what can he do to resolve this error? Basically nothing right? Would it not be better to before sending this cycle through and check which of the waiting requests have no interest and discard them - or maybe just shift the oldest one off? 

Worried about the UX of this going full and users just being stuck with nothing they can do, some rogue/badly behaving consumer can DOS this easily this way.

Also I think this would be something people will want to change at runtime so perhaps the consumer max waiting would end up being something a consumer udpate can edit.

But as per slack I am not sure I fully grasp what the use case you were aiming at here so lets catch up and make sure I understand what this is intended to achieve

server/jetstream_api.go
```diff
@@ -413,6 +416,13 @@ type JSApiConsumerListResponse struct {
 
 const JSApiConsumerListResponseType = "io.nats.jetstream.api.v1.consumer_list_response"
 
+// JSApiConsumerGetNextRequest is for getting next messages for pull based consumers.
+type JSApiConsumerGetNextRequest struct {
+	Expires time.Time `json:"expires,omitempty"`
```
ripienaar, line 421: difficult for other languages, can we make this seconds?

ripienaar (changes requested):
some comments and clarifications but I like where this is going

server/consumer.go
derekcollison: Possibly, although that is the server spending time and we really do not know much more than that. Could distinguish a bit more but not sure its worth it.

server/consumer.go
derekcollison: You are correct, needed to add one more test to take care of that and try to purge requests before failing. Will add that test and change now.

server/jetstream_api.go
derekcollison: This is a common RFC for time, but yes we could. Should look into how other represent timeouts.

server/consumer.go
derekcollison: After our call I agree with you we should pop oldest and signal and never return a requests full error. Stabilizes the clients much better, otherwise the later ones would spin on the full error.

server/consumer.go
ripienaar: It's not a lot of extra work - we already return a err there, sute its a bit no question but it would be a big help for users when debugging things.

derekcollison: We only have two options, could not parse JSON or raw batchSize number..

ripienaar: Yup. So should be easy to add user friendly behaviour :) it’s not like users know we have only these 2 without reading code

ripienaar:
LGTM

PR #1659, opened 2020-10-21
Do not forward service import messages to a stream export.

derekcollison:
Addresses stack overflow issue wally was seeing with configs
that mix and match streams and services between each other.

/cc @nats-io/core


wallyqs (approved):
LGTM

kozlovic (approved):
LGTM!

PR #1660, opened 2020-10-21
Prevent panic when accepting TLS leafnode connections

kozlovic:
This is an addition to PR #1652. I have simply added a check but
at this point in time there is no risk that connection is closed
this early.
I also renamed the small helper function and fixed a test that
had an improper `s.mu.Unlock()` in an error condition.


derekcollison (approved):
LGTM

PR #1662, opened 2020-10-22
[FIXED] Do not report bad latency results on auto-unsubscribe triggers

derekcollison:
When old requests were utilized and the requestor was on a different server then the responder, the auto-unsubscribe log on the requestors server when the reply was delivered would trigger a behavior that would report bad latency results since we did not have our second measurement m2 from the remote server yet.

This fixes that by tracking if we have delivered a response to the request, and if so will wait for the m2 measurement to arrive and not report bad results immediately.

/cc @nats-io/core

### DIFF
--- a/server/accounts.go
+++ b/server/accounts.go
@@ -130,6 +130,7 @@ type serviceImport struct {
 	invalid     bool
 	share       bool
 	tracking    bool
+	didDeliver  bool
 	trackingHdr http.Header // header from request
 }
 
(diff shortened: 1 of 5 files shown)


ripienaar (approved):
LGTM, tested my setup using this branch and all is fine

derekcollison:
I pushed a small change to a flapper test but decided to squash these all down into one commit.

kozlovic (approved):
LGTM

PR #1666, opened 2020-10-23
Add store.SkipMsg() and update no interest retention streams

derekcollison:
Previously interest based retention streams would save and the delete a message from the store if there were no consumers interested. This could lead to issues with messages showing up in the stream when none should be there.

This could also be the effect seen here: https://github.com/nats-io/jetstream/issues/314

We introduce SkipMsg() for the store interface to allow bumping the sequence number but never saving a message, which will guarantee there can never be a state observation of non-zero.

During this investigation I also realized that the filestore did not properly select a new starting sequence upon removal if the removal's block and the true next first sequence were not in the same message block, so fixed that too.

/cc @nats-io/core

### DIFF
--- a/server/filestore.go
+++ b/server/filestore.go
@@ -639,6 +639,26 @@ func (fs *fileStore) StoreMsg(subj string, hdr, msg []byte) (uint64, int64, erro
 	return seq, ts, nil
 }
 
+// SkipMsg will use the next sequence number but not store anything.
+func (fs *fileStore) SkipMsg() uint64 {
+	// Grab time.
+	now := time.Now().UTC()
+	fs.mu.Lock()
+	seq := fs.state.LastSeq + 1
+	fs.state.LastSeq = seq
+	fs.state.LastTime = now
+	if fs.state.Msgs == 0 {
+		fs.state.FirstSeq = seq
+		fs.state.FirstTime = now
+	}
+	if seq == fs.state.FirstSeq {
+		fs.state.FirstSeq = seq + 1
+		fs.state.FirstTime = now
+	}
+	fs.mu.Unlock()
+	return seq
+}
+
 // Will check the msg limit and drop firstSeq msg if needed.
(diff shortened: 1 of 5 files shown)


ripienaar (approved):
LGTM, solves the problem and makes it all nice and fast.

server/filestore.go
```diff
@@ -639,6 +639,26 @@ func (fs *fileStore) StoreMsg(subj string, hdr, msg []byte) (uint64, int64, erro
...
+	now := time.Now().UTC()
+	fs.mu.Lock()
+	seq := fs.state.LastSeq + 1
+	fs.state.LastSeq = seq
+	fs.state.LastTime = now
+	if fs.state.Msgs == 0 {
+		fs.state.FirstSeq = seq
+		fs.state.FirstTime = now
+	}
+	if seq == fs.state.FirstSeq {
```
kozlovic, line 654: I fail to see when this condition can be true with state.Msgs != 0? Could you describe in which situation would we have one of these 2 "if"'s true, but not the other? (disclaimer, I am not versed on the internal of JS file store implementation).

server/memstore.go
```diff
@@ -297,6 +315,29 @@ func (ms *memStore) EraseMsg(seq uint64) (bool, error) {
 	return removed, nil
 }
 
+// Performs logic tp update first sequence number.
```
kozlovic, line 318: tp -> to?

derekcollison: Thanks will fix.

server/filestore.go
derekcollison: 0 is not legal and denotes the start, so that case is special opt. When running, if you have been using a stream for awhile but all messages are now consumed, expired etc. the first seq will be last seq + 1, meaning you are caught up. So if I send 10 messages and consume them, the state of the stream is first = 11, last = 10.

PR #1668, opened 2020-10-23
Purge clears dedupe window as well

derekcollison:

Resolves https://github.com/nats-io/jetstream/issues/340

/cc @nats-io/core


ripienaar (approved):
LGTM

PR #1669, opened 2020-10-23
use milliseconds for pull expire

ripienaar:

/cc @nats-io/core

### DIFF
--- a/server/consumer.go
+++ b/server/consumer.go
@@ -1062,7 +1062,7 @@ func nextReqFromMsg(msg []byte) (time.Time, int, bool, error) {
 		if err := json.Unmarshal(msg, &cr); err != nil {
 			return time.Time{}, -1, false, err
 		}
-		return cr.Expires, cr.Batch, cr.NoWait, nil
+		return time.Now().Add(time.Duration(cr.Expires) * time.Millisecond), cr.Batch, cr.NoWait, nil
 	}
 	// Naked batch size here for backward compatibility.
 	bs := 1
--- a/server/jetstream_api.go
+++ b/server/jetstream_api.go
@@ -418,9 +418,9 @@ const JSApiConsumerListResponseType = "io.nats.jetstream.api.v1.consumer_list_re
 
 // JSApiConsumerGetNextRequest is for getting next messages for pull based consumers.
 type JSApiConsumerGetNextRequest struct {
-	Expires time.Time `json:"expires,omitempty"`
-	Batch   int       `json:"batch,omitempty"`
-	NoWait  bool      `json:"no_wait,omitempty"`
+	Expires int  `json:"expire_ms,omitempty"`
+	Batch   int  `json:"batch,omitempty"`
+	NoWait  bool `json:"no_wait,omitempty"`
 }
 
 // JSApiStreamTemplateCreateResponse for creating templates.
--- a/test/jetstream_test.go
+++ b/test/jetstream_test.go
@@ -1662,7 +1662,7 @@ func TestJetStreamWorkQueueRequest(t *testing.T) {
 			// Now do expiration
 			req.Batch = 1
 			req.NoWait = false
-			req.Expires = time.Now().Add(10 * time.Millisecond)
+			req.Expires = 10
 			jreq, _ = json.Marshal(req)
 
 			nc.PublishRequest(getSubj, reply, jreq)


server/consumer.go
```diff
@@ -1062,7 +1062,7 @@ func nextReqFromMsg(msg []byte) (time.Time, int, bool, error) {
 		if err := json.Unmarshal(msg, &cr); err != nil {
 			return time.Time{}, -1, false, err
 		}
-		return cr.Expires, cr.Batch, cr.NoWait, nil
+		return time.Now().Add(time.Duration(cr.Expires) * time.Millisecond), cr.Batch, cr.NoWait, nil
```
ripienaar, line 1065: I feel we need to impose a lower limit here maybe? What about a max and default for when none is set?

derekcollison: This would be the only place we ever express timeouts in ms..

server/jetstream_api.go
```diff
@@ -418,9 +418,9 @@ const JSApiConsumerListResponseType = "io.nats.jetstream.api.v1.consumer_list_re
 
 // JSApiConsumerGetNextRequest is for getting next messages for pull based consumers.
 type JSApiConsumerGetNextRequest struct {
-	Expires time.Time `json:"expires,omitempty"`
-	Batch   int       `json:"batch,omitempty"`
-	NoWait  bool      `json:"no_wait,omitempty"`
+	Expires int  `json:"expire_ms,omitempty"`
```
derekcollison, line 421: See comment above.

ripienaar: Yeah I realise. We can try change them all? But asking people to state this kind of thing jn nano seconds (duration) is really not very friendly and neither is time.Time

derekcollison: Its expressed as a deadline time via RFC spec.. No ns..

ripienaar: Sure this one isn’t ns but I mean the problem in the general case here applies to time and duration cases.

derekcollison: That is why I did a deadline and not expiration/ttl.

server/consumer.go
aricart: I think we also have nanosecond resolution on duplicate_window, the time stamp of a message, optional start time for playing a stream.

server/jetstream_api.go
aricart: If we do "#{unit}" with a default of ms for a number, I think is a reasonable option for configuration options. I am even ready to concede that timestamps can be in ns, and the library provides a way to convert to the native time representation.

Start time for playback will likely be used often - and UNIX time seems a reasonable resolution - if we are doing formatted ISO date strings, etc, then that itself might be another challenge for the different client languages.

ripienaar: This should be a duration if possible. But yeah rather time than ns. In JavaScript it’s just the output from toISOString()

server/consumer.go
derekcollison: start time is a time like the deadline, but I think you may be right on the duplicate window maybe. I am not saying we should not change, I am simply suggesting we do them all at one instead of just this one.

server/jetstream_api.go
derekcollison: Let's maybe sync up Monday am and discuss.

server/consumer.go
aricart: I agree. I just shot the ones that I remembered.

server/jetstream_api.go
ripienaar: Ones to consider in addition to this one:

ConsumerConfig.OptStartTime -> unix seconds
ConsumerConfig.AckWait -> int as seconds
ConsmerInfo.Created -> unix nano
StoredMsg.Time -> unix nano
StreamConfig.MaxAge -> seconds 
StreamConfig.Duplicates -> seconds
StreamState.FirstTime -> unix nano
StreamState.LastTime -> unix nano

derekcollison: ok this is helpful. We should rationalize sooner rather than later..

ripienaar: Some more, but less nb:

TypedEvent.Time -> unix nano
JSRestoreCompleteAdvisory.Start and End -> unix nano
JSSnapshotCompleteAdvisory.Start and End -> unix nano

derekcollison: If they are time.Time then that is ok since we use RFC to format and most (I think all) clients can read and parse and will do the right thing with nanos for their platform.

The time.Duration ones are the ones I am concerned about resolving in a consistent manner.

aricart: > JSRestoreCompleteAdvisory.Start and End -> unix nano
> JSSnapshotCompleteAdvisory.Start and End -> unix nano

Since these are generic operations for a time, possibly UnixTime on those? - I understand in highspeed messages the timestamp is possibly better as nanos, but this is a tool that will report/start/end times. Regardless, this particular setting is not critical.

ColinSullivan1: ms is good imo.

We'll want to do some research into clients.  It seems java and .NET prefer ISO-8601, although RFC 3339 can be parsed in other ways.

ColinSullivan1: rfc 3339 is poorly supported on some platforms.  tbh it's been problematic on java, and for .NET it's looking like it'll be challenging.  I think it'll be a much lower level of effort in client development using nanos since epoch.

ripienaar: So to avoid guesswork about units lets always use nanos like:

```
ConsumerConfig.OptStartTime -> unix nano seconds since epoch
ConsumerConfig.AckWait -> unix nano as duration
ConsmerInfo.Created -> unix nano seconds since epoch
StoredMsg.Time -> unix nano seconds since epoch
StreamConfig.MaxAge -> unix nano as duration
StreamConfig.Duplicates -> unix nano as duration
StreamState.FirstTime -> unix nano seconds since epoch
StreamState.LastTime -> unix nano seconds since epoch
TypedEvent.Time -> unix nano seconds since epoch
JSRestoreCompleteAdvisory.Start and End -> unix nano seconds since epoch
JSSnapshotCompleteAdvisory.Start and End -> unix nano seconds since epoch
```

So all things are always nanos, it's just a choice between durations or timestamps

ripienaar: @derekcollison would you be happy if I send a PR with this change?

ripienaar: @aricart any concerns with above?

derekcollison: Not yet, let's discuss today.

PR #1673, opened 2020-10-26
Fixed no_advertise comment in `nats-server -help`

### LINKED ISSUE #1647: --no_advertise flag documented twice with different descriptions
Hi, 

We have notice that the `--no_advertise` flag is documented twice once in [1] described as "Do not advertise known cluster information to clients" and once in [2] described as "When set to true, do not advertise this server to clients" . We have checked in code and were only able to find the [flag](https://github.com/nats-io/nats-server/blob/e846dab45dad399eefb0ac46b05be7c92c56e568/main.go#L70) as a cluster option which matches the former description in the documentation.

We are raising the question because the option has been recently added to the nats-release with this [PR](https://github.com/cloudfoundry/nats-release/commit/8cd7d93a8fc29d8226a31a5aecd87e8701da3c0e). The flag is introduced with the description "When configured to true, this nats server will not be advertised to any nats clients." and we are not sure how to consume it.

Could you please clarify what is exactly the expected behaviour if the flag is set to true? which of these descriptions is valid? 

[1] https://docs.nats.io/nats-server/flags#cluster-options
[2] https://docs.nats.io/nats-server/configuration/clustering/cluster_config

kozlovic:
Resolves #1647

[ci skip]

### DIFF
--- a/main.go
+++ b/main.go
@@ -67,7 +67,7 @@ Cluster Options:
         --routes <rurl-1, rurl-2>    Routes to solicit and connect
         --cluster <cluster-url>      Cluster URL for solicited routes
         --cluster_name <string>      Cluster Name, if not set one will be dynamically generated
-        --no_advertise <bool>        Advertise known cluster IPs to clients
+        --no_advertise <bool>        Do not advertise known cluster information to clients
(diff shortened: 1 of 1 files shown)


derekcollison (approved):
LGTM

PR #1675, opened 2020-10-27
Stability and performance updates for filestore.

derekcollison:
The original design had a shared filestore write buffer and individual message blocks had a read cache.
This presented some performance and stability issues when lots of reads and deletes were happening to a
message block that was also being written to actively.

This change eliminates the shared write buffer and uses the message block's cache as a write through as
well as read cache and handles partials correctly.

This PR also introduces a dynamic autotuned block size for file based streams aiming at 25% overhead in most cases.
We also now dynamically discover the available disk space by storage directory.

/cc @nats-io/core

### DIFF
--- a/server/filestore.go
+++ b/server/filestore.go
@@ -911,11 +911,12 @@ func (fs *fileStore) removeMsg(seq uint64, secure bool) (bool, error) {
 	if firstSeqNeedsUpdate {
 		fs.selectNextFirst()
 	}
+	cb := fs.scb
 	fs.mu.Unlock()
 
-	if fs.scb != nil {
+	if cb != nil {
 		delta := int64(msz)
-		fs.scb(-delta)
+		cb(-1, -delta, seq)
 	}
 
 	return true, nil
@@ -1113,9 +1114,11 @@ func (mb *msgBlock) expireCache() {
 	mb.mu.Lock()
 	defer mb.mu.Unlock()
 
-	if mb.cache == nil && mb.ctmr != nil {
-		mb.ctmr.Stop()
-		mb.ctmr = nil
+	if mb.cache == nil {
+		if mb.ctmr != nil {
+			mb.ctmr.Stop()
+			mb.ctmr = nil
+		}
 		return
 	}
 
--- a/server/filestore.go
+++ b/server/filestore.go
@@ -839,10 +839,23 @@ func (fs *fileStore) removeMsg(seq uint64, secure bool) (bool, error) {
 		}
 	}
 
+	// Set cache timestamp for last remove.
+	mb.lrts = time.Now().UnixNano()
+
 	// Grab record length from idx.
 	slot := seq - mb.cache.fseq
 	ri, rl, _, _ := mb.slotInfo(int(slot))
 	msz := uint64(rl)
+
+	// Call this early so that we can look this message up if needed.
+	if cb := fs.scb; cb != nil {
+		mb.mu.Unlock()
+		fs.mu.Unlock()
+		delta := int64(msz)
+		cb(-1, -delta, seq)
+		fs.mu.Lock()
+		mb.mu.Lock()
+	}
 
 	// Global stats
 	fs.state.Msgs--
@@ -851,9 +864,6 @@ func (fs *fileStore) removeMsg(seq uint64, secure bool) (bool, error) {
 	// Now local mb updates.
 	mb.msgs--
 	mb.bytes -= msz
-
-	// Set cache timestamp for last remove.
-	mb.lrts = time.Now().UnixNano()
 
 	var shouldWriteIndex bool
 	var firstSeqNeedsUpdate bool
@@ -913,13 +923,7 @@ func (fs *fileStore) removeMsg(seq uint64, secure bool) (bool, error) {
 	if firstSeqNeedsUpdate {
 		fs.selectNextFirst()
 	}
-	cb := fs.scb
 	fs.mu.Unlock()
-
-	if cb != nil {
-		delta := int64(msz)
-		cb(-1, -delta, seq)
-	}
 
 	return true, nil
 }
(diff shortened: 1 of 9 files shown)


server/stream.go
```diff
@@ -201,6 +206,68 @@ func (a *Account) AddStreamWithStore(config *StreamConfig, fsConfig *FileStoreCo
...
+// Helper to determine the max msg size for this stream if file based.
+func (mset *Stream) maxMsgSize() uint64 {
+	maxMsgSize := mset.config.MaxMsgSize
+	if maxMsgSize <= 0 {
+		// Pull from server.
+		if mset.jsa != nil {
+			if acc := mset.jsa.acc(); acc != nil {
+				if s := acc.server(); s != nil {
+					opts := s.getOpts()
+					maxMsgSize = opts.MaxPayload
```
kozlovic, line 218: Is the MaxPayload not possibly a limit set in the account JWT?

kozlovic (approved):
Just wonder if using max_payload from server is always the right thing to do?

server/filestore.go
```diff
@@ -639,6 +685,35 @@ func (fs *fileStore) StoreMsg(subj string, hdr, msg []byte) (uint64, int64, erro
 	return seq, ts, nil
 }
 
+// skipMsg will update message block for a skipped message. If first
```
kozlovic, line 688: Can you rephrase starting at "If first.."? I can't understand that sentence.

server/stream.go
derekcollison: Good point, should pull from internal client or account first. Thanks will update.

server/filestore.go
derekcollison: Yes will rephrase. If block is empty them this is just a metadata update, no placeholder records need to be written.

server/filestore.go
```diff
@@ -819,137 +1078,73 @@ func (fs *fileStore) selectNextFirst() {
...
-			mb.qch = make(chan struct{})
-			mb.dch = make(chan struct{})
-			go fs.flushWriteIndexLoop(mb, mb.dch, mb.qch)
-			// Do a blocking kick here.
-			mb.dch <- struct{}{}
-		} else {
-			mb.kickWriteFlusher()
-		}
+	if mb.cache == nil {
+		mb.ctmr = nil
```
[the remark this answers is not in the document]
derekcollison: Yes I could do that.

derekcollison:
I just noticed a performance regression, will need to address.

derekcollison:
ok performance regression has been identified.

kozlovic (approved):
LGTM

kozlovic (approved):
LGTM

kozlovic (approved):
LGTM.. that being said, this is just moving the timing a little bit, meaning if that was the source for flappers, I think that the tests should be tuned instead.

derekcollison:
It was a bug for sure. I have one more after the merge w/ pending I am looking at as well.

kozlovic:
> It was a bug for sure. I have one more after the merge w/ pending I am looking at as well.

Then how this is fixing anything? You just guarantee that the go routine is started and reached the point of closing the given channel, but after that the exact same timing can occur. Basically imagine a big sleep in the go routine after closing the channel and before the "for loop" starts. If that breaks things (as in: this is causing a bug), then this change does not fix anything. I assumed that this was just to fix some tests timing issues where a check would be made before the flusher had a chance to write things.

derekcollison:
When running constrained such that the Go routine could not be scheduled quick enough we would miss flush events and leave caches that needed to be written stale.

server/filestore.go
```diff
@@ -229,7 +229,7 @@ func newFileStoreWithCreated(fcfg FileStoreConfig, cfg StreamConfig, created tim
 	fs := &fileStore{
 		fcfg: fcfg,
 		cfg:  FileStreamInfo{Created: created, StreamConfig: cfg},
-		fch:  make(chan struct{}),
+		fch:  make(chan struct{}, 1), // Make buffered on purpose.
```
kozlovic, line 232: Could you also please check: ConsumerStore() where we create an fch but that one seem to not even be used?

You may have the same issue with consumer.go, AddConsumer() where we create fch the same way (non buffered) but use select{} to signal, which can lead to loss signal.

kozlovic (changes requested):
I looked for places where we create an unbuffered `chan struct{}` and I found a place in consumer.go that may be having the same issue, and another in filestore.go but that one I am not sure it is even used.

server/filestore.go
```diff
@@ -2634,7 +2638,7 @@ func (fs *fileStore) ConsumerStore(name string, cfg *ConsumerConfig) (ConsumerSt
 		name: name,
 		odir: odir,
 		ifn:  path.Join(odir, consumerState),
-		fch:  make(chan struct{}),
+		fch:  make(chan struct{}, 1), // Make buffered on purpose.
```
kozlovic, line 2641: This is the one that when I search all references in VS Code, I can't find any. It seems to me that `fch` in this struct is never used?

kozlovic (approved):
LGTM.. but there is a `fch` that I believe is not being used at all...

server/filestore.go
derekcollison: I think you are right. Weird staticcheck did not catch it.

server/filestore.go
```diff
@@ -2634,7 +2638,6 @@ func (fs *fileStore) ConsumerStore(name string, cfg *ConsumerConfig) (ConsumerSt
 		name: name,
 		odir: odir,
 		ifn:  path.Join(odir, consumerState),
-		fch:  make(chan struct{}),
```
kozlovic, line 2359: But remove from `consumerFileStore` struct too ;-)

kozlovic (changes requested):

server/filestore.go
kozlovic: Same for qch btw... unless I am missing something these 2 did not seem to be referenced anywhere.

kozlovic (approved):
LGTM

PR #1676, opened 2020-10-28
Track number of stream pending for each consumer.

derekcollison:
This will track the stream pending state for each consumer.
This code does account for filtered consumers.

/cc @nats-io/core

### DIFF
--- a/server/consumer.go
+++ b/server/consumer.go
@@ -33,15 +33,16 @@ import (
 )
 
 type ConsumerInfo struct {
-	Stream         string         `json:"stream_name"`
-	Name           string         `json:"name"`
-	Created        time.Time      `json:"created"`
-	Config         ConsumerConfig `json:"config"`
-	Delivered      SequencePair   `json:"delivered"`
-	AckFloor       SequencePair   `json:"ack_floor"`
-	NumPending     int            `json:"num_pending"`
-	NumRedelivered int            `json:"num_redelivered"`
-	NumWaiting     int            `json:"num_waiting"`
+	Stream           string         `json:"stream_name"`
+	Name             string         `json:"name"`
+	Created          time.Time      `json:"created"`
+	Config           ConsumerConfig `json:"config"`
+	Delivered        SequencePair   `json:"delivered"`
+	AckFloor         SequencePair   `json:"ack_floor"`
+	NumPending       int            `json:"num_pending"`
+	NumRedelivered   int            `json:"num_redelivered"`
+	NumWaiting       int            `json:"num_waiting"`
+	NumStreamPending uint64         `json:"num_stream_pending"`
 }
 
 type ConsumerConfig struct {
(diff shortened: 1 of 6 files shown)


server/consumer.go
```diff
@@ -33,15 +33,16 @@ import (
...
+	Stream           string         `json:"stream_name"`
+	Name             string         `json:"name"`
+	Created          time.Time      `json:"created"`
+	Config           ConsumerConfig `json:"config"`
+	Delivered        SequencePair   `json:"delivered"`
+	AckFloor         SequencePair   `json:"ack_floor"`
+	NumPending       int            `json:"num_pending"`
+	NumRedelivered   int            `json:"num_redelivered"`
+	NumWaiting       int            `json:"num_waiting"`
+	NumStreamPending uint64         `json:"num_stream_pending"`
```
ripienaar, line 45: hmm, not sure I think the name of this is confusing between num_pending and num_stream_pending. Maybe its num_pending thats confusing but together its very confusing.

ripienaar (commented):
This looks good, lots of knock on changes needs doing in other tools and libraries but a good step.  Pity about the initial startup speed hit gues it's unavoidable though.

It should be possible for streams to track the numer of messages matching every subject they know about and consumers can then select matching ones out and add them up for initial start.  But this will blow up on streams with lets say a subject per stock item or something so probably a bad idea :(

Reason I am thinking along those lines is because we hoped this work would get us nearer to per subject limits tracking etc, and having those would enable this feature - but as built here we have a O(n) operation that, once those feature come, would be one of 2 O(n) operations or this code would need a complete redo?

server/consumer.go
derekcollison: Agree but we had that one already, NumPending. That is really NumAckPending and the NumStreamPending is NumPendning for me.

derekcollison:
The O(n) concerns me as well. Not that this is only for Filtered and even if the stream has wildcards if the filtered subject is literal this is a simple compare, still.

If the filtered subject is a wildcard as well its still O(n) but more expensive per single comparison.

As far as better approaches, we could track on the stream per literal subject, but since a consumer gets to pick its start sequence we would need multiple states with sequence number semantics for this to really help.

ripienaar:
yeah didnt think about the start time/sequence/time deltas.  We could though hit O(n) then only if those are set. Difficult one to speed up.

derekcollison:
Difficult for sure, always a trade of space and time. There may be something we could do though but not sure now is the time.

Also no matter what we do since consumers can have filtered subjects with wildcards we still need to check since any given set for a stream can be very different based on a consumer's wildcard subject filter.

derekcollison:
I think we can merge and watch O(n) stuff. We do not have O(n) but something close with selecting a starting sequence for a consumer by time.

ripienaar:
OK, LGTM then, I'll follow up with jsm and nats.go changes.

PR #1677, opened 2020-10-28
rename consumer pending fields for clarity

ripienaar:

/cc @nats-io/core

### DIFF
--- a/server/consumer.go
+++ b/server/consumer.go
@@ -33,16 +33,16 @@ import (
 )
 
 type ConsumerInfo struct {
-	Stream           string         `json:"stream_name"`
-	Name             string         `json:"name"`
-	Created          time.Time      `json:"created"`
-	Config           ConsumerConfig `json:"config"`
-	Delivered        SequencePair   `json:"delivered"`
-	AckFloor         SequencePair   `json:"ack_floor"`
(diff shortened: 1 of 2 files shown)


server/consumer.go
```diff
@@ -33,16 +33,16 @@ import (
...
+	Stream         string         `json:"stream_name"`
+	Name           string         `json:"name"`
+	Created        time.Time      `json:"created"`
+	Config         ConsumerConfig `json:"config"`
+	Delivered      SequencePair   `json:"delivered"`
+	AckFloor       SequencePair   `json:"ack_floor"`
+	NumAckPending  int            `json:"num_ack_pending"`
+	NumRedelivered int            `json:"num_redelivered"`
+	NumWaiting     int            `json:"num_waiting"`
+	NumPending     uint64         `json:"num_pending"`
```
derekcollison, line 45: I like AckPending one, this one though feels to vague, WDYT?

NumOutstanding? NumStreamPending?

ripienaar: Stream seems wrong since its not really stream pending its Consumer pending - multiple consumers on the same stream would have different values, totals etc.  Thus StreamPending seems bad to me. Pending looks obvious as long as we specialise all the others num_x_pending ones.

PR #1682, opened 2020-10-29
Would not properly handle wildcard streams with interest retention and filtered consumers

derekcollison:

/cc @nats-io/core

### DIFF
--- a/server/stream.go
+++ b/server/stream.go
@@ -905,9 +905,9 @@ func (mset *Stream) processInboundJetStreamMsg(_ *subscription, pc *client, subj
 			return
 		}
 	}
-	mset.mu.Unlock()
 
 	if c == nil {
+		mset.mu.Unlock()
 		return
 	}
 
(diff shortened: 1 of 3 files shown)


server/stream.go
```diff
@@ -931,18 +932,35 @@ func (mset *Stream) processInboundJetStreamMsg(_ *subscription, pc *client, subj
...
-		if doAck && len(reply) > 0 {
-			response = append(pubAck, strconv.FormatUint(seq, 10)...)
-			response = append(response, '}')
-			mset.sendq <- &jsPubMsg{reply, _EMPTY_, _EMPTY_, nil, response, nil, 0}
+	if interestRetention {
+		var noInterest bool
+
+		if numConsumers == 0 {
+			noInterest = true
+		} else if mset.numFilter > 0 {
```
kozlovic, line 940: We access mset.numFilter and mset.consumers without the mset lock here. Would recommend determining if there is noInterest for interestRetention early when still under the mset lock (when we determine the consumer count). It will also avoid differences between the time we evaluate the count and when we are here (since lock has been released number of consumers may have changed - up or down).

kozlovic (changes requested):
check of mset fields without lock...

kozlovic (approved):
LGTM

kozlovic (approved):
LGTM

PR #1683, opened 2020-10-30
ensure bare AckNxt requesting >1 is supported

ripienaar:

/cc @nats-io/core

### DIFF
--- a/server/consumer.go
+++ b/server/consumer.go
@@ -1076,8 +1076,8 @@ func nextReqFromMsg(msg []byte) (time.Time, int, bool, error) {
 	bs := 1
 	// Naked batch size here for backward compatibility.
 	switch {
-	case bytes.HasPrefix(msg, AckNext):
-		if n, _ := fmt.Sscanf(string(msg), ackNextCnt, &bs); n == 0 {
+	case strings.HasPrefix(req, string(AckNext)):
+		if n, _ := fmt.Sscanf(req, ackNextCnt, &bs); n == 0 {
 			bs = 1
 		}
 	default:
--- a/server/consumer.go
+++ b/server/consumer.go
@@ -158,6 +158,8 @@ var (
 	AckNext = []byte("+NXT")
 	// Terminate delivery of the message.
 	AckTerm = []byte("+TERM")
+
+	ackNextCnt = "+NXT %d+"
 )
 
 // Consumer is a jetstream consumer.
@@ -718,7 +720,7 @@ func (o *Consumer) processAck(_ *subscription, _ *client, subject, reply string,
 	switch {
 	case len(msg) == 0, bytes.Equal(msg, AckAck), bytes.Equal(msg, AckOK):
 		o.ackMsg(sseq, dseq, dcount)
-	case bytes.Equal(msg, AckNext):
+	case bytes.HasPrefix(msg, AckNext):
 		o.ackMsg(sseq, dseq, dcount)
 		o.processNextMsgReq(nil, nil, subject, reply, msg)
 		skipAckReply = true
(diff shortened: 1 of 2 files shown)


ripienaar:
looks to me the intention was that this is now supported but it didnt work

server/consumer.go
```diff
@@ -1070,11 +1072,20 @@ func nextReqFromMsg(msg []byte) (time.Time, int, bool, error) {
...
 		return cr.Expires, cr.Batch, cr.NoWait, nil
 	}
-	// Naked batch size here for backward compatibility.
+
 	bs := 1
-	if n, err := strconv.Atoi(req); err == nil {
-		bs = n
+	// Naked batch size here for backward compatibility.
+	switch {
+	case bytes.HasPrefix(msg, AckNext):
```
derekcollison, line 1079: I think use `req` not `msg` here.

server/consumer.go
```diff
@@ -1070,11 +1072,20 @@ func nextReqFromMsg(msg []byte) (time.Time, int, bool, error) {
...
+		if n, _ := fmt.Sscanf(string(msg), ackNextCnt, &bs); n == 0 {
```
derekcollison, line 1080: Same here. Not for prefix of course but possible trailing spaces.

test/jetstream_test.go
```diff
@@ -82,6 +83,7 @@ func RunBasicJetStreamServer() *server.Server {
 	opts := DefaultTestOptions
 	opts.Port = -1
 	opts.JetStream = true
+	opts.NoLog = true
```
derekcollison, line 86: This should be part of DefaultTestOptions no?

test/jetstream_test.go
```diff
@@ -3013,10 +3015,92 @@ func TestJetStreamConsumerAckAck(t *testing.T) {
 	testAck(server.AckAck)
 	testAck(server.AckNak)
 	testAck(server.AckProgress)
-	testAck(server.AckNext)
 	testAck(server.AckTerm)
 }
 
+func TestJetStreamNack(t *testing.T) {
```
derekcollison, line 3021: Test name wrong?

derekcollison (changes requested):

test/jetstream_test.go
ripienaar: yeah not sure how that crept in there, removing

test/jetstream_test.go
```diff
@@ -83,7 +83,7 @@ func RunBasicJetStreamServer() *server.Server {
 	opts := DefaultTestOptions
 	opts.Port = -1
 	opts.JetStream = true
-	opts.NoLog = true
+	opts.NoLog = false
```
derekcollison, line 86: Do you want the test to print log statements? This should be set to true.

derekcollison (changes requested):

test/jetstream_test.go
kozlovic: Actually, for websocket/mqtt tests, I do set this intentionally to `NoLog = false` and set a dummy logger with debug and trace enabled:
```
    opts.NoLog = false
    l := &DummyLogger{}
    s.SetLogger(l, true, true)
```
The reason? This exercises all code that has debug/trace, making sure that we don't have situations where we may be tracing something that causes a dereference of a nil pointer. As an added bonus, it helps with code coverage ;-)

test/jetstream_test.go
kozlovic: Also see my comment about testing debug/trace statements with the use of a dummy logger. Note, we may not have a DummyLogger in the test/ package, but that could be easily copied from server tests.

ripienaar: restored as it was before, i think this is some line of code i put in on some other day while debugging and just didnt notice :) gone now.

test/jetstream_test.go
derekcollison: Might be a good idea.

derekcollison (approved):
LGTM

ripienaar:
merging after travis passed

Issue #1442, opened 2020-06-02
Continuous fuzzing


AdamKorcz:
#### Use Case:
[This bug](https://github.com/nats-io/nats-server/pull/1441) was found by use of a fuzz test, and fuzzing the parser is an effective way to search for other crashes in the nats-server. 

#### Proposed Change:
I have written a fuzzer for the parser. I propose to include it in the nats-server repository and setup continuous fuzzing through oss-fuzz. This will allow Google to run the fuzzer continuously to look for bugs. oss-fuzz is a free service that is offered with an implied expectation that bugs are found, so that the resources spent on fuzzing nats-server are put to good use. If any bugs are found, maintainers are emailed a link to a detailed bug report.

#### Who Benefits From The Change(s)?
All nats users

AdamKorcz:
Is it possible to ask kindly for feedback on this?

derekcollison:
Please submit PR, we would be happy to take a look!

matthiashanel:
@AdamKorcz we would welcome such a PR. The only thing would be new dependencies, but @wallyqs has recently found a way to separate out test dependencies.

PR #1684, opened 2020-10-31
Added 2 fuzzers

AdamKorcz:
This PR addresses https://github.com/nats-io/nats-server/issues/1442 by adding two fuzzers.

The parser in `/server` has been fuzzed before and as a result of this, bugs and vulnerabilities have been found. 

With this PR I suggest running these fuzzers continuously by way of oss-fuzz. This will allow Google to run the fuzzers and notify maintainers in case any bugs or vulnerabilities are found. The service is offered free of charge with an implied expectation that bugs are fixed, so that the resources spent on fuzzing NATS are put to good use. I have previously reported bugs in NATS and they were fixed incredibly fast, so I presume this is not a problem.

I can confirm that these fuzzers work on oss-fuzz's infrastructure, and I have the setup ready to integrate NATS into oss-fuzz. All I need to proceed are the email addresses of the maintainers that are to receive potential bug reports. These will be added to a public list and can be changed at any time.

Furthermore, I will be happy to write a fuzzer for the NATS client as well and include that in the integration for continuous fuzzing.

### DIFF
--- a/conf/fuzz.go
+++ b/conf/fuzz.go
@@ -1,4 +1,4 @@
-// Copyright 2012-2020 The NATS Authors
+// Copyright 2020 The NATS Authors
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
 // You may obtain a copy of the License at
--- a/conf/fuzz.go
+++ b/conf/fuzz.go
@@ -1,0 +1,24 @@
+// Copyright 2012-2020 The NATS Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// +build gofuzz
+
+package conf
+
+func Fuzz(data []byte) int {
+	_, err := Parse(string(data))
+	if err != nil {
+		return 0
+	}
+	return 1
+}
--- a/server/fuzz.go
+++ b/server/fuzz.go
@@ -1,4 +1,4 @@
-// Copyright 2012-2020 The NATS Authors
+// Copyright 2020 The NATS Authors
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
 // You may obtain a copy of the License at
--- a/server/fuzz.go
+++ b/server/fuzz.go
@@ -1,0 +1,34 @@
+// Copyright 2012-2020 The NATS Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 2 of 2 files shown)


server/fuzz.go
```diff
@@ -0,0 +1,34 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// +build gofuzz
+
+package server
+
+func Fuzz(data []byte) int {
```
matthiashanel, line 18: Hey @AdamKorcz, looking at this general structure, Currently you are picking up dummyClient() from parser_test.go. What if I wanted to add dummyRouteClient() as well or wanted to add fuzzing for an altogether different scenario say fuzzing that also includes receiving a message?

Similar to Test, do I just add more functions named `Fuzz...(daba []byte) int` or by switching cases on say the first byte handed to `Fuzz`?
I'm asking as I found one article that suggested as much but couldn't quite find that supported elsewhere.

matthiashanel (commented):
Local runs are then built and executed with `go-fuzz-build` and `go-fuzz`?

server/fuzz.go
AdamKorcz: Are you looking to fuzz the parser with the `dummyRouteClient` as well?

Regarding fuzzing and receiving a message: Could you point me to the part of the code that you are thinking of? 

To write more fuzzers, you write new functions named `FuzzSOMETHING (data []byte) int`. 

Might I suggest that in the early stages of fuzzing that I write the fuzzers and merge them in? I will be happy to do it to get things started, so if you have any parts of the code you would like to see fuzzed, please throw it at me and I will look into it.

Regarding running the fuzzers locally: I will add the instructions to do that in comments in the fuzzer, is that okay? - And yes, if you have added `$GOPATH/bin` to your `$PATH` variable it would be `go-fuzz-build && go-fuzz`

matthiashanel: > Are you looking to fuzz the parser with the dummyRouteClient as well?

Sort of, when looking at the implementation of dummyClient and dummyRouteClient I would add tests where kind has the values ROUTER, GATEWAY, LEAF & CLIENT (client has the value 0), essentially testing if the connection is a client connection, cluster connection, gateway or leaf node connection.
`&client{srv: New(&defaultServerOptions), msubs: -1, mpay: -1, mcl: MAX_CONTROL_LINE_SIZE, kind: CLIENT}`

```
func dummyClient() *client {
	return &client{srv: New(&defaultServerOptions), msubs: -1, mpay: -1, mcl: MAX_CONTROL_LINE_SIZE}
}

func dummyRouteClient() *client {
	return &client{srv: New(&defaultServerOptions), kind: ROUTER}
}
```

> Regarding fuzzing and receiving a message: Could you point me to the part of the code that you are thinking of?
Well, nothing concrete yet, but am wondering how much mileage we'd get out of doing more than just parsing. 
Like parse, but also have a subscriber that is expected to receive one message. 
I want to give a hypothetical example. If the server has no subscriber at all, why look at at a messages subject during fuzzing ever?

matthiashanel: That said, testing the client connection is the more important one to test.
Code wise I'd happy with what is submitted. 
Maybe rename Fuzz to FuzzClient.

conf/fuzz.go
```diff
@@ -0,0 +1,24 @@
+// Copyright 2012-2020 The NATS Authors
```
kozlovic, line 1: Sorry we missed that earlier. Since this is a new file, we usually put only the year the file is created, so that would be 2020 here.

server/fuzz.go
```diff
@@ -0,0 +1,34 @@
+// Copyright 2012-2020 The NATS Authors
```
kozlovic, line 1: Same here.

kozlovic (changes requested):
Sorry, just small changes to copyright banner and we will be ready to merge.

kozlovic (approved):
LGTM. Thanks, and I believe that you needed some email to report issues: you can use security@nats.io

PR #1685, opened 2020-10-31
Update to write architecture and bug fixes and stability improvements.

derekcollison:
We had a left over artifact from the shared write buffer architecture where the filestore was still controlling the flush loop for the special last message block. This changes it such that each message block is running their own flush loop.

/cc @nats-io/core

### DIFF
--- a/server/filestore.go
+++ b/server/filestore.go
@@ -971,6 +971,7 @@ func (mb *msgBlock) spinUpFlushLoop() {
 		mb.mu.Unlock()
 		return
 	}
+	mb.flusher = true
 	mb.fch = make(chan struct{}, 1)
 	mb.qch = make(chan struct{})
 	fch, qch := mb.fch, mb.qch
--- a/server/filestore.go
+++ b/server/filestore.go
@@ -74,7 +74,6 @@ type fileStore struct {
 	lmb      *msgBlock
 	blks     []*msgBlock
 	hh       hash.Hash64
-	fch      chan struct{}
 	qch      chan struct{}
 	cfs      []*consumerFileStore
 	closed   bool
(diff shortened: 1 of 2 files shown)


kozlovic (changes requested):
A bit concerned on how the flusher loop is started.

server/filestore.go
```diff
@@ -895,26 +905,20 @@ func (fs *fileStore) removeMsg(seq uint64, secure bool) (bool, error) {
...
+		qch = mb.qch
+		fch = mb.fch
 	}
 	mb.mu.Unlock()
 
 	// Kick outside of lock.
-	if shouldWriteIndex && dch != nil {
+	if shouldWriteIndex {
+		if qch == nil {
+			mb.spinUpFlushLoop()
```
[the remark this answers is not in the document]
derekcollison: True will set flusher bool in spinup after test.

server/filestore.go
```diff
@@ -959,14 +963,106 @@ func (fs *fileStore) isClosed() bool {
...
 
-// Loop on requests to write out our index file. This is used when calling
-// remove for a message. Updates to the last.seq etc are handled by main
-// flush loop when storing messages.
-func (fs *fileStore) flushWriteIndexLoop(mb *msgBlock, dch, qch chan struct{}) {
+// Will spin up our flush loop.
+func (mb *msgBlock) spinUpFlushLoop() {
+	mb.mu.Lock()
+	// Are we already running?
+	if mb.flusher {
```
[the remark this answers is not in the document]
derekcollison: Yep will change as described above.

derekcollison:
Posted changes. Please take a look when you have a chance.

server/filestore.go
```diff
@@ -971,6 +971,7 @@ func (mb *msgBlock) spinUpFlushLoop() {
 		mb.mu.Unlock()
 		return
 	}
+	mb.flusher = true
```
kozlovic, line 974: May not need the `setInFlusher()` function and its call in flushLoop() anymore?

derekcollison: I looked at that but liked the symmetry there so left it. Can remove it and put comment there if you feel strongly about it.

kozlovic: that's fine.

kozlovic (approved):

kozlovic:
LGTM

PR #1686, opened 2020-10-31
Separate out disk available for Windows

derekcollison:

/cc @nats-io/core


kozlovic (approved):
LGTM

PR #1689, opened 2020-11-02
support next structure in acknxt

ripienaar:

/cc @nats-io/core

### DIFF
--- a/server/consumer.go
+++ b/server/consumer.go
@@ -158,8 +158,6 @@ var (
 	AckNext = []byte("+NXT")
 	// Terminate delivery of the message.
 	AckTerm = []byte("+TERM")
-
-	ackNextCnt = "+NXT %d+"
 )
 
 // Consumer is a jetstream consumer.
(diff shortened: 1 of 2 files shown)


server/consumer.go
```diff
@@ -1061,32 +1059,30 @@ func (o *Consumer) needAck(sseq uint64) bool {
 
 // Helper for the next message requests.
 func nextReqFromMsg(msg []byte) (time.Time, int, bool, error) {
-	req := strings.TrimSpace(string(msg))
-	if len(req) == 0 {
-		return time.Time{}, 1, false, nil
+	req := bytes.TrimSpace(msg)
```
derekcollison, line 1062: Would make `var req []byte` and set only once based on if.

derekcollison: Might pull off prefix at upper layers and not here.

derekcollison (changes requested):

ripienaar:
updated

derekcollison (approved):
LGTM

PR #1692, opened 2020-11-03
[CHANGED] Gateway connections now always send PINGs

kozlovic:
Connections normally suppress sending PINGs if there was some
activity. We now force GATEWAY connections to send PINGs at the
configured interval or 15 seconds, whichever is the smallest.


derekcollison (approved):
LGTM

PR #1693, opened 2020-11-05
Partial filestore cache bug

derekcollison:
Fix for partial cache overwrite bug.
    
When we moved to a write through cache architecture we also moved the cache write to offset based instead of APPEND.

We were inadvertently clearing our offset from our cache when we would clear which meant if the next operation was another write we would have the wrong offset and overwrite previous messages.
    
/cc @nats-io/core


kozlovic (approved):
LGTM

PR #1694, opened 2020-11-05
Redeliver pending 

derekcollison:
We want to redeliver all pending immediately and in stream order if a consumer changes its delivery subject.

/cc @nats-io/core


kozlovic (approved):
LGTM

PR #1697, opened 2020-11-10
Updates to consumer store to support delta updates.

derekcollison:
In preparation for clustering we need to have the consumer filestore update state with deltas vs original design.

Not sure I will stick with this long term since we could do delta updates for clustering at the consumer level above the store implementation but this includes some fixes and stability improvements and moves us closer to clustering.

/cc @nats-io/core

### DIFF
--- a/server/consumer.go
+++ b/server/consumer.go
@@ -171,14 +171,14 @@ type Consumer struct {
 	dseq              uint64
 	adflr             uint64
 	asflr             uint64
-	spending          uint64
+	sgap              uint64
 	dsubj             string
 	rlimit            *rate.Limiter
 	reqSub            *subscription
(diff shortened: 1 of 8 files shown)


kozlovic:
@derekcollison I think you have have worked on a branch that got a bit stale. There is a conflict on filestore.go, but also, if I try to merge that with a branch (to test effects on mqtt branch), I get a compilation error with consumer.go where you recently added forceExpirePending() which does `off := now - o.pending[expired[0]]`. Should you try to rebase from master first?

derekcollison:
I can rebase, looking at Phil's report from last night as well.

kozlovic (approved):
LGTM

PR #1699, opened 2020-11-10
Updates to index cache handling for message blocks.

derekcollison:
We can have partial caches and we can also remove the idx cache. This was causing a bug where we would get the wrong slotInfo from the cache.idx. This code fixes the bug and detects idx partials.

/cc @nats-io/core

### DIFF
--- a/server/filestore_test.go
+++ b/server/filestore_test.go
@@ -1358,6 +1358,54 @@ func TestFileStorePartialCacheExpiration(t *testing.T) {
 	}
 }
 
+func TestFileStorePartialIndexes(t *testing.T) {
+	storeDir, _ := ioutil.TempDir("", JetStreamStoreDir)
+	os.MkdirAll(storeDir, 0755)
+	defer os.RemoveAll(storeDir)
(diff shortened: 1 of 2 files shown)


server/filestore_test.go
```diff
@@ -1358,6 +1358,54 @@ func TestFileStorePartialCacheExpiration(t *testing.T) {
...
+		}
+		return fmt.Errorf("Index not empty")
+	})
+
+	// Create a partial cache by adding more msgs.
+	for i := 0; i < toSend; i++ {
+		fs.StoreMsg("foo", nil, []byte("ok-2"))
+	}
+	// If we now load in a message in second half if we do not
+	// detect idx is a prtial correctly this will panic.
```
kozlovic, line 1403: typo: `prtial` -> `partial`

kozlovic (approved):
LGTM

server/filestore_test.go
derekcollison: Will fix and squash.

PR #1702, opened 2020-11-11
Remove conditional and broadcast for signalling consumers

derekcollison:

/cc @nats-io/core

### DIFF
--- a/server/consumer.go
+++ b/server/consumer.go
@@ -191,6 +191,7 @@ type Consumer struct {
 	filterWC          bool
 	dtmr              *time.Timer
 	dthresh           time.Duration
+	mch               chan struct{}
 	qch               chan struct{}
 	inch              chan bool
 	sfreq             int32
@@ -384,6 +385,7 @@ func (mset *Stream) AddConsumer(config *ConsumerConfig) (*Consumer, error) {
 		dsubj:   config.DeliverSubject,
 		active:  true,
 		qch:     make(chan struct{}),
+		mch:     make(chan struct{}, 1),
 		sfreq:   int32(sampleFreq),
 		maxdc:   uint64(config.MaxDeliver),
 		created: time.Now().UTC(),
@@ -510,7 +512,7 @@ func (mset *Stream) AddConsumer(config *ConsumerConfig) (*Consumer, error) {
 	// If push mode, register for notifications on interest.
 	if o.isPushMode() {
 		o.dthresh = JsDeleteWaitTimeDefault
-		o.inch = make(chan bool, 4)
+		o.inch = make(chan bool, 8)
 		a.sl.RegisterNotification(config.DeliverSubject, o.inch)
 		o.active = o.hasDeliveryInterest(<-o.inch)
 		// Check if we are not durable that the delivery subject has interest.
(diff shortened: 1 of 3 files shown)


server/consumer.go
```diff
@@ -1238,7 +1247,8 @@ func (o *Consumer) processNextMsgReq(_ *subscription, c *client, _, reply string
 	if o.replay {
 		o.waiting.add(&wr)
 		o.mu.Unlock()
-		mset.signalConsumers()
+		o.signalNewMessages()
+		//mset.signalConsumers()
```
kozlovic, line 1251: Remove?

server/consumer.go
```diff
@@ -1238,7 +1247,8 @@ func (o *Consumer) processNextMsgReq(_ *subscription, c *client, _, reply string
 	if o.replay {
 		o.waiting.add(&wr)
 		o.mu.Unlock()
-		mset.signalConsumers()
+		o.signalNewMessages()
```
kozlovic, line 1250: In some places this is invoked inside of consumer lock, but here it is outside. Note that in signalNewMessages() you are accessing consumer's field that may or may not be immutable, not sure.

server/stream.go
```diff
@@ -1013,25 +1009,12 @@ func (mset *Stream) processInboundJetStreamMsg(_ *subscription, pc *client, subj
 		for _, o := range obs {
 			o.incStreamPending(seq, subject)
 			if !o.deliverCurrentMsg(subject, hdr, msg, seq, ts) {
-				needSignal = true
+				o.signalNewMessages()
```
kozlovic, line 1012: Should we have a locked/unlocked version?

kozlovic (commented):
Not sure about locking...

server/consumer.go
derekcollison: geez, I even looked for all of them ;)

derekcollison: Will fix.

server/consumer.go
derekcollison: Yes understand. I rely on Go's race detector for things setup once at creation time and never changed such that locks are not required if the underlying pointer is safe like channels.

server/stream.go
derekcollison: I don't think so since these are channels. Just can't nil them out ever which is fine.

derekcollison: Read from multiple Go routines ok.

server/consumer.go
kozlovic: So unless consumer object is created in some tests without proper call to AddConsumer - and store is possibly not set, maybe we should remove the check to `o.store == nil` in signalNewMessages() to simply avoid confusion or second guessing.

kozlovic (approved):
LGTM

derekcollison:
That is one way we tell we are closed, but could remove, won't matter. Will do that, squash and merge.

kozlovic:
> That is one way we tell we are closed, but could remove, won't matter. Will do that, squash and merge.

Not anymore. Remember I added the `closed` boolean to avoid a race that was causing issues. But anyway it would prove my point that if we had a test that closes the consumer while there was a new message that cause signal, then we would have gotten the race detector to show the issue (if we replaced with `if o.closed {` since o.store is no longer set to nil).

kozlovic (approved):
LGTM

PR #1703, opened 2020-11-12
Fix flapper, be pedantic about signaling on delivery subject update

derekcollison:

/cc @nats-io/core


kozlovic (approved):
LGTM

PR #1705, opened 2020-11-12
Make interest notifications explicit match only

derekcollison:

/cc @nats-io/core

### DIFF
--- a/server/sublist.go
+++ b/server/sublist.go
@@ -138,9 +138,10 @@ func (s *Sublist) CacheEnabled() bool {
 // RegisterNotification will register for notifications when interest for the given
 // subject changes. The subject must be a literal publish type subject. The
 // notification is true for when the first interest for a subject is inserted,
-// and false when all interest in the subject is removed. The sublist will not
-// block when trying to send the notification. Its up to the caller to make sure
-// the channel send will not block.
+// and false when all interest in the subject is removed. Note that this interest
+// needs to be exact and that wildcards will not trigger the notifications. The sublist
+// will not block when trying to send the notification. Its up to the caller to make
+// sure the channel send will not block.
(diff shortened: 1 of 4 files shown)


server/sublist.go
```diff
@@ -149,8 +150,24 @@ func (s *Sublist) RegisterNotification(subject string, notify chan<- bool) error
 		return ErrNilChan
 	}
 
+	var hasInterest bool
 	r := s.Match(subject)
-	hasInterest := len(r.psubs)+len(r.qsubs) > 0
+
+	if len(r.psubs)+len(r.qsubs) > 0 {
+		for _, sub := range r.psubs {
+			if string(sub.subject) == subject {
```
ripienaar, line 158: not sure all the cases that would hit this code so just a question - this will prevent `>` from matching, good, but what about checking for `x.>` for a message on `x.y`, should we support those wildcards?

derekcollison: Slippery slope, I think we need to only match explicit.

derekcollison:
I see test failure on TestJetStreamAckExplicitMsgRemoval, looking now.

derekcollison:
ok fixed and force pushed.

kozlovic (approved):
LGTM. Given that delivery subjects cannot have wildcard, I believe that this is a good approach. Suppose a consumer that has delivery subject "bar" and filters subjects "foo.bar" from a stream that captures "foo.>". It is running and receives messages 1 and 2 published on foo.bar. The app is stopped, messages 2 and 3 are received. Starting a subscriber on ">" would cause messages to be delivered to this sub that is not meant to be the associated with JS durable. With this change, nothing would happen. Only after restarting the sub on "bar" would messages be delivered to the "real" application AND the subscriber on ">".

ripienaar (approved):

PR #1707, opened 2020-11-12
Fixed bug and a race condition

derekcollison:
When under load a call to `LoadMsg()` could return the wrong error since we did not snapshot the last sequence properly and would read it after.

Fixed race condition on clearing cache.

/cc @nats-io/core

### DIFF
--- a/server/filestore.go
+++ b/server/filestore.go
@@ -1513,13 +1513,6 @@ func (mb *msgBlock) clearFlushing() {
 	}
 }
 
-// Lock should not be held.
-func (mb *msgBlock) clearFlushingUnlocked() {
-	mb.mu.Lock()
-	defer mb.mu.Unlock()
-	mb.clearFlushing()
-}
-
 // Lock should be held.
 func (mb *msgBlock) setFlushing() {
 	if mb.cache != nil {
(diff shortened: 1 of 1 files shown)


server/filestore.go
```diff
@@ -1770,7 +1777,7 @@ func (mb *msgBlock) flushPendingMsgs() error {
 
 	// Only one can be flushing at a time.
 	mb.setFlushing()
-	defer mb.clearFlushing()
+	defer mb.clearFlushingUnlocked()
```
kozlovic, line 1780: Would it not be worth moving the original "defer mb.clearFlushing()" after line 1811 and call clearFlushingUnlocked() after write error before returning (around line 1795).

Arguably, defer could even be omitted because after the write and when reacquiring the lock there is currently only 2 places were we return, so could call in place.

The current approach seem just a bit weird because we release the lock and at the execution of this defer will reacquire the lock just to clear the "flag".

kozlovic (approved):
LGTM, but maybe different approach for clear flushing?

server/filestore.go
derekcollison: Yes feels weird(ish) but wanted the defer to make sure it gets done and we do not return and forget..

derekcollison: However let me take another look.

derekcollison: ok take a look now.

kozlovic (approved):
LGTM

PR #1709, opened 2020-11-14
Fix bug when removing the same message from a stream.

derekcollison:
We would release locks and call into upper layers when removing a message.
The upper layers may call back into the lower layers to get more information, such as the subject.
This fix has the storage updates optionally supply the subject for filtered consumers and fixes the bug of double deletes.

/cc @nats-io/core


kozlovic (approved):
LGTM

PR #1710, opened 2020-11-14
Add support for MaxAckPending

derekcollison:

/cc @nats-io/core


ripienaar (approved):

ripienaar:
I think we need to report pending in consumer info too - a consumer might stall and a administrator will want to find out why there is an apparent stall, reporting outstanding acks will help

PR #1711, opened 2020-11-14
Allow stream lookup by subject.

derekcollison:
Allow an API endpoint and public API to lookup a stream by subject. The subject needs to be an exact match or a subset. If the subject is considered a filtered subject for the stream that will also be returned.

/cc @nats-io/core

### DIFF
--- a/server/jetstream_api.go
+++ b/server/jetstream_api.go
@@ -75,6 +75,10 @@ const (
 	JSApiStreamInfo  = "$JS.API.STREAM.INFO.*"
 	JSApiStreamInfoT = "$JS.API.STREAM.INFO.%s"
 
+	// JSApiStreamLookup is for obtaining a stream by a target subject.
+	// Will return JSON response.
+	JSApiStreamLookup = "$JS.API.STREAM.LOOKUP"
+
 	// JSApiStreamDelete is the endpoint to delete streams.
 	// Will return JSON response.
 	JSApiStreamDelete  = "$JS.API.STREAM.DELETE.*"
@@ -112,7 +116,7 @@ const (
 	JSApiConsumerCreate  = "$JS.API.CONSUMER.CREATE.*"
 	JSApiConsumerCreateT = "$JS.API.CONSUMER.CREATE.%s"
 
-	// JSApiDurableCreate is the endpoint to create ephemeral consumers for streams.
+	// JSApiDurableCreate is the endpoint to create durable consumers for streams.
 	// You need to include the stream and consumer name in the subject.
 	JSApiDurableCreate  = "$JS.API.CONSUMER.DURABLE.CREATE.*.*"
 	JSApiDurableCreateT = "$JS.API.CONSUMER.DURABLE.CREATE.%s.%s"
@@ -261,6 +265,15 @@ type JSApiStreamInfoResponse struct {
 }
 
 const JSApiStreamInfoResponseType = "io.nats.jetstream.api.v1.stream_info_response"
+
+// JSApiStreamLookupResponse.
+type JSApiStreamLookupResponse struct {
+	ApiResponse
+	Stream   string `json:"stream"`
+	Filtered bool   `json:"is_filtered"`
+}
+
+const JSApiStreamLookupResponseType = "io.nats.jetstream.api.v1.stream_lookup_response"
 
 // Maximum entries we will return for streams or consumers lists.
 // TODO(dlc) - with header or request support could request chunked response.
@@ -480,6 +493,7 @@ var allJsExports = []string{
 	JSApiStreams,
 	JSApiStreamList,
 	JSApiStreamInfo,
+	JSApiStreamLookup,
 	JSApiStreamDelete,
 	JSApiStreamPurge,
 	JSApiStreamSnapshot,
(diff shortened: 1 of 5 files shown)


server/jetstream_api.go
```diff
@@ -953,6 +968,44 @@ func (s *Server) jsStreamInfoRequest(sub *subscription, c *client, subject, repl
...
+		resp.Error = jsNotEnabledErr
+		s.sendAPIResponse(c, subject, reply, string(msg), s.jsonResponse(&resp))
+		return
+	}
+	if isEmptyRequest(msg) {
+		resp.Error = &ApiError{Code: 400, Description: "subject required"}
+		s.sendAPIResponse(c, subject, reply, string(msg), s.jsonResponse(&resp))
+		return
+	}
+	subj := string(msg)
```
ripienaar, line 988: I imagine in time we'll add additional methods for looking up streams - ones hosted on certain servers for example - so maybe we need to start with this being a JSON request with a property set indicating we're looking up by subject.

Otherwise if we keep this then we should make the API call subject more specific

server/jetstream_api.go
```diff
@@ -262,6 +266,15 @@ type JSApiStreamInfoResponse struct {
 
 const JSApiStreamInfoResponseType = "io.nats.jetstream.api.v1.stream_info_response"
 
+// JSApiStreamLookupResponse.
+type JSApiStreamLookupResponse struct {
+	ApiResponse
+	Stream   string `json:"stream"`
+	Filtered bool   `json:"is_filtered"`
```
ripienaar, line 273: If we go on the assumption that we'll later lookup by other properties, then this key is probably too generic

server/jetstream_api.go
derekcollison: As we need to we can formalize but no need at this point for one string IMO.

server/jetstream_api.go
derekcollison: What key? This is specific to allow developers to potentially only deal with subjects and not streams and consumers (although that should be possible).

We need to know if the subject was filtered (a subset of the streams subjects). Maybe call it Subset instead?

And reason I only return name and force another call for possibly creating consumer was to not sidestep subject based authorization for stream info etc.

server/jetstream_api.go
ripienaar: For a JSON api to have random unstructured input really isn’t the done thing - it’s a major pain for @aricart as well this kind of thing. 

Worse is APIs who have a behaviour for unstructured and then also having to support some structured later and you end with having to maintain both going forward. 

Yes, seems unnecessary when looked at in isolation but at the scope of a larger API this design is a mistake

derekcollison: Why would this be a pain to simple pass a simple string vs constructing JSON to hold just a string? We can add in formal requests if needed later on like we did with others.

We have done that before and no big deal.

Making things more complex than they need to be is a mistake in my opinion.

server/jetstream_api.go
derekcollison: This is specific lookup by subject only. I had that in the signature but we already have way too many function and variable names that remind me of a java code base.

server/jetstream_api.go
ripienaar: Just passing along feedback I had from both internal and external users on the existing API. Accepting bare strings in a JSON API makes those calls weird and different and hard to document using standard means. 

But as you wish.

derekcollison: I understand, I just don't prefer complexity just for complexity sake. If we add more we can formalize request or if folks feel that strongly can add that in now as well but I think will keep plain version like we have in the past.

ripienaar: This is a API to support users, we have to look at the comlexity from the perspective of a user not the 10 lines of code I am asking for.

A easy to use API from a user perspective is one that has a consistent predictable design, with machine readable API descriptions and input validations that can be automated.  Documentation thats generated ideally from the same schemas and a known way to find a schema for any API endpoint. The benefits of this is HUGE, look at kubectl, it's entirely machine generated, look at countless terraform providers like those for major cloud properties - all machine generated.  Only achievable by having a consistent API with machine readable descriptions. Look at our own work in the schemas, input validtions and more.  All only possible by following well defined API design standards.

A complex and difficult to use API is one where there are multiple competing API designs standards in the same API, where some API endpoints defy all efforts to be documented using machine readable schemas because they dont follow compatible standards.  APIs that ignore the design docs (the ADR) and endpoints where there is no way to discover the inputs, validate them and where documentation are not standard or discoverable.  

We are really close to having a API that will deliver a good user experience, but unfortunately those efforts are undermined by this kind of API design.  Please support the others working to create a pleasant API. I know you think its "no big deal", but its literally making achieving our goals impossible.

derekcollison: We can add it in for openapi and doc support. Not sure it makes sense for client implementations to be forced to generate JSON for simple one argument calls though.

This is in support of a possible subject only approach to client API for JetStream as we have been discussing/debating. So normal users will not see this per se at all. We would use it ourselves to grab a stream name from a subject in order to create a consumer without subverting the subject based authorizations.

ripienaar (approved):
LGTM

PR #1712, opened 2020-11-15
Decouple consumer from using stream's sendq.

derekcollison:
We will be replacing this soon (channels) but wanted a separate PR for decoupling consumer from stream first.

/cc @nats-io/core

### DIFF
--- a/server/consumer.go
+++ b/server/consumer.go
@@ -180,6 +180,7 @@ type Consumer struct {
 	ackReplyT         string
 	nextMsgSubj       string
 	maxp              int
+	sendq             chan *jsPubMsg
 	pending           map[uint64]*Pending
 	ptmr              *time.Timer
 	rdq               []uint64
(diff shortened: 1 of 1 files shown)


server/consumer.go
```diff
@@ -548,12 +551,7 @@ func (mset *Stream) AddConsumer(config *ConsumerConfig) (*Consumer, error) {
 // Do all advisory sends here.
 // Lock should be held on entry but will be released.
 func (o *Consumer) sendAdvisory(subj string, msg []byte) {
-	if o.mset != nil && o.mset.sendq != nil {
-		sendq := o.mset.sendq
-		o.mu.Unlock()
-		sendq <- &jsPubMsg{subj, subj, _EMPTY_, nil, msg, nil, 0}
-		o.mu.Lock()
-	}
+	o.sendq <- &jsPubMsg{subj, subj, _EMPTY_, nil, msg, nil, 0}
```
kozlovic, line 554: As you know, I don't like releasing lock then reacquire in situations where the caller or function assumes that the state is not going to change. So happy to see above disappear, but at the same time very nervous about use of the go channel send under lock because the go routine reading may want to grab the lock at one point maybe? (can processInboundClientMsg() at one point end-up wanting to acquire the lock of this consumer?)

kozlovic (approved):
Just general concerns (not just here, but for all our use of go channel send queues) that can fill up and lead to deadlock.

server/consumer.go
derekcollison: Yes understand. In this case it can not since this is not a consumer delivered message. On consumer delivered messages because we process `did not deliver` status we can re-acquire the lock for the consumer in that case so deliverMsg will release the lock. 

For safe keeping can do so here as well even though not needed. Will make that change.

kozlovic (approved):
LGTM. Wonder how "bad" would it be to have replaced with a simple linked list where we could add to it under lock and signal with go channel and have the go routine that dequeues grab the lock to "steal" whatever is in the list at that time and then process outside of lock. Because that would solve the issue of having to unlock to send, list would be unbounded and we would still use go channel for notifications (with select/default on buffered size 1 channel) which can help with throughput compared to condition variable.

derekcollison:
I have a decent MPSC approach I wanted to take and we could use channels for signaling. LinkedLists in general don't perform well on modern machines.

I want to take a look at the Go client as well since there are still issues there in my opinion on receiving messages at a fast rate in bursts.

PR #1714, opened 2020-11-17
Make sure to honor MaxAckPending when streaming directly to consumers

derekcollison:
When I moved the MaxAckPending logic to getNextMsg() it no longer enforced on streaming current inbound messages directly to consumers.

/cc @nats-io/core


kozlovic (approved):
LGTM

PR #1715, opened 2020-11-17
support json requests for stream lookup

ripienaar:
/cc @nats-io/core

### DIFF
--- a/server/jetstream_api.go
+++ b/server/jetstream_api.go
@@ -997,13 +997,11 @@ func (s *Server) jsStreamLookupRequest(sub *subscription, c *client, subject, re
 	if bytes.HasPrefix(bytes.TrimSpace(msg), []byte("{")) {
 		var req JSApiStreamLookupRequest
 		err := json.Unmarshal(msg, &req)
-		if err != nil {
-			resp.Error = &ApiError{Code: 400, Description: "invalid request"}
-			s.sendAPIResponse(c, subject, reply, string(msg), s.jsonResponse(&resp))
-			return
+		// happy path we set it, otherwise we try whatever was in the
+		// msg as if its a subject this is because '{' is a valid subject
+		if err == nil {
+			subj = req.Subject
 		}
-
-		subj = req.Subject
 	}
 
 	if !IsValidSubject(subj) {
--- a/server/jetstream_api.go
+++ b/server/jetstream_api.go
@@ -266,6 +266,11 @@ type JSApiStreamInfoResponse struct {
 
 const JSApiStreamInfoResponseType = "io.nats.jetstream.api.v1.stream_info_response"
 
+type JSApiStreamLookupRequest struct {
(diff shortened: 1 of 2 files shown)


server/jetstream_api.go
```diff
@@ -980,14 +985,29 @@ func (s *Server) jsStreamLookupRequest(sub *subscription, c *client, subject, re
...
+
 	if isEmptyRequest(msg) {
 		resp.Error = &ApiError{Code: 400, Description: "subject required"}
 		s.sendAPIResponse(c, subject, reply, string(msg), s.jsonResponse(&resp))
 		return
 	}
+
 	subj := string(msg)
+
+	if bytes.HasPrefix(msg, []byte("{")) {
```
ripienaar, line 997: it's all a bit fragile since `{"subject":"x"}` is a valid subject via`IsValidSubject` and `HasPrefix(...."{")` isnt exactly robust, could just always try to first parse as JSON I guess and check if req.Subject is empty at a bit of a cost for bare word requests.

derekcollison: We want to avoid json if we can, very expensive in its current form unfortunately, so anything in fast path needs to be scrutinized.

ripienaar: Yes, so thats what this PR does, use it only when it seems likely that its JSON.

derekcollison: Take a look at `nextReqFromMsg()` in consumer.go

derekcollison (commented):
Depending on API discussion we may want to extend the response to this one and return a matching durable consumer if one exists.

server/jetstream_api.go
ripienaar: only real difference there is `req[0] == '{'` we cant convert these 3 if statements into a switch like there, shall I use that instead of HasPrefix?

derekcollison: That line works because of Trim above.

ripienaar: added trim

derekcollison: ok, why not just check msg[0] then for '{' vs calling out to HasPrefix? Just curious.

derekcollison (approved):
LGTM

server/jetstream_api.go
ripienaar: no real reason, can do if you prefer, it's still fragile though since "{" is a valid stream name thats more what I was on about earlier saying this check is inherintly not robust.

derekcollison: If parsing fails you can fall back after detecting a '{'

ripienaar: add that

derekcollison (approved):
LGTM

ripienaar:
@derekcollison can I merge this, think enough test runners are passing atm right?

derekcollison:
Yes. Go for it.

PR #1717, opened 2020-11-17
Optimize ackReply parsing

derekcollison:
Went from ~2400ns with 8 allocations to 60ns and 0 allocations.

/cc @nats-io/core

### DIFF
--- a/server/consumer.go
+++ b/server/consumer.go
@@ -751,9 +751,9 @@ func (o *Consumer) sendAckReply(subj string) {
 
 // Process a message for the ack reply subject delivered with a message.
 func (o *Consumer) processAck(_ *subscription, _ *client, subject, reply string, msg []byte) {
-	sseq, dseq, dcount, _, _ := o.ReplyInfo(subject)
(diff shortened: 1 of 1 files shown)


server/consumer.go
```diff
@@ -1830,23 +1830,73 @@ func (o *Consumer) checkPending() {
...
 // StreamSeqFromReply will extract the stream sequence from the reply subject.
 func (o *Consumer) StreamSeqFromReply(reply string) uint64 {
-	seq, _, _, _, _ := o.ReplyInfo(reply)
-	return seq
+	sseq, _, _ := ackReplyInfo(reply)
+	return sseq
 }
 
+// Quick parser for positive numbers in ack reply encoding.
+func parseAckReplyNum(d string) (n int64) {
```
kozlovic, line 1844: We have parseInt64() already in util. You would still have to pass `[]byte(d)` instead.

kozlovic (approved):
LGTM

server/consumer.go
derekcollison: Yes and that can cause a copy..

PR #1719, opened 2020-11-17
Only fetch account jwt if the name is a proper public account key

matthiashanel:

This only cuts down on obviously bad requests that would have otherwise resulted in failures/timeouts on the receiving end

### DIFF
--- a/server/accounts.go
+++ b/server/accounts.go
@@ -3125,7 +3125,7 @@ func buildInternalNkeyUser(uc *jwt.UserClaims, acts map[string]struct{}, acc *Ac
 
 const fetchTimeout = 2 * time.Second
 
-func FetchAccount(res AccountResolver, name string) (string, error) {
+func fetchAccount(res AccountResolver, name string) (string, error) {
 	if !nkeys.IsValidPublicAccountKey(name) {
 		return "", fmt.Errorf("will only fetch valid account keys")
 	}
--- a/server/accounts.go
+++ b/server/accounts.go
@@ -3125,6 +3125,13 @@ func buildInternalNkeyUser(uc *jwt.UserClaims, acts map[string]struct{}, acc *Ac
 
 const fetchTimeout = 2 * time.Second
 
+func FetchAccount(res AccountResolver, name string) (string, error) {
+	if !nkeys.IsValidPublicAccountKey(name) {
+		return "", fmt.Errorf("will only fetch valid account keys")
+	}
+	return res.Fetch(name)
+}
+
 // AccountResolver interface. This is to fetch Account JWTs by public nkeys
 type AccountResolver interface {
 	Fetch(name string) (string, error)
--- a/server/server.go
+++ b/server/server.go
@@ -397,7 +397,7 @@ func NewServer(opts *Options) (*Server, error) {
 			s.mu.Unlock()
 			var a *Account
 			// perform direct lookup to avoid warning trace
-			if _, err := FetchAccount(ar, s.opts.SystemAccount); err == nil {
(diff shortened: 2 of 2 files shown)


server/accounts.go
```diff
@@ -3125,6 +3125,13 @@ func buildInternalNkeyUser(uc *jwt.UserClaims, acts map[string]struct{}, acc *Ac
 
 const fetchTimeout = 2 * time.Second
 
+func FetchAccount(res AccountResolver, name string) (string, error) {
```
kozlovic, line 3128: From my look at references in server code for Fetch() I see that we call it 4 times and here you have replaced 3 calls. Is that intentional that one is not replaced?

If the answer is that you missed one and they should all be replaced, then why not update Fetch() directly? Is it because this function is exposed and you don't want to break usage? It should be used only by the server, so should not be a problem. If you were expecting to have replaced all calls if Fetch to FetchAccount but don't want to break the old behavior of Fetch(), it could be an option of the account resolver. So all Fetch() would either verify or not based on that option.

server/server.go
```diff
@@ -1305,7 +1305,12 @@ func (s *Server) fetchAccountClaims(name string) (*jwt.AccountClaims, string, er
 	if err != nil {
 		return nil, _EMPTY_, err
 	}
-	return s.verifyAccountClaims(claimJWT)
+	var claim *jwt.AccountClaims
+	claim, claimJWT, err = s.verifyAccountClaims(claimJWT)
+	if claim != nil && claim.Subject != name {
```
kozlovic, line 1310: Are you sure about that? (I think I recall once that you wanted to add some check and during a meeting you realized that assumption may have been wrong, but don't know if it was about that specifically).

server/accounts.go
matthiashanel: I assume the extra fetch is part of a test? (see my search results below)
I have not changed tests to keep this change small. (the presumption is that fetch in a test will be using the correct names).

I wrap the call to `res.Fetch` inside of `FetchAccount` so the code to check the key does not need to repeat in each of the account resolver individually.

Method specification
    Fetch
Found usages  (4 usages found)
    nats-server  (4 usages found)
        server  (2 usages found)
            accounts.go  (1 usage found)
                FetchAccount  (1 usage found)
                    3132 return res.Fetch(name)
            events_test.go  (1 usage found)
                Fetch  (1 usage found)
                    2075 return sr.AccountResolver.Fetch(name)
        test  (2 usages found)
            leafnode_test.go  (2 usages found)
                TestLeafNodeSignerUser  (2 usages found)
                    1581 token, err := s.AccountResolver().Fetch(apk1)
                    1604 tt, err := s.AccountResolver().Fetch(apk1)

server/server.go
matthiashanel: This extra check assures that we actually return what we asked for, not smth. we are made to believe we asked for.
Imagine this scenario. Requesting account key AA... but get a response AB..., without this check we would end up with an account we did not ask for.
Subject contains the account public key we are asking for.

server/accounts.go
kozlovic: There is another call to Fetch() but I realize now that we pass "", which is just to check if the account resolver is reachable.

server/server.go
kozlovic: Ok

server/accounts.go
kozlovic: Since so far this only used in the server package, does this need to be exposed? Could it be fetchAccount() instead?

matthiashanel: Absolutely, will make that change. Had it public essentially because Fetch was public.

kozlovic (approved):
LGTM

PR #1721, opened 2020-11-17
Fix for not properly recovering first sequence number on recovery.

derekcollison:
When all messages where deleted we could forget the sequence number that we should be starting back off with on restart/receovery.

This also fixes a few minor other bugs.

/cc @nats-io/core


kozlovic (approved):
LGTM

PR #1725, opened 2020-11-18
[FIXED] LeafNode: duplicate queue messages in complex routing setup

### LINKED ISSUE #1722: Duplicate message occurs when configuring leaf nodes between clusters
## Defect

### Versions of `nats-server` and affected client libraries used:

2.2.0-beta.32 (latest master branch, 958c0484)
nats sample client (nats-req/nats-rply)

### OS/Container environment:

windows 10

### Steps or code to reproduce the issue:

There are two clusters, a srv cluster and a leaf cluster.
Each cluster has 3 nodes. (srv-1,2,3/leaf-1,2,3)
As shown in the figure below, The two clusters are connected by leaf connections.

[nats-config.zip](https://github.com/nats-io/nats-server/files/5556939/nats-config.zip)

#### Issue 1)

Initially I used the release 2.1.7 version.

1-1) Client(nats-rply) subscribe Foo on leaf-1
1-2) Client(nats-req) publish one message to Foo on leaf-2
1-3) Client(nats-rply) receive two message on leaf-1

I refer to the issue https://github.com/nats-io/nats-server/issues/1585, updated the nats-server version to 2.2.0-beta.32 and configured the same cluster name as 'a' on the leaf cluster and srv cluster. 
After setting it, the problem was solved.
Is this solution right?
Is this configuration correct?

#### Issue 2)

The same subscription (subject: foo queuegroup: NATS-RPLY-22) is on srv-1 and leaf-1.
Client(nats-req) publish message to foo on leaf-2.
Randomly, srv-1 or leaf-1 receive the message or both srv-1 and leaf-1 receive the message. 
This issue also occurred in 2.2.0-beta.
I attach nats-server -DV log and capture file.

[nats-log.zip](https://github.com/nats-io/nats-server/files/5556959/nats-log.zip)

### Expected result:

One node(srv-1 or leaf-1) receive the message. 
(It's better for leaf-1 always receive it.
Is there a priority for routing messages between leaf cluster and remote cluster?)

### Actual result:

Randomly, srv-1 or leaf-1 receive the message or both srv-1 and leaf-1 receive the message.

kozlovic:
Suppose a cluster of 2 servers, let's call them leaf1 and leaf2.
These servers are routed and have a leaf connection to another
server, let's call it srv1.
They share the same cluster name.

If a queue subscriber runs on srv1 and a queue subscriber on the
same subject/group name runs on leaf1, if a requestor runs on
leaf2, the request should reach only one of the 2 queue subs.

The defect was that sometimes both queue subs would receive the
message.

The added test checks that only one reply is ever received and
that the local "leaf" cluster is preferred.

Resolves #1722

### DIFF
--- a/server/client.go
+++ b/server/client.go
@@ -3818,6 +3818,8 @@ func (c *client) processMsgResults(acc *Account, r *SublistResult, msg, deliver,
 					continue
 				} else {
 					c.addSubToRouteTargets(sub)
+					// Clear rsub since we added a sub.
+					rsub = nil
 					if flags&pmrCollectQueueNames != 0 {
 						queues = append(queues, sub.queue)
 					}
--- a/server/events_test.go
+++ b/server/events_test.go
@@ -416,6 +416,7 @@ func runSolicitWithCredentials(t *testing.T, opts *Options, creds string) (*Serv
 
 // Helper function to check that a leaf node has connected to our server.
 func checkLeafNodeConnected(t *testing.T, s *Server) {
+	t.Helper()
 	checkLeafNodeConnectedCount(t, s, 1)
 }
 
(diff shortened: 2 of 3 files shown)


derekcollison (approved):
LGTM

PR #1726, opened 2020-11-19
[jetstream] Allow consumer FilterSubject wildcards for non-WorkQueue Streams

wallyqs:
This adds support in the server to allow `FilterSubjects` to use wildcards when the stream does not have a WorkQueue retention policy configuration. Let's say if a Stream as subjects `["test.one", "test.two", "test.three"]`, then it would be possible to create a consumer that only matches `test.one` explicitly, or fetch all via a wildcard like `test.*`.

 - [X] Tests added

/cc @nats-io/core

### DIFF
--- a/test/jetstream_test.go
+++ b/test/jetstream_test.go
@@ -2182,6 +2182,161 @@ func TestJetStreamWildcardSubjectFiltering(t *testing.T) {
 	}
 }
 
+func TestJetStreamWildcardSubjectFilteringEphemeralConsumer(t *testing.T) {
+	subjects := []string{"js.test.one", "js.test.two", "js.test.three"}
+	cases := []struct {
+		name    string
+		mconfig *server.StreamConfig
+	}{
+		{"MemoryStore", &server.StreamConfig{Name: "TEST", Storage: server.MemoryStorage, Subjects: subjects}},
+		{"FileStore", &server.StreamConfig{Name: "TEST", Storage: server.FileStorage, Subjects: subjects}},
+	}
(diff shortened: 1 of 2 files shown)


derekcollison (commented):
Currently we do not allow overlaps, which means the subject should be a strict subset of the streams subjects. With looking at the ability to connect streams and consumers via just subjects like we are possibly suggesting would mean that this may not work.

What I mean is that let's say I have stream A with subjects foo, bar and I have a stream B with baz. In the case below '*' should be a valid subject but I would argue it should not be since its not a strict subset of any given streams subjects and if the system was asked to match a stream and consumer to the subject '*' we would not know what to do.

For the stream with foo, bar and baz you can just create a non-filtered consumer as well.

test/jetstream_test.go
```diff
@@ -2182,6 +2182,161 @@ func TestJetStreamWildcardSubjectFiltering(t *testing.T) {
 	}
 }
 
+func TestJetStreamWildcardSubjectFilteringEphemeralConsumer(t *testing.T) {
+	subjects := []string{"js.test.one", "js.test.two", "js.test.three"}
```
wallyqs, line 2186: Changing this into `[]string{"js.test.*"}` would also make it possible to use a wildcard in the FilterSubject and the test would be the same.

wallyqs:
Thanks, I see now... changing the Subjects from a StreamConfig to include a wildcard like `"js.test.*` would also make it possible to narrow down results via the FilterSubjects as well so closing.

PR #1727, opened 2020-11-19
[ADDED] verify_cert_and_check_known_urls to tie subject alt name to url in cfg

matthiashanel:
Only works for gateways and routes. When true the subject alt DNS name
must match one url in the corresponding configuration

As discussed. At the very least have a look at the unit test cases in TestDNSAltNameMatching. Looking at certs to write my other tests I started to wonder if we should extend this to IPs too?

```
        X509v3 extensions:
            X509v3 Subject Alternative Name:
                DNS:localhost, IP Address:127.0.0.1
```

Also look for 127.0.0.1 and it's IPv6 variant. Opinions?

If yes, I'd be open to suggestions on which urls to provide in test/tls_test.go line 823, 891, 892

What I intend to include in the doc
verify_and_implicit_allow| Only settable in a non client context where verify=true is the default (gateways/cluster). The incoming connections certificate's `X509v3 Subject Alternative Name` DNS entries will be matched against all urls in the configuration context that contains this tls map. If a match is found the connection is accepted, otherwise rejected. Meaning for gateways we will match all DNS entries in the certificate against all gateway urls. For cluster we will match against all route urls. As a consequence of this, dynamic cluster growth may require config changes in other cluster where this flag is true. DNS name checking is performed according to https://tools.ietf.org/html/rfc6125#section-6.4.1 Only the full wildcard `*` is supported for the left most label.  This would be a way to keep cluster growth flexible.

### DIFF
--- a/server/auth.go
+++ b/server/auth.go
@@ -808,12 +808,12 @@ func (s *Server) isRouterAuthorized(c *client) bool {
 		return s.opts.CustomRouterAuthentication.Check(c)
 	}
 
-	if opts.Cluster.TLSMap || opts.Cluster.TLSAcceptKnownUrls {
+	if opts.Cluster.TLSMap || opts.Cluster.TLSCheckKnwonURLs {
 		return checkClientTLSCertSubject(c, func(user string, _ *ldap.DN, isDNSAltName bool) (string, bool) {
 			if user == "" {
 				return "", false
 			}
-			if opts.Cluster.TLSAcceptKnownUrls && isDNSAltName {
+			if opts.Cluster.TLSCheckKnwonURLs && isDNSAltName {
 				if dnsAltNameMatches(dnsAltNameLabels(user), opts.Routes) {
 					return "", true
 				}
@@ -844,12 +844,12 @@ func (s *Server) isGatewayAuthorized(c *client) bool {
 	opts := s.getOpts()
 
 	// Check whether TLS map is enabled, otherwise use single user/pass.
-	if opts.Gateway.TLSMap || opts.Gateway.TLSAcceptKnownUrls {
+	if opts.Gateway.TLSMap || opts.Gateway.TLSCheckKnownURLs {
 		return checkClientTLSCertSubject(c, func(user string, _ *ldap.DN, isDNSAltName bool) (string, bool) {
 			if user == "" {
 				return "", false
 			}
-			if opts.Gateway.TLSAcceptKnownUrls && isDNSAltName {
+			if opts.Gateway.TLSCheckKnownURLs && isDNSAltName {
 				labels := dnsAltNameLabels(user)
 				for _, gw := range opts.Gateway.Gateways {
 					if gw != nil && dnsAltNameMatches(labels, gw.URLs) {
--- a/server/auth.go
+++ b/server/auth.go
@@ -20,6 +20,7 @@ import (
 	"encoding/base64"
 	"fmt"
 	"net"
+	"net/url"
 	"regexp"
 	"strings"
 	"time"
@@ -418,7 +419,7 @@ func (s *Server) processClientOrLeafAuthentication(c *client, opts *Options) boo
 	} else if hasUsers {
 		// Check if we are tls verify and are mapping users from the client_certificate.
 		if tlsMap {
-			authorized := checkClientTLSCertSubject(c, func(u string, certRDN *ldap.DN) (string, bool) {
+			authorized := checkClientTLSCertSubject(c, func(u string, certRDN *ldap.DN, _ bool) (string, bool) {
 				// First do literal lookup using the resulting string representation
 				// of RDNSequence as implemented by the pkix package from Go.
 				if u != "" {
@@ -667,7 +668,7 @@ func getTLSAuthDCs(rdns *pkix.RDNSequence) string {
 	return strings.Join(dcs, ",")
 }
 
-type tlsMapAuthFn func(string, *ldap.DN) (string, bool)
+type tlsMapAuthFn func(string, *ldap.DN, bool) (string, bool)
 
 func checkClientTLSCertSubject(c *client, fn tlsMapAuthFn) bool {
 	tlsState := c.GetTLSConnectionState()
@@ -696,7 +697,7 @@ func checkClientTLSCertSubject(c *client, fn tlsMapAuthFn) bool {
 	switch {
 	case hasEmailAddresses:
 		for _, u := range cert.EmailAddresses {
-			if match, ok := fn(u, nil); ok {
+			if match, ok := fn(u, nil, false); ok {
 				c.Debugf("Using email found in cert for auth [%q]", match)
 				return true
 			}
@@ -704,7 +705,7 @@ func checkClientTLSCertSubject(c *client, fn tlsMapAuthFn) bool {
 		fallthrough
 	case hasSANs:
 		for _, u := range cert.DNSNames {
-			if match, ok := fn(u, nil); ok {
+			if match, ok := fn(u, nil, true); ok {
 				c.Debugf("Using SAN found in cert for auth [%q]", match)
 				return true
 			}
@@ -712,7 +713,7 @@ func checkClientTLSCertSubject(c *client, fn tlsMapAuthFn) bool {
 		fallthrough
 	case hasURIs:
 		for _, u := range cert.URIs {
-			if match, ok := fn(u.String(), nil); ok {
+			if match, ok := fn(u.String(), nil, false); ok {
 				c.Debugf("Using URI found in cert for auth [%q]", match)
 				return true
 			}
@@ -726,7 +727,7 @@ func checkClientTLSCertSubject(c *client, fn tlsMapAuthFn) bool {
 	// Match that follows original order from the subject takes precedence.
 	dn, err := ldap.FromCertSubject(cert.Subject)
 	if err == nil {
-		if match, ok := fn("", dn); ok {
+		if match, ok := fn("", dn, false); ok {
 			c.Debugf("Using DistinguishedNameMatch for auth [%q]", match)
 			return true
 		}
@@ -745,7 +746,7 @@ func checkClientTLSCertSubject(c *client, fn tlsMapAuthFn) bool {
 		dcs := getTLSAuthDCs(&rdns)
 		if len(dcs) > 0 {
 			u := strings.Join([]string{rdn, dcs}, ",")
-			if match, ok := fn(u, nil); ok {
+			if match, ok := fn(u, nil, false); ok {
 				c.Debugf("Using RDNSequence for auth [%q]", match)
 				return true
 			}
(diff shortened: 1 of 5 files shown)


server/opts.go
```diff
@@ -3253,6 +3258,16 @@ func parseTLS(v interface{}) (t *TLSConfigOpts, retErr error) {
 			}
 			tc.Verify = verify
 			tc.Map = verify
+		case "verify_and_implicit_allow":
+			verify, ok := mv.(bool)
+			if !ok {
+				return nil, &configErr{tk, "error parsing tls config, expected 'verify_and_map' to be a boolean"}
```
kozlovic, line 3264: `verify_an_map` is not correct field name.

test/tls_test.go
```diff
@@ -786,6 +786,167 @@ func TestTLSGatewaysCertificateCNBasedAuth(t *testing.T) {
...
+func TestTLSRoutesCertificateImplicitAllowPass(t *testing.T) {
+	testTLSRoutesCertificateImplicitAllow(t, true)
+}
+
+func TestTLSRoutesCertificateImplicitAllowFail(t *testing.T) {
+	testTLSRoutesCertificateImplicitAllow(t, false)
+}
+
+func testTLSRoutesCertificateImplicitAllow(t *testing.T, pass bool) {
+	// Base config for the servers
```
kozlovic, line 798: Wonder if you don't need `t.Helper()` here?

test/tls_test.go
```diff
@@ -786,6 +786,167 @@ func TestTLSGatewaysCertificateCNBasedAuth(t *testing.T) {
...
+	cfg, err := ioutil.TempFile("", "cfg")
+	if err != nil {
+		t.Fatal(err)
+	}
+	if err = cfg.Sync(); err != nil {
```
kozlovic, line 803: Add a `defer os.Remove(cfg)` before this line.

test/tls_test.go
```diff
@@ -786,6 +786,167 @@ func TestTLSGatewaysCertificateCNBasedAuth(t *testing.T) {
...
+func TestTLSGatewaysCertificateImplicitAllowPass(t *testing.T) {
+	testTLSGatewaysCertificateImplicitAllow(t, true)
+}
+
+func TestTLSGatewaysCertificateImplicitAllowFail(t *testing.T) {
+	testTLSGatewaysCertificateImplicitAllow(t, false)
+}
+
+func testTLSGatewaysCertificateImplicitAllow(t *testing.T, pass bool) {
+	// Base config for the servers
```
kozlovic, line 865: Same than for other test, add `t.Helper()`?

test/tls_test.go
```diff
@@ -786,6 +786,167 @@ func TestTLSGatewaysCertificateCNBasedAuth(t *testing.T) {
...
+	cfg, err := ioutil.TempFile("", "cfg")
+	if err != nil {
+		t.Fatal(err)
+	}
+	if err = cfg.Sync(); err != nil {
```
kozlovic, line 870: `defer os.Remove(cfg)`

server/auth.go
```diff
@@ -803,9 +844,23 @@ func (s *Server) isGatewayAuthorized(c *client) bool {
 	}
 
 	// Check whether TLS map is enabled, otherwise use single user/pass.
-	if opts.Gateway.TLSMap {
-		return checkClientTLSCertSubject(c, func(user string, _ *ldap.DN) (string, bool) {
-			return "", opts.Gateway.Username == user
+	if opts.Gateway.TLSMap || opts.Gateway.TLSImplicitAllow {
+		return checkClientTLSCertSubject(c, func(user string, _ *ldap.DN, isDNSAltName bool) (string, bool) {
+			if user == "" {
```
kozlovic, line 849: In original code, is that possible that `user` and `Gateway.Username` be empty in TLSMap context? If so, this is changing the behavior. Is that intended?

server/auth.go
```diff
@@ -803,9 +844,23 @@ func (s *Server) isGatewayAuthorized(c *client) bool {
...
+				return "", false
+			}
+			if opts.Gateway.TLSImplicitAllow && isDNSAltName {
+				labels := dnsAltNameLabels(user)
+				for _, gw := range opts.Gateway.Gateways {
+					if gw != nil && dnsAltNameMatches(labels, gw.URLs) {
+						return "", true
+					}
+				}
```
kozlovic, line 858: Help me understand if we are in this `if` statement, and we did not return true yet, should then we fail here? Or is it intended that we fall back to TLSMap check.

server/auth.go
```diff
@@ -764,6 +765,38 @@ func checkClientTLSCertSubject(c *client, fn tlsMapAuthFn) bool {
...
+func dnsAltNameLabels(dnsAltName string) []string {
+	return strings.Split(strings.ToLower(dnsAltName), ".")
+}
+
+// Check DNS name according to https://tools.ietf.org/html/rfc6125#section-6.4.1
+func dnsAltNameMatches(dnsAltNameLabels []string, urls []*url.URL) bool {
+URLS:
+	for _, url := range urls {
+		if url == nil {
+			continue URLS
```
kozlovic, line 777: What is your intent here? Moving to the next url? If so, you don't need the `URLS:` label and `continue` alone will just do that.

server/auth.go
```diff
@@ -764,6 +765,38 @@ func checkClientTLSCertSubject(c *client, fn tlsMapAuthFn) bool {
...
+		}
+		i := 0
+		// only match wildcard on left most label
+		if dnsAltNameLabels[0] == "*" {
+			i++
+		}
+		for ; i < len(dnsAltNameLabels); i++ {
+			if dnsAltNameLabels[i] != hostLabels[i] {
+				continue URLS
```
kozlovic, line 792: Oh, I guess because of this one?

server/opts.go
```diff
@@ -57,20 +57,21 @@ func NoErrOnUnknownFields(noError bool) {
...
+	Host             string            `json:"addr,omitempty"`
+	Port             int               `json:"cluster_port,omitempty"`
+	Username         string            `json:"-"`
+	Password         string            `json:"-"`
+	AuthTimeout      float64           `json:"auth_timeout,omitempty"`
+	Permissions      *RoutePermissions `json:"-"`
+	TLSTimeout       float64           `json:"-"`
+	TLSConfig        *tls.Config       `json:"-"`
+	TLSMap           bool              `json:"-"`
+	TLSImplicitAllow bool              `json:"-"`
```
kozlovic, line 70: To be honest, it is not clear to me what this option does. Default will be false, so can you describe to me what setting this boolean to true will mean?

kozlovic (changes requested):
Some tests changes.

server/opts.go
matthiashanel: fixed

test/tls_test.go
matthiashanel: added

test/tls_test.go
matthiashanel: added

test/tls_test.go
matthiashanel: added

test/tls_test.go
matthiashanel: added

server/auth.go
matthiashanel: I don't think that user should ever be empty. In that case the CA would have signed a cert with empty fields.
this is more like a guard I added out of habit. remove?

I'm also not sure. if you can even configure an empty user name.

server/auth.go
matthiashanel: Yes, I added it primarily because I need the label in one place, so I opted to provide it wherever it applies.
remove where unnecessary?

server/auth.go
matthiashanel: yes

server/auth.go
kozlovic: I was just asking, because maybe we need this check and it wasn't there before. Actually, looking at isGatewayAuthorized() top of func we already return true if `opts.Gateway.Username == ""`

server/auth.go
matthiashanel: I'm have to go through all gateways. 
Just because GW1 didn't contain a matching url does not mean that GW2 wouldn't.
Your comment would apply to the cluster case as well, there only one list exists.

I decided to support setting both flags (verify_and_map verify_and_implicit_allow) to true.
I did so primarily to avoid having to write the check to cause a config error if both are set.

server/auth.go
kozlovic: I would.

server/opts.go
matthiashanel: the doc will say:
verify_and_implicit_allow| Only settable in a non client context where verify=true is the default (gateways/cluster). The incoming connections certificate's `X509v3 Subject Alternative Name` DNS entries will be matched against all urls in the configuration context that contains this tls map. If a match is found the connection is accepted, otherwise rejected. Meaning for gateways we will match all DNS entries in the certificate against all gateway urls. For cluster we will match against all route urls. As a consequence of this, dynamic cluster growth may require config changes in other cluster where this flag is true. DNS name checking is performed according to https://tools.ietf.org/html/rfc6125#section-6.4.1 Only the full wildcard `*` is supported for the left most label.

matthiashanel: clearer?

kozlovic: The explanation seem clear to me, but I am not sure about the option name. Maybe @philpennock and @davidkemper can comment on that?

server/auth.go
matthiashanel: removed

server/auth.go
kozlovic: I think this function (outside of tests at least) is always called with a user, so we should not need this test.
But if we think that cert.EmailAddress, etc.. may contain (legally or maliciously) empty values and if that would cause a failure for dnsAltNameLabels(), then you could leave the check.

matthiashanel: I just guard against something weird coming out of the cert. that said. I realized that it is missing for the cluster

server/auth.go
kozlovic: My question is more: if opts.Gateway.TLSImplicitAllow is true, but isDNSAltName is false, it means what? That we are not in a client context and we should not perform this check anyway. But if isDNSAltName is true and we did not find a match, does that makes sense to fallback to tls_map check only? I don't know.

matthiashanel: if isDNSAltName is false then the argument user does not contain a value coming out of a DNS entry, hence it can't be used for matching, even if isDNSAltName allow is true.

 When presented with an email for example, we shouldn't try to tokenize the ip address.
```
        X509v3 extensions:
            X509v3 Subject Alternative Name:
                IP Address:127.0.0.1
```
We get a call where the name is 127.0.0.1 and isDNSAltName = false.
If tls_map happens to be true we check if 127.0.0.1 is configured as user.

if isDNSAltName is true and say the user is localhost, and we fail to find localhost in our config and if the user also enabled tls_map, why not see if the user localhost is configured?
if tls_map is not false, we'll default to false at the end of the function.

matthiashanel:
@kozlovic @philpennock @davidkemper 

Talked to @philpennock , let's hold off on IPs for now.
As for the name, here are more suggestions, any (partly/fully) sound particular appealing or descriptive?:
verify_and_inherit_allowed_names
verify_and_map_listed_names
verify_and_map_existing_hostnames
verify_and_map_names

kozlovic:
Maybe `verify_and_map_listed_names` or `verify_and_map_existing_hostnames`? But I am very bad with names, so I would let maybe @davidkemper weigh in.

kozlovic (approved):
LGTM.. let's see if others have some input for the new config option name.

server/opts.go
```diff
@@ -81,20 +81,20 @@ type ClusterOpts struct {
...
+	Name               string               `json:"name"`
+	Host               string               `json:"addr,omitempty"`
+	Port               int                  `json:"port,omitempty"`
+	Username           string               `json:"-"`
+	Password           string               `json:"-"`
+	AuthTimeout        float64              `json:"auth_timeout,omitempty"`
+	TLSConfig          *tls.Config          `json:"-"`
+	TLSTimeout         float64              `json:"tls_timeout,omitempty"`
+	TLSMap             bool                 `json:"-"`
+	TLSAcceptKnownUrls bool                 `json:"-"`
```
kozlovic, line 93: For some reason, I feel like this name sounds less restrictive when enabled than when it is disabled, which is not the case. Meaning that the default is `false`, which sounds like the server is not accepting something, while in reality, when false it means that it may accept things that it should not.

Would `TLSCheckKnownUrls` be better? I know we agreed on `verify_and_accept_known_urls` but maybe we could also change that one with `verify_and_check_known_urls`. If the `verify_and_check_` seem strange, maybe having `verify_cert_and_check_known_urls` lays out explicitly what this is about.

kozlovic (approved):
LGTM

PR #1728, opened 2020-11-20
[Adding] ability to reference $SYS in leaf node config

matthiashanel:
only do so if $SYS is the system account

shortcut if all I want is forward it. Avoid unnecessary configuration.


kozlovic (approved):
LGTM

PR #1729, opened 2020-11-20
Enable JetStream streams and consumer access to be exported

derekcollison:
This allows basic access to publish into a stream and receive the pub ack across accounts, as well as pull and push based receipt of consumer messages and the ability to ack those messages.

/cc @nats-io/core


kozlovic (approved):
LGTM

PR #1731, opened 2020-11-21
[FIXED] Detect service import cycles.

derekcollison:
Simple configs that represent a service export/import cycle between accounts could crash the server.

/cc @nats-io/core

### DIFF
--- a/server/accounts.go
+++ b/server/accounts.go
@@ -1343,8 +1343,24 @@ func (a *Account) AddServiceImportWithClaim(destination *Account, from, to strin
 		return ErrServiceImportAuthorization
 	}
 
+	if a.importFormsCycle(destination, from, to) {
+		return ErrServiceImportFormsCycle
+	}
+
 	_, err := a.addServiceImport(destination, from, to, imClaim)
 	return err
+}
+
+// Detects if we have a cycle.
+func (a *Account) importFormsCycle(destination *Account, from, to string) bool {
+	// Check that what we are importing is not something we also export.
+	if a.serviceExportOverlaps(to) {
+		// So at this point if destination account is also importing from us, that forms a cycle.
+		if destination.serviceImportOverlaps(from) {
+			return true
+		}
+	}
+	return false
 }
 
 // SetServiceImportSharing will allow sharing of information about requests with the export account.
@@ -1583,6 +1599,30 @@ func (a *Account) checkForReverseEntry(reply string, si *serviceImport, checkInt
 			}
 		}
 	}
+}
+
+// Internal check to see if the to subject overlaps with another export.
+func (a *Account) serviceExportOverlaps(to string) bool {
+	a.mu.RLock()
+	defer a.mu.RUnlock()
+	for subj := range a.exports.services {
+		if to == subj || SubjectsCollide(to, subj) {
+			return true
+		}
+	}
+	return false
+}
+
+// Internal check to see if the from subject overlaps with another import.
+func (a *Account) serviceImportOverlaps(from string) bool {
+	a.mu.RLock()
+	defer a.mu.RUnlock()
+	for subj := range a.imports.services {
+		if from == subj || SubjectsCollide(from, subj) {
+			return true
+		}
+	}
+	return false
 }
 
 // Internal check to see if a service import exists.
--- a/server/client.go
+++ b/server/client.go
@@ -3705,6 +3705,7 @@ func (c *client) processMsgResults(acc *Account, r *SublistResult, msg, deliver,
 			}
 			continue
 		}
+
 		// Assume delivery subject is the normal subject to this point.
 		dsubj = subj
 
--- a/test/service_latency_test.go
+++ b/test/service_latency_test.go
@@ -1303,6 +1303,72 @@ func TestServiceAndStreamStackOverflow(t *testing.T) {
 	})
 }
 
+func TestServiceCycle(t *testing.T) {
+	t.Skip("Remove the skip to run this test and shows the issue")
+
+	conf := createConfFile(t, []byte(`
+		accounts {
+		  A {
+		    users = [ { user: "a", pass: "a" } ]
+		    exports [
+		      { service: help }
+			]
+			imports [
+		      { service { subject: help, account: B } }
+		    ]
+		  }
+		  B {
+		    users = [ { user: "b", pass: "b" } ]
+		    exports [
+		      { service: help }
+			]
+			imports [
+		      { service { subject: help, account: A } }
+		    ]
+		  }
+		}
+	`))
+	defer os.Remove(conf)
+
+	srv, opts := RunServerWithConfig(conf)
+	defer srv.Shutdown()
+
+	// Responder (just request sub)
+	nc, err := nats.Connect(fmt.Sprintf("nats://a:a@%s:%d", opts.Host, opts.Port))
+	if err != nil {
+		t.Fatalf("Error on connect: %v", err)
+	}
+	defer nc.Close()
+
+	cb := func(m *nats.Msg) {
+		m.Respond(nil)
+	}
+	sub, _ := nc.Subscribe("help", cb)
+	nc.Flush()
+
+	// Requestor
+	nc2, err := nats.Connect(fmt.Sprintf("nats://b:b@%s:%d", opts.Host, opts.Port))
+	if err != nil {
+		t.Fatalf("Error on connect: %v", err)
+	}
+	defer nc2.Close()
+
+	// Send a single request.
+	if _, err := nc2.Request("help", []byte("hi"), time.Second); err != nil {
+		t.Fatal("Did not get the reply")
+	}
+
+	// Make sure works for queue subscribers as well.
+	sub.Unsubscribe()
+	sub, _ = nc.QueueSubscribe("help", "prod", cb)
+	nc.Flush()
+
+	// Send a single request.
+	if _, err := nc2.Request("help", []byte("hi"), time.Second); err != nil {
+		t.Fatal("Did not get the reply")
+	}
+}
+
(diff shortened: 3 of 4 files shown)


kozlovic (approved):
LGTM

server/accounts.go
```diff
@@ -1343,6 +1343,14 @@ func (a *Account) AddServiceImportWithClaim(destination *Account, from, to strin
 		return ErrServiceImportAuthorization
 	}
 
+	// Check that what we are importing is not something we also export.
+	if a.serviceExportExists(to) {
+		// So at this point if destination account is also importing from us, that forms a cycle.
```
matthiashanel, line 1348: Would this also happen if the cycle would be bigger, A - B -C - A?

matthiashanel (commented):
change LGTM.
But would this have happened with larger cycles as well?

server/accounts.go
derekcollison: You raise a good point. I have to detect cycles at runtime. I will make it similar to routes in that once you are mapped via a service import you can't be mapped again. Will keep what we have but do a test like above.

derekcollison:
Ok I updated and do not try to detect configuration cycles but instead catch it at runtime.

Please take another look. Thanks.

test/service_latency_test.go
```diff
@@ -1303,6 +1303,62 @@ func TestServiceAndStreamStackOverflow(t *testing.T) {
 	})
 }
 
+func TestServiceCycle(t *testing.T) {
+	conf := createConfFile(t, []byte(`
+		accounts {
+		  A {
+		    users = [ { user: "a", pass: "a" } ]
+		    exports [ { service: help } ]
+			imports [ { service { subject: help, account: B } } ]
```
kozlovic, line 1312: Is tabulation here intentional?

test/service_latency_test.go
```diff
@@ -1303,6 +1303,62 @@ func TestServiceAndStreamStackOverflow(t *testing.T) {
...
+		  }
+		  B {
+		    users = [ { user: "b", pass: "b" } ]
+		    exports [ { service: help } ]
+			imports [ { service { subject: help, account: A } } ]
```
kozlovic, line 1317: Is tabulation here intentional?

kozlovic (approved):
LGTM. To be clear, we are not failing this as a misconfiguration now, just make sure we are not cycling, right?

server/client.go
```diff
@@ -3705,6 +3705,13 @@ func (c *client) processMsgResults(acc *Account, r *SublistResult, msg, deliver,
 			}
 			continue
 		}
+
+		// If this sub is an internal account subscription, meaning a service import, and this
+		// message has already been mapped via a previous service import, skip.
+		if flags&pmrMsgImportedFromService != 0 && sub.icb != nil && sub.client.kind == ACCOUNT {
+			continue
```
matthiashanel, line 3712: Not an actual code change but maybe add a todo in all the places in this and the previous PR so that in a subsequent change we can set a flag in either the client or the account to maybe record this event?
Similar to incomplete in the account. If a cycle is the only way this can happen then maybe the account.
if this can happen for some other reason (say reconnect) maybe in the client.

test/service_latency_test.go
```diff
@@ -1303,6 +1303,62 @@ func TestServiceAndStreamStackOverflow(t *testing.T) {
 	})
 }
 
+func TestServiceCycle(t *testing.T) {
```
matthiashanel, line 1306: Seems like a test with 3 accounts is warranted as well.

test/service_latency_test.go
derekcollison: yes

test/service_latency_test.go
derekcollison: same

server/client.go
derekcollison: Now that I am looking at my next problem I think this won't work since we do want to daisy chain service imports, we just do not want cycles.

derekcollison: So I may go back to what I had before and try to expand to detect complex loops.

matthiashanel (commented):
For simple self imports, in jwt I intend to turn this into a validation error.

server/client.go
matthiashanel: > Now that I am looking at my next problem I think this won't work since we do want to daisy chain service imports, we just do not want cycles.

@derekcollison wouldn't you want to go hub/spoke vs daisy chain? Seems to me that the workaround would be to import from the original account directly. What's the daisy chain use case, proxy accounts, easy forwarding from private to public imports?

derekcollison:
Yes I am going back to detecting the cycle when you add a new import.

server/client.go
derekcollison: Daisy chain is export $JS.API.>. Will go from account A to account B to $SYS.

derekcollison:
ok pls take one more look.

server/accounts.go
```diff
@@ -1343,10 +1343,26 @@ func (a *Account) AddServiceImportWithClaim(destination *Account, from, to strin
...
+	}
+
 	_, err := a.addServiceImport(destination, from, to, imClaim)
 	return err
 }
 
+// Detects if we have a cycle.
+func (a *Account) importFormsCycle(destination *Account, from, to string) bool {
+	// Check that what we are importing is not something we also export.
+	if a.serviceExportOverlaps(to) {
```
matthiashanel, line 1357: This covers export -> mapping -> import ?

server/accounts.go
```diff
@@ -1343,10 +1343,26 @@ func (a *Account) AddServiceImportWithClaim(destination *Account, from, to strin
...
+		// So at this point if destination account is also importing from us, that forms a cycle.
+		if destination.serviceImportOverlaps(from) {
```
matthiashanel, line 1359: looks like this would trigger if:  a importFormsCycle b
a serviceExportOverlaps with c
b serviceImportOverlaps with d

Seems the checks do not take accounts fully into account?

server/accounts.go
```diff
@@ -1343,10 +1343,26 @@ func (a *Account) AddServiceImportWithClaim(destination *Account, from, to strin
...
+	if a.importFormsCycle(destination, from, to) {
+		return ErrServiceImportFormsCycle
+	}
+
 	_, err := a.addServiceImport(destination, from, to, imClaim)
 	return err
 }
 
+// Detects if we have a cycle.
+func (a *Account) importFormsCycle(destination *Account, from, to string) bool {
```
matthiashanel, line 1355: Are we doing direct cycles here?
I would have expected a depth first search alternating between exports/imports facilitating a queue to avoid recursion.

test/service_latency_test.go
```diff
@@ -1303,6 +1303,66 @@ func TestServiceAndStreamStackOverflow(t *testing.T) {
...
+	conf = createConfFile(t, []byte(`
+		accounts {
+		  A {
+		    exports [ { service: * } ]
+			imports [ { service { subject: help, account: B } } ]
+		  }
+		  B {
+		    exports [ { service: help } ]
+			imports [ { service { subject: help, account: C } } ]
+		  }
```
matthiashanel, line 1352: I'd say the above check finds the loop here?

how about A export a, B import a to b and export b, C import b to c to d and export d

Then another test that rewrites subjects via mappings as well.

matthiashanel:
Also

> Daisy chain is export $JS.API.>. Will go from account A to account B to $SYS.

Why not A -> B and B to $SYS instead and avoid having to support this?
Also, would A to B and C to B and everything in B to $SYS work?

server/accounts.go
derekcollison: Mapping is for client inbound messages only, so that takes effect before any service import processing.

server/accounts.go
derekcollison: It does take accounts into consideration since it looks at itself first during an import and sees if we have an overlapping export. AFAIK this is required to form any cycle. We then check the destination which is the destination account against the "from" subject to see if it overlaps in the destination account with its imports.

server/accounts.go
derekcollison: Yes we detect direct cycles here, hence the original test with just help and two accounts.

test/service_latency_test.go
derekcollison: I have a test with 3 accounts. I could add this one if you feel strongly about it.

derekcollison:
I figured out the daisy chaining and export/import for all of JS in another PR.

kozlovic (approved):
LGTM

server/accounts.go
matthiashanel: I meant more like only direct cycles (involving 2 accounts)  vs all cycles (involving more accounts)

test/service_latency_test.go
matthiashanel: Well, if we only detect direct cycles, then a test with 3 accounts - no direct cycle - is not needed.
If we do want to detect cycles between more than two accounts I feel strongly about such a test.
My reading of the above code is that you get an error because of these two accounts alone.

server/accounts.go
matthiashanel: ok

server/accounts.go
matthiashanel: Can a.serviceExportOverlaps(to) return true for a reason that has nothing to do with destination?
Similarly, can destination.serviceImportOverlaps(from) return true for a reason that has nothing to do with a?

On further reflection, I also wonder if we should prevent cycles irrespective of the subjects involved.

test/service_latency_test.go
derekcollison: I have tests with wildcards and import mappings as well IIRC>

flymoondust:
I want to know if this bug will release a new version?

PR #1733, opened 2020-11-22
Allow complete $JS.API to be imported from another account.

derekcollison:

/cc @nats-io/core

### DIFF
--- a/server/client.go
+++ b/server/client.go
@@ -3534,6 +3534,13 @@ func (c *client) processServiceImport(si *serviceImport, acc *Account, msg []byt
 	// so we only lock once.
 	to, _ = si.acc.selectMappedSubject(to)
 
+	oreply, oacc := c.pa.reply, c.acc
+	c.pa.reply = nrr
+	if !si.isSysAcc {
+		c.mu.Lock()
+		c.acc = si.acc
+		c.mu.Unlock()
+	}
 	// FIXME(dlc) - Do L1 cache trick like normal client?
 	rr := si.acc.sl.Match(to)
 
@@ -3568,6 +3575,13 @@ func (c *client) processServiceImport(si *serviceImport, acc *Account, msg []byt
 
 	// Put what was there back now.
 	c.in.rts = orts
+	c.pa.reply = oreply
+
+	if !si.isSysAcc {
+		c.mu.Lock()
+		c.acc = oacc
+		c.mu.Unlock()
+	}
 
 	// Determine if we should remove this service import. This is for response service imports.
 	// We will remove if we did not deliver, or if we are a response service import and we are
--- a/server/accounts.go
+++ b/server/accounts.go
@@ -131,6 +131,7 @@ type serviceImport struct {
 	share       bool
 	tracking    bool
 	didDeliver  bool
+	isSysAcc    bool
 	trackingHdr http.Header // header from request
 }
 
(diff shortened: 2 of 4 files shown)


ripienaar:
This works, but I have some questions thats perhaps not directly about this PR.

What are the plans for the metadata here? The reply subject is of course munged now so we cant extract from that, hoped we can avoid going to headers but seems likely?

Advisories being published when accessing the API from the account that does the import has:

```json
  "client": {
    "host": "::1",
    "port": 36248,
    "cid": 17,
    "account": "test",
    "user": "jsi",
    "name": "NATS CLI Version development",
    "lang": "go",
    "version": "1.11.0"
  },
```

This is incorrect, the user is indeed `jsi` but that user is in the account `jsi` not `test` - `test` is doing the export.

```
test {
  jetstream: {
    max_mem: 1G
    max_file: 1G
    max_streams: 5
    max_consumers: 5
  }

  exports = [
    { service: "$JS.API.>" }
  ]

  users = [
    {user: test, password: test}
  ]
}

jsi {
  imports = [
    {service: {account: test, subject: "$JS.API.>" }, to: "$JS.API.>" } ]

  users = [
    {user: jsi, password: test}
  ]
}
```

derekcollison:
Let me think about the munging a bit more.

For the account how are you pulling that one in?

ripienaar:
I was just subscribing to the advisory subject in the account thats enabled for jetstream - the exporting 'test' account

derekcollison:
What did your server config look like?

ripienaar:
as above, here's the full thing

```
jetstream: {
  max_mem: 1G
  max_file: 10G
}

accounts {
  test {
    jetstream: {
      max_mem: 1G
      max_file: 1G
      max_streams: 5
      max_consumers: 5
    }

    exports = [
      { service: "$JS.API.>" }
      { stream: "js.out.test.push" }
    ]

    users = [
      {user: test, password: test}
    ]
  }

  jsi {
    imports = [
      {service: {account: test, subject: "$JS.API.>" }, to: "$JS.API.>" }
      {stream: {account: test, subject: "js.out.test.push" }, to "js.push" }
    ]


    users = [
      {user: jsi, password: test}
    ]
  }

  system {
    users = [
      {user: system, password: system}
    ]
  }
}

system_account: system
~
```

derekcollison:
How can you see the advisories? Aren't they a stream export?

derekcollison:
My apologies, thought you were looking at advisories from the jsi account.

ripienaar:
LGTM in general though, follow up PRs can fix the weirdness with advisories etc

derekcollison:
Agree.

server/client.go
```diff
@@ -3534,6 +3534,13 @@ func (c *client) processServiceImport(si *serviceImport, acc *Account, msg []byt
 	// so we only lock once.
 	to, _ = si.acc.selectMappedSubject(to)
 
+	oreply, oacc := c.pa.reply, c.acc
+	c.pa.reply = nrr
+	if !si.isSysAcc {
+		c.mu.Lock()
+		c.acc = si.acc
```
kozlovic, line 3541: For some reason, I find this scary.. And if I look down the code, I see that we pass si.acc to processMsgResults/sendMsgToGateways. So is it that somewhere deep into those calls there is a reference to c.acc where we should have instead use the passed account? Would that not be safer to fix those places?

derekcollison: We needed to do something like this for daisy chaining service imports together that involved the $SYS account like JS. The bug was allowing JS to work.

I will keep working on this with followup PRs as well since complete behavior is not correct just yet for full import of JS.

PR #1734, opened 2020-11-22
Unbreak FreeBSD compilation (syscall.Statfs_t)

philpennock:
The types of fields in syscall.Statfs_t varies between platforms.
fs.Bavail is uint64 on Linux and int64 on FreeBSD.  This is the opposite way
around to fs.Bsize.

For now, just coerce the Bavail to be uint64.

If the VFS layer might return -1 for one or the other of these then these casts
will be problematic and we'll need more safeguards.

/cc @nats-io/core


derekcollison (approved):
LGTM

PR #1735, opened 2020-11-23
Stream publish changes

derekcollison:
This PR does two things. 

1. Changes the ack from a publish to a formal api response.
2. We add in the ability to set headers to check for expected stream name, last sequence and last msg ID.

Second item reference: #1648 

/cc @nats-io/core

### DIFF
--- a/server/stream.go
+++ b/server/stream.go
@@ -52,11 +52,19 @@ type StreamConfig struct {
 	Duplicates   time.Duration   `json:"duplicate_window,omitempty"`
 }
 
+const JSApiPubAckResponseType = "io.nats.jetstream.api.v1.pub_ack_response"
+
+// JSPubAckResponse is a formal response to a publish operation.
+type JSPubAckResponse struct {
+	ApiResponse
+	*PubAck
+}
+
 // PubAck is the detail you get back from a publish to a stream that was successful.
 // e.g. +OK {"stream": "Orders", "seq": 22}
 type PubAck struct {
 	Stream    string `json:"stream"`
-	Seq       uint64 `json:"seq"`
+	Sequence  uint64 `json:"seq"`
 	Duplicate bool   `json:"duplicate,omitempty"`
 }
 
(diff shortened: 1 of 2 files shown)


server/stream.go
```diff
@@ -181,13 +195,14 @@ func (a *Account) AddStreamWithStore(config *StreamConfig, fsConfig *FileStoreCo
...
 
-	// Create our pubAck here. This will be reused and for +OK will contain JSON
-	// for stream name and sequence.
-	longestSeq := strconv.FormatUint(math.MaxUint64, 10)
-	lpubAck := len(OK) + len(cfg.Name) + len("{\"stream\": ,\"seq\": }") + len(longestSeq)
-	mset.pubAck = make([]byte, 0, lpubAck)
-	mset.pubAck = append(mset.pubAck, OK...)
-	mset.pubAck = append(mset.pubAck, fmt.Sprintf(" {\"stream\": %q, \"seq\": ", cfg.Name)...)
+	// Create our pubAck template here. Better than json marshal each time on success.
+	b, _ := json.Marshal(&JSPubAckResponse{
```
kozlovic, line 199: So as it stands, this is breaking "clients" that check for +OK in puback response. I don't think that there were client releases that have that yet, but just checking with you guys. (The GO client's jetstream.go does check for +OK, but that is in main branch but not released yet, right? @ripienaar)
May also have to check with Colin, Alberto, etc.. if my understanding is correct.

kozlovic (approved):
LGTM... just note about breaking code that already exists checking for +OK.

server/stream.go
ripienaar: lots in jsm, but thats fine.

ripienaar:
@aricart note publish api changes

server/stream.go
ColinSullivan1: Just discovered it in java, all good, will fix.  Building the main branch for CI detects breaking changes quickly.  :)

derekcollison:
We may pull type out of the response for PubAck, will keep folks updated.

Issue #1080, opened 2019-07-24
please add freebsd support package in release


blusewang:
see title

kozlovic:
Ironically, we would not be able to cross-compile (we use CGO_ENABLED=off) because of this commit that added freebsd support :-) https://github.com/nats-io/nats-server/commit/a1bda5cad507e8ae09d1a4757848bb0585ff8d7a

Adding freebsd in our cross-compile script would produce error:
```
Stderr: go build github.com/nats-io/nats-server/server/pse: build constraints exclude all Go files in /Users/ivan/dev/go/src/github.com/nats-io/nats-server/server/pse
```
which leads to: https://github.com/golang/go/issues/24068.

That being said, I believe that NATS Server is part of the FreeBSD. There is a maintainer in our Slack that use to say that he updated it at each new release. Will try to reach out on Slack.

cougarwww:
please add freebsd support package in release

kozlovic:
@cougarwww We can't for the reason stated in previous comments. We build using CGO_ENABLED=off and the server/pse for freebsd uses C, so we would have to change that first.

philpennock:
For clarity: the above PR which fixes this issue does so by doing exactly what @kozlovic said, because he is of course right.

The above PR adds support for cross-compiling _only_ for FreeBSD/amd64.  There are struct offsets which need to be pre-calculated on any architecture where we want to be able to avoid cgo, and I only have access to an amd64 box for this.

If you care about other hardware architectures, please take a look at the PR (which I'm cautiously optimistic will land) and create a per-architecture file in `server/pse/pse_freebsd_${GOARCH}.go` which does the same things.  We can figure out as those arise if we want to just copy `pse_freebsd_amd64.go` or if it makes more sense to abstract out the per-architecture constants and endianness variable to reduce code duplication.

philpennock:
The next release of binaries for `nats-server` should include a FreeBSD/amd64 binary.

PR #1737, opened 2020-11-23
Cross-compile for FreeBSD/amd64 without cgo

philpennock:
(These commits should be preserved in merge, not squashed.)

The first commit fixes FreeBSD %cpu calculation, which was wrong by a factor of
2048, and provides an alternative implementation on amd64 which avoids the need
for cgo.  The framework setup is extensible to other architectures.

The second commit adds goreleaser targeting FreeBSD/amd64 as part of our usual
release artifact generation.

f088660d
   goreleaser target freebsd/amd64

80d11fa1
   FreeBSD fixes: %cpu scale; no-cgo for amd64

   It would be convenient if we did not need to use cgo to compile for FreeBSD
   for common architectures, allowing builds to be generated in Linux CI
   flows.

   The sysctl interface uses offsets which can vary by build architecture, so
   doing this in the general case without cgo is not realistic.  But we can
   front-load the C work to get the offsets for a given architecture, then use
   encoding/binary at run-time.

   While doing this, I saw that the existing FreeBSD code was mis-calculating
   `%cpu` by converting the `fixpt_t` scaled int straight to a double without
   dividing by the scaling factor, so we also fix for all other architectures
   by introducing a division by `FSCALE`.

   The offsets-emitting code is in `freebsd.txt`, with the filename chosen to
   keep the Go toolchain from picking it up and trying to compile.

   The result is unsafe-free cgo-free builds for FreeBSD/amd64.

/cc @nats-io/core

### DIFF
--- a/server/pse/pse_freebsd.go
+++ b/server/pse/pse_freebsd.go
@@ -11,6 +11,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+// +build !amd64
+
 package pse
 
 /*
@@ -54,7 +56,7 @@ int getusage(double *pcpu, unsigned int *rss, unsigned int *vss)
 
     *rss = pagetok(kp.ki_rssize);
     *vss = kp.ki_size;
-    *pcpu = kp.ki_pctcpu;
+    *pcpu = (double)kp.ki_pctcpu / FSCALE;
 
     return 0;
 }
--- a/server/pse/pse_freebsd_amd64.go
+++ b/server/pse/pse_freebsd_amd64.go
@@ -1,0 +1,91 @@
+// Copyright 2015-2020 The NATS Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// This is the amd64-specific FreeBSD implementation, with hard-coded offset
+// constants derived by running freebsd.txt; having this implementation allows
+// us to compile without CGO, which lets us cross-compile for FreeBSD from our
+// CI system and so supply binaries for FreeBSD amd64.
+//
+// To generate for other architectures:
+//   1. Update pse_freebsd.go, change the build exclusion to exclude your arch
+//   2. Copy this file to be built for your arch
+//   3. Update `nativeEndian` below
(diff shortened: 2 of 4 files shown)


philpennock:
Fixes #1080

server/pse/pse_freebsd.go
```diff
@@ -54,7 +56,7 @@ int getusage(double *pcpu, unsigned int *rss, unsigned int *vss)
 
     *rss = pagetok(kp.ki_rssize);
     *vss = kp.ki_size;
-    *pcpu = kp.ki_pctcpu;
+    *pcpu = (double)kp.ki_pctcpu / FSCALE;
```
derekcollison, line 59: IIRC this number can go up to 100% X num cores, so if you have 8 cores you can see numbers like 640%. Not sure if that is what is being done here.

server/pse/pse_freebsd_amd64.go
```diff
@@ -0,0 +1,87 @@
...
+	f_pcpu := float64(r_pcpu) / float64(2048)
+
+	// To get the physical page size, the C library checks two places:
+	//   process ELF auxiliary info, AT_PAGESZ
+	//   as a fallback, the hw.pagesize sysctl
+	// In looking closely, I found that the Go runtime support is handling
+	// this for us, and exposing that as syscall.Getpagesize, having checked
+	// both in the same ways, at process start, so a call to that should return
+	// a memory value without even a syscall bounce.
+	pagesize := syscall.Getpagesize()
```
derekcollison, line 74: Does this change? Maybe call once and remember it?

server/pse/pse_freebsd.go
philpennock: The data stored is not a floating point, it's a scaled integer.  `FSCALE` is the amount it was multiplied by to get an integer.  We need to divide by `FSCALE` when restoring to a float.

server/pse/pse_freebsd_amd64.go
philpennock: Okay.  I considered that as an optimization.  Will do.

server/pse/pse_freebsd.go
philpennock: FYI, this same calculation but in the Golang amd64 version, `f_pcpu := float64(r_pcpu) / float64(2048)`, yielded "reasonable" values for me in testing just now.

While running `nats bench foo` in one tty, `nats server request variables` showed `.data.cpu` rise up as high as `0.24658203125`.

philpennock:
Latest push was just because the FreeBSD box where I wrote/tested the fixes didn't have git `user.email` set to my work address.  After fixing that, I ran `git rebase master --exec 'git commit --amend --reset-author --no-edit'`, reviewed with `git log` and then did `git push --force-with-lease` -- there's no change to code at all, only to the author of the first commit.

kozlovic (approved):
LGTM. Verified with following command:
```
goreleaser --skip-publish --skip-validate --rm-dist -p 8
```
that executable for freebsd OS is correctly built and only the amd64 arch.

PR #1738, opened 2020-11-23
[FIXED] LeafNode reject duplicate remote

### LINKED ISSUE #1730: leaf node eviction of old connections seems racy (for mis configured setups)
## Defect

Leaf node configuration where server A accepts incomming leaf nodes connections on $G.
Server B has two remotes to server A. This is a misconfiguration but:
Server A only sometimes disconnects server B. (there is a race. initial startup usually shows no issue) 
Server B is configured, but none of it's output even hints at an issue.

The picture shows the config and server A and B not having an issue. That's the race condition in A.

Screen Shot 2020-11-20 at 12 39 52 PM

kozlovic:
There was a test to prevent an errorneous loop detection when a
remote would reconnect (due to a stale connection) while the accepting
side did not detect the bad connection yet.

However, this test was racy because the test was done prior to add
the connections to the map.

In the case of a misconfiguration where the remote creates 2 different
remote connections that end-up binding to the same account in the
accepting side, then it was possible that this would not be detected.
And when it was, the remote side would be unaware since the disconnect/
reconnect attempts would not show up if not running in debug mode.

This change makes sure that the detection is no longer racy and returns
an error to the remote so at least the log/console of the remote will
show the "duplicate connection" error messages.

Resolves #1730

### DIFF
--- a/server/client.go
+++ b/server/client.go
@@ -177,6 +177,7 @@ const (
 	MsgHeaderViolation
 	NoRespondersRequiresHeaders
 	ClusterNameConflict
+	DuplicateRemoteLeafnodeConnection
 )
 
 // Some flags passed to processMsgResults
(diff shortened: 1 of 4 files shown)


derekcollison (approved):
LGTM

matthiashanel (approved):
LGTM

PR #1739, opened 2020-11-23
[Added] account name checks for leaf nodes in operator mode

matthiashanel:
Rules out implausible ones

### DIFF
--- a/server/leafnode.go
+++ b/server/leafnode.go
@@ -127,22 +127,19 @@ func validateLeafNode(o *Options) error {
 	}
 	// In local config mode, check that leafnode configuration refers to accounts that exist.
 	if len(o.TrustedOperators) == 0 {
+		accNames := map[string]struct{}{}
+		for _, a := range o.Accounts {
+			accNames[a.Name] = struct{}{}
+		}
+		// global account is always created
+		accNames[DEFAULT_GLOBAL_ACCOUNT] = struct{}{}
+		accNames[_EMPTY_] = struct{}{} // in the context of leaf nodes, empty account means global account
+		// system account either exists or, if not disabled, will be created
+		if o.SystemAccount == _EMPTY_ && !o.NoSystemAccount {
+			accNames[DEFAULT_SYSTEM_ACCOUNT] = struct{}{}
+		}
 		checkAccountExists := func(accName string, cfgType string) error {
-			found := false
-			if accName == "$G" || accName == _EMPTY_ {
-				found = true
-			} else {
-				for _, a := range o.Accounts {
-					if a.Name == accName {
-						found = true
-						break
-					}
-				}
-			}
-			if !found {
-				if accName == DEFAULT_SYSTEM_ACCOUNT && o.SystemAccount == _EMPTY_ && !o.NoSystemAccount {
-					return nil
-				}
+			if _, ok := accNames[accName]; !ok {
 				return fmt.Errorf("cannot find local account %q specified in leafnode %s", accName, cfgType)
 			}
 			return nil
@@ -151,7 +148,7 @@ func validateLeafNode(o *Options) error {
 			return err
 		}
 		for _, lu := range o.LeafNode.Users {
-			if lu.Account == nil {
+			if lu.Account == nil { // means global account
 				continue
 			}
 			if err := checkAccountExists(lu.Account.Name, "authorization"); err != nil {
--- a/server/leafnode.go
+++ b/server/leafnode.go
@@ -133,7 +133,8 @@ func validateLeafNode(o *Options) error {
 		}
 		// global account is always created
 		accNames[DEFAULT_GLOBAL_ACCOUNT] = struct{}{}
-		accNames[_EMPTY_] = struct{}{} // in the context of leaf nodes, empty account means global account
+		// in the context of leaf nodes, empty account means global account
+		accNames[_EMPTY_] = struct{}{}
 		// system account either exists or, if not disabled, will be created
 		if o.SystemAccount == _EMPTY_ && !o.NoSystemAccount {
 			accNames[DEFAULT_SYSTEM_ACCOUNT] = struct{}{}
--- a/server/leafnode.go
+++ b/server/leafnode.go
@@ -125,6 +125,58 @@ func validateLeafNode(o *Options) error {
 	if err := validateLeafNodeAuthOptions(o); err != nil {
 		return err
 	}
+	// In local config mode, check that leafnode configuration refers to accounts that exist.
+	if len(o.TrustedOperators) == 0 {
+		checkAccountExists := func(accName string, cfgType string) error {
+			found := false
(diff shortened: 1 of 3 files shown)


server/server.go
```diff
@@ -451,6 +451,18 @@ func NewServer(opts *Options) (*Server, error) {
 				return nil, fmt.Errorf("no local account %q for remote leafnode", r.LocalAccount)
 			}
 		}
+	} else {
+		if len(opts.LeafNode.Users) != 0 {
```
kozlovic, line 455: Would that be possible to move those checks in `validateLeafNode()` which checks for options?

server/leafnode_test.go
```diff
@@ -1883,3 +1883,45 @@ func TestLeafNodeNoDuplicateWithinCluster(t *testing.T) {
 		}
 	}
 }
+
+func TestLeafNodeOperatorBadCfg(t *testing.T) {
+	dir := os.TempDir()
```
kozlovic, line 1888: Needs something like this instead:
```
tmpDir, err := ioutil.TempDir("", "_nats-server")
if err != nil {
    t.Fatal("Could not create tmp dir")
}
defer os.RemoveAll(tmpDir)
```
right now you are making this test remove the OS temp dir, which explains the weird failures you got.

kozlovic (changes requested):

server/server.go
matthiashanel: absolutely, I placed them here because of the non operator mode checks. Ok to move all those checks?

kozlovic: That being said, I am not sure since we call this in NewServer()
```
if err := validateOptions(opts); err != nil {
    return nil, err
}
```
but before we actually process the trusted keys:
```
	// Trusted root operator keys.
	if !s.processTrustedKeys() {
		return nil, fmt.Errorf("Error processing trusted operator keys")
	}
```
would have to make sure that the validation of leafnode would still work if you move them there.

server/leafnode_test.go
matthiashanel: fixed

server/leafnode.go
```diff
@@ -125,6 +125,58 @@ func validateLeafNode(o *Options) error {
 	if err := validateLeafNodeAuthOptions(o); err != nil {
 		return err
 	}
+	// In local config mode, check that leafnode configuration refers to accounts that exist.
+	if len(o.TrustedOperators) == 0 {
```
kozlovic, line 129: I was referring to moving the `else` part that you introduced, not the whole thing. I see that you had to replace the s.accounts.Load() to iterate through the options account.. and will have to do that for LN's account, possibly every LN user, every remote's local account.

server/leafnode.go
```diff
@@ -125,6 +125,58 @@ func validateLeafNode(o *Options) error {
...
+		checkAccountExists := func(accName string, cfgType string) error {
+			found := false
+			if accName == "$G" || accName == _EMPTY_ {
```
kozlovic, line 132: If you keep this here, and I am not convinced, honestly, there is no argument in favor of a `if/else` here:
```
if accName == globalAccountName || accName == _EMPTY_ {
    return nil
}
var found bool
for _, a := range o.Accounts {
...
```

kozlovic (changes requested):

server/leafnode.go
matthiashanel: that's what I meant by move all checks. move it back?

kozlovic: Sorry for the misunderstanding. Yes, I would feel more comfortable leaving the non operator mode checks where they were. You can have the ones specific to operator mode in validate options. But even then, if you feel that there may be some initialization not done at the time we call validate, then leave the check where you originally had them.
Tthe one I have in mind is this:
```
	if !s.processTrustedKeys() {
		return nil, fmt.Errorf("Error processing trusted operator keys")
	}

	s.mu.Lock()
	defer s.mu.Unlock()
        ...
```
We call validate before call to s.processTrustedKeys(). Maybe we can move call to validateOptions() after that call?

matthiashanel: I'll create a map of account names on the fly, shouldn't be to bad, considering we build it up later as well

kozlovic: But if you move it back, you don't need that since you can access s.accounts.Load().. Basically undo your changes and just move the new checks into validate()

matthiashanel: the advantage would be that validation code stays together, wouldn't you agree?

kozlovic: I would agree... but just want to make sure that we are not missing things that may be updated somewhere during the NewServer() init and before the original test was located.

matthiashanel: in this part we only check if accounts are configured, so far we generate only $G and $SYS

server/leafnode.go
matthiashanel: got simplified by use of a map

server/leafnode.go
```diff
@@ -125,6 +125,55 @@ func validateLeafNode(o *Options) error {
...
 	}
+	// In local config mode, check that leafnode configuration refers to accounts that exist.
+	if len(o.TrustedOperators) == 0 {
+		accNames := map[string]struct{}{}
+		for _, a := range o.Accounts {
+			accNames[a.Name] = struct{}{}
+		}
+		// global account is always created
+		accNames[DEFAULT_GLOBAL_ACCOUNT] = struct{}{}
+		accNames[_EMPTY_] = struct{}{} // in the context of leaf nodes, empty account means global account
```
kozlovic, line 136: Please move the comment before the `accNames[_EMPTY_] = ...` line.

server/leafnode.go
```diff
@@ -125,6 +125,55 @@ func validateLeafNode(o *Options) error {
...
+		// system account either exists or, if not disabled, will be created
+		if o.SystemAccount == _EMPTY_ && !o.NoSystemAccount {
+			accNames[DEFAULT_SYSTEM_ACCOUNT] = struct{}{}
+		}
+		checkAccountExists := func(accName string, cfgType string) error {
+			if _, ok := accNames[accName]; !ok {
+				return fmt.Errorf("cannot find local account %q specified in leafnode %s", accName, cfgType)
```
kozlovic, line 143: Maybe remove `local` here since this message now becomes common to local and non local accounts.

kozlovic (changes requested):
Almost there :-)

server/leafnode.go
matthiashanel: it's always a local account we are looking for. I use cfgType to make the distinction between remotes and authorization

server/leafnode.go
matthiashanel: fixed

server/leafnode.go
kozlovic: Ok, my mistake then.  I was looking at original code..

kozlovic (approved):
LGTM

PR #1741, opened 2020-11-23
fixed bad issuer check performed against account issuer instead account subject (Name)

### LINKED ISSUE #1740: user jwts that set account issuer to be same as the account public key fail authentication
an user JWT which sets "issuer_account" to be the same as the account id, gets rejected.
The server has a check for that tests signing keys and the issuer to match - however the public key for the account is set on the 'Name` field, and so the test fails rejecting the user.

aricart:
fixed check of user issuer against the account id (Name) - was set to Issuer instead (issuer is always an operator)
added tests

FIX #1740

/cc @nats-io/core

### DIFF
--- a/server/accounts.go
+++ b/server/accounts.go
@@ -2692,7 +2692,7 @@ func (a *Account) hasIssuer(issuer string) bool {
 
 // hasIssuerNoLock is the unlocked version of hasIssuer
 func (a *Account) hasIssuerNoLock(issuer string) bool {
-	// same issuer
+	// same issuer -- keep this for safety on the calling code
 	if a.Name == issuer {
 		return true
 	}
--- a/server/accounts.go
+++ b/server/accounts.go
@@ -2693,7 +2693,7 @@ func (a *Account) hasIssuer(issuer string) bool {
 // hasIssuerNoLock is the unlocked version of hasIssuer
 func (a *Account) hasIssuerNoLock(issuer string) bool {
 	// same issuer
-	if a.Issuer == issuer {
+	if a.Name == issuer {
 		return true
 	}
 	for i := 0; i < len(a.signingKeys); i++ {
--- a/server/auth.go
+++ b/server/auth.go
@@ -525,6 +525,7 @@ func (s *Server) processClientOrLeafAuthentication(c *client, opts *Options) boo
 			c.Debugf("Account JWT not signed by trusted operator")
 			return false
 		}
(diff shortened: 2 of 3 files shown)


kozlovic:
@aricart Asking @matthiashanel instead since he is more familiar with that part of the code than I am.

server/auth.go
```diff
@@ -525,6 +525,7 @@ func (s *Server) processClientOrLeafAuthentication(c *client, opts *Options) boo
 			c.Debugf("Account JWT not signed by trusted operator")
 			return false
 		}
+		// this only executes IF there's an issuer on the Juc - otherwise the account is already
```
matthiashanel, line 528: already what?

I know what, but seems like smth. is missing in this sentence

server/accounts.go
```diff
@@ -2693,7 +2693,7 @@ func (a *Account) hasIssuer(issuer string) bool {
 // hasIssuerNoLock is the unlocked version of hasIssuer
 func (a *Account) hasIssuerNoLock(issuer string) bool {
 	// same issuer
```
matthiashanel, line 2695: change the name of the function to match its functionality?
account IsIssuing?

server/auth.go
```diff
@@ -525,6 +525,7 @@ func (s *Server) processClientOrLeafAuthentication(c *client, opts *Options) boo
...
 		if juc.IssuerAccount != "" && !acc.hasIssuer(juc.Issuer) {
```
matthiashanel, line 529: why go back to this from what we ended up with?

aricart: because in the end the function that we dispatch is doing that check. Here we are propagating knowledge of an edge condition where the issuer account is set, but it also matches the account name. Let the `hasIssuer()` verification do what it does.

server/accounts.go
aricart: that name is correct - it is checking the issuer of the user JWT. The account has such an issuer key.

server/auth.go
aricart: Fixed.

server/accounts.go
matthiashanel: ok

server/auth.go
matthiashanel: ok

matthiashanel (approved):
LGTM

kozlovic:
Deferring to Matthias, so LGTM.

matthiashanel:
@kozlovic I'm fine with the change

PR #1742, opened 2020-11-24
Servicecyclewithmapping

matthiashanel:
as pointed out int the review to https://github.com/nats-io/nats-server/pull/1731 importFormsCycle does not work and:
cycles are found where there are none
the 3 account test in TestServiceCycle does not fail because a cycle between A - B - C is detected. 
rewriting the imported subject is not taken into account
If we do not allow cycles for services, we shouldn't allow cycles for streams either.

Frankly we shouldn't look at subjects or import types, we should require that no information flow in a cyclic way, thereby requiring accounts to be hierarchical.  
I would certainly consider this a best practice practice as it avoids common errors (see lock hierarchies).
Anything tightly coupled should end up in the same account.

This should be detected as cycle. 
```
accounts {
A {
exports [ { stream: strm } ]
imports [ { service { subject: srv, account: B } } ]
}
B {
exports [ { service: srv } ]
imports [ { stream { subject: strm, account: A } } ]
}
}
```
Having the assertion that there are no reference cycles between accounts is a handy property to have and would simplify other tasks like account cleanup as well.

Taking subjects and types into account or not, cycle detection can't be done by just looking at two nodes in a graph.

Here is a rough draft:
A list of nodes to visit (to implement depth or breadth first search) is needed as well as a map[*Account]strict{} (to mark visited Accounts).
As an import is added to account X, put the imported account into the to visit queue and the visited map
begin:
Take a node from the to visit list
for each import in the account, 
     if the imported account is X, then we have a loop, return
     if the imported account is not in the visited map (this also takes care of imports from the same account)
          add the imported account to the visited map
          add the imported account to the to visit list
goto begin 

This should be fine as, the checks are not expensive (especially when subjects are not looked at) and I would not expect many accounts to be involved in this process. Also, imports are not added all too often.

### DIFF
--- a/test/service_latency_test.go
+++ b/test/service_latency_test.go
@@ -1406,14 +1406,17 @@ func TestServiceCycleWithMapping(t *testing.T) {
 func TestServiceNonCycle(t *testing.T) {
 	t.Skip("will fail without an error otherwise")
 	/*  Occasionally this test passes, probably because the order at which accounts are inspected matters
-	> go test -v "-run=TestServiceNonCycle" ./test -count 100 --failfast
+	go test -v "-run=TestServiceNonCycle" ./test -count 100 --failfast
 	=== RUN   TestServiceNonCycle
 	--- PASS: TestServiceNonCycle (0.00s)
+	=== RUN   TestServiceNonCycleChain
+	    TestServiceNonCycleChain: service_latency_test.go:1443: will fail without an error otherwise
+	--- SKIP: TestServiceNonCycleChain (0.00s)
 	=== RUN   TestServiceNonCycle
-	    TestServiceNonCycle: service_latency_test.go:1427: Expected no error but got /tmp/807881498:2:3: Error adding service import "*": service import forms cycle
+	    TestServiceNonCycle: service_latency_test.go:1438: Expected no error but got /var/folders/9h/6g_c9l6n6bb8gp331d_9y0_w0000gn/T/218571535:2:3: Error adding service import "*": service import forms cycle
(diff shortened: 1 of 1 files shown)


derekcollison:
This is really well done and thought out.

derekcollison:
I don't think we can force the hierarchical account strategy though.

derekcollison:
I will try to take a look later today.

matthiashanel:
> I don't think we can force the hierarchical account strategy though.

If we where, we'd be doing our users a favor. Hierarchies keep things from getting too messy and complex.

derekcollison:
They could also prevent normal use cases which is what I am more concerned about.

matthiashanel:
> They could also prevent normal use cases which is what I am more concerned about.

I'd call that bad design. Sometimes a limit is a good thing. 
Programming languages and import cycles between packages (=account) don't go well together.
go does not compile if you have an import cycle. 
When using locks, establishing a hierarchy is well advised.

https://stackoverflow.com/questions/2052579/circular-dependencies (calls them bad design)
https://stackoverflow.com/questions/28256923/import-cycle-not-allowed 
https://en.wikipedia.org/wiki/Circular_dependency#Problems
https://stackabuse.com/python-circular-imports/ (calls them bad design)

This would be perfectly fine:
A import B 
A import C 
B import C

derekcollison:
I have it on my list of TODOs to look more closely at this. Will do so hopefully tomorrow.

Designers of systems definitely would want easy and straightforward, agree. However it could be very sane to have something that I want to import and something I have that you may want to import and they do not collide. I don't think that should be an error.

PR #1747, opened 2020-11-25
JetStream changes.

derekcollison:
Made several changes based on feedback.

1. Made PubAckResponse only optionally include an ApiError and not force an API type.
2. Allow FilterSubject to be set on a consumer config and cleared if it matches the only stream subject.
3. Remove LookupStream by subject, and add in filters for stream names API.

/cc @nats-io/core

### DIFF
--- a/server/consumer.go
+++ b/server/consumer.go
@@ -265,12 +265,21 @@ func (mset *Stream) AddConsumer(config *ConsumerConfig) (*Consumer, error) {
 
 	// Make sure any partition subject is also a literal.
 	if config.FilterSubject != "" {
-		// Make sure this is a valid partition of the interest subjects.
-		if !mset.validSubject(config.FilterSubject) {
-			return nil, fmt.Errorf("consumer filter subject is not a valid subset of the interest subjects")
+		// If this is a direct match for the streams only subject clear the filter.
+		mset.mu.RLock()
+		if len(mset.config.Subjects) == 1 && mset.config.Subjects[0] == config.FilterSubject {
+			config.FilterSubject = _EMPTY_
 		}
-		if config.AckPolicy == AckAll {
-			return nil, fmt.Errorf("consumer with filter subject can not have an ack policy of ack all")
+		mset.mu.RUnlock()
+
+		if config.FilterSubject != "" {
+			// Make sure this is a valid partition of the interest subjects.
+			if !mset.validSubject(config.FilterSubject) {
+				return nil, fmt.Errorf("consumer filter subject is not a valid subset of the interest subjects")
+			}
+			if config.AckPolicy == AckAll {
+				return nil, fmt.Errorf("consumer with filter subject can not have an ack policy of ack all")
+			}
 		}
 	}
 
--- a/server/jetstream_api.go
+++ b/server/jetstream_api.go
@@ -75,10 +75,6 @@ const (
 	JSApiStreamInfo  = "$JS.API.STREAM.INFO.*"
 	JSApiStreamInfoT = "$JS.API.STREAM.INFO.%s"
 
-	// JSApiStreamLookup is for obtaining a stream by a target subject.
-	// Will return JSON response.
-	JSApiStreamLookup = "$JS.API.STREAM.LOOKUP"
-
 	// JSApiStreamDelete is the endpoint to delete streams.
 	// Will return JSON response.
 	JSApiStreamDelete  = "$JS.API.STREAM.DELETE.*"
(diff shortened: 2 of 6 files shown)


server/consumer.go
```diff
@@ -265,12 +265,21 @@ func (mset *Stream) AddConsumer(config *ConsumerConfig) (*Consumer, error) {
 
 	// Make sure any partition subject is also a literal.
 	if config.FilterSubject != "" {
-		// Make sure this is a valid partition of the interest subjects.
-		if !mset.validSubject(config.FilterSubject) {
-			return nil, fmt.Errorf("consumer filter subject is not a valid subset of the interest subjects")
+		// If this is a direct match for the streams only subject clear the filter.
+		mset.mu.RLock()
+		if len(mset.config.Subjects) == 1 && mset.config.Subjects[0] == config.FilterSubject {
+			config.FilterSubject = _EMPTY_
```
ripienaar, line 271: nice, discovered this annoyance too when I did a subject based client POC

server/consumer.go
```diff
@@ -265,12 +265,21 @@ func (mset *Stream) AddConsumer(config *ConsumerConfig) (*Consumer, error) {
...
 		}
-		if config.AckPolicy == AckAll {
-			return nil, fmt.Errorf("consumer with filter subject can not have an ack policy of ack all")
+		mset.mu.RUnlock()
+
+		if config.FilterSubject != "" {
+			// Make sure this is a valid partition of the interest subjects.
+			if !mset.validSubject(config.FilterSubject) {
```
ripienaar, line 277: Do we do a similar check on existing consumers if I update a stream config to drop a subject that the consumer is filtered on? What happens to those consumers then?

server/consumer.go
```diff
@@ -477,7 +486,7 @@ func (mset *Stream) AddConsumer(config *ConsumerConfig) (*Consumer, error) {
 		}
 		// If we are here we have already registered this durable. If it is still active that is an error.
 		if eo.Active() {
-			return nil, fmt.Errorf("consumer already exists and is still active")
+			return nil, fmt.Errorf("consumer already 1exists and is still active")
```
ripienaar, line 489: typo

server/jetstream_api.go
```diff
@@ -266,27 +257,20 @@ type JSApiStreamInfoResponse struct {
...
-const JSApiStreamLookupResponseType = "io.nats.jetstream.api.v1.stream_lookup_response"
-
 // Maximum entries we will return for streams or consumers lists.
 // TODO(dlc) - with header or request support could request chunked response.
 const JSApiNamesLimit = 1024
 const JSApiListLimit = 256
 
 type JSApiStreamNamesRequest struct {
 	ApiPagedRequest
+	Filters *Filters `json:"filters,omitempty"`
```
ripienaar, line 267: why make this a extra struct? You were complaining about how many structs we have :) this will be hit infrequently so some empty strings with omitempty wouldnt be a big overhead vs making everyone do this extra struct?

ripienaar (changes requested):

server/consumer.go
derekcollison: Was fat fingering 'goto line 1' in emacs alot last night :/

server/jetstream_api.go
derekcollison: I figured we would have more (I think I put that comment in). Otherwise it becomes

FilterSubject
FilterHeaders
Filer...

Its not too bad IMO.

```go
req, _ := json.Marshal(&server.JSApiStreamNamesRequest{Filters: &server.Filters{Subject: filter}})
```

derekcollison: I can change it though if we feel strongly.

server/consumer.go
derekcollison: Added to the test to make sure.

server/jetstream_api.go
ripienaar: doesnt matter enough to change, just wanted to know your reasoning, I imagine probably the only keys we'd add here is filter related, but worth being cautious so we can leave as is

ripienaar (approved):

ripienaar (approved):
LGTM

server/jetstream_api.go
derekcollison: I flattened.

server/consumer.go
kozlovic: But do we plan maybe in the future to allow the modification of a stream as to be able to add a non overlapping subject to it? Say stream "STREAM" captures "foo" at the moment, then later the user decides to add "bar" to this stream. If we were to want to allow that in the future, would this change here not prevent that feature from ever existing?

derekcollison: If decide to allow that we can update to correctly work but I do not see us doing that anytime soon.

PR #1748, opened 2020-11-25
Fixed bug restoring consumer state

derekcollison:
This was also the flapper test we had for **TestJetStreamConsumerMaxDeliveryAndServerRestart**.
    
We were not properly restoring our state for consumers and we also had a bug where we would not properly encode and write redelivered state.
    
/cc @nats-io/core


kozlovic (approved):
LGTM

PR #1749, opened 2020-11-25
[FIXED] System account was not properly tracking GW routed replies

kozlovic:
In some cases, the reply of a request message is prefixed when
going over a gateway so that if it comes back to a different
server than when the request originates, it can be routed back.

For system accounts, this routed reply subject was not tracked
so the server would reply to the inbox and may reach a server
that had not yet processed (through the route) the interest
on that inbox. If the reply came with the GW routed info, that
server would know to route it to the original server.


derekcollison (approved):
LGTM

PR #1750, opened 2020-11-26
ensure the stream originating a pub error is reported

ripienaar:

/cc @nats-io/core

### DIFF
--- a/server/stream.go
+++ b/server/stream.go
@@ -940,6 +940,7 @@ func (mset *Stream) processInboundJetStreamMsg(_ *subscription, pc *client, subj
 		if sname := getExpectedStream(hdr); sname != _EMPTY_ && sname != name {
 			mset.mu.Unlock()
 			if doAck && len(reply) > 0 {
+				resp.PubAck = &PubAck{Stream: name}
 				resp.Error = &ApiError{Code: 400, Description: "expected stream does not match"}
 				b, _ := json.Marshal(resp)
 				sendq <- &jsPubMsg{reply, _EMPTY_, _EMPTY_, nil, b, nil, 0}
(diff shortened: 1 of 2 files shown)


ripienaar:
I didnt change the whole structure instead just make sure that the variable used to return errors have the stream name set for additional details

server/stream.go
```diff
@@ -919,7 +919,7 @@ func (mset *Stream) processInboundJetStreamMsg(_ *subscription, pc *client, subj
 	numConsumers := len(mset.consumers)
 	interestRetention := mset.config.Retention == InterestPolicy
 
-	var resp = &JSPubAckResponse{}
+	var resp = &JSPubAckResponse{PubAck: &PubAck{Stream: name}}
```
derekcollison, line 922: I wonder if we should only attach this PubAck inside the code where detect an error? Avoid creating and throwing away in normal case.

derekcollison (approved):
LGTM, one minor comment but can test perf hit and escape analysis later.

server/stream.go
ripienaar: Yeah, can do, did think this is a bit of a waste here, but wanted to avoid accidentally forgetting it on one.  Will change round

ripienaar: instead of allocating `resp` each time and then discarding it in the happy path:

```go
resp.PubAck = &PubAck{Stream: name}
resp.Error = &ApiError{Code: 400, Description: err.Error()}
b, _ = json.Marshal(resp)
sendq <- &jsPubMsg{reply, _EMPTY_, _EMPTY_, nil, b, nil, 0}
```

I was thinking maybe this?

```go
errorResp = func(c int, d string) []byte {
  resp := &JSPubAckResponse{
    PubAck: &PubAck{Stream: name},
    Error: &ApiError{Code: c, Description: d}
  }

  b, _ := json.Marshal(resp)
  return b
}

.
.
.

sendq <- &jsPubMsg{reply, _EMPTY_, _EMPTY_, nil, errorResp(400, "foo"), nil, 0}
```

derekcollison: Would not make it a closure inside the processInboundStream message but if external func for errors yes.

derekcollison: The cost of setting up the closure function may be worse then the original allocation strategy ;)

ripienaar: Hehe could be not sure how clever the compiler is :) anyway will just set this in every call and hope we don’t make a mistake

derekcollison: Sounds good.

derekcollison (approved):
LGTM

PR #1753, opened 2020-11-28
Remove limitation on AckAll for consumers with filtered subjects.

derekcollison:

/cc @nats-io/core


ripienaar (approved):
LGTM

PR #1754, opened 2020-11-30
[ADDED] MQTT Support

kozlovic:
This PR introduces native support for MQTT clients. It requires use
of accounts with JetStream enabled. Since as of now clustering is
not available, MQTT will be limited to single instance.

Only QoS 0 and 1 are supported at the moment. MQTT clients can
exchange messages with NATS clients and vice-versa.

Since JetStream is required, accounts with JetStream enabled must
exist in order for an MQTT client to connect to the NATS Server.
The administrator can limit the users that can use MQTT with the
allowed_connection_types option in the user section. For instance:
```
accounts {
  mqtt {
    users [
      {user: all, password: pwd, allowed_connection_types: ["STANDARD", "WEBSOCKET", "MQTT"]}
      {user: mqtt_only, password: pwd, allowed_connection_types: "MQTT"}
    ]
    jetstream: enabled
  }
}
```
The "mqtt_only" can only be used for MQTT connections, which the user
"all" accepts standard, websocket and MQTT clients.

Here is what a configuration to enable MQTT looks like:
```
mqtt {
  # Specify a host and port to listen for websocket connections
  #
  # listen: "host:port"

  # It can also be configured with individual parameters,
  # namely host and port.
  #
  # host: "hostname"
  port: 1883

  # TLS configuration section
  #
  # tls {
  #  cert_file: "/path/to/cert.pem"
  #  key_file: "/path/to/key.pem"
  #  ca_file: "/path/to/ca.pem"
  #
  #  # Time allowed for the TLS handshake to complete
  #  timeout: 2.0
  #
  #  # Takes the user name from the certificate
  #  #
  #  # verify_an_map: true
  #}

  # Authentication override. Here are possible options.
  #
  # authorization {
  #   # Simple username/password
  #   #
  #   user: "some_user_name"
  #   password: "some_password"
  #
  #   # Token. The server will check the MQTT's password in the connect
  #   # protocol against this token.
  #   #
  #   # token: "some_token"
  #
  #   # Time allowed for the client to send the MQTT connect protocol
  #   # after the TCP connection is established.
  #   #
  #   timeout: 2.0
  #}

  # If an MQTT client connects and does not provide a username/password and
  # this option is set, the server will use this client (and therefore account).
  #
  # no_auth_user: "some_user_name"

  # This is the time after which the server will redeliver a QoS 1 message
  # sent to a subscription that has not acknowledged (PUBACK) the message.
  # The default is 30 seconds.
  #
  # ack_wait: "1m"

  # This limits the number of QoS1 messages sent to a session without receiving
  # acknowledgement (PUBACK) from that session. MQTT specification defines
  # a packet identifier as an unsigned int 16, which means that the maximum
  # value is 65535. The default value is 1024.
  #
  # max_ack_pending: 100
}
```

### DIFF
--- a/server/auth.go
+++ b/server/auth.go
@@ -253,6 +253,8 @@ func (s *Server) configureAuthorization() {
 
 	// Do similar for websocket config
 	s.wsConfigAuth(&opts.Websocket)
+	// And for mqtt config
+	s.mqttConfigAuth(&opts.MQTT)
 }
 
 // Takes the given slices of NkeyUser and User options and build
@@ -343,11 +345,15 @@ func (s *Server) processClientOrLeafAuthentication(c *client, opts *Options) boo
 	)
 	s.mu.Lock()
 	authRequired := s.info.AuthRequired
-	// c.ws is immutable, but may need lock if we get race reports.
-	if !authRequired && c.ws != nil {
-		// If no auth required for regular clients, then check if
-		// we have an override for websocket clients.
-		authRequired = s.websocket.authOverride
+	// c.ws/mqtt is immutable, but may need lock if we get race reports.
+	if !authRequired {
+		if c.mqtt != nil {
+			authRequired = s.mqtt.authOverride
+		} else if c.ws != nil {
+			// If no auth required for regular clients, then check if
+			// we have an override for websocket clients.
+			authRequired = s.websocket.authOverride
+		}
 	}
 	if !authRequired {
 		// TODO(dlc) - If they send us credentials should we fail?
@@ -361,7 +367,20 @@ func (s *Server) processClientOrLeafAuthentication(c *client, opts *Options) boo
 		noAuthUser string
 	)
 	tlsMap := opts.TLSMap
-	if c.ws != nil {
+	if c.mqtt != nil {
+		mo := &opts.MQTT
+		// Always override TLSMap.
+		tlsMap = mo.TLSMap
+		// The rest depends on if there was any auth override in
+		// the mqtt's config.
+		if s.mqtt.authOverride {
+			noAuthUser = mo.NoAuthUser
+			username = mo.Username
+			password = mo.Password
+			token = mo.Token
+			ao = true
+		}
+	} else if c.ws != nil {
 		wo := &opts.Websocket
 		// Always override TLSMap.
 		tlsMap = wo.TLSMap
@@ -998,7 +1017,7 @@ func validateAllowedConnectionTypes(m map[string]struct{}) error {
 	for ct := range m {
 		ctuc := strings.ToUpper(ct)
 		switch ctuc {
-		case jwt.ConnectionTypeStandard, jwt.ConnectionTypeWebsocket, jwt.ConnectionTypeLeafnode:
+		case jwt.ConnectionTypeStandard, jwt.ConnectionTypeWebsocket, jwt.ConnectionTypeLeafnode, jwt.ConnectionTypeMqtt:
 		default:
 			return fmt.Errorf("unknown connection type %q", ct)
 		}
--- a/server/auth.go
+++ b/server/auth.go
@@ -346,11 +346,12 @@ func (s *Server) processClientOrLeafAuthentication(c *client, opts *Options) boo
 	s.mu.Lock()
 	authRequired := s.info.AuthRequired
 	if !authRequired {
-		if c.isMqtt() {
+		// If no auth required for regular clients, then check if
+		// we have an override for MQTT or Websocket clients.
+		switch c.clientType() {
+		case MQTT:
 			authRequired = s.mqtt.authOverride
-		} else if c.isWebsocket() {
-			// If no auth required for regular clients, then check if
-			// we have an override for websocket clients.
+		case WS:
 			authRequired = s.websocket.authOverride
 		}
 	}
@@ -366,33 +367,36 @@ func (s *Server) processClientOrLeafAuthentication(c *client, opts *Options) boo
 		noAuthUser string
 	)
 	tlsMap := opts.TLSMap
-	if c.isMqtt() {
-		mo := &opts.MQTT
-		// Always override TLSMap.
-		tlsMap = mo.TLSMap
-		// The rest depends on if there was any auth override in
-		// the mqtt's config.
-		if s.mqtt.authOverride {
-			noAuthUser = mo.NoAuthUser
-			username = mo.Username
-			password = mo.Password
-			token = mo.Token
-			ao = true
+	if c.kind == CLIENT {
+		switch c.clientType() {
+		case MQTT:
+			mo := &opts.MQTT
+			// Always override TLSMap.
+			tlsMap = mo.TLSMap
+			// The rest depends on if there was any auth override in
+			// the mqtt's config.
+			if s.mqtt.authOverride {
+				noAuthUser = mo.NoAuthUser
+				username = mo.Username
+				password = mo.Password
+				token = mo.Token
+				ao = true
+			}
+		case WS:
+			wo := &opts.Websocket
+			// Always override TLSMap.
+			tlsMap = wo.TLSMap
+			// The rest depends on if there was any auth override in
+			// the websocket's config.
+			if s.websocket.authOverride {
+				noAuthUser = wo.NoAuthUser
+				username = wo.Username
+				password = wo.Password
+				token = wo.Token
+				ao = true
+			}
 		}
-	} else if c.isWebsocket() {
-		wo := &opts.Websocket
-		// Always override TLSMap.
-		tlsMap = wo.TLSMap
-		// The rest depends on if there was any auth override in
-		// the websocket's config.
-		if s.websocket.authOverride {
-			noAuthUser = wo.NoAuthUser
-			username = wo.Username
-			password = wo.Password
-			token = wo.Token
-			ao = true
-		}
-	} else if c.kind == LEAF {
+	} else {
 		tlsMap = opts.LeafNode.TLSMap
 	}
 	if !ao {
--- a/server/auth.go
+++ b/server/auth.go
@@ -345,11 +345,10 @@ func (s *Server) processClientOrLeafAuthentication(c *client, opts *Options) boo
 	)
 	s.mu.Lock()
 	authRequired := s.info.AuthRequired
-	// c.ws/mqtt is immutable, but may need lock if we get race reports.
 	if !authRequired {
-		if c.mqtt != nil {
+		if c.isMqtt() {
 			authRequired = s.mqtt.authOverride
-		} else if c.ws != nil {
+		} else if c.isWebsocket() {
 			// If no auth required for regular clients, then check if
 			// we have an override for websocket clients.
 			authRequired = s.websocket.authOverride
@@ -367,7 +366,7 @@ func (s *Server) processClientOrLeafAuthentication(c *client, opts *Options) boo
 		noAuthUser string
 	)
 	tlsMap := opts.TLSMap
-	if c.mqtt != nil {
+	if c.isMqtt() {
 		mo := &opts.MQTT
 		// Always override TLSMap.
 		tlsMap = mo.TLSMap
@@ -380,7 +379,7 @@ func (s *Server) processClientOrLeafAuthentication(c *client, opts *Options) boo
 			token = mo.Token
 			ao = true
 		}
-	} else if c.ws != nil {
+	} else if c.isWebsocket() {
 		wo := &opts.Websocket
 		// Always override TLSMap.
 		tlsMap = wo.TLSMap
--- a/server/client.go
+++ b/server/client.go
@@ -2991,7 +2991,12 @@ func (c *client) deliverMsg(sub *subscription, subject, reply, mh, msg []byte, g
 	// Update statistics
 
 	// The msg includes the CR_LF, so pull back out for accounting.
-	msgSize := int64(len(msg) - LEN_CR_LF)
+	msgSize := int64(len(msg))
+	prodIsMQTT := c.isMqtt()
+	// MQTT producers send messages without CR_LF, so don't remove it for them.
+	if !prodIsMQTT {
+		msgSize -= int64(LEN_CR_LF)
+	}
 
 	// No atomic needed since accessed under client lock.
 	// Monitor is reading those also under client's lock.
@@ -3066,6 +3071,10 @@ func (c *client) deliverMsg(sub *subscription, subject, reply, mh, msg []byte, g
 	// Queue to outbound buffer
 	client.queueOutbound(mh)
 	client.queueOutbound(msg)
+	if prodIsMQTT {
+		// Need to add CR_LF since MQTT producers don't send CR_LF
+		client.queueOutbound([]byte(CR_LF))
+	}
 
 	client.out.pm++
 
--- a/server/client.go
+++ b/server/client.go
@@ -51,6 +51,20 @@ const (
 	ACCOUNT
 )
 
+// Extended type of a CLIENT connection. This is returned by c.clientType()
+// and indicate what type of client connection we are dealing with.
+// If invoked on a non CLIENT connection, NON_CLIENT type is returned.
+const (
+	// If the connection is not a CLIENT connection.
+	NON_CLIENT = iota
+	// Regular NATS client.
+	NATS
+	// MQTT client.
+	MQTT
+	// Websocket client.
+	WS
+)
+
 const (
 	// ClientProtoZero is the original Client protocol from 2009.
 	// http://nats.io/documentation/internals/nats-protocol/
@@ -427,6 +441,26 @@ func (c *client) GetTLSConnectionState() *tls.ConnectionState {
 	return &state
 }
 
+// For CLIENT connections, this function returns the client type, that is,
+// NATS (for regular clients), MQTT or WS for websocket.
+// If this is invoked for a non CLIENT connection, NON_CLIENT is returned.
+//
+// This function does not lock the client and accesses fields that are supposed
+// to be immutable and therefore it can be invoked outside of the client's lock.
+func (c *client) clientType() int {
+	switch c.kind {
+	case CLIENT:
+		if c.isMqtt() {
+			return MQTT
+		} else if c.isWebsocket() {
+			return WS
+		}
+		return NATS
+	default:
+		return NON_CLIENT
+	}
+}
+
 // This is the main subscription struct that indicates
 // interest in published messages.
 // FIXME(dlc) - This is getting bloated for normal subs, need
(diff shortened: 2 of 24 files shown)


kozlovic:
@derekcollison Obviously too much to review, but if you could focus on areas that existed before but have been modified for MQTT. Namely consumer.go, jetstream.go. I also needed a reverse-match for sublist, so also have a look at that.

derekcollison:
Will take a look.

derekcollison:
I can't reproduce from head of main branch either.

server/consumer.go
```diff
@@ -2177,9 +2184,22 @@ func (mset *Stream) deliveryFormsCycle(deliverySubject string) bool {
 	return false
 }
 
-// This is same as check for delivery cycle.
+// Check that the subject is a subset of the stream's configured subjects,
+// or returns true if the stream has been created with no subject.
 func (mset *Stream) validSubject(partitionSubject string) bool {
-	return mset.deliveryFormsCycle(partitionSubject)
```
derekcollison, line 2182: This used anymore? If so move to `RLock` as well.

server/jetstream.go
```diff
@@ -521,7 +521,12 @@ func (a *Account) EnableJetStream(limits *JetStreamAccountLimits) error {
 				s.Warnf("  Error adding Stream %q to Template %q: %v", cfg.Name, cfg.Template, err)
 			}
 		}
-		mset, err := a.AddStream(&cfg.StreamConfig)
+		// TODO: We should not rely on the stream name.
+		// However, having a StreamConfig property, such as AllowNoSubject,
```
derekcollison, line 525: Could be a non-public field. `allowNoSubject`

server/opts.go
```diff
@@ -316,6 +317,49 @@ type WebsocketOpts struct {
...
+	NoAuthUser string
+
+	// Authentication section. If anything is configured in this section,
+	// it will override the authorization configuration of regular clients.
+	Username string
+	Password string
+	Token    string
+
+	// Timeout for the authentication process.
+	AuthTimeout float64
```
derekcollison, line 338: This b/c of mqtt protocol? vs `time.Duration`.

server/opts.go
```diff
@@ -3831,6 +3953,14 @@ func setBaselineOptions(opts *Options) {
 			opts.Websocket.Host = DEFAULT_HOST
 		}
 	}
+	if opts.MQTT.Port != 0 {
```
derekcollison, line 3956: Should we use default that folks expect here by default? For port I mean..

server/server.go
```diff
@@ -1516,6 +1522,11 @@ func (s *Server) Start() {
 		s.startWebsocketServer()
 	}
 
+	// MQTT
+	if opts.MQTT.Port != 0 {
```
derekcollison, line 1526: Be good if we could configure this to be the default easily since its already established.

server/server.go
```diff
@@ -2073,7 +2091,7 @@ func (c *tlsMixConn) Read(b []byte) (int, error) {
 	return c.Conn.Read(b)
 }
 
-func (s *Server) createClient(conn net.Conn, ws *websocket) *client {
+func (s *Server) createClient(conn net.Conn, ws *websocket, mqtt *mqtt) *client {
```
derekcollison, line 2094: Feels like `createClient` should stay the same and have specialized version for ws and mqtt vs passing them in.

server/server.go
```diff
@@ -2086,32 +2104,49 @@ func (s *Server) createClient(conn net.Conn, ws *websocket) *client {
 	now := time.Now()
 
 	c := &client{srv: s, nc: conn, opts: defaultOpts, mpay: maxPay, msubs: maxSubs, start: now, last: now, ws: ws}
+	if mqtt != nil {
+		c.mqtt = mqtt
+		// Set some of the options here since MQTT clients don't
+		// send a regular CONNECT (but have their own).
+		c.opts.Lang = "mqtt"
```
derekcollison, line 2111: Wonder if we should leave blank?

server/stream.go
```diff
@@ -125,13 +127,20 @@ func (a *Account) AddStream(config *StreamConfig) (*Stream, error) {
 
 // AddStreamWithStore adds a stream for the given account with custome store config options.
 func (a *Account) AddStreamWithStore(config *StreamConfig, fsConfig *FileStoreConfig) (*Stream, error) {
+	if strings.HasPrefix(config.Name, mqttStreamNamePrefix) {
```
derekcollison, line 130: Wonder if we should add in internal streams etc such that these could flow easier and not reply on names?

derekcollison: Meaning a non-public `internal` bool in stream config. The consumers could inherit that and bypass some of the subject stuff off of that.

server/stream.go
```diff
@@ -125,13 +127,20 @@ func (a *Account) AddStream(config *StreamConfig) (*Stream, error) {
...
+		return nil, fmt.Errorf("prefix %q is reserved for MQTT, unable to create stream %q", mqttStreamNamePrefix, config.Name)
```
derekcollison, line 131: Would just say reserved, not say anything about MQTT per se.

server/stream.go
```diff
@@ -125,13 +127,20 @@ func (a *Account) AddStream(config *StreamConfig) (*Stream, error) {
...
+
+func (a *Account) addStreamWithStore(config *StreamConfig, fsConfig *FileStoreConfig, noSubjectsOK bool) (*Stream, error) {
 	s, jsa, err := a.checkForJetStream()
 	if err != nil {
 		return nil, err
 	}
 
 	// Sensible defaults.
-	cfg, err := checkStreamCfg(config)
+	cfg, err := checkStreamCfg(config, noSubjectsOK)
```
derekcollison, line 143: Same here with internal option.

server/stream.go
```diff
@@ -168,7 +177,7 @@ func (a *Account) AddStreamWithStore(config *StreamConfig, fsConfig *FileStoreCo
 
 	// Setup the internal client.
 	c := s.createInternalJetStreamClient()
-	mset := &Stream{jsa: jsa, config: cfg, client: c, consumers: make(map[string]*Consumer)}
+	mset := &Stream{jsa: jsa, config: cfg, client: c, consumers: make(map[string]*Consumer), nosubj: noSubjectsOK}
```
derekcollison, line 180: Let's move to stream and consumer config as internal fields in config struct.

`interestNotRequired`

server/stream.go
```diff
@@ -691,7 +705,7 @@ func (mset *Stream) subscribeInternal(subject string, cb msgHandler) (*subscript
 	mset.sid++
 
 	// Now create the subscription
-	return c.processSub([]byte(subject), nil, []byte(strconv.Itoa(mset.sid)), cb, false)
+	return c.processSub(c.createSub([]byte(subject), nil, []byte(strconv.Itoa(mset.sid)), cb), false)
```
derekcollison, line 708: Why did you change first arg to `*subscription`?

server/stream.go
```diff
@@ -901,7 +915,7 @@ func getExpectedLastSeq(hdr []byte) uint64 {
 }
 
 // processInboundJetStreamMsg handles processing messages bound for a stream.
-func (mset *Stream) processInboundJetStreamMsg(_ *subscription, pc *client, subject, reply string, msg []byte) {
+func (mset *Stream) processInboundJetStreamMsg(sub *subscription, pc *client, subject, reply string, msg []byte) {
```
derekcollison, line 918: I see you set sub but I don't think you use it.

server/sublist.go
```diff
@@ -1392,3 +1392,65 @@ func (s *Sublist) collectAllSubs(l *level, subs *[]*subscription) {
 		s.collectAllSubs(l.fwc.next, subs)
 	}
 }
+
+func (s *Sublist) ReverseMatch(subject string) *SublistResult {
```
derekcollison, line 1396: Description here be good.

derekcollison (commented):
In general LGTM. Did not look too closely at mqtt code and tests but integration points.

server/client.go
```diff
@@ -2212,32 +2226,30 @@ func (c *client) parseSub(argo []byte, noForward bool) error {
...
 }
 
-func (c *client) processSub(subject, queue, bsid []byte, cb msgHandler, noForward bool) (*subscription, error) {
-	// Create the subscription
-	sub := &subscription{client: c, subject: subject, queue: queue, sid: bsid, icb: cb}
+func (c *client) createSub(subject, queue, sid []byte, cb msgHandler) *subscription {
+	return &subscription{client: c, subject: subject, queue: queue, sid: sid, icb: cb}
+}
+
+func (c *client) processSub(sub *subscription, noForward bool) (*subscription, error) {
```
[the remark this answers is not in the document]
kozlovic: This is because a subscription on "foo" may already exist, so if I call processSub() with a new subscription object but for the same subject/sid, then the existing one is returned (and some of the sub's field are updated - under the client lock. For mqtt it means the qos and other).

server/client.go
```diff
@@ -3983,7 +4003,7 @@ func (c *client) processPingTimer() {
 	c.mu.Lock()
 	c.ping.tmr = nil
 	// Check if connection is still opened
-	if c.isClosed() {
+	if c.isClosed() || c.mqtt != nil {
```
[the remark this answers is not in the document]
kozlovic: c.kind is still CLIENT. As for websocket, mqtt clients are CLIENT connections. They just have extra field to maintain specific state. I can however create an isMqtt() that does the check though.

server/client.go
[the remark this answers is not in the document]
derekcollison: ok then why do you need to build the sub beforehand?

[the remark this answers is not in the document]
kozlovic: This has 2 advantages: reduce the number of parameters that we would have to had to processSub(). Granted we could create a suboption struct that we pass, but this is really equivalent to passing a sub object, even better if happen to not exist so we don't create the suboption object AND the subscription.
The second is that I need the subscription object to "add" to it retained message prior to be added to the sublist and possibly starting to receive incoming messages.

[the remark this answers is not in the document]
derekcollison: What additional params do you need to pass?
Can you not ass in retained message post this without causing issues?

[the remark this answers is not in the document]
kozlovic: There is qos, but not only that, the subscription has a `mqtt` field that holds the JS consumer it is attached to. I want this to be set before I call processSub() because I want to make sure that this is set prior to add the sub to sublist. You can't set things to the sub after processSub() returns, it is simply not safe. Remember that I had to fix races where you would set sub.icb after calling processSub().
For the retained message, I don't think in the spec they say that they must be sent before any incoming messages, although my experience with some mqtt brokers is that they are sent first.

server/consumer.go
kozlovic: Not sure what you mean, but yes, deliveryFormsCycle is still used in this branch. 
The thing is that I want validSubject() to return true if the stream is created with "no subject allowed", but if I were to put the check in deliveryFormsCycle I would have to return "true" (that it forms cycle) which does not make sense to me. This is why I originally I duplicated the code from deliveryFormsCycle but made it about validating the subject.

server/opts.go
kozlovic: No, simply to be consistent with all the auth timeout that we have in all other configs where we have auth block.

server/opts.go
kozlovic: You already mentioned the same for Websocket and my answer will be the same: setting the port is how users will enable MQTT, so I can't set a default port... We could have a "enabled" flag in mqtt{} instead, but that is a departure of what we have been doing (which is fine). For instance, this is how routes, leafnodes, etc.. are enabled, by setting the port, not having an "enabled" option.

server/server.go
kozlovic: No, again, like we do for routes, etc.. this is how the server knows if it should start MQTT, it is by checking if the port is set.

server/server.go
kozlovic: There would be a LOT of code to copy/paste and have to fix issues in 3 different places (we have been fixing issues in that code very recently).

server/server.go
kozlovic: No because some code check that Lang not empty means it is a "client" connection, as opposed to route, etc...

server/stream.go
kozlovic: As I mentioned in the past, I wanted to make this a configuration but if it is not exported, it won't be persisted (json does not marshall non exported fields, right?). If so, on recovery I won't be able to rely on the config and have to check the name anyway. Of course, in some other places, like here, I could use the internal's config option instead of checking name.

server/stream.go
kozlovic: What about the issue with persisting the config? I will experiment to see if a non exported config option works, but I believe that it would not be persisted..

server/stream.go
kozlovic: Left over from some earlier code, will remove.

server/stream.go
kozlovic: Explained earlier.

server/jetstream_api.go
```diff
@@ -952,7 +952,14 @@ func (s *Server) jsStreamInfoRequest(sub *subscription, c *client, subject, repl
 		s.sendAPIResponse(c, subject, reply, string(msg), s.jsonResponse(&resp))
 		return
 	}
-	resp.StreamInfo = &StreamInfo{Created: mset.Created(), State: mset.State(), Config: mset.Config()}
+	config := mset.Config()
+	// MQTT streams are created without subject, but "nats" tooling would then
+	// fail to display them since it uses validation and expect the config's
+	// Subjects to not be empty.
```
ripienaar, line 958: technically the nats utility is wrong, since the server accepts it and it then defaults to stream name. I could allow it but at the time decided it would be better to be explicit

server/client.go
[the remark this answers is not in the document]
derekcollison: ok that makes sense. Maybe we need to change the name now though from `procesSub`

server/consumer.go
derekcollison: Understood, just saying the lock should be an RLock in `deliveryFormsCycle` too. That was my mistake.

server/stream.go
derekcollison: Good point, this just feels weird passing around a simple bool to the internal functions etc.

server/jetstream_api.go
derekcollison: I will need, as part of clustering work, to have streams not visible to the outside and used for internal purposes as well for a WAL.

server/stream.go
kozlovic: I really would want to make those configuration options, and not have to carry over some boolean in the function calls, but we have the problem of persistence. I can overcome that with implementing MarshalJSON/UnmarshalJSON for FileStreamInfo and FileConsumerInfo if you are ok with that approach?

kozlovic:
@derekcollison I have addressed some of the comments, namely using non public stream/consumer config options. I am now going to look at others, like createClient() and if I really need the subscription to be created before calling processSub, etc..

kozlovic:
@derekcollison Good news, I was able to revert changes to processSub() and still maintain code correctness/no race with mqtt subscriptions. So now moving to other remarks such as createClient(), etc..

server/stream.go
derekcollison: Yes I agree.

derekcollison:
Just ping me when you want me to take another look. We are very close IMO..

server/opts.go
kozlovic: Wanted to circle back on that: what do you want me to do there? Do we introduce an "enabled" flag that enables/disables MQTT? I don't see how make use of a default port if we don't have that, no? (again, in the server we normally use Port != 0 to know that we start listening for connections). Or do I leave like this can mark this as resloved?

derekcollison: What do you think of this. We allow variables in the conf system and they need to normally be externally defined, but what if we had builtins for default ports etc?

```conf
listen: 127.0.0.1:$NATS_DEFAULT_PORT
mqtt {
  listen: 127.0.0.1:$MQTT_DEFAULT_PORT
}
```

If the variable is not defined we use our internal version of it, but you can still override it.

kozlovic: Not sure I like it. I was buying the idea that one would not have to specify the port at all and server would use a default, but again would mean that server has to be told that it should start MQTT, so an "enabled" flag of some sort.

Technically, your approach would be difficult since the parser is in its own package and if it does not find the variable it errors out:
```
nats-server: variable reference for 'MQTT_DEFAULT_PORT' on line 11 can not be found
```
In order for parser code to look for those specific variables and then associate default ports it means that the default should be in the `conf` package, not `server`, which may be a bit weird, no?

kozlovic: but as a POC, in parser.go, lookupVariable(), code similar to this would work:
```
	switch varReference {
	case "MQTT_DEFAULT_PORT":
		return int64(1883), true, nil
	}
        return nil, false, nil
```
Would have to make those constants, but again, likely need to be defined in `conf` package...

kozlovic: .. well, looks like parsing of variable works only if standalone. That is this would work:
```
port: $NATS_DEFAULT_PORT
```
but this won't:
```
listen: 0.0.0.0:$NATS_DEFAULT_PORT
```
get error:
```
nats-server: Parse error on line 1: 'Expected a top-level value to end with a new line, comment or EOF, but got '$' instead.'
```

derekcollison: We could make that work. Also, before you call into conf just look for those env vars and if not found set them for the process then call into conf.

server/server.go
kozlovic: Done.

server/server.go
kozlovic: I have removed it.

server/auth.go
```diff
@@ -367,7 +366,7 @@ func (s *Server) processClientOrLeafAuthentication(c *client, opts *Options) boo
 		noAuthUser string
 	)
 	tlsMap := opts.TLSMap
-	if c.mqtt != nil {
+	if c.isMqtt() {
```
derekcollison, line 369: Maybe make more generic?

```go
switch c.clientType() {
  case NATS:
  case MQTT:
  case WS:
}
```

Or something like that?

derekcollison (approved):
LGTM - minor comments.

server/auth.go
kozlovic: Please check commit https://github.com/nats-io/nats-server/pull/1754/commits/2ca0794b48ce10cfa5a14c54ca915595edc1507e when you get a chance. I am not sold on that and I am afraid that it gets mis-used.. meaning you still want to check for c.kind == CLIENT when you deal with routing of messages for instance. So you would need extendedKind() only in specific situations (like the places I use it in this commit).
I can easily revert this commit or have another approach.

derekcollison: The enum for what gets returned should include a NON_CLIENT option as well.

kozlovic: I guess you miss the fact that this is extendedKind(), not Type() which means that it can be invoked on any type of connection and will return say ROUTER, LEAF, but for a client, it may return CLIENT (if regular) or the new CLIENT_MQTT or CLIENT_WS.
What you suggest (and I do have a version of that) is to have a clientType() that should be invoked only for c.kind == CLIENT and that will then return NATS/MQTT/WS or NON_CLIENT if incorrectly invoked on a non CLIENT connection.

derekcollison: Yes the second option is what I had in mind, but minor so..

kozlovic: Ok, replaced that commit with 035cffae375e34eed0020aa6f37a533823721601, which I believe does what you asked.

derekcollison (approved):
LGTM

kozlovic:
@derekcollison I could have merged and then work on optimization, but I decided to include this one in the main PR. If could could review https://github.com/nats-io/nats-server/pull/1754/commits/1d7c4712a5cfb4384a1a6498e75382f7be975f6a when you have time. I still have some work to do with conversion of topic/subject for delivery of messages to subscriptions. If I am not done soon with that one, we could merge this PR and will continue on perf improvement later.

derekcollison (approved):
LGTM

PR #1755, opened 2020-11-30
Send LS- without origin to route

### LINKED ISSUE #1751: No response from any moment in leaf node configuration
## Defect

 - [v ] Included `nats-server -DV` output

2.2.0-beta.33
master branch (2e28de50)

windows 2010

The configuration is the same as https://github.com/nats-io/nats-server/issues/1743
While testing the fixed, I found another issue.

I used nats-rply and a simple request client written by go.

simple client code is as follows.
```
func main() {
    var urls = flag.String("s", nats.DefaultURL, "The nats server URLs (separated by comma)")
    var userCreds = flag.String("creds", "", "User Credentials File")
    var showHelp = flag.Bool("h", false, "Show help message")

    log.SetFlags(0)
    flag.Usage = usage
    flag.Parse()

    if *showHelp {
        showUsageAndExit(0)
    }

    args := flag.Args()
    if len(args) < 2 {
        showUsageAndExit(1)
    }
    subj, payload, numStr := args[0], []byte(args[1]), args[2]
    num, err := strconv.Atoi(numStr)

    // Connect Options.
    opts := []nats.Option{nats.Name("NATS Sample Requestor")}

    // Use UserCredentials
    if *userCreds != "" {
        opts = append(opts, nats.UserCredentials(*userCreds))
    }

    // Connect to NATS
    nc, err := nats.Connect(*urls, opts...)
    if err != nil {
        log.Fatal(err)
    }
    defer nc.Close()
    
    var reqCount, resCount int = 0, 0
    
    // Subscribe reply subject
    reSubject := "RE"
    subRe, err := nc.QueueSubscribe(reSubject, reSubject, func(m *nats.Msg) {
        resCount++
        log.Printf("Received [%v,%d] : '%s'", m.Subject, resCount, string(m.Data))
    })
    subRe.SetPendingLimits(-1, -1)
    
    for i :=0; i < num; i++ {
        err := nc.PublishRequest(subj, "RE", payload)
        if err != nil {
            if nc.LastError() != nil {
                log.Fatalf("%v for request", nc.LastError())
            }
            log.Fatalf("%v for request", err)
        }

        reqCount++
        log.Printf("Published [%s,%d] : '%s'", subj, reqCount, payload)
    }

    for (resCount < reqCount) {}
}
```
This simple client subscribe 'RE' to receive response.
And publish_request num times.

**Step of execution**
1) nats-rply subscribe 'foo' on srv-2
2) execute request-client foo hello 10 on leaf-2
3) execute request-client repeatedly

#### Expected result:

The request-client receives all responses.

#### Actual result:

The request-client doesn't receive responses from any moment.
err2

Looking at the srv-2 log, count of publish response to 'RE' is 20, but count of RMSG is 17.
And 'processRemoteUnsub parse error' logged at srv-2.
```
[22940] 2020/11/27 08:15:42.506173 [TRC] [::1]:6222 - rid:2 - <<- [RS- ncs $G RE RE]
[22940] 2020/11/27 08:15:42.506173 [ERR] [::1]:6222 - rid:2 - processRemoteUnsub parse error: 'ncs $G RE RE'
[22940] 2020/11/27 08:15:42.506173 [INF] [::1]:6222 - rid:2 - Router connection closed: Protocol Violation
[22940] 2020/11/27 08:15:42.506173 [DBG] Attempting reconnect for solicited route "nats://localhost:6222"
```
It is not reproduced in the release version.
Full logs are attatched.
[1127_natslog.zip](https://github.com/nats-io/nats-server/files/5605450/1127_natslog.zip)

kozlovic:
When cluster origin code was added, a server may send LS+ with
an origin cluster name in the protocol. Parsing code from a ROUTER
connection was adjusted to understand this LS+ protocol.
However, the server was also sending an LS- with origin but the
parsing code was not able to understand that. When the unsub was
for a queue subscription, this would cause the parser to error out
and close the route connection.

This PR sends an LS- without the origin in this case (so that tracing
makes sense in term of LS+/LS- sent to a route). The receiving side
then traces appropriate LS- but processes as a normal RS-.

Resolves #1751

### DIFF
--- a/server/leafnode_test.go
+++ b/server/leafnode_test.go
@@ -1947,7 +1947,7 @@ func TestLeafNodeNoDuplicateWithinCluster(t *testing.T) {
 	}
 }
 
-func TestLeafNodeNoRouteParserError(t *testing.T) {
+func TestLeafNodeLMsgSplit(t *testing.T) {
 	// This set the cluster name to "abc"
 	oSrv1 := DefaultOptions()
 	oSrv1.LeafNode.Host = "127.0.0.1"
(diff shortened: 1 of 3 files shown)


derekcollison (approved):
LGTM

PR #1759, opened 2020-12-02
Better implementation to detect various cycles from account imports/exports.

derekcollison:

/cc @nats-io/core


kozlovic (approved):
LGTM

server/accounts.go
```diff
@@ -1344,26 +1345,85 @@ func (a *Account) AddServiceImportWithClaim(destination *Account, from, to strin
...
+func (a *Account) hasStreamExportMatching(to string) bool {
+	for subj := range a.exports.streams {
+		if subjectIsSubsetMatch(to, subj) {
 			return true
 		}
 	}
 	return false
 }
 
+func (a *Account) checkStreamImportsForCycles(to string, visited map[string]bool) bool {
```
matthiashanel, line 1398: So there is no communality between the two that could be added or used to avoid duplicating what is essentially the same code?

server/accounts.go
```diff
@@ -1344,26 +1345,85 @@ func (a *Account) AddServiceImportWithClaim(destination *Account, from, to strin
 		return ErrServiceImportAuthorization
 	}
 
-	if a.importFormsCycle(destination, from, to) {
-		return ErrServiceImportFormsCycle
+	// Check if this introduces a cycle before proceeding.
+	if a.serviceImportFormsCycle(destination, from) {
+		return ErrImportFormsCycle
```
matthiashanel, line 1350: Will this cause a warning trace to help detect these issues?

matthiashanel (changes requested):
remove recursion

server/accounts.go
```diff
@@ -1344,26 +1345,85 @@ func (a *Account) AddServiceImportWithClaim(destination *Account, from, to strin
...
+			a.mu.RUnlock()
+			if visited[si.acc.Name] {
+				return true
+			}
+			// Push ourselves and check si.acc
+			visited[a.Name] = true
+			if subjectIsSubsetMatch(si.from, from) {
+				from = si.from
+			}
+			if si.acc.checkServiceImportsForCycles(from, visited) {
```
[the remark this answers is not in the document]
derekcollison: Yes I am ok with recursion here.

server/accounts.go
derekcollison: I had them combined for awhile but was hard to read back TBH, so split them back up again. The way we track down the cycles is different enough IMO.

server/accounts.go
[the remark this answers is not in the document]
matthiashanel: @derekcollison  Then limit the recursion depth by checking the size of the visited map being smaller than an arbitrary value

server/accounts.go
derekcollison: Yes I believe so..

server/accounts.go
matthiashanel: duplicated code is never good

server/accounts.go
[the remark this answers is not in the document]
derekcollison: We need to detect cycles so track down as far as we need, why would we limit? I mean we could and error out but it could be legit (although very unlikely unless someone trying to do something malicious).

server/accounts.go
derekcollison: Read them a but more closely..

server/accounts.go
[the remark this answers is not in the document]
matthiashanel: @derekcollison so that a malicious person can't upload linearly chained accounts and force out to exceed stack space. 

make that pass in a recursion depth vs checking the size of visited.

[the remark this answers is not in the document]
derekcollison: That might be good, but we can just check `len(visited)` I think.

server/accounts.go
matthiashanel: So why is it that streams are being iterated over in a separate pass and services are not?

matthiashanel:
@derekcollison 
> That might be good, but we can just check len(visited) I think.

When you pass in a debth or a level you limit the maximum recursion (max chain you follow).
When you use len(visited) you limit the size of the graph and thus how much time you will spend before giving up. 
This would hit accounts with many imports, but maybe good too?

derekcollison:
We only descend if we know the subjects would potentially collide and we add into visited before descending, so I think its fine to just check len(visited).

server/accounts.go
derekcollison: Not following that statement, we iterate over each one looking for cycles separately.

derekcollison:
Will add a test and a bail statement if we get too far down the rabbit hole.

matthiashanel:
> We only descend if we know the subjects would potentially collide and we add into visited before descending, so I think its fine to just check len(visited).

would make sense for streams too then or not?

Types and subjects asside, I'm still a firm believer that we be better off ignoring both and just check for cycles between accounts. 
just imagine the setup, especially when using jwt.
two accounts that mutually end up importing each others information.
In config it's easy to write down. but in jwt, you need to create both with exports first, then add the imports.
If the functionalities do not mutually depend on each other for functioning, when adding an account you will always have to modify the other. Not sure if that is desirable from a config point of view (hub and spoke?)
functionality that tightly coupled would best be in the same account.

It would also simplify the logic and be beneficial for other logic.

derekcollison:
Yes will do a test for streams too.

I know you are a fan of a simple DAG, I get it, but at the user experience level that would break down very quickly and not make much sense to our users that we disallowed simple and valid use cases.

PR #1760, opened 2020-12-02
[FIXES] https://github.com/nats-io/jetstream/issues/396

derekcollison:
Had a deadlock with new preconditions. We need to hold lock across Store() call but that call could call into storeUpdate() such that we may need to acquire the lock. We can enter this callback from the storage layer itself and the lock would not be held so added an atomic.

Resolves https://github.com/nats-io/jetstream/issues/396

/cc @nats-io/core

### DIFF
--- a/server/stream.go
+++ b/server/stream.go
@@ -775,11 +775,11 @@ func (mset *Stream) setupStore(fsCfg *FileStoreConfig) error {
 // Called for any updates to the underlying stream. We pass through the bytes to the
 // jetstream account. We do local processing for stream pending for consumers, but only
 // for removals.
-// Lock should not ne held.
+// Lock should not be held.
 func (mset *Stream) storeUpdates(md, bd int64, seq uint64, subj string) {
 	// If we have a single negative update then we will process our consumers for stream pending.
 	// Purge and Store handled separately inside individual calls.
-	if md == -1 {
+	if md == -1 && seq > 0 {
 		mset.mu.RLock()
 		for _, o := range mset.consumers {
 			o.decStreamPending(seq, subj)
@@ -1043,14 +1043,25 @@ func (mset *Stream) processInboundJetStreamMsg(_ *subscription, pc *client, subj
 		hdr = msg[:pc.pa.hdr]
 		msg = msg[pc.pa.hdr:]
 	}
-	seq, ts, err = store.StoreMsg(subject, hdr, msg)
-	if err == nil && seq > 0 {
-		mset.lseq = seq
-		mset.lmsgId = msgId
-	}
+
+	// Assume this will succeed.
+	olseq, olmsgId := mset.lseq, mset.lmsgId
+	mset.lseq++
+	mset.lmsgId = msgId
 
 	// We hold the lock to this point to make sure nothing gets between us since we check for pre-conditions.
 	mset.mu.Unlock()
+
+	// Store actual msg.
+	seq, ts, err = store.StoreMsg(subject, hdr, msg)
+
+	// If we did not succeed put those values back.
+	if err != nil || seq == 0 {
+		mset.mu.Lock()
+		mset.lseq = olseq
+		mset.lmsgId = olmsgId
+		mset.mu.Unlock()
+	}
 
 	if err != nil {
 		if err != ErrStoreClosed {
@@ -1148,16 +1159,16 @@ func (mset *Stream) Name() string {
 }
 
 func (mset *Stream) internalSendLoop() {
-	mset.mu.Lock()
+	mset.mu.RLock()
 	c := mset.client
 	if c == nil {
-		mset.mu.Unlock()
+		mset.mu.RUnlock()
 		return
 	}
 	s := c.srv
 	sendq := mset.sendq
 	name := mset.config.Name
-	mset.mu.Unlock()
+	mset.mu.RUnlock()
 
 	// Warn when internal send queue is backed up past 75%
 	warnThresh := 3 * msetSendQSize / 4
--- a/server/consumer.go
+++ b/server/consumer.go
@@ -2265,13 +2265,9 @@ func (o *Consumer) incStreamPending(sseq uint64, subj string) {
 
 func (o *Consumer) decStreamPending(sseq uint64, subj string) {
 	o.mu.Lock()
-	defer o.mu.Unlock()
-
 	// Ignore if we have already seen this one.
-	if sseq < o.sseq || o.sgap == 0 {
-		return
-	}
-	if o.isFilteredMatch(subj) {
+	if sseq >= o.sseq && o.sgap > 0 && o.isFilteredMatch(subj) {
 		o.sgap--
 	}
+	o.mu.Unlock()
 }
--- a/test/jetstream_test.go
+++ b/test/jetstream_test.go
@@ -7546,8 +7546,8 @@ func TestJetStreamAPIStreamListPaging(t *testing.T) {
 		defer os.RemoveAll(config.StoreDir)
 	}
 
-	// Create 4X limit
-	streamsNum := 4 * server.JSApiNamesLimit
+	// Create 2X limit
+	streamsNum := 2 * server.JSApiNamesLimit
 	for i := 1; i <= streamsNum; i++ {
 		name := fmt.Sprintf("STREAM-%06d", i)
 		cfg := server.StreamConfig{Name: name}
@@ -7601,7 +7601,6 @@ func TestJetStreamAPIStreamListPaging(t *testing.T) {
 
 	checkResp(reqList(0), server.JSApiNamesLimit, 0)
 	checkResp(reqList(server.JSApiNamesLimit), server.JSApiNamesLimit, server.JSApiNamesLimit)
-	checkResp(reqList(2*server.JSApiNamesLimit), server.JSApiNamesLimit, 2*server.JSApiNamesLimit)
 	checkResp(reqList(streamsNum), 0, streamsNum)
 	checkResp(reqList(streamsNum-22), 22, streamsNum-22)
 	checkResp(reqList(streamsNum+22), 0, streamsNum)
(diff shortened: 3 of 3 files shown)


server/stream.go
```diff
@@ -780,11 +782,16 @@ func (mset *Stream) storeUpdates(md, bd int64, seq uint64, subj string) {
 	// If we have a single negative update then we will process our consumers for stream pending.
 	// Purge and Store handled separately inside individual calls.
 	if md == -1 {
-		mset.mu.RLock()
+		needLock := atomic.LoadInt32(&mset.lh) == 0
```
kozlovic, line 785: I think this is wrong. If 2 threads are doing store operations that lead to the invocation of this callback, one thread may be under the mset lock and has stored the atomic, while the other thread is not under the mset lock. If the second thread evaluates the atomic it will think that it does not need the lock - because it has been set by the first thread. It is possible that the first thread finishes processing of callback, unsets atomic and releases lock, and then the second thread resumes execution and proceeds with executing this code without the lock, which is technically wrong. (granted, it is unlikely in this sequence of events since it means that thread 2 execution would have to be suspended after evaluation of the atomic and before getting the mset RLock, but...)

So either all invocations of this callback would have to be made under the mset lock,  which clearly we don't always (like Purge, etc..), or if there is no need for high accuracy, then these callback executions could be serialized and dispatched from separate long running go routine (similar to async callbacks in go client).

NOTE: maybe we could ensure that cases where `md==-1` is invoked from code that does have the mset lock, in which case we would not need this code at all. I am guessing that the case is for removing of a message or purge when there was a single message, no?

kozlovic (approved):
Since I don't think that any of the 2 alternative approaches are likely, I guess we don't have much choice, but I do believe that current approach is not technically correct. See comments.

server/stream.go
derekcollison: The atomic is only set when the mset lock is held and is also only cleared before said lock is released. Yes this happens when we have a limit on number of messages or bytes and storing the new message causes us to drop one.

derekcollison: If we do not protect the store with the mset lock though I could see a callback here from a RemoveMsg() that may cause issues. We could make sure to protect and single removals.

What I wanted to avoid is releasing lock and reacquiring since I need those preconditions set before I release the lock.

kozlovic: I understand that the atomic is set under the mset lock, the issue is that a thread could be doing say a Purge() which is executed outside of the mset lock, which another thread may be processing inbound msg and calling StoreMsg() (where this atomic is set/unset) under the mset lock.

derekcollison: Purge() is ok. Its RemoveMsg() we need to make sure is protected by mset lock.

kozlovic: https://github.com/nats-io/nats-server/pull/1760#discussion_r534440035

Right, this is why I mentioned that we should either always have the storeUpdates callback be invoked from under the mset lock, which means all store operations that can trigger the store update callback should be under the mset lock, or at least look at cases where callback is invoked with md==-1, which I assume is remove msg and purge of single message?

kozlovic: Purge() calls with "-(purged)", so if purged is 1, then you have the problem.

derekcollison: Let me see if there is a better way. Thanks for the feedback, helpful (and correct)..

derekcollison: ok please take another look. Thanks..

server/stream.go
```diff
@@ -1043,15 +1043,30 @@ func (mset *Stream) processInboundJetStreamMsg(_ *subscription, pc *client, subj
...
+	mset.lmsgId = msgId
 
 	// We hold the lock to this point to make sure nothing gets between us since we check for pre-conditions.
 	mset.mu.Unlock()
 
+	// Can not hold lock here since storeUpdates() could need lock.
+	seq, ts, err = store.StoreMsg(subject, hdr, msg)
+
+	// Put back if we did not complete correctly.
+	if mset.lseq != seq {
```
kozlovic, line 1059: you can't access mset.lseq unprotected..

Also here are some sequences that worry me. Multiple producers (different server readloops) send to same stream:
- thread 1 does mset.lseq++, which say it then equal to 1, releases lock but does not call StoreMsg() yet.
- thread 2 does mset.lseq++, which means it is now equal to 2, releases and calls StoreMsg(), which returns seq==1.
- thread 1 does call StoreMsg(), which returns seq==2, grabs the lock and says (assuming that the check is now under lock and does not use the value that was evaluated before unlocking) mset.lseq == seq, so it will leave that alone.
- thread 2 resumes, get lock and check that mset.lseq(==2) is different from seq(==1), so it will then do mset.lseq=1, which is wrong!

If you capture mset.lseq value after mset.lseq++ and before releasing lock, then you have still a problem:
- thread 1 does mset.lseq++ => savedLseq==1, releases lock
- thread 2 does mset.lseq++ => savedLseq==2, releases lock
- thread 2 calls StoreMsg(), returns seq==1
- thread 1 calls StoreMsg(), returns seq==2
- thread 1 grabs lock see that savedLseq(==1) != seq(==2), so set mset.lseq=2
- thread 2 grabs lock, see that savedLseq(==2) != seq(==1), so set mset.lseq=1

kozlovic (changes requested):
This does not work either..

server/stream.go
derekcollison: Good catch, let me think some more about it.

derekcollison: ok hopefully third time is a charm.

server/stream.go
```diff
@@ -1207,6 +1226,41 @@ func (mset *Stream) internalSendLoop() {
...
+	mset.mu.RUnlock()
+
+	// Warn when internal send queue is backed up past 75%
+	warnThresh := 3 * msetSendQSize / 4
+	warnFreq := time.Second
+	last := time.Now().Add(-warnFreq)
+
+	for {
+		if len(decq) > warnThresh && time.Since(last) >= warnFreq {
+			s.Warnf("Jetstream internal decrement pending queue > 75%% for account: %q stream: %q", c.acc.Name, name)
```
kozlovic, line 1248: Should we capture the c.acc.Name at the top of function? Is this client an internal client that never will have c.acc changed/etc.. even during reload?

server/stream.go
```diff
@@ -779,12 +784,8 @@ func (mset *Stream) setupStore(fsCfg *FileStoreConfig) error {
...
 	// If we have a single negative update then we will process our consumers for stream pending.
 	// Purge and Store handled separately inside individual calls.
-	if md == -1 {
-		mset.mu.RLock()
-		for _, o := range mset.consumers {
-			o.decStreamPending(seq, subj)
-		}
-		mset.mu.RUnlock()
+	if md == -1 && seq > 0 {
+		mset.decq <- &decp{seq, subj}
```
kozlovic, line 788: Yeah, but you now have the problem that this channel may be full and will block! And if this gets called under mset lock, then the "internalDecLoop" may be blocked wanted to get the RLock()..

kozlovic (commented):
This is not what I meant with having a long running go routine. I meant that invocation of callbacks themselves should be serialized and invoked from that routine. I understand that it may not be possible because you may need "synchronous" feedback on store updates.
But only if the callbacks themselves were invoked asynchronously from a dedicated go routine would you be able to safely acquire the mset lock in storeUpdates.

server/stream.go
derekcollison: name won't change.

kozlovic: Not the name of the account, I meant will `c.acc` ever be set to nil/reset during a config reload?

server/stream.go
derekcollison: Yes I thought about that.. Unfortunately this has folks blocked, main branch is broke for any stream with limits.
Let me try again since I have a fresher set of eyes.

kozlovic: I understand. In that case, the version with atomic (although wrong) is less dangerous. The worst case is that the map would be iterated without locking - which a test may report as data race - but will not deadlock like this approach could.

derekcollison: Sorry, no acc will not change either.

server/stream.go
derekcollison: Sorry, no acc will not change either.

server/stream.go
derekcollison: Agree deadlock is what needs to be first priority, then the preconditions, etc.

derekcollison:
ok went back to approach that StoreMsg can not be called with the stream lock held. Assume we will succeed and if not will put back what we remember. Under heavy load we may not put back the exact right thing but this is only for precondition checks so they may fail when there might have been a chance they could succeed.

server/stream.go
```diff
@@ -1043,15 +1043,26 @@ func (mset *Stream) processInboundJetStreamMsg(_ *subscription, pc *client, subj
...
+	mset.lseq++
+	mset.lmsgId = msgId
 
 	// We hold the lock to this point to make sure nothing gets between us since we check for pre-conditions.
 	mset.mu.Unlock()
 
+	// Store actual msg.
+	seq, ts, err = store.StoreMsg(subject, hdr, msg)
+
+	// If we did not succedd put those values back.
```
kozlovic, line 1058: typo

server/stream.go
```diff
@@ -1043,15 +1043,26 @@ func (mset *Stream) processInboundJetStreamMsg(_ *subscription, pc *client, subj
...
+	if err != nil || seq == 0 {
+		mset.mu.Lock()
+		mset.lseq = olseq
```
kozlovic, line 1061: Understand that this may completely mess up the state of the stream with concurrent calls to this function and if any of the thread failed to store a message. 

My strong opinion is that I would rather have the original approach with the atomic check (again, although wrong) than this one. As I said, the worse that can happen in the atomic approach is the unprotected map iteration of the consumers... 
Btw, you could always have a dedicated lock to protect the consumer map or use a sync.map - I understand that you use len() so adding/checking in some cases would still require to be done under the mset lock, but for storeUpdates, since it is just iterating, the sync.map itself may be ok).

kozlovic (commented):
I am sorry, but I find this approach really dangerous. This will leave the stream in a very bad state if StoreMsg() returns an error. Better panic then that continue with corrupting the stream's runtime state..

server/stream.go
derekcollison: thx will fix

server/stream.go
derekcollison: Understood, but was not a fan of the atomic approach. I think we go with this for now.

derekcollison: We used to not hold lock for StoreMsg() fyi, I moved where we released the lock just recently due to precondition stuff.

kozlovic: Well, there was a reason to move it under lock. I fear that this is going to be really bad if a failure occurs. At least we log the error, so that would be an indication. I see that you do that if `err != nil || seq == 0`. Is there a case where an error is not returned but seq==0 still indicates an error? If so, then in this case we would not even know from the logs..

derekcollison: I only moved it under the lock for the preconditions. It has always been outside, which is what I put it back to and assume preconditions will do the right thing. If there are multiple producers the preconditions on a failure could be put back to wrong values (temporarily), but if you have multiple producers you are probably not doing pre-conditions. At least with highly contending multiple producers..

kozlovic: So those values are used just in case publishers provide headers that require expected last sequence and/or last message ID? That's the only thing this would affect?

derekcollison: Correct.

kozlovic (approved):
LGTM

PR #1761, opened 2020-12-02
Limit search depth for account cycles for imports

derekcollison:
Although Go stacks are not limited per se (just by memory), we wanted to put a limit to the depth to check. I chose 1024, which in my opinion if we hit this from a chaining perspective something is probably wrong.

/cc @nats-io/core


kozlovic (approved):
LGTM

matthiashanel (approved):
LGTM

derekcollison:
Thanks @matthiashanel for being diligent here.

matthiashanel:
You're welcome. Thank you @derekcollison

PR #1763, opened 2020-12-03
Added Compact to store interface for WAL functionality

derekcollison:

/cc @nats-io/core

### DIFF
--- a/server/filestore.go
+++ b/server/filestore.go
@@ -2319,11 +2319,11 @@ func (fs *fileStore) dmapEntries() int {
 
 // Purge will remove all messages from this store.
 // Will return the number of purged messages.
-func (fs *fileStore) Purge() uint64 {
+func (fs *fileStore) Purge() (uint64, error) {
 	fs.mu.Lock()
 	if fs.closed {
 		fs.mu.Unlock()
-		return 0
+		return 0, ErrStoreClosed
 	}
 
 	purged := fs.state.Msgs
@@ -2370,7 +2370,43 @@ func (fs *fileStore) Purge() uint64 {
 		cb(-int64(purged), -rbytes, 0, _EMPTY_)
 	}
 
-	return purged
+	return purged, nil
+}
+
+// Compact will remove all messages from this store up to
+// but not including the seq parameter.
+// Will return the number of purged messages.
+func (fs *fileStore) Compact(seq uint64) (uint64, error) {
+	if seq == 0 {
+		return fs.Purge()
+	}
+
+	if _, err := fs.msgForSeq(seq); err != nil {
+		return 0, err
+	}
+
+	var purged uint64
+	for fseq := fs.firstSeq(); fseq < seq; fseq = fs.firstSeq() {
+		if found, err := fs.removeMsg(fseq, false); err != nil {
+			if err == ErrStoreMsgNotFound {
+				continue
+			} else if err == ErrStoreEOF {
+				err = nil
+			}
+			return purged, err
+		} else if found {
+			purged++
+		}
+	}
+
+	return purged, nil
+}
+
+func (fs *fileStore) firstSeq() uint64 {
+	fs.mu.RLock()
+	fseq := fs.state.FirstSeq
+	fs.mu.RUnlock()
+	return fseq
 }
 
 // Returns number of msg blks.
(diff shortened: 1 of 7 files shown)


server/filestore.go
```diff
@@ -2370,7 +2370,41 @@ func (fs *fileStore) Purge() uint64 {
...
+
+// Compact will remove all messages from this store up to
+// but not including the seq parameter.
+// Will return the number of purged messages.
+func (fs *fileStore) Compact(seq uint64) (uint64, error) {
+	if seq == 0 {
+		return fs.Purge()
+	}
+
+	if fs.isClosed() {
```
kozlovic, line 2384: Why bother? The store could be closed after this call. The functions isClosed(), msgForSeq(), firstSeq() and removeMsg() all individually get rlock/lock, and by the look of it check for closed state (except firstSeq).
To be honest, I think that this function should have locking from start to finish.. which means likely having "locked" versions of those calls.
(what is going to happen during the compact if messages are removed due to limits because new messages are added?)

server/filestore.go
```diff
@@ -2370,7 +2370,41 @@ func (fs *fileStore) Purge() uint64 {
...
+	}
+
+	if _, err := fs.msgForSeq(seq); err != nil {
+		return 0, ErrStoreMsgNotFound
+	}
+
+	var purged uint64
+
+	for fseq := fs.firstSeq(); fseq < seq; fseq = fs.firstSeq() {
+		if found, _ := fs.removeMsg(fseq, false); found {
```
kozlovic, line 2395: Either rewrite this function with full lock or check error here since store may close. Also I would not recommend ignoring an error here since this may be critical info for caller.

server/memstore.go
```diff
@@ -270,7 +270,46 @@ func (ms *memStore) Purge() uint64 {
...
+		purged++
+		delete(ms.msgs, seq)
+	}
+	ms.state.Msgs -= purged
+	ms.state.Bytes -= bytes
+
+	cb := ms.scb
+	ms.mu.Unlock()
+
+	if cb != nil {
```
kozlovic, line 308: Is it copy/paste from Purge() and was not intending to do that here? If it is intentional, should you do the same for the filestore implementation?

server/filestore.go
derekcollison: True..

derekcollison: I don't agree it would be easier to hold the lock the whole time. Would be all new and custom code.

server/filestore.go
derekcollison: Fair. Will update.

server/memstore.go
derekcollison: Filestore is totally different impl. This is meant to be here since memstore version has to update since it is not calling remove individually like fs version. fs version is way better off calling into the low level delete since that layer knows about message blocks, jumping blocks to update first seq etc..

kozlovic: ok

server/filestore.go
derekcollison: fixed.

server/filestore.go
derekcollison: Removed isClosed() check.

server/filestore.go
```diff
@@ -2370,7 +2370,38 @@ func (fs *fileStore) Purge() uint64 {
...
+	if seq == 0 {
+		return fs.Purge()
+	}
+
+	if _, err := fs.msgForSeq(seq); err != nil {
+		return 0, err
+	}
+
+	var purged uint64
+	for fseq := fs.firstSeq(); fseq < seq; fseq = fs.firstSeq() {
```
kozlovic, line 2389: So is the caller responsible to make sure that `seq` is valid? What would happen if I call Compact(100) on a file that has 1 to 10. Will we get to try to removeMsg(11) and that will fail with not found (or similar error)? If so, it means that Compact() did remove messages but still returns a failure. It maybe ok since the caller asked to remove up to 100 (not included) and that functions only removed 1 to 10. But just making sure that this is what you expect.

derekcollison: We check that it is valid on line 2384 so would get an error right away. However what you asked me to add, checking the error the way I do is wrong. I need to skip across these messages if they are simply not there. The code before did that since it only cared if we removed the message. 

Let me add more error code handling to piece out different errors.

derekcollison: ok I think this is correct now.

derekcollison: Once we are good will squash.

kozlovic (approved):
LGTM

PR #1764, opened 2020-12-04
fix test timing issue and flapper caused by unnecessary pop/push

matthiashanel:
fixed timing issue in unit test.
added another unit test for lru.
fixed pop / push issue. this caused the lru list to be reordered, which tripped of test assumptions.

### DIFF
--- a/server/dirstore.go
+++ b/server/dirstore.go
@@ -686,18 +686,15 @@ func (store *DirJWTStore) startExpiring(reCheck time.Duration, limit int64, evic
 			now := time.Now()
 			store.Lock()
 			if pq.Len() > 0 {
-				if it := heap.Pop(pq).(*jwtItem); it.expiration <= now.Unix() {
+				if it := pq.heap[0]; it.expiration <= now.Unix() {
 					path := store.pathForKey(it.publicKey)
-					if err := os.Remove(path); err != nil {
-						heap.Push(pq, it) // retry later
-					} else {
+					if err := os.Remove(path); err == nil {
+						heap.Pop(pq)
 						pq.unTrack(it.publicKey)
 						xorAssign(&pq.hash, it.hash)
 						store.Unlock()
-						continue // we removed an entry, check next one
+						continue // we removed an entry, check next one right away
 					}
-				} else {
-					heap.Push(pq, it)
 				}
 			}
 			store.Unlock()
--- a/server/dirstore_test.go
+++ b/server/dirstore_test.go
@@ -457,7 +457,7 @@ func TestExpiration(t *testing.T) {
 			return
 		}
 	}
-	t.Errorf("Waited more than 4 seconds for the file with expiration 1 second to expire")
+	t.Fatalf("Waited more than 4 seconds for the file with expiration 1 second to expire")
 }
 
 func TestLimit(t *testing.T) {
@@ -886,7 +886,7 @@ func TestNotificationOnPack(t *testing.T) {
 	notificationChan := make(chan struct{}, len(jwts)) // set to same len so all extra will block
 	notification := func(pubKey string) {
 		if _, ok := jwts[pubKey]; !ok {
-			t.Errorf("Key not found: %s", pubKey)
+			t.Fatalf("Key not found: %s", pubKey)
 		}
 		notificationChan <- struct{}{}
 	}
--- a/server/dirstore_test.go
+++ b/server/dirstore_test.go
@@ -22,6 +22,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"math"
+	"math/rand"
 	"os"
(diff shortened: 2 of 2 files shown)


server/dirstore.go
```diff
@@ -686,18 +686,15 @@ func (store *DirJWTStore) startExpiring(reCheck time.Duration, limit int64, evic
 			now := time.Now()
 			store.Lock()
 			if pq.Len() > 0 {
-				if it := heap.Pop(pq).(*jwtItem); it.expiration <= now.Unix() {
+				if it := pq.heap[0]; it.expiration <= now.Unix() {
```
derekcollison, line 689: So you have second resolution only here, which I understand is fine based on usage. For tests with higher time based resolutions sometimes we do stuff like.

```go
now := time.Now().Sub(900*time.Millisecond)
```

This allows the test to run in ~100ms versus over a second. Just food for thought.

server/dirstore_test.go
```diff
@@ -56,7 +57,7 @@ func require_NoError(t *testing.T, err error) {
 func require_Error(t *testing.T, err error) {
 	t.Helper()
 	if err == nil {
-		t.Errorf("require no error, but got: %v", err)
+		t.Errorf("require error, but got none")
```
derekcollison, line 60: Why Errorf vs Fatalf? Different Go routine?

derekcollison (approved):
LGTM - a few comments.

server/dirstore_test.go
kozlovic: Agreed. I usually don't use anything but Fatal() because there is no point in continuing the test.

kozlovic (approved):
LGTM

server/dirstore.go
matthiashanel: unfortunately I am stuck with second resolution due to jwt

derekcollison: Understood, but you can still use the trick above to speed up the test.

server/dirstore_test.go
matthiashanel: not sure, this code was moved from account server and originally dependent on the require library.
probably at the time I saw t.Errorf more often.

Fixed for the entire file

server/dirstore_test.go
```diff
@@ -436,28 +437,27 @@ func TestExpiration(t *testing.T) {
...
-	for i := 1; i <= 3; i++ {
-		account(i)
-		nh := dirStore.Hash()
-		require_NotEqual(t, h, nh)
-		h = nh
-	}
-	time.Sleep(500 * time.Millisecond)
-	for i := 3; i > 0; i-- {
+	failAt := time.Now().Add(4 * time.Second)
+	for time.Now().Before(failAt) {
```
derekcollison, line 450: So this makes this test run for at least 4 seconds?

server/dirstore_test.go
```diff
@@ -436,28 +437,27 @@ func TestExpiration(t *testing.T) {
...
-		require_NotEqual(t, h, nh)
-		h = nh
+		if len(f) == 1 {
+			lh := dirStore.Hash()
+			require_Equal(t, string(hNoExp[:]), string(lh[:]))
+			return
+		}
 	}
-	assertStoreSize(t, dirStore, 0)
+	t.Errorf("Waited more than 4 seconds for the file with expiration 1 second to expire")
```
derekcollison, line 460: t.Fatalf

server/dirstore_test.go
kozlovic: No, as you can see there is a return in the loop once the assertion is done. The 4sec is the max before reporting error

derekcollison: So should only run 100ms from first sleep etc in optimum case yes?

server/dirstore.go
kozlovic: But this is actual code, not test code. Are you saying that we do that in server code in some places? I don't think so.

kozlovic: in test code we sometimes change expiration time of a jwt, or some other things. we can sometimes access server internals and change values, but this is done in test code itself, not in server code.

matthiashanel: are you suggesting to round down the current time? Is this done in other places of the code?

derekcollison: No apologies, thought it was test code stuff.

server/dirstore_test.go
matthiashanel: about a second.
In part expiration in the jwt has to be bigger than issue at. and so we have to do rounding when issuing the jwt.

server/dirstore_test.go
matthiashanel: fixed, thanks

kozlovic (approved):
LGTM

PR #1765, opened 2020-12-04
use the natscli repo instead of jetstream

ripienaar:

/cc @nats-io/core


ripienaar:
This should also ensure that natscli uses the dependencies declared in its go.mod.

derekcollison (approved):
LGTM

PR #1767, opened 2020-12-10
MQTT updates

kozlovic:
- Keep track of published topic to avoid conversion in some case
- Rework tracing of some MQTT protocols
- Rework topic-to-subject conversion
- Update some tests

### DIFF
--- a/server/mqtt.go
+++ b/server/mqtt.go
@@ -175,6 +175,7 @@ type mqttPersistedSession struct {
 }
 
 type mqttRetainedMsg struct {
+	Topic  []byte `json:"topic,omitempty"`
 	Msg    []byte `json:"msg,omitempty"`
 	Flags  byte   `json:"flags,omitempty"`
 	Source string `json:"source,omitempty"`
@@ -232,6 +233,7 @@ type mqttWriter struct {
 
 type mqttWill struct {
 	topic   []byte
+	subject []byte
 	message []byte
 	qos     byte
 	retain  bool
@@ -240,9 +242,12 @@ type mqttWill struct {
 type mqttFilter struct {
 	filter string
 	qos    byte
+	// Used only for tracing and should not be used after parsing of (un)sub protocols.
+	ttopic []byte
 }
 
 type mqttPublish struct {
+	topic   []byte
 	subject []byte
 	msg     []byte
 	sz      int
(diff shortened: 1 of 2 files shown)


server/consumer.go
```diff
@@ -961,6 +961,7 @@ func (o *Consumer) loopAndDeliverMsgs() {
 			c.pa.size = len(pm.msg) + len(pm.hdr)
 			c.pa.szb = []byte(strconv.Itoa(c.pa.size))
 			c.pa.reply = []byte(pm.reply)
+			c.pa.topic = []byte(pm.topic)
```
derekcollison, line 964: What exactly is the role of the topic here?

server/parser.go
```diff
@@ -47,6 +47,7 @@ type pubArg struct {
 	queues  [][]byte
 	size    int
 	hdr     int
+	topic   []byte // This is MQTT published topic but is not restricted to MQTT producers!
```
derekcollison, line 50: Isn't there a fixed way to go between the two?

server/consumer.go
kozlovic: When parsing an MQTT publish message, there is a topic, say "foo/bar". It is converted to foo.bar to do sublist match, etc.. but when it is time to deliver to MQTT subscriptions, if we have only `subject` which is foo.bar, we would need to convert foo.bar back to foo/bar. Instead, we save reference to topic foo/bar in c.pa.topic. However, for QoS1 when JS is the publisher (internal client) we pass the jsPubMsg.topic to c.pa.topic for processing in mqttDeliverMsgCb().
How we get it into jsPubMsg? This is in processInboundJetStreamMsg() where we pass pc.pa.topic if not empty.

server/parser.go
kozlovic: Not sure what you mean here. The comment was supposed to explain that although `topic` is used for the "subject" that is passed to the mqtt subscriptions, it can be used by non MQTT producers.
The case is a NATS publisher that has matching MQTT subscriptions. c.pa.topic will be empty but the first time we get to a mqttDeliverMsgCb(), we will do the conversion from subject to topic, but only once regardless of how many matching MQTT subscriptions there are for this NATS message.

derekcollison: Meaning there is a well known way to get from MQTT topic to NATS subject and back, why do we need to store it? strings.Replace() may not be too fast but wondering if saving makes a noticeable difference.

kozlovic: Unfortunately, it is not a simple strings.Replace(). There are more complex rules due to the how `/` can be used in MQTT. For instance, this is a valid MQTT topic: `/foo//bar/`, which a straight strings.Replace() would create an invalid NATS subject.

That being said, I am fine removing c.pa.topic and have the subject-to-topic conversion being made every time for NATS->MQTT but also (unfortunately) from MQTT QoS1 -> MQTT QoS1. The one case I can store in the specific MQTT producer's struct is MQTT Qos0.

Based on your response, I will update the PR (I am going to push an update anyway since I am adding some cases to the conversion test).

derekcollison: Let's do that and mark it with comment around potential future optimization.

derekcollison (approved):
LGTM

PR #1770, opened 2020-12-11
Break test up into shorter ones

matthiashanel:

 each test takes about 2 seconds. one second to observe expiration does not happen, and one to observe expiration.


kozlovic:
We could maybe use UnixNano() in dirstore.go which would help for tests that don't rely on JWT's expiration time to have a shorter running time. I have sent you some proposed changes and you will have a look and if that works, you will update this PR.

kozlovic (approved):
LGTM. For info, running all tests in dirstore_test.go now takes under 7s, as opposed to 27s.

PR #1771, opened 2020-12-12
[FIXED] Config reload for gateways/leaf remote TLS configurations

kozlovic:
Presence of TLS config in any remote gateway or leafnode would
cause the config reload to fail (because TLS config internal
content may change which fails the DeepEqual check).

This PR excludes the TLS configs in such case to check for
changes in gateways and leafnodes.

Although GW and LN config reload is technically supported, this
PR updates the internal remotes' TLS configuration so that
changes/updates to TLS certificates would take effect after
a configuration reload.

### DIFF
--- a/server/gateway.go
+++ b/server/gateway.go
@@ -370,6 +370,29 @@ func (s *Server) newGateway(opts *Options) error {
 	return nil
 }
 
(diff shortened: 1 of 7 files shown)


derekcollison (approved):
LGTM

PR #1775, opened 2020-12-13
Fix for https://github.com/nats-io/jetstream/issues/406

derekcollison:

Resolves #https://github.com/nats-io/jetstream/issues/406

/cc @nats-io/core

### DIFF
--- a/server/filestore.go
+++ b/server/filestore.go
@@ -2289,6 +2289,12 @@ func (mb *msgBlock) readIndexInfo() error {
 	mb.last.ts = readTimeStamp()
 	dmapLen := readCount()
 
+	// Check if this is a short write index file.
+	if bi < 0 || bi+checksumSize > len(buf) {
+		defer os.Remove(mb.ifn)
+		return fmt.Errorf("short index file")
+	}
+
 	// Checksum
 	copy(mb.lchk[0:], buf[bi:bi+checksumSize])
 	bi += checksumSize
--- a/server/filestore_test.go
+++ b/server/filestore_test.go
@@ -17,6 +17,7 @@ import (
 	"archive/tar"
 	"bytes"
 	"compress/gzip"
+	"encoding/base64"
 	"encoding/hex"
 	"encoding/json"
 	"fmt"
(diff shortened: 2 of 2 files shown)


server/filestore.go
```diff
@@ -2289,6 +2289,12 @@ func (mb *msgBlock) readIndexInfo() error {
 	mb.last.ts = readTimeStamp()
 	dmapLen := readCount()
 
+	// Check if this is a short write index file.
```
ripienaar, line 2292: I am not too familiar with the storage code - if we just nuke the index will it be rebuilt? Whats the impact of that?

derekcollison: Yes will rebuild from raw data.

derekcollison: I want to understand a bit more why she got the short write, I think there may be a timing bug somewhere as well.

ripienaar: Could also just be a kill -9 at the wrong time? I think there was a lot of hard resetting the server because the other bug is that she caused it to become unresponsive (which I now cant reproduce)

ripienaar:
LGTM

server/filestore_test.go
```diff
@@ -2410,3 +2412,16 @@ func TestFileStoreConsumerPerf(t *testing.T) {
...
+	// https://github.com/nats-io/jetstream/issues/406
+	badIdxBytes, _ := base64.StdEncoding.DecodeString("FgGBkw7D/f8/772iDPDIgbU=")
+	dir, _ := ioutil.TempDir("", "js-bad-idx-")
+	defer os.RemoveAll(dir)
+	fn := path.Join(dir, "1.idx")
+	ioutil.WriteFile(fn, badIdxBytes, 0644)
+	mb := &msgBlock{index: 1, ifn: fn}
+	if err := mb.readIndexInfo(); err == nil || !strings.Contains(err.Error(), "short index") {
+		t.Fatalf("Expected error during readIndexInfo(): %v", err)
+	}
```
kozlovic, line 2426: You could add a check that the file has been removed here..

kozlovic (approved):
LGTM

PR #1776, opened 2020-12-16
[Added] ability to use jwt latency sampling properties headers/share

matthiashanel:

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -2,7 +2,7 @@ module github.com/nats-io/nats-server/v2
 
 require (
 	github.com/minio/highwayhash v1.0.0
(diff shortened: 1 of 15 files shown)


kozlovic (changes requested):
With the go mod changes (tidy/vendor), it seems that we have now vendor/golang.org/x/sys/unix/ (as seen in the update to vendor/modules.txt) but I guess you did not `git add vendor/` it?

matthiashanel:
> With the go mod changes (tidy/vendor), it seems that we have now vendor/golang.org/x/sys/unix/ (as seen in the update to vendor/modules.txt) but I guess you did not `git add vendor/` it?

@kozlovic Yes, thanks for catching that

kozlovic (approved):
LGTM

PR #1777, opened 2020-12-16
Updated go mod/vendor due to PR #1737

kozlovic:
The PR #1737 added a new file with an import of "golang.org/x/sys/unix"
but vendor directory was not updated.

Related to #1737


philpennock (approved):
Can't wait for a modules switch which lets us get rid of the vendoring.

PR #1779, opened 2020-12-17
updated iteration of signing keys (previously a list, now a map).

aricart:
[update] jwt2 now signing keys is a map, not a list - internal check for what changed when account loads and update copies them out, so the extraction needed to be different. Note this is referencing a non-released JWT library.

/cc @nats-io/core


kozlovic:
@aricart As you can see, there is a test that was using this as a list. So this test should be updated.

aricart:
> @aricart As you can see, there is a test that was using this as a list. So this test should be updated.

yes! - didn't show in my usages for some reason. Fixed now.

server/accounts.go
```diff
@@ -2844,11 +2844,10 @@ func (s *Server) updateAccountClaimsWithRefresh(a *Account, ac *jwt.AccountClaim
 	// update account signing keys
 	a.signingKeys = nil
 	signersChanged := false
-	if len(ac.SigningKeys) > 0 {
-		// insure copy the new keys and sort
-		a.signingKeys = append(a.signingKeys, ac.SigningKeys...)
-		sort.Strings(a.signingKeys)
+	for k, _ := range ac.SigningKeys {
```
kozlovic, line 2847: staticcheck flags this: replace with `for k := ..`

kozlovic (changes requested):

matthiashanel (commented):
LGTM - pending staticcheck change

kozlovic (approved):
LGTM

matthiashanel (approved):
LGTM

PR #1785, opened 2020-12-28
Fixed GW implicit reconnection

kozlovic:
PR #1412 had a fix for races during implicit GW reconnection.
However, the fix was a bit too simplistic in that it was checking
only if there was any inbound gateway to decide to try to reconnect
an implicit disconnected GW. We need to check the name, not only
presence of inbound GW connections.

Related to #1412

### DIFF
--- a/server/gateway.go
+++ b/server/gateway.go
@@ -679,7 +679,7 @@ func (s *Server) solicitGateway(cfg *gatewayCfg, firstConnect bool) {
 				s.gateway.Lock()
 				// We could have just accepted an inbound for this remote gateway.
 				// So if there is an inbound, let's try again to connect.
-				if len(s.gateway.in) > 0 {
+				if s.gateway.hasInbound(cfg.Name) {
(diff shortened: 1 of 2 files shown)


server/gateway.go
```diff
@@ -697,6 +697,20 @@ func (s *Server) solicitGateway(cfg *gatewayCfg, firstConnect bool) {
 	}
 }
 
+// Returns true if there is an inbound for the given `name`.
+// Lock held on entry.
+func (g *srvGateway) hasInbound(name string) bool {
+	for _, ig := range g.in {
+		ig.mu.Lock()
```
derekcollison, line 704: Can this be a readlock?

derekcollison (approved):
LGTM - Could not remember if those locks were RW.

server/gateway.go
kozlovic: No, `*client` has only a `sync.Mutex`

PR #1788, opened 2021-01-03
[FIXED]: syslog Warnf call Notice() instead of Warning()

paoloteti:
Use `Warning()` instead of `Notice()` and fix documentation for both syslog and Windows event log.

Signed-off-by: Paolo Teti <paolo.teti@gmail.com>

/cc @nats-io/core


ColinSullivan1 (approved):
LGTM.  Thanks for the contribution.

Looks good for windows, tested using Jetstream beta which produces a warning:
`pid[872][WARN]:          JetStream is a Beta feature`

I also double-checked that the warning messages are printed in the system log on a mac.

PR #1790, opened 2021-01-04
[Added] support for wildcard services and import remapping by JWT.

matthiashanel:
Imports in JWT where extended to contain a new filed LocalSubject.
This Change pulls the new JWT library version in.
It was needed as prefix did not exist in the JWT library and the
original field could not be used. The field To has been deprecated.

When LocalSubject is set, service imports can be configured the same way
they are in config. Meaning, no reversal due to the type.

This change also ensures that wildcard references in transforms are only
set in To/LocalSubject. Before, for services, $1 would have to be set in Subject.
This simplifies tools/parsing as only one field will have to be sanitized.


server/accounts.go
```diff
@@ -2936,17 +2941,33 @@ func (s *Server) updateAccountClaimsWithRefresh(a *Account, ac *jwt.AccountClaim
...
+				// set local subject implies to is empty
+				to = string(i.LocalSubject)
+				s.Debugf("Adding stream import %s:%q for %s:%q", acc.Name, from, a.Name, to)
+				if err := a.AddMappedStreamImportWithClaim(acc, from, to, i); err != nil {
+					s.Debugf("Error adding stream import to account [%s]: %v", a.Name, err.Error())
+					incompleteImports = append(incompleteImports, i)
+				}
+			} else {
+				s.Debugf("Adding stream import %s:%q for %s:%q", acc.Name, from, a.Name, to)
+				if err := a.AddStreamImportWithClaim(acc, from, to, i); err != nil {
```
kozlovic, line 2958: Could have made the error handling (debug/adding to incompleteImeports slice) common to the if/else cases, but that's ok.

kozlovic (approved):
LGTM (just little note about maybe reducing some code duplication, but no big deal).

server/accounts.go
matthiashanel: Fixed

matthiashanel:
@derekcollison , @kozlovic wanted to get your ok as well as this change is one that @aricart wanted to do about a year ago.
I'm implementing it this way to make jwt and file based config identical and avoid the need to sanitize different fields (this applies to tooling as well as the server)

kozlovic:
I am ok with the changes but wanted @derekcollison to have a look.

derekcollison:
ok will take a look.

server/accounts.go
```diff
@@ -2936,17 +2941,33 @@ func (s *Server) updateAccountClaimsWithRefresh(a *Account, ac *jwt.AccountClaim
 			incompleteImports = append(incompleteImports, i)
 			continue
 		}
+		from := string(i.Subject)
```
derekcollison, line 2944: This not already a string?

server/accounts.go
```diff
@@ -2936,17 +2941,33 @@ func (s *Server) updateAccountClaimsWithRefresh(a *Account, ac *jwt.AccountClaim
...
+		to := i.GetTo()
 		switch i.Type {
 		case jwt.Stream:
-			s.Debugf("Adding stream import %s:%q for %s:%q", acc.Name, i.Subject, a.Name, i.To)
-			if err := a.AddStreamImportWithClaim(acc, string(i.Subject), string(i.To), i); err != nil {
-				s.Debugf("Error adding stream import to account [%s]: %v", a.Name, err.Error())
-				incompleteImports = append(incompleteImports, i)
+			if i.LocalSubject != "" {
```
derekcollison, line 2948: Maybe use _EMPTY_ which is convention in server.

server/accounts.go
```diff
@@ -2936,17 +2941,33 @@ func (s *Server) updateAccountClaimsWithRefresh(a *Account, ac *jwt.AccountClaim
...
+				// set local subject implies to is empty
+				to = string(i.LocalSubject)
```
derekcollison, line 2950: Not already a string?

server/accounts.go
```diff
@@ -2936,17 +2941,33 @@ func (s *Server) updateAccountClaimsWithRefresh(a *Account, ac *jwt.AccountClaim
...
+					s.Debugf("Error adding stream import to account [%s]: %v", a.Name, err.Error())
+					incompleteImports = append(incompleteImports, i)
+				}
 			}
 		case jwt.Service:
 			// FIXME(dlc) - need to add in respThresh here eventually.
-			s.Debugf("Adding service import %s:%q for %s:%q", acc.Name, i.Subject, a.Name, i.To)
-			if err := a.AddServiceImportWithClaim(acc, string(i.Subject), string(i.To), i); err != nil {
+			if i.LocalSubject != "" {
+				from = string(i.LocalSubject)
```
derekcollison, line 2964: Same comment as above.

vendor/modules.txt
```diff
@@ -1,6 +1,6 @@
 # github.com/minio/highwayhash v1.0.0
 github.com/minio/highwayhash
-# github.com/nats-io/jwt/v2 v2.0.0-20201217175639-911e216824d5
+# github.com/nats-io/jwt/v2 v2.0.0-20201217175639-911e216824d5 => ../jwt/v2
```
derekcollison, line 3: That looks strange to me.

derekcollison (commented):
In general LGTM, thank you for that! Some small comments, once you comment back can do official LGTM.

vendor/modules.txt
matthiashanel: fixed

server/accounts.go
matthiashanel: no, it's coming from the jwt library where we have a subject type

server/accounts.go
matthiashanel: fixed

server/accounts.go
matthiashanel: same as for Subject

derekcollison (approved):
LGTM

kozlovic (approved):
LGTM

PR #1791, opened 2021-01-05
[fixed] two minor configuration lexer issues found by oss-fuzz

matthiashanel:
after eof isVariable had a buffer overflow
stringStateFn was not set when lexString was returned


kozlovic (approved):
LGTM

derekcollison (approved):
LGTM

PR #1792, opened 2021-01-06
[fixed] timeout in configuration lexer found by oss-fuzz

matthiashanel:
Peek followed by next resulted in an infinite loop when eof was not
checked.
I looked for all instances of this pattern and return an error when eof is not already checked or skip was used.

### DIFF
--- a/conf/parse_test.go
+++ b/conf/parse_test.go
@@ -384,6 +384,10 @@ func TestIncludeVariablesWithChecks(t *testing.T) {
 	expectKeyVal(t, m, "CAROL_PASS", "foo", 6, 3)
 }
 
-func TestParser(t *testing.T) {
-	Parse(`A@@Føøøø?˛ø:{øøøø˙˙`)
+func TestParserNoInfiniteLoop(t *testing.T) {
+	if _, err := Parse(`A@@Føøøø?˛ø:{øøøø˙˙`); err == nil {
+		t.Fatal("expected an error")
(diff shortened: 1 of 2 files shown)


conf/parse_test.go
```diff
@@ -383,3 +383,7 @@ func TestIncludeVariablesWithChecks(t *testing.T) {
 	expectKeyVal(t, m, "BOB_PASS", "$2a$11$dZM98SpGeI7dCFFGSpt.JObQcix8YHml4TBUZoge9R1uxnMIln5ly", 3, 1)
 	expectKeyVal(t, m, "CAROL_PASS", "foo", 6, 3)
 }
+
+func TestParser(t *testing.T) {
+	Parse(`A@@Føøøø?˛ø:{øøøø˙˙`)
```
kozlovic, line 388: Maybe be a bit more explicit on what this test is doing (name the test appropriately) and check error returned, assuming one is expected :-)

kozlovic (changes requested):
Note about the test.

conf/parse_test.go
kozlovic: For instance, it could be `TestNoInfiniteLoop` or something like that. And for that test, check that returned error contains something about "Unexpected EOF"?

matthiashanel: Actually that one slipped in by accident. However, I'll keep it and fix as suggested

kozlovic (approved):
LGTM

PR #1796, opened 2021-01-07
use go 1.15 for nightly builds

ripienaar:
2.2 will be built with 1.15 so preparing nightlies to match

/cc @nats-io/core


derekcollison (approved):
LGTM

PR #1797, opened 2021-01-07
Added tests for cfg/jwt based queue restrictions and updated jwt lib

matthiashanel:

### DIFF
--- a/server/jwt_test.go
+++ b/server/jwt_test.go
@@ -4923,54 +4923,59 @@ func TestJWTQueuePermissions(t *testing.T) {
 	if err != nil {
 		t.Fatalf("Received unexpected error %s", err)
 	}
+	opts.NoLog, opts.NoSigs = true, true
 	errChan := make(chan error, 1)
 	defer close(errChan)
 	s := RunServer(opts)
 	defer s.Shutdown()
 
-	for permType, tests := range map[string][][2]interface{}{
-		"allow": {{"queue.dev", false}, {"", true}, {"bad", true}},
-		"deny":  {{"", false}, {"queue.dev", true}},
+	for _, test := range []struct {
+		permType    string
+		queue       string
+		errExpected bool
+	}{
+		{"allow", "queue.dev", false},
+		{"allow", "", true},
+		{"allow", "bad", true},
+		{"deny", "", false},
+		{"deny", "queue.dev", true},
 	} {
-		for _, test := range tests {
-			queue := test[0].(string)
-			errExpected := test[1].(bool)
-			t.Run(permType+queue, func(t *testing.T) {
-				usrCreds := newUser(t, permType)
-				defer os.Remove(usrCreds)
-				nc, err := nats.Connect(fmt.Sprintf("nats://127.0.0.1:%d", opts.Port),
-					nats.ErrorHandler(func(conn *nats.Conn, s *nats.Subscription, err error) {
-						errChan <- err
-					}),
-					nats.UserCredentials(usrCreds))
-				if err != nil {
-					t.Fatalf("No error expected: %v", err)
+		t.Run(test.permType+test.queue, func(t *testing.T) {
+			usrCreds := newUser(t, test.permType)
+			defer os.Remove(usrCreds)
+			nc, err := nats.Connect(fmt.Sprintf("nats://127.0.0.1:%d", opts.Port),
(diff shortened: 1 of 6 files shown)


server/jwt_test.go
```diff
@@ -4890,3 +4890,87 @@ func TestJWTResponseThreshold(t *testing.T) {
...
+		resolver = MEMORY
+		resolver_preload = {
+			%s : %s
+		}`, ojwt, aExpPub, aExpJwt)))
+	defer os.Remove(confFileName)
+	opts, err := ProcessConfigFile(confFileName)
+	if err != nil {
+		t.Fatalf("Received unexpected error %s", err)
+	}
+	errChan := make(chan error, 1)
```
kozlovic, line 4926: Add this:
```
opts.NoLog, opts.NoSigs = true, true
```
to avoid the test printing the server's log/output.

server/jwt_test.go
```diff
@@ -4890,3 +4890,87 @@ func TestJWTResponseThreshold(t *testing.T) {
...
+	defer close(errChan)
+	s := RunServer(opts)
+	defer s.Shutdown()
+
+	for permType, tests := range map[string][][2]interface{}{
```
kozlovic, line 4931: I find this a bit difficult to read, I think this is easier:
```
for _, test := range []struct{
    permType string
    queue string
    errExpected bool
}{
  {"allow", "queue.dev", false},
  {"allow", "", true},
  {"allow", "bad", false},
  {"deny", "", false},
  {"deny", "queue.dev", true},
} {
  t.Run(test.permType+test.queue, func(t *testing.T) {
```
Then you can use test.queue and test.errExpected instead of getting from the map and doing type casting.

server/opts_test.go
```diff
@@ -2826,3 +2826,66 @@ func TestDefaultAuthTimeout(t *testing.T) {
...
+		listen: 127.0.0.1:-1
+		no_auth_user: u
+		authorization {
+			users [{ user: u, password: pwd, permissions: {
+						sub: { %s: ["foo.> *.dev"] }
+					}
+				}]}`
+	errChan := make(chan error, 1)
+	defer close(errChan)
+	for permType, tests := range map[string][][2]interface{}{
```
kozlovic, line 2841: Same comment than before.

server/opts_test.go
```diff
@@ -2826,3 +2826,66 @@ func TestDefaultAuthTimeout(t *testing.T) {
...
 	}
 }
+
+func TestQueuePermissions(t *testing.T) {
+	cfgFmt := `
+		listen: 127.0.0.1:-1
+		no_auth_user: u
+		authorization {
+			users [{ user: u, password: pwd, permissions: {
+						sub: { %s: ["foo.> *.dev"] }
```
kozlovic, line 2836: Seem that there is some tabs/spaces issues here and next line?

server/opts_test.go
```diff
@@ -2826,3 +2826,66 @@ func TestDefaultAuthTimeout(t *testing.T) {
...
+			queue := test[0].(string)
+			errExpected := test[1].(bool)
+			t.Run(permType+queue, func(t *testing.T) {
+				confFileName := createConfFile(t, []byte(fmt.Sprintf(cfgFmt, permType)))
+				defer os.Remove(confFileName)
+				opts, err := ProcessConfigFile(confFileName)
+				if err != nil {
+					t.Fatalf("Received unexpected error %s", err)
+				}
+				s := RunServer(opts)
```
kozlovic, line 2855: Add `opts.NoLog, opts.NoSig = true, true` before starting server.

kozlovic (changes requested):
Need to suppress logging, and I would suggest using []struct{} instead of map for the multiple tests.

server/jwt_test.go
matthiashanel: done

server/opts_test.go
matthiashanel: done

server/opts_test.go
matthiashanel: done

server/opts_test.go
matthiashanel: done

server/jwt_test.go
matthiashanel: done

kozlovic (approved):
LGTM

PR #1798, opened 2021-01-08
[fiixed] infinite loop in config parser for include with quote

matthiashanel:
and double quote.

similar to last, even more straight forward


derekcollison (approved):
LGTM

kozlovic (approved):
LGTM

PR #1801, opened 2021-01-12
Propose going back to condition variable to notify writeLoop

### LINKED ISSUE #1786: publish same subject with multiple clients will lead to lower performance
I have a cluster with 3 nats nodes. and I use nats-bench tool to publish and subscribe to test the performance
There are 3 subscription with nats-bench CLI:
nats-bench -np 0 -ns 1 -n 100000000 -s nats://127.0.0.1:4421 foo
nats-bench -np 0 -ns 1 -n 100000000 -s nats://127.0.0.1:4421 foo
nats-bench -np 0 -ns 1 -n 100000000 -s nats://127.0.0.1:4421 foo

And, When I publish with CLI:
nats-bench -np 1 -n 100000000 -s nats://127.0.0.1:4421 foo
I got the result: 
> Starting benchmark [msgs=100000000, msgsize=128, pubs=1, subs=0]
> Pub stats: 1,265,771 msgs/sec ~ 154.51 MB/sec

but when I increase the publish client to 2, I got lower performance
nats-bench -np 2 -n 100000000 -s nats://127.0.0.1:4421 foo
> Starting benchmark [msgs=100000000, msgsize=128, pubs=2, subs=0]
> Pub stats: 994,603 msgs/sec ~ 121.41 MB/sec
>  [1] 497,817 msgs/sec ~ 60.77 MB/sec (50000000 msgs)
>  [2] 497,306 msgs/sec ~ 60.71 MB/sec (50000000 msgs)
>  min 497,306 | avg 497,561 | max 497,817 | stddev 255 msgs

I test multiple times with different size of publish client, when the publish size is 1, the performance is the best

I read the code, I found the function client.deliverMsg have a lock operation and the granularity of the lock is really too big. Could you consider to refine the granularity of the lock.  I am not sure that is the root cause, but it is really weird, why it performs worse when increasing the publishing client size

Thanks and waiting for your response : )

kozlovic:
This is how it was up to v2.1.2 included (changed in v2.1.4 onward).
I added a benchmark that has 3 subscribers running and increase
the number of publishers: 1, 2, 5 and 10. This is the comparison
between the pre-PR and post-PR:

```
benchcmp old.txt new.txt
benchmark                           old ns/op     new ns/op     delta
Benchmark___BumpPubCount_1x3-16     396           385           -2.78%
Benchmark___BumpPubCount_2x3-16     495           406           -17.98%
Benchmark___BumpPubCount_5x3-16     542           395           -27.12%
Benchmark__BumpPubCount_10x3-16     549           515           -6.19%

benchmark                           old MB/s     new MB/s     speedup
Benchmark___BumpPubCount_1x3-16     717.27       737.54       1.03x
Benchmark___BumpPubCount_2x3-16     574.31       699.02       1.22x
Benchmark___BumpPubCount_5x3-16     524.35       718.80       1.37x
Benchmark__BumpPubCount_10x3-16     517.26       551.53       1.07x
```

It is inline with what the user reported, seeing a 20% drop in performance
when going from 1 publisher to 2. But, as we can see, the difference
between go channel and cond variable reduces with the increased number
of publishers after a certain number.

I am not sure of the performance impact on other situations, so this
PR is more of a proposal than a fix.

Resolves #1786

### DIFF
--- a/server/client.go
+++ b/server/client.go
@@ -288,7 +288,7 @@ type outbound struct {
 	pb  int64         // Total pending/queued bytes.
 	pm  int32         // Total pending/queued messages.
 	fsp int32         // Flush signals that are pending per producer from readLoop's pcd.
-	sch chan struct{} // To signal writeLoop that there is data to flush.
+	sg  *sync.Cond    // To signal writeLoop that there is data to flush.
 	wdl time.Duration // Snapshot of write deadline.
 	mp  int64         // Snapshot of max pending for client.
 	lft time.Duration // Last flush time for Write.
@@ -536,7 +536,7 @@ func (c *client) initClient() {
 
 	// Outbound data structure setup
 	c.out.sz = startBufSize
-	c.out.sch = make(chan struct{}, 1)
+	c.out.sg = sync.NewCond(&(c.mu))
 	opts := s.getOpts()
 	// Snapshots to avoid mutex access in fast paths.
 	c.out.wdl = opts.WriteDeadline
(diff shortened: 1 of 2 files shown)


derekcollison (approved):
LGTM

PR #1802, opened 2021-01-14
Fixed sublist notification

kozlovic:
The insert notification was done based on the creation of a node
during an insert, which was wrong since the node may have already
existed and still the subscription could be all new. For instance,
suppose that there is a subscription on "foo.bar".
We register an notification interest for "foo", which does not
notify, which is normal. Then we create a subscription on "foo".
During the insert, "foo" node already exists so notification would
not be sent, but it should.
Fixed also removed by having removeFromNode() returning a boolean
to indicate if the subscription was the last in that node.
However, it seems that we again check for interest in
chkForRemoveNotification(), so not sure if that is required.

### DIFF
--- a/server/sublist.go
+++ b/server/sublist.go
@@ -349,7 +349,6 @@ func (s *Sublist) Insert(sub *subscription) error {
 			}
 		}
 		if n == nil {
-			isnew = true
 			n = newNode()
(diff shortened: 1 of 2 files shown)


derekcollison (approved):
LGTM - Thanks!

PR #1803, opened 2021-01-14
JetStream Clustering

derekcollison:
This is the first pass PR at JS clustering. 

/cc @nats-io/core


server/client.go
```diff
@@ -3587,22 +3624,35 @@ func (c *client) processServiceImport(si *serviceImport, acc *Account, msg []byt
...
 	// Can't use non-locked trick like in processInboundClientMsg, so just call into selectMappedSubject
 	// so we only lock once.
 	to, _ = si.acc.selectMappedSubject(to)
 
-	// Save off some of our state.
-	oreply, oacc, opsi := c.pa.reply, c.acc, c.pa.psi
-	c.pa.reply = nrr
-	c.pa.psi = si
+	// Copy our pubArg and account
+	pacopy := c.pa
```
kozlovic, line 3633: But that is not a deep copy. Meaning that `int` etc.. will be copied, but reference to slices will be the same. So if c.pa.<any slice> change, then pacopy's slices will reflect the change. Is that ok?

derekcollison: I don't change the strings or byte slices, but can replace them in the copy of c.pa which will be put back when done.

server/gateway.go
```diff
@@ -45,7 +45,7 @@ const (
 	// hash of origin cluster name and <server> is 6 characters hash of origin server pub key.
 	gwReplyPrefix    = "_GR_."
 	gwReplyPrefixLen = len(gwReplyPrefix)
-	gwHashLen        = 6
+	gwHashLen        = sysHashLen
```
kozlovic, line 48: Just making a note here that this may break pre GWs between pre 2.2.0 and 2.2.0 servers. Not sure, will have to experiment/dig a bit more.

kozlovic: Alternatively, we could have a getHash() with a count and default would be 8 but for GW code, I would call with 6 to maintain compatibility. No other code in gateway.go has been changed it seems, so that would be safer I think.

derekcollison: Possibly, need to know if its really a problem.

kozlovic: Will check and we can always adjust if it happens to be a problem.

kozlovic (approved):
LGTM

server/jetstream_api.go
```diff
@@ -747,90 +798,110 @@ func jsNotFoundError(err error) *ApiError {
...
 	resp.StreamInfo = &StreamInfo{Created: mset.Created(), State: mset.State(), Config: mset.Config()}
-	s.sendAPIResponse(c, subject, reply, string(msg), s.jsonResponse(resp))
+	s.sendAPIResponse(ci, acc, subject, reply, string(msg), s.jsonResponse(resp))
 }
 
 // Request to update a stream.
-func (s *Server) jsStreamUpdateRequest(sub *subscription, c *client, subject, reply string, msg []byte) {
-	if c == nil || c.acc == nil {
+func (s *Server) jsStreamUpdateRequest(sub *subscription, c *client, subject, reply string, rmsg []byte) {
+	if c == nil {
```
ripienaar, line 847: shouldnt these also be `if c == nil || !s.JetStreamIsLeader() {` - few places here of that

derekcollison: I need to flesh out the rest of the API, not all implemented yet.

derekcollison: Marked here - https://github.com/nats-io/nats-server/projects/2

derekcollison: Part of https://github.com/nats-io/nats-server/pull/1819

PR #1804, opened 2021-01-14
Fixed service import cycle detection that broke with JS clustering

kozlovic:
Also added some no-op error handler for some tests to silence the
error report in the log.


derekcollison (approved):
LGTM - Thanks!

PR #1805, opened 2021-01-14
[added] enforcement and usage of scoped signing keys

matthiashanel:

I added a lookup of the scoped signing key to applyAccountLimits. 
passing in the values into the function would have made the change a lot bigger.

Currently the update strategy of scoped keys is to disconnect the clients

### DIFF
--- a/server/accounts.go
+++ b/server/accounts.go
@@ -2857,14 +2857,13 @@ func (s *Server) updateAccountClaimsWithRefresh(a *Account, ac *jwt.AccountClaim
 	}
 	if len(a.signingKeys) != len(old.signingKeys) {
 		signersChanged = true
-	} else {
-		for k, scope := range a.signingKeys {
-			if oldScope, ok := old.signingKeys[k]; !ok {
-				signersChanged = true
-			} else if !reflect.DeepEqual(scope, oldScope) {
-				signersChanged = true
-				alteredScope[k] = struct{}{}
-			}
+	}
+	for k, scope := range a.signingKeys {
+		if oldScope, ok := old.signingKeys[k]; !ok {
+			signersChanged = true
+		} else if !reflect.DeepEqual(scope, oldScope) {
+			signersChanged = true
+			alteredScope[k] = struct{}{}
 		}
 	}
 	a.mu.Unlock()
--- a/server/accounts.go
+++ b/server/accounts.go
@@ -74,7 +74,7 @@ type Account struct {
 	limits
 	expired      bool
 	incomplete   bool
-	signingKeys  []string
+	signingKeys  map[string]jwt.Scope
 	srv          *Server // server this account is registered with (possibly nil)
 	lds          string  // loop detection subject for leaf nodes
 	siReply      []byte  // service reply prefix, will form wildcard subscription.
@@ -2596,7 +2596,8 @@ func (a *Account) isIssuerClaimTrusted(claims *jwt.ActivationClaims) bool {
 		}
 		return false
 	}
-	return a.hasIssuerNoLock(claims.Issuer)
+	_, ok := a.hasIssuerNoLock(claims.Issuer)
+	return ok
 }
 
 // Returns true if `a` and `b` stream imports are the same. Note that the
@@ -2742,26 +2743,23 @@ func (a *Account) checkExpiration(claims *jwt.ClaimsData) {
 }
 
 // hasIssuer returns true if the issuer matches the account
+// If the issuer is a scoped signing key, the scope will be returned as well
 // issuer or it is a signing key for the account.
-func (a *Account) hasIssuer(issuer string) bool {
+func (a *Account) hasIssuer(issuer string) (jwt.Scope, bool) {
 	a.mu.RLock()
-	hi := a.hasIssuerNoLock(issuer)
+	scope, ok := a.hasIssuerNoLock(issuer)
 	a.mu.RUnlock()
-	return hi
+	return scope, ok
 }
 
 // hasIssuerNoLock is the unlocked version of hasIssuer
-func (a *Account) hasIssuerNoLock(issuer string) bool {
+func (a *Account) hasIssuerNoLock(issuer string) (jwt.Scope, bool) {
 	// same issuer -- keep this for safety on the calling code
 	if a.Name == issuer {
-		return true
+		return nil, true
 	}
-	for i := 0; i < len(a.signingKeys); i++ {
-		if a.signingKeys[i] == issuer {
-			return true
-		}
-	}
-	return false
+	scope, ok := a.signingKeys[issuer]
+	return scope, ok
 }
 
 // Returns the loop detection subject used for leafnodes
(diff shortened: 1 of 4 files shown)


server/accounts.go
```diff
@@ -2846,20 +2844,26 @@ func (s *Server) updateAccountClaimsWithRefresh(a *Account, ac *jwt.AccountClaim
 	// Reset any notion of export revocations.
 	a.actsRevoked = nil
 
+	alteredScope := map[string]struct{}{}
+
 	// update account signing keys
 	a.signingKeys = nil
+	if len(ac.SigningKeys) > 0 {
+		a.signingKeys = map[string]jwt.Scope{}
```
kozlovic, line 2852: Should we have a make here? Also, we could initialize the size to ac.SigningKeys?

server/accounts.go
```diff
@@ -3153,7 +3157,12 @@ func (s *Server) updateAccountClaimsWithRefresh(a *Account, ac *jwt.AccountClaim
 			c.mu.Lock()
 			sk := c.user.SigningKey
 			c.mu.Unlock()
-			if sk != "" && !a.hasIssuer(sk) {
+			if sk == "" {
```
kozlovic, line 3160: Maybe use `_EMPTY_` here, just to be consistent with most of code checking for empty string..

server/auth.go
```diff
@@ -561,9 +561,21 @@ func (s *Server) processClientOrLeafAuthentication(c *client, opts *Options) boo
 			return false
 		}
 		// this only executes IF there's an issuer on the Juc - otherwise the account is already vetted
-		if juc.IssuerAccount != "" && !acc.hasIssuer(juc.Issuer) {
-			c.Debugf("User JWT issuer is not known")
-			return false
+		if juc.IssuerAccount != "" {
```
kozlovic, line 564: _EMPTY_

server/client.go
```diff
@@ -688,6 +688,15 @@ func (c *client) applyAccountLimits() {
 		if uc, _ := jwt.DecodeUserClaims(c.opts.JWT); uc != nil {
 			c.mpay = int32(uc.Limits.Payload)
 			c.msubs = int32(uc.Limits.Subs)
+			if uc.IssuerAccount != "" && uc.IssuerAccount != uc.Issuer {
```
kozlovic, line 691: _EMPTY_

server/jwt_test.go
```diff
@@ -4563,17 +4563,27 @@ func encodeClaim(t *testing.T, claim *jwt.AccountClaims, pub string) string {
...
-func newUser(t *testing.T, accKp nkeys.KeyPair) string {
+// returns user creds
+func newUserEx(t *testing.T, accKp nkeys.KeyPair, scoped bool, issuerAccount string) string {
 	ukp, _ := nkeys.CreateUser()
 	seed, _ := ukp.Seed()
 	upub, _ := ukp.PublicKey()
 	uclaim := newJWTTestUserClaims()
 	uclaim.Subject = upub
+	uclaim.SetScoped(scoped)
+	if issuerAccount != "" {
```
kozlovic, line 4574: _EMPTY_

kozlovic (changes requested):
Maybe a change or two, otherwise, LGTM.

server/accounts.go
matthiashanel: fixed

server/accounts.go
matthiashanel: fixed

server/auth.go
matthiashanel: fixed

server/client.go
matthiashanel: fixed

server/jwt_test.go
matthiashanel: fixed

server/accounts.go
```diff
@@ -2846,20 +2844,26 @@ func (s *Server) updateAccountClaimsWithRefresh(a *Account, ac *jwt.AccountClaim
...
+
 	// update account signing keys
 	a.signingKeys = nil
+	if len(ac.SigningKeys) > 0 {
+		a.signingKeys = make(map[string]jwt.Scope, len(ac.SigningKeys))
+	}
 	signersChanged := false
-	for k := range ac.SigningKeys {
-		a.signingKeys = append(a.signingKeys, k)
+	for k, scope := range ac.SigningKeys {
```
aricart, line 2855: this doesn't look right it is simply counting - it could have the same count, but a value is changed

server/accounts.go
```diff
@@ -2846,20 +2844,26 @@ func (s *Server) updateAccountClaimsWithRefresh(a *Account, ac *jwt.AccountClaim
...
+		a.signingKeys[k] = scope
 	}
-	sort.Strings(a.signingKeys)
 	if len(a.signingKeys) != len(old.signingKeys) {
```
aricart, line 2858: this doesn't look right - the count of the keys may be the same but the keys could have changed?

server/auth.go
```diff
@@ -561,9 +561,21 @@ func (s *Server) processClientOrLeafAuthentication(c *client, opts *Options) boo
...
-		if juc.IssuerAccount != "" && !acc.hasIssuer(juc.Issuer) {
-			c.Debugf("User JWT issuer is not known")
-			return false
+		if juc.IssuerAccount != _EMPTY_ {
+			if scope, ok := acc.hasIssuer(juc.Issuer); !ok {
+				c.Debugf("User JWT issuer is not known")
+				return false
+			} else if scope != nil {
+				if err := scope.ValidateScopedSigner(juc); err != nil {
+					c.Debugf("User JWT is not valid")
```
aricart, line 570: would printing the error in the debug message be useful?

server/jwt_test.go
```diff
@@ -4497,9 +4497,9 @@ func TestJWTAccountOps(t *testing.T) {
 		t.Run("", func(t *testing.T) {
 			var syspub, sysjwt, sysCreds string
 			createAccountAndUser(&syspub, &sysjwt, &sysCreds)
+			defer os.Remove(sysCreds)
```
aricart, line 4500: should use the https://golang.org/pkg/testing/#T.TempDir

server/jwt_test.go
```diff
@@ -4979,3 +4989,132 @@ func TestJWTQueuePermissions(t *testing.T) {
 
 	}
 }
+
+func TestJWScopedSigningKeys(t *testing.T) {
+	sysKp, syspub := createKey(t)
+	sysJwt := encodeClaim(t, jwt.NewAccountClaims(syspub), syspub)
+	sysCreds := newUser(t, sysKp)
+	defer os.Remove(sysCreds)
```
aricart, line 4997: store in t.TempDir

aricart (changes requested):
A test in the signing keys is simply counting - when signing key could be added and removed before the server sees it. So the count is the same but contents are different.

server/accounts.go
kozlovic: That's what the else is doing. This basic check is to say that if count is different, there is surely something that has changed. However, since now we keep track of the one that changed, we may need to do the "else" case anyway.

server/jwt_test.go
```diff
@@ -5048,8 +5049,9 @@ func TestJWScopedSigningKeys(t *testing.T) {
 			}
 		}
 	}
-	errHdlr := nats.DisconnectErrHandler(func(conn *nats.Conn, err error) {
+	errHdlr := nats.ErrorHandler(func(conn *nats.Conn, s *nats.Subscription, err error) {
 		errChan <- err
+		t.Log(err)
```
kozlovic, line 5054: Actually we should not log things with `t.` (unless when you are debugging). This is adding noise to the test run logfile.

server/auth.go
matthiashanel: added

server/accounts.go
matthiashanel: not the same but when different we don't compare the scope. fixed.

kozlovic (changes requested):
Would remote t.Log()..

server/jwt_test.go
matthiashanel: not going to change this as this will drag though to a lot of utility functions that use ioutil.tempfile...

server/jwt_test.go
matthiashanel: same

server/jwt_test.go
matthiashanel: fixed

kozlovic (approved):
LGTM

server/accounts.go
```diff
@@ -3153,7 +3156,12 @@ func (s *Server) updateAccountClaimsWithRefresh(a *Account, ac *jwt.AccountClaim
 			c.mu.Lock()
 			sk := c.user.SigningKey
 			c.mu.Unlock()
-			if sk != "" && !a.hasIssuer(sk) {
+			if sk == _EMPTY_ {
+				continue
+			}
+			if _, ok := alteredScope[sk]; ok {
+				c.closeConnection(AuthenticationViolation)
```
aricart, line 3163: so here we are simply d/c because their configuration changed, and a reconnect will make them have the correct permissions?

aricart: this seems reasonable.

aricart (approved):
LGTM - we are simply going to d/c user's whose scope has changed since they connected. On reconnect, they will have the proper permissions.

PR #1806, opened 2021-01-14
Fixing latency sharing which was overwritten

matthiashanel:
Also adjusting unit test to not check for renamed values

### DIFF
--- a/server/accounts.go
+++ b/server/accounts.go
@@ -1736,17 +1736,13 @@ func (a *Account) addServiceImport(dest *Account, from, to string, claim *jwt.Im
 			}
 		}
 	}
-
-	share := false
+	// Turn on sharing by default if importing from system services.
+	share := isSysAcc
 	if claim != nil {
 		share = claim.Share
 	}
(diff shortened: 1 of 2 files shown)


server/accounts.go
```diff
@@ -1741,12 +1741,12 @@ func (a *Account) addServiceImport(dest *Account, from, to string, claim *jwt.Im
 	if claim != nil {
 		share = claim.Share
 	}
-
+	if !share {
```
kozlovic, line 1744: Help me understand: is it that share should be true if isSysAcc, but if there is a claim, then share becomes whatever claim.Share is?
If so, would not it be simpler to have:
```
// Turn on sharing by default if importing from system services.
share := isSysAcc
if claim != nil {
    share = claim.Share
}
si := &serviceImport{...
```

matthiashanel: the problem was that if share was true an disSysAcc was false, the earlier code would have resulted in false.
yes, will change to your suggestion

PR #1812, opened 2021-01-15
Fixed some leafnode issues introduced from JS cluster work

kozlovic:
Also fixed a flapper.

### DIFF
--- a/server/leafnode.go
+++ b/server/leafnode.go
@@ -1178,7 +1178,7 @@ type leafConnectInfo struct {
 	TLS     bool   `json:"tls_required"`
 	Comp    bool   `json:"compression,omitempty"`
 	ID      string `json:"server_id,omitempty"`
-	Name    string `json:"server_name,omitempty"`
+	Name    string `json:"name,omitempty"`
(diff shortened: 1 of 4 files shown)


server/leafnode.go
```diff
@@ -1218,7 +1229,15 @@ func (c *client) processLeafNodeConnect(s *Server, arg []byte, lang string) erro
 	c.headers = supportHeaders && proto.Headers
 
 	// Remember the remote server.
-	c.leaf.remoteServer = proto.Name
+	// We changed the leafnode CONNECT.Name json tag from "name" to "server_name",
```
kozlovic, line 1232: Note: I could remove this comment and code (just have c.leaf.remoteServer = proto.Name) if I revert the change from "name" to "server_name" that you did in JSC.

derekcollison (approved):
LGTM

server/leafnode.go
derekcollison: ok sounds good.

kozlovic: Just double checking: sounds good mean that you agree on changing back "server_name" json tag to "name"?

PR #1815, opened 2021-01-16
Routes send subscriptions by utilizing random clients from an account.

derekcollison:
There was a bug where the client chosen under the $SYS account could have a different account.

/cc @nats-io/core

### DIFF
--- a/server/accounts.go
+++ b/server/accounts.go
@@ -878,7 +878,9 @@ func (a *Account) randomClient() *client {
 	}
 	var c *client
 	for c = range a.clients {
-		break
+		if c.acc == a {
+			break
+		}
 	}
 	return c
 }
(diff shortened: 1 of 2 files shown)


kozlovic (approved):
LGTM

server/accounts.go
```diff
@@ -878,7 +878,9 @@ func (a *Account) randomClient() *client {
 	}
 	var c *client
 	for c = range a.clients {
-		break
+		if c.acc == a {
```
kozlovic, line 881: But now you may get DATA RACE reports because c.acc is sometimes changed. I wonder if we should not rework that whole sending subs through route.

derekcollison: I wanted to see if travis picked it up, I know we do things around being able to access c.acc without a lock.
Agree we should on routes, I even have a comment there about it ;)

derekcollison: I am seeing another test failure. Looking into that one now.

kozlovic (approved):
LGTM

PR #1816, opened 2021-01-16
Fixed gateway reply mapping following changes in JetStream clustering

kozlovic:
Those changes are required to maintain backward compatibility.
Since the replies are "_G_.<gateway name hash>.<server ID hash>"
and the hash were 6 characters long, changing to 8 the hash function
would break things.


derekcollison (approved):
LGTM

PR #1818, opened 2021-01-16
Stabilize restart/catchup for raft.

derekcollison:

/cc @nats-io/core


derekcollison:
Part of the other one, so closing..

PR #1819, opened 2021-01-16
Clustering updates

derekcollison:
Added in support for stream update as well as stream and consumer detailed lists which use a scatter and gather approach.

/cc @nats-io/core

### DIFF
--- a/server/consumer.go
+++ b/server/consumer.go
@@ -228,13 +228,13 @@ func (mset *Stream) AddConsumer(config *ConsumerConfig) (*Consumer, error) {
 
 func (mset *Stream) addConsumer(config *ConsumerConfig, oname string, node RaftNode) (*Consumer, error) {
 	mset.mu.RLock()
-	jsa := mset.jsa
+	s, jsa := mset.srv, mset.jsa
 	mset.mu.RUnlock()
 
 	// If we do not have the consumer assigned to us in cluster mode we can not proceed.
 	// Running in single server mode this always returns true.
 	if oname != _EMPTY_ && !jsa.consumerAssigned(mset.Name(), oname) {
-		return nil, ErrJetStreamNotAssigned
+		s.Debugf("Consumer %q > %q does not seem to be assigned to this server", mset.Name(), oname)
 	}
 
 	if config == nil {
--- a/server/consumer.go
+++ b/server/consumer.go
@@ -233,8 +233,9 @@ func (mset *Stream) addConsumer(config *ConsumerConfig, oname string, node RaftN
 	s, jsa := mset.srv, mset.jsa
 	mset.mu.RUnlock()
 
-	// If we do not have the consumer assigned to us in cluster mode we can not proceed.
-	// Running in single server mode this always returns true.
+	// If we do not have the consumer currently assigned to us in cluster mode we will proceed but warn.
+	// This can happen on startup with restored state where on meta replay we still do not have
+	// the assignment. Running in single server mode this always returns true.
 	if oname != _EMPTY_ && !jsa.consumerAssigned(mset.Name(), oname) {
 		s.Debugf("Consumer %q > %q does not seem to be assigned to this server", mset.Name(), oname)
 	}
@@ -604,16 +605,17 @@ func (o *Consumer) setLeader(isLeader bool) {
 			return
 		}
 
+		mset.mu.RLock()
+		s, jsa, stream := mset.srv, mset.jsa, mset.config.Name
+		mset.mu.RUnlock()
+
 		o.mu.Lock()
 		// Restore our saved state. During non-leader status we just update our underlying store.
 		o.readStoredState()
 
 		// Do info sub.
-		if o.infoSub == nil && mset != nil && mset.jsa != nil {
-			mset.mu.RLock()
-			s, jsa, stream := mset.srv, mset.jsa, mset.config.Name
-			mset.mu.RUnlock()
-			isubj := fmt.Sprintf("$JSC.CI.%s.%s.%s", jsa.acc(), stream, o.name)
+		if o.infoSub == nil && jsa != nil {
+			isubj := fmt.Sprintf(clusterConsumerInfoT, jsa.acc(), stream, o.name)
 			// Note below the way we subscribe here is so that we can send requests to ourselves.
 			o.infoSub, _ = s.systemSubscribe(isubj, _EMPTY_, false, o.sysc, o.handleClusterConsumerInfoRequest)
 		}
--- a/server/consumer.go
+++ b/server/consumer.go
@@ -171,6 +171,7 @@ type Consumer struct {
 	mset              *Stream
 	acc               *Account
 	client            *client
+	sysc              *client
 	sid               int
 	name              string
 	stream            string
@@ -211,7 +212,8 @@ type Consumer struct {
 	closed            bool
 
 	// Clustered.
-	node RaftNode
+	node    RaftNode
+	infoSub *subscription
 }
 
 const (
@@ -433,6 +435,7 @@ func (mset *Stream) addConsumer(config *ConsumerConfig, oname string, node RaftN
 		mset:    mset,
 		acc:     a,
 		client:  s.createInternalJetStreamClient(),
+		sysc:    s.createInternalJetStreamClient(),
 		config:  *config,
 		dsubj:   config.DeliverSubject,
 		active:  true,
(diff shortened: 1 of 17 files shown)


derekcollison:
I had these alot more stable but now seem to be flapping again. Taking a short break and then will look into getting this stable again.

server/consumer.go
```diff
@@ -228,13 +230,13 @@ func (mset *Stream) AddConsumer(config *ConsumerConfig) (*Consumer, error) {
...
 	mset.mu.RLock()
-	jsa := mset.jsa
+	s, jsa := mset.srv, mset.jsa
 	mset.mu.RUnlock()
 
 	// If we do not have the consumer assigned to us in cluster mode we can not proceed.
 	// Running in single server mode this always returns true.
 	if oname != _EMPTY_ && !jsa.consumerAssigned(mset.Name(), oname) {
-		return nil, ErrJetStreamNotAssigned
+		s.Debugf("Consumer %q > %q does not seem to be assigned to this server", mset.Name(), oname)
```
kozlovic, line 239: So now we are not returning with an error like we used to. Is that intentional? The comment above seem to indicate that if not assigned we cannot proceed, yet we do.

server/consumer.go
```diff
@@ -603,6 +608,16 @@ func (o *Consumer) setLeader(isLeader bool) {
 		// Restore our saved state. During non-leader status we just update our underlying store.
 		o.readStoredState()
 
+		// Do info sub.
+		if o.infoSub == nil && mset != nil && mset.jsa != nil {
+			mset.mu.RLock()
```
kozlovic, line 613: I don't know enough of the locking rules between stream and consumer, but here it seem we have o.mu.Lock() above and now trying to get mset's lock. Could that cause a lock inversion (again without knowing, I would think that it seem more logical to have stream -> consumer than the opposite).

(if that is an issue, would move those mset's fields capture before getting o.mu.Lock(), even if we end-up not needing them).

server/jetstream_api.go
```diff
@@ -1024,6 +1026,12 @@ func (s *Server) jsStreamListRequest(sub *subscription, c *client, subject, repl
 		offset = req.Offset
 	}
 
+	// Clustered mode will invoke a scatter and gather.
+	if s.JetStreamIsClustered() {
+		s.startGoRoutine(func() { s.jsClusteredStreamListRequest(acc, ci, offset, subject, reply, msg) })
```
kozlovic, line 1031: Don't we need to copy some of the content of those variables here? Meaning when the stack unrolls, those fields may be overwritten (say msg points to wire, etc..).

server/jetstream_api.go
```diff
@@ -1845,6 +1855,13 @@ func (s *Server) jsConsumerListRequest(sub *subscription, c *client, subject, re
 	}
 
 	streamName := streamNameFromSubject(subject)
+
+	// Clustered mode will invoke a scatter and gather.
+	if s.JetStreamIsClustered() {
+		s.startGoRoutine(func() { s.jsClusteredConsumerListRequest(acc, ci, offset, streamName, subject, reply, msg) })
```
kozlovic, line 1861: Same here.

server/jetstream_cluster.go
```diff
@@ -1001,10 +1013,10 @@ func (js *jetStream) applyStreamEntries(mset *Stream, ce *CommittedEntry) (bool,
 				if err != nil {
 					panic(err.Error())
 				}
-				if lseq == 0 && mset.lastSeq() != 0 { // Very first msg
+				// Skip by hand here since first msg special case.
```
kozlovic, line 1016: Not sure I fully understand this comment :-)

server/jetstream_cluster.go
```diff
@@ -1136,7 +1148,10 @@ func (js *jetStream) processStreamAssignment(sa *streamAssignment) {
 	accStreams[stream] = sa
 	cc.streams[acc.Name] = accStreams
 
-	isMember := sa.Group.isMember(cc.meta.ID())
+	var isMember bool
+	if sa.Group != nil && cc.meta != nil {
```
kozlovic, line 1152: If those are possible legitimate conditions (group and/or meta being nil), a small comment indicating what that means may help the reader.

server/jetstream_cluster.go
```diff
@@ -1918,6 +1936,207 @@ func (s *Server) jsClusteredStreamPurgeRequest(ci *ClientInfo, stream, subject,
...
+		case rc <- &si:
+		default:
+			s.Warnf("Failed placing stream info result on internal chan")
+		}
+	})
+	defer s.sysUnsubscribe(rsub)
+
+	// Send out our requests here.
+	for _, sa := range streams {
+		isubj := fmt.Sprintf("$JSC.SI.%s.%s", sa.Client.Account, sa.Config.Name)
```
kozlovic, line 1994: Should "$JSC.SI.." made it a constant (like the template that we have)?

server/jetstream_cluster.go
```diff
@@ -1918,6 +1936,207 @@ func (s *Server) jsClusteredStreamPurgeRequest(ci *ClientInfo, stream, subject,
...
+		case rc <- &ci:
+		default:
+			s.Warnf("Failed placing consumer info result on internal chan")
+		}
+	})
+	defer s.sysUnsubscribe(rsub)
+
+	// Send out our requests here.
+	for _, ca := range consumers {
+		isubj := fmt.Sprintf("$JSC.CI.%s.%s.%s", ca.Client.Account, stream, ca.Name)
```
kozlovic, line 2097: Same here.

server/stream.go
```diff
@@ -153,7 +155,7 @@ func (a *Account) addStream(config *StreamConfig, fsConfig *FileStoreConfig, sa
 	// If we do not have the stream assigned to us in cluster mode we can not proceed.
 	// Running in single server mode this always returns true.
 	if !jsa.streamAssigned(config.Name) {
-		return nil, ErrJetStreamNotAssigned
+		s.Debugf("Stream %q does not seem to be assigned to this server", config.Name)
```
kozlovic, line 158: Again, here the comment says that we cannot proceed but we do and just log a debug message. If it's ok, let fix the comment.

server/stream.go
```diff
@@ -297,14 +302,25 @@ func (mset *Stream) setLeader(isLeader bool) error {
 }
 
 // Lock should be held.
-func (mset *Stream) startSyncSub() {
+func (mset *Stream) startClusterSubs() {
+	if mset.infoSub == nil {
+		if jsa := mset.jsa; jsa != nil {
+			isubj := fmt.Sprintf("$JSC.SI.%s.%s", jsa.acc(), mset.config.Name)
```
kozlovic, line 308: If we make this a constant, replace the string here.

server/consumer.go
derekcollison: This is all for raciness during startup. We should warn but could be that we process a recovered stream or consumer before we process the assignment if it was the last. In normal practice does not happen and if it is truly not supposed to be there the deleteOp will follow from meta and delete it anyway.

server/consumer.go
derekcollison: Good point, will cleanup.

server/jetstream_api.go
derekcollison: Apologies, same what?

server/jetstream_cluster.go
derekcollison: ok will make clearer. sequences are unsigned, and each message we replicate has the prior sequence it expects when being placed in the stream storage. So 0 can't have prev of -1.

server/jetstream_cluster.go
derekcollison: It should not happen, if it does, at least for now, that is an error.

server/jetstream_cluster.go
derekcollison: Yes will do that.

server/jetstream_cluster.go
derekcollison: Yep.

server/stream.go
derekcollison: Yes will correct.

server/stream.go
derekcollison: Done

server/jetstream_api.go
kozlovic: Same than this comment I made that you missed: https://github.com/nats-io/nats-server/pull/1819#discussion_r559226595

server/jetstream_api.go
derekcollison: Ah I see what you mean. Will take a look more closely. Thanks.

server/events.go
```diff
@@ -1673,7 +1673,9 @@ func (s *Server) debugSubscribers(sub *subscription, c *client, subject, reply s
 		}
 		// Cleanup the WC entry.
 		s.mu.Lock()
-		delete(s.sys.replies, replySubj)
+		if s.sys != nil && s.sys.replies != nil {
+			delete(s.sys.replies, replySubj)
+		}
```
kozlovic, line 1678: Not sure it makes sense to send the reply if we did not delete from the map, but I guess it does not hurt..

derekcollison: True will fix..

kozlovic (approved):
LGTM

PR #1825, opened 2021-01-19
Simplify sendSubsToRoute()

kozlovic:
Since we were creating subs on the fly, sub.im would always be nil.
We were passing a client because it was needed in sendRouteSubOrUnSubProtos().

This PR simply fills the buffer with each account's subscriptions.
There is also no need to have subs sent from different go routine
based on some threshold. Routes are no longer subject to max pending.

Some code has been made into a function so that they can be shared
by sendSubsToRoute() and sendRouteSubOrUnSubProtos(). The function
is simply adding to given buffer the RS+/- protocol.

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -1048,14 +1048,6 @@ func (s *Server) setSystemAccount(acc *Account) error {
 	if acc.imports.services == nil {
 		acc.imports.services = make(map[string]*serviceImport)
 	}
-
-	// Create a dummy internal client for fast lookup for
-	// randomClient used in route xfer of subs. Also will
-	// be stable with account.
-	if acc.ic == nil {
-		acc.ic = s.createInternalAccountClient()
-		acc.ic.acc = acc
-	}
 	acc.mu.Unlock()
 
 	s.sys = &internal{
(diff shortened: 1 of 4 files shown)


server/server.go
```diff
@@ -1048,14 +1048,6 @@ func (s *Server) setSystemAccount(acc *Account) error {
 	if acc.imports.services == nil {
 		acc.imports.services = make(map[string]*serviceImport)
 	}
-
-	// Create a dummy internal client for fast lookup for
```
kozlovic, line 1052: Since I have removed randomClient() (not needed anymore), I removed this because you added with this comment. However, if this is still required (regardless of randomClient()), I can put that back.
(I notice that there is quite a bit of flapping, so not sure if this is due to that code being removed).

derekcollison (approved):
LGTM - Might look closely at the test failures in case we broke something subtley.

kozlovic:
@derekcollison Not sure if you saw this comment: https://github.com/nats-io/nats-server/pull/1825#discussion_r560508971

server/server.go
derekcollison: I believe I had just added that with last PR, so should be fine to remove.

PR #1826, opened 2021-01-20
Fix stop of consumer's delivery loop

kozlovic:
I noticed that some consumer go routines were left running at the end
of the test suite.
It turns out that there was a race the way the consumer's qch was closed.
Since it was closed and then set to nil, it is possible that the go
routines that are started and then try to capture o.qch would actually
get qch==nil, wich then when doing a select on that nil channel would
block forever.

So we know pass the qch to the 2 go routines loopAndGatherMsgs() and
loopAndDeliverMsgs() so that when we close the channel there is
no risk of that race happening.

I do believe that there is still something that should be looked at:
it seems that a consumer's delivery loop can now be started/stopped
many times based on leadership acquired/lost. If that is the case,
I think that the consumer should wait for previous go routine to
complete before trying to start new ones.

Also moved 3 JetStream tests to the test/norace_test.go file because
they would consumer several GB of memory when running with the -race flag.


derekcollison (approved):
LGTM - Thanks!

kozlovic:
@derekcollison Found some issues with the consumer's sendq being nil on shutdown which would hang the tests. Please check 2nd commit: https://github.com/nats-io/nats-server/pull/1826/commits/a1f011747442667c10f872dab4a9400891a1ca80

PR #1832, opened 2021-01-22
[added] support for tags and filter ping monitoring requests by tags

### LINKED ISSUE #1588: Add a tag map to the server configuration
## Feature Request

Add support for an arbitrary key/value map for configuration properties per server.

#### Use Case:

From an ops perspective it is useful to be able to associate meta-tags to an instance that would be queryable through the various mechanisms supported by the server.

The server name (which defaults to the server id) is a singular instance of why you'd want to tag an instance with more human-friendly metadata. But there are many other examples.

- Tag the server with the latitude and longitude of its data center.
- Tag the server with its cloud, region, and availability zone.
- Tag the server with its environment (e.g., test or production).

Note that we wouldn't expect this tagging to be unmanageably large, nor to change often. Perhaps limits as to number of tags and key/value sizing would be reasonable.

#### Proposed Change:

Add a string to string key/value dictionary to the nats server configuration. Read it on startup, and perhaps re-read it during reload signal processing. Emit its values in, say, `varz`, and/or make it queryable through server messages.

Anyone who needs more domain-specific tagging in a large cloud of nats servers.

#### Alternative Approaches

The configured server name could be used as a key or key fragment to some external key/value store. But that adds extra moving parts where none might be needed.

matthiashanel:
fixes #1588

original issue asks for map, but I'd rather stick to the same label format we have in other places.
For now I also refrained from sending them between server.

### DIFF
--- a/server/opts.go
+++ b/server/opts.go
@@ -255,6 +255,8 @@ type Options struct {
 	// private fields, used for testing
 	gatewaysSolicitDelay time.Duration
 	routeProto           int
+
+	Tags jwt.TagList `json:"-"`
 }
 
 // WebsocketOpts are options for websocket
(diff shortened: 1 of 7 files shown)


server/opts.go
```diff
@@ -255,6 +255,8 @@ type Options struct {
 	// private fields, used for testing
 	gatewaysSolicitDelay time.Duration
 	routeProto           int
+
+	Tags jwt.TagList `json:"-"`
```
kozlovic, line 259: Let's move this above the private fields. Also, since this is a public option, could you have a little comment? 
```
// Tags <whatever comment>
Tags jwt.TagList `json:"-"`

// private fields, used for testing
...
```

kozlovic (changes requested):
Just move Tags in the server.Options struct and add a comment to it.

server/opts.go
matthiashanel: fixed

kozlovic (approved):
LGTM

PR #1833, opened 2021-01-22
Added support for clustered account info and limit enforcement

derekcollison:

/cc @nats-io/core

### DIFF
--- a/server/jetstream.go
+++ b/server/jetstream.go
@@ -15,6 +15,7 @@ package server
 
 import (
 	"crypto/sha256"
+	"encoding/binary"
 	"encoding/hex"
 	"encoding/json"
 	"fmt"
@@ -77,13 +78,24 @@ type jsAccount struct {
 	account       *Account
 	limits        JetStreamAccountLimits
 	memReserved   int64
-	memUsed       int64
 	storeReserved int64
-	storeUsed     int64
+	memTotal      int64
+	storeTotal    int64
+	usage         jsaUsage
+	rusage        map[string]*jsaUsage
 	storeDir      string
 	streams       map[string]*Stream
 	templates     map[string]*StreamTemplate
 	store         TemplateStore
+
+	// Cluster support
+	updatesPub string
+	updatesSub *subscription
+}
+
+type jsaUsage struct {
+	mem   int64
+	store int64
 }
 
 // EnableJetStream will enable JetStream support on this server with the given configuration.
@@ -418,7 +430,6 @@ func (a *Account) EnableJetStream(limits *JetStreamAccountLimits) error {
 	if s == nil {
 		return fmt.Errorf("jetstream account not registered")
 	}
-	// FIXME(dlc) - cluster mode
 	js := s.getJetStream()
 	if js == nil {
 		return ErrJetStreamNotEnabled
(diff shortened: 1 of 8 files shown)


server/jetstream.go
```diff
@@ -829,32 +854,83 @@ func (a *Account) JetStreamEnabled() bool {
...
+	var rnode string
+	if li := strings.LastIndexByte(subject, btsep); li > 0 && li < len(subject) {
+		rnode = subject[li+1:]
+	}
+	if rnode == _EMPTY_ {
+		jsa.mu.Unlock()
+		s.Warnf("Received remote usage update with no remote node")
+		return
+	}
+	var le = binary.LittleEndian
```
kozlovic, line 876: Since this is a subscription on '.*', should this function ignores if the rnode is itself?

server/jetstream.go
```diff
@@ -829,32 +854,83 @@ func (a *Account) JetStreamEnabled() bool {
...
+	// Update the usage for this remote.
+	if usage := jsa.rusage[rnode]; usage != nil {
+		// Decrement our old values.
+		jsa.memTotal -= usage.mem
+		jsa.storeTotal -= usage.store
+		usage.mem, usage.store = memUsed, storeUsed
+	} else {
+		jsa.rusage[rnode] = &jsaUsage{memUsed, storeUsed}
+	}
+	jsa.memTotal += memUsed
```
kozlovic, line 891: If we don't exclude the current node above, are we not double accounting here with what is done in updateUsage()?

server/jetstream.go
```diff
@@ -829,32 +854,83 @@ func (a *Account) JetStreamEnabled() bool {
...
 	} else {
-		jsa.storeUsed += delta
+		jsa.usage.store += delta
+		jsa.storeTotal += delta
+	}
+	// Publish our local updates if in clustered mode.
+	if jsa.js != nil && jsa.js.cluster != nil && jsa.js.srv != nil {
+		s, b := jsa.js.srv, make([]byte, 16)
+		var le = binary.LittleEndian
+		le.PutUint64(b[0:], uint64(jsa.usage.mem))
```
kozlovic, line 912: I understand that this is called often, hence the desire to make it binary encoded, however, this is pretty limiting future changes (due to the fact that the receiving side would bail out if the size of the thing we send here were different). I wonder if we need every single updates or should do it say every second, in which case using JSON would be more extensible.

server/jetstream_cluster.go
```diff
@@ -2608,12 +2629,46 @@ func (mset *Stream) snapshot() []byte {
...
 func (mset *Stream) processClusteredInboundMsg(subject, reply string, hdr, msg []byte) error {
-	mset.mu.Lock()
 
 	// For possible error response.
 	var response []byte
-	canRespond := !mset.config.NoAck && len(reply) > 0 && mset.isLeader()
-	sendq := mset.sendq
+
+	mset.mu.RLock()
+	canRespond := !mset.config.NoAck && len(reply) > 0
```
kozlovic, line 2637: The `mst.isLeader()` condition disappeared. Is that intentional?
I understand that if all nodes can check limits, any can respond, but `canRespond` is not only used here, it is used also when processing the message. If all nodes reply, is that not wasting traffic?

server/jetstream.go
derekcollison: We won't hear ourselves, no echo by default.

server/jetstream.go
derekcollison: See above.

server/jetstream.go
derekcollison: I thought about that quite a bit. We will only ever send absolutes of usage for the account per server. We could add a version in there but not worth it imo.

Also thought about not sending everyone, and came up with something pretty complicated to speed up as you got close to the limit etc. but abandoned it for the simpler version. 

We could slow it down if need be in the future but for now I prefer the send everytime version.

server/jetstream_cluster.go
kozlovic: ok

kozlovic (approved):
LGTM

kozlovic (approved):
LGTM

PR #1835, opened 2021-01-22
Change over last to active and a duration

derekcollison:

/cc @nats-io/core


ripienaar:
LGTM

PR #1836, opened 2021-01-22
Use client version for stream and consumer extended info

derekcollison:

/cc @nats-io/core


wallyqs (approved):
LGTM

PR #1838, opened 2021-01-23
Fixes for cluster info when R=1 and consumer names lists in clustered mode.

derekcollison:

/cc @nats-io/core


wallyqs (approved):
LGTM

PR #1840, opened 2021-01-25
[ADDED] Allow bearer token as mqtt authentication method

angiglesias:

 - Allow parsing password field on MQTT Connect request as a JWT to use bearer token client authentication/authorization

### Tests added

 - Test MQTT authentication with tokens with different "allowed connection types" to ensure correct user authorization

/cc @nats-io/core


kozlovic:
@angiglesias Thank you for your contribution! I was planning on doing just that, so thanks! I will have a look later today.

kozlovic (approved):
LGTM. Thank you for your contribution! This approach will allow user to use MQTT clients with NATS operator mode.

PR #1841, opened 2021-01-25
Remove mkpasswd build tag

variadico:
Currently, mkpasswd has a build tag. The tag doesn't allow mkpasswd to
be built by default. This makes go get fail when it tries to install
mkpasswd. [Our docs](https://docs.nats.io/nats-tools/mkpasswd#installing-mkpasswd) tell people to go get mkpasswd.

Resolves https://github.com/nats-io/nats.docs/issues/183

/cc @nats-io/core


variadico:
Aah, wait. `mkpasswd` has some extra dependencies. I guess I already had them on my system that's why I was able to build.

Should we add the deps to nats-server?
Or should we move `mkpasswd` to its own repo?

kozlovic:
Humm... having even more dependencies may not be the best approach, so I wonder indeed if we should move this tool our of the nats-server repo. Do we have some util repo? Or maybe RI's `nats` tool could do that (or already does).

(If we were to move it out I believe that we would need to update nats.io docs because I believe we reference it in some places).

ripienaar:
Lets put this here instead and we can remove it from this repo https://github.com/nats-io/natscli/issues/52

ripienaar:
I added this to the cli

```
$ nats server passwd -g
Generated password: eo8EdxzoA8zV%vKceWAcgu
       bcrypt hash: $2a$11$VPExyGCSHGHHZ0uu9.ZK0.aPXUkx1N4UoF1RqkxrTfUiXGFY7o4Iq
```

```
$ nats server passwd -p W#OZwVN-UjMb8nszwvT2LQ
$2a$11$oS.ygA2Nu0miT76ydTiWNub/5dCBgvf508LhOZt7opc5wohEiPgYq
```

```
$ nats server passwd
? Enter password [? for help] **********************
? Reenter password [? for help] **********************

$2a$11$uMiop3/RsLSrMs3e1U6crurxRzRI8hYgE/Rtmii5yeBxclpG4CF8q
```

derekcollison:
That looks great! We should consider supporting argon2 as well at some point. /cc @philpennock

derekcollison:
I agree let's remove.

variadico:
Closing in favor of https://github.com/nats-io/nats-server/pull/1904

PR #1842, opened 2021-01-26
LDM Support, Raft updates, fixes

derekcollison:
Some raft changes for split vote scenarios which leads to some flappers. LDM support for raft leaders and some other fixes.

/cc @nats-io/core

### DIFF
--- a/server/raft.go
+++ b/server/raft.go
@@ -1532,9 +1532,9 @@ func (n *raft) processAppendEntry(ae *appendEntry, sub *subscription) {
 	if n.state == Candidate {
 		n.debug("Received append entry in candidate state from %q, converting to follower", ae.leader)
 		n.term = ae.term
-		n.Unlock()
+		n.vote = noVote
+		n.writeTermVote()
 		n.stepdown <- ae.leader
-		return
 	}
 
 	// Catching up state.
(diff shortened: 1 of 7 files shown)


server/raft.go
```diff
@@ -369,6 +369,31 @@ func (s *Server) lookupRaftNode(group string) RaftNode {
...
+
+	var nodes []RaftNode
+	s.rnMu.RLock()
+	if len(s.raftNodes) > 0 {
+		s.Noticef("Transferring any raft leaders")
+	}
+	for _, n := range s.raftNodes {
+		nodes = append(nodes, n)
+	}
+	s.rnMu.RUnlock()
```
kozlovic, line 385: But is it not a problem that we are releasing here? Meaning say we capture 3 nodes, but then 1 is added after that, and it happens to be the leader?

Should we also collect only leaders here? or we can't because node.Leader() cannot run under s.rnMu lock?

kozlovic (approved):
LGTM but not sure if there is a "race" of adding a new node while doing LDM.

kozlovic (approved):
LGTM

server/raft.go
derekcollison: Correct..

PR #1843, opened 2021-01-26
Fix for blocking on a channel for the raft impl, various test fixes for flappers.

derekcollison:
/cc @nats-io/core

### DIFF
--- a/test/jetstream_test.go
+++ b/test/jetstream_test.go
@@ -4365,6 +4365,8 @@ func TestJetStreamSnapshotsAPI(t *testing.T) {
 	// Now connect through a cluster server and make sure we can get things to work this way as well.
 	nc2 := clientConnectToServer(t, ls)
 	defer nc2.Close()
+	// Wait a bit for interest to propagate.
+	time.Sleep(100 * time.Millisecond)
 
 	snapshot = snapshot[:0]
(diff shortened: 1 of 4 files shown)


ripienaar (approved):
LGTM

test/jetstream_test.go
```diff
@@ -4365,6 +4365,8 @@ func TestJetStreamSnapshotsAPI(t *testing.T) {
 	// Now connect through a cluster server and make sure we can get things to work this way as well.
 	nc2 := clientConnectToServer(t, ls)
 	defer nc2.Close()
+	// Wait a bit for interest to propagate.
+	time.Sleep(100 * time.Millisecond)
```
kozlovic, line 4369: Not sure which interest you are looking for to propagate here, but keep in mind that we have helpers (maybe not in /test package?) to check/wait for interest on a literal on a server for a given account.

kozlovic (commented):
LGTM. May need to replace time.Sleep() when waiting for interest propagation with proper helper that really check that interest is present on a given server/account.

test/jetstream_test.go
derekcollison: ok will look into that. Thanks..

PR #1845, opened 2021-01-26
[added] support for StrictSigningKeyUsage and updated jwt library

matthiashanel:
This will cause the server to not trust accounts/user signed by an
identity key

To be sure I changed using len( s.trustedKeys) > 0 to a check for not nil.
To support this I moved all keys trusted into signingKeys.


server/jwt.go
```diff
@@ -132,7 +132,9 @@ func validateTrustedOperators(o *Options) error {
 		if o.TrustedKeys == nil {
 			o.TrustedKeys = make([]string, 0, 4)
 		}
-		o.TrustedKeys = append(o.TrustedKeys, opc.Issuer)
+		if !opc.StrictSigningKeyUsage {
```
kozlovic, line 135: So during a config reload, this function is invoked, but prior to that the o.TrustedKeys slice is set to nil, so this code would work. However, I am wondering what happens to s.trustedKeys, I am not sure this is handled at all. (same with the new map you are adding). I am not sure if this is supposed to be supported with a config reload, but just wondering if this is not putting things into a bad state.

kozlovic (commented):
Wonder if you thought about config reload and see if there is an issue (maybe existing prior to this change).

server/jwt.go
matthiashanel: It's my understanding that operators won't be reloaded. 
processTrustedKeys which is called in NewServer copies trusted keys from options into the server.
(I made a change to make sure to make a deep copy at this place)

What is set to nil are the old options but processTrustedKeys and thus copying them over is not called.

kozlovic: validateOptions() is called on reload. This in turn calls validateTrustedOperators(), so o.TrustedKeys will be reconstructed (because it is set to nil on reload), but s.trustedKeys is not. It may be ok if we don't support that, but wanted to make sure that this is not going to cause problems.

server/jwt.go
```diff
@@ -127,7 +127,7 @@ func validateTrustedOperators(o *Options) error {
 	}
 	// If we have operators, fill in the trusted keys.
 	// FIXME(dlc) - We had TrustedKeys before TrustedOperators. The jwt.OperatorClaims
-	// has a DidSign(). Use that longer term. For now we can expand in place.
+	// has a DidSign(). Use that longesr term. For now we can expand in place.
```
kozlovic, line 130: typo?

server/server.go
```diff
@@ -823,7 +825,11 @@ func (s *Server) processTrustedKeys() bool {
 				return false
 			}
 		}
-		s.trustedKeys = s.opts.TrustedKeys
+		tk := make([]string, len(s.opts.TrustedKeys))
```
kozlovic, line 828: If this is a copy you want to make, I believe a more idiomatic way would be:
```
s.trustedKeys = append([]string(nil), s.opts.TrustedKeys...)
```
Not sure if Go optimize and pre-allocate the right size (which it probably does), but regardless this is not a fast path (executed only once), so the more compact form is better I think.

kozlovic (changes requested):

server/jwt.go
matthiashanel: yes, fixed

server/server.go
matthiashanel: that's neat, fixed

server/jwt.go
matthiashanel: I believe this is a side effect of how the code is structured and reloading keys or the operator is not supported.

kozlovic (approved):
LGTM

PR #1849, opened 2021-01-26
Raft fixes

derekcollison:
Had some flappers determining state to be current because applied and commit were going backwards.

/cc @nats-io/core


kozlovic (approved):
LGTM

PR #1850, opened 2021-01-26
[CHANGED] Enforce max_control_line for client connections only

kozlovic:
Only check limit for CLIENT connection types, however, the check
is done for CLIENT in all conditions, not only in the case of a
split buffer as it was the case so far.

### DIFF
--- a/server/client.go
+++ b/server/client.go
@@ -544,6 +544,13 @@ func (c *client) initClient() {
 	// Snapshots to avoid mutex access in fast paths.
 	c.out.wdl = opts.WriteDeadline
 	c.out.mp = opts.MaxPending
+	// Snapshot max control line since currently can not be changed on reload and we
+	// were checking it on each call to parse. If this changes and we allow MaxControlLine
+	// to be reloaded without restart, this code will need to change.
+	c.mcl = int32(opts.MaxControlLine)
+	if c.mcl == 0 {
+		c.mcl = MAX_CONTROL_LINE_SIZE
+	}
 
 	c.subs = make(map[string]*subscription)
 	c.echo = true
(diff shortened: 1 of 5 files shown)


server/const.go
```diff
@@ -124,6 +124,9 @@ const (
 	// PROTO_SNIPPET_SIZE is the default size of proto to print on parse errors.
 	PROTO_SNIPPET_SIZE = 32
 
+	// MAX_CONTOL_LINE_SNIPPET_SIZE is the default size of proto to print on max control line errors.
+	MAX_CONTOL_LINE_SNIPPET_SIZE = 128
```
wallyqs, line 128:
```suggestion
	MAX_CONTROL_LINE_SNIPPET_SIZE = 128
```

kozlovic: Oops.. good catch.

server/client.go
```diff
@@ -544,6 +544,13 @@ func (c *client) initClient() {
 	// Snapshots to avoid mutex access in fast paths.
 	c.out.wdl = opts.WriteDeadline
 	c.out.mp = opts.MaxPending
+	// Snapshot max control line since currently can not be changed on reload and we
+	// were checking it on each call to parse. If this changes and we allow MaxControlLine
+	// to be reloaded without restart, this code will need to change.
+	c.mcl = int32(opts.MaxControlLine)
+	if c.mcl == 0 {
```
derekcollison, line 551: This is consistent for all clients that we will check yes? Seems like we should just use the singleton, maybe pass it to the parse function or since clients have a server pointer set it at server start and just access that way without locks?

derekcollison (commented):
In general looks good, wondering if we can just access via server instance versus having to put it into every client which are already in need of a diet.

derekcollison (approved):
LGTM

server/client.go
kozlovic: As we discussed, I did not add it to the client, just had to move it from where it was set (readLoop() to createClient() because of all the test that use newClientForServer() which don't start the readLoop).
We will revisit later when trying to shrink the size of the client struct.

PR #1853, opened 2021-01-27
Added in raft quorum tracking and stepdown and API responses.

derekcollison:
Also various fixes for bugs and data races.

/cc @nats-io/core

### DIFF
--- a/server/events.go
+++ b/server/events.go
@@ -417,6 +417,9 @@ func (s *Server) sendInternalMsg(sub, rply string, si *ServerInfo, msg interface
 
 // Locked version of checking if events system running. Also checks server.
 func (s *Server) eventsRunning() bool {
+	if s == nil {
+		return false
+	}
 	s.mu.Lock()
 	er := s.running && s.eventsEnabled()
 	s.mu.Unlock()
--- a/server/jetstream_api.go
+++ b/server/jetstream_api.go
@@ -476,6 +476,7 @@ var (
 	jsNoConsumerErr       = &ApiError{Code: 404, Description: "consumer not found"}
 	jsStreamMismatchErr   = &ApiError{Code: 400, Description: "stream name in subject does not match request"}
 	jsNoClusterSupportErr = &ApiError{Code: 503, Description: "not currently supported in clustered mode"}
+	jsClusterNotAvailErr  = &ApiError{Code: 503, Description: "JetStream system temporarily unavailable"}
 )
 
 // For easier handling of exports and imports.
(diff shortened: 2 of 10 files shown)


server/jetstream_api.go
```diff
@@ -836,6 +855,24 @@ func (s *Server) jsStreamCreateRequest(sub *subscription, c *client, subject, re
 	}
 
 	var resp = JSApiStreamCreateResponse{ApiResponse: ApiResponse{Type: JSApiStreamCreateResponseType}}
+
+	// Determine if we should proceed here when we are in clustered mode.
+	if s.JetStreamIsClustered() {
```
kozlovic, line 860: This seem to be an exact copy of the code you added above, and looks like you do the same again in many different places. Should it be a function that maybe returns a boolean to indicate if calling code should proceed or not?

server/events.go
```diff
@@ -417,6 +417,9 @@ func (s *Server) sendInternalMsg(sub, rply string, si *ServerInfo, msg interface
 
 // Locked version of checking if events system running. Also checks server.
 func (s *Server) eventsRunning() bool {
+	if s == nil {
```
kozlovic, line 420: Weird that we have to check for `s == nil` here.. I would instead have fixed the call stack to find out when this gets invoked with a nil server.

kozlovic (commented):
Should you make the check for availability a function to avoid so much copy/paste (unless I misread and the code is slightly different)

server/jetstream_api.go
derekcollison: I think so longer term. The response is different but I could see if I could reduce some of the repeated parts.. WDYT?

server/events.go
derekcollison: Was in a test. So nothing real per se but added it since it crashed when I did not check that the server was nil in the test and called Shutdown on it.

server/jetstream_api.go
kozlovic: I would make it a function because right now this is a lot of duplicate code. If long term one of the API need to be different then you don't call that function and have specific code. When the number of calls to this common function falls to 1 then we can remove :-)

server/events.go
kozlovic: I would fix the test to be honest. This check is happening all the time (for no good reason) but more importantly, as a reader, it says that it is "normal" in some situations to have s being nil. I am guilty of doing that too sometimes, but I really try hard not to check against nil if the object is not meant to be nil, ever.

derekcollison: I did fix test but don't mind the protection for nil receivers tbh.

server/jetstream_api.go
derekcollison: ok will do.

kozlovic: Ok, I realize that the response objects are different.. sorry about that.

kozlovic: But, they all embed ApiResponse and  jsonResponse() already works with an interface{}, so you could make that new function generic?

kozlovic: Ok, we can always do that later. The code is there now, so that's ok.

kozlovic (approved):
LGTM

PR #1855, opened 2021-01-27
[FIXED] Fixed bug to allow memory store for clustering

derekcollison:
Small bug prevented memory store backed streams and consumers. 

/cc @nats-io/core

### DIFF
--- a/server/jetstream_cluster.go
+++ b/server/jetstream_cluster.go
@@ -897,7 +897,7 @@ func (rg *raftGroup) setPreferred() {
 }
 
 // createRaftGroup is called to spin up this raft group if needed.
-func (js *jetStream) createRaftGroup(rg *raftGroup) {
+func (js *jetStream) createRaftGroup(rg *raftGroup) error {
 	js.mu.Lock()
 	defer js.mu.Unlock()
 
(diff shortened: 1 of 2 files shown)


server/jetstream_cluster.go
```diff
@@ -906,31 +906,31 @@ func (js *jetStream) createRaftGroup(rg *raftGroup) {
...
-		return
+		s.Debugf("JetStream cluster detected shutdown processing raft group: %+v", rg)
+		return errors.New("shutting down")
 	}
 
 	stateDir := path.Join(js.config.StoreDir, sysAcc.Name, defaultStoreDirName, rg.Name)
 	fs, bootstrap, err := newFileStore(
 		FileStoreConfig{StoreDir: stateDir, BlockSize: 32 * 1024 * 1024},
-		StreamConfig{Name: rg.Name, Storage: rg.Storage},
+		StreamConfig{Name: rg.Name, Storage: FileStorage},
```
kozlovic, line 929: This is `newFileStore()` so I understand why you would put FileStorage here. What is confusing is that this PR is about allowing memstore for clustering but then here we see that creating a raft group will use file store.

And looking at existing code, this is the only place where rg.Storage was used, so I am not sure I understand...

derekcollison: This is for the WAL which right now defaults to file but was trying to pick up based on the asset (stream or consumer), so when I passed in memory it failed since we always (for now) try to create fileStore for the WAL. 

I will change that here shortly but just wanted to unlock for now. Also we talked about the WAL and out of band catchup for streams running concurrently with active and being memory based so thinking through some of that before making the changes for WALs to be memory based too.

kozlovic (approved):
LGTM

PR #1857, opened 2021-01-28
Add in leader elected and lost quorum advisories.

derekcollison:
/cc @nats-io/core

### DIFF
--- a/server/jetstream_cluster.go
+++ b/server/jetstream_cluster.go
@@ -1273,7 +1273,6 @@ func (js *jetStream) processStreamLeaderChange(mset *Stream, sa *streamAssignmen
 					Time: time.Now().UTC(),
 				},
 				Stream:   stream,
-				State:    mset.State(),
 				Leader:   s.serverNameForNode(node.GroupLeader()),
 				Replicas: s.replicas(node),
 			})
@@ -1290,7 +1289,6 @@ func (js *jetStream) processStreamLeaderChange(mset *Stream, sa *streamAssignmen
 					Time: time.Now().UTC(),
 				},
 				Stream:   stream,
-				State:    mset.State(),
 				Replicas: s.replicas(node),
 			})
 		}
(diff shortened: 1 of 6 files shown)


server/jetstream_cluster.go
```diff
@@ -1241,21 +1241,67 @@ func (js *jetStream) applyStreamEntries(mset *Stream, ce *CommittedEntry) (bool,
...
+		if node := mset.raftNode(); node != nil && !node.Quorum() {
+			s.Warnf("JetStream cluster stream '%s > %s' has lost quorum, stalled.", sa.Client.Account, stream)
+			s.publishAdvisory(mset.account(), JSAdvisoryStreamQuorumLostPre+"."+stream, &JSStreamQuorumLostAdvisory{
+				TypedEvent: TypedEvent{
+					Type: JSStreamQuorumLostAdvisoryType,
+					ID:   nuid.Next(),
+					Time: time.Now().UTC(),
+				},
+				Stream:   stream,
+				State:    mset.State(),
```
ripienaar, line 1293: without consumer state i wonder if its worth having this really? I am not sure what I'll do with this info if I had it, what did you have in mind as a usecase?

ripienaar (approved):
Just a q, otherwise, LGTM

server/jetstream_cluster.go
derekcollison: What would you do with state here?

derekcollison:
FYI this approach is a bit more fragile than I want, requires old leader to not be the one lost to lose quorum. Have an idea I am going to try to make it more robust. Will update PR if it works.

server/jetstream_cluster.go
ripienaar: I can't think of anything I'd do with it :)

ripienaar (approved):
LGTM

PR #1859, opened 2021-01-29
Extended AccountInfo to track API calls and errors.

derekcollison:

/cc @nats-io/core

### DIFF
--- a/server/jetstream.go
+++ b/server/jetstream.go
@@ -1,4 +1,4 @@
-// Copyright 2019-2020 The NATS Authors
+// Copyright 2019-2021 The NATS Authors
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
 // You may obtain a copy of the License at
@@ -40,7 +40,6 @@ type JetStreamConfig struct {
 	StoreDir  string
 }
 
-// TODO(dlc) - need to track and rollup against server limits, etc.
 type JetStreamAccountLimits struct {
 	MaxMemory    int64 `json:"max_memory"`
 	MaxStore     int64 `json:"max_storage"`
@@ -50,10 +49,17 @@ type JetStreamAccountLimits struct {
 
 // JetStreamAccountStats returns current statistics about the account's JetStream usage.
 type JetStreamAccountStats struct {
-	Memory  uint64                 `json:"memory"`
-	Store   uint64                 `json:"storage"`
-	Streams int                    `json:"streams"`
-	Limits  JetStreamAccountLimits `json:"limits"`
+	Memory    uint64                 `json:"memory"`
+	Store     uint64                 `json:"storage"`
+	Streams   int                    `json:"streams"`
+	Consumers int                    `json:"consumers"`
+	API       JetStreamAPIStats      `json:"api"`
+	Limits    JetStreamAccountLimits `json:"limits"`
+}
+
+type JetStreamAPIStats struct {
+	Ok  uint64
+	Err uint64
 }
 
 // This is for internal accounting for JetStream for this server.
@@ -81,6 +87,8 @@ type jsAccount struct {
 	storeReserved int64
 	memTotal      int64
 	storeTotal    int64
+	apiOk         uint64
+	apiErr        uint64
 	usage         jsaUsage
 	rusage        map[string]*jsaUsage
 	storeDir      string
@@ -93,9 +101,12 @@ type jsAccount struct {
 	updatesSub *subscription
 }
 
+// Track general usage for this account.
 type jsaUsage struct {
 	mem   int64
 	store int64
+	api   uint64
+	err   uint64
 }
 
 // EnableJetStream will enable JetStream support on this server with the given configuration.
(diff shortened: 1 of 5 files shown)


server/jetstream.go
```diff
@@ -855,7 +878,7 @@ func (a *Account) JetStreamEnabled() bool {
 }
 
 func (jsa *jsAccount) remoteUpdateUsage(sub *subscription, c *client, subject, _ string, msg []byte) {
-	const usageSize = 16
+	const usageSize = 32
```
kozlovic, line 881: So you see what I was commenting few days ago. If it had been released, you would not be able to add those without having to deal with knowing the servers you are sending the update to (because they would fail to decode).

So maybe as a way to plan for future, the check few lines down should not be `if len(msg) != usageSize` but something like `if len(msg) < usageSize` and then a server will always be able to decode the fields that it knows about. This requires obviously that we don't change size of existing fields.

server/jetstream_api.go
```diff
@@ -555,10 +556,16 @@ func (s *Server) setJetStreamExportSubs() error {
...
 
 func (s *Server) sendAPIResponse(ci *ClientInfo, acc *Account, subject, reply, request, response string) {
+	acc.trackAPI()
 	s.sendInternalAccountMsg(nil, reply, response)
 	s.sendJetStreamAPIAuditAdvisory(ci, acc, subject, request, response)
 }
 
+func (s *Server) sendAPIErrResponse(ci *ClientInfo, acc *Account, subject, reply, request, response string) {
+	acc.trackAPIErr()
+	s.sendAPIResponse(ci, acc, subject, reply, request, response)
```
kozlovic, line 566: So an error is calling sendAPIResponse() which calls trackAPI(), which means that this will send 2 updates ok++ and err++?

kozlovic (commented):
Some notes and a question..

server/jetstream.go
```diff
@@ -787,15 +798,27 @@ func (a *Account) JetStreamUsage() JetStreamAccountStats {
 
 	var stats JetStreamAccountStats
 	if jsa != nil {
+		js := jsa.js
```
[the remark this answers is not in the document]
derekcollison: It's only ever set once on creation.

server/jetstream.go
derekcollison: Sounds good. Will update.

server/jetstream_api.go
derekcollison: Thanks, I had it as total and err and changed it but did not update this. Will fix.

server/jetstream.go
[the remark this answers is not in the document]
kozlovic: Ok, I am fine with immutable fields. (I did find places where we check for `jsa.js != nil`, which made it look like it then should be protected under the rlock).

server/jetstream_api.go
```diff
@@ -581,6 +589,32 @@ func (s *Server) getRequestInfo(c *client, raw []byte) (pci *ClientInfo, acc *Ac
...
+	}
+}
+
+func (a *Account) trackAPIErr() {
+	a.mu.RLock()
+	jsa := a.js
+	a.mu.RUnlock()
+	if jsa != nil {
+		jsa.mu.Lock()
+		jsa.usage.err++
```
kozlovic, line 611: One could argue that this is still an API request, so we could have jsa.usage.api++ too (assuming that usage.api does not mean usage.apiOk, but simply the number of api calls).

kozlovic (approved):
LGTM

server/jetstream_api.go
derekcollison: That is the way I had it. Will let @ripienaar weigh in when he wakes up.

ripienaar: yeah I think increment both here

derekcollison: So then should not be called Ok. Right now in user facing results we have..

```go
API : {
  Ok: 22,
  Err: 1
}
```

We should probably name it something else yes?

derekcollison: LMK your thoughts and I can update.

ripienaar: yeah good call, so make it Count or similar

kozlovic: If user facing of `usage.api` has the meaning of API OK, then no, I would not bump it. I was thinking that this was just the number of API calls. I was confused because I saw this for the "api OK" case:
```
jsa.usage.api++
jsa.apiOk++
```
so I thought they were different counters.

derekcollison: How about this..

```go
API : {
  Total: 22,
  Errors: 1
}
```

kozlovic: Sure.

ripienaar:
LGTM

server/jetstream_api.go
derekcollison: ok will make that change.

derekcollison: ok updated, please take a look when you have a moment.

ripienaar: LGTM

kozlovic (approved):
LGTM

PR #1860, opened 2021-01-29
[FIXED] Return no match result if subject contains empty token

kozlovic:
A subject such as `foo..bar` is invalid, but if it is published
from a connection that has disabled pedantic, then the message
is matched against subscriptions and will be delivered.

This change causes Sublist.Match() to return no result.


derekcollison (approved):
LGTM

PR #1861, opened 2021-01-29
js: consumer list request fix

wallyqs:
Noticed in the logs a few of these entries when running tests:

```
Did not receive all stream info results
```

In case there are no consumers for a stream, sends response with the empty list without waiting.

/cc @nats-io/core


derekcollison (approved):
LGTM - Thanks!

PR #1862, opened 2021-01-29
[added] field to varz output containing the operator jwt

matthiashanel:

### DIFF
--- a/server/jwt.go
+++ b/server/jwt.go
@@ -48,23 +48,23 @@ func readOperatorJWT(jwtfile string) (string, *jwt.OperatorClaims, error) {
 	}
 	defer wipeSlice(contents)
 
-	var claim string
+	var theJWT string
 	items := nscDecoratedRe.FindAllSubmatch(contents, -1)
 	if len(items) == 0 {
-		claim = string(contents)
+		theJWT = string(contents)
 	} else {
 		// First result should be the JWT.
 		// We copy here so that if the file contained a seed file too we wipe appropriately.
 		raw := items[0][1]
 		tmp := make([]byte, len(raw))
 		copy(tmp, raw)
-		claim = string(tmp)
+		theJWT = string(tmp)
 	}
-	opc, err := jwt.DecodeOperatorClaims(claim)
+	opc, err := jwt.DecodeOperatorClaims(theJWT)
 	if err != nil {
 		return "", nil, err
 	}
-	return claim, opc, nil
+	return theJWT, opc, nil
 }
 
 // Just wipe slice with 'x', for clearing contents of nkey seed file.
--- a/server/jwt.go
+++ b/server/jwt.go
@@ -32,11 +32,16 @@ const jwtPrefix = "eyJ"
 
 // ReadOperatorJWT will read a jwt file for an operator claim. This can be a decorated file.
 func ReadOperatorJWT(jwtfile string) (*jwt.OperatorClaims, error) {
+	_, claim, err := readOperatorJWT(jwtfile)
+	return claim, err
+}
+
+func readOperatorJWT(jwtfile string) (string, *jwt.OperatorClaims, error) {
 	contents, err := ioutil.ReadFile(jwtfile)
 	if err != nil {
 		// Check to see if the JWT has been inlined.
 		if !strings.HasPrefix(jwtfile, jwtPrefix) {
-			return nil, err
+			return "", nil, err
 		}
 		// We may have an inline jwt here.
 		contents = []byte(jwtfile)
@@ -57,9 +62,9 @@ func ReadOperatorJWT(jwtfile string) (*jwt.OperatorClaims, error) {
 	}
 	opc, err := jwt.DecodeOperatorClaims(claim)
 	if err != nil {
-		return nil, err
+		return "", nil, err
 	}
-	return opc, nil
+	return claim, opc, nil
 }
 
 // Just wipe slice with 'x', for clearing contents of nkey seed file.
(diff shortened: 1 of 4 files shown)


server/jwt.go
```diff
@@ -31,7 +31,7 @@ var nscDecoratedRe = regexp.MustCompile(`\s*(?:(?:[-]{3,}[^\n]*[-]{3,}\n)(.+)(?:
 const jwtPrefix = "eyJ"
 
 // ReadOperatorJWT will read a jwt file for an operator claim. This can be a decorated file.
-func ReadOperatorJWT(jwtfile string) (*jwt.OperatorClaims, error) {
+func ReadOperatorJWT(jwtfile string) (*MergedClaimAndJWT, error) {
```
kozlovic, line 34: You can't do that. This function is unfortunately public, so you would possibly break users that embed NATS Server and use this API.
If you need a function to return both the OperatorClaims and the JWT from reading the JWT file, and you think only the server right now would need it, create a private one that return both and have this function call it, but return only the OperatorClaims from it.

server/monitor.go
```diff
@@ -1281,6 +1282,13 @@ func (s *Server) createVarz(pcpu float64, rss int64) *Varz {
 		}
 		varz.LeafNode.Remotes = rlna
 	}
+	if l := len(opts.TrustedOperators); l > 0 {
```
kozlovic, line 1285: We can't change TrustedOperators signature, this is an option that exists in public release. Need to rethink your approach.

server/opts.go
```diff
@@ -225,10 +230,10 @@ type Options struct {
 	MaxTracedMsgLen int `json:"-"`
 
 	// Operating a trusted NATS server
-	TrustedKeys              []string              `json:"-"`
-	TrustedOperators         []*jwt.OperatorClaims `json:"-"`
-	AccountResolver          AccountResolver       `json:"-"`
-	AccountResolverTLSConfig *tls.Config           `json:"-"`
+	TrustedKeys              []string             `json:"-"`
+	TrustedOperators         []*MergedClaimAndJWT `json:"-"`
```
kozlovic, line 234: See previous comment. We can't do that. You would need to leave as-is and add a new option.
So looking at the code it seems that during the parsing of `operator` option, we get a list of files, then we call ReadOperatorJWT() on those to create the TrustedOperators slice. After option parsing, we would not have the list of files so we would not be able to get the jwt.
What we could do is either add a new option, deprecate this one (but keep it for ever, just at one point it may not have any content since it would be superseded by the new one), or we could have a private option (not exported) so we can build both in the configuration parsing.
But it seems that what you "expose" in the monitoring is just the JWT and so I am not sure that we need to have the option containing the merge of the two.
We can discuss all this since I may not make too much sense here :-)

kozlovic (changes requested):

server/jwt.go
matthiashanel: I went this way as I had smth in the back of my mind that embedding the server and operator mode causes an error. 
But that was probably the wrong assumption. I fixed it

server/monitor.go
matthiashanel: fixed

server/opts.go
matthiashanel: fixed

server/jwt.go
```diff
@@ -57,9 +62,9 @@ func ReadOperatorJWT(jwtfile string) (*jwt.OperatorClaims, error) {
 	}
 	opc, err := jwt.DecodeOperatorClaims(claim)
 	if err != nil {
-		return nil, err
+		return "", nil, err
 	}
-	return opc, nil
+	return claim, opc, nil
```
kozlovic, line 67: For readability (at the expense of more changes that are not really changes), I would have renamed the "claim" variable to "jwt" if this is what it is. The "opc" variable seem to be the operator claim.

server/monitor.go
```diff
@@ -992,59 +992,61 @@ func (s *Server) HandleStacksz(w http.ResponseWriter, r *http.Request) {
...
+	OutMsgs           int64                 `json:"out_msgs"`
+	InBytes           int64                 `json:"in_bytes"`
+	OutBytes          int64                 `json:"out_bytes"`
+	SlowConsumers     int64                 `json:"slow_consumers"`
+	Subscriptions     uint32                `json:"subscriptions"`
+	HTTPReqStats      map[string]uint64     `json:"http_req_stats"`
+	ConfigLoadTime    time.Time             `json:"config_load_time"`
+	Tags              jwt.TagList           `json:"tags,omitempty"`
+	OperatorJwt       []string              `json:"operator_jwt,omitempty"`
+	OperatorClaim     []*jwt.OperatorClaims `json:"operator_claim,omitempty"`
```
kozlovic, line 1049: Feels like those 2 should be plural, but not sure where to put the "s" (is this the collection of claims and jwts for all operators or a single operator)

server/opts.go
```diff
@@ -227,8 +227,9 @@ type Options struct {
 	// Operating a trusted NATS server
 	TrustedKeys              []string              `json:"-"`
 	TrustedOperators         []*jwt.OperatorClaims `json:"-"`
-	AccountResolver          AccountResolver       `json:"-"`
-	AccountResolverTLSConfig *tls.Config           `json:"-"`
+	operatorJWT              []string
```
kozlovic, line 230: I would move this one (and while at it `resolverPreloads`) down after the comment saying that here are the private fields.

kozlovic (changes requested):
Maybe move the private options down and not sure if we should use plural or not.

server/jwt.go
matthiashanel: fixed

server/opts.go
matthiashanel: fixed

kozlovic (approved):
LGTM

PR #1864, opened 2021-01-29
added test demonstrating how js breaks imports

matthiashanel:

All that is necessary is to enable jetstream on the server.
When enabled imports start to break, even if the accounts do not use js themselves.
This is not just happening for the last case. 
It happens for every subtest, I just added it to the last one.


derekcollison:
Have fix PR off of this branch, so closing.

PR #1868, opened 2021-01-30
[FIXED] Fix for #1864

derekcollison:
When trying to make sure we properly created all subs for service imports we would check the internal client to see if we should process. With JS enabled on the server we would place system imports that would break that check and orphan other service imports.

Resolves #1864 

/cc @nats-io/core


kozlovic (approved):
LGTM

kozlovic (approved):
LGTM

matthiashanel (commented):
LGTM

PR #1869, opened 2021-01-30
[FIXED] Bug on single replica restore would deliver duplicate messages.

derekcollison:
Fixed a bug that on restore of single streams in clustered mode would subscribe to the stream subjects twice.

/cc @nats-io/core

### DIFF
--- a/server/stream.go
+++ b/server/stream.go
@@ -122,6 +122,7 @@ type Stream struct {
 	ddindex   int
 	ddtmr     *time.Timer
 	qch       chan struct{}
+	active    bool
 
 	// Clustered mode.
 	sa      *streamAssignment
@@ -266,7 +267,8 @@ func (a *Account) addStream(config *StreamConfig, fsConfig *FileStoreConfig, sa 
 	}
 
 	// Call directly to set leader if not in clustered mode.
-	if !s.JetStreamIsClustered() {
+	// This can be called though before we actually setup clustering, so check both.
+	if !s.JetStreamIsClustered() && s.standAloneMode() {
 		if err := mset.setLeader(true); err != nil {
 			mset.Delete()
 			return nil, err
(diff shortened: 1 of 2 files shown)


server/stream.go
```diff
@@ -854,6 +860,7 @@ func (mset *Stream) unsubscribeToStream() error {
 	for _, subject := range mset.config.Subjects {
 		mset.unsubscribeInternal(subject)
 	}
+	mset.active = true
```
kozlovic, line 863: Should it be false here???

test/jetstream_cluster_test.go
```diff
@@ -2612,6 +2612,53 @@ func TestJetStreamRestartAdvisories(t *testing.T) {
...
+
+	sub, err := js.SubscribeSync("foo", nats.Durable("dlc"))
+	if err != nil {
+		t.Fatalf("Unexpected error: %v", err)
+	}
+
+	js.Publish("foo", []byte("msg1"))
+	if _, err := sub.NextMsg(time.Second); err != nil {
+		t.Fatalf("Unexpected error: %v", err)
+	}
```
kozlovic, line 2638: The message is not ack'ed and therefore could be redelivered. Should we ack it and wait for ack processing before continuing?

test/jetstream_cluster_test.go
```diff
@@ -2612,6 +2612,53 @@ func TestJetStreamRestartAdvisories(t *testing.T) {
...
+
+	// Send second msg
+	js.Publish("foo", []byte("msg2"))
+	msg, err := sub.NextMsg(time.Second)
+	if err != nil {
+		t.Fatalf("Error getting message: %v", err)
+	}
+	if string(msg.Data) != "msg2" {
+		t.Fatalf("Unexpected message: %s", msg.Data)
+	}
```
kozlovic, line 2653: Same here, we should ack here?

test/jetstream_cluster_test.go
derekcollison: I had that in the test but default will never have us see it in this test so removed. Could add back though..

test/jetstream_cluster_test.go
derekcollison: Fixed. Thanks.

test/jetstream_cluster_test.go
derekcollison: ok added Akc() back.

kozlovic (approved):
LGTM

PR #1870, opened 2021-01-30
Fixed bug on raft inline catchup when apply channel was full.

derekcollison:

/cc @nats-io/core

### DIFF
--- a/server/raft.go
+++ b/server/raft.go
@@ -204,6 +204,8 @@ var (
 	errCorruptPeers    = errors.New("raft: corrupt peer state")
 	errStepdownFailed  = errors.New("raft: stepdown failed")
 	errPeersNotCurrent = errors.New("raft: all peers are not current")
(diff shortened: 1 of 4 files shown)


server/raft.go
```diff
@@ -1833,6 +1848,11 @@ func (n *raft) storeToWAL(ae *appendEntry) error {
 		return err
 	}
 
+	// Sanity checking for now.
+	if ae.pindex != seq-1 {
+		panic(fmt.Sprintf("[%s] Placed an entry at the wrong index, ae is %+v, index is %d\n\n", n.s, ae, seq))
```
ripienaar, line 1853: surely not panic? are we recovering this higher up?

ripienaar (approved):
LGTM, just wondering about the panic

derekcollison:
Diagnostic for now as we continue to find all the issues. This way if someone runs into it they can report it. I also saw this with the consumer bug we are chasing..

PR #1871, opened 2021-01-31
Make sure to suppress duplicate create/delete audit events

derekcollison:

/cc @nats-io/core


ripienaar (approved):
LGTM

PR #1874, opened 2021-02-01
[Adding] support for account_token_position

matthiashanel:

This was added to support ngs.

This change does 4 things.
Refactor to only have one function to validate imports.
Have This function support the jwt field account_token_position.
For completeness make this value configurable as well. 
unit tests

### DIFF
--- a/go.sum
+++ b/go.sum
@@ -3,9 +3,11 @@ github.com/golang/protobuf v1.4.0-rc.1.0.20200221234624-67d41d38c208/go.mod h1:x
 github.com/golang/protobuf v1.4.0-rc.2/go.mod h1:LlEzMj4AhA7rCAGe4KMBDvJI+AwstrUpVNzEA03Pprs=
 github.com/golang/protobuf v1.4.0-rc.4.0.20200313231945-b860323f09d0/go.mod h1:WU3c8KckQ9AFe+yFwt9sWVRKCVIyN9cPHBJSNnbL67w=
 github.com/golang/protobuf v1.4.0/go.mod h1:jodUvKwWbYaEsadDk5Fwe5c77LiNKVO9IDvqG2KuDX0=
+github.com/golang/protobuf v1.4.2 h1:+Z5KGCizgyZCbGh1KZqA0fcLLkwbsjIzS4aV2v7wJX0=
 github.com/golang/protobuf v1.4.2/go.mod h1:oDoupMAO8OvCJWAcko0GGGIgR6R6ocIYbsSw735rRwI=
 github.com/google/go-cmp v0.3.0/go.mod h1:8QqcDgzrUqlUb/G2PQTWiueGozuR1884gddMywk6iLU=
 github.com/google/go-cmp v0.3.1/go.mod h1:8QqcDgzrUqlUb/G2PQTWiueGozuR1884gddMywk6iLU=
+github.com/google/go-cmp v0.4.0 h1:xsAVV57WRhGj6kEIi8ReJzQlHHqcBYCElAvkovg3B/4=
 github.com/google/go-cmp v0.4.0/go.mod h1:v8dTdLbMG2kIc/vJvl+f65V22dbkXbowE6jgT/gNBxE=
 github.com/klauspost/compress v1.11.7 h1:0hzRabrMN4tSTvMfnL3SCv1ZGeAP23ynzodBgaHeMeg=
 github.com/klauspost/compress v1.11.7/go.mod h1:aoV0uJVorq1K+umq18yTdKaF57EivdYsUV+/s2qKfXs=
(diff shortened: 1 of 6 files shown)


aricart (approved):
LGTM - this will simplify the securing of imports without requiring a token.

server/accounts.go
```diff
@@ -895,33 +931,28 @@ func (a *Account) AddServiceExportWithResponse(subject string, respType ServiceR
 		a.exports.services = make(map[string]*serviceExport)
 	}
 
-	se := a.exports.services[subject]
+	ea := a.exports.services[subject]
```
kozlovic, line 934: Not sure why you felt the need to change `se` for service export to `ea`?

server/accounts.go
```diff
@@ -2280,8 +2311,16 @@ func (a *Account) AddStreamImport(account *Account, from, prefix string) error {
...
+// it will signify a public export, meaning anyone can import.
 func (a *Account) AddStreamExport(subject string, accounts []*Account) error {
+	return a.AddStreamExportWithAccountPos(subject, accounts, 0)
+}
+
+// AddStreamExport will add an export to the account. If accounts is nil
+// it will signify a public export, meaning anyone can import.
+// if accountPos is > 0, all imports will be granted where the following holds:
+// strings.Split(subject, ".")[accountPos] == account id will be granted.
+func (a *Account) AddStreamExportWithAccountPos(subject string, accounts []*Account, accountPos uint) error {
```
kozlovic, line 2323: Again, would keep private until we need to make this public.

server/accounts.go
```diff
@@ -2329,11 +2360,17 @@ func (a *Account) checkStreamImportAuthorizedNoLock(account *Account, subject st
...
 	// if ea is nil or ea.approved is nil, that denotes a public export
-	if ea == nil || (ea.approved == nil && !ea.tokenReq) {
+	if ea == nil || (ea.approved == nil && !ea.tokenReq && ea.accountPos == 0) {
 		return true
 	}
+	// Check if the export is protected and enforces presence of importing account identity
+	if ea.accountPos > 0 && ea.accountPos <= uint(len(tokens)) {
+		if tokens[ea.accountPos-1] == account.Name {
+			return true
+		}
```
kozlovic, line 2372: Is it possible that `ea.approved == nil` and `ea.tokenReq == false` and `ea.accountPos != 0` but not found in tokens? Code will not panic because Go does not when doing `m[key]` on a nil map, but is that not risky? Should we return false here?

server/accounts.go
```diff
@@ -2362,44 +2400,33 @@ func (a *Account) checkStreamExportApproved(account *Account, subject string, im
...
+			return a.checkAuth(&ea.exportAuth, account, imClaim, tokens)
 		}
 	}
 	return false
 }
 
 func (a *Account) checkServiceExportApproved(account *Account, subject string, imClaim *jwt.Import) bool {
 	// Check direct match of subject first
-	se, ok := a.exports.services[subject]
+	ea, ok := a.exports.services[subject]
```
kozlovic, line 2411: Again, would have kept `se` for service export. It just happens that serviceExport embeds an `exportAuth` struct, but really here you get the service export, not the exportAuth.

server/accounts.go
```diff
@@ -2868,7 +2895,8 @@ func (s *Server) updateAccountClaimsWithRefresh(a *Account, ac *jwt.AccountClaim
 		switch e.Type {
 		case jwt.Stream:
 			s.Debugf("Adding stream export %q for %s", e.Subject, a.Name)
-			if err := a.AddStreamExport(string(e.Subject), authAccounts(e.TokenReq)); err != nil {
+			if err := a.AddStreamExportWithAccountPos(
```
kozlovic, line 2898: Would change to private if you agree that it does not have to be exported at this point.

kozlovic (changes requested):
Some comments. If the new API does not need to be public, my comment on using private name applies everywhere where the new API is referenced. I did not comment in all occurrences.

server/accounts.go
```diff
@@ -877,13 +878,48 @@ func (a *Account) removeClient(c *client) int {
 	return n
 }
 
+func setExportAuth(ea *exportAuth, subject string, accounts []*Account, accountPos uint) error {
+	if accountPos > 0 {
+		if token := strings.Split(subject, "."); len(token) < int(accountPos) {
```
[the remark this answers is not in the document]
matthiashanel: fixed

server/accounts.go
```diff
@@ -877,13 +878,48 @@ func (a *Account) removeClient(c *client) int {
...
+}
+
 // AddServiceExport will configure the account with the defined export.
 func (a *Account) AddServiceExport(subject string, accounts []*Account) error {
-	return a.AddServiceExportWithResponse(subject, Singleton, accounts)
+	return a.AddServiceExportWithResponseAndAccountPos(subject, Singleton, accounts, 0)
+}
+
+// AddServiceExport will configure the account with the defined export.
+func (a *Account) AddServiceExportWithAccountPos(subject string, accounts []*Account, accountPos uint) error {
```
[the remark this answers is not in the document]
matthiashanel: fixed, same for addStreamExportWithAccountPos

server/accounts.go
```diff
@@ -877,13 +878,48 @@ func (a *Account) removeClient(c *client) int {
...
+	return a.AddServiceExportWithResponseAndAccountPos(subject, Singleton, accounts, accountPos)
 }
 
 // AddServiceExportWithResponse will configure the account with the defined export and response type.
 func (a *Account) AddServiceExportWithResponse(subject string, respType ServiceRespType, accounts []*Account) error {
+	return a.AddServiceExportWithResponseAndAccountPos(subject, respType, accounts, 0)
+}
+
+// AddServiceExportWithresponse will configure the account with the defined export and response type.
+func (a *Account) AddServiceExportWithResponseAndAccountPos(
```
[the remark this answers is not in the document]
matthiashanel: fixed

server/accounts.go
matthiashanel: fixed

server/accounts.go
matthiashanel: fixed

server/accounts.go
matthiashanel: Tbh, I'm not sure as this was done a long time ago. altered it back.

kozlovic: `ea` is for `exportAuth` I assume, but see my other comment. This is still a serviceExport...

server/accounts.go
matthiashanel: fixed. 
I probably renamed to ea (which is what checkStreamExportApproved named it) to see how similar they are.

server/accounts.go
```diff
@@ -877,13 +878,48 @@ func (a *Account) removeClient(c *client) int {
...
+	if accountPos > 0 {
+		if token := strings.Split(subject, "."); len(token) < int(accountPos) {
+			return ErrInvalidSubject
+		} else if token[accountPos-1] != "*" {
+			return ErrInvalidSubject
+		}
+	}
+	ea.accountPos = accountPos
+	// empty means auth required but will be import token.
+	if accounts == nil {
```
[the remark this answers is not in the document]
matthiashanel: fixed

server/accounts.go
matthiashanel: Agreed, will return false

go.sum
```diff
@@ -3,9 +3,11 @@ github.com/golang/protobuf v1.4.0-rc.1.0.20200221234624-67d41d38c208/go.mod h1:x
 github.com/golang/protobuf v1.4.0-rc.2/go.mod h1:LlEzMj4AhA7rCAGe4KMBDvJI+AwstrUpVNzEA03Pprs=
 github.com/golang/protobuf v1.4.0-rc.4.0.20200313231945-b860323f09d0/go.mod h1:WU3c8KckQ9AFe+yFwt9sWVRKCVIyN9cPHBJSNnbL67w=
 github.com/golang/protobuf v1.4.0/go.mod h1:jodUvKwWbYaEsadDk5Fwe5c77LiNKVO9IDvqG2KuDX0=
+github.com/golang/protobuf v1.4.2 h1:+Z5KGCizgyZCbGh1KZqA0fcLLkwbsjIzS4aV2v7wJX0=
```
kozlovic, line 6: By the way, any reason for changes in the go.sum? Could you make sure that you have rebased from master and that these changes are required?

matthiashanel: Maybe missing in an earlier commit?

```
> git co account_protected_export
Switched to branch 'account_protected_export'
> git rebase origin/master
Current branch account_protected_export is up to date.
> go mod vendor
> go mod tidy
> git status
On branch account_protected_export
Untracked files:
  (use "git add <file>..." to include in what will be committed)
	mytest
	server.test
nothing added to commit but untracked files present (use "git add" to track)
>

kozlovic (approved):
LGTM

PR #1875, opened 2021-02-02
Removing peers from streams and consumers and stepdown functionality

derekcollison:
Added in JS disable per server on reload. This will cause all peer allocations from this server to be re-assigned.
Added ability to disable JS for a server but have the server properly route JS traffic.
Fixed ability to reload debug setting and have raft nodes react.
Added API to allow removing a peer from a stream (and hence the consumers as well)
Add stepdown api for streams and consumers which will cause leader election to happen. 

Various bug fixes, stability improvements.

/cc @nats-io/core

### DIFF
--- a/server/jetstream_api.go
+++ b/server/jetstream_api.go
@@ -1719,7 +1719,7 @@ func (s *Server) jsStreamDeleteRequest(sub *subscription, c *client, subject, re
 
 	// Clustered.
 	if s.JetStreamIsClustered() {
-		s.jsClusteredStreamDeleteRequest(ci, stream, reply, msg)
+		s.jsClusteredStreamDeleteRequest(ci, stream, subject, reply, msg)
 		return
 	}
 
--- a/server/jetstream_api.go
+++ b/server/jetstream_api.go
@@ -142,6 +142,21 @@ const (
 	jsSnapshotAckT    = "$JS.SNAPSHOT.ACK.%s.%s"
 	jsRestoreDeliverT = "$JS.SNAPSHOT.RESTORE.%s.%s"
 
+	// JSApiStreamRemovePeer is the endpoint to remove a peer from a clustered stream and its consumers.
+	// Will return JSON response.
+	JSApiStreamRemovePeer  = "$JS.API.STREAM.PEER.REMOVE.*"
+	JSApiStreamRemovePeerT = "$JS.API.STREAM.PEER.REMOVE.%s"
+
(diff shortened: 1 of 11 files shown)


server/jetstream_api.go
```diff
@@ -1307,7 +1362,297 @@ func (s *Server) jsStreamInfoRequest(sub *subscription, c *client, subject, repl
...
+	if !acc.JetStreamEnabled() {
+		resp.Error = jsNotEnabledErr
+		s.sendAPIErrResponse(ci, acc, subject, reply, string(msg), s.jsonResponse(&resp))
+		return
+	}
+	if !isEmptyRequest(msg) {
+		resp.Error = jsBadRequestErr
+		s.sendAPIErrResponse(ci, acc, subject, reply, string(msg), s.jsonResponse(&resp))
+		return
+	}
```
ripienaar, line 1428: these 2 checks can probably move up above 1384 right? Avoids some read locks and such for bad requests, similar in the consumer one

server/jetstream_api.go
```diff
@@ -1307,7 +1362,297 @@ func (s *Server) jsStreamInfoRequest(sub *subscription, c *client, subject, repl
...
+	if js.isGroupLeaderless(sa.Group) {
+		resp.Error = jsClusterNotAvailErr
+		s.sendAPIErrResponse(ci, acc, subject, reply, string(msg), s.jsonResponse(&resp))
+		return
+	}
+
+	// We have the stream assigned and a leader, so only the stream leader should answer.
+	if !acc.JetStreamIsStreamLeader(name) {
+		return
+	}
```
ripienaar, line 1440: Can also be done earlier I think

ripienaar (approved):
So verbose writing these wow, but it seems fine, I might be missing some subtelties so 2 comments about failing faster for invalid requests - smaller DOS surface area by doing so, fewer locks etc

server/jetstream_api.go
derekcollison: Depends. With these its not as clear but since all servers running JS are receiving the high level API calls we need to determine when as a given server we should be elected to respond. Some are for the metacontroller leader, some are for asset leader (streams and consumers) to respond to. Some calls have both gates working depending on the failure.

server/jetstream_api.go
derekcollison: Some errors we want to report but before we assess if we are the stream leader in case there is none.

kozlovic (approved):
LGTM

PR #1880, opened 2021-02-03
[FIXED] Bug with erased messages and re-indexing the message block cache.

derekcollison:
We needed to clear the erase bit to properly set fseq in the cache structure on re-indexing.

/cc @nats-io/core

### DIFF
--- a/server/jetstream_api.go
+++ b/server/jetstream_api.go
@@ -333,7 +333,8 @@ const JSApiStreamUpdateResponseType = "io.nats.jetstream.api.v1.stream_update_re
 
 // JSApiMsgDeleteRequest delete message request.
 type JSApiMsgDeleteRequest struct {
-	Seq uint64 `json:"seq"`
+	Seq     uint64 `json:"seq"`
+	NoErase bool   `json:"no_erase,omitempty"`
 }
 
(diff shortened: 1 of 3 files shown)


kozlovic (approved):
LGTM

server/jetstream_api.go
```diff
@@ -333,7 +333,8 @@ const JSApiStreamUpdateResponseType = "io.nats.jetstream.api.v1.stream_update_re
 
 // JSApiMsgDeleteRequest delete message request.
 type JSApiMsgDeleteRequest struct {
-	Seq uint64 `json:"seq"`
+	Seq     uint64 `json:"seq"`
+	NoErase bool   `json:"no_erase, omitempty"`
```
kozlovic, line 337: is space ok in `"no_erase, omitempty"`?

kozlovic (approved):
LGTM. Good that we can select erase vs simple delete!

server/jetstream_api.go
derekcollison: Will fix.

derekcollison: ok force pushed the fix.

PR #1882, opened 2021-02-04
Stream Message delete and purge

derekcollison:
We did not handle message delete or stream purge where R=1 properly. We would also respond during recovery.

/cc @nats-io/core


kozlovic (approved):
LGTM

PR #1883, opened 2021-02-04
First pass at supercluster enablement.

derekcollison:
This allows metacontrollers to span superclusters. Also includes placement directives for streams. By default they select the request origin cluster.

/cc @nats-io/core

### DIFF
--- a/server/jetstream_cluster.go
+++ b/server/jetstream_cluster.go
@@ -46,6 +46,12 @@ type jetStreamCluster struct {
 	consumerResults *subscription
 }
 
+// Used to guide placement of streams in clustered JetStream.
+type Placement struct {
+	Cluster string   `json:"cluster"`
+	Tags    []string `json:"tags,omitempty"`
+}
+
 // Define types of the entry.
 type entryOp uint8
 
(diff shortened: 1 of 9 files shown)


kozlovic (approved):
LGTM

server/jetstream_cluster.go
```diff
@@ -46,6 +46,12 @@ type jetStreamCluster struct {
 	consumerResults *subscription
 }
 
+// Used to guide placement of streams in clustered JetStream.
+type Placement struct {
+	Cluster string   `json:"cluster"`
+	Tags    []string `json:"tags,omitempty"`
```
ripienaar, line 52: cluster not having omitempty, does this imply that cluster is required? seems empty is valid so just checking

derekcollison: We could make optional, for now it is required.

ripienaar: hmm, appears to me we do handle it being empty, which looks like it then does random assignment? That would be what I'd expect anyway

derekcollison: Yes if its empty we pick point of origin. What I meant was if you are filling this in you should have cluster for now.. But I can make optional or you can submit quick PR for omitempty addition.

ripienaar: can wait till we do tag matching and all :)

PR #1884, opened 2021-02-04
Fixed some issues with JS consumers

kozlovic:
- Should check if there is interest at top of deleteNotActive timer
callback.
- Use RLock/RUnlock for no interest check
- Remove some checks in cluster mode regarding internal stream/consumers


derekcollison (approved):
LGTM

PR #1886, opened 2021-02-04
Purge operation replay on restart

derekcollison:
Purge operations would be replayed on restart regardless if they had already been processed.

/cc @nats-io/core

### DIFF
--- a/server/jetstream_cluster.go
+++ b/server/jetstream_cluster.go
@@ -119,6 +119,7 @@ type consumerAssignment struct {
 type streamPurge struct {
 	Client  *ClientInfo `json:"client,omitempty"`
 	Stream  string      `json:"stream"`
+	LastSeq uint64      `json:"last_seq"`
 	Subject string      `json:"subject"`
 	Reply   string      `json:"reply"`
 }
@@ -1342,7 +1343,9 @@ func (js *jetStream) applyStreamEntries(mset *Stream, ce *CommittedEntry, isReco
 					continue
 				}
 				if err := mset.processJetStreamMsg(subject, reply, hdr, msg, lseq, ts); err != nil {
-					js.srv.Debugf("Got error processing JetStream msg: %v", err)
+					if err != errLastSeqMismatch || !isRecovering {
+						js.srv.Debugf("Got error processing JetStream msg: %v", err)
+					}
 				}
 			case deleteMsgOp:
 				md, err := decodeMsgDelete(buf[1:])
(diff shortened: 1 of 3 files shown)


ripienaar:
looks good

server/jetstream_cluster.go
```diff
@@ -1383,6 +1386,11 @@ func (js *jetStream) applyStreamEntries(mset *Stream, ce *CommittedEntry, isReco
 				if err != nil {
 					panic(err.Error())
 				}
+				// Ignore if we are recovering and we have already processed.
+				if isRecovering && mset.State().LastSeq > sp.LastSeq {
```
kozlovic, line 1390: I think this is right, however, say that purge was done at a time that stream had messages 1..100, and now stream is at seq 150. Then on replay there would be no purge. But it is possible that this node did not play the logs in the first place so it would add messages 1..150. I thought you had a purge(seq) that would have purged messages up to that point. I think you need that.
Even with a purge(seq) you should still verify that the mset's state *FirstSeq* is above the sp.LastSeq to completely ignore.

kozlovic (changes requested):
Better than before where that would blow incorrectly the stream on replay, but I think you are still missing the cases where some messages should be purged...

server/jetstream_cluster.go
kozlovic: In other words, you still should ensure that messages up to sp.LastSeq are no longer in the mset.

derekcollison: ok will take a different look at it..

derekcollison: ok will compact regardless which will be a no-op if was successful first time processing.

server/jetstream_cluster.go
```diff
@@ -1383,6 +1386,13 @@ func (js *jetStream) applyStreamEntries(mset *Stream, ce *CommittedEntry, isReco
...
+					// Make sure all messages from the purge are gone.
+					mset.store.Compact(sp.LastSeq)
```
kozlovic, line 1392: So I read that Compact() removes "up to but not including the seq parameter".
Is `LastSeq` in the purge request supposed to be included in the purge or is it the sequence in the stream that is supposed to be the new first after the purge? If the former, I wonder if it should not be `Compact(sp.LastSeq+1)` here then?

derekcollison: Yes good point, will fix.

server/jetstream_cluster.go
derekcollison: ok I think I have it correct now.

kozlovic (approved):
LGTM

PR #1887, opened 2021-02-04
Noticed stream snapshots were being processed on recovery

derekcollison:
This would cause high cpu usage and memory usage.
Also made consumer Created() UTC.

/cc @nats-io/core


kozlovic (approved):
LGTM

PR #1888, opened 2021-02-05
Make direct message retrieval from stream cluster aware.

derekcollison:

/cc @nats-io/core

### DIFF
--- a/test/jetstream_cluster_test.go
+++ b/test/jetstream_cluster_test.go
@@ -3216,6 +3216,45 @@ func TestJetStreamClusterPurgeReplayAfterRestart(t *testing.T) {
 	}
 }
 
+func TestJetStreamClusterStreamGetMsg(t *testing.T) {
+	c := createJetStreamClusterExplicit(t, "R3F", 3)
+	defer c.shutdown()
+
+	// Client based API
+	s := c.randomServer()
+	nc, js := jsClientConnect(t, s)
+	defer nc.Close()
+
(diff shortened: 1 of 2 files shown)


ripienaar (approved):
LGTM

test/jetstream_cluster_test.go
```diff
@@ -3216,6 +3216,45 @@ func TestJetStreamClusterPurgeReplayAfterRestart(t *testing.T) {
 	}
 }
 
+func TestJetStreamClusterStreamGetMsg(t *testing.T) {
+	c := createJetStreamClusterExplicit(t, "R3F", 3)
+	defer c.shutdown()
+
+	// Client based API
+	s := c.randomServer()
```
kozlovic, line 3224: Actually, for this test (and probably others), I would have made sure that the client connects to the non stream leader. After all, without the changes you've made, if the client happened to connect to the stream leader, this test would have passed.
So this test (and again, likely all others) should have a t.Run() with 2 cases, connecting to leader and to a follower. I guess we should add this task to the long TODO list :-)

derekcollison: Before fix would fail with count as little as 5, but I did 50.. ;)

kozlovic: I understand, my point is that when we run the suite on travis, this test is executed only once. So if it happens that the random pick is the leader, then the test would pass, even without the fix for cluster-aware-lookup.

derekcollison: Understood.

PR #1890, opened 2021-02-05
Cleaning up unused accounts

matthiashanel:
So far, having exports or js means cleanup is skipped

These are the beginnings of account removal.
So far accounts with jetstream or exports are skipped.

If you approve of this one I'd look into that next. 
(suggestions on how to remove in the presence of jetstream are welcome)
For accounts with exports I'd check if the account can just be removed and importer set to incomplete.

### DIFF
--- a/server/accounts.go
+++ b/server/accounts.go
@@ -184,6 +184,7 @@ type streamExport struct {
 // serviceExport holds additional information for exported services.
 type serviceExport struct {
 	exportAuth
+	// back pointer to account containing this serviceExport
 	acc        *Account
 	respType   ServiceRespType
 	latency    *serviceLatency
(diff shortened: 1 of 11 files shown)


derekcollison:
How do you determine they should be cleaned up?

kozlovic:
Could this wait post 2.2? I honestly don't know (and don't have the resource to check) if removing an account is going to cause problems (think gateways, soon mqtt, etc..).

derekcollison:
I am feeling same tbh..

matthiashanel:
@derekcollison @kozlovic sure post 2.2 is fine

The removal process (putting the account name into cleanupChan) is initiated when we remove the last client of an account.
In the go routine accountCleanup a lru list of these account names to check is maintained. 
Once the list size exceeds MaxUnusedAccounts we pick an account that was unused the longest.
We check if the initial conditions (account has no client) still holds.
If so, the account is removed and cleaned up once we exceed MaxUnusedAccounts. 
If not we wait for the next time a las client disconnects.

matthiashanel:
As a reminder to myself. To get rid of accounts with exports remove the account and make the importing accounts appear as if they failed to download the exporting account. (put the discarded account into s.incompleteAccExporterMap.Store(i.Account, struct{}{}))

Consider pass through accounts where import => export

PR #1891, opened 2021-02-05
Prevent non-exported messages from leaking into importer account

variadico:
Currently when someone imports a subject with a wildcard, uses an empty string for `LocalSubject`, and if the importer account subscribes to `>`, then non-exported messages from the exported account leak to the importer account.

This happens because a shadow subscription is created on `>` (subscriber subject), instead of the import subject. From what I could see, I think these are the relevant code sections in master, in chronological order.

1. https://github.com/nats-io/nats-server/blob/master/server/accounts.go#L2272-L2283
    (value of `usePub` relevant)
2. https://github.com/nats-io/nats-server/blob/master/server/client.go#L2466-L2469
    (value of `ime.dyn` relevant [boolean value of `ime{}`])
3. https://github.com/nats-io/nats-server/blob/master/server/client.go#L2507-L2520
    (subscription created here based on `usePub` and `ime.dyn`)

This change allows the shadow subscription to be created for the import subject, instead of the subscription subject.

/cc @nats-io/core

### DIFF
--- a/server/accounts_test.go
+++ b/server/accounts_test.go
@@ -25,6 +25,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/nats-io/jwt/v2"
 	"github.com/nats-io/nats.go"
 	"github.com/nats-io/nkeys"
 )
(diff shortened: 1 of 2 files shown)


server/accounts_test.go
```diff
@@ -106,6 +107,195 @@ func TestAccountIsolation(t *testing.T) {
...
+		t.Helper()
+
+		buildMemAccResolver(s)
+		okp, err := nkeys.FromSeed(oSeed)
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		// Setup exporter account.
+		accAPair, err := nkeys.CreateAccount()
```
matthiashanel, line 131: this would be the sorter version: accAPair, accAPub := createKey(t)

server/accounts_test.go
```diff
@@ -106,6 +107,195 @@ func TestAccountIsolation(t *testing.T) {
...
+		if exp != "" {
+			accAClaims.Limits.WildcardExports = true
+			accAClaims.Exports.Add(&jwt.Export{
+				Name:    fmt.Sprintf("%s-stream-export", exp),
+				Subject: jwt.Subject(exp),
+				Type:    jwt.Stream,
+			})
+		}
+		accAJWT, err := accAClaims.Encode(okp)
+		if err != nil {
```
matthiashanel, line 149: shorter version: require_NoError(t, err)

server/accounts_test.go
```diff
@@ -106,6 +107,195 @@ func TestAccountIsolation(t *testing.T) {
...
+
+		// Setup importer account.
+		accBPair, err := nkeys.CreateAccount()
+		if err != nil {
+			t.Fatal(err)
+		}
+		accBPub, err := accBPair.PublicKey()
+		if err != nil {
+			t.Fatal(err)
+		}
```
matthiashanel, line 162: same as above

server/accounts_test.go
```diff
@@ -106,6 +107,195 @@ func TestAccountIsolation(t *testing.T) {
...
+				Name:    fmt.Sprintf("%s-stream-import", imp),
+				Subject: jwt.Subject(imp),
+				Account: accAPub,
+				Type:    jwt.Stream,
+			})
+		}
+		accBJWT, err := accBClaims.Encode(okp)
+		if err != nil {
+			t.Fatal(err)
+		}
```
matthiashanel, line 176: same as above

server/accounts_test.go
```diff
@@ -106,6 +107,195 @@ func TestAccountIsolation(t *testing.T) {
...
+			name: "export mid partial wildcard, import mid partial wildcard",
+			exp:  "foo.*.bar", imp: "foo.*.bar",
+			pubSubj: "foo.whizz.bar",
+		},
+	}
+	for _, c := range cases {
+		t.Run(c.name, func(t *testing.T) {
+			s := getServer(t)
+			defer s.Shutdown()
+			ncA, ncB := getAccountConns(t, s, c.exp, c.imp)
```
matthiashanel, line 236: Is it possible to add another version of getAccountConns that's for non jwt?
It appears that this could happen in config as well.

```
if useJWT (
    ncA, ncB = getAccountConnsJWT()
    s = getServerJWT()
} else {
    ncA, ncB = getAccountConnsCfg()
    s = getServerCfg() 
}
```
possibly have the setup function return the server as well

server/client.go
```diff
@@ -2517,6 +2517,8 @@ func (c *client) addShadowSub(sub *subscription, ime *ime) (*subscription, error
 		} else {
 			nsub.subject = []byte(im.from)
 		}
+	} else if !ime.dyn {
+		nsub.subject = []byte(im.from)
```
matthiashanel, line 2521: would it make sense to switch outer and inner conditions to simplify this ?

```
if ime.dyn {
    if !im.usePub {
        // same code as before
    }
} else {
    // same code for usePub and !usePub
}
```

server/accounts_test.go
```diff
@@ -106,6 +107,195 @@ func TestAccountIsolation(t *testing.T) {
...
+			}
+
+			for key, gotCnt := range gotSubjs {
+				if wantCnt := wantSubjs[key]; gotCnt != wantCnt {
+					t.Errorf("unexpected receive count for subject %q, got=%d, want=%d", key, gotCnt, wantCnt)
+				}
+			}
+			if t.Failed() {
+				t.Logf("exported=%q; imported=%q", c.exp, c.imp)
+			}
```
matthiashanel, line 294: use t.Fatal and avoid t.Failed()

derekcollison (approved):
LGTM - Do local tests pass ok?

server/accounts_test.go
```diff
@@ -106,6 +107,195 @@ func TestAccountIsolation(t *testing.T) {
...
+			exp:  "foo.>", imp: "foo.bar.>",
+			pubSubj: "foo.bar.whizz",
+		},
+		// FAIL - dropping "foo.bar" from different account.
+		// What is the correct behavior here?
+		// Should this be allowed?
+		// If so, probably need to edit accounts.go:2427
+		// {
+		// 	name: "export literal, import full wildcard",
+		// 	exp:  "foo.bar", imp: "foo.>",
```
matthiashanel, line 213: this looks like a violation imp should be a subset of exp, which this one wouldn't be

server/accounts_test.go
```diff
@@ -106,6 +107,195 @@ func TestAccountIsolation(t *testing.T) {
...
+		if !s.ReadyForConnections(5 * time.Second) {
+			t.Fatal("failed to be ready for connections")
+		}
+		return s
+	}
+	getAccountConns := func(t *testing.T, s *Server, exp, imp string) (*nats.Conn, *nats.Conn) {
+		t.Helper()
+
+		buildMemAccResolver(s)
+		okp, err := nkeys.FromSeed(oSeed)
```
matthiashanel, line 125: try using globally defined oKp instead

kozlovic (approved):
LGTM

matthiashanel (changes requested):

server/accounts_test.go
variadico: Ah, so I'm using this differently than a `t.Fatal`. This if statement is to only print the export and import subjects if the previous loop encounters any errors.

You get an error log like this:
```
unexpected receive count for subject "what", got=1, want=2
unexpected receive count for subject "ever", got=3, want=4
unexpected receive count for subject "mysubj", got=5, want=6
exported="whatever"; imported="whatever"
```

variadico:
> Do local tests pass ok?
Yeah, tests pass locally. The CI failures seem to be JetStream tests.

derekcollison (approved):
LGTM

matthiashanel (approved):
LGTM

PR #1892, opened 2021-02-07
JetStream updates and improvments

derekcollison:
More stability improvements to the raft core.
Made filestore better at handling write errors and propagating them.
Removed majority of MarshalIndent.

/cc @nats-io/core

### DIFF
--- a/server/store.go
+++ b/server/store.go
@@ -110,14 +110,21 @@ const (
 
 // StreamState is information about the given stream.
 type StreamState struct {
-	Msgs      uint64    `json:"messages"`
-	Bytes     uint64    `json:"bytes"`
-	FirstSeq  uint64    `json:"first_seq"`
-	FirstTime time.Time `json:"first_ts"`
-	LastSeq   uint64    `json:"last_seq"`
-	LastTime  time.Time `json:"last_ts"`
-	Deleted   []uint64  `json:"deleted,omitempty"`
-	Consumers int       `json:"consumer_count"`
+	Msgs      uint64          `json:"messages"`
+	Bytes     uint64          `json:"bytes"`
+	FirstSeq  uint64          `json:"first_seq"`
+	FirstTime time.Time       `json:"first_ts"`
+	LastSeq   uint64          `json:"last_seq"`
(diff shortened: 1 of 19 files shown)


kozlovic (approved):
LGTM

server/stream.go
```diff
@@ -97,7 +97,9 @@ type ClusterInfo struct {
 type PeerInfo struct {
 	Name    string        `json:"name"`
 	Current bool          `json:"current"`
+	Offline bool          `json:"offline,omitempty"`
 	Active  time.Duration `json:"active"`
+	Lag     uint64        `json:"lag,omitempty"`
```
ripienaar, line 102: @derekcollison would you say "How many uncomitted operations this peer is behind the leader" accurately describes this?

server/store.go
```diff
@@ -110,14 +110,21 @@ const (
...
-	LastTime  time.Time `json:"last_ts"`
-	Deleted   []uint64  `json:"deleted,omitempty"`
-	Consumers int       `json:"consumer_count"`
+	Msgs      uint64          `json:"messages"`
+	Bytes     uint64          `json:"bytes"`
+	FirstSeq  uint64          `json:"first_seq"`
+	FirstTime time.Time       `json:"first_ts"`
+	LastSeq   uint64          `json:"last_seq"`
+	LastTime  time.Time       `json:"last_ts"`
+	Deleted   []uint64        `json:"deleted,omitempty"`
```
ripienaar, line 119: @derekcollison hmm, thought we werent going to return this because it can be so much data? Less now since we dont MarshalIndent but still quite a lot I imagine in real world uses.  I also dont know what we'd do with this information, we cant undelete.  So maybe a counter of deleted messages could be good instead

server/stream.go
derekcollison: Yes correct, not necessarily things measured by the FSM.

server/store.go
derekcollison: Interior deletes should be very rare, I think including them is important. Most deletes happen at the beginning dues to limits etc. They do not show up here since FirstSeq and FirstTime move.

PR #1895, opened 2021-02-08
Fixes for supercluster startup and JetStream clustering.

derekcollison:
For larger superclusters we send out our server information after a short delay on startup.
For determining leaderless status for API calls make sure the raft node has been running for long enough.
    
/cc @nats-io/core


ripienaar:
appears to solve my problem

PR #1896, opened 2021-02-08
Added leader only response for jsz.

derekcollison:
Speaking with @ripienaar it seemed like a good option to ask the jsz endpoint to only respond if its the leader, since for the metacluster information only the leaders information is useful.

@matthiashanel could you review and possibly add some tests (apologies for that).

/cc @nats-io/core

### DIFF
--- a/server/monitor.go
+++ b/server/monitor.go
@@ -2212,15 +2212,16 @@ func (s *Server) accountInfo(accName string) (*AccountInfo, error) {
 	}, nil
 }
 
-// LeafzOptions are options passed to Jsz
+// JSzOptions are options passed to Jsz
 type JSzOptions struct {
-	Account  string `json:"account,omitempty"`
-	Accounts bool   `json:"accounts,omitempty"`
-	Streams  bool   `json:"streams,omitempty"`
-	Consumer bool   `json:"consumer,omitempty"`
-	Config   bool   `json:"config,omitempty"`
-	Offset   int    `json:"offset,omitempty"`
-	Limit    int    `json:"limit,omitempty"`
+	Account    string `json:"account,omitempty"`
+	Accounts   bool   `json:"accounts,omitempty"`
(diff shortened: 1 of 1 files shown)


server/monitor.go
```diff
@@ -2449,12 +2469,18 @@ func (s *Server) HandleJsz(w http.ResponseWriter, r *http.Request) {
 	if err != nil {
 		return
 	}
+	leader, err := decodeBool(w, r, "leader_only")
```
matthiashanel, line 2472: in urls? leader-only

matthiashanel (approved):
LGTM. name change up to you.

server/monitor.go
derekcollison: sure, or just leader=yes|1?

matthiashanel: yes just leader should be fine.

matthiashanel (approved):
LGTM

server/monitor.go
derekcollison: Went with leader-only

server/monitor.go
```diff
@@ -2337,6 +2338,25 @@ func (s *Server) Jsz(opts *JSzOptions) (*JSInfo, error) {
 	if opts.Streams {
 		opts.Accounts = true
 	}
+
+	// Check if we want a response from the leader only.
+	if opts.LeaderOnly {
+		js, cc := s.getJetStreamCluster()
+		if js == nil {
+			// Ignore
+			return nil, ErrJetStreamNotEnabled
```
ripienaar, line 2347: Here and on line 2355 if we return Err the error is sent back to the clients, meaning the !leaders still respond but in error.

These 2 lines should be `return nil, nil`

ripienaar: Ugh, not even ... will dig deeper

ripienaar: https://github.com/nats-io/nats-server/pull/1902

PR #1898, opened 2021-02-08
[added] system_account to varz/accounts and is_system_account to accountz

matthiashanel:

### DIFF
--- a/server/monitor.go
+++ b/server/monitor.go
@@ -1991,28 +1991,28 @@ type ExtVrIssues struct {
 type ExtMap map[string][]*MapDest
 
 type AccountInfo struct {
-	AccountName     string               `json:"account_name"`
-	LastUpdate      time.Time            `json:"update_time,omitempty"`
-	IsSystemAccount bool                 `json:"is_system_account,omitempty"`
-	Expired         bool                 `json:"expired"`
-	Complete        bool                 `json:"complete"`
-	JetStream       bool                 `json:"jetstream_enabled"`
-	LeafCnt         int                  `json:"leafnode_connections"`
-	ClientCnt       int                  `json:"client_connections"`
-	SubCnt          uint32               `json:"subscriptions"`
-	Mappings        ExtMap               `json:"mappings,omitempty"`
-	Exports         []ExtExport          `json:"exports,omitempty"`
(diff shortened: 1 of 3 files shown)


server/monitor.go
```diff
@@ -1987,34 +1991,36 @@ type ExtVrIssues struct {
...
-	Tags        jwt.TagList          `json:"tags,omitempty"`
-	Claim       *jwt.AccountClaims   `json:"decoded_jwt,omitempty"`
-	Vr          []ExtVrIssues        `json:"validation_result_jwt,omitempty"`
-	RevokedUser map[string]time.Time `json:"revoked_user,omitempty"`
-	RevokedAct  map[string]time.Time `json:"revoked_activations,omitempty"`
-	Sublist     *SublistStats        `json:"sublist_stats,omitempty"`
-	Responses   map[string]ExtImport `json:"responses,omitempty"`
+	AccountName     string               `json:"account_name"`
+	LastUpdate      time.Time            `json:"update_time,omitempty"`
+	IsSystemAccount bool                 `json:"is_system_account,omitempty"`
```
kozlovic, line 1996: We are in "Account", so "IsSystem" (and `is_system` for the json tag) would be enough in my opinion. (we can't change now, but would be the same for Name instead of AccountName, etc..)

server/monitor_test.go
```diff
@@ -3791,6 +3793,10 @@ func TestMonitorAccountz(t *testing.T) {
 		t.Fatalf("Body missing value. Contains: %s", body)
 	} else if !strings.Contains(body, `"subscriptions": 36,`) {
 		t.Fatalf("Body missing value. Contains: %s", body)
+	} else if !strings.Contains(body, `"is_system_account": true,`) {
```
kozlovic, line 3796: Would have to change that if you change the json tag name.

kozlovic (changes requested):

server/monitor.go
matthiashanel: fixed

server/monitor_test.go
matthiashanel: fixed

kozlovic (approved):
LGTM

PR #1899, opened 2021-02-08
Remap subject only for service imports

kozlovic:
Also optimized a test that was taking too long to run.

/cc @nats-io/core


variadico (approved):

derekcollison (approved):
LGTM

PR #1902, opened 2021-02-09
skip responses from non leaders in jsz

ripienaar:
Adds an error type and a check to avoid sending them
to clients

/cc @nats-io/core

### DIFF
--- a/server/monitor.go
+++ b/server/monitor.go
@@ -2357,7 +2357,7 @@ func (s *Server) Jsz(opts *JSzOptions) (*JSInfo, error) {
 		js, cc := s.getJetStreamCluster()
 		if js == nil {
 			// Ignore
-			return nil, ErrJetStreamNotEnabled
+			return nil, fmt.Errorf("%w: no cluster", errSkipZreq)
 		}
 		// So if we have JS but no clustering, we are the leader so allow.
 		if cc != nil {
@@ -2365,7 +2365,7 @@ func (s *Server) Jsz(opts *JSzOptions) (*JSInfo, error) {
 			isLeader := cc.isLeader()
 			js.mu.RUnlock()
 			if !isLeader {
-				return nil, errNotLeader
+				return nil, fmt.Errorf("%w: not leader", errSkipZreq)
 			}
 		}
 	}
--- a/server/events.go
+++ b/server/events.go
@@ -1155,6 +1155,8 @@ func (s *Server) statszReq(sub *subscription, _ *client, subject, reply string, 
 	s.mu.Unlock()
 }
 
+var errSkipZreq = errors.New("filtered response")
+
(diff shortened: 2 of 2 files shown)


server/monitor.go
```diff
@@ -2357,15 +2357,15 @@ func (s *Server) Jsz(opts *JSzOptions) (*JSInfo, error) {
 		js, cc := s.getJetStreamCluster()
 		if js == nil {
 			// Ignore
-			return nil, ErrJetStreamNotEnabled
+			return nil, fmt.Errorf("%w: no cluster", skipZreq)
```
ripienaar, line 2360: I do it this way as I didnt want to change the behavior for HTTP requests - I dont know what we want to do exactly when making a request over HTTP that isnt for us?

Anyway now it does:

```
# curl "0:8222/jsz?leader-only=true"
filtered response: not leader
```

server/monitor.go
```diff
@@ -2357,15 +2357,15 @@ func (s *Server) Jsz(opts *JSzOptions) (*JSInfo, error) {
 		js, cc := s.getJetStreamCluster()
 		if js == nil {
 			// Ignore
-			return nil, ErrJetStreamNotEnabled
+			return nil, fmt.Errorf("%w: no cluster", errSkipZreq)
```
ripienaar, line 2360: I do it this way as I didnt want to change the behavior for HTTP requests - I dont know what we want to do exactly when making a request over HTTP that isnt for us?

Anyway now it returns a 400 status with:

```
# curl "0:8222/jsz?leader-only=true"
filtered response: not leader
```

derekcollison: Yes does not make too much sense on http.

server/monitor.go
```diff
@@ -2357,15 +2357,15 @@ func (s *Server) Jsz(opts *JSzOptions) (*JSInfo, error) {
...
 		}
 		// So if we have JS but no clustering, we are the leader so allow.
 		if cc != nil {
 			js.mu.RLock()
 			isLeader := cc.isLeader()
 			js.mu.RUnlock()
 			if !isLeader {
-				return nil, errNotLeader
+				return nil, fmt.Errorf("%w: not leader", errSkipZreq)
```
derekcollison, line 2368: Not sure if this will work with the errors.Is above..

ripienaar: It works, thats the point of `%w` it wraps the error and errors.Is() checks the wrapped errors.

derekcollison: Gotcha

derekcollison (approved):
LGTM - Make sure all tests pass locally as best you can.

ripienaar:
Failures here are race conditions in TestJetStreamClusterRemoveServer so I reckon its fine to merge as those are surely predating, wydt @derekcollison

derekcollison:
Yes go ahead. Will sort it out..

PR #1904, opened 2021-02-10
Remove mkpasswd from server

variadico:
The functionality of mkpasswd has been added to nats-io/natscli instead.

We'll need to update the docs as well: https://docs.nats.io/?q=mkpasswd


derekcollison (approved):
LGTM

wallyqs (approved):
LGTM

kozlovic:
Would be good to update nats.io doc in all places where mkpasswd is referenced: https://docs.nats.io/nats-tools/mkpasswd?q=mkpasswd

PR #1906, opened 2021-02-12
Snapshot and Compaction Improvements

derekcollison:
Previously we did a snapshot and compaction strategy that would require pauses to the proposals from clients which was not ideal. This changes that to a raft node maintained external snapshots that do not require pauses. We also moved to memory based WALs for streams and consumers.

Various other bug fixes, data race fixes and stability improvements.

/cc @nats-io/core


kozlovic (approved):
LGTM

PR #1907, opened 2021-02-13
Changes needed to properly support multi-layer service imports

derekcollison:

/cc @nats-io/core

### DIFF
--- a/server/client.go
+++ b/server/client.go
@@ -3568,9 +3568,23 @@ func (c *client) setupResponseServiceImport(acc *Account, si *serviceImport, tra
 	return rsi
 }
 
+// Will remove a header if present.
+func removeHeaderIfPresent(hdr []byte, key string) []byte {
+	start := bytes.Index(hdr, []byte(key))
+	if start < 0 {
+		return hdr
+	}
+	end := bytes.Index(hdr[start:], []byte(_CRLF_))
+	if end < 0 {
+		return hdr
+	}
+	return append(hdr[:start], hdr[start+end+len(_CRLF_):]...)
+}
+
 // This will set a header for the message.
 // Lock does not need to be held but this should only be called
 // from the inbound go routine. We will update the pubArgs.
+// This will replace any previously set header and not add to it per normal spec.
 func (c *client) setHeader(key, value string, msg []byte) []byte {
 	const hdrLine = "NATS/1.0\r\n"
 	var bb bytes.Buffer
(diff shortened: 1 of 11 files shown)


server/client.go
```diff
@@ -3597,6 +3613,33 @@ func (c *client) setHeader(key, value string, msg []byte) []byte {
...
+		return nil
+	}
+	index := bytes.Index(hdr, []byte(key))
+	if index < 0 {
+		return nil
+	}
+
+	var value []byte
+	hdrLen := len(hdr)
+	index += len(key) + 1
```
kozlovic, line 3629: In the mqtt_clustering branch I have modified getHdrVal() with the following:
```
if index >= hdrLen {
    return nil
}
if hdr[index] != ':' {
    return nil
}
index++
for hdr[index] == ' ' && index < hdrLen {
    index++
}
for index < hdrLen {
    ...
```
The reason for that is that since we don't do proper lifting, at least lets make sure that the header name matches what we are looking for. Say you look for key `foo` in the header `foobar: some_value`, I don't want this function to return `ar: somevalue` but return nil because `foo:` was not found.

kozlovic (approved):
LGTM, but would harden a bit the detection of a header key in the helper function.

server/client.go
derekcollison: Good idea, will fix.

derekcollison: ok fixed.

kozlovic: Not, it is not, you did not check for `if index >= hdrLen` so then it would panic. See this: https://play.golang.org/p/98u-DLyOd-_3

server/client.go
```diff
@@ -3597,6 +3613,37 @@ func (c *client) setHeader(key, value string, msg []byte) []byte {
...
+func getHeader(key string, hdr []byte) []byte {
+	if len(hdr) == 0 {
+		return nil
+	}
+	index := bytes.Index(hdr, []byte(key))
+	if index < 0 {
+		return nil
+	}
+	index += len(key)
+	if hdr[index] != ':' {
```
kozlovic, line 3627: after moving the index, you need to check that it is not passed the length of `hdr` because otherwise it would panic.

kozlovic (changes requested):

matthiashanel (approved):
LGTM

server/client.go
derekcollison: Will fix.

kozlovic (approved):
LGTM

PR #1908, opened 2021-02-13
Set pindex to wrong setting on snapshot restore with no WAL

derekcollison:

/cc @nats-io/core

### DIFF
--- a/server/jetstream_test.go
+++ b/server/jetstream_test.go
@@ -8822,7 +8822,7 @@ func TestJetStreamPubPerf(t *testing.T) {
 	nc := clientConnectToServer(t, s)
 	defer nc.Close()
 
-	toSend := 5000000
+	toSend := 5_000_000
 	numProducers := 1
 
 	payload := []byte("Hello World")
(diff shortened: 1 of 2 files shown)


ripienaar:
LGTM, couldnt repro crash

server/jetstream_test.go
```diff
@@ -8895,6 +8895,38 @@ func TestJetStreamPubWithAsyncResponsePerf(t *testing.T) {
 	fmt.Printf("%.0f msgs/sec\n", float64(toSend)/tt.Seconds())
 }
 
+func TestJetStreamPubWithSyncPerf(t *testing.T) {
+	// Comment out to run, holding place for now.
+	//t.SkipNow()
```
kozlovic, line 8900: You need to put that back: it will probably fail all -race runs and even without race (the coverall run) takes too long:
```
=== RUN   TestJetStreamPubWithSyncPerf
time is 42.463880496s
23549 msgs/sec
--- PASS: TestJetStreamPubWithSyncPerf (42.49s)
```

derekcollison: Thanks will do.

PR #1909, opened 2021-02-15
Restoration of streams would possibly block route and client connections.

derekcollison:

/cc @nats-io/core


kozlovic (approved):
LGTM

PR #1911, opened 2021-02-15
enable manual triggers of nightly builds

ripienaar:

/cc @nats-io/core


derekcollison (approved):
LGTM

PR #1914, opened 2021-02-16
revert to previous checkout version

ripienaar:

/cc @nats-io/core


ripienaar:
I am not really sure if this will fix the nightly builds - but its the only thing I changed other than supporting a button to trigger it should seems unrelated.

derekcollison (approved):
LGTM

ripienaar:
@derekcollison ok that works again, on demand builds too

PR #1916, opened 2021-02-16
Better restore logic

derekcollison:
General stability improvements on catch up, restore and snapshot.

/cc @nats-io/core


kozlovic (approved):
LGTM

PR #1918, opened 2021-02-17
Some stability improvements to raft lib and catchup stream processing.

derekcollison:

/cc @nats-io/core

### DIFF
--- a/server/raft.go
+++ b/server/raft.go
@@ -369,6 +369,11 @@ func (s *Server) startRaftNode(cfg *RaftConfig) (RaftNode, error) {
 
 		for index := state.FirstSeq; index <= state.LastSeq; index++ {
 			ae, err := n.loadEntry(index)
+			if ae.pindex != index-1 {
+				n.warn("Corrupt WAL, truncating")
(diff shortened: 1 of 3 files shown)


server/raft.go
```diff
@@ -2268,6 +2275,15 @@ func (n *raft) storeToWAL(ae *appendEntry) error {
 		fmt.Printf("[%s] n is %+v\n\n", n.s, n)
 		fmt.Printf("[%s] n.catchup is %+v\n", n.s, n.catchup)
 		fmt.Printf("[%s] n.wal is %+v\n", n.s, n.wal.State())
+		if state := n.wal.State(); state.Msgs > 0 {
+			for index := state.FirstSeq; index <= state.LastSeq; index++ {
+				nae, _ := n.loadEntry(index)
+				fmt.Printf("INDEX %d is %+v\n", index, nae)
```
ripienaar, line 2281: some debug left here

derekcollison: Yes on purpose to print out when the condition trips. There are several nuanced bugs I believe that can lead to this condition.

PR #1920, opened 2021-02-18
General stability improvements.

derekcollison:
Original thought to move to memory based WALs was ill-advised and caused issues with stability around restarts.
Returned to file based but with async flush for the WAL itself.
Also the raft inline catchup has been improved.

/cc @nats-io/core


kozlovic (approved):
LGTM

PR #1923, opened 2021-02-18
[fixed] out of range issue in processMsgResults during subject rewrite

matthiashanel:

doc of LastIndex says it will return -1 when none found.


kozlovic (approved):
LGTM

PR #1924, opened 2021-02-19
Cross cluster consumer interest.

derekcollison:
In certain circumstances we were not taking gateway interest into consideration for consumers.
Also upgrade Go client and more raft stability fixes around split votes under heavy load.

/cc @nats-io/core


server/raft.go
```diff
@@ -2090,7 +2104,7 @@ func (n *raft) processAppendEntry(ae *appendEntry, sub *subscription) {
 			if eae, err := n.loadEntry(ae.pindex); err == nil && eae != nil {
 				// If terms mismatched, delete that entry and all others past it.
 				if ae.pterm > eae.pterm {
-					n.wal.Truncate(ae.pindex - 1)
+					n.wal.Truncate(ae.pindex)
```
kozlovic, line 2107: again an off by one? :-)

kozlovic (approved):
LGTM

server/raft.go
derekcollison: yea

matthiashanel (approved):
LGTM

PR #1925, opened 2021-02-19
[fixed] LeafNode sending message using stream's import subject.

matthiashanel:
A publish on "a" becomes an LMSG on ">" which
is the stream import's subject. The subscriber on "a" on the other
side did not receive the message.

This change adds @kozlovic unit test and adds a missing check.
The check is now identical to other regions of code doing smth similar.
```
func (c *client) processMsgResults(acc *Account, r *SublistResult, msg, deliver, subject, reply []byte, flags int) (bool, [][]byte) {
...
	// Loop over all normal subscriptions that match.
	for _, sub := range r.psubs {
...
		// Assume delivery subject is normal subject to this point.
		dsubj = subj
		// Check for stream import mapped subs (shadow subs). These apply to local subs only.
		if sub.im != nil {
			if sub.im.tr != nil {
				to, _ := sub.im.tr.transformSubject(string(subj))
				dsubj = append(_dsubj[:0], to...)
			} else if sub.im.usePub {
				dsubj = append(_dsubj[:0], subj...)
			} else {
				dsubj = append(_dsubj[:0], sub.im.to...)
			}
		}
...
	}
...
	selectQSub:
...

		// Find a subscription that is able to deliver this message starting at a random index.
		for i := 0; i < lqs; i++ {
...
			if sub.im != nil {
				if sub.im.tr != nil {
					to, _ := sub.im.tr.transformSubject(string(subj))
					dsubj = append(_dsubj[:0], to...)
				} else if sub.im.usePub {
					dsubj = append(_dsubj[:0], subj...)
				} else {
					dsubj = append(_dsubj[:0], sub.im.to...)
				}
			}
...
		}
...
}
```

### DIFF
--- a/server/client.go
+++ b/server/client.go
@@ -2802,7 +2802,7 @@ func (c *client) msgHeaderForRouteOrLeaf(subj, reply []byte, rt *routeTarget, ac
 			if rt.sub.im.tr != nil {
 				to, _ := rt.sub.im.tr.transformSubject(string(subj))
 				subj = []byte(to)
-			} else {
+			} else if !rt.sub.im.usePub {
 				subj = []byte(rt.sub.im.to)
 			}
 		}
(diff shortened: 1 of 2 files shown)


kozlovic (approved):
LGTM

PR #1926, opened 2021-02-19
MQTT clustering

kozlovic:
Made changes needed to support JetStream clustering.
I am no longer making any call to direct stream/consumer internal APIs.
I have no longer the need for "internal/no_interest/no_subject" private stream/consumer configuration properties, so I have removed them and remove the JSON marshaller that was added to persist those (since they were not exported).

### DIFF
--- a/server/client.go
+++ b/server/client.go
@@ -3606,10 +3606,10 @@ func (c *client) setHeader(key, value string, msg []byte) []byte {
 	// Put the original message back.
 	// FIXME(dlc) - This is inefficient.
 	bb.Write(msg[omi:])
-	nsize := bb.Len()
-	// MQTT producers don't have CRLF
-	if !c.isMqtt() {
-		nsize -= LEN_CR_LF
+	nsize := bb.Len() - LEN_CR_LF
+	// MQTT producers don't have CRLF, so add it back.
+	if c.isMqtt() {
+		nsize += LEN_CR_LF
 	}
 	// Update pubArgs
 	// If others will use this later we need to save and restore original.
--- a/server/client.go
+++ b/server/client.go
@@ -117,6 +117,8 @@ var readLoopReportThreshold = readLoopReport
 // Represent client booleans with a bitmask
 type clientFlag uint16
 
+const hdrLine = "NATS/1.0\r\n"
+
 // Some client state represented as flags
 const (
 	connectReceived   clientFlag = 1 << iota // The CONNECT proto has been received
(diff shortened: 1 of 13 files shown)


server/client.go
```diff
@@ -117,6 +117,8 @@ var readLoopReportThreshold = readLoopReport
 // Represent client booleans with a bitmask
 type clientFlag uint16
 
+const hdrLine = "NATS/1.0\r\n"
```
derekcollison, line 120: We already have this somewhere I thought defined.

server/client.go
```diff
@@ -3603,7 +3606,11 @@ func (c *client) setHeader(key, value string, msg []byte) []byte {
 	// Put the original message back.
 	// FIXME(dlc) - This is inefficient.
 	bb.Write(msg[omi:])
-	nsize := bb.Len() - LEN_CR_LF
+	nsize := bb.Len()
```
derekcollison, line 3609: Would reverse and set to -LEN and add back in when MQTT.

server/consumer.go
```diff
@@ -386,7 +378,7 @@ func (mset *stream) addConsumerWithAssignment(config *ConsumerConfig, oname stri
 			} else {
 				// If we are a push mode and not active and the only difference
 				// is deliver subject then update and return.
-				if configsEqualSansDelivery(ocfg, *config) && (config.allowNoInterest || eo.hasNoLocalInterest()) {
+				if configsEqualSansDelivery(ocfg, *config) && eo.hasNoLocalInterest() {
```
derekcollison, line 381: Should make this SC aware like we did yesterday for the consumer fix.

derekcollison (approved):
LGTM - Minor comments.

server/client.go
kozlovic: I don't think so, this is why I moved it here from inside setHeader()

server/client.go
kozlovic: I can do that. Right now, I am not using setHeader() with an MQTT client, so I could even completely remove the change, but if one day I use that, I would get parse errors on the other side if I don't have such code.

server/consumer.go
kozlovic: This code is rebased from main.. If you made a change here, and is not yet merged, then it will be added during the merge. But maybe I am misunderstanding what you are asking.

server/client.go
derekcollison: I can see it in master. But was scoped inside a function.

server/client.go
derekcollison: I am saying default single op should be normal NATS case and MQTT case can put it back vs 3 ops for normal NATS flow.

server/consumer.go
derekcollison: ok so maybe this is something I need to change on my side for GW interest checks etc.

server/client.go
kozlovic: Oh sure.. I was just saying that I could even remove the changes in this function since I am not using it. But I will make the changes you mention and keep the MQTT case just in case I use it in the future.

derekcollison (approved):
LGTM

PR #1927, opened 2021-02-19
Allow placement directives for meta controllers

derekcollison:
Allow metacontroller leaders to be moved to specific clusters in a supercluster setup.

/cc @nats-io/core

### DIFF
--- a/server/jetstream_api.go
+++ b/server/jetstream_api.go
@@ -18,6 +18,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
+	"math/rand"
 	"os"
 	"sort"
 	"strconv"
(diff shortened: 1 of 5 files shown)


server/jetstream_api.go
```diff
@@ -1699,10 +1707,47 @@ func (s *Server) jsLeaderStepDownRequest(sub *subscription, c *client, subject,
...
+			// Tags currently not supported.
+			resp.Error = jsClusterTagsErr
+			s.sendAPIErrResponse(ci, acc, subject, reply, string(msg), s.jsonResponse(&resp))
+			return
+		}
+		cn := req.Placement.Cluster
+		var peers []string
+		for _, p := range cc.meta.Peers() {
+			si, ok := s.nodeToInfo.Load(p.ID)
+			ni := si.(*nodeInfo)
```
kozlovic, line 1730: if `!ok` then `si` would be nil, no? It may be a case where you need 2 separate `if`'s.

ripienaar (approved):
LGTM feature wise

kozlovic (changes requested):
Danger of `si` being nil...

server/jetstream_api.go
derekcollison: Good point, will fix.

derekcollison: ok take another look.

kozlovic (approved):
LGTM

PR #1929, opened 2021-02-22
Add optional DeliverTo field to ConsumerGetNextRequest JS API

wallyqs:
This adds a `DeliverTo` which can be used to signal the server where the next message should be delivered after sending the next message request.  This way when a client makes a request to pull messages, it can receive the response from a different response handler than the subscription that is processing the messages.

 - [X] Tests added

/cc @nats-io/core

### DIFF
--- a/server/consumer.go
+++ b/server/consumer.go
@@ -1540,27 +1540,27 @@ func (o *consumer) needAck(sseq uint64) bool {
 }
 
 // Helper for the next message requests.
-func nextReqFromMsg(msg []byte) (time.Time, int, bool, error) {
+func nextReqFromMsg(msg []byte) (time.Time, int, bool, string, error) {
 	req := bytes.TrimSpace(msg)
 
 	switch {
 	case len(req) == 0:
-		return time.Time{}, 1, false, nil
+		return time.Time{}, 1, false, "", nil
 
 	case req[0] == '{':
 		var cr JSApiConsumerGetNextRequest
 		if err := json.Unmarshal(req, &cr); err != nil {
-			return time.Time{}, -1, false, err
+			return time.Time{}, -1, false, "", err
 		}
-		return cr.Expires, cr.Batch, cr.NoWait, nil
+		return cr.Expires, cr.Batch, cr.NoWait, cr.DeliverTo, nil
 
 	default:
 		if n, err := strconv.Atoi(string(req)); err == nil {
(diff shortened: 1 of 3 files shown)


wallyqs:
As is this wouldn't work when the pull request is done across accounts, it needs to be aware of the remapped inbox:

```
[83750] 2021/02/25 17:13:39.893837 [TRC] [::1]:57987 - cid:19 - "v1.11.0:go" - <<- [PUB $JS.API.CONSUMER.MSG.NEXT.foo.p1 _INBOX.FdjTGbgmxBWWMDXJTDGiA6.jejKsJcA 56]
[83750] 2021/02/25 17:13:39.893852 [TRC] [::1]:57987 - cid:19 - "v1.11.0:go" - <<- MSG_PAYLOAD: ["{\"batch\":1,\"deliver_to\":\"_INBOX.FdjTGbgmxBWWMDXJTDGiCm\"}"]
[83750] 2021/02/25 17:13:39.893926 [TRC] [::1]:57987 - cid:19 - "v1.11.0:go" - ->> [MSG _R_.Jqf851.3tLjgP 1 0]
```

server/consumer.go
```diff
@@ -1718,7 +1729,14 @@ func (o *consumer) processNextMsgReq(_ *subscription, c *client, _, reply string
 			break
 		}
 	}
+	sendq := o.sendq
 	o.mu.Unlock()
+
+	if shouldRespond {
+		// Empty message meaning that there were no errors.
+		pmsg := &jsPubMsg{reply, reply, _EMPTY_, nil, nil, nil, 0}
```
wallyqs, line 1737: In case the `sub.Pull()` is done from another account, then this would not deliver the response.

derekcollison: This should work ok once I push the current branch. Sync with that should work fine.

derekcollison: This should not be needed. Let's chat in am.

wallyqs: Just tried with latest and seems that for the approach in this PR, it still needs to transform the reply inbox. An alternative that I was thinking to make the sub.Poll() synchronous is to handle as an old style request on the client side: https://github.com/nats-io/nats.go/blob/118341e3023b7bf5bf01b3f332b98fc52e2c22b1/js.go#L913-L980

wallyqs:
Closing since taking a different approach.

PR #1931, opened 2021-02-23
Stream Mirrors, Sources and Updates Support

derekcollison:
Also various fixes, stability improvements, etc.

/cc @nats-io/core


server/const.go
```diff
@@ -1,202 +0,0 @@
-// Copyright 2012-2021 The NATS Authors
```
ripienaar, line 1: this is probably a mistake right?

server/jetstream_api.go
```diff
@@ -1020,6 +1024,55 @@ func (s *Server) jsStreamCreateRequest(sub *subscription, c *client, subject, re
...
+			s.sendAPIErrResponse(ci, acc, subject, reply, string(msg), s.jsonResponse(&resp))
+			return
+		}
+		if cfg.Mirror.FilterSubject != _EMPTY_ {
+			resp.Error = &ApiError{Code: 400, Description: "stream mirrors can not contain filtered subjects"}
+			s.sendAPIErrResponse(ci, acc, subject, reply, string(msg), s.jsonResponse(&resp))
+			return
+		}
+		if cfg.Mirror.OptStartSeq > 0 && cfg.Mirror.OptStartTime != nil {
+			resp.Error = &ApiError{Code: 400, Description: "stream mirrors can not have both start seq and start time configured"}
```
ripienaar, line 1045: hmm, so then stream mirrors can in fact have no config at all, so the *StreamSource only contributes Name at that point, are we keeping *StreamSource there as a nod to the future when we somehow point to another account in there too?

server/const.go
derekcollison: yes moved.

server/jetstream_api.go
derekcollison: Not following..

PR #1932, opened 2021-02-23
Stream Mirrors, Sources and Updates Support

derekcollison:
Also various fixes, stability improvements, etc.

/cc @nats-io/core

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,8 +3,8 @@ dist: focal
 
 language: go
 go:
+- 1.16.x
 - 1.15.x
-- 1.14.x
 addons:
   apt:
     packages:
@@ -29,7 +29,7 @@ script:
 - set -e
 - go test -i ./...
 - go test -v -run=TestNoRace --failfast -p=1 ./...
-- if [[ "$TRAVIS_GO_VERSION" =~ 1.14 ]]; then ./scripts/cov.sh TRAVIS; else go test -v -race -p=1 --failfast ./...; fi
+- if [[ "$TRAVIS_GO_VERSION" =~ 1.15 ]]; then ./scripts/cov.sh TRAVIS; else go test -v -race -p=1 --failfast ./...; fi
 - set +e
 
 deploy:
@@ -38,4 +38,4 @@ deploy:
   script: curl -sL http://git.io/goreleaser | bash
   on:
     tags: true
-    condition: $TRAVIS_GO_VERSION =~ 1.14
+    condition: $TRAVIS_GO_VERSION =~ 1.16
--- a/server/jetstream_api.go
+++ b/server/jetstream_api.go
@@ -549,7 +549,7 @@ var (
 	jsNotEnabledErr        = &ApiError{Code: 503, Description: "JetStream not enabled for account"}
 	jsBadRequestErr        = &ApiError{Code: 400, Description: "bad request"}
 	jsNotEmptyRequestErr   = &ApiError{Code: 400, Description: "expected an empty request payload"}
-	jsInvalidJSONErr       = &ApiError{Code: 400, Description: "invalid JSON request"}
+	jsInvalidJSONErr       = &ApiError{Code: 400, Description: "invalid JSON"}
 	jsInsufficientErr      = &ApiError{Code: 503, Description: "insufficient resources"}
 	jsNoConsumerErr        = &ApiError{Code: 404, Description: "consumer not found"}
 	jsStreamMismatchErr    = &ApiError{Code: 400, Description: "stream name in subject does not match request"}
@@ -633,13 +633,17 @@ func (s *Server) setJetStreamExportSubs() error {
 
 func (s *Server) sendAPIResponse(ci *ClientInfo, acc *Account, subject, reply, request, response string) {
 	acc.trackAPI()
-	s.sendInternalAccountMsg(nil, reply, response)
+	if reply != _EMPTY_ {
+		s.sendInternalAccountMsg(nil, reply, response)
+	}
 	s.sendJetStreamAPIAuditAdvisory(ci, acc, subject, request, response)
 }
 
 func (s *Server) sendAPIErrResponse(ci *ClientInfo, acc *Account, subject, reply, request, response string) {
 	acc.trackAPIErr()
-	s.sendInternalAccountMsg(nil, reply, response)
+	if reply != _EMPTY_ {
+		s.sendInternalAccountMsg(nil, reply, response)
+	}
 	s.sendJetStreamAPIAuditAdvisory(ci, acc, subject, request, response)
 }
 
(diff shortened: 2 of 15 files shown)


server/stream.go
```diff
@@ -132,12 +161,23 @@ type stream struct {
...
+	last time.Time
+}
+
 // Headers for published messages.
 const (
 	JSMsgId             = "Nats-Msg-Id"
 	JSExpectedStream    = "Nats-Expected-Stream"
 	JSExpectedLastSeq   = "Nats-Expected-Last-Sequence"
 	JSExpectedLastMsgId = "Nats-Expected-Last-Msg-Id"
+	JSStreamSource      = "Nats-Stream-Source"
```
ripienaar, line 180: I saw these and wasnt sure what the user story this fixes? Would the user parse this subjet (its the ack subject) to figure out metadata? This would have original time stamp I guess, the mirrored timestamp would be different?

server/jetstream_api.go
```diff
@@ -1020,6 +1024,55 @@ func (s *Server) jsStreamCreateRequest(sub *subscription, c *client, subject, re
...
+		if cfg.Mirror.FilterSubject != _EMPTY_ {
+			resp.Error = &ApiError{Code: 400, Description: "stream mirrors can not contain filtered subjects"}
+			s.sendAPIErrResponse(ci, acc, subject, reply, string(msg), s.jsonResponse(&resp))
+			return
+		}
+		if cfg.Mirror.OptStartSeq > 0 && cfg.Mirror.OptStartTime != nil {
+			resp.Error = &ApiError{Code: 400, Description: "stream mirrors can not have both start seq and start time configured"}
+			s.sendAPIErrResponse(ci, acc, subject, reply, string(msg), s.jsonResponse(&resp))
+			return
+		}
```
ripienaar, line 1048: hmm, so then stream mirrors can in fact have no config at all, so the *StreamSource only contributes Name at that point, are we keeping *StreamSource there as a nod to the future when we somehow point to another account in there too?

Your 4 checks here - no subjects, no sources, no filter and no start seq or time - effectively makes the cfg.Mirror only use Name nothing else - since everything else errors.

So why does Mirrors have it? So that we can later add things to point to a stream in another account?

ripienaar: Sorry, My bad, dont read code at 9pm

server/raft.go
```diff
@@ -1069,7 +1069,18 @@ func (n *raft) shutdown(shouldDelete bool) {
...
+		var subs []*subscription
+		c.mu.Lock()
+		for _, sub := range c.subs {
+			subs = append(subs, sub)
+		}
+		c.mu.Unlock()
+		for _, sub := range subs {
+			n.unsubscribe(sub)
+		}
+		c.closeConnection(InternalClient)
```
kozlovic, line 1082: Note that unsubscribe and closing a client will end-up grabbing the server lock (or others due to update of some objects, route maps, etc..), so it may need to be done outside of n.Lock() *if* you have situations where you have server lock -> raft lock.

server/stream.go
derekcollison: They could, we need it for sources to properly figure out where to pick back up for any given source. We walk backwards through our own stream and look at that header for stream sequence for the source stream. Then we recreate the consumer with that starting sequence +1.

server/jetstream_api.go
derekcollison: Just quick check if you accidentally set both start seq and start time.

kozlovic (approved):
LGTM

server/raft.go
derekcollison: Will take a look, and want to get a green test run so will look at the flappers in a bit.

server/raft.go
```diff
@@ -1449,6 +1449,9 @@ func (n *raft) handleForwardedProposal(sub *subscription, c *client, _, reply st
 
 func (n *raft) runAsLeader() {
 	n.RLock()
+	if n.state == Closed {
+		return
```
kozlovic, line 1453: Forgot the n.RUnlock() here.

kozlovic (changes requested):

.travis.yml
```diff
@@ -3,8 +3,8 @@ dist: focal
 
 language: go
 go:
+- 1.16.x
 - 1.15.x
-- 1.14.x
```
kozlovic, line 7: If you remove 1.14, please check line 32 where we run code coverage for 1.14 (and race tests for 1.15). Adjust to the new versions.

kozlovic (changes requested):

.travis.yml
kozlovic: While we are at it, if we plan to release as 1.16, then put 1.16 line 41

server/stream.go
ripienaar: OK, its like 70 bytes overhead per message. We could pick out just the bits we need and put it in a header we dont show the user - and so with a much shorter name?

Usually I am not too concerned about messages sizes - a lot less than you :P - but 70/message is a lot.

ripienaar:
```+----------------+---------+----------+-----------+----------+---------+------+---------+----------------------+
| Stream         | Storage | Template | Consumers | Messages | Bytes   | Lost | Deleted | Cluster              |
+----------------+---------+----------+-----------+----------+---------+------+---------+----------------------+
| ORDERS_6       | File    |          | 5         | 0        | 0 B     | 0    | 0       | n1-c1, n2-c1, n3-c1* |
| ORDERS_7       | File    |          | 5         | 0        | 0 B     | 0    | 0       | n1-c1, n2-c1*, n3-c1 |
| ORDERS_2       | File    |          | 5         | 0        | 0 B     | 0    | 0       | n1-c1, n2-c1*, n3-c1 |
| ORDERS_3       | File    |          | 5         | 0        | 0 B     | 0    | 0       | n1-c1, n2-c1*, n3-c1 |
| ORDERS_4       | File    |          | 5         | 0        | 0 B     | 0    | 0       | n1-c1*, n2-c1, n3-c1 |
| ORDERS_5       | File    |          | 5         | 0        | 0 B     | 0    | 0       | n1-c1*, n2-c1, n3-c1 |
| ORDERS_8       | File    |          | 5         | 0        | 0 B     | 0    | 0       | n1-c1, n2-c1*, n3-c1 |
| ORDERS_9       | File    |          | 5         | 0        | 0 B     | 0    | 0       | n1-c1, n2-c1*, n3-c1 |
| ORDERS_0       | File    |          | 6         | 100      | 4.5 KiB | 0    | 0       | n1-c1, n2-c1*, n3-c1 |
| ORDERS_1       | File    |          | 6         | 100      | 4.5 KiB | 0    | 0       | n1-c1*, n2-c1, n3-c1 |
| ORDERS_ARCHIVE | File    |          | 0         | 200      | 27 KiB  | 0    | 0       | n2-c2*               |
+----------------+---------+----------+-----------+----------+---------+------+---------+----------------------+```

Worst case since my messages are tiny though

derekcollison: After our discussion let me know if you still want me to look into shrinking or do we think this could be valuable as is to end users?

/cc @kozlovic

ripienaar: I think its ok, it should be useful. 1 billion messages = 70GB of this header :P

derekcollison: Compression ;)

PR #1933, opened 2021-02-24
General stability improvements. 

derekcollison:
Tracked down the expected not what we wanted state with snapshots etc. Was a term mismatch so properly truncate now. Still may be some lingering strange states but this seemed to help.

Found some general subscription management issues with leaking subscriptions.

Made shutdown more bearable, and also the reporting on stalled quorums got slowed down for when we have lots of streams and consumers.

/cc @nats-io/core

### DIFF
--- a/server/raft.go
+++ b/server/raft.go
@@ -130,6 +130,7 @@ type raft struct {
 	acks     map[uint64]map[string]struct{}
 	elect    *time.Timer
 	active   time.Time
+	llqrt    time.Time
 	term     uint64
 	pterm    uint64
 	pindex   uint64
@@ -371,13 +372,13 @@ func (s *Server) startRaftNode(cfg *RaftConfig) (RaftNode, error) {
 
 		for index := state.FirstSeq; index <= state.LastSeq; index++ {
 			ae, err := n.loadEntry(index)
-			if ae.pindex != index-1 {
-				n.warn("Corrupt WAL, truncating")
-				n.wal.Truncate(index - 1)
-				break
-			}
 			if err != nil {
 				panic("err loading entry from WAL")
+			}
+			if ae.pindex != index-1 {
+				n.warn("Corrupt WAL, truncating and fixing")
+				n.truncateWal(ae)
+				break
 			}
 			n.processAppendEntry(ae, nil)
 		}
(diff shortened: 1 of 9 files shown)


ripienaar (commented):
LGTM, doesnt change behaviour etc

server/raft.go
```diff
@@ -473,6 +474,28 @@ func (s *Server) reloadDebugRaftNodes() {
...
+		s.Debugf("Shutting down all raft nodes")
+	}
+	for _, n := range s.raftNodes {
+		nodes = append(nodes, n)
+	}
+	s.rnMu.RUnlock()
+
+	for _, node := range nodes {
+		if node.Leader() {
+			node.StepDown()
```
kozlovic, line 493: I guess that this is the right thing to do since you know this server is going away. The "bad" thing is that when one shutdown all servers, then this is likely to cause another server to become leader, etc.. but again, I believe this is the right thing to do in general.

kozlovic (approved):
LGTM

server/raft.go
ripienaar: Yeah, I do see re-elects on shutdown/delete of consumers and streams too as you describe.

Iirc we dont yet have a API to permentantly evict a node, only step-down, we'd need that

derekcollison: By default we want to assume this server going away is a singelton event and try to transfer leadership for any leader based nodes.

We can permanently evict a peer. Just remove JS from its config and do a config reload. JS will remap around it and that peer is considered gone.

So thought there was on ops side.

Reload config with JS disabled.
LDM

ripienaar: If a machine just dies, power off, dead? We need to help things around a bit by manually killing it off?

derekcollison: Yes I remember, I thought we had that already but I could be mistaken.

derekcollison:
Waiting for tests to be green.

server/raft.go
ripienaar: JS can be enabled on a machine not specifically listed in the routes, and will be remembered ofcourse, so suspect we'll need a way to get rid of those that die unexpectedly

PR #1935, opened 2021-02-24
Cross account JetStream pull consumers.

derekcollison:
Reply subject rewrite behaviors when crossing accounts via service imports would break JetStream pull consumers.
For now make special case. Need longer term solution for opt-in or opt-opt of reply subject rewrite.

Also do not store request info headers into JetStream streams.

/cc @nats-io/core

### DIFF
--- a/server/client.go
+++ b/server/client.go
@@ -117,7 +117,10 @@ var readLoopReportThreshold = readLoopReport
 // Represent client booleans with a bitmask
 type clientFlag uint16
 
-const hdrLine = "NATS/1.0\r\n"
+const (
+	hdrLine      = "NATS/1.0\r\n"
+	emptyHdrLine = "NATS/1.0\r\n\r\n"
+)
 
 // Some client state represented as flags
 const (
@@ -3587,7 +3590,11 @@ func removeHeaderIfPresent(hdr []byte, key string) []byte {
 	if end < 0 {
 		return hdr
 	}
-	return append(hdr[:start], hdr[start+end+len(_CRLF_):]...)
+	nhdr := append(hdr[:start], hdr[start+end+len(_CRLF_):]...)
+	if len(nhdr) <= len(emptyHdrLine) {
+		return nil
+	}
+	return nhdr
 }
 
 // Generate a new header based on optional original header and key value.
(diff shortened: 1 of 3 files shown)


server/client.go
```diff
@@ -3587,7 +3590,11 @@ func removeHeaderIfPresent(hdr []byte, key string) []byte {
 	if end < 0 {
 		return hdr
 	}
-	return append(hdr[:start], hdr[start+end+len(_CRLF_):]...)
+	nhdr := append(hdr[:start], hdr[start+end+len(_CRLF_):]...)
```
kozlovic, line 3593: Few things:
- if the hdr was just the empty line (meaning "NATS/1.0\r\n\r\n"), then this function will return that (since the key is not found). However, if we find the key, and it was the only one, then we return nil. That is not consistent.
- Again, as mentioned previously, you should be careful of not removing a key that is not an exact match, that is, key "Test" would remove the key "Testx", which is bad. Now, this is likely ok since we are looking for a server "key", but since we don't enforce usage of header keys from users, at least let's be strict here.

kozlovic (commented):
Not sure about the implementation of "removeHeaderIfPresent".

server/client.go
derekcollison: if we do not find anything we should not take any action, regardless of what was passed in. 
I could pass it back after removal and have upper layers null it out if needed.

For making sure its the right key will update.

matthiashanel: append the separator to key? so we search for "key=" instead of "key"

This still won't help against a key= being some other key's value.

derekcollison: ok please take another look, kept the auto return nil if we remove and this was only header present.

matthiashanel (commented):
see comment on searching for a key

derekcollison:
Updated based on feedback to make sure its the right key.

server/client.go
```diff
@@ -3586,15 +3586,19 @@ func removeHeaderIfPresent(hdr []byte, key string) []byte {
 	if start < 0 {
 		return hdr
 	}
+	index := start + len(key)
+	if index >= len(hdr) || hdr[index] != ':' {
+		return hdr
+	}
```
matthiashanel, line 3592: Is it possible to add hdr[start -1] == '\n' ? 
Keys start at the beginning of a line.

kozlovic (approved):
LGTM

server/client.go
derekcollison: The suffix is ':' which we now check, but agree could have another prefix. Let me try one more time.

server/client.go
derekcollison: ok take one more look, thanks..

kozlovic (approved):
LGTM

server/client.go
derekcollison: Good idea, done.

PR #1936, opened 2021-02-25
Properly handle local deliver subjects

derekcollison:
When pulling consumer messages if the consumer and the client were o the same server we needed to process local deliver subject.

/cc @nats-io/core


kozlovic (approved):
LGTM

PR #1937, opened 2021-02-25
[FIXED] Bug where followers would not snapshot/compact WAL.

derekcollison:
Also tweaked some write/sync params.

/cc @nats-io/core

### DIFF
--- a/server/filestore.go
+++ b/server/filestore.go
@@ -173,7 +173,7 @@ const (
 	// cache idx expiration
 	defaultCacheIdxExpiration = 5 * time.Minute
 	// default sync interval
-	defaultSyncInterval = 10 * time.Second
+	defaultSyncInterval = 60 * time.Second
 	// coalesceMinimum
 	coalesceMinimum = 16 * 1024
 	// maxFlushWait is maximum we will wait to gather messages to flush.
--- a/server/filestore.go
+++ b/server/filestore.go
@@ -1370,6 +1370,7 @@ func (mb *msgBlock) truncate(sm *fileStoredMsg) (nmsgs, nbytes uint64, err error
 		mb.mfd.ReadAt(lchk[:], eof-8)
 		copy(mb.lchk[0:], lchk[:])
 	} else {
+		mb.mu.Unlock()
 		return 0, 0, fmt.Errorf("failed to truncate msg block %d, file not open", mb.index)
 	}
 
@@ -1632,7 +1633,6 @@ func (fs *fileStore) checkMsgs() *LostStreamData {
 // filestore lock will be held.
 func (mb *msgBlock) writeMsgRecord(rl, seq uint64, subj string, mhdr, msg []byte, ts int64, flush bool) error {
 	mb.mu.Lock()
-
 	// Make sure we have a cache setup.
 	if mb.cache == nil {
 		mb.cache = &cache{}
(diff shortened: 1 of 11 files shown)


server/raft.go
```diff
@@ -330,6 +338,8 @@ func (s *Server) startRaftNode(cfg *RaftConfig) (RaftNode, error) {
 		c:        s.createInternalSystemClient(),
 		sendq:    sendq,
 		quit:     make(chan struct{}),
+		wtvch:    make(chan struct{}),
+		wpsch:    make(chan struct{}),
```
kozlovic, line 342: Need to make those at least size 1, no?

server/raft.go
```diff
@@ -2563,22 +2597,63 @@ func (n *raft) readTermVote() (term uint64, voted string, err error) {
...
+	b := buf[:8+len(n.vote)]
+
+	// If same as what we have we can ignore.
+	if bytes.Equal(n.wtv, b) {
+		return
+	}
+	// Stamp latest and kick writer.
+	n.wtv = b
+	select {
+	case n.wtvch <- struct{}{}:
```
kozlovic, line 2654: With being unbuffered, are you ok losing it if the fileWriter routine is not actively receiving?

server/filestore.go
```diff
@@ -173,7 +173,7 @@ const (
 	// cache idx expiration
 	defaultCacheIdxExpiration = 5 * time.Minute
 	// default sync interval
-	defaultSyncInterval = 10 * time.Second
+	defaultSyncInterval = 60 * time.Second
```
kozlovic, line 176: If sync has really a negative impact, this is just moving the issue from 10sec to 60sec.
Wonder if you should not expose the (auto)sync params so users can decide.

server/jetstream_cluster.go
```diff
@@ -1108,7 +1120,7 @@ func (js *jetStream) createRaftGroup(rg *raftGroup) error {
 
 	storeDir := path.Join(js.config.StoreDir, sysAcc.Name, defaultStoreDirName, rg.Name)
 	fs, err := newFileStore(
-		FileStoreConfig{StoreDir: storeDir, BlockSize: 8_000_000, AsyncFlush: true},
+		FileStoreConfig{StoreDir: storeDir, BlockSize: 4_000_000, AsyncFlush: true, SyncInterval: 2 * time.Minute},
```
kozlovic, line 1123: Maybe we should expose through config the sync interval so it can be disabled if performance impact is too much. One could rely on the idea that with R<n> it would need several failures to lose data.

server/raft.go
derekcollison: No always do a select around populating.

server/raft.go
derekcollison: Ah I see what you mean, maybe should make at least 1 buffered.

server/filestore.go
derekcollison: Its is configurable but not all the way to user at this time.

server/jetstream_cluster.go
derekcollison: These are for the internal WAL, so probably would not expose them.

derekcollison:
Please take another look when you can.

kozlovic (approved):
LGTM

kozlovic (approved):
LGTM

kozlovic (commented):
LGTM

PR #1940, opened 2021-02-25
always store the filter subject

ripienaar:
This avoids a situation where a consumer asks for
ORDERS.new today on a stream of ORDERS.new but later
someone makes the same stream ORDERS.* and the new
consumer would then get messages for other purposes
as well

/cc @nats-io/core

### DIFF
--- a/server/consumer.go
+++ b/server/consumer.go
@@ -291,19 +291,7 @@ func (mset *stream) addConsumerWithAssignment(config *ConsumerConfig, oname stri
 
 	// Make sure any partition subject is also a literal.
 	if config.FilterSubject != _EMPTY_ {
-		var checkSubject bool
-
-		mset.mu.RLock()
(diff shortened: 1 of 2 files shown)


server/consumer.go
```diff
@@ -1564,7 +1552,7 @@ func nextReqFromMsg(msg []byte) (time.Time, int, bool, error) {
 		if err := json.Unmarshal(req, &cr); err != nil {
 			return time.Time{}, -1, false, err
 		}
-		return cr.Expires, cr.Batch, cr.NoWait, nil
+		return cr.Expires.UTC(), cr.Batch, cr.NoWait, nil
```
ripienaar, line 1555: well, this crept in, will remove- MH will also remove it soon I bet

derekcollison (approved):
LGTM

PR #1941, opened 2021-02-26
extend the out of space advisory

ripienaar:

/cc @nats-io/core

### DIFF
--- a/server/jetstream.go
+++ b/server/jetstream.go
@@ -249,7 +249,7 @@ func (s *Server) setJetStreamDisabled() {
 	js.mu.Unlock()
 }
 
-func (s *Server) handleOutOfSpace() {
+func (s *Server) handleOutOfSpace(stream string) {
 	if s.JetStreamEnabled() {
 		s.Errorf("JetStream out of space, will be DISABLED")
 		go s.DisableJetStream()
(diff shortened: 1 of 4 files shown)


server/jetstream_events.go
```diff
@@ -210,5 +210,8 @@ const JSServerOutOfStorageAdvisoryType = "io.nats.jetstream.advisory.v1.server_o
 // JSServerOutOfSpaceAdvisory indicates that a stream has lost quorum and is stalled.
 type JSServerOutOfSpaceAdvisory struct {
 	TypedEvent
-	Server string `json:"server"`
+	Server   string `json:"server"`
+	ServerID string `json:"server_id"`
+	Stream   string `json:"stream,omitempty"`
+	Cluster  string `json:"cluster"`
```
ripienaar, line 216: basically just adding enough here so someone can really super easily figure out exactly what server it is.

Since we know which stream caused the failure I figured we might as well add that when possible, this is if people do weird mounting things as per recent calls.

server/jetstream.go
```diff
@@ -260,7 +260,10 @@ func (s *Server) handleOutOfSpace() {
 				ID:   nuid.Next(),
 				Time: time.Now().UTC(),
 			},
-			Server: s.Name(),
+			Server:   s.Name(),
+			ServerID: s.ID(),
+			Stream:   stream,
```
derekcollison, line 265: Not sure this matters, just chance TBH. The whole server is out of space.

derekcollison (approved):
LGTM

server/jetstream.go
ripienaar: Stream? It's there if someone mounts a different volume onto the stream specific storage dir, so not the whole server always.

derekcollison: This advisory signals the whole server shutdown jetstream. We could have the stream name who did it I guess..

PR #1942, opened 2021-02-26
Allow sources and mirrors to not be pre-defined.

derekcollison:
Now will keep trying to establish contact.
Fixed bug where double source headers for sourced messages were causing issues.

/cc @nats-io/core


ripienaar (approved):
LGTM

PR #1947, opened 2021-03-01
Stability Fixes, bug fixes, RC features.

derekcollison:

/cc @nats-io/core


kozlovic (approved):
LGTM.

server/stream.go
```diff
@@ -1069,10 +1160,10 @@ func (mset *stream) setupMirrorConsumer() error {
 			DeliverSubject: string(sub.subject),
 			DeliverPolicy:  DeliverByStartSequence,
 			OptStartSeq:    state.LastSeq,
-			AckPolicy:      AckExplicit,
+			AckPolicy:      AckNone,
 			AckWait:        48 * time.Hour,
 			MaxDeliver:     1,
-			MaxAckPending:  mirrorMaxAckPending,
+			RateLimit:      1 * 1024 * 1024 * 1024, // 1 Gbit
```
ripienaar, line 1166: wont this increase likelyhood of triggering slow consumers on links incapable of handling that?

server/jetstream_api.go
```diff
@@ -1742,6 +1773,73 @@ func (s *Server) jsStreamRemovePeerRequest(sub *subscription, c *client, subject
...
+		resp.Error = jsInvalidJSONErr
+		s.sendAPIErrResponse(ci, acc, subject, reply, string(msg), s.jsonResponse(&resp))
+		return
+	}
+
+	var found string
+	js.mu.RLock()
+	for _, p := range cc.meta.Peers() {
+		si, ok := s.nodeToInfo.Load(p.ID)
+		if ok && si.(*nodeInfo).name == req.Server {
```
ripienaar, line 1821: do we guarantee name is unique? I am thinking for this API we should take a ID and not a name to be 100% sure we get what we want?

server/stream.go
derekcollison: I lowered it and still thinking through better ways we might approach this.

server/jetstream_api.go
derekcollison: That may be a good point, forcing the peer removal by server ID vs name..

server/jetstream_api.go
```diff
@@ -440,6 +448,20 @@ type JSApiLeaderStepDownResponse struct {
...
+	Server string `json:"peer"`
+}
+
+// JSApiLeaderServerRemoveResponse is the response to a metaleader peer removal request.
+type JSApiLeaderServerRemoveResponse struct {
+	ApiResponse
+	Success bool `json:"success,omitempty"`
+}
+
+const JSApiLeaderServerRemovalResponseType = "io.nats.jetstream.api.v1.meta_leader_server_removal"
```
ripienaar, line 463: should be `io.nats.jetstream.api.v1.meta_leader_server_remove_response`

PR #1948, opened 2021-03-01
Stability Updates

derekcollison:
Bug fix for lockup in raft on blocking channels.
Bug fix for off by one bug on out of band catchup.

/cc @nats-io/core


kozlovic (approved):
LGTM

PR #1951, opened 2021-03-02
Enable cross account behaviors for mirrors and sources.

derekcollison:

/cc @nats-io/core

### DIFF
--- a/server/stream.go
+++ b/server/stream.go
@@ -111,10 +111,17 @@ type StreamSourceInfo struct {
 
 // StreamSource dictates how streams can source from other streams.
 type StreamSource struct {
-	Name          string     `json:"name"`
-	OptStartSeq   uint64     `json:"opt_start_seq,omitempty"`
-	OptStartTime  *time.Time `json:"opt_start_time,omitempty"`
-	FilterSubject string     `json:"filter_subject,omitempty"`
+	Name          string          `json:"name"`
+	OptStartSeq   uint64          `json:"opt_start_seq,omitempty"`
+	OptStartTime  *time.Time      `json:"opt_start_time,omitempty"`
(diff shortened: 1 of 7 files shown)


server/stream.go
```diff
@@ -111,10 +111,17 @@ type StreamSourceInfo struct {
...
+	OptStartSeq   uint64          `json:"opt_start_seq,omitempty"`
+	OptStartTime  *time.Time      `json:"opt_start_time,omitempty"`
+	FilterSubject string          `json:"filter_subject,omitempty"`
+	External      *ExternalStream `json:"external,omitempty"`
+}
+
+// ExternalStream allows you to qualify access to a stream source in another account.
+type ExternalStream struct {
+	ApiPrefix     string `json:"api"`
+	DeliverPrefix string `json:"deliver"`
```
ripienaar, line 124: so just chekcing for doc purposes, deliver basically have to be exported as a service and needs to support like `out.>` yeah?

derekcollison: image

ripienaar:
LGTM

server/stream.go
```diff
@@ -1347,6 +1387,11 @@ func (mset *stream) setSourceConsumer(sname string, seq uint64) {
 
 	b, _ := json.Marshal(req)
 	subject = fmt.Sprintf(JSApiDurableCreateT, sname, durable)
+	if ext != nil {
+		subject = strings.Replace(subject, JSApiPrefix, ext.ApiPrefix, 1)
```
kozlovic, line 1391: This code is repeated 4 times, so it could have been worth a function. Even the one with either `$JS.M` or `$JS.S` could have been made a function.

derekcollison: Agree and post GA lots of little cleanup opportunities like that.

PR #1952, opened 2021-03-03
Fixed linter issues

matthiashanel:
```
From top 
accounts.go: 
    se is never nil
    complained about differing receiver names (I only fixed a few of these. To keep the change smaller )
    unnecessary conversion
    unused constant
    acc will be nil
    p is always nil
    complaining about errors starting with lower case (There are more. I only changed one of these)
    receiver name again
    jwt overloaded the package. (there are more than this one, but only fixed a few to keep this change smaller)

auth.go
    regex was bad , is not part of the regex. {n} is only needed when n > 1

client.go
    return is always false
    srv is never nil (scroll up quite far)
    err is never nil  (scroll up quite far)

dirstore.go
    differing receiver name

jetstream_cluster.go
    || threw off the linter thought both conditions where possible
    err is not set after being checked (scroll up quite far) essentially only else branch will be taken
    
monitor.go
    switched to non deprecated value

mqtt.go
    for depends on the if, thus moved for into the if

route.go
    srv.info.Cluster == clusterName. the name only changes if shouldreject == false

server.go
    claims was nil when err != nil
    l == nil if err != nil
```

### DIFF
--- a/server/accounts.go
+++ b/server/accounts.go
@@ -943,9 +943,6 @@ func (a *Account) addServiceExportWithResponseAndAccountPos(
 	}
 
 	if accounts != nil || accountPos > 0 {
-		if se == nil {
-			se = &serviceExport{}
-		}
 		if err := setExportAuth(&se.exportAuth, subject, accounts, accountPos); err != nil {
 			return err
 		}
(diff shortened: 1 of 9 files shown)


kozlovic (approved):
LGTM!

PR #1953, opened 2021-03-03
JetStream API Changes

derekcollison:
In clustered JetStream we need to move API calls out of routes/gateways/leafnodes path.

This moves from explicit imports and subscriptions to one wildcard subscription and a single wildcard export.

/cc @nats-io/core

### DIFF
--- a/server/consumer.go
+++ b/server/consumer.go
@@ -1677,35 +1677,22 @@ func (o *consumer) processNextMsgReq(_ *subscription, c *client, _, reply string
 	// In case we have to queue up this request. This is all on stack pre-allocated.
 	wr := waitingRequest{client: c, reply: reply, n: batchSize, noWait: noWait, expires: expires}
 
-	// If we are in replay mode, defer to processReplay for delivery.
-	if o.replay {
-		o.waiting.add(&wr)
+	if wr.noWait {
+		if o.maxp > 0 && len(o.pending) >= o.maxp {
+			sendErr(409, "Exceeded MaxAckPending")
+		}
 		o.mu.Unlock()
-		o.signalNewMessages()
-		return
-	}
-
-	for i := 0; i < batchSize; i++ {
-		// See if we have more messages available.
-		if subj, hdr, msg, seq, dc, ts, err := o.getNextMsg(); err == nil {
-			o.deliverMsg(reply, subj, hdr, msg, seq, dc, ts)
-			// Need to discount this from the total n for the request.
-			wr.n--
-		} else {
-			if wr.noWait {
-				switch err {
-				case errMaxAckPending:
-					sendErr(409, "Exceeded MaxAckPending")
-				default:
-					sendErr(404, "No Messages")
-				}
-				return
-			}
-			o.waiting.add(&wr)
-			break
+		empty := mset.state().Msgs == 0
+		o.mu.Lock()
+		if empty {
+			sendErr(404, "No Messages")
 		}
 	}
+
+	o.waiting.add(&wr)
 	o.mu.Unlock()
+
+	o.signalNewMessages()
 }
 
 // Increase the delivery count for this message.
--- a/server/consumer.go
+++ b/server/consumer.go
@@ -1680,12 +1680,14 @@ func (o *consumer) processNextMsgReq(_ *subscription, c *client, _, reply string
 	if wr.noWait {
 		if o.maxp > 0 && len(o.pending) >= o.maxp {
 			sendErr(409, "Exceeded MaxAckPending")
+			return
 		}
 		o.mu.Unlock()
 		empty := mset.state().Msgs == 0
 		o.mu.Lock()
 		if empty {
 			sendErr(404, "No Messages")
+			return
 		}
 	}
 
--- a/server/jetstream_api.go
+++ b/server/jetstream_api.go
@@ -667,7 +667,9 @@ func (js *jetStream) apiDispatch(sub *subscription, c *client, subject, reply st
 	}
 
 	// If we are here we can properly dispatch this API call.
-	// Copy the message and the client. Client for the pubArgs.
+	// Copy the message and the client. Client for the pubArgs
+	// but note the JSAPI only uses the hdr index to piece apart
+	// the header from the msg body. No other references are needed.
 	// FIXME(dlc) - Should cleanup eventually and make sending
 	// and receiving internal messages more formal.
 	rmsg = append(rmsg[:0:0], rmsg...)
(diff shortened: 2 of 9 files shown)


server/jetstream_api.go
```diff
@@ -620,7 +625,79 @@ var allJsExports = []string{
...
+		s.Warnf("JetStream API Overloaded: %d calls outstanding", apiOut)
+		return
+	}
+
+	// If we are here we can properly dispatch this API call.
+	// Copy the message and the client. Client for the pubArgs.
+	// FIXME(dlc) - Should cleanup eventually and make sending
+	// and receiving internal messages more formal.
+	rmsg = append(rmsg[:0:0], rmsg...)
+	client := &client{srv: s, kind: JETSTREAM}
```
kozlovic, line 674: This will create additional GC pressure...

server/jetstream_api.go
```diff
@@ -620,7 +625,79 @@ var allJsExports = []string{
...
+	client.pa = c.pa
```
kozlovic, line 675: Are you sure about that? I mean the content will still reference buffer that is going to be overwritten.
`client.pa = c.pa` causes a copy since `pa` is `pubArg` which is a struct, but all `[]byte` inside (may) reference the read buffer. I don't think that this ensuring that those are copied...

server/jetstream_api.go
```diff
@@ -620,7 +625,79 @@ var allJsExports = []string{
...
+
+	// Dispatch the API call to its own Go routine.
+	go func() {
```
kozlovic, line 678: Can't help but think that this is going to cause out of order issues.

server/monitor.go
```diff
@@ -2259,11 +2259,12 @@ type JSInfo struct {
 	Disabled bool            `json:"disabled,omitempty"`
 	Config   JetStreamConfig `json:"config,omitempty"`
 	JetStreamStats
-	StreamCnt    int          `json:"total_streams,omitempty"`
```
kozlovic, line 2262: This is ok to change since I believe that these were added just in main and not in public release.

kozlovic (commented):
Really not sure about the go routine and use of temporary client with "shallow" copy of c.pa.

server/jetstream_api.go
derekcollison: These API calls are low. All high volumes bypass this like sending to streams and getting next messages.

server/jetstream_api.go
derekcollison: JSApi strictly only uses it to piece apart hdr and msg body. In clustered mode can not use or depend on passed in client for any more than that.

server/jetstream_api.go
derekcollison: If you rapid fire a request without a reply and do not wait for the reply. Possible. I am ok with that for now.

server/monitor.go
derekcollison: Yes exactly..

server/jetstream_api.go
kozlovic: Ok, a simple comment would help clear the confusion.

server/consumer.go
```diff
@@ -1677,35 +1677,22 @@ func (o *consumer) processNextMsgReq(_ *subscription, c *client, _, reply string
 	// In case we have to queue up this request. This is all on stack pre-allocated.
 	wr := waitingRequest{client: c, reply: reply, n: batchSize, noWait: noWait, expires: expires}
 
-	// If we are in replay mode, defer to processReplay for delivery.
-	if o.replay {
-		o.waiting.add(&wr)
+	if wr.noWait {
+		if o.maxp > 0 && len(o.pending) >= o.maxp {
+			sendErr(409, "Exceeded MaxAckPending")
```
kozlovic, line 1682: Not clear to me that in this error situation, we still continue. Is that expected?

server/consumer.go
```diff
@@ -1677,35 +1677,22 @@ func (o *consumer) processNextMsgReq(_ *subscription, c *client, _, reply string
...
-					sendErr(404, "No Messages")
-				}
-				return
-			}
-			o.waiting.add(&wr)
-			break
+		empty := mset.state().Msgs == 0
+		o.mu.Lock()
+		if empty {
+			sendErr(404, "No Messages")
```
kozlovic, line 1688: Same than above, and also difficult to understand how we could be both max pending and no message?

server/jetstream_api.go
derekcollison: Will do.

server/consumer.go
derekcollison: Yes, fixed.

server/consumer.go
```diff
@@ -1680,12 +1680,14 @@ func (o *consumer) processNextMsgReq(_ *subscription, c *client, _, reply string
 	if wr.noWait {
 		if o.maxp > 0 && len(o.pending) >= o.maxp {
 			sendErr(409, "Exceeded MaxAckPending")
+			return
```
kozlovic, line 1683: Unlock first?

server/consumer.go
```diff
@@ -1680,12 +1680,14 @@ func (o *consumer) processNextMsgReq(_ *subscription, c *client, _, reply string
...
 		}
 		o.mu.Unlock()
 		empty := mset.state().Msgs == 0
 		o.mu.Lock()
 		if empty {
 			sendErr(404, "No Messages")
+			return
```
kozlovic, line 1690: Unlock

kozlovic (changes requested):

server/consumer.go
kozlovic: Also, same Unlock missing here: https://github.com/nats-io/nats-server/blob/b9498b8c0e0983e9c72eb8f0a3494fb76b1fff02/server/consumer.go#L1673

derekcollison: sendErr does it.

kozlovic: Oh, sendErr is unlocking, right?

kozlovic (approved):
LGTM

server/monitor.go
```diff
@@ -2259,11 +2259,12 @@ type JSInfo struct {
...
-	ConsumerCnt  int          `json:"total_consumers,omitempty"`
-	MessageCnt   uint64       `json:"total_messages,omitempty"`
-	MessageBytes uint64       `json:"total_message_bytes,omitempty"`
-	Meta         *ClusterInfo `json:"meta_cluster,omitempty"`
+	APICalls  int64        `json:"current_api_calls"`
```
ripienaar, line 2262: what does this indicate exactly? Looks like in-flight api calls right?

derekcollison: Yes for the ones we spin up Go routines for, but thinking about it we should account for them regardless of inline or not. Will also add this logic to getNextMsg as well in separate PR.

PR #1954, opened 2021-03-03
extend peer info and cluster info with node IDs

ripienaar:
In order to make meta peer removal safe we prefer to operate on IDs
rather than names, we would therefor need to be able to get the ID
of an offline node, with this change we report the IDs - even of
offline nodes - and so we can build the UI around IDs even for
nodes that are not around any more

/cc @nats-io/core

### DIFF
--- a/server/jetstream_cluster.go
+++ b/server/jetstream_cluster.go
@@ -4143,18 +4143,16 @@ func (js *jetStream) clusterInfo(rg *raftGroup) *ClusterInfo {
 
 	if rg == nil || rg.node == nil {
 		return &ClusterInfo{
-			Name:     s.ClusterName(),
-			Leader:   s.Name(),
-			LeaderID: s.ID(),
+			Name:   s.ClusterName(),
+			Leader: &PeerInfo{Name: s.Name(), ID: s.ID(), Active: 0, Current: true},
 		}
 	}
 	n := rg.node
 
 	leader := n.GroupLeader()
 	ci := &ClusterInfo{
-		Name:     s.ClusterName(),
-		Leader:   s.serverNameForNode(leader),
-		LeaderID: s.serverIdForNode(leader),
+		Name:   s.ClusterName(),
+		Leader: &PeerInfo{Name: s.serverNameForNode(leader), ID: s.serverIdForNode(leader), Active: 0, Current: true},
 	}
 
 	now := time.Now()
--- a/server/jetstream_cluster.go
+++ b/server/jetstream_cluster.go
@@ -1535,7 +1535,7 @@ func (s *Server) replicas(node RaftNode) []*PeerInfo {
 	for _, rp := range node.Peers() {
 		if sir, ok := s.nodeToInfo.Load(rp.ID); ok && sir != nil {
 			si := sir.(*nodeInfo)
-			pi := &PeerInfo{Name: si.name, Current: rp.Current, Active: now.Sub(rp.Last), Offline: si.offline, Lag: rp.Lag}
+			pi := &PeerInfo{Name: si.name, ID: si.id, Current: rp.Current, Active: now.Sub(rp.Last), Offline: si.offline, Lag: rp.Lag}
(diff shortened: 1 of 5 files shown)


server/jetstream_cluster.go
```diff
@@ -4143,15 +4143,17 @@ func (js *jetStream) clusterInfo(rg *raftGroup) *ClusterInfo {
...
 		}
 	}
 	n := rg.node
 
 	ci := &ClusterInfo{
-		Name:   s.ClusterName(),
-		Leader: s.serverNameForNode(n.GroupLeader()),
+		Name:     s.ClusterName(),
+		Leader:   s.serverNameForNode(n.GroupLeader()),
+		LeaderID: s.ID(),
```
ripienaar, line 4156: hmm, I guess this is wrong, will check.

server/jetstream_cluster_test.go
```diff
@@ -4330,7 +4330,7 @@ func TestJetStreamCrossAccountMirrorsAndSources(t *testing.T) {
 		t.Fatalf("Did not receive correct response: %+v", scResp.Error)
 	}
 
-	checkFor(t, 2*time.Second, 100*time.Millisecond, func() error {
+	checkFor(t, 4*time.Second, 100*time.Millisecond, func() error {
```
ripienaar, line 4333: this was flapping

server/jetstream_cluster.go
```diff
@@ -4143,15 +4143,18 @@ func (js *jetStream) clusterInfo(rg *raftGroup) *ClusterInfo {
 
 	if rg == nil || rg.node == nil {
 		return &ClusterInfo{
-			Name:   s.ClusterName(),
-			Leader: s.Name(),
+			Name:     s.ClusterName(),
+			Leader:   s.Name(),
```
derekcollison, line 4147: Maybe make sub struct?

```go
Leader {
 Name: string
 ID: string
}```

derekcollison (changes requested):

server/jetstream_cluster.go
ripienaar: Then I'd rather just make it a `*PeerInfo`? there would be overlap then

ripienaar: Pushed a commit doing that, it's what I wanted from the beginning :) but you didnt like the Current and Active in there that doesnt really mean much for a Leader.  Seems better though than adding a Leader struct thats only a few fields different from PeerInfo

ripienaar:
Chicken and Egg here, nats.go needs an update also before that test will pass, that test wont pass in nats.go because server would first need updating.

derekcollison:
I might suggest just going back to server name and use the audit tool etc.

server/jetstream_cluster.go
ripienaar: > I might suggest just going back to server name and use the audit tool etc.

We don't really know what the roadmap/plan is with audit tool, but we do need to ship peer membership before then

derekcollison:
I mean revert back to just using server name to remove peer vs ID.

ripienaar:
> I mean revert back to just using server name to remove peer vs ID.

Think that's the only option - since even adding a Leader field to PeerInfo is impossable because again we have a chicken and egg between nats.go and the server.  Using nats.go for the tests is quite unfortunate.

We'll need to ensure names are unique then

Will make a new PR

kozlovic:
@ripienaar I may misunderstand, but if it is just a matter of test, there is no need to depend on nats.go. Can't you send requests directly (marshal and then publish)?

ripienaar:
@kozlovic yeah - however in this case, changing Leader from `string` to a struct would cause all attempts in nats.go to decode a StreamInfo to fail.  And we have many many tests using the nats.go stream create/maintain functions.

Thus we have a impossible situation where we cannot update anything its coupled - we'd have to update the vendored nats.go here before updating actual nats.go

kozlovic:
Understood.

PR #1956, opened 2021-03-03
Consumer Updates.

derekcollison:
Make sure to not process next message (batch) requests inline if non client connection.
Added an index to the redeliver queue for large rdq scenarios.
Cleanup service exports and imports for JetStream.

/cc @nats-io/core


kozlovic (approved):
LGTM... still feel uncomfortable with the individual go routines.

derekcollison:
Noted, but after review I did not feel Go routine with channels per account actually buys us anything and actually could provide undesirable behavior.

derekcollison:
One last check on last CL that will error back to a pull request on push based consumer.

ripienaar:
LGTM, I tested a few things manually like offline nodes

kozlovic (approved):
LGTM (you have a conflict to resolve: server/jetstream_cluster_test.go)

PR #1958, opened 2021-03-03
Prevent pull batch larger than max ack pending

wallyqs:
Noticed that when making a batch request, it was possible to ask for a batch larger than ack pending so resulting in redeliveries as part of the batch.

### DIFF
--- a/server/consumer.go
+++ b/server/consumer.go
@@ -1637,6 +1637,11 @@ func (o *consumer) processNextMsgReq(_ *subscription, c *client, _, reply string
 		return
 	}
 
+	if o.maxp > 0 && batchSize > o.maxp {
+		sendErr(409, "Exceeded MaxAckPending")
+		return
+	}
+
 	// In case we have to queue up this request.
 	wr := waitingRequest{client: c, reply: reply, n: batchSize, noWait: noWait, expires: expires}
 
--- a/server/jetstream_test.go
+++ b/server/jetstream_test.go
@@ -8999,6 +8999,23 @@ func TestJetStreamPushConsumersPullError(t *testing.T) {
 	if m.Header.Get("Status") != "409" {
 		t.Fatalf("Expected a 409 status code, got %q", m.Header.Get("Status"))
 	}
+
+	// Should not be possible to ask for more messages than MaxAckPending limit.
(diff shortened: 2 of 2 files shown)


derekcollison (commented):
Should we just adjust here to be under the limit?

wallyqs:
That sounds good to me

wallyqs:
Changed to cap to the number of max ack pending, but then something that I noticed is that the pull request from the client side could be waiting for the expected batch size until the timeout, even though the server is already done sending the messages from the batch request.  In case there was an adjustment, then I think maybe we need to send a 404 of no more messages to the client so that it knows to stop waiting for messages.

derekcollison:
ok so maybe just error like you had originally?

wallyqs:
Yes... For now I reverted to sending a 409 in case the batch request is larger, might be useful to have a signal so that the pull requests stops in case there are no more messages after sometime but that is a separate issue.

derekcollison (approved):
LGTM

wallyqs:
Thanks, included test and adding doc update here: https://github.com/nats-io/jetstream/pull/494

server/consumer.go
```diff
@@ -1677,6 +1677,11 @@ func (o *consumer) processNextMsgReq(_ *subscription, c *client, _, reply string
 		return
 	}
 
+	if o.maxp > 0 && batchSize > o.maxp {
```
derekcollison, line 1680: I think you want to look at current pending + batchSize and if that is > o.maxp no?

wallyqs: Not sure... That seems to change the behavior of redeliveries, gave it a try but cannot get `TestJetStreamPullConsumerMaxAckPendingRedeliveries` to pass since think that would mean that having too many unacked inflight messages prevent making pull requests:

https://github.com/nats-io/nats-server/compare/master...wallyqs:consumer-ack-pending-pull-cap

wallyqs:
Rebased against latest version of the server.

wallyqs:
There is a flapper about stream mirrors but unrelated to this branch, ok to merge for now? 
Updating some of the stream mirror tests here as well: https://github.com/nats-io/nats-server/pull/1963

PR #1959, opened 2021-03-03
Fixed data race with nodeInfo

kozlovic:
Took the approach of storing struct instead of pointer. Of course,
when changing the offline bool from false to true, it means that
we need to call Store again (with same key).

This is based on the assumption that those Load/Store are not too
frequent. Otherwise, we may need to use locking (and keep *nodeInfo)


matthiashanel (approved):
LGTM

derekcollison (approved):
LGTM

PR #1961, opened 2021-03-04
Switch Gateway JS accounts to interest-only mode + some other fixes

kozlovic:
- Fixed the close of a TLS connection which starting Go 1.16
set the deadline to 5 seconds.

- Fixed an issue with setHeader that was causing these error messages
```
=== RUN   TestServiceImportReplyMatchCycleMultiHops
nats: message could not decode headers on connection [4] for subscription on "foo"
--- PASS: TestServiceImportReplyMatchCycleMultiHops (0.04s)
```

- Fixed names of tests in norace_test.go since they must start with
TestNoRace in order to make sure that we execute them in Travis:
```
go test -v -run=TestNoRace --failfast -p=1 ./...
```

### DIFF
--- a/server/gateway.go
+++ b/server/gateway.go
@@ -1072,7 +1072,26 @@ func (c *client) processGatewayInfo(info *Info) {
 		// connect events to switch those accounts into interest only mode.
 		s.mu.Lock()
 		s.ensureGWsInterestOnlyForLeafNodes()
+		js := s.js
 		s.mu.Unlock()
+
+		// Switch JetStream accounts to interest-only mode.
+		if js != nil {
+			var accounts []*Account
+			js.mu.Lock()
+			if len(js.accounts) > 0 {
+				accounts = make([]*Account, 0, len(js.accounts))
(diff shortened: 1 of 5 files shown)


server/gateway.go
```diff
@@ -1072,7 +1072,26 @@ func (c *client) processGatewayInfo(info *Info) {
 		// connect events to switch those accounts into interest only mode.
 		s.mu.Lock()
 		s.ensureGWsInterestOnlyForLeafNodes()
+		js := s.js
 		s.mu.Unlock()
+
+		// Switch JetStream accounts to interest-only mode.
+		if js != nil {
```
kozlovic, line 1079: I should do that only if s.gateway.enabled is true, let me fix that.

kozlovic: Well, clearly it is enabled if we process an inbound GW, so no need to change anything :-)

derekcollison (approved):
LGTM - Thanks!

PR #1962, opened 2021-03-04
Various bug fixes and improvements to out of disk handling.

derekcollison:
Better handling on out of disk.
Suppress some stream and consumer bad results since they delete the asset.
Allow rehup to re-enable JetStream.
Changes to handle short writes. Bug fixes to truncation.
Fixed leaking consumer info sub
    
/cc @nats-io/core

### DIFF
--- a/server/jetstream.go
+++ b/server/jetstream.go
@@ -127,13 +127,11 @@ type jsaUsage struct {
 
 // EnableJetStream will enable JetStream support on this server with the given configuration.
 // A nil configuration will dynamically choose the limits and temporary file storage directory.
-// If this server is part of a cluster, a system account will need to be defined.
 func (s *Server) EnableJetStream(config *JetStreamConfig) error {
-	s.mu.Lock()
-	if s.js != nil {
-		s.mu.Unlock()
+	if s.JetStreamEnabled() {
 		return fmt.Errorf("jetstream already enabled")
 	}
+
 	s.Noticef("Starting JetStream")
 	if config == nil || config.MaxMemory <= 0 || config.MaxStore <= 0 {
 		var storeDir string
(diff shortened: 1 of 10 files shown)


server/jetstream_cluster_test.go
```diff
@@ -2995,19 +2995,15 @@ func TestJetStreamClusterPeerOffline(t *testing.T) {
 		var found bool
 		for _, s := range ml.JetStreamClusterPeers() {
 			if s == rs.Name() {
-				if shouldBeOffline {
```
kozlovic, line 2998: I think we need to keep this, however, maybe move this for-loop, etc.. inside the checkFor() (and therefore replace t.Fatalf() with return fmt.Errorf()).

server/jetstream.go
```diff
@@ -276,41 +292,28 @@ func (s *Server) DisableJetStream() error {
...
-				case <-maxWait.C:
-					break LOOP
-				case <-t.C:
-					js.mu.RLock()
-					isCurrent := cc.isCurrent()
-					js.mu.RUnlock()
-					if isCurrent {
-						break LOOP
-					}
+				case <-time.After(2 * time.Second):
```
kozlovic, line 308: So it will always wait for 2secs here, there is no notification that it no longer needs to wait, right?

server/jetstream.go
```diff
@@ -276,41 +292,28 @@ func (s *Server) DisableJetStream() error {
...
+				if !s.JetStreamIsCurrent() {
+					s.Warnf("JetStream timeout waiting for meta leader transfer")
 				}
 			}
-		}
-		// Once here we can forward our proposal to remove ourselves.
-		if meta != nil {
+			// Once here we can forward our proposal to remove ourselves.
 			meta.ProposeRemovePeer(meta.ID())
+			time.Sleep(250 * time.Millisecond)
```
kozlovic, line 316: Seem a bit arbitrary...

kozlovic (changes requested):

server/jetstream_cluster_test.go
derekcollison: That just checks that if we expect it to be there its there immediately regardless of offline status.

server/jetstream.go
derekcollison: Yes this is for the case when we are the metaleader and are going, so arbitrary to wait for the remappings to be distributed and committed.

derekcollison: BTW this is optimistic, meaning the system is operational if we fail to remap on our way out the door.

server/jetstream_cluster_test.go
kozlovic: I added this test yesterday. The idea of `shouldBeOffline` means that I want to fail if I find it in the list, because it should not be. The issue maybe is that it should be retried (hence the move to checkFor).
Removing this code means that you are going to fail only if it should be present and is not, but will not capture the fact that it is present when it should not.

kozlovic (approved):
LGTM

PR #1964, opened 2021-03-04
Raft changes and inbound stream msgs out of fastpath for non-clients.

derekcollison:
Move raft comms off of internal sendq. Move our stream msg processing out of Go routines servicing routes and gateways.

/cc @nats-io/core

### DIFF
--- a/server/client.go
+++ b/server/client.go
@@ -271,6 +271,8 @@ type client struct {
 	trace bool
 	echo  bool
 
+	internal bool
+
 	tags    jwt.TagList
 	nameTag string
 }
@@ -562,6 +564,7 @@ func (c *client) initClient() {
 
 	c.subs = make(map[string]*subscription)
 	c.echo = true
+	c.internal = true
 
 	c.setTraceLevel()
 
@@ -3026,7 +3029,7 @@ func (c *client) deliverMsg(sub *subscription, subject, reply, mh, msg []byte, g
 	client.outBytes += msgSize
 
 	// Check for internal subscriptions.
-	if sub.icb != nil {
+	if sub.icb != nil && c.internal {
 		if gwrply {
 			// Note that we keep track of the GW routed reply in the destination
 			// connection (`client`). The routed reply subject is in `c.pa.reply`,
--- a/server/client.go
+++ b/server/client.go
@@ -270,8 +270,7 @@ type client struct {
 
 	trace bool
 	echo  bool
-
-	internal bool
+	noIcb bool
 
 	tags    jwt.TagList
 	nameTag string
@@ -564,7 +563,6 @@ func (c *client) initClient() {
 
 	c.subs = make(map[string]*subscription)
 	c.echo = true
-	c.internal = true
 
 	c.setTraceLevel()
 
@@ -3029,7 +3027,7 @@ func (c *client) deliverMsg(sub *subscription, subject, reply, mh, msg []byte, g
 	client.outBytes += msgSize
 
 	// Check for internal subscriptions.
-	if sub.icb != nil && c.internal {
+	if sub.icb != nil && !c.noIcb {
 		if gwrply {
 			// Note that we keep track of the GW routed reply in the destination
 			// connection (`client`). The routed reply subject is in `c.pa.reply`,
--- a/server/consumer.go
+++ b/server/consumer.go
@@ -187,7 +187,7 @@ type consumer struct {
 	ackSubj           string
 	nextMsgSubj       string
 	maxp              int
-	sendq             chan *jsPubMsg
+	outq              *jsOutQ
 	pending           map[uint64]*Pending
 	ptmr              *time.Timer
 	rdq               []uint64
@@ -623,7 +623,7 @@ func (o *consumer) setLeader(isLeader bool) {
 		}
 
 		mset.mu.RLock()
-		s, jsa, stream := mset.srv, mset.jsa, mset.cfg.Name
+		s, jsa, stream, outq := mset.srv, mset.jsa, mset.cfg.Name, mset.outq
 		mset.mu.RUnlock()
 
 		o.mu.Lock()
@@ -670,8 +670,9 @@ func (o *consumer) setLeader(isLeader bool) {
 			o.replay = true
 		}
 
-		// Recreate internal sendq
-		o.sendq = make(chan *jsPubMsg, msetSendQSize)
+		// We borrow this from the mset.
+		o.outq = outq
+
 		// Recreate quit channel.
 		o.qch = make(chan struct{})
 		qch := o.qch
@@ -679,8 +680,6 @@ func (o *consumer) setLeader(isLeader bool) {
 
 		// Now start up Go routine to deliver msgs.
 		go o.loopAndGatherMsgs(qch)
-		// Startup our deliver loop.
-		go o.loopAndDeliverMsgs(qch)
 
 	} else {
 		// Shutdown the go routines and the subscriptions.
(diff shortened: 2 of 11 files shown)


server/sendq.go
```diff
@@ -0,0 +1,134 @@
...
+		msg = append(msg[:0:0], msg...)
+		out.msg = msg
+	}
+	sq.mu.Lock()
+
+	sq.msgs++
+	sq.bytes += len(subj) + len(rply) + len(hdr) + len(msg)
+
+	if sq.head == nil {
+		sq.head = out
```
kozlovic, line 120: Capture this event and notify only when adding first element in the list.

server/sendq.go
```diff
@@ -0,0 +1,134 @@
...
+	} else {
+		sq.tail.next = out
+	}
+	sq.tail = out
+	msgs, bytes := sq.msgs, sq.bytes
+	sq.mu.Unlock()
+
+	select {
+	case sq.mch <- struct{}{}:
```
kozlovic, line 129: Could notify only when adding first element.

server/stream.go
```diff
@@ -1909,10 +1913,63 @@ func (mset *stream) isClustered() bool {
...
+	} else {
+		mset.msgs.tail.next = m
+	}
+	mset.msgs.tail = m
+	mset.msgs.msgs++
+	mset.msgs.bytes += len(subj) + len(rply) + len(hdr) + len(msg)
+	mset.mu.Unlock()
+
+	select {
+	case mset.mch <- struct{}{}:
```
kozlovic, line 1966: Same comment than for other list.

server/sendq.go
```diff
@@ -0,0 +1,134 @@
...
+					c.pa.hdb = nil
+					msg = append(pm.msg, _CRLF_...)
+				}
+				c.processInboundClientMsg(msg)
+				c.pa.szb = nil
+				// Do this here to nil out below vs up in for loop.
+				next := pm.next
+				pm.next, pm.hdr, pm.msg = nil, nil, nil
+				if pm = next; pm == nil {
+					pm = sq.pending()
```
kozlovic, line 86: Should not be necessary since we would be notified that there is added messages to the list if we were going back to select. This would also give a chance to select on quitCh...

server/stream.go
```diff
@@ -2287,6 +2353,21 @@ func (mset *stream) internalSendLoop() {
...
+				if isClustered {
+					mset.processClusteredInboundMsg(im.subj, im.rply, im.hdr, im.msg)
+				} else {
+					mset.processJetStreamMsg(im.subj, im.rply, im.hdr, im.msg, 0, 0)
+				}
+				// Do this here to nil out below vs up in for loop.
+				next := im.next
+				im.next, im.hdr, im.msg = nil, nil, nil
+				if im = next; im == nil {
+					im = mset.pending()
```
kozlovic, line 2368: Same comment than before, and even more important here since you would possibly starve the other channel you are receiving on.

server/sendq.go
```diff
@@ -0,0 +1,134 @@
...
+	next *outMsg
+}
+
+type sendq struct {
+	mu    sync.Mutex
+	mch   chan struct{}
+	head  *outMsg
+	tail  *outMsg
+	s     *Server
+	msgs  int
```
kozlovic, line 36: Does not look like you are doing anything with msgs and bytes, so maybe remove?

server/sendq.go
```diff
@@ -0,0 +1,134 @@
...
+			c.flushClients(10 * time.Millisecond)
+		}
+	}
+}
+
+func (sq *sendq) pending() *outMsg {
+	sq.mu.Lock()
+	head := sq.head
+	sq.head, sq.tail = nil, nil
+	sq.msgs, sq.bytes = 0, 0
```
kozlovic, line 98: Those are incremented and then set to 0 here, and returned by send() but nobody using those.

server/stream.go
```diff
@@ -1909,10 +1913,63 @@ func (mset *stream) isClustered() bool {
...
+	hdr  []byte
+	msg  []byte
+	next *inMsg
+}
+
+// Linked list for inbound messages.
+type inbound struct {
+	head  *inMsg
+	tail  *inMsg
+	msgs  int
```
kozlovic, line 1929: Same than above, I am not sure msgs/bytes are used.

kozlovic (changes requested):
Concerned about possible "starvation" of the select{} if you keep dequeuing from the pending list without going back to select.

server/client.go
```diff
@@ -271,6 +271,8 @@ type client struct {
 	trace bool
 	echo  bool
 
+	internal bool
```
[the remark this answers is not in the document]
derekcollison: Will do.

server/sendq.go
derekcollison: I thought about that then was a bit nervous we may miss a kick. Will give it a shot though, good idea.

server/stream.go
derekcollison: Will remove this one.

server/sendq.go
derekcollison: I track them but do not act on them atm.. Will remove.

kozlovic (approved):
Changes LGTM.

server/consumer.go
```diff
@@ -737,14 +736,14 @@ func (o *consumer) unsubscribe(sub *subscription) {
...
 // Do all advisory sends here.
 // Lock should be held on entry but will be released.
 func (o *consumer) sendAdvisory(subj string, msg []byte) {
-	sendq := o.sendq
+	outq := o.outq
 	o.mu.Unlock()
-	if sendq != nil {
-		sendq <- &jsPubMsg{subj, subj, _EMPTY_, nil, msg, nil, 0}
+	if outq != nil {
+		outq.send(&jsPubMsg{subj, subj, _EMPTY_, nil, msg, nil, 0, nil})
```
kozlovic, line 746: Since this will never block, you don't have to release lock anymore (still need to check for `outq != nil` since you set this to nil when stopping).
I am sure if we were using this everywhere (and not releasing lock to "send") then it may remove some obscure bugs that are due to unlocking/locking where state may have changed.

server/consumer.go
```diff
@@ -1645,18 +1577,17 @@ func (o *consumer) processNextMsgReq(_ *subscription, c *client, _, reply string
...
+	outq := o.outq
 
 	sendErr := func(status int, description string) {
 		// Needs to be unlocked to send err.
 		o.mu.Unlock()
 		defer o.mu.Lock()
 		hdr := []byte(fmt.Sprintf("NATS/1.0 %d %s\r\n\r\n", status, description))
-		pmsg := &jsPubMsg{reply, reply, _EMPTY_, hdr, nil, nil, 0}
-		sendq <- pmsg // Send error message.
+		outq.send(&jsPubMsg{reply, reply, _EMPTY_, hdr, nil, nil, 0, nil})
```
kozlovic, line 1590: Not going to repeat that everywhere, but here again you would not need the "unlock/defer lock" in this function, etc..

kozlovic (approved):
LGTM. Could go an extra step and remove the unlock/lock that we had because of sending to go channel that could have been full.

server/consumer.go
derekcollison: Good idea, will also not nil it out anymore and set at construction.

kozlovic (approved):
LGTM. Again, if we generalize that (or other method that gets us away from the buffered channel method) then we can remove some of the unlock/lock that we have.

kozlovic (commented):
LGTM

PR #1965, opened 2021-03-05
Push consumer idle heartbeats

derekcollison:
Allow an option to push based consumers to have idle heartbeats delivered.
This allows an endpoint to know the consumer is still alive.
    
/cc @nats-io/core

### DIFF
--- a/server/consumer.go
+++ b/server/consumer.go
@@ -1958,7 +1958,7 @@ func (o *consumer) loopAndGatherMsgs(qch chan struct{}) {
 		case <-mch:
 			// Messages are waiting.
 		case <-hbc:
-			hdr := []byte("NATS/1.0 200 Idle Heartbeat\r\n\r\n")
+			hdr := []byte("NATS/1.0 100 Idle Heartbeat\r\n\r\n")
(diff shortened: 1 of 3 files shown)


server/consumer.go
```diff
@@ -1926,6 +1957,9 @@ func (o *consumer) loopAndGatherMsgs(qch chan struct{}) {
 			return
 		case <-mch:
 			// Messages are waiting.
+		case <-hbc:
+			hdr := []byte("NATS/1.0 200 Idle Heartbeat\r\n\r\n")
```
ripienaar, line 1961: 200 seems weird, general optics is that it means things are done, no std code seems to fit the bill though

ripienaar: Actually 202 Accepted seems better ```The request has been accepted for processing, but the processing has not been completed. The request might or might not be eventually acted upon, and may be disallowed when processing occurs.```

derekcollison: I wanted to make sure folks did not think it was an error.. ;) Let me look for something better.

derekcollison: Will go with 100 Continue.. Should I update description to Continue?

ripienaar (approved):
LGTM

PR #1968, opened 2021-03-05
Replace listers with receive only channels

variadico:
This uses the following two APIs from JSM instead of the listers, which have been removed.

```
type JetStreamManager interface {
	// StreamsInfo can be used to retrieve a list of StreamInfo objects.
	StreamsInfo(ctx context.Context) <-chan *StreamInfo

	// ConsumersInfo is used to retrieve a list of ConsumerInfo objects.
	ConsumersInfo(ctx context.Context, stream string) <-chan *ConsumerInfo
}
```

Waiting on https://github.com/nats-io/nats.go/pull/659 to be merged first. Then I can update this branch.

### DIFF
--- a/server/jetstream_cluster_test.go
+++ b/server/jetstream_cluster_test.go
@@ -15,6 +15,7 @@ package server
 
 import (
 	"bytes"
+	"context"
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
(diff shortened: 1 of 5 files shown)


server/jetstream_cluster_test.go
```diff
@@ -1088,15 +1089,16 @@ func TestJetStreamClusterStreamInfoList(t *testing.T) {
 	sendBatch("baz", 33)
 
 	// Now get the stream list info.
-	sl := js.NewStreamLister()
-	if !sl.Next() {
-		t.Fatalf("Unexpected error: %v", sl.Err())
+	ctx, cancel := context.WithTimeout(context.Background(), 3*time.Second)
+	var infos []*nats.StreamInfo
+	for info := range js.StreamsInfo(ctx) {
```
wallyqs, line 1094: just passing `nil` is fine no? to use the default timeout from the JS context

```suggestion
	for info := range js.StreamsInfo(nil) {
```

server/jetstream_cluster_test.go
```diff
@@ -1168,15 +1170,16 @@ func TestJetStreamClusterConsumerInfoList(t *testing.T) {
 	}
 
 	// Now get the consumer list info.
-	cl := js.NewConsumerLister("TEST")
-	if !cl.Next() {
-		t.Fatalf("Unexpected error: %v", cl.Err())
+	ctx, cancel := context.WithTimeout(context.Background(), 3*time.Second)
+	var infos []*nats.ConsumerInfo
+	for info := range js.ConsumersInfo(ctx, "TEST") {
```
wallyqs, line 1175:
```suggestion
	for info := range js.ConsumersInfo(nil, "TEST") {
```

wallyqs: to simplify the testing code we should use the `nil` default and rely on the iterator/page stop

variadico:
Closing in favor of https://github.com/nats-io/nats-server/pull/2022

PR #1969, opened 2021-03-05
Fix check from next request msg

wallyqs:
Found that the following panic can happen with `m.Respond([]byte("+NXT"))`.

```
panic: runtime error: slice bounds out of range [:57] with capacity 3

goroutine 6441 [running]:
github.com/nats-io/nats-server/v2/server.(*client).msgParts(...)
	/Users/wallyqs/repos/nats-dev/pkg/mod/github.com/nats-io/nats-server/v2@v2.1.8-0.20210305172713-34f66748770a/server/client.go:2136
github.com/nats-io/nats-server/v2/server.(*consumer).processNextMsgReq(0xc001a4c900, 0x0, 0xc000abe600, 0xc001410870, 0x2e, 0xc000a47c38, 0x11, 0xc0002f69fd, 0x0, 0x3)
	/Users/wallyqs/repos/nats-dev/pkg/mod/github.com/nats-io/nats-server/v2@v2.1.8-0.20210305172713-34f66748770a/server/consumer.go:1596 +0x835
```

/cc @nats-io/core

### DIFF
--- a/server/consumer.go
+++ b/server/consumer.go
@@ -1036,7 +1036,16 @@ func (o *consumer) processAck(_ *subscription, c *client, subject, reply string,
 		o.ackMsg(sseq, dseq, dc)
 	case bytes.HasPrefix(msg, AckNext):
 		o.ackMsg(sseq, dseq, dc)
+		// processNextMsgReq can be invoked from an internal subscription or from here.
+		// Therefore, it has to call msgParts(), so we can't simply pass msg[len(AckNext):]
+		// with current c.pa.hdr because it would cause a panic.  We will save the current
+		// c.pa.hdr value and disable headers before calling processNextMsgReq and then
+		// restore so that we don't mess with the calling stack in case it is used
+		// somewhere else.
+		phdr := c.pa.hdr
+		c.pa.hdr = -1
 		o.processNextMsgReq(nil, c, subject, reply, msg[len(AckNext):])
+		c.pa.hdr = phdr
 		skipAckReply = true
 	case bytes.Equal(msg, AckNak):
 		o.processNak(sseq, dseq)
--- a/server/jetstream_cluster_test.go
+++ b/server/jetstream_cluster_test.go
@@ -4480,6 +4480,28 @@ func TestJetStreamClusterJSAPIImport(t *testing.T) {
 	if meta.Consumer != 1 || meta.Stream != 1 || meta.Delivered != 1 || meta.Pending != 0 {
 		t.Fatalf("Bad meta: %+v", meta)
 	}
+
+	js.Publish("TEST", []byte("Second"))
+	js.Publish("TEST", []byte("Third"))
+
+	// Ack across accounts.
+	m, err = nc.Request("$JS.API.CONSUMER.MSG.NEXT.TEST.tr", []byte("+NXT"), 2*time.Second)
+	if err != nil {
+		t.Fatalf("Unexpected error: %v", err)
+	}
+	meta, err = m.MetaData()
+	if err != nil {
+		t.Fatalf("Unexpected error: %v", err)
+	}
+	if meta.Consumer != 2 || meta.Stream != 2 || meta.Delivered != 1 || meta.Pending != 1 {
+		t.Fatalf("Bad meta: %+v", meta)
+	}
+
+	// AckNext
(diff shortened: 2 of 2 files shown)


server/client.go
```diff
@@ -2132,7 +2132,7 @@ func (c *client) processPong() {
 
 // Will return the parts from the raw wire msg.
 func (c *client) msgParts(data []byte) (hdr []byte, msg []byte) {
-	if c != nil && c.pa.hdr > 0 {
+	if c != nil && c.pa.hdr > 0 && len(data) > 0 {
```
kozlovic, line 2135: @derekcollison Well, my understanding is that this is worrisome because if `c.pa.hdr > 0` then data should not be empty because data should contain both header and payload.
I think the issue is (o *consumer) processAck() that calls processNextMsgReq() with msg[len(AckNext):]. That is wrong, we need to differentiate processNextMsgReq() when coming from processAck and from subscribeInternal().

kozlovic (changes requested):

server/client.go
derekcollison: Correct something is off there, this is more of a bandaid.

wallyqs: specifically, this is happening when sending AckNext on a message that was pulled from another account as in this test: https://github.com/nats-io/nats.go/blob/master/test/js_test.go#L1729-L1730

kozlovic: I would propose doing that instead:
```
index 0055f170..bb563c46 100644
--- a/server/client.go
+++ b/server/client.go
@@ -2132,7 +2132,7 @@ func (c *client) processPong() {
 
 // Will return the parts from the raw wire msg.
 func (c *client) msgParts(data []byte) (hdr []byte, msg []byte) {
-       if c != nil && c.pa.hdr > 0 && len(data) > 0 {
+       if c != nil && c.pa.hdr > 0 {
                return data[:c.pa.hdr], data[c.pa.hdr:]
        }
        return nil, data
diff --git a/server/consumer.go b/server/consumer.go
index 33ea9be3..18507c66 100644
--- a/server/consumer.go
+++ b/server/consumer.go
@@ -1036,7 +1036,10 @@ func (o *consumer) processAck(_ *subscription, c *client, subject, reply string,
                o.ackMsg(sseq, dseq, dc)
        case bytes.HasPrefix(msg, AckNext):
                o.ackMsg(sseq, dseq, dc)
+               phdr := c.pa.hdr
+               c.pa.hdr = -1
                o.processNextMsgReq(nil, c, subject, reply, msg[len(AckNext):])
+               c.pa.hdr = phdr
                skipAckReply = true
        case bytes.Equal(msg, AckNak):
                o.processNak(sseq, dseq)
```
I believe that processNextMsgReq() ignores the header after calling msgParts(), so just let trick it into thinking there is no header when called from processAck?

server/consumer.go
```diff
@@ -1036,7 +1036,10 @@ func (o *consumer) processAck(_ *subscription, c *client, subject, reply string,
 		o.ackMsg(sseq, dseq, dc)
 	case bytes.HasPrefix(msg, AckNext):
 		o.ackMsg(sseq, dseq, dc)
+		phdr := c.pa.hdr
```
kozlovic, line 1039: Let's add a comment here so we understand why we are doing this:
```
// processNextMsgReq can be invoked from an internal subscription or from here.
// Therefore, it has to call msgParts(), so we can't simply pass msg[len(AckNext):]
// with current c.pa.hdr because it would cause a panic. We will save the current
// c.pa.hdr value and disable headers before calling processNextMsgReq and then
// restore so that we don't mess with the calling stack in case it is used somewhere else.
c.pa.hdr = -1
...
```

kozlovic (changes requested):

server/consumer.go
wallyqs: Thanks, added the comment.

server/jetstream_cluster_test.go
```diff
@@ -4816,15 +4838,17 @@ var jsClusterImportsTempl = `
 			jetstream: enabled
 			users = [ { user: "rip", pass: "pass" } ]
 			exports [
-				{ service: "$JS.API.>" }
+				{ service: "$JS.API.>", response: stream }
```
wallyqs, line 4841: These updates to the test were needed so that the second message gets delivered.

kozlovic (approved):
LGTM

PR #1971, opened 2021-03-06
Rework Mirrors and Sources

derekcollison:
Take advantage of consumer changes.

/cc @nats-io/core


server/jetstream_cluster.go
```diff
@@ -1254,8 +1254,8 @@ func (js *jetStream) monitorStream(mset *stream, sa *streamAssignment) {
 
 	const (
 		compactInterval = 2 * time.Minute
-		compactSizeMin  = 32 * 1024 * 1024
-		compactNumMin   = 4
+		compactSizeMin  = 64 * 1024 * 1024
```
kozlovic, line 1257: During our zoom and working on this, I said that I believe it was set to 64MB, not 32MB, but the diff shows that I was wrong. So we could change it back to 32MB.

server/jetstream_cluster.go
```diff
@@ -1315,8 +1315,8 @@ func (js *jetStream) monitorStream(mset *stream, sa *streamAssignment) {
 			if err := js.applyStreamEntries(mset, ce, isRecovering); err == nil {
 				n.Applied(ce.Index)
 				ne := ce.Index - lastApplied
-				// If over our compact min and we have at least min entries to compact, go ahead and snapshot/compact.
-				if _, b := n.Size(); b > compactSizeMin && ne >= compactNumMin {
+				// If we have at least min entries to compact, go ahead and snapshot/compact.
+				if ne >= compactNumMin {
```
kozlovic, line 1319: The `monitorConsumer()` has the same logic and compacts if size is above 4MB actually. Knowing that we saw cases where stream was compacted way too often (size was wrong metric and count was set to 4 instead of higher number), should we do the same there?

kozlovic (approved):
LGTM

PR #1972, opened 2021-03-06
monitorConsumer perform snapshot similar to monitorStream

kozlovic:
Changed the stream min size default value back to 32MB and removed
the one for consumer since we don't use it anymore but set the
count size same than for stream (8192).


derekcollison (approved):
LGTM - Thanks!

PR #1973, opened 2021-03-06
Skip TERM signal during lame duck mode

wallyqs:
In some platforms like K8S / Docker, the TERM signal is sent after calling `docker stop` / `kubectl delete pod`, letting the process terminate for some time before stopping the container.  

In order to terminate gracefully in NATS, it is needed to first send an USR2 signal to the server so that it enters lame duck mode before exiting. For example in k8s the `preStop` command can be used to send the signal:

```yaml
# Gracefully stop NATS Server on pod deletion or image upgrade.
#
lifecycle:
  preStop:
     exec:
        command: ["/bin/sh", "-c", "/nats-server -sl=ldm=/var/run/nats/nats.pid"]
```

But soon after sending the signal, K8S will send a `TERM` signal and abort the LDM graceful stop. In this PR, we add a check to the TERM signal handler so that LDM is able to wrap up and exit on its own before being forced to terminate (like `terminationGracePeriodSeconds` in K8S).

/cc @nats-io/core

### DIFF
--- a/server/const.go
+++ b/server/const.go
@@ -29,6 +29,7 @@ const (
 
 	// private for now
 	commandLDMode = Command("ldm")
+	commandTerm   = Command("term")
 )
 
 var (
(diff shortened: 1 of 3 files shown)


ripienaar:
LGTM

kozlovic (approved):
LGTM

PR #1974, opened 2021-03-06
Memory cache for commits and WAL size constraints.

derekcollison:

/cc @nats-io/core

### DIFF
--- a/server/raft.go
+++ b/server/raft.go
@@ -130,6 +130,7 @@ type raft struct {
 	qn       int
 	peers    map[string]*lps
 	acks     map[uint64]map[string]struct{}
+	pae      map[uint64]*appendEntry
 	elect    *time.Timer
 	active   time.Time
 	llqrt    time.Time
@@ -332,6 +333,7 @@ func (s *Server) startRaftNode(cfg *RaftConfig) (RaftNode, error) {
 		hash:     hash,
 		peers:    make(map[string]*lps),
 		acks:     make(map[uint64]map[string]struct{}),
+		pae:      make(map[uint64]*appendEntry),
 		s:        s,
 		c:        s.createInternalSystemClient(),
 		sq:       sq,
@@ -1858,22 +1860,13 @@ func (n *raft) applyCommit(index uint64) error {
 		delete(n.acks, index)
 	}
 
-	// FIXME(dlc) - Can keep this in memory if this too slow.
-	ae, err := n.loadEntry(index)
-	if err != nil {
-		state := n.wal.State()
-		if index < state.FirstSeq {
-			return nil
-		}
-		if err != ErrStoreClosed && err != ErrStoreEOF {
-			if err == errBadMsg {
-				n.setWriteErrLocked(err)
-			}
-			n.warn("Got an error loading %d index: %v", index, err)
-		}
+	ae := n.pae[index]
+	if ae == nil {
+		n.warn("Could not load append entry %d", index)
 		n.commit = original
 		return errEntryLoadFailed
 	}
+	delete(n.pae, index)
 	ae.buf = nil
 
 	var committed []*Entry
@@ -2362,6 +2355,12 @@ func (n *raft) processAppendEntry(ae *appendEntry, sub *subscription) {
 				n.Unlock()
 				return
 			}
+			// Save in memory for faster processing during applyCommit.
+			n.pae[n.pindex] = ae
+			if len(n.pae) > paeWarnThreshold {
+				n.warn("%d append entries pending", len(n.pae))
+			}
+
 		} else {
 			// This is a replay on startup so just take the appendEntry version.
 			n.pterm = ae.term
@@ -2506,6 +2505,8 @@ func (n *raft) storeToWAL(ae *appendEntry) error {
 	return nil
 }
 
+const paeWarnThreshold = 1024
+
 func (n *raft) sendAppendEntry(entries []*Entry) {
 	n.Lock()
 	defer n.Unlock()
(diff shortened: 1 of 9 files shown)


server/raft.go
```diff
@@ -1858,22 +1878,13 @@ func (n *raft) applyCommit(index uint64) error {
...
-			return nil
-		}
-		if err != ErrStoreClosed && err != ErrStoreEOF {
-			if err == errBadMsg {
-				n.setWriteErrLocked(err)
-			}
-			n.warn("Got an error loading %d index: %v", index, err)
-		}
+	ae := n.pae[index]
+	if ae == nil {
```
kozlovic, line 1882: I assumed that you would have had a cache (not expected to maintain ALL of them), so on cache miss you would resort to call loadEntry.
Testing with this branch I see quickly the warning down here and mirrors do not complete...

server/raft.go
```diff
@@ -1858,22 +1878,13 @@ func (n *raft) applyCommit(index uint64) error {
...
+		n.warn("Could not load append entry %d", index)
 		n.commit = original
 		return errEntryLoadFailed
 	}
+	delete(n.pae, index)
```
kozlovic, line 1887: I experimented by adding a "deleted pae" map here (that is, keeping track of the index that we delete here) and when above we have `ae==nil` condition, do a lookup on that the deleted map and I found it. So it means that either there is a bug somewhere, or is it normal to have this function invoked with the same index.

server/raft.go
```diff
@@ -2362,6 +2373,12 @@ func (n *raft) processAppendEntry(ae *appendEntry, sub *subscription) {
 				n.Unlock()
 				return
 			}
+			// Save in memory for faster processing during applyCommit.
+			n.pae[n.pindex] = ae
+			if len(n.pae) > paeWarnThreshold {
+				n.warn("%d append entries pending", len(n.pae))
```
kozlovic, line 2379: Those warnings show up very quickly, so if it was not expected, something is wrong.

server/raft.go
```diff
@@ -2520,6 +2539,12 @@ func (n *raft) sendAppendEntry(entries []*Entry) {
 		// We count ourselves.
 		n.acks[n.pindex] = map[string]struct{}{n.id: struct{}{}}
 		n.active = time.Now()
+
+		// Save in memory for faster processing during applyCommit.
+		n.pae[n.pindex] = ae
+		if len(n.pae) > paeWarnThreshold {
+			n.warn("%d append entries pending", len(n.pae))
```
kozlovic, line 2546: See comment above.

kozlovic (changes requested):
Quickly get the "could not load append entry" and mirror does not complete.

server/raft.go
derekcollison: Just for the ones we have outstanding waiting to commit but I see the error loading sometimes too so will make some changes.

derekcollison: Made some changes.

server/raft.go
derekcollison: If we fail processing the entry trying to put the entry on the response channel maybe, will look.

server/raft.go
kozlovic: Still a lot of warnings about the number of append entries pending, and on the mirror leader, "AppendEntryResponse failed to be placed on internal channel" warnings are plenty.
But at least this time the mirroring completes, but there is 1 message missing in the run I just made.
I just tried with file and same outcome (1 missing).  What I see also is that while "nats s report" shows that the lag falls to 0, it takes way more time for the count to go to the correct number, so that I believe shows the number of entries being already replicated but not yet applied?

server/raft.go
```diff
@@ -2362,6 +2386,12 @@ func (n *raft) processAppendEntry(ae *appendEntry, sub *subscription) {
 				n.Unlock()
 				return
 			}
+			// Save in memory for faster processing during applyCommit.
+			n.pae[n.pindex] = ae
+			if len(n.pae) > paeWarnThreshold {
+				n.warn("%d append entries pending", len(n.pae))
+			}
```
kozlovic, line 2393:
```suggestion
			if npae := len(n.pae); npae > paeWarnThreshold && npae%10000 == 0 {
				n.warn("%d append entries pending", npae)
			}
```

server/raft.go
```diff
@@ -2520,6 +2552,12 @@ func (n *raft) sendAppendEntry(entries []*Entry) {
 		// We count ourselves.
 		n.acks[n.pindex] = map[string]struct{}{n.id: struct{}{}}
 		n.active = time.Now()
+
+		// Save in memory for faster processing during applyCommit.
+		n.pae[n.pindex] = ae
+		if len(n.pae) > paeWarnThreshold {
+			n.warn("%d append entries pending", len(n.pae))
+		}
```
kozlovic, line 2560:
```suggestion
		if npae := len(n.pae); npae > paeWarnThreshold && npae%10000 == 0 {
			n.warn("%d append entries pending", npae)
		}
```

kozlovic (commented):
To reduce the number of number of "AppendEntryResponse failed to be placed on internal channel" warnings, we could also add this in handleAppendEntryResponse:
```
                if time.Since(n.laerw) > time.Second {
			n.warn("AppendEntryResponse failed to be placed on internal channel")
			n.laerw = time.Now()
		}
```
with `laerw time.Time` added to raft struct definition.

This is of course, if we think that these warnings are inevitable and not hiding a more serious issue...

kozlovic (approved):
LGTM - last CL done online with @derekcollison

PR #1984, opened 2021-03-09
Fix unit test that received an extra connect message

matthiashanel:

travis ran into this:
```
=== RUN   TestSystemAccountWithGateways
    events_test.go:1625: Expected subject $SYS.ACCOUNT.$G.SERVER.CONNS but got $SYS.ACCOUNT.AASFGYFS64ODVDAUGE6DRHTGWQK4REMJF2FC2RSZCBSI6IULEPPWF6N6.SERVER.CONNS
--- FAIL: TestSystemAccountWithGateways (0.06s)
```

Without the flush, we'd receive the connect message of the connect in line 1588.
This messed up the test logic. 
I cleaned that up as well and added more tests. (check if the Connect message is from the correct cluster)


kozlovic (approved):
LGTM

PR #1989, opened 2021-03-10
Make source and mirror consumers direct.

derekcollison:

/cc @nats-io/core

### DIFF
--- a/server/consumer.go
+++ b/server/consumer.go
@@ -236,7 +236,7 @@ const (
 	JsDeleteWaitTimeDefault = 5 * time.Second
 	// JsFlowControlMaxPending specifies default pending bytes during flow control that can be
 	// outstanding.
-	JsFlowControlMaxPending = 32 * 1024 * 1024
+	JsFlowControlMaxPending = 16 * 1024 * 1024
 )
 
 func (mset *stream) addConsumer(config *ConsumerConfig) (*consumer, error) {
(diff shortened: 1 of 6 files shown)


server/consumer.go
```diff
@@ -63,6 +63,9 @@ type ConsumerConfig struct {
 	MaxAckPending   int           `json:"max_ack_pending,omitempty"`
 	Heartbeat       time.Duration `json:"idle_heartbeat,omitempty"`
 	FlowControl     bool          `json:"flow_control,omitempty"`
+
+	// Don't add to general clients.
+	Direct bool `json:"direct,omitempty"`
```
ripienaar, line 68: hmm, is this something we'd want to expose to clients?

derekcollison: I don't think so..

ripienaar: do we need to export it then?

server/stream.go
```diff
@@ -992,7 +992,7 @@ func (mset *stream) mirrorInfo() *StreamSourceInfo {
 	return mset.sourceInfo(mset.mirror)
 }
 
-const sourceHealthCheckInterval = 10 * time.Second
+const sourceHealthCheckInterval = 2 * time.Second
```
kozlovic, line 995: I had changed that in my tests to not have to wait too long to reproduce the issue I was seeing. I would not have changed it otherwise.

server/stream.go
```diff
@@ -1434,12 +1434,7 @@ func (mset *stream) setSourceConsumer(sname string, seq uint64) {
 			}
 			mset.mu.Unlock()
 		case <-time.After(10 * time.Second):
-			// Make sure things have not changed.
-			mset.mu.Lock()
-			if si := mset.sources[sname]; si != nil && si.cname == _EMPTY_ {
-				mset.setSourceConsumer(sname, seq)
-			}
-			mset.mu.Unlock()
+			return
```
kozlovic, line 1437: Still think that we should remove the retry that we do when getting a bad response (line 1429), and same for mirrors. Again, the long lived go routines that are checking for stalled consumers *should* be able to handle those cases.

kozlovic (approved):
LGTM

kozlovic (approved):
LGTM

PR #1992, opened 2021-03-11
Fix ephemeral removal across gateways

kozlovic:

### DIFF
--- a/server/consumer.go
+++ b/server/consumer.go
@@ -704,7 +704,7 @@ func (o *consumer) setLeader(isLeader bool) {
 		// If push mode, register for notifications on interest.
 		if o.isPushMode() {
 			o.inch = make(chan bool, 8)
-			o.acc.sl.RegisterNotification(o.cfg.DeliverSubject, o.inch)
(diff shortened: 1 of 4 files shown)


server/consumer.go
```diff
@@ -745,6 +745,47 @@ func (o *consumer) setLeader(isLeader bool) {
...
+	}
+	gw := s.gateway
+	if !gw.enabled {
+		return
+	}
+	gw.RLock()
+	defer gw.RUnlock()
+	for _, c := range gw.outo {
+		c.mu.Lock()
+		if ei, ok := c.gw.outsim.Load(acc.Name); ok && ei != nil {
```
derekcollison, line 773: What about changes after this call?

derekcollison (approved):
LGTM - I had one minor question but this looks good.

server/consumer.go
kozlovic: Anyway, I realize that registering on both acc.sl and then the gw.sl is bad because for local interest then the register on gateway would produce a "no interest" in the o.inch channel, which would make delivery stop. My test did not catch that because there was a single message to sync.
Closing this PR for now, need to think of a different way.

PR #1993, opened 2021-03-11
Lost quorum changes for streams

derekcollison:
This PR changes our behaviors for streams and peer removals in several ways.
    
First we no longer try to auto-remap stream assignments on peer removals from the system.
We also now can always respond to stream info requests if at least a member is running.

So when a stream loses quorum, either for simple server down or automated peer removals like out of disk storage, the streams are still visible and operators can remove the peer assignments for the stream manually.

/cc @nats-io/core

### DIFF
--- a/server/const.go
+++ b/server/const.go
@@ -41,7 +41,7 @@ var (
 
 const (
 	// VERSION is the current version for the server.
-	VERSION = "2.2.0-RC.7.5"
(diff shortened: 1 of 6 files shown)


ripienaar (approved):
LGTM

kozlovic (approved):
LGTM

PR #1994, opened 2021-03-12
Fixed a quorum formation issue that caused truncation

matthiashanel:
When a new leader is elected it has to give everyone a chance to reply,
so that we can observe rejections with higher term.

The maximum election timeout is 7.5 seconds.
The new behavior of waiting for the election timeout caused unit tests
to fail. Hence upping the timeout there as well.

Unit test wise this might cause more failures due to the altered timing. (only ran locally so far)
I am also missing the knowledge to implement the TODO that I added. 
But if we could, this would speed things back up for cases where the leader does not change.


derekcollison (approved):
LGTM

PR #1995, opened 2021-03-12
raft: Fixes to cluster size check for streams

wallyqs:
Under some failure conditions a stream replica set could end up with `cluster size < 2` and failing to restart with a `raft: cluster too small` error unless removing a peers.idx file out of band.  We relax this check to allow starting the cluster if we are aware of other known peers which would help recover the state as well.
Also, sometimes the clustersize could end up in 0 or -1 values when restarting JS in nodes were it had become disabled:

```
[70376] 2021/03/11 12:06:03.265984 [DBG] RAFT [NUmM6cRx - S-R3F-hT2jragG] Expanding our clustersize: -1 -> 0
```

prevent that from happening as well by capping the number to be positive (consider at least one node in the cluster always).

/cc @nats-io/core

### DIFF
--- a/server/raft.go
+++ b/server/raft.go
@@ -317,7 +317,7 @@ func (s *Server) startRaftNode(cfg *RaftConfig) (RaftNode, error) {
 	if err != nil {
 		return nil, err
 	}
-	if ps == nil || ps.clusterSize < 2 {
+	if ps == nil || (len(ps.knownPeers) < 2 && ps.clusterSize < 2) {
 		return nil, errors.New("raft: cluster too small")
 	}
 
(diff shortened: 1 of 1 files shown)


server/raft.go
```diff
@@ -1933,9 +1933,13 @@ func (n *raft) applyCommit(index uint64) error {
 			// FIXME(dlc) - Check if this is us??
 			if _, ok := n.peers[oldPeer]; ok {
 				// We should decrease our cluster size since we are tracking this peer.
-				n.debug("Decreasing our clustersize: %d -> %d", n.csz, n.csz-1)
-				n.csz--
-				n.qn = n.csz/2 + 1
+				if n.csz > 1 {
+					n.debug("Decreasing our clustersize: %d -> %d", n.csz, n.csz-1)
```
derekcollison, line 1937: This means we could still go to 1, is that what we wanted here just to prevent going to 0, or did we want to also prevent going to 1?

derekcollison (approved):
LGTM

PR #1996, opened 2021-03-12
JSC Updates

derekcollison:
Tweaked raft leader election under transfer or campaign.
Changed raft startup to not require certain cluster size.
Tweaked default block size for streams.

/cc @nats-io/core

### DIFF
--- a/server/raft.go
+++ b/server/raft.go
@@ -158,6 +158,9 @@ type raft struct {
 	sq    *sendq
 	aesub *subscription
 
+	// Are we doing a leadership transfer.
+	lxfer bool
+
 	// For holding term and vote and peerstate to be written.
 	wtv   []byte
 	wps   []byte
@@ -204,7 +207,7 @@ type lps struct {
 }
 
 const (
-	minElectionTimeout = 1500 * time.Millisecond
+	minElectionTimeout = 1000 * time.Millisecond
 	maxElectionTimeout = 5 * minElectionTimeout
 	minCampaignTimeout = 100 * time.Millisecond
 	maxCampaignTimeout = 4 * minCampaignTimeout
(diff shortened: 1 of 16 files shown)


matthiashanel (changes requested):
I am a hard no on lxfer at this time.
Given that we are so close to GA, I'd rather not introduce this quite yet as it appears too risky.

If at all, we should avoid the full wait if there was no leadership change. (the TODO)
I know that this will improve things, but wouldn't necessary want to force that in just yet.

The reason I'm weary about skipping the full wait until after is the following:
When the replacement is picked there is no guarantee it will still be current by the time the operation is received. 
**And being the server that is current is what my change is all about.**
This could potentially introduce short truncation which is hard to test for.

The reason why a full wait is not needed when there was no leadership change is that the only server where assumptions about being current are guaranteed to hold true.
This will not be true for any other server.

---- edit:

The more I contemplate the change the less I like it. 
Also how often is this going to improve things?
Yes we look at this a lot right now but we specifically tested truncation.
Again the thing to do here is to make sure we don't wait if there was no leader change, this will apply every time a follower experiences issues.

server/raft.go
```diff
@@ -2084,7 +2088,7 @@ func (n *raft) runAsCandidate() {
 				if n.wonElection(votes) {
 					// TODO If this server was also leader in n.term-1, then we could skip the timer as well.
 					// This would be ok as we'd be guaranteed to have the latest history.
-					if len(n.peers) == votes {
+					if len(n.peers) == votes || n.lxfer {
```
kozlovic, line 2091: I mentioned to Matthias that I feel that this addition "negates" his previous fix, so I am not sure about this change..

matthiashanel: I am a hard no on this.

derekcollison: This is when a current leader with quorum selects us to be the new leader, in that case there is no reason to wait. IMO the system is painful to work with when you just ctrl-c a server without this. If we have lost quorum and do not have a leader Matthias code will apply..

kozlovic (approved):
LGTM

PR #1998, opened 2021-03-13
Speed up raft leader election

matthiashanel:
Skip wait if node was previous leader
Shorten the response time to minElectionTimeout

I noticed that I used the wrong timer. election timer purpose is that nodes avoid election at the same time and fight unnecessarily. Thus bad as a timeout. 

All these changes remain valid with Dereks pr.

### DIFF
--- a/server/raft.go
+++ b/server/raft.go
@@ -147,6 +147,9 @@ type raft struct {
 	c        *client
 	dflag    bool
 
+	// last term this node was leader
+	llterm uint64
+
 	// Subjects for votes, updates, replays.
 	psubj  string
 	rpsubj string
@@ -2048,6 +2051,7 @@ func (n *raft) runAsCandidate() {
 	for len(n.votes) > 0 {
 		<-n.votes
 	}
+	lastTermAsLeader := n.llterm
 	n.Unlock()
 
 	// Send out our request for votes.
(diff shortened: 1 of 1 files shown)


server/raft.go
```diff
@@ -2069,23 +2079,28 @@ func (n *raft) runAsCandidate() {
 		case <-n.quit:
 			return
 		case <-elect.C:
+			n.switchToCandidate()
+			return
+		case <-tikChan:
+			// disable timeout and receipt of more votes
+			voteChan = nil
+			tikChan = nil
```
kozlovic, line 2087: tikChan is time.After() and will fire only once, not sure setting it to nil here does anything, but doesn't hurt.

kozlovic (approved):
LGTM

derekcollison (commented):
The better timer makes sense, but the last term I do not think is possible. That means a leader stepped down for some reason and then won the election again, which does not make any sense. In any other scenario being a leader in the last term is not possible.

server/raft.go
matthiashanel: remove

matthiashanel:
> The better timer makes sense, but the last term I do not think is possible. That means a leader stepped down for some reason and then won the election again, which does not make any sense. In any other scenario being a leader in the last term is not possible.

@derekcollison , removed the code

kozlovic (approved):
LGTM

derekcollison:
I changed the logic around this in the updates branch. Let's hold on this one for now.

PR #1999, opened 2021-03-13
On RaftNode shutdown, wait for raft specific go routines

matthiashanel:

some of the routines checked for having RaftNode. 
This check was moved to where the go routine is started.

QuitC now returns the waitgroup used for the RaftNode.
Figured tat wherever the channel is used, the waitgroup needs to be decremented.


kozlovic (commented):
Not feeling confident with those changes so late in the process. I tried this branch and that does not help with a node removed/added back that does not commit.

bruth:
Many changes to the Raft layer have changed. Closing as stale.

PR #2000, opened 2021-03-13
Ephemeral cleanup across GWs

kozlovic:
Watch for interest loss across GWs so ephemeral consumers are removed
when there is no longer local and GW interest.

### DIFF
--- a/server/consumer.go
+++ b/server/consumer.go
@@ -713,6 +713,7 @@ func (o *consumer) setLeader(isLeader bool) {
 					// There is no local interest, but there is GW interest, we
 					// will watch for interest disappearing.
 					// TODO: may need to revisit...
+					stopAndClearTimer(&o.gwdtmr)
 					o.gwdtmr = time.AfterFunc(o.dthresh, func() { o.watchGWinterest() })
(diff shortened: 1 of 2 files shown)


server/consumer.go
```diff
@@ -708,6 +709,12 @@ func (o *consumer) setLeader(isLeader bool) {
 			if o.active = <-o.inch; !o.active {
 				// Check gateways in case they are enabled.
 				o.active = s.hasGatewayInterest(o.acc.Name, o.cfg.DeliverSubject)
+				if o.active {
+					// There is no local interest, but there is GW interest, we
+					// will watch for interest disappearing.
+					// TODO: may need to revisit...
+					o.gwdtmr = time.AfterFunc(o.dthresh, func() { o.watchGWinterest() })
```
derekcollison, line 716: Could this timer already be running?
Should we call stopAndClearTimer(&o.gwdtmr) first?

derekcollison (approved):
In general LGTM, one minor comment. Agree we should revisit.

derekcollison (approved):
LGTM

PR #2002, opened 2021-03-14
[CHANGED] Move Gateway interest-only mode switch from INF to DBG

kozlovic:
Also fixed a test that would sometimes fail depending on timing.


wallyqs (approved):
LGTM

derekcollison (approved):
LGTM

PR #2005, opened 2021-03-14
Delete Dockerfile.alpine

wallyqs:
Since we now have alpine based official images released, plus a [nightly image](https://github.com/nats-io/nats-server/blob/master/docker/Dockerfile.nightly) based from alpine, we can delete this one now.

/cc @nats-io/core


kozlovic (approved):
LGTM

PR #2008, opened 2021-03-15
Don't set to nil on stop to avoid any panics accessing that map.

derekcollison:

/cc @nats-io/core


ripienaar (approved):
LGTM

Issue #2011, opened 2021-03-16
unaligned 64-bit atomic operation on arm7 architecture


GuangchaoDeng:
panic : unaligned 64-bit atomic operation on arm7 architecture

**Describe**:

i deploy cluster on ARMV7l use nats-server-v2.2.0-linux-arm7,but  the nast-server crash when i create stream by natscli

Crash information is as follows：

the panic info: unaligned 64-bit atomic operation

![crash](https://user-images.githubusercontent.com/26615322/111271353-0b5a7680-866c-11eb-986d-5cf9c677b5ab.png)

**Environment**：

nats-server package : nats-server-v2.2.0-linux-arm7.tar.gz

my environment cpu info:

![cpu-info](https://user-images.githubusercontent.com/26615322/111271388-144b4800-866c-11eb-8ca6-2f2a6bc16a59.png)

derekcollison:
Thanks, booting up my RPI-Zero now.. Will get it fixed. Apologies.

derekcollison:
ok should be fixed and is merged into main branch. Will be part of next 2.2.1 release and of course you can build yourself now and should work fine.

mlorenz-tug:
This seems to have reappeared in the latest version on the `armv7l`-architecture using the binary from [nats-server-v2.7.3-linux-arm7.zip](https://github.com/nats-io/nats-server/releases/download/v2.7.3/nats-server-v2.7.3-linux-arm7.zip). 

The `/connz` monitoring endpoint works as expected, but calling the `/varz` endpoint results in the `unaligned 64-bit atomic operation` panic. After the panic the NATS server also has to be killed with `SIGKILL` as the server no longer reacts to `SIGINT`s.

![image](https://user-images.githubusercontent.com/74900368/157398755-a855d315-2d00-4c4d-b252-bca0490f8c60.png)

![image](https://user-images.githubusercontent.com/74900368/157398723-4412a526-41f0-4462-9f7d-fd444bbecf2e.png)

PR #2012, opened 2021-03-16
Fix for #2011 and atomics on 32bit systems.

derekcollison:

/cc @nats-io/core

### DIFF
--- a/server/filestore.go
+++ b/server/filestore.go
@@ -86,20 +86,20 @@ type fileStore struct {
 
 // Represents a message store block and its data.
 type msgBlock struct {
-	mu    sync.RWMutex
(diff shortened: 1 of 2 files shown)


philpennock (approved):
Do we have stats (from CI on this PR?) on any performance impact from moving the mutex away from the zero-offset and how that changes addressing modes used in the generated code?

I think the mutex is always 24 octets wide, so IF there is a performance regression, that could be moved back to the front.  But I don't know that there is any performance impact.

derekcollison:
We have the mutex moved already for the client struct which is our most sensitive to perf so I think we are good. But I do not have direct numbers of impact.

PR #2013, opened 2021-03-16
When a stream is a mirror or has sources we need to check the upstream streams.

derekcollison:

/cc @nats-io/core

### DIFF
--- a/server/consumer.go
+++ b/server/consumer.go
@@ -246,7 +246,7 @@ func (mset *stream) addConsumer(config *ConsumerConfig) (*consumer, error) {
 
 func (mset *stream) addConsumerWithAssignment(config *ConsumerConfig, oname string, ca *consumerAssignment) (*consumer, error) {
 	mset.mu.RLock()
-	s, jsa := mset.srv, mset.jsa
+	s, jsa, acc := mset.srv, mset.jsa, mset.acc
 	mset.mu.RUnlock()
 
 	// If we do not have the consumer currently assigned to us in cluster mode we will proceed but warn.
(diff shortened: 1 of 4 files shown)


wallyqs (approved):
LGTM

server/consumer.go
```diff
@@ -330,7 +330,25 @@ func (mset *stream) addConsumerWithAssignment(config *ConsumerConfig, oname stri
 
 	// Make sure any partition subject is also a literal.
 	if config.FilterSubject != _EMPTY_ {
-		if !mset.validSubject(config.FilterSubject) {
+		streams, cfg, valid := []*stream{mset}, mset.config(), false
```
ripienaar, line 333: we cant lookup or verify this for streams with external set right? So I think we should handle those here

derekcollison: I also need to do clustered version which I will work on now.

ripienaar (approved):
LGTM

server/consumer.go
```diff
@@ -330,7 +330,24 @@ func (mset *stream) addConsumerWithAssignment(config *ConsumerConfig, oname stri
...
+			} else {
+				subjects = acc.streamSourceSubjects(cfg.Mirror)
+			}
+		} else if len(cfg.Sources) > 0 {
+			for _, si := range cfg.Sources {
+				if si.External != nil {
+					hasExt = true
+					break
+				}
+				subjects = append(subjects, acc.streamSourceSubjects(si)...)
```
wallyqs, line 347: In case we have stream that it itself is sourced from another stream like:

```
Stream M1 <-OriginA
Stream M2 <- OriginB
Stream S1 <- M1, M2
```

Then it will fail to match since we need to check against the subjects from either OriginA or OriginB.

derekcollison: Good point, need to make it recursive it seems to some degree.

PR #2015, opened 2021-03-18
Updates to memory retention in filestore.

derekcollison:
Updates to memory pressure.
    
    1. Release cache during heavy writes when we move to a new block if no read activity.
    2. If we detect a linear scan reading dump the cache on the last seq read for the mb.
    
/cc @nats-io/core


wallyqs (approved):
LGTM

ripienaar (approved):
LGTM

PR #2017, opened 2021-03-18
Mirror consumers were not able to restart after initial failure.

derekcollison:

/cc @nats-io/core


ripienaar (approved):
LGTM

PR #2018, opened 2021-03-18
Fix for filtered consumers and idle heartbeats.

derekcollison:
If a filtered consumer was backed by a busy stream then idle heartbeats for the idle consumer would be suppressed.

/cc @nats-io/core


ripienaar (approved):
LGTM

PR #2020, opened 2021-03-18
ensure connect events are always sent

### LINKED ISSUE #2019: Account connection events not sent for custom authentication
## Defect

Account connect events is not sent when a custom authenticator is set:

https://github.com/nats-io/nats-server/blob/4784205c2d60ee57f1c539e3eb5f19d03974a71d/server/auth.go#L323-L334

Above the events are only sent in `processClientOrLeafAuthentication`

ripienaar:
Fixes #2019 

/cc @nats-io/core

### DIFF
--- a/server/auth.go
+++ b/server/auth.go
@@ -326,11 +326,20 @@ func (s *Server) isClientAuthorized(c *client) bool {
 	// Check custom auth first, then jwts, then nkeys, then
 	// multiple users with TLS map if enabled, then token,
 	// then single user/pass.
-	if opts.CustomClientAuthentication != nil {
-		return opts.CustomClientAuthentication.Check(c)
+	if opts.CustomClientAuthentication != nil && !opts.CustomClientAuthentication.Check(c) {
+		return false
 	}
 
-	return s.processClientOrLeafAuthentication(c, opts)
+	if opts.CustomClientAuthentication == nil && !s.processClientOrLeafAuthentication(c, opts) {
+		return false
+	}
+
+	if c.kind == CLIENT || c.kind == LEAF {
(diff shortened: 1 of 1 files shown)


derekcollison (changes requested):

server/auth.go
```diff
@@ -327,10 +327,21 @@ func (s *Server) isClientAuthorized(c *client) bool {
 	// multiple users with TLS map if enabled, then token,
 	// then single user/pass.
 	if opts.CustomClientAuthentication != nil {
-		return opts.CustomClientAuthentication.Check(c)
+		if !opts.CustomClientAuthentication.Check(c) {
```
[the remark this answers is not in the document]
ripienaar: done

.travis.yml
```diff
@@ -4,7 +4,6 @@ dist: focal
 language: go
 go:
 - 1.16.x
-- 1.15.x
```
derekcollison, line 7: I think we do different tests based on 1.15 for now so might want to revert this one.

ripienaar: Looks like just 

https://github.com/nats-io/nats-server/blob/4784205c2d60ee57f1c539e3eb5f19d03974a71d/.travis.yml#L32

I can update that too? Already go.mod declares 1.16 required so worth sorting it out

derekcollison: We run different tests on each platform so we always have 2... Let's leave for now IMO. Agree be nice to remove but will want Ivan to weigh in.

ripienaar: ok will revert

derekcollison (approved):
LGTM

PR #2021, opened 2021-03-18
Direct consumer updates.

derekcollison:
1. Reduce maximum sliding window for large N->1.
2. Do not create any store for directs.
3. Relax hard requirement for interest at creation for directs.

/cc @nats-io/core


ripienaar (approved):
LGTM

PR #2023, opened 2021-03-19
Make sure the jetstream accounts use the name as the key to the map.

derekcollison:
This prevents possible double adds under reload or restart scenarios.

/cc @nats-io/core


wallyqs (approved):
LGTM

PR #2024, opened 2021-03-20
In operator mode with JetStream we want to load all JS accounts.

derekcollison:
Previously since accounts were lazy loaded they would not be loaded into the system on startup.

/cc @nats-io/core


server/jetstream.go
```diff
@@ -427,25 +427,44 @@ func (s *Server) configAllJetStreamAccounts() error {
...
 	}
+
+	// Now walk all the storage we have and resolve any accounts that we did not process already.
+	// This is important in resolver/operator models.
+	fis, _ := ioutil.ReadDir(js.config.StoreDir)
+	for _, fi := range fis {
+		if accName := fi.Name(); accName != _EMPTY_ {
+			// Only load up ones not already loaded since they are processed above.
+			if _, ok := accounts.Load(accName); !ok {
+				if acc, err := s.lookupAccount(accName); err != nil {
```
ripienaar, line 459: so accounts we cant lookup here will later lazy load and enable? say operator gets it from some other server into its cache etc?

derekcollison: Yes we will load on-demand all accounts that are represented by stable storage at startup.

ripienaar:
LGTM

PR #2025, opened 2021-03-20
Can not use sub comparisons for old messages from direct consumers.

derekcollison:

/cc @nats-io/core


ripienaar (approved):
LGTM

PR #2026, opened 2021-03-20
Reworked sources and mirrors on missed data.

derekcollison:
Add last delivered sequence to consumer idle heartbeats.

/cc @nats-io/core


ripienaar (approved):
LGTM

PR #2027, opened 2021-03-21
Reduce sliding window for direct consumers and catchup stream windows.

derekcollison:
Remove another possible wire blocking operation in raft.

/cc @nats-io/core


ripienaar:
LGTM

PR #2028, opened 2021-03-22
Fix for bad behavior with flow control and multiple sources streams.

derekcollison:
Also allow chaining of sources and mirrors with filtered consumers.

/cc @nats-io/core

### DIFF
--- a/server/stream.go
+++ b/server/stream.go
@@ -980,24 +980,127 @@ func (mset *stream) sourcesInfo() (sis []*StreamSourceInfo) {
 	return sis
 }
 
-// Return the subjects for a stream source.
-func (a *Account) streamSourceSubjects(ss *StreamSource) (subjects []string) {
-	s, js, _ := a.getJetStreamFromAccount()
-	if !s.JetStreamIsClustered() {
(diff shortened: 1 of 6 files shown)


server/stream.go
```diff
@@ -980,24 +980,128 @@ func (mset *stream) sourcesInfo() (sis []*StreamSourceInfo) {
...
-		if sa := js.streamAssignment(a.Name, ss.Name); sa != nil {
-			js.mu.RLock()
-			if len(sa.Config.Subjects) > 0 {
-				subjects = append(sa.Config.Subjects[:0:0], sa.Config.Subjects...)
+		return js.streamSourceSubjectsClustered(a.Name, ss.Name, seen)
+	}
+}
+
+func (js *jetStream) streamSourceSubjectsClustered(accountName, streamName string, seen map[string]bool) (subjects []string, hasExt bool) {
+	fmt.Printf("AAA\n\n")
```
ripienaar, line 1026: some debug stuff here

ripienaar (approved):
LGTM besides the debug print

server/stream.go
derekcollison: Fixed

PR #2029, opened 2021-03-22
tls: support multiple key/cert pairs, for SNI

philpennock:
This does not add tests or docs, this is to sound out how people feel about the
API.  If we like it, we can dress it up formally.

Turn the TLS options cert_file and key_file into list-separated paths (so
colon-delimited if on Unix, like `$PATH`); the count of entries must match, and
the keys and certs should zip together.

With this change, TLS SNI works to pick the correct cert to return for a given
connection, allowing a NATS server to have multiple identities.

/cc @nats-io/core

### DIFF
--- a/server/opts.go
+++ b/server/opts.go
@@ -4624,16 +4624,28 @@ func GenTLSConfig(tc *TLSConfigOpts) (*tls.Config, error) {
 	case tc.CertFile == _EMPTY_ && tc.KeyFile != _EMPTY_:
 		return nil, fmt.Errorf("missing 'cert_file' in TLS configuration")
 	case tc.CertFile != _EMPTY_ && tc.KeyFile != _EMPTY_:
-		// Now load in cert and private key
-		cert, err := tls.LoadX509KeyPair(tc.CertFile, tc.KeyFile)
-		if err != nil {
-			return nil, fmt.Errorf("error parsing X509 certificate/key pair: %v", err)
+		// To support multiple certs (for SANs) without rearchitecting our
+		// configuration, we accept a list, separated by the platform list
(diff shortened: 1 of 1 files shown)


matthiashanel (changes requested):
add a unit test so this functionality does not disappear?

wallyqs:
Thanks @philpennock ok sounds like we should add a field to support multiple certs optionally then:

```hcl
tls {
  certs = [{cert_file = "first.pem", key_file = "first.key"}]
}
```

ripienaar:
Agree with @wallyqs that's a lot clearer

philpennock:
It's still my intent to find time to work on the new config layout and write the tests (but I'm happy for someone else to take it if they have time).  The push just now is not me doubling-down on the quick hack proof-of-concept, it's just that I have this live at home and I wanted to install 2.2.3 so I rebased and pushed after confirming it works.

derekcollison:
What is the status here?

philpennock:
This is still useful functionality to have, but needs to be done more cleanly, as wally requested.  I'm still running with this at home (to dual-name with a .lan hostname from my personal cert authority, and with a Let's Encrypt cert for the current external IP) and we today used this patch for a fiddly migration, to be able to cut DNS over to a box at our leisure.

So the use-case is definitely there and it needs a better integration than the proof-of-concept with the path-separator strings, and tests to protect against regression.

bruth:
Is this worth revisiting and prioritizing?

derekcollison:
You should sync with @philpennock

philpennock:
Closing in favor of Wally's #4889 which does the configuration the Right Way for nats-server, instead of the old-fashioned Unix way.  :laughing:

PR #2033, opened 2021-03-26
RC8

derekcollison:
Changes and stability improvements.

/cc @nats-io/core

### DIFF
--- a/server/jetstream.go
+++ b/server/jetstream.go
@@ -1272,9 +1272,9 @@ func (jsa *jsAccount) checkBytesLimits(addBytes int64, storage StorageType) erro
 }
 
 func (jsa *jsAccount) acc() *Account {
-	jsa.mu.RLock()
+	//jsa.mu.RLock()
 	acc := jsa.account
-	jsa.mu.RUnlock()
+	//jsa.mu.RUnlock()
 	return acc
 }
 
--- a/server/jetstream.go
+++ b/server/jetstream.go
@@ -1302,10 +1302,7 @@ func (jsa *jsAccount) checkBytesLimits(addBytes int64, storage StorageType) erro
 }
 
 func (jsa *jsAccount) acc() *Account {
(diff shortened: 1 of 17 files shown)


server/jetstream.go
```diff
@@ -1272,9 +1272,9 @@ func (jsa *jsAccount) checkBytesLimits(addBytes int64, storage StorageType) erro
 }
 
 func (jsa *jsAccount) acc() *Account {
-	jsa.mu.RLock()
+	//jsa.mu.RLock()
```
kozlovic, line 1301: What is that?

server/stream.go
```diff
@@ -1222,13 +1223,28 @@ func (mset *stream) processInboundMirrorMsg(m *inMsg) bool {
...
+	olag, odseq, oclseq := mset.mirror.lag, mset.mirror.dseq, mset.mirror.clseq
+	if dseq == mset.mirror.dseq+1 {
+		mset.mirror.dseq++
+		mset.mirror.sseq = sseq
+	} else if dseq > mset.mirror.dseq {
+		if mset.mirror.cname == _EMPTY_ {
+			mset.mirror.cname = tokenAt(m.rply, 4)
+			mset.mirror.dseq, mset.mirror.sseq = dseq, sseq
+		} else {
+			mset.retryMirrorConsumer()
```
kozlovic, line 1238: Does that not require lock? If so, should release before?

kozlovic (changes requested):

server/jetstream.go
derekcollison: Left over, removed lock since it is only set at creation and can't change. Will fix.

server/stream.go
derekcollison: Yes thanks, will fix.

kozlovic (approved):
LGTM

kozlovic (approved):
LGTM

PR #2034, opened 2021-03-26
Calling out to the storage update should not hold lock.

derekcollison:
We had lock inversion with consumers. Not best solution but good for now.

/cc @nats-io/core


ripienaar (approved):
LGTM

ripienaar:
LGTM

PR #2035, opened 2021-03-26
fix nightly builds

ripienaar:
goreleaser now only function with GO111MODULES on, as will
much of the go eco system soon and we will need to move also.

For now use a version that still supports it

/cc @nats-io/core


derekcollison (approved):
LGTM

PR #2039, opened 2021-03-26
Durable consumers with R>1 had performance challenges.

derekcollison:
This code changes the way we handle raft based proposals for consumers.

/cc @nats-io/core


kozlovic (approved):
LGTM

PR #2040, opened 2021-03-26
[FIXED] Websocket: do not generate empty frames + LN corruption

kozlovic:
- It was possible that when the server was sending frames to a
webbrowser, it would send empty frames. While technically not wrong,
prevent that from happening.
- Not copying enqueued buffers could cause corruption with LN+WS.

### DIFF
--- a/server/websocket.go
+++ b/server/websocket.go
@@ -1235,12 +1235,17 @@ func (c *client) wsCollapsePtoNB() (net.Buffers, int64) {
 					continue
 				}
 				for len(b) > 0 {
-					endFrame(fhIdx, total)
+					endStart := total != 0
+					if endStart {
+						endFrame(fhIdx, total)
+					}
 					total = len(b)
 					if total >= mfs {
 						total = mfs
 					}
-					fhIdx = startFrame()
+					if endStart {
+						fhIdx = startFrame()
+					}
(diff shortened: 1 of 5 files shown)


server/websocket.go
```diff
@@ -1230,18 +1230,23 @@ func (c *client) wsCollapsePtoNB() (net.Buffers, int64) {
 			for i := 0; i < len(nb); i++ {
 				b := nb[i]
 				if total+len(b) <= mfs {
-					bufs = append(bufs, b)
+					bufs = append(bufs, copyIfMasking(mask, b))
```
derekcollison, line 1233: The copy here still needed?

derekcollison (approved):
LGTM - One minor question..

server/websocket.go
kozlovic: I guess we may not need since this is a buffer for a particular client and it is unique to that client (has been copied during queueOutbound), and the buffer is discarded after being sent. Let me revisit that...

kozlovic: I have removed that and updated PR (with force push)...

kozlovic: (also rebased from main branch prior to force push)

derekcollison (approved):
LGTM

PR #2041, opened 2021-03-27
[Added] error when mirror/source stream prefix overlaps with stream subs

matthiashanel:

This change collects prefixes from mirror or sources, then checks them against collected subjects.

This does not work if the streams are in different accounts.
I can look into adding that as well. 
Just haven't yet as I was unsure why it was not done already.
Should I add it?

I am also wondering if this (already  present) check should be done for Sources as well? (since I added code to obtain the stream config there as well)
```
		if exists && cfg.MaxMsgSize > 0 && maxMsgSize > 0 && cfg.MaxMsgSize < maxMsgSize {
			resp.Error = &ApiError{Code: 400, Description: "stream mirror must have max message size >= source"}
			return
		}
```

### DIFF
--- a/server/jetstream_api.go
+++ b/server/jetstream_api.go
@@ -1142,6 +1142,28 @@ func (s *Server) jsStreamCreateRequest(sub *subscription, c *client, subject, re
 		return
 	}
 
+	hasStream := func(streamName string) (bool, int32, []string) {
+		var exists bool
+		var maxMsgSize int32
+		var subs []string
+		if s.JetStreamIsClustered() {
+			if js, _ := s.getJetStreamCluster(); js != nil {
+				js.mu.RLock()
+				if sa := js.streamAssignment(acc.Name, streamName); sa != nil {
+					maxMsgSize = sa.Config.MaxMsgSize
+					subs = sa.Config.Subjects
+					exists = true
+				}
+				js.mu.RUnlock()
+			}
+		} else if mset, err := acc.lookupStream(streamName); err == nil {
+			maxMsgSize = mset.cfg.MaxMsgSize
+			subs = mset.cfg.Subjects
+			exists = true
+		}
+		return exists, maxMsgSize, subs
+	}
+
 	var streamSubs []string
 	var deliveryPrefixes []string
 	var apiPrefixes []string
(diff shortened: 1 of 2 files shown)


server/jetstream_api.go
```diff
@@ -1168,22 +1172,75 @@ func (s *Server) jsStreamCreateRequest(sub *subscription, c *client, subject, re
...
+		}
+	}
+	if len(cfg.Sources) > 0 {
+		for _, src := range cfg.Sources {
+			if src.External == nil {
+				continue
+			}
+			if s.JetStreamIsClustered() {
+				js, _ := s.getJetStreamCluster()
+				if sa := js.streamAssignment(acc.Name, src.Name); sa != nil {
```
kozlovic, line 1208: I don't think you can call js.streamAssignment() like that (it needs locking)

server/jetstream_api.go
```diff
@@ -1168,22 +1172,75 @@ func (s *Server) jsStreamCreateRequest(sub *subscription, c *client, subject, re
 		// We do not require other stream to exist anymore, but if we can see it check payloads.
 		var exists bool
 		var maxMsgSize int32
-
 		if s.JetStreamIsClustered() {
 			js, _ := s.getJetStreamCluster()
 			if sa := js.streamAssignment(acc.Name, cfg.Mirror.Name); sa != nil {
```
kozlovic, line 1177: Missing locking here!

kozlovic (changes requested):
`streamAssignment()` needs to be protected (I understand that before your change this function calls it without locking, but that is a mistake).
Will defer the rest to Derek.

server/jetstream_api.go
matthiashanel: fixed

server/jetstream_api.go
matthiashanel: fixed

server/jetstream_api.go
```diff
@@ -1173,11 +1173,14 @@ func (s *Server) jsStreamCreateRequest(sub *subscription, c *client, subject, re
 		var exists bool
 		var maxMsgSize int32
 		if s.JetStreamIsClustered() {
-			js, _ := s.getJetStreamCluster()
-			if sa := js.streamAssignment(acc.Name, cfg.Mirror.Name); sa != nil {
-				maxMsgSize = sa.Config.MaxMsgSize
-				streamSubs = sa.Config.Subjects
-				exists = true
+			if js, _ := s.getJetStreamCluster(); js != nil {
+				js.mu.RLock()
```
derekcollison, line 1177: This can go inside conditional below. ok to get to the pointer and check for nil without lock.

server/jetstream_api.go
```diff
@@ -1204,9 +1207,12 @@ func (s *Server) jsStreamCreateRequest(sub *subscription, c *client, subject, re
 				continue
 			}
 			if s.JetStreamIsClustered() {
-				js, _ := s.getJetStreamCluster()
-				if sa := js.streamAssignment(acc.Name, src.Name); sa != nil {
-					streamSubs = append(streamSubs, sa.Config.Subjects...)
+				if js, _ := s.getJetStreamCluster(); js != nil {
+					js.mu.RLock()
```
derekcollison, line 1211: Same here.

server/jetstream_api.go
```diff
@@ -1168,22 +1172,75 @@ func (s *Server) jsStreamCreateRequest(sub *subscription, c *client, subject, re
...
+				deliveryPrefixes = append(deliveryPrefixes, src.External.DeliverPrefix)
+			}
+			if src.External.ApiPrefix != _EMPTY_ {
+				apiPrefixes = append(apiPrefixes, src.External.ApiPrefix)
+			}
+		}
+	}
+	// check prefix overlap with subjects
+	for _, pfx := range deliveryPrefixes {
+		if pfx == "*" || pfx == ">" || strings.Contains(pfx, ".*") || strings.HasSuffix(pfx, ".>") {
```
derekcollison, line 1230: We have const for these I believe and helper functions for testing if they have wildcards etc.

derekcollison (approved):
LGTM - Some minor comments but nothing critical.

derekcollison (approved):
LGTM

server/jetstream_api.go
matthiashanel: The lock is needed around streamAssignment

server/jetstream_api.go
matthiashanel: I believe same as above

server/jetstream_api.go
matthiashanel: The lock is needed around streamAssignment
```
// Will lookup a stream assignment.
// Lock should be held.
func (js *jetStream) streamAssignment(account, stream string) (sa *streamAssignment) {
```

server/jetstream_api.go
matthiashanel: same as above

server/jetstream_api.go
matthiashanel: fixed

PR #2045, opened 2021-03-29
RC11

derekcollison:
Move raft panic to a warning and make sure bad stream updates do not delete the stream.

/cc @nats-io/core


kozlovic (approved):
LGTM

PR #2047, opened 2021-03-29
[FIXED] Websocket: print correct random port instead of `0`

kozlovic:
When a random port is selected (likely in tests), the banner would
print the port as `0` instead of its real value.


derekcollison (approved):
LGTM

PR #2048, opened 2021-03-30
Jwt remove

matthiashanel:
2 commits:
1: adding ability to remove account jwt.
2: when doing so disconnect clients and disable account. 
In lieu of account removal inside the server (preferred), I had to add this stop gap measure as this would have been the next ask.

### DIFF
--- a/server/accounts.go
+++ b/server/accounts.go
@@ -3590,6 +3590,28 @@ func claimValidate(claim *jwt.AccountClaims) error {
 	return nil
 }
 
+func removeCb(s *Server, pubKey string) {
+	v, ok := s.accounts.Load(pubKey)
+	if !ok {
+		return
+	}
+	a := v.(*Account)
+	s.Debugf("Disable account %s due to remove", pubKey)
+	a.mu.Lock()
+	// lock out new clients
+	a.msubs = 0
+	a.mpay = 0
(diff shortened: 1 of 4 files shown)


derekcollison (approved):
LGTM

kozlovic (approved):
LGTM - however, if we merge that and we do a 2.2.1 release, I need to understand if this is more of a fix or an addition: I don't see new API or nothing that would require a 2.3.0 release, but I would like to double check. Otherwise we have 2 options: make a 2.3.0 release instead of 2.2.1 or delay the merge until after 2.2.1?

matthiashanel:
@kozlovic previously the delete request could only be signed by the operator primary key. 
Now signing keys will work as well.

It is reasonable for a customer to expect clients to the removed account to be disconnected. 
The initial plan was to add this as part of account cleanup, which was rejected.
So now I am adding a bandaid that sets the account residing in server memory to expired and close all client connections.

kozlovic:
@matthiashanel Thanks, so it looks almost more like a fix than a new feature and regardless, there is no new API or breaking changes that I can see, so inclusion to a 2.2.1 release is ok.

matthiashanel:
@ikozlovic  will merge then

Issue #1976, opened 2021-03-08
400 instead of 500 on error of STREAM.MSG.GET requests


scottf:
## Feature Request
When making a STREAM.MSG.GET request with an invalid sequence respond with a 400 error code instead of a 500 error code. Currently, the response has error code 500, which suggests a server error.

Also, maybe combine the use cases into the same error as it's irrelevant to the client why the message is not found.

#### Use Case:

The first can happen when the requested sequence is 0 or for a message that was deleted. The second is when the requested sequence is greater than the last actual sequence number.

```
{"type":"io.nats.jetstream.api.v1.stream_msg_get_response","error":{"code":500,"description":"no message found"}}
{"type":"io.nats.jetstream.api.v1.stream_msg_get_response","error":{"code":500,"description":"stream store EOF"}}
```

#### Who Benefits From The Change(s)?
More standard error code, where 4xx represents a user/input error instead of 5xx representing a server error.

ripienaar:
Agree, should be 404

```
17:47:37 >>> $JS.API.STREAM.MSG.GET.CHORIA_REGISTRATION
{"seq":10}

17:47:37 <<< $JS.API.STREAM.MSG.GET.CHORIA_REGISTRATION
{"type":"io.nats.jetstream.api.v1.stream_msg_get_response","error":{"code":500,"description":"no message found"}}
```

derekcollison:
This with no wait set or no wait is not set? Batch size?

ripienaar:
It’s the stream get APi not consumer pull. 

That api does a message lookup and return 500 for all errors - msg not found is a error from the storage

derekcollison:
Ah gotcha.. ok that makes more sense ;)

kozlovic:
Closing this since it is addressed with #2053

PR #2053, opened 2021-03-31
Fix for #1976

derekcollison:

/cc @nats-io/core


server/jetstream_api.go
```diff
@@ -2351,7 +2352,7 @@ func (s *Server) jsMsgGetRequest(sub *subscription, c *client, subject, reply st
 
 	subj, hdr, msg, ts, err := mset.store.LoadMsg(req.Seq)
 	if err != nil {
-		resp.Error = jsError(err)
+		resp.Error = jsNoMessageFoundErr
```
ripienaar, line 2355: not sure, i mean LoadMsg can fail for many reasons, perhaps LoadMsg should return a specific error for not found vs other kinds of failure - if it can detect it, then we can have 404 and 500 appropriately?

derekcollison: Most of the others are captured above. This is what we do for a pull consumer also, essentially turn everything into a `not found` error.

ripienaar: Ok. Consistency matters lost then. LGTM

PR #2056, opened 2021-04-01
Aggressive cache expiration was hindering performance

derekcollison:
When publishing and receiving from a stream concurrently performance was hindered due to aggressive cache expiration.

/cc @nats-io/core


kozlovic (approved):
LGTM

PR #2057, opened 2021-04-02
[Adding] (fetch) timeout to nats resolver

matthiashanel:

Just in case as per our conversation
The change in signature to NewCacheDirAccResolver and NewDirAccResolver are backwards compatible.
I left URLAccResolver unchanged for now.
The resolver being public, made this a little more involved

### DIFF
--- a/server/accounts.go
+++ b/server/accounts.go
@@ -3784,10 +3784,10 @@ func (dr *DirAccResolver) Store(name, jwt string) error {
 	return dr.saveIfNewer(name, jwt)
 }
 
-type dirResOption func(s *DirAccResolver) error
+type DirResOption func(s *DirAccResolver) error
 
 // limits the amount of time spent waiting for an account fetch to complete
-func FetchTimeout(to time.Duration) dirResOption {
+func FetchTimeout(to time.Duration) DirResOption {
 	return func(r *DirAccResolver) error {
 		if to <= time.Duration(0) {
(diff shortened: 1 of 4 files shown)


server/accounts.go
```diff
@@ -3772,19 +3771,42 @@ func (dr *DirAccResolver) Fetch(name string) (string, error) {
...
 
 func (dr *DirAccResolver) Store(name, jwt string) error {
 	return dr.saveIfNewer(name, jwt)
 }
 
-func NewDirAccResolver(path string, limit int64, syncInterval time.Duration, delete bool) (*DirAccResolver, error) {
+type dirResOption func(s *DirAccResolver) error
+
+// limits the amount of time spent waiting for an account fetch to complete
+func FetchTimeout(to time.Duration) dirResOption {
```
wallyqs, line 3790: Need to either make `fetchTimeout` private or `disResOption` public, since not recommended to return private types:

```
// golint
accounts.go:3790:37: exported func FetchTimeout returns unexported type server.dirResOption, which can be annoying to use
```


```suggestion
func fetchTimeout(to time.Duration) dirResOption {
```

matthiashanel: Since it's an argument to a public function that should probably stay public.

matthiashanel: fixed

derekcollison (approved):
LGTM

wallyqs (approved):
LGTM

PR #2058, opened 2021-04-02
During repeated server restarts or failures consumer state could drift between replicas.

derekcollison:

/cc @nats-io/core


kozlovic (approved):
LGTM

PR #2061, opened 2021-04-02
[FIXED] MQTT: make sure unsubscribe is propagated

kozlovic:
In error conditions or when replacing an existing sub qos1 to qos0,
we were unsubscribing the NATS subscription, but that would not
have been propagated across the cluster.

Also fixed a flapper


derekcollison (approved):
LGTM

PR #2064, opened 2021-04-05
put typestring to map and add tests

alexpantyukhin:
 - [x] Tests added

Resolves #

 - substitute switch-case with map in the `typeString` function.

/cc @nats-io/core

### DIFF
--- a/server/client.go
+++ b/server/client.go
@@ -4361,23 +4361,21 @@ func (c *client) flushAndClose(minimalFlush bool) {
 	}
 }
 
+var typeStringMap = map[int]string{
+	CLIENT:    "Client",
+	ROUTER:    "Router",
+	GATEWAY:   "Gateway",
+	LEAF:      "Leafnode",
+	JETSTREAM: "JetStream",
+	ACCOUNT:   "Account",
+	SYSTEM:    "System",
+}
+
 func (c *client) typeString() string {
-	switch c.kind {
-	case CLIENT:
-		return "Client"
-	case ROUTER:
-		return "Router"
-	case GATEWAY:
-		return "Gateway"
-	case LEAF:
-		return "Leafnode"
-	case JETSTREAM:
-		return "JetStream"
-	case ACCOUNT:
-		return "Account"
-	case SYSTEM:
-		return "System"
+	if typeStringVal, ok := typeStringMap[c.kind]; ok {
+		return typeStringVal
 	}
+
 	return "Unknown Type"
 }
 
--- a/server/client_test.go
+++ b/server/client_test.go
@@ -852,6 +852,54 @@ func TestSplitSubjectQueue(t *testing.T) {
 	}
 }
 
+func TestTypeString(t *testing.T) {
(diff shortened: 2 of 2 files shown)


derekcollison (approved):
LGTM - Thanks.

server/client.go
```diff
@@ -4361,23 +4361,21 @@ func (c *client) flushAndClose(minimalFlush bool) {
 	}
 }
 
+var typeStringMap = map[int]string{
+	CLIENT: "Client",
```
kozlovic, line 4365: You need to run `go fmt` on this. You can see from the PR run that this file is not properly formatted.

server/client_test.go
```diff
@@ -852,6 +852,35 @@ func TestSplitSubjectQueue(t *testing.T) {
 	}
 }
 
+func TestTypeString(t *testing.T) {
+	_, c, _ := setupClient()
```
kozlovic, line 856: Don't need this one.

server/client_test.go
```diff
@@ -852,6 +852,35 @@ func TestSplitSubjectQueue(t *testing.T) {
...
+		{intType: GATEWAY, stringType: "Gateway"},
+		{intType: LEAF, stringType: "Leafnode"},
+		{intType: JETSTREAM, stringType: "JetStream"},
+		{intType: ACCOUNT, stringType: "Account"},
+		{intType: SYSTEM, stringType: "System"},
+		{intType: -1, stringType: "Unknown Type"},
+	}
+	for _, cs := range cases {
+		_, c, _ := setupClient()
+		c.kind = cs.intType
```
kozlovic, line 873: I ran this test and it would leave client's go routine running. Doing a defer c.Close() after setup would cause some failures because in certain case it would try to reconnect, etc.. We don't need the client to be setup for this test, so simple replace with:
```suggestion
		c := &client{kind: cs.intType}
```

server/client_test.go
```diff
@@ -852,6 +852,35 @@ func TestSplitSubjectQueue(t *testing.T) {
...
+		typeStringVal := c.typeString()
+
+		if typeStringVal != cs.stringType {
+			t.Fatalf("Expected typeString  value  %q, but instead received %q", cs.stringType, typeStringVal)
+		}
+	}
+
+	defer c.close()
```
kozlovic, line 881: Remove this.

kozlovic (changes requested):

server/client_test.go
alexpantyukhin: fixed

server/client_test.go
alexpantyukhin: fixed

server/client_test.go
alexpantyukhin: fixed

alexpantyukhin:
@kozlovic Thank you very much for your feedback! I have fixed your review notes. Please re-review.

kozlovic (approved):
LGTM. Thank you for your contribution!

PR #2065, opened 2021-04-05
extract_server_update

alexpantyukhin:
 - [ ] Tests added

Resolves #

 - Extract server updates functionality into function

/cc @nats-io/core

### DIFF
--- a/server/client.go
+++ b/server/client.go
@@ -2522,11 +2522,7 @@ func (c *client) addShadowSub(sub *subscription, ime *ime) (*subscription, error
 	}
 
 	// Update our route map here.
-	c.srv.updateRouteSubscriptionMap(im.acc, &nsub, 1)
-	if c.srv.gateway.enabled {
-		c.srv.gatewayUpdateSubInterest(im.acc.Name, &nsub, 1)
-	}
-	c.srv.updateLeafNodes(im.acc, &nsub, 1)
+	c.srv.updateRemoteSubscription(im.acc, &nsub, 1)
 
 	return &nsub, nil
 }
@@ -2972,11 +2968,7 @@ func (c *client) deliverMsg(sub *subscription, subject, reply, mh, msg []byte, g
 				// Due to defer, reverse the code order so that execution
 				// is consistent with other cases where we unsubscribe.
 				if shouldForward {
-					if srv.gateway.enabled {
-						defer srv.gatewayUpdateSubInterest(client.acc.Name, sub, -1)
-					}
-					defer srv.updateRouteSubscriptionMap(client.acc, sub, -1)
-					defer srv.updateLeafNodes(client.acc, sub, -1)
+					defer srv.updateRemoteSubscription(client.acc, sub, -1)
 				}
 				defer client.unsubscribe(client.acc, sub, true, true)
 			} else if sub.nm > sub.max {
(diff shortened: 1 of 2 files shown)


server/client.go
```diff
@@ -3125,6 +3117,16 @@ func (c *client) deliverMsg(sub *subscription, subject, reply, mh, msg []byte, g
 	return true
 }
 
+func (c *client) serveUpdate(client *Client, sub *subscription) {
```
kozlovic, line 3120: I would rename this to make it a server receiver since when we call above, you already have `srv`. Also, maybe a rename to `func (s *Server) updateRemoteSubscription(c *client, sub *subscription)`?

server/client.go
```diff
@@ -3125,6 +3117,16 @@ func (c *client) deliverMsg(sub *subscription, subject, reply, mh, msg []byte, g
...
+	srv := client.srv
+
+	if srv.gateway.enabled {
```
kozlovic, line 3123: I don't think ordering matters much, but trying to be consistent with most of the rest of the code (mainly in (*client).processUnsub(), I would change the order to do update route, then update gateway, then update leafnodes.

kozlovic (changes requested):
`*Client` does not exist. Anyway, had some more comments.

server/client.go
alexpantyukhin: I tried user the following sequence:
https://github.com/nats-io/nats-server/blob/master/server/client.go#L2972
It says that we need that sequence of execution.  There we update gateway, route, and leafs.
By the way, if we have to have different sequences in different places, then the PR is not needed. 

Should we have the order `route, then update gateway, then update leafnodes.` on both places?

kozlovic: Like I said, the order may not matter much, but the comment was saying something but then doing something else. See, in non defer cases, we would call update route, gateway then leafnodes. But in the defer (which reverse the order), we have leafnodes, route and gateway. I am guessing that the "defer srv.updateLeafNodes(client.acc, sub, -1)" was added later, and not at the right position.

Like I said, I think ordering is not mandatory, but I like to try to be consistent. So in your new function, you should have the "route, gateway and leafnodes" in that that order please. Order in that function will always be the same, and the good thing is that the whole function is called as a defer, so we don't have to think about reordering list of defers..

alexpantyukhin:
@kozlovic All mentions are fixed. Ready for review.

server/server.go
```diff
@@ -3356,3 +3356,12 @@ func (s *Server) setFirstPingTimer(c *client) {
 	d += time.Duration(addDelay)
 	c.ping.tmr = time.AfterFunc(d, c.processPingTimer)
 }
+
+func (s *Server) updateRemoteSubscription(c *client, sub *subscription) {
```
kozlovic, line 3360: Sorry for mentioning that now, but since we are just removing here, the name of the function should reflect that, or, maybe better, keep this name but add a `delta int` as the last parameter of this function (and pass `delta` in updateRouteSubscriptionMap(), gatewayUpdateSubInterest() and updateLeafNodes().

In the places you called updateRemoteSubscription() in this PR you would pass `-1`. There are only 1 or 2 places in the rest of the server code where we could possibly call this function with a positive delta, because in some cases we do some of the things based on the client's kind (CLIENT, etc..).
We internally need to have a deeper look at that, but I think having this function with a `delta` could make it be reusable later when we cleanup the rest of the code.

alexpantyukhin: Good input! I will fix asap

alexpantyukhin: @kozlovic  fixed.

kozlovic (approved):
LGTM. Thank you for your contribution!

PR #2066, opened 2021-04-05
[FIXED] Leafnode: incorrect loop detection in multi-cluster setup

kozlovic:
If leafnodes from a cluster were to reconnect to a server in
a different cluster, it was possible for that server to send
to the leafnodes some their own subscriptions that could cause
an inproper loop detection error.

There was also a defect that would cause subscriptions over route
for leafnode subscriptions to be registered under the wrong key,
which would lead to those subscriptions not being properly removed
on route disconnect.

Finally, during route disconnect, the leafnodes map was not updated.
This PR fixes that too.

### DIFF
--- a/server/leafnode.go
+++ b/server/leafnode.go
@@ -1256,10 +1256,13 @@ func (s *Server) initLeafNodeSmapAndSendSubs(c *client) {
 
 	// Now walk the results and add them to our smap
 	c.mu.Lock()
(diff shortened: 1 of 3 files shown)


derekcollison (approved):
LGTM

matthiashanel (approved):
LGTM

PR #2067, opened 2021-04-06
[fixed] issue with concurrent account fetch when account was incomplete

matthiashanel:
This happened when a dummy (expired/incomplete) account was created during
a route operation. The dummy was to avoid fetching the account, which would
cause a lock inversion.
When a non route request required the account, we'd download it as it is
set to expired.
A concurrent request would result in ErrAccountResolverSameClaims which
the caller did not handle.
Fix is to remove ErrAccountResolverSameClaims.

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -1318,9 +1318,12 @@ func (s *Server) updateAccountWithClaimJWT(acc *Account, claimJWT string) error 
 	if acc == nil {
 		return ErrMissingAccount
 	}
-	if acc.claimJWT != "" && acc.claimJWT == claimJWT && !acc.incomplete {
+	acc.mu.Lock()
+	sameClaim := acc.claimJWT != "" && acc.claimJWT == claimJWT && !acc.incomplete
+	acc.mu.Unlock()
+	if sameClaim {
 		s.Debugf("Requested account update for [%s], same claims detected", acc.Name)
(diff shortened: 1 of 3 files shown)


server/server.go
```diff
@@ -1318,9 +1318,12 @@ func (s *Server) updateAccountWithClaimJWT(acc *Account, claimJWT string) error
 	if acc == nil {
 		return ErrMissingAccount
 	}
-	if acc.claimJWT != "" && acc.claimJWT == claimJWT && !acc.incomplete {
+	acc.mu.Lock()
```
derekcollison, line 1321: I think this may be a RW lock.

derekcollison (approved):
LGTM

server/server.go
kozlovic: Looks like it indeed.

kozlovic (approved):
LGTM, but looks like lock could be made into rlock.

server/server.go
matthiashanel: fixed

derekcollison (approved):
LGTM

kozlovic (approved):
LGTM

kozlovic (commented):
Should not remove since this was exported. Please have a PR to add this back.

PR #2072, opened 2021-04-06
[FIXED] Websocket: TLS configuration not updated on reload

kozlovic:
Although we do not support websocket configuration changes, we usually
try to support changes to TLS certificates, etc..

The way websocket is handled (using an http server), the TLS config
was given on startup and updates following a configuration reload
would not be reflected.
Using a tls.Config function that allows passing the tls config
prior to handshake seem to workaround this issue.

I have added a test that demonstrate that the TLS configuration
is really updated after the reload.

### DIFF
--- a/server/websocket.go
+++ b/server/websocket.go
@@ -955,6 +955,7 @@ func (s *Server) startWebsocketServer() {
 	if o.TLSConfig != nil {
 		proto = wsSchemePrefixTLS
 		config := o.TLSConfig.Clone()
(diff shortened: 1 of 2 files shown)


derekcollison (approved):
LGTM

davidkemper (approved):
LGTM

kozlovic:
I have rebased from master to use @variadico new helper to remove config file in the new test (since helper checks/report errors).

PR #2076, opened 2021-04-07
Report an error if we fail to load an account on a stream assignment.

derekcollison:
We expected an account load to not fail but this is possible with resolvers.
If we can not load the account report back to the metaleader the error.

Consumers overlay the same peer group and need to exist prior, so I believe we only need to check on creating a stream, not on update or creating a consumer.

/cc @nats-io/core

### DIFF
--- a/server/jetstream_cluster_test.go
+++ b/server/jetstream_cluster_test.go
@@ -5284,6 +5284,26 @@ func TestJetStreamClusterServerLimits(t *testing.T) {
 	}
 }
 
+func TestJetStreamClusterAccountLoadFailure(t *testing.T) {
+	c := createJetStreamClusterWithTemplate(t, jsClusterLimitsTempl, "R3L", 3)
+	defer c.shutdown()
+
+	// Client based API
+	nc, js := jsClientConnect(t, c.leader())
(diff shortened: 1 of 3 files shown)


server/jetstream_cluster_test.go
```diff
@@ -5284,6 +5284,26 @@ func TestJetStreamClusterServerLimits(t *testing.T) {
...
+
+	// Remove the "ONE" account from non-leader
+	s := c.randomNonLeader()
+	s.mu.Lock()
+	s.accounts.Delete("ONE")
+	s.mu.Unlock()
+
+	_, err := js.AddStream(&nats.StreamConfig{Name: "F", Replicas: 3})
+	if err == nil || !strings.Contains(err.Error(), "account not found") {
+		t.Fatalf("Expected an 'account not found' error but got none")
```
kozlovic, line 5303: Would change to `.. but got %v`, since it could be `nil` or an unexpected error.

kozlovic (approved):
LGTM but would do a little change to the test failure report.

kozlovic (approved):
LGTM

PR #2077, opened 2021-04-07
Quiet race detector

derekcollison:

/cc @nats-io/core


kozlovic (changes requested):
I think that this should have been part of #2076 or then just fix the race here without the rest.

derekcollison:
Must not have gone from main branch and accidentally branched from another.

PR #2078, opened 2021-04-07
If a mirror was removed or changed this would panic.

derekcollison:
This protects against that.

/cc @nats-io/core


kozlovic (approved):
LGTM

PR #2079, opened 2021-04-07
Check errors when removing test directories and files

variadico:
Currently in tests, we have calls to `os.Remove` and `os.RemoveAll` where we don't check the returned error. This hides useful error messages when tests fail to run, such as "too many open files".

This change checks for more filesystem related errors and calls `t.Fatal` if there is an error.

Call updates
* `os.Remove(path)` gets updated to `removeFile(t, path)`
* `os.RemoveAll(path)` gets updated to `removeDir(t, path)`

Other minor changes:
* In `reload_test.go/TestConfigReloadEnableClusterAuthorization` and `reload_test.go/TestConfigReloadClusterRemoveSolicitedRoutes`, we were double removing files, resulting in an error. This change only removes that file once.
* In `reload_test.go/TestConfigReloadRotateFiles`, we were removing files that don't exist, resulting in an error. This change remove those unneeded calls.
* Some calls to `ProcessConfigFile` were ignoring the returned error. This change checks the error.

**Before change**
```
$ go test -p=1 -count=1 ./server/

--- FAIL: TestJetStreamTemplateFileStoreRecovery (10.35s)
panic: Unable to start NATS Server in Go Routine [recovered]
        panic: Unable to start NATS Server in Go Routine

```

**After change** - now "too many open files" error is surfaced. This is a good clue to set `ulimit` higher. (Higher `ulimit` fixes the panic.)
```
$ go test -p=1 -count=1 ./server/
--- FAIL: TestJetStreamTemplateFileStoreRecovery (10.58s)
    panic.go:965: open /tmp/nats-server/jstests-storedir-883699902: too many open files
--- FAIL: TestJetStreamSingleInstanceRemoteAccess (10.00s)
panic: Unable to start NATS Server in Go Routine [recovered]
        panic: Unable to start NATS Server in Go Routine
```

/cc @nats-io/core

### DIFF
--- a/logger/log_test.go
+++ b/logger/log_test.go
@@ -106,7 +106,7 @@ func TestStdLoggerTraceWithOutDebug(t *testing.T) {
 
 func TestFileLogger(t *testing.T) {
 	tmpDir := createDir(t, "_nats-server")
-	defer os.RemoveAll(tmpDir)
+	defer removeDir(t, tmpDir)
 
 	file := createFileAtDir(t, tmpDir, "nats-server:log_")
 	file.Close()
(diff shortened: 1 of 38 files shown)


kozlovic (approved):
LGTM. Thank you for this big task of updating all tests with these helpers. It will indeed help us better figure out why some tests fail. Thanks again!

PR #2084, opened 2021-04-08
expose the connection kind to CustomClientAuthentication

ripienaar:

/cc @nats-io/core

### DIFF
--- a/server/auth.go
+++ b/server/auth.go
@@ -47,6 +47,8 @@ type ClientAuthentication interface {
 	RegisterUser(*User)
 	// RemoteAddress expose the connection information of the client
 	RemoteAddress() net.Addr
+	// Kind indicates what type of connection this is matching defined constants like CLIENT, ROUTER, GATEWAY, LEAF etc
+	Kind() int
 }
 
 // NkeyUser is for multiple nkey based users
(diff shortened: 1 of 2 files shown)


kozlovic (commented):
I would approve, but knowing we are planning on a 2.2.2, a patch release, are we "allowed" to change APIs (even if in this case it is an addition, so I personally think it should be ok)?

ripienaar:
I believe adding is fine, and also there's no way that we've been following semver moving from 2.1.x to 2.2.x so I think that ship has sailed imo

kozlovic:
>  following semver moving from 2.1.x to 2.2.x

Maybe not, but there was no backward compatibility changes, if I am not mistaken, meaning that you can run a 2.2.x against a 2.1.x and vice versa. Of course some features are not enabled if not at the same level, but not a break like we had between 1.4 and 2.0 with routes, etc (hence the bump from v1 to v2). Anyway, I will LGTM it.

kozlovic (approved):
LGTM

ripienaar:
The protocol did not change that is correct and we made efforts to make things like headers degrade gracefully(ish).  But in numerous other data structures and behaviours I think were not quite up to semver standards.

We're following GoVer - do whatever it takes to avoid a major release :P

PR #2085, opened 2021-04-08
Fix for #2083

### LINKED ISSUE #2083: stream expired messages are mot removed from consumer pending ack list
A stream with age limits on the messages will remove messages once they hit that time.  However any messages that at that point were not acked on a consumer will continue to be considered as ack outstanding.

If the max ack outstanding is reached by such a consumer/stream the consmer is forever dead as it can never recover from that scenario.

Here is a stream that recently expired all its messages:

```
State:

             Messages: 0
                Bytes: 0 B
             FirstSeq: 101
              LastSeq: 100 @ 2021-04-08T11:09:40 UTC
     Active Consumers: 1
```

And a consumer on the same stream

```
State:

   Last Delivered Message: Consumer sequence: 100 Stream sequence: 10
     Acknowledgment floor: Consumer sequence: 0 Stream sequence: 0
         Outstanding Acks: 10 out of maximum 10
     Redelivered Messages: 10
     Unprocessed Messages: 0
```

At this point the consumer will never again deliver messages even if new messages are added to the stream

derekcollison:
Resolves #2083 

/cc @nats-io/core

### DIFF
--- a/server/consumer.go
+++ b/server/consumer.go
@@ -2916,10 +2916,22 @@ func (o *consumer) setInitialPending() {
 func (o *consumer) decStreamPending(sseq uint64, subj string) {
 	o.mu.Lock()
 	// Ignore if we have already seen this one.
-	if sseq >= o.sseq && o.sgap > 0 && o.isFilteredMatch(subj) && o.sgap > 0 {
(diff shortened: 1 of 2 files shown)


ripienaar:
LGTM but at least one of the new tests failed in one case - so they might be flappers

PR #2089, opened 2021-04-09
Fix flappers

derekcollison:

/cc @nats-io/core

### DIFF
--- a/server/jetstream_cluster_test.go
+++ b/server/jetstream_cluster_test.go
@@ -2298,6 +2298,9 @@ func TestJetStreamClusterUserSnapshotAndRestore(t *testing.T) {
 	if err := js.DeleteStream("TEST"); err != nil {
 		t.Fatalf("Unexpected error: %v", err)
 	}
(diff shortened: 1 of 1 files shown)


server/jetstream_cluster_test.go
```diff
@@ -5386,7 +5389,7 @@ func TestJetStreamClusterAckPendingWithMaxRedelivered(t *testing.T) {
 	}
 
 	// Send in 100 messages.
-	msg, toSend := make([]byte, 32), 100
+	msg, toSend := []byte("HELLO"), 100
 	rand.Read(msg)
```
kozlovic, line 5393: Is that necessary?

kozlovic (approved):
LGTM. Just use of rand.Read(msg) that does not seem to be required.

server/jetstream_cluster_test.go
derekcollison: Not as it is but when I was debugging and printing things out it was much easier.

derekcollison: Does not affect the test.

PR #2091, opened 2021-04-09
[Fixed] leaf node subscription permission negotiation.

matthiashanel:
On connect all subscription where sent by the soliciting leaf node.
If creds contains sub deny permissions, the leaf node would be
disconnected.
This waits for the permissions to be exchanged and checks permissions
before sending subscriptions.

split up leafNodeFinishConnectProcess into two functions.
When expecting multiple messages in a row, use a buffer to store everything after the left most match.
This can be used a lot more. But there are unrelated tests that rely on the old behavior. (will address separately)

### DIFF
--- a/server/leafnode.go
+++ b/server/leafnode.go
@@ -2185,10 +2185,7 @@ func (s *Server) leafNodeResumeConnectProcess(c *client) {
 			c.mu.Unlock()
 			return
 		}
-		if !c.ping.tmr.Stop() {
-			<-c.ping.tmr.C
-			c.ping.tmr = nil
-		}
+		clearTimer(&c.ping.tmr)
 		closed := c.isClosed()
 		c.mu.Unlock()
 		if !closed {
@@ -2223,10 +2220,7 @@ func (s *Server) leafNodeFinishConnectProcess(c *client) {
 	// Capture account before releasing lock
 	acc := c.acc
 	// cancel connectProcessTimeout
-	if !c.ping.tmr.Stop() {
-		<-c.ping.tmr.C
-		c.ping.tmr = nil
-	}
+	clearTimer(&c.ping.tmr)
 	c.mu.Unlock()
 
 	// Make sure we register with the account here.
--- a/server/leafnode.go
+++ b/server/leafnode.go
@@ -916,7 +916,7 @@ func (c *client) processLeafnodeInfo(info *Info) {
 	s = c.srv
 	c.mu.Unlock()
 
-	finishConnect := info.ConnectInfo && !firstINFO
+	finishConnect := info.ConnectInfo
 	if resumeConnect && s != nil {
 		s.leafNodeResumeConnectProcess(c)
 		if !info.InfoOnConnect {
@@ -1364,7 +1364,7 @@ func (s *Server) updateLeafNodes(acc *Account, sub *subscription, delta int32) {
 		ln.mu.Lock()
 		skip := sub.origin != nil && string(sub.origin) == ln.remoteCluster()
 		// do not skip on !ln.canSubscribe(string(sub.subject))
-		// Given allow:foo, > would be rejected. For leaf nodes filtering is done on the (soliciting) end (
+		// Given allow:foo, > would be rejected. For leaf nodes filtering is done on the (soliciting) end.
 		ln.mu.Unlock()
 		if skip {
 			continue
@@ -2124,6 +2124,8 @@ func (c *client) leafNodeSolicitWSConnection(opts *Options, rURL *url.URL, remot
 	return preBuf, 0, nil
 }
 
+const connectProcessTimeout = 2 * time.Second
+
 // This is invoked for remote LEAF remote connections after processing the INFO
 // protocol. This will do the TLS handshake (if needed be)
 func (s *Server) leafNodeResumeConnectProcess(c *client) {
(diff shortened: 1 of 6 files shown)


server/leafnode.go
```diff
@@ -2164,8 +2175,44 @@ func (s *Server) leafNodeFinishConnectProcess(c *client) {
 	// Spin up the write loop.
 	s.startGoRoutine(func() { c.writeLoop() })
 
+	cid := c.cid
+	c.mu.Unlock()
 	c.Debugf("Remote leafnode connect msg sent")
 
+	// timeout leafNodeFinishConnectProcess
+	time.AfterFunc(s.getOpts().PingInterval, func() {
```
kozlovic, line 2183: The PingInterval can be quite high (like 2min by default). Should we have a smaller interval instead (as a constant for now, not exposed)?

server/server.go
```diff
@@ -83,9 +83,11 @@ type Info struct {
 	LameDuckMode      bool     `json:"ldm,omitempty"`
 
 	// Route Specific
-	Import *SubjectPermission `json:"import,omitempty"`
-	Export *SubjectPermission `json:"export,omitempty"`
-	LNOC   bool               `json:"lnoc,omitempty"`
+	Import        *SubjectPermission `json:"import,omitempty"`
+	Export        *SubjectPermission `json:"export,omitempty"`
+	LNOC          bool               `json:"lnoc,omitempty"`
+	InfoOnConnect bool               `json:"info_on_connect,omitempty"` // When true the server will respond connect to with an INFO
```
kozlovic, line 89: Maybe: `When true the server will respond to CONNECT with an INFO`?

server/server.go
```diff
@@ -83,9 +83,11 @@ type Info struct {
...
+	ConnectInfo   bool               `json:"connect_info,omitempty"`    // When true this is the servers response to CONNECT
```
kozlovic, line 90: And `When true this is the server INFO response to CONNECT`

kozlovic (changes requested):
Some changes, but the important one would be to make sure that leafNodeFinishConnectProcess() is never invoked more than once, which right now we could be at risk to do in sendPermsInfo() ends-up be called more than once.

server/leafnode.go
```diff
@@ -1348,6 +1363,8 @@ func (s *Server) updateLeafNodes(acc *Account, sub *subscription, delta int32) {
 		// Check to make sure this sub does not have an origin cluster than matches the leafnode.
 		ln.mu.Lock()
 		skip := sub.origin != nil && string(sub.origin) == ln.remoteCluster()
+		// do not skip on !ln.canSubscribe(string(sub.subject))
+		// Given allow:foo, > would be rejected. For leaf nodes filtering is done on the (soliciting) end (
```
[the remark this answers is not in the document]
matthiashanel: fixed

server/leafnode.go
matthiashanel: A constant of 2 seconds? Or AuthTimeout?

server/server.go
matthiashanel: fixed

server/server.go
matthiashanel: fixed

server/leafnode.go
```diff
@@ -902,20 +903,27 @@ func (c *client) processLeafnodeInfo(info *Info) {
...
 		c.nc.SetDeadline(time.Time{})
-		finishConnect = true
-		s = c.srv
+		resumeConnect = true
 	}
+	s = c.srv
 	c.mu.Unlock()
 
-	if finishConnect && s != nil {
+	finishConnect := info.ConnectInfo && !firstINFO
```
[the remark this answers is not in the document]
matthiashanel: Would the check I do for the timer be an ok solution? essentially be a no op if already registerd.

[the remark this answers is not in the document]
matthiashanel: will add a flag

server/leafnode.go
kozlovic: Maybe AuthTimeout but 2sec at minimum?

matthiashanel: went with 2 seconds

server/leafnode.go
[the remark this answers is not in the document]
matthiashanel: added a flag

server/leafnode.go
matthiashanel: fixed

server/leafnode.go
```diff
@@ -2215,6 +2222,11 @@ func (s *Server) leafNodeFinishConnectProcess(c *client) {
 
 	// Capture account before releasing lock
 	acc := c.acc
+	// cancel connectProcessTimeout
+	if !c.ping.tmr.Stop() {
+		<-c.ping.tmr.C
```
kozlovic, line 2227: I know that this is what Go asks us to do, but I have seen too many times this block. So if you do that, I would recommend having a select{} here. But honestly, I would just have something like:
```
clearTimer(&c.ping.tmr)
```
which takes care of checking if timer is nil or not and then Stop() and set it to nil.
If the timer can't be stopped because it is about to fire, it will do nothing since it will check the flag that you have set at top of function.

kozlovic (changes requested):
Just a final change I think.

server/leafnode.go
matthiashanel: done

kozlovic (approved):
LGTM

PR #2093, opened 2021-04-09
[CHANGED] Flush in place only if producer and consumer are client connections

### LINKED ISSUE #2082: prevent stalled wait for client to gateway
## Feature Request

recently, I experienced network issue between clusters affect main cluster. gateway slowed down > client to gateway flushing slowed and client stalledWait occur.
it blocked some clients inbound message, as a result whole system blocked.

#### Use Case:
one main cluster & other sub clusters connected with gateway
massive messages produces on main cluster and copied to sub clusters

#### Proposed Change:
flushing gateway (maybe network issue) don't block client inbound message
 1. when client kind is GATEWAY flushClients() call flushSignal(), not flushOutbound()
 2. Add gateway option, not stall channel when client kind is GATEWAY
may need an option for proposal

gateway network issue don't affect to local cluster

#### Alternative Approaches
small write_deadline & max_pending reduce stalledWait but exist

shkim-will:
 - [ ] Tests added

Resolves #2082

 - flushing outbound is not to block the client read loop
 -
 -

/cc @nats-io/core

### DIFF
--- a/server/client.go
+++ b/server/client.go
@@ -1005,7 +1005,8 @@ func (c *client) flushClients(budget time.Duration) time.Time {
 			continue
 		}
 
-		if budget > 0 && cp.out.lft < 2*budget && cp.flushOutbound() {
+		// Try to flush in place, if producer and consumer are client.
+		if c.kind == CLIENT && cp.kind == CLIENT && budget > 0 && cp.out.lft < 2*budget && cp.flushOutbound() {
 			budget -= cp.out.lft
 		} else {
 			cp.flushSignal()
--- a/server/client.go
+++ b/server/client.go
@@ -981,10 +981,18 @@ func (c *client) writeLoop() {
 
 // flushClients will make sure to flush any clients we may have
 // sent to during processing. We pass in a budget as a time.Duration
-// for how much time to spend in place flushing for this client. This
-// will normally be called in the readLoop of the client who sent the
-// message that now is being delivered.
+// for how much time to spend in place flushing for this client.
 func (c *client) flushClients(budget time.Duration) time.Time {
+	return c.flushClientsWithCheck(budget, false)
+}
+
+// flushClientsWithCheck will make sure to flush any clients we may have
+// sent to during processing. We pass in a budget as a time.Duration
+// for how much time to spend in place flushing for this client.
+// The 'clientsKindOnly' boolean indicates whether to check kind of client
+// and pending client to run flushOutbound in flushClientsWithCheck.
+// flushOutbound could be block caller when pending client is 'stuck'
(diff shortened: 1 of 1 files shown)


server/client.go
```diff
@@ -1005,7 +1005,8 @@ func (c *client) flushClients(budget time.Duration) time.Time {
 			continue
 		}
 
-		if budget > 0 && cp.out.lft < 2*budget && cp.flushOutbound() {
+		// Try to flush in place, if producer and consumer are client.
+		if c.kind == CLIENT && cp.kind == CLIENT && budget > 0 && cp.out.lft < 2*budget && cp.flushOutbound() {
```
kozlovic, line 1009: No, we can't do that here in this function since it is called in various place with a budget where the expectation is that message is possibly flushed in place (or effort to do so is done). And it can be situations where the kind is not CLIENT (think internal send queues, JetStream, etc...).
To minimize risks, I would then limit the single call that we make in readLoop() where we pass a budget, and only that one behaving this way. This probably means having a function flushClients() that behaves the way it does today, and a different version that can have this check.
flushClients(budget) invokes flushClientsWithCheck(budget, clientsKindOnly) with clientsKindOnly==false.
Then the check could be:
```
if budget > 0 && (!clientsKindOnly || c.kind == CLIENT && cp.kind == CLIENT) && cp.out.lft...
```
Note that I am bad at naming things, so if you can think of a better function name and field, go for it :-)

kozlovic (changes requested):

server/client.go
```diff
@@ -981,10 +981,18 @@ func (c *client) writeLoop() {
...
 func (c *client) flushClients(budget time.Duration) time.Time {
+	return c.flushClientsWithCheck(budget, false)
+}
+
+// flushClientsWithCheck will make sure to flush any clients we may have
+// sent to during processing. We pass in a budget as a time.Duration
+// for how much time to spend in place flushing for this client.
+// The 'clientsKindOnly' boolean indicates whether to check kind of client
+// and pending client to run flushOutbound in flushClientsWithCheck.
+// flushOutbound could be block caller when pending client is 'stuck'
```
kozlovic, line 994: Would rephrase:
```
flushOutbound() could block the caller up to the write deadline when the receiving client cannot drain data from the socket fast enough.
```

kozlovic:
@derekcollison I have reviewed and I think it is ok, but may want to delay merge post 2.2.2 to give this a bit more mileage? wdyt?

derekcollison:
Yes agree, let's move to next release.

kozlovic:
Since we have now released 2.2.2, I am going to merge this.

PR #2094, opened 2021-04-09
JS Headers and case-sensitivity

wallyqs:
Currently JS Headers such as `Nats-Expected-Last-Sequence` are case-sensitive in the server to have a fast path and avoid using `ReadMIMEHeader`.

In this change, we try to keep the fast path but then add another check for cases when the header may have been sent in a different case than expected.

/cc @nats-io/core

### DIFF
--- a/server/client.go
+++ b/server/client.go
@@ -3653,9 +3653,16 @@ func getHeader(key string, hdr []byte) []byte {
 	if len(hdr) == 0 {
 		return nil
 	}
+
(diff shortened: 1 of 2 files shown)


wallyqs:
Most of the time this should not be a problem since it is controlled in the client how these headers are being used via options, so we could potentially keep as is as well for now.

derekcollison:
Can this be closed?

PR #2095, opened 2021-04-09
Mixed mode improvements.

derekcollison:
1. When in mixed mode and only running the global account we now will check the account for JS.
2. Added code to decrease the cluster set size if we guessed wrong in mixed mode setup.

/cc @nats-io/core

### DIFF
--- a/server/jetstream_cluster.go
+++ b/server/jetstream_cluster.go
@@ -786,7 +786,7 @@ func (js *jetStream) monitorCluster() {
 // This is called on first leader transition to double check the peers and cluster set size.
 func (js *jetStream) checkClusterSize() {
 	s, n := js.server(), js.getMetaGroup()
-	if n == nil || !n.Leader() {
+	if n == nil {
 		return
 	}
 	// We will check that we have a correct cluster set size by checking for any non-js servers
--- a/server/jetstream_cluster.go
+++ b/server/jetstream_cluster.go
@@ -740,6 +740,7 @@ func (js *jetStream) monitorCluster() {
 	}
 
 	isRecovering := true
+	beenLeader := false
 
 	for {
 		select {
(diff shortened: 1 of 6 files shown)


server/jetstream_cluster.go
```diff
@@ -767,12 +768,55 @@ func (js *jetStream) monitorCluster() {
...
+			if si.(nodeInfo).js {
+				totalJS++
+			}
+		}
+	}
+	// If we have less then our cluster size adjust that here. Can not do individual peer removals since
+	// they will not be in the tracked peers.
+	if totalJS < ps.clusterSize {
+		s.Debugf("Adjusting JetStream cluster size from %d to %d", ps.clusterSize, totalJS)
+		if err := n.AdjustClusterSize(totalJS); err != nil {
```
kozlovic, line 814: Question: if size is adjusted down, can it ever go back up? If so where would that happen? (since here we do it only on first time this node goes to a leader state)

kozlovic (commented):
I am not sure why, but this test fails in this branch (even when running locally):
```
=== RUN   TestLeafNodePermissions
    leafnode_test.go:1225: Expected 2 subs, got 4
--- FAIL: TestLeafNodePermissions (1.07s)
```
and does not fail in this branch if reseting to before your commit, or does not fail in main branch...

server/jetstream_cluster.go
derekcollison: Yes we support adding and removing peers.

`TestJetStreamExpandCluster` is a test that shows how we add in peers.

kozlovic (changes requested):
Debugging the LeafNode failure in this branch, we have 4 subs instead of 2, and they are:
```
@@IK: sub="$JS.API.>"
@@IK: sub="$JS.API.>"
@@IK: sub="$LDS.b1Lr0PAfkHZJVoDM889ABs"
@@IK: sub=">"
```
which is related to the change were you enable JS.. Not sure why we endup with 2 subs on the same subject.

Note that this test has changed in main branch, so you could rebase to see...

kozlovic:
Also maybe rebase so you benefit from your previous PR about flappers since you still get failure about the ack pending tests.

server/server.go
```diff
@@ -1582,6 +1583,12 @@ func (s *Server) Start() {
 			}
 			return true
 		})
+
+		// We may be in a mixed mode here.
+		if !s.standAloneMode() && s.globalAccountOnly() {
```
kozlovic, line 1588: This can affect lots of test because as soon as you use DefaultOptions() which set a cluster port, then this test will be true: that is, stanAloneMode() is false because cluster port is defined. And if test does not specify specific accounts, then we add JS subscriptions.. maybe it is correct, but seem to be a bit intrusive..

derekcollison: I will check to make sure its not processed twice from above. Will also rebase.

derekcollison: Without that change you can have a case where some $G are enabled and those on servers that do not have JS enabled will not properly allow things to be forwarded.

I could take a different approach now with clustering and not turn anything on at all.. WDYT?

kozlovic: Not sure what the approach would be, but I hope you get my point that in a situation where I have simply clustering enabled (as in routes) and no JS at all, it seem a bit intrusive to have the system add things related to JetStream. Now it could be that there is something JS related in this test or something that has to do with default options, etc..
Now if we want to support mixed-mode without the user configuring anything in the non-js node, then I understand that the server has to do some automatic configuration...

server/jetstream_cluster.go
```diff
@@ -767,12 +768,55 @@ func (js *jetStream) monitorCluster() {
...
 		case <-t.C:
 			doSnapshot()
 		}
 	}
 }
 
+// This is called on first leader transition to double check the peers and cluster set size.
+func (js *jetStream) checkClusterSize() {
+	s, n := js.server(), js.getMetaGroup()
+	if n == nil || !n.Leader() {
```
kozlovic, line 789: The check of `n.Leader()` bothers me. We are called because we became leader. Having the test here means to me that the state can change underneath us.. if that is the case, then what's the validity of this call if the state suddenly change after this check?

derekcollison: Will remove.

server/server.go
derekcollison: This was a misconfiguration, will remove all of this code.

kozlovic (approved):
LGTM!

PR #2096, opened 2021-04-09
Run gofmt -s to simplify code

variadico:
This runs `gofmt -s -w .` at the root of this repo. `gofmt` is part of the default Go distribution and it's what `go fmt` runs under the hood.

/cc @nats-io/core

### DIFF
--- a/server/auth_test.go
+++ b/server/auth_test.go
@@ -108,7 +108,7 @@ func TestUserCloneNil(t *testing.T) {
 
 func TestUserUnknownAllowedConnectionType(t *testing.T) {
 	o := DefaultOptions()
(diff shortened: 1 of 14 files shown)


kozlovic (approved):
LGTM! Clearly, based on the quantity of changes in mqtt_test.go, which I wrote, I am not good at using the simple version in those slice of pointers to structures. It is not natural to me to have something like: `[]*User{{Username: "user}}`  compared to `[]*User{&User{Username: "user"}}` because in the second form I can clearly see that I am adding a pointer to a User struct{}, where as in the first form (gofmt -s -w), it looks like we are adding/initializing a struct...

PR #2098, opened 2021-04-10
Add mixedmode test back in

derekcollison:
Added back the mixed mode test with accounts to have a test for cluster size stepdown.

/cc @nats-io/core


kozlovic (approved):
LGTM

Issue #2068, opened 2021-04-06
When the messages cached in the stream are empty, kill -9 nats-server and start again, and firstSeq and firstSeq become 0


GuangchaoDeng:
## Defect

When the messages cached in the stream are empty, kill -9 nats-server and start again, and firstSeq and firstSeq become 0

#### Versions of `nats-server` and affected client libraries used:
nats-server : 2.2.1
nats : 0.0.22
#### OS/Container environment:
Red Hat Enterprise Linux Server release 7.2 (Maipo)  or other
#### Steps or code to reproduce the issue:
**Describe：**

1、Build a leaf node and persist it in file mode

2、Create stream:denggc8, Consumer :denggc8, for subject: denggc8

`nats  -s nats://leaf_1:leaf_1@localhost:5222 str add denggc8  --subjects="denggc8  " --storage=file --retention=workq --discard=new --max-msgs=-1 --max-bytes=-1 --max-age=-1 --max-msg-size=-1 --dupe-window=5s --replicas=1`

`nats  -s nats://leaf_1:leaf_1@localhost:5222 con add denggc8   denggc8  2  --deliver=all --replay=original --filter="denggc8  " --max-deliver=-1 --max-pending=1 --pull`

3、Send 10 messages to Stream Denggc8; And through Consumer: Denggc8 to complete the consumption

stream和consumer info：
![before](https://user-images.githubusercontent.com/26615322/113706083-d85a4e80-9710-11eb-966e-274a1d6a6ba7.jpg)
4、kill -9  leaf node nats-server

5、start leaf node nats-server again

6、stream和consumer info：

![after](https://user-images.githubusercontent.com/26615322/113706126-e6a86a80-9710-11eb-8485-24cf67c882a5.jpg)

#### Expected result:
![before](https://user-images.githubusercontent.com/26615322/113706166-f1fb9600-9710-11eb-9f45-8192d047d729.jpg)

#### Actual result:
![after](https://user-images.githubusercontent.com/26615322/113706180-f6c04a00-9710-11eb-9505-468d6035b3a0.jpg)

ripienaar:
The kill -9 will corrupt the index, startup will rebuild the index from scratch.

Because the stream is empty this will result in a reset state for sequences. We might need to do some resetting on consumers or derive the correct sequences some other way.  Not sure what our options are

The tl;dr here is:

 * WQ stream
 * Consume all the messages so consumers advance in sequences
 * kill -9 the server
[...]

PR #2099, opened 2021-04-10
Partial fix for #2068

derekcollison:
Signed-off-by: Derek Collison derek@nats.io

/cc @nats-io/core

### DIFF
--- a/server/filestore.go
+++ b/server/filestore.go
@@ -616,6 +616,11 @@ func (mb *msgBlock) rebuildState() (*LostStreamData, error) {
 		index += rl
 	}
 
+	// For empty msg blocks make sure we recover last seq correctly based off of first.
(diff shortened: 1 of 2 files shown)


ripienaar (approved):
LGTM

server/filestore.go
```diff
@@ -616,6 +616,11 @@ func (mb *msgBlock) rebuildState() (*LostStreamData, error) {
 		index += rl
 	}
 
+	// For empty msg blocks make sure we recover last seq correctly based off of first.
+	if mb.msgs == 0 {
+		mb.last.seq = mb.first.seq - 1
```
kozlovic, line 621: Do you need to take care of special case with mb.first.seq could be 0, and therefore mb.last.seq would end-up being -1?

PR #2100, opened 2021-04-10
Only adjust cluster size when we have heard from all peers.

derekcollison:

/cc @nats-io/core


kozlovic (approved):
LGTM

PR #2101, opened 2021-04-12
[fixing] subscription issue when subscribing to a super set of deny_i…

matthiashanel:
…mport

If the subscription was foo. > but the server also had an import deny of foo.bar
It was legal to send the subscription. But the other server was unaware
of the restriction and sent the message anyway. The check of the
incoming message did not happen.

Fixing by ignoring messages the server is not supposed to receive.
And exchange deny_import so that the non soliciting leaf node knows to not
send these messages in the first place.

NB. merging of deny_ export/import with perms from INFO happens in processLeafnodeInfo

### DIFF
--- a/server/client.go
+++ b/server/client.go
@@ -900,6 +900,25 @@ func (c *client) setPermissions(perms *Permissions) {
 	}
 }
 
+// Merge client.perms structure with additional pub deny permissions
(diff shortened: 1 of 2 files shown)


kozlovic (approved):
LGTM

PR #2103, opened 2021-04-12
Make TERM exit 1 as pre v2.2 series

wallyqs:
Before 2.2 series, the TERM signal used to not be handled by the server, so it would not have been a clean exit. In 2.2, it was changed to process TERM signal as a clean exit but this affects the behavior of some tools that were expecting TERM to be exit 1.

/cc @nats-io/core

### DIFF
--- a/server/signal.go
+++ b/server/signal.go
@@ -59,7 +59,7 @@ func (s *Server) handleSignals() {
 
 					if !ldm {
 						s.Shutdown()
-						os.Exit(0)
+						os.Exit(1)
 					}
 				case syscall.SIGUSR1:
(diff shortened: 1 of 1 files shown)


ripienaar (approved):
LGTM

kozlovic:
@wallyqs In the PR description and summary you are talking about exit code `-1`, yet in the code you use `1`. So which one is right?

wallyqs:
@kozlovic sorry, I meant exit 1. Updated the PR description...

philpennock (approved):

kozlovic (approved):
LGTM

Issue #2097, opened 2021-04-10
the messeges  are deleted  after rm consumer on InterestPolicy stream


GuangchaoDeng:
1、create stream:t24 and consumer:t24，The details are as follows:

Configuration:

             Subjects: t24
     Acknowledgements: true
            Retention: File - Interest
             Replicas: 1
       Discard Policy: New
     Duplicate Window: 2m0s
     Maximum Messages: unlimited
        Maximum Bytes: unlimited
          Maximum Age: 0.00s
 Maximum Message Size: unlimited
    Maximum Consumers: unlimited

State:

             Messages: 0
                Bytes: 0 B
             FirstSeq: 0
              LastSeq: 0
     Active Consumers: 0
     
     Configuration:
    
            Durable Name: t24
        Delivery Subject: t244
             Deliver All: true
              Ack Policy: All
                Ack Wait: 30s
           Replay Policy: Instant
         Max Ack Pending: 3
    
    State:
    
       Last Delivered Message: Consumer sequence: 0 Stream sequence: 0
         Acknowledgment floor: Consumer sequence: 0 Stream sequence: 0
             Outstanding Acks: 0 out of maximum 3
         Redelivered Messages: 0
         Unprocessed Messages: 0

2、pub 11 messeges to subject ：t24; Subscribe to subject: t244, get 3 messages, and don't reply to ack.

The details are as follows:

`[leaf_1@host5 nats-cli]$ ./nats  -s nats://leaf_1:leaf_1@localhost:5222 sub t244`
`22:29:14 Subscribing on t244`
`[#1] Received JetStream message: consumer: t24 > t24 / subject: t24 / delivered: 1 / consumer seq: 1 / stream seq: 1 / ack: false`
`1`

`[#2] Received JetStream message: consumer: t24 > t24 / subject: t24 / delivered: 1 / consumer seq: 2 / stream seq: 2 / ack: false`
`2`

`[#3] Received JetStream message: consumer: t24 > t24 / subject: t24 / delivered: 1 / consumer seq: 3 / stream seq: 3 / ack: false`

Configuration:

        Durable Name: t24
    Delivery Subject: t244
         Deliver All: true
          Ack Policy: All
            Ack Wait: 30s
       Replay Policy: Instant
     Max Ack Pending: 3

State:

   Last Delivered Message: Consumer sequence: 3 Stream sequence: 3
     Acknowledgment floor: Consumer sequence: 0 Stream sequence: 0
[...]

PR #2105, opened 2021-04-12
Pull based consumers when deleted were not removing messages from an interest retention stream.

derekcollison:
Fix for #2097.

/cc @nats-io/core


kozlovic (approved):
LGTM

PR #2106, opened 2021-04-12
Consolidate test running logic

variadico:
Currently, we have different commands to run tests spread across
different YAML files and comments in Go files.

This change consolidates all the different commands, environment
variables, and OS limits into 1 file that Travis, GitHub, and engineers
can use to run tests.

This change doesn't require engineers to alter their workflow.

/cc @nats-io/core

### DIFF
--- a/.github/workflows/go-test.yaml
+++ b/.github/workflows/go-test.yaml
@@ -9,7 +9,6 @@ jobs:
 
     env:
       GOPATH: /home/runner/work/nats-server
-      GO111MODULE: "off"
 
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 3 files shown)


derekcollison (commented):
Not a fan of Makefiles in Go projects. Like the idea, maybe post 2.2.2

kozlovic (approved):
LGTM. But @derekcollison if desired, we can merge only after 2.2.2 is released?

variadico:
Converted to draft so we don't accidentally merge for 2.2.2 release.

ripienaar:
If not make then maybe https://magefile.org/ or just a bash script.

PR #2108, opened 2021-04-13
Improve JS when a leafnode cluster extends and shares a system account.

derekcollison:
This PR improves the setup that has a remote leafnode cluster extending a cluster or supercluster and the JetStream domain. We want to ensure the remote cluster places correctly and does not try to elect its own leader.

/cc @nats-io/core


kozlovic (approved):
LGTM

matthiashanel (approved):
LGTM

PR #2110, opened 2021-04-13
Mirrors failed when upstream messages had expired.

derekcollison:
This fixes that scenario by detecting the situation and inserting skip msgs to catch up.

/cc @nats-io/core


ripienaar (approved):
LGTM

kozlovic (approved):
LGTM

PR #2111, opened 2021-04-14
Add in more extensive test on extending hubs with leafnode clusters

derekcollison:
Added more test to make sure we properly scale up and down peer set sizes when extending JetStream across hub and leafnode clusters.

/cc @nats-io/core


matthiashanel (approved):
LGTM

PR #2116, opened 2021-04-14
Allow control messages like heartbeats to pass the old sub test.

derekcollison:

/cc @nats-io/core


kozlovic (approved):
LGTM

matthiashanel (approved):
LGTM

PR #2117, opened 2021-04-14
Fix calling FailNow in parent test

variadico:
Currently, TestJetStreamClusterMirrorExpirationAndMissingSequences calls
FailNow from the parent test when a subtest fails. This causes an error.

This change calls FailNow to fail the subtest from the subtest itself.

Fixes this error in CI:

```
Error:     testing.go:1103: test executed panic(nil) or runtime.Goexit: subtest may have called FailNow on a parent test
--- FAIL: TestJetStreamClusterMirrorExpirationAndMissingSequences (7.49s)
```

/cc @nats-io/core


derekcollison (approved):
LGTM

kozlovic (approved):
LGTM

PR #2118, opened 2021-04-14
Cleanup request subscription if no answer

derekcollison:

/cc @nats-io/core


matthiashanel (approved):
LGTM

PR #2119, opened 2021-04-15
Do not limit expansion of new peers

derekcollison:
Under scenarios where peers would not be known prior to the metagroup leader being elected, they would artificially not be allowed to properly be added.

/cc @nats-io/core


kozlovic:
Just ran JS tests locally and got a failure:
```
=== RUN   TestJetStreamClusterPeerRemovalAPI
    jetstream_cluster_test.go:3050: Still in the peer list
--- FAIL: TestJetStreamClusterPeerRemovalAPI (0.47s)
```
suspect that this is related to this change, no?

derekcollison:
Looking now.

matthiashanel (approved):
LGTM

PR #2121, opened 2021-04-15
js: msgblock close fix

wallyqs:
This fixes the issue of sometimes file descriptors remaining open after shutdown in tests. Also swaps order of cleaning up resource directory in cluster tests to prevent artifacts from previous tests remain in temp storage dir.

/cc @nats-io/core


kozlovic (approved):
LGTM

variadico (approved):

PR #2125, opened 2021-04-16
[fixed] only become observer if the leaf config has raft not restricted

matthiashanel:
If a subject in the system accounts leafnode deny_imports matches $NRG.>
then jetstream is explicitly disconnected and the server can become
leader.

### DIFF
--- a/server/jetstream_cluster.go
+++ b/server/jetstream_cluster.go
@@ -487,10 +487,17 @@ func (js *jetStream) setupMetaGroup() error {
 	// but do not form our own.
 	if ln := s.getOpts().LeafNode; len(ln.Remotes) > 0 {
 		sys := s.SystemAccount().GetName()
+	ENDFOR:
 		for _, r := range ln.Remotes {
 			if r.LocalAccount == sys {
+				for _, denySub := range r.DenyImports {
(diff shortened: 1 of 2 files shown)


server/jetstream_cluster.go
```diff
@@ -487,10 +487,17 @@ func (js *jetStream) setupMetaGroup() error {
 	// but do not form our own.
 	if ln := s.getOpts().LeafNode; len(ln.Remotes) > 0 {
 		sys := s.SystemAccount().GetName()
+	ENDFOR:
 		for _, r := range ln.Remotes {
 			if r.LocalAccount == sys {
+				for _, denySub := range r.DenyImports {
+					if subjectIsSubsetMatch(denySub, "$NRG.>") {
```
derekcollison, line 494: Make const that loves in raft.go so we don't break things later on?

derekcollison (approved):
LGTM - Maybe make prefix a const that is from raft.go.

server/jetstream_cluster.go
matthiashanel: fixed.

derekcollison (approved):
LGTM

PR #2127, opened 2021-04-19
Declare required JetStream resources in test

variadico:
Currently, tests that use js-op.conf use dynamic JetStream limits,
calculated from the machine where the test is run. However, this file
also has an account JWT that requires an explicit amount of resources.
This causes some tests to pass on some machine, but not others.

This change explicitly declares the max storage and memory limits
JetStream requires to match the limits specified in the account JWT, so
that the host environment doesn't influence the tests.

tl;dr; Dynamic limits correctly determines MaxStore should be 2.75GB for my Linux laptop, but the account requires 10GB. This causes the test to fail. For `/tmp/nats/jetstream`, my laptop only has 3.8GB available because tmpfs is in memory, not disk.

/cc @nats-io/core

### DIFF
--- a/server/configs/js-op.conf
+++ b/server/configs/js-op.conf
@@ -7,6 +7,8 @@ server_name: "S1"
 # The test will cleanup this directory so if you change it search for the test.
 jetstream {
   store_dir: "/tmp/nats-server/js-op-test"
+  max_memory_store: 1073741824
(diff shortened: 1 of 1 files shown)


derekcollison (approved):
LGTM - Thanks!

kozlovic (approved):
LGTM

PR #2128, opened 2021-04-19
[added] ability to respond to stream/consumer names/info

matthiashanel:
If a follower was disconnected long enough from the meta group leader,
it will respond to stream/consumer names/info requests with a delay.

If a leader responded or if the response is complete is included in the response.

in it's current form TestJetStreamClusterNoQuorumStepdown fails as it tries to list all streams past having no leader.
I can fix the test or we can look at the client and have it filter non leader or non complete responses.

### DIFF
--- a/server/jetstream_api.go
+++ b/server/jetstream_api.go
@@ -335,7 +335,8 @@ type JSApiStreamNamesRequest struct {
 type JSApiStreamNamesResponse struct {
 	ApiResponse
 	ApiPaged
-	Streams []string `json:"streams"`
+	Streams  []string `json:"streams"`
+	ByLeader bool     `json:"by_meta_leader"`
 }
 
 const JSApiStreamNamesResponseType = "io.nats.jetstream.api.v1.stream_names_response"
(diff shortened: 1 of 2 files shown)


derekcollison (commented):
In general I am in favor of an approach that permits reads from other members that are not the leader. 

I do believe these should be forming a distributed queue group however vs random delays. This is more work but is better in the long run IMO.

Also, we should probably only consider the names version of streams and consumers. The reports will have the responder probe for the actual asset leaders to get detailed information. In a situation where the meta leader is not reachable this could have quite a few probes running all in parallel.

I like the idea but I think we need to strategize a bit more on the approach we want.

server/jetstream_api.go
```diff
@@ -1480,7 +1502,17 @@ func (s *Server) jsStreamNamesRequest(sub *subscription, c *client, subject, rep
 	resp.Total = numStreams
 	resp.Limit = JSApiNamesLimit
 	resp.Offset = offset
-	s.sendAPIResponse(ci, acc, subject, reply, string(msg), s.jsonResponse(resp))
+	if responseDelay != 0 {
+		// Need to copy these off before sending..
+		msg = append(msg[:0:0], msg...)
```
kozlovic, line 1507: For info if later this PR is accepted or revisited: Since sendAPIResponse() seem to accept a string for the "msg" and not a slice, (see below `string(msg)`) you should capture the conversion as a string here, before starting the go routine. Something like this:
```
strMsg := string(msg)
s.startGoRoutine(func() {
			defer s.grWG.Done()
			time.Sleep(responseDelay)
			s.sendAPIResponse(ci, acc, subject, reply, strMsg, s.jsonResponse(resp))
		})
```
This applies to all places where you made a copy because of sending in a separate go routine.

PR #2129, opened 2021-04-19
Return not ready for connection reason

variadico:
Currently, we use ReadyForConnections in server tests to wait for the
server to be ready. However, when this fails we don't get a clue about
why it failed.

This change adds a new unexported method called readyForConnections that
returns an error describing which checks failed. The exported
ReadyForConnections version works exactly as before. The unexported
version gets used in internal tests only.

Ulimit 256 too low error - before
```
panic: Unable to start NATS Server in Go Routine
```
Ulimit 256 too low error - with this change
```
panic: failed to be ready for connections after 10s: leafNode, server, route
```

Interestingly when I get to higher ulimit values, I can now see that the route check
consistently fails. Need to investigate further.

Ulimit 65535 - before
```
panic: Unable to start NATS Server in Go Routine
```
Ulimit 65535 - with this change
```
panic: failed to be ready for connections after 10s: route
```

/cc @nats-io/core

### DIFF
--- a/server/accounts_test.go
+++ b/server/accounts_test.go
@@ -214,8 +214,8 @@ func TestAccountIsolationExportImport(t *testing.T) {
 			s := opTrustBasicSetup()
 			defer s.Shutdown()
 			go s.Start()
-			if !s.ReadyForConnections(5 * time.Second) {
-				t.Fatal("failed to be ready for connections")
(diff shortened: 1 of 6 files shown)


variadico:
Actually, closing for now.

PR #2131, opened 2021-04-20
General Updates and Stability Improvements

derekcollison:

/cc @nats-io/core


server/raft.go
```diff
@@ -2564,7 +2564,7 @@ func (n *raft) processAppendEntry(ae *appendEntry, sub *subscription) {
 			select {
 			case n.applyc <- &CommittedEntry{n.commit, ae.entries[:1]}:
 			default:
-				n.debug("Failed to place snapshot entry onto our apply channel")
+				n.warn("Failed to place snapshot entry onto our apply channel")
```
ripienaar, line 2567: If i saw this, I would have no way to figure out what I could do about that. Is it just a wait and see kind of situation or what would be the remedy? Warn level logs need to be actionable some how - or indicate that they will hopefully clear

server/norace_test.go
```diff
@@ -1755,3 +1755,248 @@ func TestNoRaceJetStreamClusterSourcesMuxd(t *testing.T) {
...
+
+	c.restartServer(ms)
+	c.checkClusterFormed()
+	c.waitOnStreamLeader("$G", "M")
+
+	sendBatch(10)
+	checkMirror(20)
+}
+
+func TestNoRaceJetStreamClusterSuperClusterRIPStress(t *testing.T) {
```
ripienaar, line 1842: 🤣

ripienaar (approved):
generally seems fine - as in I ran these things in my tests but really, one for Ivan to review :) Currently running the latest in my cluster and so far seem stable

server/raft.go
derekcollison: I promoted it so it could be more easily reported to us via support channels etc.

server/stream.go
```diff
@@ -1352,7 +1386,7 @@ func (mset *stream) setupMirrorConsumer() error {
 	if ext != nil {
 		deliverSubject = strings.ReplaceAll(ext.DeliverPrefix+syncSubject(".M"), "..", ".")
 	} else {
-		deliverSubject = syncSubject("$JS.M")
+		deliverSubject = syncSubject("$JSDC.M")
```
kozlovic, line 1389: Is that permanent or what is part of some debug/test?

kozlovic (commented):
Question about $JS.M -> $JSDC.M and same for $JS.S

server/stream.go
derekcollison: Tracking GW sub leaks I thought it was better to not be under the $JS.> envelope since ephemerals go away when they detect interest goes away.

I could put it back.

kozlovic: Just trying to reduce or at least be aware of all subjects that potentially need to be allowed/denied when it comes to permissions or import/export rules.

derekcollison: Running a sub leak test now that R.I. showed me, so trying to reproduce that now.

kozlovic (approved):
LGTM

kozlovic (approved):
LGTM

PR #2134, opened 2021-04-20
Fix account resolver lock ordering on shutdown

matthiashanel:
This is benign as the inversion was between start and stop.

with this in place the race detector is happy on shutdown


derekcollison (approved):
LGTM

kozlovic (approved):
LGTM

PR #2135, opened 2021-04-20
Check that we have mirror still to not panic and deadlock/hang

derekcollison:

/cc @nats-io/core

### DIFF
--- a/server/stream.go
+++ b/server/stream.go
@@ -1139,6 +1139,10 @@ func (mset *stream) processMirrorMsgs() {
 
 	// Grab stream quit channel.
 	mset.mu.Lock()
+	if mset.mirror == nil {
+		mset.mu.Unlock()
+		return
+	}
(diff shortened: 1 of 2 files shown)


server/stream.go
```diff
@@ -1139,6 +1139,10 @@ func (mset *stream) processMirrorMsgs() {
 
 	// Grab stream quit channel.
 	mset.mu.Lock()
+	if mset.mirror == nil {
```
kozlovic, line 1142: and is it possible that mset.mirror.msgs be nil? because if so, then you would get a panic also because of that.

kozlovic: we don't apparently.

derekcollison: I double checked and we are good, I just nil out mset.mirror.

kozlovic (approved):
LGTM

PR #2136, opened 2021-04-21
[FIXED] Possible panic due to concurrent access to unlocked map

kozlovic:
This could happen when a leafnode has permissions set and another
connection (client, etc..) is about to assign a message to the
leafnode while the leafnode itself is receiving messages and they
both check permissions at the same time.


server/client.go
```diff
@@ -3269,32 +3270,38 @@ func (c *client) pruneDenyCache() {
...
+		c.perms.pcache.Delete(k)
+		if r++; (r > pruneSize && atomic.LoadInt32(&c.perms.pcsz) < int32(maxPermCacheSize)) ||
+			(r > maxPruneAtOnce) {
+			return false
 		}
-	}
+		return true
+	})
+	atomic.AddInt32(&c.perms.pcsz, -int32(r))
+	atomic.CompareAndSwapInt32(&c.perms.prun, 1, 0)
```
derekcollison, line 3284: This could just be a set to 0 yes?

derekcollison (approved):
LGTM

server/client.go
kozlovic: Yes, will change.

matthiashanel, line 3284: if we are in this function c.perms.prun has the value of 1, meaning a simple atomic store would work as well.

server/client.go
```diff
@@ -3325,12 +3334,17 @@ func (c *client) pubAllowedFullCheck(subject string, fullCheck bool) bool {
...
 		// Update our cache here.
-		c.perms.pcache[string(subject)] = allowed
-		// Prune if needed.
-		if len(c.perms.pcache) > maxPermCacheSize {
+		c.perms.pcache.Store(string(subject), allowed)
+		// There is a case where we can invoke this from multiple go routines,
+		// (in deliverMsg() if sub.client is a LEAF), so we make sure to prune
+		// from only one go routine at a time.
+		if n := atomic.AddInt32(&c.perms.pcsz, 1); n > maxPermCacheSize &&
+			atomic.CompareAndSwapInt32(&c.perms.prun, 0, 1) {
```
matthiashanel, line 3347: I wonder if the compare and swap portion here should move into prunePubPermsCache, where we also clear the value.
```
if !atomic.CompareAndSwapInt32(&c.perms.prun, 0, 1) {
    return
}
```

matthiashanel (approved):
LGTM, comments are optional

server/client.go
kozlovic: Fixed.

server/client.go
kozlovic: Agreed.

kozlovic:
I have rebased from main branch to see if "ports" fix helps, so had to push force, but kept the recommended changes as different commit.

derekcollison (approved):
LGTM

matthiashanel (approved):
LGTM

PR #2139, opened 2021-04-21
Fix for consumer on restore being deleted

derekcollison:
New code dealing with concurrent consumer requests inadvertently broke tests that were restoring consumers with same config.

/cc @nats-io/core


ripienaar (approved):
LGTM

PR #2140, opened 2021-04-21
Make sure that ephemerals are cleaned up across gateways.

derekcollison:

/cc @nats-io/core

### DIFF
--- a/server/stream.go
+++ b/server/stream.go
@@ -3017,8 +3017,10 @@ func (mset *stream) state() StreamState {
 	return mset.stateWithDetail(false)
 }
 
-// Lock should be held
 func (mset *stream) numDirectConsumers() (num int) {
+	mset.mu.RLock()
+	defer mset.mu.RUnlock()
+
 	// Consumers that are direct are not recorded at the store level.
 	for _, o := range mset.consumers {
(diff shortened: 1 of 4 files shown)


server/stream.go
```diff
@@ -3029,6 +3042,7 @@ func (mset *stream) stateWithDetail(details bool) StreamState {
 	if !details {
 		state.Deleted = nil
 	}
+	state.Consumers += ndc
```
ripienaar, line 3045: so here Consumers counter will be lets say 10 but we wil not see those in consumer list still which would be confusing don't you think? Definitely welcome more visibility but maybe an additional key on state would be good (ugh), somthing for next release perhaps.

ripienaar (approved):
LGTM other than the one comment.

server/stream.go
derekcollison: Should I make it separate in State?

kozlovic (approved):
LGTM, but suspect that this will raise questions from user as to the count does not match what one gets from the list.

derekcollison:
Yes changing test and putting back old behavior in terms of stream state as we speak.

ripienaar (approved):
LGTM

PR #2142, opened 2021-04-21
Added implementation hints for header manipulation APIs to promote cross-client compatibility

aricart:
/cc @nats-io/core

### DIFF
--- a/doc/adr/0004-nats-headers.md
+++ b/doc/adr/0004-nats-headers.md
@@ -92,3 +92,52 @@ Implemented and merged to master.
 ## Consequences
 
 Use of headers is possible.
+
+## Compatibility Across NATS Clients
+
+The following is a list of features to insure compatibility across NATS clients that support headers.
+Because the feature in Go client and nats-server leverage the Go implementation as described above,
+the API used will determine how header names are serialized.
+
+The Go implementation of `http.Header` is a `map[string][]string` If the `Set`, `Get`, `Values`, `Del`
+functions are used, the specified key arguments will be converted into their canonical format. This means
+that the serialized Header name will _not_ be what the application specified. Go's documentation
+motivates direct map access if the client doesn't wish to have the header modified. This has the implication
+that non-canonical Headers are invisible to `http.Header`'s API, as the API performs a canonical
+key conversion prior to any operation.
+
+In order to promote compatibility across clients, this section describes how other language clients
+should behave:
+
+- `GET` returns a `string` of the first value found matching the specified key in a case-insensitive lookup.
+- `SET` removes all case-insensitive matching headers, with the specified value stored under the key 
+  specified without any manipulation.
+- `VALUES` returns a list of all values that case-insensitively match the specified key.
+- `DELETE` removes all headers that match case-insensitively the specified key.
+
+To prevent over-designing the API, it should be noted that if clients must have case-sensitive semantics,
+they should be able to iterate over the keys and values to enable client-side filtering of the values as
+necessary.
+
+#### Go Client
+
+Note that built-in Go client does not implement the above manipulations correctly. Users requiring
+case-preservation on a headers can easily implement these operations against the underlying map.
+
+### Multiple Header Values Serialization
+
+When serializing, entries that have more than one value should be serialized one per line. 
+While the http Header standard, prefers values to be a comma separated list, this introduces additional parsing
+requirements and ambiguity from client code. HTTP itself doesn't implement this requirement on headers such 
+as `Set-Cookie`. Libraries, such as Go, do not interpret comma-separated values as lists.
+
+
+
+
+
+
+
+
+
+
+


matthiashanel (approved):
LGTM

doc/adr/0004-nats-headers.md
```diff
@@ -92,3 +92,52 @@ Implemented and merged to master.
...
+- `DELETE` removes all headers that match case-insensitively the specified key.
+
+To prevent over-designing the API, it should be noted that if clients must have case-sensitive semantics,
+they should be able to iterate over the keys and values to enable client-side filtering of the values as
+necessary.
+
+#### Go Client
+
+Note that built-in Go client does not implement the above manipulations correctly. Users requiring
+case-preservation on a headers can easily implement these operations against the underlying map.
```
wallyqs, line 125: yes we can change this in the client since can move from types and internally still use http.Header

wallyqs (changes requested):
Overall ok, but we can fix some things before merging.

doc/adr/0004-nats-headers.md
aricart: @wallyqs this is simply the ADR, if this is OK, then we need an issue created for nats-server.

scottf: The Java client is released with `GET` returning a List of strings. Changing this would be a breaking change. Frankly this is less important than the serialization / deserialization. Also, in java `PUT` is used instead of `SET` and `REMOVE` is used instead of `DELETE` as these are the existing names in the java Map interface and makes sense to have our api match

doc/adr/0004-nats-headers.md
```diff
@@ -92,3 +92,52 @@ Implemented and merged to master.
...
+
+### Multiple Header Values Serialization
+
+When serializing, entries that have more than one value should be serialized one per line. 
+While the http Header standard, prefers values to be a comma separated list, this introduces additional parsing
+requirements and ambiguity from client code. HTTP itself doesn't implement this requirement on headers such 
+as `Set-Cookie`. Libraries, such as Go, do not interpret comma-separated values as lists.
+
```
scottf, line 133: Java already does multi line. There is an issue already created for .NET

doc/adr/0004-nats-headers.md
kozlovic: Same for the C client, which has already been publicly released. All my APIs (Set, Add, Get, Values, Keys, Delete) convert given key to canonical. I would have no choice but to either add new APIs (for case-insensitive) or some sort of library-level option that disable converting to canonical form.

wallyqs: I think it might be fine that it was already released since we are supposed to be treating keys as case-insensitive, the main area where this API addendum could help is in making sure that in client APIs when fetching the key it matches even if it is in canonical form or not (making Get case-insensitive).

wallyqs: Problem with case-insensitive Get though is that it would mean slower lookups in a map... one of the reasons why HTTP2 later on made headers lowercase by default instead since also easier to implement. (Go v1 which the http.Header API this ADR is basing from was done before HTTP2 was released).

scottf: I think the `SET` should be case sensitive. It we actually want to maintain user intention, then setting `FOO` is different than setting `foo`.

wallyqs: agree, would mean to insert the key as is and no other side effect

kozlovic: Well, but with implementation using maps, storing "FOO" and "foo" is different entries. So then Get() cannot be doing case-insensitive lookup because then what is being returned? Value of "FOO" or "foo"? So it would be more like those APIs would have to work with user-input as-is, but not doing case-insensitive insert/search, actually one could say that they are case-sensitive, in that they would respect the user's provided case. But then we are back to problems where basically user MUST know the case of the key that was added by another application..

wallyqs: So maybe we need to match exactly first (like JS headers which are case-sensitive) and that would be the fast path and preferred, then could be that if there is no hit then try case insensitive match

scottf: So we agree that we don't want to touch user data, so keys will have to be maintained case sensitive, i.e. FOO and foo. Since this has to be maintained anyway. I'm starting to think this is how it should be done.

1. keys are case sensitive, i.e. SET is case sensitive
2. default GET/VALUES is case sensitive
3. we should provide case insensitive GET/VALUES
4. remove is case sensitive

aricart: you are redesigning what we talked about.

scottf: My proposal requires that:

* We have to make the least decisions about what to do
* It maintains the data

By strictly not touching the data we support the charter as Colin summarized
> preserve headers provided by applications (namely casing) and allow both a case sensitive and case insensitive lookup

As I see it, 

* Standard data structures (hashtable/map/dictionary) are already case sensitive
* some clients may have to change how they serialize/deserialize, but this was already going to happen.
* clients will all have to code to support case-insensitive lookup, but that was expected
* It will be easier to change later than if we apply case-insensitive now

aricart:
After talking to Ivan I realized that I was not communicating correctly and the message was getting lost

Please review 
https://docs.google.com/presentation/d/1DDmvefaRVv-p0sX0mReTAh6UUsG9mKHfKf6vxnuKRKc/edit?usp=sharing

doc/adr/0004-nats-headers.md
ColinSullivan1: Thanks all for these updates.  I can appreciate Ivan's comment about case sensitive puts and case insensitive lookups.  I would think that for most cases (no pun intended), users will be aware of casing.  I'm primarily thinking of the http translation case where lookups are case insensitive and that'd be the path of least surprise.  We could default to case sensitive lookups and provide a case insensitive alternative.

scottf:
I'm convinced removing in a case-insensitive manner destroys data. If you look on page 6 of your presentation, your set example deletes 3 values for keys that do not match the key you are setting. This does not satisfy preserving data. I am however, for providing explicit case insensitive operations, i.e. 
* getInsensitive
* valuesInsensitive
* putInsensitive
* removeInsensitive 
 
I also am for an add that creates or updates in a case sensitive manner

ColinSullivan1:
After much discussion, we've decided to go with preserving case in the headers, defaulting to case sensitive operations, and later/optionally providing a case insensitive get/delete/etc for convenience (e.g. after a HTTP conversion).  API details about returning lists, etc, can be client specific and could be idiomatic to the language.  So long as there is a way to access all values in a client we're good in the ADR.  Primary concern for the ADR is around serialization / wire format imo.

aricart:
@ColinSullivan1 we need to nail the description of the ADR. the conversation here is adding additional proposals when what we are looking for is consensus on the document. I will close and open a new one.

PR #2143, opened 2021-04-21
Make sure to return no consumer error if we are still waiting on a leader.

derekcollison:

/cc @nats-io/core

### DIFF
--- a/server/jetstream_api.go
+++ b/server/jetstream_api.go
@@ -3324,9 +3324,8 @@ func (s *Server) jsConsumerInfoRequest(sub *subscription, c *client, subject, re
 			if js.isLeaderless() {
 				resp.Error = jsClusterNotAvailErr
 				s.sendAPIErrResponse(ci, acc, subject, reply, string(msg), s.jsonResponse(&resp))
-			}
-			// Check here if we are a member and this is just a new consumer that does not have a leader yet.
-			if rg := ca.Group; rg.node != nil && rg.isMember(cc.meta.ID()) {
+			} else if rg := ca.Group; rg != nil && rg.node != nil && rg.isMember(cc.meta.ID()) {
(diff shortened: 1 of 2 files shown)


server/jetstream_api.go
```diff
@@ -3320,6 +3325,13 @@ func (s *Server) jsConsumerInfoRequest(sub *subscription, c *client, subject, re
 				resp.Error = jsClusterNotAvailErr
 				s.sendAPIErrResponse(ci, acc, subject, reply, string(msg), s.jsonResponse(&resp))
 			}
+			// Check here if we are a member and this is just a new consumer that does not have a leader yet.
+			if rg := ca.Group; rg.node != nil && rg.isMember(cc.meta.ID()) {
```
kozlovic, line 3329: Is `rg` guaranteed to not be nil? I see that we check for `rg.node != nil` but not sure if `rg` itself can be nil.

server/jetstream_api.go
```diff
@@ -3320,6 +3325,13 @@ func (s *Server) jsConsumerInfoRequest(sub *subscription, c *client, subject, re
 				resp.Error = jsClusterNotAvailErr
 				s.sendAPIErrResponse(ci, acc, subject, reply, string(msg), s.jsonResponse(&resp))
 			}
```
kozlovic, line 3327: Do we need a `return` after sending the error in the `if js.isLeaderLess()` case?

server/jetstream_api.go
derekcollison: Should not be but will add a check for saftey.

derekcollison: Will reflow.

server/jetstream_api.go
derekcollison: Will reflow this part.

kozlovic (approved):
LGTM

PR #2145, opened 2021-04-21
Check for overlapping subjects on stream update.

### LINKED ISSUE #2144: stream subjects overlap is not checked on stream edit/update
Streams are not allowed to listen to the same subjects as other streams. When trying to add a stream that listens to the same subjects of an existing stream the error `subjects overlap with an existing stream` is produced.

This check does not occur when editing a streams subject.

To reproduce:

Add a stream that listens to a subject.
Add a second stream that listens to a different subject.
Edit the second stream to listen to the same subject as the first stream.

This is confirmed to be the case using nats.go UpdateStream() as well ass the natscli nats stream edit.

derekcollison:
Resolves #2144 

/cc @nats-io/core

### DIFF
--- a/server/jetstream_cluster.go
+++ b/server/jetstream_cluster.go
@@ -3466,6 +3466,22 @@ func (s *Server) jsClusteredStreamUpdateRequest(ci *ClientInfo, acc *Account, su
 		return
 	}
 
+	// Check for subject collisions here.
+	for _, sa := range cc.streams[acc.Name] {
+		if sa == osa {
+			continue
+		}
+		for _, subj := range sa.Config.Subjects {
+			for _, tsubj := range newCfg.Subjects {
+				if SubjectsCollide(tsubj, subj) {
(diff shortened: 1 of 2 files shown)


server/jetstream_cluster.go
```diff
@@ -3466,6 +3466,22 @@ func (s *Server) jsClusteredStreamUpdateRequest(ci *ClientInfo, acc *Account, su
 		return
 	}
 
+	// Check for subject collisions here.
+	for _, sa := range cc.streams[acc.Name] {
+		if sa == osa {
```
kozlovic, line 3471: This same loop (without this check) exists in jsClusteredStreamRequest(). Could have been made a little function, although this check is specific to jsClusteredStreamUpdateRequest() I presume.

kozlovic (approved):
LGTM

server/jetstream_cluster.go
derekcollison: Agree, we could make a function but this one does need to skip the original assignment unlike create.

PR #2146, opened 2021-04-21
Make sure to not have the raft layer block on apply channel on exit.

derekcollison:
In the case when raft layer can run for a bit as we exit the monitorStream that pulls from the apply channel make sure we drain that to prevent a lockup at the raft layer.

/cc @nats-io/core


kozlovic (approved):
LGTM

PR #2148, opened 2021-04-21
Delaying stream/consumer error response when there is no contact to meta

matthiashanel:
this gives a stream/consumer leader time to respond first

I went with 500 ms as I still saw the issue with 250. (the leader is under full load)
Since the group is not available in all cases, I do not use that channel.  (don't think responding would be a big problem)
When we address responding by non leader via a queue, I think this code will change

### DIFF
--- a/server/jetstream_api.go
+++ b/server/jetstream_api.go
@@ -776,6 +776,28 @@ func (s *Server) sendAPIErrResponse(ci *ClientInfo, acc *Account, subject, reply
 	s.sendJetStreamAPIAuditAdvisory(ci, acc, subject, request, response)
 }
 
+const errRespDelay = 500 * time.Millisecond
+
+func (s *Server) sendDelayedAPIErrResponse(ci *ClientInfo, acc *Account, subject, reply, request, response string, rg *raftGroup) {
+	var quitCh <-chan struct{}
+	if rg != nil && rg.node != nil {
+		quitCh = rg.node.QuitC()
+	}
+	s.startGoRoutine(func() {
+		defer s.grWG.Done()
+		select {
+		case <-quitCh:
+		case <-s.quitCh:
+		case <-time.After(errRespDelay):
+			acc.trackAPIErr()
+			if reply != _EMPTY_ {
+				s.sendInternalAccountMsg(nil, reply, response)
+			}
+			s.sendJetStreamAPIAuditAdvisory(ci, acc, subject, request, response)
+		}
+	})
+}
+
 func (s *Server) getRequestInfo(c *client, raw []byte) (pci *ClientInfo, acc *Account, hdr, msg []byte, err error) {
 	hdr, msg = c.msgParts(raw)
 	var ci ClientInfo
@@ -1614,7 +1636,8 @@ func (s *Server) jsStreamInfoRequest(sub *subscription, c *client, subject, repl
 		} else if sa == nil {
 			if js.isLeaderless() {
 				resp.Error = jsClusterNotAvailErr
-				s.sendAPIErrResponse(ci, acc, subject, reply, string(msg), s.jsonResponse(&resp))
+				// Delaying an error response gives the leader a chance to respond before us
+				s.sendDelayedAPIErrResponse(ci, acc, subject, reply, string(msg), s.jsonResponse(&resp), nil)
 			}
 			return
 		}
@@ -1626,7 +1649,8 @@ func (s *Server) jsStreamInfoRequest(sub *subscription, c *client, subject, repl
 		if !acc.JetStreamIsStreamLeader(streamName) && !isLeaderless {
 			if js.isLeaderless() {
 				resp.Error = jsClusterNotAvailErr
-				s.sendAPIErrResponse(ci, acc, subject, reply, string(msg), s.jsonResponse(&resp))
+				// Delaying an error response gives the leader a chance to respond before us
+				s.sendDelayedAPIErrResponse(ci, acc, subject, reply, string(msg), s.jsonResponse(&resp), sa.Group)
 			}
 			return
 		}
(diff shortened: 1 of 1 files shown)


server/jetstream_api.go
```diff
@@ -776,6 +776,23 @@ func (s *Server) sendAPIErrResponse(ci *ClientInfo, acc *Account, subject, reply
...
 }
 
+const errRespDelay = 500 * time.Millisecond
+
+func (s *Server) sendDelayedAPIErrResponse(ci *ClientInfo, acc *Account, subject, reply, request, response string) {
+	s.startGoRoutine(func() {
+		defer s.grWG.Done()
+		select {
+		case <-s.quitCh:
+		case <-time.After(errRespDelay):
```
derekcollison, line 786: Only issue here is we do not bail early if asset goes away like stream or consumer.

server/jetstream_api.go
```diff
@@ -1614,7 +1631,9 @@ func (s *Server) jsStreamInfoRequest(sub *subscription, c *client, subject, repl
 		} else if sa == nil {
 			if js.isLeaderless() {
 				resp.Error = jsClusterNotAvailErr
-				s.sendAPIErrResponse(ci, acc, subject, reply, string(msg), s.jsonResponse(&resp))
+				// Delaying an error response gives the leader a chance to respond before us
+				strMsg := string(msg)
+				s.sendDelayedAPIErrResponse(ci, acc, subject, reply, strMsg, s.jsonResponse(&resp))
```
derekcollison, line 1636: Just inline string cast I think.

server/jetstream_api.go
```diff
@@ -1626,7 +1645,9 @@ func (s *Server) jsStreamInfoRequest(sub *subscription, c *client, subject, repl
 		if !acc.JetStreamIsStreamLeader(streamName) && !isLeaderless {
 			if js.isLeaderless() {
 				resp.Error = jsClusterNotAvailErr
-				s.sendAPIErrResponse(ci, acc, subject, reply, string(msg), s.jsonResponse(&resp))
+				// Delaying an error response gives the leader a chance to respond before us
+				strMsg := string(msg)
```
derekcollison, line 1649: Same

derekcollison (commented):
In general looks good, should we try to detect when the asset goes away? Pass in an asset quit chan or nil if not known etc?

In this case the stream and consumer both have quitCh IIRC.

server/jetstream_api.go
```diff
@@ -1614,7 +1631,9 @@ func (s *Server) jsStreamInfoRequest(sub *subscription, c *client, subject, repl
 		} else if sa == nil {
 			if js.isLeaderless() {
 				resp.Error = jsClusterNotAvailErr
-				s.sendAPIErrResponse(ci, acc, subject, reply, string(msg), s.jsonResponse(&resp))
+				// Delaying an error response gives the leader a chance to respond before us
+				strMsg := string(msg)
```
kozlovic, line 1635: You did not have to do this, could pass string(msg) directly to sendDelayedAPIErrResponse() since that function is not itself invoked in a different go routine, it only invoke something in a go routine.

matthiashanel: fixed

server/jetstream_api.go
matthiashanel: fixed

server/jetstream_api.go
matthiashanel: fixed

server/jetstream_api.go
matthiashanel: fixed, passing in raft group, will use `rg.node.QuitC()`

server/jetstream_api.go
```diff
@@ -3323,12 +3348,14 @@ func (s *Server) jsConsumerInfoRequest(sub *subscription, c *client, subject, re
...
 				resp.Error = jsClusterNotAvailErr
-				s.sendAPIErrResponse(ci, acc, subject, reply, string(msg), s.jsonResponse(&resp))
+				// Delaying an error response gives the leader a chance to respond before us
+				s.sendDelayedAPIErrResponse(ci, acc, subject, reply, string(msg), s.jsonResponse(&resp), ca.Group)
 			} else if rg := ca.Group; rg != nil && rg.node != nil && rg.isMember(cc.meta.ID()) {
 				// Check here if we are a member and this is just a new consumer that does not have a leader yet.
 				if rg.node.GroupLeader() == _EMPTY_ && !rg.node.HadPreviousLeader() {
 					resp.Error = jsNoConsumerErr
-					s.sendAPIErrResponse(ci, acc, subject, reply, string(msg), s.jsonResponse(&resp))
+					strMsg := string(msg)
```
kozlovic, line 3357: Missed one here and you are calling sendAPIErrResponse, not the delayed one..

server/jetstream_api.go
```diff
@@ -776,6 +776,28 @@ func (s *Server) sendAPIErrResponse(ci *ClientInfo, acc *Account, subject, reply
 	s.sendJetStreamAPIAuditAdvisory(ci, acc, subject, request, response)
 }
 
+const errRespDelay = 500 * time.Millisecond
+
+func (s *Server) sendDelayedAPIErrResponse(ci *ClientInfo, acc *Account, subject, reply, request, response string, rg *raftGroup) {
+	var quitCh <-chan struct{}
+	if rg != nil {
```
kozlovic, line 783: I think I see often `rg.node != nil` check, so would add that too...

matthiashanel: added

server/jetstream_api.go
matthiashanel: fixed

server/jetstream_api.go
derekcollison: Asset may not have a raft group, could be R=1.

derekcollison: We need the stream and consumer quit chan I think. Or we punt with comment for next release to add in asset tracking for this.

matthiashanel: wouldn't that channel go away with the stream too?
at this point we do't really have the consumer/stream yet.
I'm not worried about responding for smth. that went away while we delayed. 
I am also convinced that this will see a revamp once we have only one disconnected follower respond.

derekcollison (approved):
LGTM

kozlovic (approved):
LGTM

PR #2149, opened 2021-04-22
Do not persist or snapshot consumer state after a restore.

derekcollison:
This can lead to a data race and is not needed after being applied.

/cc @nats-io/core


wallyqs (approved):

kozlovic (approved):
LGTM

PR #2156, opened 2021-04-22
[CHANGED] Randomize Leafnode remote URLs and add option to disable

### LINKED ISSUE #2153: NATS LeafNode remote URLs : Config to shuffle, on by default
## Feature Request
Please allow the shuffling of leaf node remote URLs.

#### Use Case:
All leaf nodes start with the same config, the same ordered list of leaf remote URLs. This causes the first server to pick up all leafs by default. Unlike other client connections, which have the option to disable node shuffling, this is on by default.

#### Proposed Change:
Add a config for LeafRemotes "noShuffle" mode.
Add a function to randomize or shuffle the pool of remote URLs.

All nats cluster operators with a multitude of leaf-node connections.

variadico:
Currently, leafnode remote URLs are kept in order when parsed. This change randomizes the order of URLs by default and adds an option to disable this behavior, returning you to the way it was before.

```
leafnodes {
  remotes = [
    {
      dont_randomize: true
      url: [
        "nats-leaf://host1:7422"
        "nats-leaf://host2:7422"
        "nats-leaf://host3:7422"
        "nats-leaf://host4:7422"
      ]
      account: fizzbuzz
      credentials: "./fizzbuzz.creds"
    }
  ]
}
```

Resolves #2153

/cc @nats-io/core

### DIFF
--- a/server/opts.go
+++ b/server/opts.go
@@ -139,6 +139,7 @@ type LeafNodeOpts struct {
 // RemoteLeafOpts are options for connecting to a remote server as a leaf node.
 type RemoteLeafOpts struct {
 	LocalAccount string      `json:"local_account,omitempty"`
+	NoRandomize  bool        `json:"-"`
 	URLs         []*url.URL  `json:"urls,omitempty"`
 	Credentials  string      `json:"-"`
 	TLS          bool        `json:"-"`
@@ -1761,6 +1762,8 @@ func parseRemoteLeafNodes(v interface{}, errors *[]error, warnings *[]error) ([]
 		for k, v := range rm {
 			tk, v = unwrapValue(v, &lt)
 			switch strings.ToLower(k) {
+			case "no_randomize", "dont_randomize":
+				remote.NoRandomize = v.(bool)
 			case "url", "urls":
 				switch v := v.(type) {
 				case []interface{}, []string:
--- a/server/opts_test.go
+++ b/server/opts_test.go
@@ -16,6 +16,7 @@ package server
 import (
 	"bytes"
 	"crypto/tls"
+	"encoding/json"
 	"flag"
 	"fmt"
 	"io/ioutil"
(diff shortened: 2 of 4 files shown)


server/opts.go
```diff
@@ -1842,6 +1846,11 @@ func parseRemoteLeafNodes(v interface{}, errors *[]error, warnings *[]error) ([]
 				}
 			}
 		}
+		if remote.ShuffleURLs {
+			rand.Shuffle(len(remote.URLs), func(i, j int) {
```
kozlovic, line 1850: I would not do the shuffling in parsing of options (which would make your test to have to be updated). The reason is that users embedding the server would not benefit from that if they configure the server directly with options.
I recommend doing checks/validations in code that is done in NewServer() which then means it applies to both config and direct options. For leafnode, this would be `validateLeafNode()`.

server/opts_test.go
```diff
@@ -2383,6 +2384,127 @@ func TestParsingLeafNodeRemotes(t *testing.T) {
 			t.Fatalf("Expected %v, got %v", expected, opts.LeafNode.Remotes[0])
 		}
 	})
+
+	t.Run("shuffle remote urls no effect", func(t *testing.T) {
```
kozlovic, line 2388: Unfortunately, you would have to change those tests if you move the shuffling. You would even have to add test with direct options (so 4 tests total): 2 with ProcessConfigFile() followed by NewServer() and then check remotes' urls array, and 2 with creation Options and setting options directly and then calling NewServer() and checking remotes.
Let me know if you need more details on that.

kozlovic (changes requested):

server/opts.go
variadico: Aaah, ok. Great thanks, will take another look.

matthiashanel:
@variadico @kozlovic , wondering if we should enable this by default and provide an option to turn it off. In part because this is the right thing to do, Why make it harder by having to enable the better behavior.

ripienaar:
Agree. Should be default.

kozlovic:
Yes, sorry I missed that too. Since I saw the discussion on Slack and issue reported by Don R., I assumed that this was the direction we would take. @variadico The flag would have to be a negative, like "NoSuffle" or something like that.

variadico:
All right. Got some updates.

* Moved `rand.Shuffle` to `validateLeafNode`
* Made `rand.Shuffle` the default behavior
* Changed the name "shuffle" to "random" to match the term the NATS client uses (any thoughts on this?)
* Updated the parsing code
* Added new tests

kozlovic (approved):
LGTM!

PR #2157, opened 2021-04-23
fix unit test crashing when leader was nil caused by quorum reformation

matthiashanel:

Locally had a crash in leader.JetStreamClusterPeers() where the first access to `s.mu` would crash.
Probably caused by leadership change between awaiting a leader and retrieving it.
Not worried about waiting forever as waitOnLeader will fail if there's none in time.

### DIFF
--- a/server/jetstream_cluster_test.go
+++ b/server/jetstream_cluster_test.go
@@ -6456,6 +6456,10 @@ func (c *cluster) waitOnPeerCount(n int) {
 	c.t.Helper()
 	c.waitOnLeader()
 	leader := c.leader()
+	for leader == nil {
+		c.waitOnLeader()
+		leader = c.leader()
+	}
 	expires := time.Now().Add(10 * time.Second)
 	for time.Now().Before(expires) {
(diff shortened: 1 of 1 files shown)


kozlovic (approved):
LGTM

server/jetstream_cluster_test.go
```diff
@@ -6456,6 +6456,10 @@ func (c *cluster) waitOnPeerCount(n int) {
 	c.t.Helper()
 	c.waitOnLeader()
 	leader := c.leader()
+	for leader == nil {
+		c.waitOnLeader()
+		leader = c.leader()
```
derekcollison, line 6461: Just pushes the problem, same thing can happen after this line.

kozlovic: That was my first though but look, it is a "for" loop. And as he explains in description, waitOnLeader() will Fatal() the test if there is no leader.

derekcollison: A gotcha, I skipped over that part. Apologies.

PR #2158, opened 2021-04-23
Alternate headers redux proposal

scottf:
In addition, we might considering changing the version number.

/cc @aricart @wallyqs @ColinSullivan1 @kozlovic

### DIFF
--- a/doc/adr/0004-nats-headers.md
+++ b/doc/adr/0004-nats-headers.md
@@ -1,6 +1,6 @@
 # 4. nats-headers
 
-Date: 2020-05-28
+Date: 2021-04-23
 
 ## Status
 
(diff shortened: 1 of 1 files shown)


doc/adr/0004-nats-headers.md
```diff
@@ -96,41 +85,22 @@ Use of headers is possible.
...
-#### Go Client
-
-Note that built-in Go client does not implement the above manipulations correctly. Users requiring
-case-preservation on a headers can easily implement these operations against the underlying map.
+- `GET` can either return a `string` of the first value found matching the specified key in a case-sensitive lookup, or it can return a list of `string`. This may vary depending on the client.
+- `VALUES` returns a list of all `string` values that case-sensitively match the specified key. This will not be present in clients where the `GET` functions like `VALUES`
+- `SET` removes all values stored under the case-sensitive key and replaces it with the new value. Some clients may refer to this a `PUT`
+- `ADD` If the case-sensitive key already has associated values, the new value is added to the list, otherwise the key is created with the new value.
+- `DELETE` removes all headers that match the specified case-sensitive key. Some clients may refer to this as `REMOVE`
+- Clients will provide case-insensitve versions of lookups such as `GET` and `VALUE`
```
wallyqs, line 96: I think what could work is for `Get` to first match case-sensitive so that it becomes the fast path without relying on a normalization into canonical (http v1 style) or lowercase (http v2 style), that would help us with processing of JS headers where the case is already determined, but then if there is no match then do the case-insensitive check against the keys in the map?

scottf: So I would ask why would we want to invent new computing behavior here when case paradigm is already baked into our languages and developers are used to it.

Seems to me that in support of case sensitive, is that fact that internally we specify an exact case for our internal header keys. In the corporate world, they tend to setup constants and organizational rules. http1 / 2 actually specify case sensitive keys with case-insensitive lookup, that's why I want to ensure clients provide the secondary ability to read case-insensitive

wallyqs: You are right, for example things like `Sec-WebSocket-X` headers in HTTP actually have a correct way to be set so that is why in Go the `Set/Add` APIs aren't used for some things, otherwise it would become `Sec-Websocket-X` which can be incompatible in some proxies, for best compatibility instead it is set with the exact case via:

```go
req.Header["Sec-WebSocket-Key"]
```

The Go HTTP client does emit keys as they are set exactly when making a request as well to be flexible in that way without normalizing:

```go
	req, err := http.NewRequest("GET", "http://localhost:3000", nil)
	req.Header = http.Header{
		"foo": []string{"A"},
		"Foo": []string{"B"},
		"FOO": []string{"C"},
	}
	// RESULT:
	// User-Agent: Go-http-client/1.1
	// FOO: C
	// Foo: B
	// foo: A
```

The NATS Server currently does not normalize the keys either and treats them opaquely, so it is data-preserving already in that sense as well (and for JS headers that means that for using them correctly need to have the exact case (e.g. to be right it has to be exactly `Nats-Msg-Id`).  Maybe rather than defining the exact Get/Set/Del API for all clients what is important is that there is a way for users to get the keys in case-sensitive and case-insensitive ways as needed and document how to do it.

scottf: > Maybe rather than defining the exact Get/Set/Del API for all clients what is important is that there is a way for users to get the keys in case-sensitive and case-insensitive ways as needed and document how to do it.
+1 I'm good with that

scottf:
I have implemented Java according to this. https://github.com/nats-io/nats.java/pull/451/files

doc/adr/0004-nats-headers.md
```diff
@@ -1,41 +1,30 @@
...
 
 ## Context
 
-This document describes NATS Headers from the perspective of clients. NATS headers allow clients to specify additional meta-data in the form of headers. The headers are effectively [HTTP Headers](https://tools.ietf.org/html/rfc7230#section-3.2). The reference Go implementation internally uses `http.Header` for their serialization and parsing.
+This document describes NATS Headers from the perspective of clients. NATS headers allow clients to specify additional meta-data in the form of headers. 
 
 The salient points of the HTTP header specification are:
 
-- Each header field consists of a case-insensitive field name followed by a colon (`:`), optional leading whitespace, the field value, and optional trailing whitespace.
+- Each header entry consists of a field name followed by a colon (`:`), optional leading space (SP), the field value, optional trailing whitespace and then carriage-return/line-feed (CR/LF). It is suggested to not use leading or trailing spaces simply to reduce the number of bytes transmitted.
```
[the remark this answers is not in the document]
scottf: this was fixed in the server b/c the java client writes without the space

scottf:
Superseded by https://github.com/nats-io/nats-server/pull/2177/

PR #2159, opened 2021-04-23
[added] leaf deny exports/imports to varz monitoring

matthiashanel:

Would have wanted that today

### DIFF
--- a/server/monitor.go
+++ b/server/monitor.go
@@ -1101,13 +1101,18 @@ type LeafNodeOptsVarz struct {
 	Remotes     []RemoteLeafOptsVarz `json:"remotes,omitempty"`
 }
 
+// Contains lists of subjects not allowed to be imported/exported through a leaf node connection
+type RemoteLeafOptDeny struct {
+	Exports []string `json:"exports,omitempty"`
+	Imports []string `json:"imports,omitempty"`
+}
+
 // RemoteLeafOptsVarz contains monitoring remote leaf node information
 type RemoteLeafOptsVarz struct {
-	LocalAccount string   `json:"local_account,omitempty"`
-	TLSTimeout   float64  `json:"tls_timeout,omitempty"`
-	URLs         []string `json:"urls,omitempty"`
-	DenyExports  []string `json:"deny_exports,omitempty"`
-	DenyImports  []string `json:"deny_imports,omitempty"`
+	LocalAccount string             `json:"local_account,omitempty"`
+	TLSTimeout   float64            `json:"tls_timeout,omitempty"`
+	URLs         []string           `json:"urls,omitempty"`
+	Deny         *RemoteLeafOptDeny `json:"deny,omitempty"`
 }
 
 // VarzOptions are the options passed to Varz().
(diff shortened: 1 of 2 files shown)


server/monitor.go
```diff
@@ -1106,6 +1106,8 @@ type RemoteLeafOptsVarz struct {
 	LocalAccount string   `json:"local_account,omitempty"`
 	TLSTimeout   float64  `json:"tls_timeout,omitempty"`
 	URLs         []string `json:"urls,omitempty"`
+	DenyExports  []string `json:"deny_exports,omitempty"`
```
derekcollison, line 1109: Should we say Deny property that has two sub properties, Exports and Imports?
Otherwise LGTM.

kozlovic (approved):
LGTM. It's true that we could make a single field that contains Exports/Imports, but I understand that Matthias probably copied from RemoteLeafOpts (not the Varz version) that contains the 2 distinct slices.
Up to you Matthias if you want to add a new struct.

server/monitor.go
matthiashanel: wrapped them in a dedicated struct

matthiashanel:
Sample output with one account having `$JS.API.>` denied for export/import

```
curl localhost:8080/varz | grep '"leaf"' -A 25
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  3615    0  3615    0     0   882k      0 --:--:-- --:--:-- --:--:--  882k
  "leaf": {
    "remotes": [
      {
        "local_account": "ABDGQHM46B4NBP4YIGKBDASU53TZHWPUVI33H5O3RQUPIGZ2QAFGJQKO",
        "urls": [
          "localhost:4224",
          "localhost:4234"
        ]
      },
      {
        "local_account": "ADXKBBHIU3HAY3VXAEETH3XSKQSFJ3YUH7NOBHGS5BN3K4SKNUNE7TUV",
        "urls": [
          "localhost:4224",
          "localhost:4234"
        ],
        "deny": {
          "exports": [
            "$JS.API.\u003e"
          ],
          "imports": [
            "$JS.API.\u003e"
          ]
        }
      }
    ]
  },
```

server/monitor.go
```diff
@@ -1101,13 +1101,18 @@ type LeafNodeOptsVarz struct {
 	Remotes     []RemoteLeafOptsVarz `json:"remotes,omitempty"`
 }
 
+// Contains lists of subjects not allowed to be imported/exported through a leaf node connection
+type RemoteLeafOptDeny struct {
```
kozlovic, line 1105: I was wondering if we should make this very generic, so that if in the future we need an exports/imports []string, we can use in different places. That being said, I have no idea how I would name this new struct.
Also, there is a way to do type aliases...

kozlovic (approved):
LGTM. See my comment about this new structure name: should it be more generic or not?

server/monitor.go
matthiashanel: how does DenyRules sound?

kozlovic: Sounds good to me.

matthiashanel: renamed

kozlovic (approved):
LGTM

PR #2164, opened 2021-04-29
[FIXED] MQTT: asset placement in origin cluster

kozlovic:
In a setup with shared system account and a cluster of leaf nodes,
the JS requests did not contain the origin cluster, which caused
assets to possibly be created in the HUB. With this change, the
assets will be created in the origin cluster.

Also, removed use of acc.JetStreamEnabled() but instead fail
start of the server if mqtt is enabled in standalone mode and JS
is not enabled. If JS is enabled, we will get proper error if
account has no JS enabled.

### DIFF
--- a/server/dirstore_test.go
+++ b/server/dirstore_test.go
@@ -44,7 +44,7 @@ func require_False(t *testing.T, b bool) {
 	}
 }
 
-func require_NoError(t *testing.T, err error) {
+func require_NoError(t testing.TB, err error) {
 	t.Helper()
 	if err != nil {
 		t.Fatalf("require no error, but got: %v", err)
--- a/server/jetstream_cluster_test.go
+++ b/server/jetstream_cluster_test.go
@@ -6333,6 +6333,8 @@ var jsClusterTemplWithLeafNode = `
 		routes = [%s]
 	}
 
+	{{mqtt}}
+
 	# For access to system account.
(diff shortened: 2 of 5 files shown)


matthiashanel (approved):
LGTM

server/dirstore_test.go
```diff
@@ -44,7 +44,7 @@ func require_False(t *testing.T, b bool) {
 	}
 }
 
-func require_NoError(t *testing.T, err error) {
+func require_NoError(t testing.TB, err error) {
```
derekcollison, line 47: What is this? Just curious.

derekcollison (approved):
LGTM

server/dirstore_test.go
kozlovic: Because I use createDir() in testMQTTRunServer() which I have with testing.TB because I also use it for benchmarks that have *testing.B. Using the interface testing.TB allows an helper to be used by both test and bench.

derekcollison: That new?

kozlovic: testing.TB interface? Has been there quite a while, but after we copied some helpers to work with both tests and bench. So moving forward, technically we should use testing.TB for helpers (at least the ones that are not embedded), but I think that is fine to still use *testing.T but then change if the needs come to use a given helper from a bench test.

PR #2165, opened 2021-04-29
allow custom authenticators to set a username

ripienaar:
A custom authenticator can call `RegisterUser`, but the Username it sets on the 
User struct is never used, setting it here will allow a custom authenticator to declare
a username that will be seen in advisories, logs etc.

I am not sure if I should also do same for password - could not see any use case
for that but I am not super familiar with this code.

/cc @nats-io/core

### DIFF
--- a/server/client.go
+++ b/server/client.go
@@ -793,10 +793,17 @@ func (c *client) RegisterUser(user *User) {
 	} else {
 		c.setPermissions(user.Permissions)
 	}
+
+	// allows custom authenticators to set a username to be reported in
+	// server events and more
+	if user.Username != _EMPTY_ {
+		c.opts.Username = user.Username
+	}
+
 	c.mu.Unlock()
 }
 
-// RegisterNkey allows auth to call back into a new nkey
+// RegisterNkeyUser allows auth to call back into a new nkey
 // client with the authenticated user. This is used to map
 // any permissions into the client and setup accounts.
 func (c *client) RegisterNkeyUser(user *NkeyUser) error {


server/client.go
```diff
@@ -793,10 +793,17 @@ func (c *client) RegisterUser(user *User) {
 	} else {
 		c.setPermissions(user.Permissions)
 	}
+
+	// allows custom authenticators to set a username to be reported in
+	// server events and more
+	if user.Username != "" {
```
kozlovic, line 799: nit: For continuity, we tend to use `_EMPTY_` constant as opposed to `""`.

So I had a closer look since I am not familiar with the use of custom authentication. The custom authenticator is invoked while processing the CONNECT protocol, so at this point, c.opts.Username (and Password) is set based on what is present in the CONNECT. The issue you are describing is that if the custom auth Check() code is calling RegisterUser() with a user made User record, then the code here would not overwrite c.opts.Username value that we got from CONNECT with whatever value the user is passing to RegisterUser's User object.

So I think this is fine and would not break regular auth where we invoke RegisterUser() from auth.go, and agree that there may be no need to set the password since in all places we report user info, password is actually redacted out.

kozlovic (approved):
LGTM, little nit about use of `""` as opposed to `_EMPTY_`.

server/client.go
ripienaar: Ah yes. Will fix empty

yes for some background if my use case - in my case I send my own JWT in and will do my own verification and auth against a trustee certificate. 

I extract identity and role information from then JWT and construct identity based private inbox paths so no user can view another users inboxes since every dynamically created user has private subject spaces. I also ensure that connections of one role can only every access subjects that role require based on my own JWT. 

Finally for advisories and logs I want to them tag those connections with the username found in my jwt. Hence setting it there

ripienaar:
Fixed the one nit, so will merge, thank you

PR #2166, opened 2021-04-29
[FIXED] JetStream pull based message delivery could drop responses.

derekcollison:
GW Routed replies by default pin the map to the destination client connection. For internal callbacks, especially ones that use different internal clients for receiving versus sending the response we would not properly track these routed replies. If proper interest arrived after the request it could be lost.

/cc @nats-io/core


kozlovic (approved):
LGTM

PR #2167, opened 2021-04-29
[FIXED] TLS: default ciphers not set when tls enabled from command line

kozlovic:
If running the server with command lines:
```
nats-server --tlsverify --tlscert "cert.pem" --tlskey "key.pem"
```
the default ciphers would not be set, however, they would using this
equivalent config:
```
tls: {
   verify: true
   cert_file: "cert.pem"
   key_file: "key.pem"
}
```

Reported by @DavidSimner


derekcollison (approved):
LGTM

Issue #1811, opened 2021-01-15
Improve API Error Response Structure for Better Programmatic Error Handling


jon-whit:
## Feature Request
The Jetstream API errors do not return a unique error code per path in the current design. This makes it difficult to programmatically react to specific Jetstream errors. I'd like to see the Jetstream API introduce unique error codes per specific error returned.

#### Use Case:
As an example, when I want to create a consumer and the consumer I am creating references a stream that does not exist, I'd like to be able to check the Jetstream API error against a specific uuid code that identifies the fact that the `"stream does not exist"`. Right now with the HTTP code approach, if the HTTP code is overloaded by multiple return paths, then my application cannot really take the appropriate action (unless it parses the error description in some way).

#### Proposed Change:
It would be nice to change the API error structure to something like:

```
{
    "code": "<uuid>",
    "cause": {...}, // underlying error (the error raised by the underlying application; whose fields are subject to change)
    "description": "<description>"
}
```

#### Who Benefits From The Change(s)?
Me, and other developers trying to programmatically react to certain error cases when trying to interact with the Jetstream API. 

For more information, @ripienaar and I had a long conversation about this on [this](https://natsio.slack.com/archives/CM3T6T7JQ/p1609872359263800) slack thread.

ripienaar:
So to summary the problem, toda we have for example `jsNotFoundErrorI()` that makes a generic 404 and you dont know if it means stream not found, consumer not found, template not found or what.

So I am not sure if we should do a huge breaking change now on this one.

Today we have 

```
type ApiError struct {
	Code        int    `json:"code"`
	Description string `json:"description,omitempty"`
}
```

We can add a field here `Kind string` and put in there something like `stream_not_found`, `consumer_not_found` etc, we promise we wont change these strings - effectively they are subject to SemVer - but the descriptions as they are today are not part of any SemVer promises.

This - adding 1 field and not changing any existing fields - doesnt break anything but gets you what you want I think?

derekcollison:
The enclosing type tells you what the 404 pertains to IIRC. Meaning stream or consumer etc. The enclosing advisory has that type IIRC.

ripienaar:
> The enclosing type tells you what the 404 pertains to IIRC. Meaning stream or consumer etc. The enclosing advisory has that type IIRC.

Unfortunately only true in broad strokes, does a 404 from consumer info mean stream or consumer is missing? Could be either.

[...]

PR #2168, opened 2021-04-30
richer api errors proposal

ripienaar:
This creates richer errors in response to https://github.com/nats-io/nats-server/issues/1811 and numerous other requests.

The main idea is that when a 404 is raised in response to accessing a consumer we now include a unique identifier that would allow callers to distinguish between stream not found, consumer not found, no message found and other errors that might raise 404.

One could argue that users might parse the Description like `stream not  found` however this makes our human friendly descriptions part of the API promises and we can never improve, bug fix or translate those strings. For example today there is an unfortunate mix of `jetstream` and `JetStream` in use in error descriptions and we would want the ability to improve errors without breaking peoples code.

Additionally I propose capturing help and urls in the code - not in actual API responses - to facilitate a single place where we maintain these errors and then adding tooling like `nats error JS1` that will produce the error text, help, additional information and links to documentation that would help users resolve those errors.  So the slow `ApiErrorByKind` would not be used in anything but the CLI to resolve these things, though even thats optional as the `ApiErrors` is exported.

/cc @nats-io/core


server/jetstream_api.go
```diff
@@ -608,6 +636,17 @@ var (
 	jsNoAccountErr         = &ApiError{Code: 503, Description: "account not found"}
 )
 
+// ApiErrorByKind used by the cli to provide `nats error JS01`
+func ApiErrorByKind(kind string) *ApiError {
+	for _, v := range ApiErrors {
+		if v.Kind == kind {
+			return v
```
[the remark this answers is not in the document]
ripienaar: They are unique yes. 

My thinking is that when writing or reading:

```
if !acc.JetStreamEnabled() {
		resp.Error = jsNotEnabledErr
		resp.Error = ApiErrors[JSNotEnabledErr]
		return
}
```

is easier to parse while scanning code due to the `JSNotEnabledErr` token than:

```
if !acc.JetStreamEnabled() {
		resp.Error = jsNotEnabledErr
		resp.Error = ApiErrors[JS1]
		return
}
```

This is perhaps not the best example but we have cases where we have like 3 or more if statements quite densely returning different errors.  For me when scanning such code I look for log lines and error strings rather than code - this retains a better readability to my eye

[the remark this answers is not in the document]
ripienaar: Also you get the benefit of IDE completing errors for you and you dont need to remember whats the code for an error, you see error constants.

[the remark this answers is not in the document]
ripienaar: Here is an earlier stab I took at doing this - we have many many errors:

https://github.com/ripienaar/nats-server/blob/cdef3e8acb07565336efda3c89b2e51be696bb99/server/jetstream_api.go#L600

I dont want to remember Kind codes.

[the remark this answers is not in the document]
kozlovic: I did not make myself clear, I was saying that we can have 2 maps, one keyed by the id like you have (say JSNotEnabledErr, etc), and another that will be keyed by the kind ("JS01", "JS02", etc..). As you can see from my code snippet, I use both maps.
The advantage of the 2nd map (keyed by kind) is that there is no longer "slow lookup". The range and checking for kind basically becomes:
```
func ApiErrorByKind(kind string) *ApiError {
    return ApiErrorsByKind[kind]
}
```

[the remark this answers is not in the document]
ripienaar: Sure, I dont think we have a need in the server for looking up by kind really and I dont want to forget to maintain one map all the time.  Can you think of a reason in the server to find by kind?

The CLI will loop when needed but its so fast - few 100 keys max - its effectively instantaneous.  I'd just remove the lookupbykind - there's no use. 

CLI will need to loop anyway as I want to be able to search/list/etc the kinds and so forth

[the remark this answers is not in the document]
kozlovic: Oh ok, I got confused, I thought that you would get this lookup in the server that the server could query, etc..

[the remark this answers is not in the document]
ripienaar: Yeah, no I dont think we really need that in the server

ripienaar:
@derekcollison I want to spend some time on this during this week, let me know what you think of the approach before I dive in on a wrong track, thanks!

derekcollison:
Probably deserves a meeting between a few folks.

ripienaar:
I added @matthiashanel and @aricart to reviewers for context, will have a call with all reviewers I guess.

derekcollison:
Add me onto the list as well if you can.

aricart:
I had looked at the PR before, and I agree that having a code that can be pointed at documentation can provide a richer experience to diagnose problems. The standard string that it represents is useful, but more so I think is the case where these values make it to a server log. Having documentation where the lookup can be performed is very useful.

server/jetstream_api.go
```diff
@@ -259,7 +259,22 @@ const JSMaxNameLen = 256
 // TODO(dlc) - Move to more generic location.
 type ApiError struct {
 	Code        int    `json:"code"`
+	Kind        string `json:"kind,omitempty"`
```
derekcollison, line 262: Pick different name?

Internal
ErrCode
Err

server/jetstream_api.go
```diff
@@ -259,7 +259,22 @@ const JSMaxNameLen = 256
...
+
+func (e *ApiError) Error() string {
+	return e.Description + " (" + e.Kind + ")"
+}
+
+func (e *ApiError) CopyF(a ...interface{}) *ApiError {
+	return &ApiError{
+		Code:        e.Code,
+		Kind:        e.Kind,
+		Description: fmt.Sprintf(e.Description, a...),
```
derekcollison, line 276: We can do the camel case here. Most are fine but if we wrap an err then staticcheck forces those to be all lowercase.

server/jetstream_api.go
```diff
@@ -588,8 +603,21 @@ type JSApiStreamTemplateNamesResponse struct {
...
+	JSNotEnabledErr
+	JSExternalDelPrefixOverlapsErrF
+)
+
 var (
-	jsNotEnabledErr        = &ApiError{Code: 503, Description: "JetStream not enabled for account"}
+	ApiErrors = map[APIErrorIdentifier]*ApiError{
+		JSUnknownErr:                    {Code: 503, Description: "Unknown error", Kind: "JS0"},
+		JSNotEnabledErr:                 {Code: 503, Description: "JetStream not enabled for account", Kind: "JS1", Help: "This error indicates that JetStream is not enabled either at a global level or at global and account level", URL: "https://docs.nats.io/jetstream"},
+		JSExternalDelPrefixOverlapsErrF: {Code: 400, Description: "stream external delivery prefix %q overlaps with stream subject %q", Kind: "JS2"},
```
derekcollison, line 618: Cap stream ?

server/jetstream_api.go
```diff
@@ -499,36 +483,31 @@ const JSApiMsgGetResponseType = "io.nats.jetstream.api.v1.stream_msg_get_respons
...
 
-// JSApiConsumerDeleteResponse.
 type JSApiConsumerDeleteResponse struct {
 	ApiResponse
 	Success bool `json:"success,omitempty"`
 }
 
 const JSApiConsumerDeleteResponseType = "io.nats.jetstream.api.v1.consumer_delete_response"
 
-// JSApiConsumerInfoResponse.
```
ripienaar, line 518: these are more or less useless comments so just remove them, they dont even make my IDE not complain

server/jetstream_errors.go
```diff
@@ -0,0 +1,220 @@
...
+
+import (
+	"fmt"
+)
+
+type APIErrorIdentifier int
+
+const (
+	JSUnknownErr APIErrorIdentifier = iota
+	JSAccountResourcesExceededErr
```
ripienaar, line 11: I kind of want to generate this file rather than hand maintain it. Given a source - json or yaml - that roughly looks like the structure a bit lower `go generate` would make this file.

The benefit of this is that I could automatically add a comment here above each of these identifiers that include the Description from the `ApiErrors` struct, this will greatly help code use because when you hover your mouse over the error you'll get some feedback about its body and such, now using these are a bit of pain.  

I doubt yous will like `go generate` based flow though.

server/jetstream_errors.go
```diff
@@ -0,0 +1,220 @@
...
+	JSStreamUpdateErrF
+	JSStreamWrongLastMsgIDErrF
+	JSStreamWrongLastSequenceErrF
+	JSTempStorageFailedErr
+	JSTemplateNameNotMatchSubjectErr
+)
+
+var (
+	ApiErrors = map[APIErrorIdentifier]*ApiError{
+		JSUnknownErr:                       {Code: 503, ErrCode: "JS001", Description: "Unknown error"},
```
ripienaar, line 87: Not decided yet if `JS001` is the right format, feedback welcome, I do think strings not numbers are better and the catagorisation that the `JS` adds is nice.

server/jetstream_errors.go
```diff
@@ -0,0 +1,220 @@
...
+
+func (e *ApiError) NewF(a ...interface{}) *ApiError {
+	return &ApiError{
+		Code:        e.Code,
+		ErrCode:     e.ErrCode,
+		Description: fmt.Sprintf(e.Description, a...),
+	}
+}
+
+func (e *ApiError) ErrOrNewF(err error, a ...interface{}) *ApiError {
```
ripienaar, line 204: will add some comments

server/jetstream_test.go
```diff
@@ -7417,17 +7417,15 @@ func TestJetStreamRequestAPI(t *testing.T) {
 		t.Fatalf("Created time seems wrong: %v\n", scResp.Created)
 	}
 
-	checkBadRequest := func(e *ApiError, description string) {
+	checkNatsError := func(e *ApiError, id APIErrorIdentifier) {
```
ripienaar, line 7420: do we have a more generic place for test helper functions? I suspec this will be used quite a lot

server/jetstream_errors.go
derekcollison: I am not a fan of it as a main build flow, but we could use it and check in the resulting files.

server/jetstream_api.go
derekcollison: What about `go vet` and `staticcheck`?

server/jetstream_errors.go
derekcollison: Numbers are directly comparable in all languages, that could be an advantage. Interested in what other's think.

server/jetstream_errors.go
derekcollison: I am the worst at naming but might consider something a bit more descriptive.

server/jetstream_errors.go
ripienaar: yes, that would be the idea - use it only when the source file changes and check in the result

server/jetstream_errors.go
ripienaar: I'd need some suggestions, as these are already attempts at improving the situation :)

derekcollison: Describe the function for us.

ripienaar: Uncommited locally I have this comment

```
// ErrOrNewF returns err if it's an ApiError else creates a new error using Sprintf
```

ripienaar:
```
err = acc.deleteStreamTemplate(name)
if err != nil {
		resp.Error = ApiErrors[JSStreamTemplateDeleteErrF].ErrOrNew(err)
		return
}
```

Here if deleteStreamTemplate() returns an apiError - that one will be used, otherwise a new `JSStreamTemplateDeleteErrF` is made via Sprintf

This way we can return a ApiError for things like template not found but a io error if deleting it from disk fails, and the fact that the problem is template not found would be surfaced correctly.

server/jetstream_api.go
kozlovic: Used that linters on my IDE would complain, but not any more. go vet and staticcheck don't seem to have a problem with those either. So agreed that if we are not adding more meaningful comment, then better removed.

server/jetstream_errors.go
kozlovic: Numbers would be easier indeed. But if the category is useful (JS), it could be made an ErrCategory: "JS" for people who care to know. But the ErrCode alone (a number) would be enough to check for the error code. However, for applications to be able to ignore the category means that error codes should be ever increasing, that is, do not repeat err codes because they belong to different categories.

PS: I get that Code already exists in ApiError, and RI proposed Kind first. But ErrCode is a bit of a stutter, as in ApiError.ErrCode.

server/mqtt.go
```diff
@@ -959,7 +959,7 @@ func (s *Server) mqttCreateAccountSessionManager(acc *Account, quitCh chan struc
 		Retention: InterestPolicy,
 		Replicas:  as.replicas,
 	}
-	if _, err := jsa.createStream(cfg); isErrorOtherThan(err, ErrJetStreamStreamAlreadyUsed) {
+	if _, err := jsa.createStream(cfg); isErrorOtherThan(err, ApiErrors[JSStreamNameExistErr]) {
```
kozlovic, line 962: Had this been in place, I would have checked for an error code here instead. The helper(s) createStream() and the like could be changed to return an error based on ApiError, but also the error code and that is what I would check here: (errors should be returned as last argument)
```
if _, errCode, err := jsa.createStream(cfg); errCode != "JS058" {
    return nil, fmt.Errorf("...", acc.GetName(), err)
}
```
(above is based on currently errCode be the ApiError.ErrCode which is a string as opposed to a number)

ripienaar: It seems quite a verbose pattern to also return a code there - making go's already awful error handling more verbose and more returned values etc. 

Mainly I want to make as little as possible burdon on those writing functions, shifting the effort instead on their callers - most callers would never need to check a specific err code.

```
if _, err := foo(); !IsNatsErr(err, "JS058") { ... }
```

Wouldnt overly complicate the functions, allow us to use richer errors progressively in functions (what if `foo()` returned a ApiErr sometimes othertimes a go error? which is happening a bit in functions already).  Probably about the same level of effort but feels more go like.

server/jetstream_errors.go
ripienaar: See comments below about how to check for errors etc, but I agree probably `JS001` is not needed.  How would you feel about codes like `1xxxx` being JS and `2xxxx` being MQTT etc?

server/jetstream_errors.go
ripienaar: Discussed with Ivan, we'll go with generate

ripienaar:
@kozlovic think I am good on this one, I updated MQTT code considering new errors and also added a ADR for how it works and how to maintain it (though I think that needs expanding in time).

for now keen to get this in so we can stop the rebasing pain :)

server/errors.go
```diff
@@ -171,46 +171,10 @@ var (
...
 
 	// ErrSubscribePermissionViolation is returned when processing of a subscription fails due to permissions.
-	ErrSubscribePermissionViolation = errors.New("subscribe permission viloation")
+	ErrSubscribePermissionViolation = errors.New("subscribe permission violation")
 
 	// ErrNoTransforms signals no subject transforms are available to map this subject.
 	ErrNoTransforms = errors.New("no matching transforms available")
-
-	// ErrJetStreamNotEnabled is returned when JetStream is not enabled.
-	ErrJetStreamNotEnabled = errors.New("jetstream not enabled")
```
kozlovic, line 180: As we discussed once, should we leave those but with a doc section that says that they are "deprecated"? I am thinking of use-case where users embed the NATS Server. I assume that internally they will no longer be used or updated.

kozlovic (approved):
LGTM, except not sure what to do with exported JS errors: removing them may break some apps (but unlikely).

server/errors.go
ripienaar: Yeah its a difficult one, we could say `ErrJetStreamNotEnabled = ApiErrors[JSWhatever]` but I am not 100% sure that will solve the problem in all cases tbh

ripienaar: Actually it should solve it for these - as they are essentially static, the ones where it wont solve it are the ones where we put in a different description via printf.  So it might be fine

ripienaar:
While we wait for @derekcollison to give this a look the one thing I've been mulling over is how we format the errors that are of the ErrorF form.

```
ApiErrors[JSStreamExternalApiOverlapErrF].ErrOrNewF(foo, bar)
```

Where the ApiErr is:

```
{Code: 400, ErrCode: 10021, Description: "stream external api prefix %q must not overlap with %s"},
```

The  problem here is that we want to be able to extend and change these errors - especially if we later need to translate them - then the order based replace becomes an error.

The typical fix is to make the description something like `"stream external api prefix {prefix} must not overlap with {api}"` and then:

```
ApiErrors[JSStreamExternalApiOverlapErrF].ErrOrNewF("{prefix}", foo, "{api}", bar)
```

This is much more robust and future proof, personally I think it's worth it, but curious what @ColinSullivan1 (who like me also seem to care for translation) and @kozlovic thinks

kozlovic:
> This is much more robust and future proof, personally I think it's worth it, but curious what @ColinSullivan1 (who like me also seem to care for translation) and @kozlovic thinks

I agree.

server/jetstream_api.go
```diff
@@ -262,7 +262,22 @@ const JSMaxNameLen = 256
 // TODO(dlc) - Move to more generic location.
 type ApiError struct {
 	Code        int    `json:"code"`
+	Kind        string `json:"kind,omitempty"`
 	Description string `json:"description,omitempty"`
+	URL         string `json:"-"`
```
derekcollison: Why are these never serialized and sent to clients?

server/jetstream_api.go
```diff
@@ -262,7 +262,22 @@ const JSMaxNameLen = 256
...
+	Help        string `json:"-"`
+}
+
+func (e *ApiError) Error() string {
+	return e.Description + " (" + e.Kind + ")"
```
derekcollison: Not sure if we expect for this to be called alot but this generates a bunch of temp objects etc. Might be better to do StringBuilder.

server/jetstream_api.go
```diff
@@ -262,7 +262,22 @@ const JSMaxNameLen = 256
 // TODO(dlc) - Move to more generic location.
 type ApiError struct {
 	Code        int    `json:"code"`
+	Kind        string `json:"kind,omitempty"`
```
derekcollison: We should think if these should be strings or numbers. For direct comparison we would have to look to C and would need numbers, not strings.

server/errors.go
derekcollison: We should leave, no harm imo..

server/jetstream_errors.go
```diff
@@ -84,100 +84,115 @@ const (
...
+
+	return ce.natsErrCode() == code
+}
+
 // ApiError is included in all responses if there was an error.
 // TODO(dlc) - Move to more generic location.
 type ApiError struct {
 	Code        int    `json:"code"`
-	Kind        string `json:"kind,omitempty"`
+	ErrCode     string `json:"err_code,omitempty"`
```
derekcollison: I think this should be u16.

derekcollison (approved):
In general LGTM. Went through them one by one so ignore repeated comments on codes need to be numbers.

server/jetstream_api.go
ripienaar: These are there mainly to assist tooling - I will add some `nats explain` or something that read these and generate documents, links to further information etc.

Doesnt seem appropraite to put in API responses - its a documentation link.

server/jetstream_errors.go
ripienaar: agree,

Not sure why I cant comment directly by the:

> We should leave, no harm imo..

I think there is plenty of harm tbh, in order to return the code we have to use a different type - we wont use this constant anymore so it'll just cause confusion.

We could as I say do `ErrJetStreamNotEnabled = ApiErrors[xxx]` but that only works for the kind of error without any dynamic content so would work in this specific case but in general not.

server/jetstream_api.go
ripienaar: not a lot but not that its a uint there we also have to do strconv etc - so I think just using fmt.Sprintf() is going to be best.

ripienaar:
I squashed this but I think I need more time to find a better solution to handle the named token replace things, forcing all things to be strings is terrible which is what strings.Replacer() wants.  So will investigate more.

server/jetstream_api.go
derekcollison: Should these be separate aggregate struct then that extends?

server/jetstream_errors.go
derekcollison: ok maybe you are correct and since we made the functions private no one could really run something and use those.

derekcollison:
We could do the string replace solution at a different time as well.

server/jetstream_api.go
ripienaar: Sure, but whats the benefit? I don't see how this is better or worse then?

derekcollison: Feels odd to have public members on a struct used to send JSON responses that includes things that we do not send.. But not super important. I would do aggregate (has-a) or not make them public, but again not that critical.

ripienaar: They are public so that other tools - like the NATS CLI can get access to them via the Go API rather than have to maintain some additional structures etc.

So for that use case I think this is exactly where they should be, even if I added them in another struct that extend, to allow the CLI to access that extending struct I would have to export that struct key and tell JSON to ignore it. Which feels about exactly the same as this.

derekcollison: Not following that part, so you are using the NATS server source code directly in the NATS cli or do you mean the Go client? Which is of course very different and will have its own version of this.

ripienaar: Only the CLI

Definitely not any clients - they only need to know the ErrCode which is in the JSON.

(something like)
```
$ nats explain 10039
Message:  JetStream not enabled for account
Help: https://docs.nats.io/jetstream

This error indicates that JetStream is not enabled either at a global level or at global and account level
```

This help url and additional short help comes out of these structs directly

ripienaar: However, to be exactly clear about why I did certain things, these errors are all in a JSON files in this repository. The build process of the CLI can pull in that json file and generate the richer structures there so this is all out of the nats server go code but we still end up with one place to maintain the data.

So I can completely remove these fields from the structs here and it would be ok(ish) though more work in the CLI.  But the JSON file is there so that any other language that wish to gain access to this stuff can do so

derekcollison: How do we plan on dealing with server / cli version mismatch in this case? Should the json be something you can get from the server's system account maybe?

ripienaar: We could for sure do that, we do have quite a bit of version coupling between CLI and Server atm so don't consider it the end of the world.

It's probably something we can improve around the time we do schema registeries - so this is some data that can also provide?

derekcollison: Might be worth discussing.

server/jetstream_errors.go
```diff
@@ -0,0 +1,105 @@
...
+func (e *ApiError) ErrOrNew(err error) *ApiError {
+	if ae, ok := err.(*ApiError); ok {
+		return ae
+	}
+
+	return e.NewT()
+}
+
+// NewT creates a new error using strings.Replacer on the Description field, arguments must be an even number like NewT("{err}", err)
+func (e *ApiError) NewT(replacements ...interface{}) *ApiError {
```
ripienaar, line 68: @kozlovic So I added tokenization in 40c9c1c and updated all the things and rebased to master.  wdyt?

derekcollison (commented):
I think we can merge once @kozlovic has signed off.

doc/adr/0007-error-codes.md
```diff
@@ -0,0 +1,142 @@
...
+
+## Design
+
+We will adopt a numbering system for our errors where every error has a unique number within a range that indicates the 
+subsystem it belongs to.
+
+|Range|Description|
+|-----|-----------|
+|1xxxx|JetStream related errors|
+|2xxxx|MQTT related errors|
```
kozlovic, line 33: To be clear, there is no such error at the moment...

server/errors.go
kozlovic: I think the point was that we would not use them anymore, but we may want to keep them as to possibly not break users that embed NATS Server and somehow were checking those error codes?

kozlovic (commented):
Last comment regarding removing server's existing `ErrXXX`...

server/errors.go
ripienaar: Yeah, I ppreciate why we would want to keep them - however, keeping them would not solve the problem completely.

For some we can keep them via an alias asignment but any of them that were dynamic would not be possible (the ones with tokens in them):

```
ErrJetStreamNotEnabled = ApiErrors[Foo]
```

Non of these had tokens of course so I need to look, might have to adjust `NewT()` and related functions to instead of a new instance return the existing pointer that is in `ApiErrors`, thats the only way it could possibly work.  This though is risky someone could adjust the "source" of these errors and bad things will happen, as there's no way to really make them read only as such

kozlovic: Yes, it silly of me to say to keep them not to "break" their code (because say `ErrJetStreamNotEnabled` is no longer defined), while it would break it in worse way by having the check for say `ErrJetStreamNotEnabled` be false because of the new errors returned.
So I would be ok if we remove in 2.3.0, but then if we have to release a 2.2.7 before that release, then we would have to create and maintain a `branch_2_2_x` to release from, which I would rather not have to do..

ripienaar: Old costants were restored in https://github.com/nats-io/nats-server/pull/2168/commits/ee9d10f40b8f16aacddd57f8142df7004422a532 with appropriate comments warning people etc

I think this one is good to go now

server/errors.json
```diff
@@ -187,9 +188,17 @@
   {
     "constant": "JSNotEnabledErr",
     "code": 503,
+    "error_code": 10076,
```
kozlovic, line 191: Changing the error code after this PR is merged is something that we can't be doing, correct?

This file is manually maintained, correct? I see that the error codes are a bit all over the places, which may be error prone when adding a new error to pick a code that is already used. But I guess the generator would check on that and fail if that is the case?

server/errors.json
```diff
@@ -187,9 +188,17 @@
...
+    "description": "JetStream not enabled",
+    "help": "This error indicates that JetStream is not enabled at a global level",
+    "deprectes": "ErrJetStreamNotEnabled"
```
kozlovic, line 194: Typo? `deprectes`

kozlovic (changes requested):
New field `deprecates` is misspelled in one of the error, also comment regarding ordering of the error codes in the json file.

server/errors.json
ripienaar: Yes, can not change.  The file is indeed a pain do you sort it by code, constant or what? :)

Generater does check and fail, I might make it report whats the highest current error code in that check though

ripienaar: Updated thanks, I imagine when we add other subsystems I'll need a tool to maintain this file

server/errors_gen.go
```diff
@@ -74,16 +74,20 @@ func goFmt(file string) error {
...
+		}
 	}
 
 	codeKeys := make(map[int]bool)
 	constKeys := make(map[string]bool)
 
 	for _, entry := range errs {
 		if _, found := codeKeys[entry.ErrCode]; found {
-			return fmt.Errorf("duplicate error code %+v", entry)
+			return fmt.Errorf("duplicate error code %+v, highest code is %d", entry, highest)
```
kozlovic, line 90: Good, but I thought that you would have an option to the tool to tell us what are the highest codes for each "category". Can be added later, so not holding the merge for that.

ripienaar: Indeed, once we have more than one cat I will add :P

PR #2169, opened 2021-04-30
JetStream extended over a leafnode fixes.

derekcollison:
We want to auto-suppress JetStream traffic on normal accounts when noticing the JetStream domain is extended.
We also now track remote accounts so that client info headers can be remapped if the accounts are different.

This PR does not have single leafnode as JetStream cluster extension yet.

/cc @nats-io/core

### DIFF
--- a/server/jetstream.go
+++ b/server/jetstream.go
@@ -228,7 +228,8 @@ func (s *Server) enableJetStream(cfg JetStreamConfig) error {
 	return nil
 }
 
-// This will check if we have a solicited leafnode that shares the system account.
+// This will check if we have a solicited leafnode that shares the system account
+// and is not denying subjects that would prevent extending JetStream.
 func (s *Server) hasSolicitLeafNodeSystemShare() bool {
 	sysAcc := s.SystemAccount().GetName()
 	for _, r := range s.getOpts().LeafNode.Remotes {
--- a/server/jetstream.go
+++ b/server/jetstream.go
@@ -228,7 +228,7 @@ func (s *Server) enableJetStream(cfg JetStreamConfig) error {
 	return nil
 }
 
-// This will check if we have the a solicited leafnode that shares the system account.
+// This will check if we have a solicited leafnode that shares the system account.
 func (s *Server) hasSolicitLeafNodeSystemShare() bool {
 	sysAcc := s.SystemAccount().GetName()
 	for _, r := range s.getOpts().LeafNode.Remotes {
--- a/server/jetstream.go
+++ b/server/jetstream.go
@@ -219,13 +219,29 @@ func (s *Server) enableJetStream(cfg JetStreamConfig) error {
 	}
 
 	// If we are in clustered mode go ahead and start the meta controller.
-	if !s.standAloneMode() {
+	if !s.standAloneMode() || s.hasSolicitLeafNodeSystemShare() {
 		if err := s.enableJetStreamClustering(); err != nil {
 			return err
 		}
 	}
 
 	return nil
+}
+
+// This will check if we have the a solicited leafnode that shares the system account.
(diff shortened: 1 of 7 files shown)


server/leafnode.go
```diff
@@ -108,8 +108,59 @@ func (c *client) isHubLeafNode() bool {
...
+		}
+	}
+	for _, dsubj := range r.DenyImports {
+		if dsubj == jsAllApi {
+			hasDI = true
+			break
+		}
+	}
+	if !hasDE {
+		s.Warnf("Adding deny export of %q for leafnode configuration on %q that bridges system account", jsAllApi, r.LocalAccount)
```
matthiashanel, line 128: Info instead of warning? (same for warn below)

server/leafnode.go
```diff
@@ -108,8 +108,59 @@ func (c *client) isHubLeafNode() bool {
...
+		if r.LocalAccount == sysAcc {
+			sysShared = true
+			break
+		}
+	}
+	if !sysShared {
+		return
+	}
+
+	s.Warnf("Sharing the system account across a leafnode remote")
```
matthiashanel, line 151: info

server/leafnode.go
derekcollison: Sure.. Will make the change.

matthiashanel (approved):
LGTM

kozlovic (approved):
LGTM

server/jetstream.go
```diff
@@ -228,6 +228,22 @@ func (s *Server) enableJetStream(cfg JetStreamConfig) error {
 	return nil
 }
 
+// This will check if we have the a solicited leafnode that shares the system account.
```
matthiashanel, line 231: system account and allow JS.

matthiashanel (commented):
Does this address having a single server in the hub? 
If that's the case, I'm ok with the change, just need to know.
I tried the leaf side, that works as expected

server/jetstream.go
derekcollison: No we don't check for JS here on purpose. Will introduce that concept per account with bridging work.
I do see comment could be worded better though, will fix that.

derekcollison:
I will double check on single server on hub side.
To make sure we are on same page, if other side is a single server we should not extend and treat each as island yes?

Also, I realized that we may want to inherit the server name as the cluster name when the single server is extending the JetStream domain, so will also add that to the test.

server/jetstream.go
matthiashanel: line 237 below checks that raft is not denied. Meaning this checks more than just if the system Account is mapped.
Thought that makes sense to mention in the comment

matthiashanel:
> I will double check on single server on hub side.
> To make sure we are on same page, if other side is a single server we should not extend and treat each as island yes?

The biggest use case for single server on the hub is to allow for smaller samples. Right now you need a cluster in order to get the leaf JS to join the hub JS.

If you have a hub of 1 and leaf of 1 and the config does not say otherwise, the. can work together.
I would treat the question about island or not separately from how big the clusters need to be on either end.

I would also be fine if we just have an option, to start a cluster of one that can be extended.
Or an option to say how many server are expected in the group. if that option says 1, then elect yourself leader.

server/jetstream.go
derekcollison: Fair point. Will do.

derekcollison:
I will offer a counter point against hub of 1 and leaf of 1 forming a cluster. The cluster is blocked if both are not up. I think we should detect it and keep them separate islands.

But in bridge mode if the leaf does not have JS but the hub does that will just work as expected.

derekcollison:
Please review last CL. Will handle 1:1 scenario from above in next PR on bridging and islands.

matthiashanel (approved):
LGTM

PR #2171, opened 2021-05-03
Add in formal support for multiple JetStream domains across leafnodes.

derekcollison:
This CL adds in support for multiple JetStream domains using mapped subjects.
Mapping subjects aligns well with the JetStream context APIPrefix in clients.

/cc @nats-io/core


server/jetstream_api.go
```diff
@@ -3420,8 +3526,12 @@ func (s *Server) jsConsumerDeleteRequest(sub *subscription, c *client, subject,
...
 
 	if !acc.JetStreamEnabled() {
-		resp.Error = jsNotEnabledErr
-		s.sendAPIErrResponse(ci, acc, subject, reply, string(msg), s.jsonResponse(&resp))
+		// This local account is not enabled, but we need to check of we are using leafnodes to bridge
+		// this account and if so not answer locally.
+		if acc.NumLeafNodes() == 0 {
+			resp.Error = jsNotEnabledErr
+			s.sendAPIErrResponse(ci, acc, subject, reply, string(msg), s.jsonResponse(&resp))
+		}
```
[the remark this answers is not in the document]
derekcollison: Will do as we discussed.

server/leafnode.go
```diff
@@ -140,18 +140,26 @@ func (s *Server) checkForSystemRemoteLeaf(remotes []*RemoteLeafOpts) {
...
+
+		// We want to deny normal JS API if we share the system account or our local account has JS enabled.
+		if sysShared {
 			s.addInJSDeny(r)
+		} else {
+			// Here we want to suppress if this local account has JS enabled.
+			// This is regardless of whether or not this server is actually running JS.
+			if acc, _ := s.lookupAccount(r.LocalAccount); acc != nil && acc.jetStreamConfigured() {
+				s.addInJSDeny(r)
+			}
```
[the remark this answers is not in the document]
derekcollison: Will change flow.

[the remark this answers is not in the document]
derekcollison: ok fixed. Will beef up test for mirrors and sources now.

server/opts.go
```diff
@@ -1518,6 +1520,10 @@ func parseJetStream(v interface{}, opts *Options, errors *[]error, warnings *[]e
 				opts.JetStreamMaxMemory = mv.(int64)
 			case "max_file_store", "max_file":
 				opts.JetStreamMaxStore = mv.(int64)
+			case "domain":
+				opts.JetStreamDomain = mv.(string)
+			case "enable", "enabled":
```
kozlovic, line 1525: Not clear on that: is that because you want to be able to specify a Domain, but not have JetStream enabled per-se? By the way, in earlier discussions with @matthiashanel, I think he thought of this setting be in the leafnodes{} accept block, no? (but may be better to lift it higher)

kozlovic (approved):
LGTM, but have a comment about adding a `enabled: true/false` option under jetstream{}

server/opts.go
derekcollison: Yes originally thought it would be around LNs but upon further thinking it belongs in the jetstream block. 
The enable is to support mixed mode clusters that all have active solicited LNs and want a separate JS domain for the LN cluster.

server/opts.go
```diff
@@ -205,6 +205,7 @@ type Options struct {
 	JetStream             bool          `json:"jetstream"`
 	JetStreamMaxMemory    int64         `json:"-"`
 	JetStreamMaxStore     int64         `json:"-"`
+	JetStreamDomain       string        `json:"-"`
```
matthiashanel, line 208: When this value is set, should the Jetstream startup list it like in the example below

```
[74483] 2021/05/06 10:00:02.844380 [INF]     _ ___ _____ ___ _____ ___ ___   _   __  __
[74483] 2021/05/06 10:00:02.844388 [INF]  _ | | __|_   _/ __|_   _| _ \ __| /_\ |  \/  |
[74483] 2021/05/06 10:00:02.844390 [INF] | || | _|  | | \__ \ | | |   / _| / _ \| |\/| |
[74483] 2021/05/06 10:00:02.844392 [INF]  \__/|___| |_| |___/ |_| |_|_\___/_/ \_\_|  |_|
[74483] 2021/05/06 10:00:02.844393 [INF]
[74483] 2021/05/06 10:00:02.844395 [INF]          https://docs.nats.io/jetstream
[74483] 2021/05/06 10:00:02.844397 [INF]
[74483] 2021/05/06 10:00:02.844400 [INF] ---------------- JETSTREAM ----------------
[74483] 2021/05/06 10:00:02.844404 [INF]   Max Memory:      48.00 GB
[74483] 2021/05/06 10:00:02.844406 [INF]   Max Storage:     509.30 GB
[74483] 2021/05/06 10:00:02.844408 [INF]   Store Directory: "store_server_3/jetstream"
[74483] 2021/05/06 10:00:02.844408 [INF]   Jetstream Domain: hub , available with prefix $JS.hub.API
[74483] 2021/05/06 10:00:02.844409 [INF] -------------------------------------------
```

server/jetstream.go
```diff
@@ -752,6 +776,17 @@ func (a *Account) EnableJetStream(limits *JetStreamAccountLimits) error {
 	s.Debugf("  Max Memory:      %s", friendlyBytes(limits.MaxMemory))
 	s.Debugf("  Max Storage:     %s", friendlyBytes(limits.MaxStore))
 
+	// Check if we have a Domain specified.
+	// If so add in a subject mapping that will allow local connected clients to reach us here as well.
+	opts := s.getOpts()
+	if opts.JetStreamDomain != _EMPTY_ {
+		s.Debugf("  Enable Domain:   %s", opts.JetStreamDomain)
+		src := fmt.Sprintf(jsDomainAPI, opts.JetStreamDomain)
+		if err := a.AddMapping(src, jsAllAPI); err != nil {
```
matthiashanel, line 785: Referring to later comment, maybe turn s.Debug into a trace and also print src

server/leafnode.go
```diff
@@ -110,57 +110,91 @@ func (c *client) isHubLeafNode() bool {
...
-		return
+}
+
+// Used for $SYS accounts when sharing but using separate JS domains.
+// r lock should be held.
+func (s *Server) addInJSDenyAll(r *leafNodeCfg) {
+	denyAll := []string{jscAllSubj, raftAllSubj, jsAllAPI}
+
+	s.Noticef("Sharing system account but utilizing separate JetStream Domains")
+	s.Noticef("Adding deny of %+v for leafnode configuration that bridges system account", denyAll, r.LocalAccount)
```
matthiashanel, line 160: missing %q at the end

matthiashanel (commented):
minor comments

server/leafnode.go
derekcollison: Will fix, thanks.

server/opts.go
```diff
@@ -1518,6 +1520,10 @@ func parseJetStream(v interface{}, opts *Options, errors *[]error, warnings *[]e
 				opts.JetStreamMaxMemory = mv.(int64)
 			case "max_file_store", "max_file":
 				opts.JetStreamMaxStore = mv.(int64)
+			case "domain":
+				opts.JetStreamDomain = mv.(string)
```
matthiashanel, line 1524: We are using this in a subject, can we check that opts.JetStreamDomain contains neither `.` nor ` ` (space) and error otherwise?

derekcollison: Dots are actually ok, does not matter. But could check if its a valid token or valid prefix subject once constructed. Happy to add test for that if we feel we should.

kozlovic (approved):
LGTM. Derek reviewed the last MQTT commit.

server/opts.go
matthiashanel: I would add the test just so we can fail fast (on startup).

matthiashanel: but can be separate pr

matthiashanel (approved):
LGTM

server/opts.go
derekcollison: I can add it in..

ripienaar: I feel quite strongly we should not allow `.` in a domain name, it significantly complicates writing ACL lists when domains can have dots, if a domain is a single token its really easy to be specific and restrctive, if its multiple tokens how would I write an access list allowing access to just one domain?

It also removes cognitive overhead since there's a single answer to what are the rules for jetstream object names - no dots anywhere. Adding a "but" to that is not good.

ripienaar:
Advisories that are sent need to be prefixed or have domain in them

derekcollison:
Domains in advisories might be good.

Question for @matthiashanel should we also segregate advisories like we do with $JS.API?

server/opts.go
ripienaar: Done in https://github.com/nats-io/nats-server/pull/2186

Issue #2170, opened 2021-05-01
Resource usage issues after upgrading Jetstream cluster to 2.2.2


cjbottaro:
We started seeing strange resource usage patterns after updating to 2.2.2. Our application code hasn't changed at all, and is very simple. It just creates a Jetstream stream and publishes messages to it (via new style requests). There are no consumers at all.

The memory usage of one of servers kept growing despite having zero connections to it. Also, it looks like the subscriptions are growing indefinitely, though that might be a reporting error.

![](https://storage.cjbotta.ro/20210501-130929.png)

After upgrading to 2.2.2, you can see the subscriptions keeps growing indefinitely, but we figured that was just a reporting issue due to metric names being changed. I.e. `gnatsd_varz_subscriptions` vs `gnatsd_subsz_num_subscriptions` (the latter doesn't grow indefinitely, but holds constant).

But then seeing memory usage grow on one server, seeming indefinitely until a sudden drop, caused us to open this issue. Something seem strange.

Also, all the server logs show this error happening a lot:
```
[1] 2021/04/30 17:11:03.145831 [WRN] JetStream resource limits exceeded for account: "$G"
```

We have very few connections to the Nats cluster (as seen in the graphs). Each connection makes a subscription to `_INBOX.<inbox-uid>.*` and makes requests using `reply_to`s like `_INBOX.<inbox-uid>.<request-uid>`. Lastly, the data volumes attached to each node are only at 20% capacity.

How to debug this further? Thanks for the help.

cjbottaro:
Some more info that might help track this down...

In 2.2.1, our stream seemed to be fine for a while, but then we'd starting seeing this in the log...
```
[WRN] JetStream resource limits exceeded for account: "$G"
```

Then the Nats servers would stop responding to any request regarding the stream, for example:
```
PUB $JS.API.STREAM.UPDATE.change-stream _INBOX.e14c4300.c9381997 332\r\n{\"discard\":\"old\",\"duplicate_window\":120000000000,\"host\":\"nats\",\"max_age\":86400000000000,\"max_bytes\":-1,\"max_consumers\":-1,\"max_msg_size\":-1,\"max_msgs\":-1,\"module\":\"Elixir.ChangeStream\",\"name\":\"change-stream\",\"num_replicas\":3,\"otp_app\":\"change_stream\",\"port\":4222,\"retention\":\"limits\",\"storage\":\"file\",\"subjects\":[\"change-stream.*\"]}\r\n
```
[...]

PR #2172, opened 2021-05-04
[FIXED] Getting varz from the http endpoint saw subscriptions always double for each fetch.

derekcollison:
Resolves part of #2170

/cc @nats-io/core


kozlovic (approved):
LGTM

PR #2173, opened 2021-05-04
Under heavy load a leader change could warn about not processing entry responses.

derekcollison:
This also adjust the minimum election timeout to 2 seconds vs just 1 for very large networks.

/cc @nats-io/core


ripienaar (approved):
LGTM

PR #2177, opened 2021-05-04
clarified client operations to promote cross-client compatibility

aricart:
Resubmission of the ADR edits with the specified clarifications as discussed with Colin and Ivan.

- Changed to be case-sensitive by default
- Removed all references to the Go implementation
- Clarified how append in a case-insensitive context should behave

[ci skip]

/cc @nats-io/core

### DIFF
--- a/doc/adr/0004-nats-headers.md
+++ b/doc/adr/0004-nats-headers.md
@@ -1,6 +1,6 @@
 # 4. nats-headers
 
-Date: 2020-05-28
+Date: 2021-05-12
 
 ## Status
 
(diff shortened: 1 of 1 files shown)


aricart:
After speaking with Ivan last week it was clear that the original intention of the manipulations was not clear. Again, the header implementation is case-preserving on writes, but case-insensitive on reads. The ambiguities are clarified. I would like the ADR to be passed before any client merges any implementations. If you are not clear about what the operations mean by all means slack me, so that we can finalize what is required of clients.

kozlovic:
> case-preserving on writes, but case-insensitive on reads

No, I think what we ended-up deciding is that it would be case-sensitive "by default". Client libraries are allowed to offer case-insensitive options, but the minimum to have is case-sensitive: "Foo" and "foo" will be different.
It is understood that some libraries have already been published with canonical, etc.. so the case-sensitive working may not be the "default" there, but they should be able to work with case-sensitivity.

aricart:
Perfect. Let me update and resubmit.

aricart:
@kozlovic @ColinSullivan1 I changed it to be optionally case-insensitive, if this is OK, let's get this in.

doc/adr/0004-nats-headers.md
```diff
@@ -8,38 +8,73 @@ Accepted
...
+The reference NATS Go client uses the `http.Header` which internally uses
+`textproto.MIMEHeader` infrastructure for serializing NATS headers. Other
+language clients may not be able to reuse header serialization infrastructure
+outside of the context of an HTTP request. Furthermore, the implementation of
+the native header library may diverge from the Go `http.Header` while still
+being standard.
+
+For example, the Go `http.Header` library, performs the following formatting
+which while not part of the spec, as header field names are case-insensitive,
+become a requirement in NATS header fields:
```
kozlovic, line 49: This is not correct. We no longer use MIME canonicalization in the Go client. Here is an example of setting a header in the Go client:
```
m := nats.NewMsg(subj)
m.Header.Set("ivan", "kozlovic")
nc.PublishMsg(m)
```
and how it is received in the server:
```
<<- MSG_PAYLOAD: ["NATS/1.0\r\nivan: kozlovic\r\n\r\n"]
```
As you can see, the key has not been canonicalized.

doc/adr/0004-nats-headers.md
```diff
@@ -92,3 +129,82 @@ Implemented and merged to master.
 ## Consequences
 
 Use of headers is possible.
+
+## Compatibility Across NATS Clients
+
+The following is a list of features to insure compatibility across NATS clients
+that support headers. Because the feature in Go client and nats-server leverage
+the Go implementation as described above, the API used will determine how header
+names are serialized.
```
kozlovic, line 138: Again, the GO client is now simply case-sensitive. The set/get/etc.. work with whatever user's input they get. Also, keys are not touched when header is decoded from the wire, or changed when encoded when sending the message out.

doc/adr/0004-nats-headers.md
```diff
@@ -92,3 +129,82 @@ Implemented and merged to master.
...
+
+The Go implementation of `http.Header` is a `map[string][]string` If the `Set`,
+`Get`, `Values`, `Del` functions are used, the specified key arguments will be
+converted into their canonical format. This means that the serialized Header
```
kozlovic, line 142: No, we no longer use that. The nats message has now its own Header type, so Set()/Add()/etc.. work directly with user key and don't do any modification.

kozlovic (changes requested):
Sorry to make you changes everything again, but I mentioned earlier that we have settled in case-sensitive API and case-preserving decode/encode of headers. A library will have to support that at a minimum, but can optionally provide case-insensitive options, etc..

doc/adr/0004-nats-headers.md
aricart: removed all references to the go library

ColinSullivan1 (approved):
LGTM.

scottf (approved):
LGTM

doc/adr/0004-nats-headers.md
```diff
@@ -65,7 +76,9 @@ HPUB <SUBJ> [REPLY] <HDR_LEN> <TOT_LEN>
 ```
 
 #### NOTES:
-- `HDR_LEN` includes the entire serialized header, from the start of the version string up to and including the ␍␊ before the payload
+
+- `HDR_LEN` includes the entire serialized header, from the start of the version
```
kozlovic, line 80: What version are we talking about here?

doc/adr/0004-nats-headers.md
```diff
@@ -82,7 +95,8 @@ HMSG <SUBJECT> <SID> [REPLY] <HDR_LEN> <TOT_LEN>
 <PAYLOAD>
 ```
 
-- `HDR_LEN` includes the entire serialized header, from the start of the version string up to and including the ␍␊ before the payload
+- `HDR_LEN` includes the entire serialized header, from the start of the version
```
kozlovic, line 98: Again, not sure why we mention version here? Are you referring to `NATS/1.0\r\n`?

kozlovic (commented):
Could I get a clarification on "version string" statement?

aricart:
Yes. This is all about the version nats is putting there

kozlovic (approved):
LGTM

PR #2178, opened 2021-05-04
[FIXED] MQTT fixes and improvements

kozlovic:
Some issues that have been fixed would manifest by timeouts on
connect, unexpected memory usage on high publish message rate.

Some details:
- Replies were not always GW routed properly because we were looking
at the wrong connection's rsubs
- GW routed replies would not be found because they were tracked
in the subscription's client object, which may not be the same used
to send the reply
- Increased the mqtt timeout to wait for JS replies since in some
tests it was sometimes taking more than the original 2 seconds
- Incoming gateway messages destined for an MQTT internal subscription
may have been rejected as a no interest if the account had service imports
- Don't use time.After(), instead create explicit timer so it can
be stopped when not timing out.
- Unnecessary copy of a slice since we were converting to a string anyway.

### DIFF
--- a/server/gateway.go
+++ b/server/gateway.go
@@ -238,6 +238,29 @@ type gwReplyMap struct {
 	exp int64
 }
 
+type gwReplyMapping struct {
+	// Indicate if we should check the map or not. Since checking the map is done
+	// when processing inbound messages and requires the lock we want to
+	// check only when needed. This is set/get using atomic, so needs to
+	// be memory aligned.
+	check int32
+	// To keep track of gateway replies mapping
+	mapping map[string]*gwReplyMap
+}
+
+// Returns the corresponding gw routed subject, and `true` to indicate that a
+// mapping was found. If no entry is found, the passed subject is returned
+// as-is and `false` is returned to indicate that no mapping was found.
+// Caller is responsible to ensure the locking.
+func (g *gwReplyMapping) get(subject []byte) ([]byte, bool) {
+	rm, ok := g.mapping[string(subject)]
+	if !ok {
+		return subject, false
+	}
+	subj := []byte(rm.ms)
+	return subj, true
+}
+
 // clone returns a deep copy of the RemoteGatewayOpts object
 func (r *RemoteGatewayOpts) clone() *RemoteGatewayOpts {
 	if r == nil {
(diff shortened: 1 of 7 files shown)


server/gateway.go
```diff
@@ -3023,31 +3027,42 @@ func (c *client) gatewaySwitchAccountToSendAllSubs(e *insie, accName string) {
...
 // Keeps track of the routed reply to be used when/if application
 // sends back a message on the reply without the prefix.
-// Client lock held on entry. This is a server receiver because
-// we use a timer interval that is avail in Server.gateway object.
-func (s *Server) trackGWReply(c *client, reply []byte) {
+// This is a server receiver because we use a timer interval that
+// is avail in Server.gateway object.
+func (s *Server) trackGWReply(acc *Account, reply, routedReply []byte) {
+	acc.mu.Lock()
+	defer acc.mu.Unlock()
```
kozlovic, line 3034: Maybe I could unlock after adding to the map after line 3062.

server/gateway.go
```diff
@@ -3023,31 +3027,42 @@ func (c *client) gatewaySwitchAccountToSendAllSubs(e *insie, accName string) {
...
 	ttl := s.gateway.recSubExp
-	rm := c.gwrm
+	rm := acc.gwrm
 	var we bool // will be true if map was empty on entry
```
derekcollison, line 3037: Can just set this to len(rm)

server/gateway.go
```diff
@@ -3023,31 +3027,42 @@ func (c *client) gatewaySwitchAccountToSendAllSubs(e *insie, accName string) {
...
 	if rm == nil {
 		rm = make(map[string]*gwReplyMap)
-		c.gwrm = rm
+		acc.gwrm = rm
 		we = true
```
derekcollison, line 3041: Can remove

server/gateway.go
```diff
@@ -3078,18 +3093,18 @@ func (s *Server) startGWReplyMapExpiration() {
...
+					for k, grm := range acc.gwrm {
 						if grm.exp <= now {
-							delete(c.gwrm, k)
-							if len(c.gwrm) == 0 {
-								atomic.StoreInt32(&c.cgwrt, 0)
-								s.gwrm.m.Delete(c)
+							delete(acc.gwrm, k)
+							if len(acc.gwrm) == 0 {
+								atomic.StoreInt32(&acc.cgwrt, 0)
+								s.gwrm.m.Delete(acc)
```
derekcollison, line 3103: nil out acc.gwrm too?

server/gateway.go
derekcollison: I like the way defer signals should be held throughout which I think is fine. Also prevents dangling locks so would keep as is.

matthiashanel (approved):
LGTM - But I have limited familiarity with the code

server/gateway.go
kozlovic: Concerned that we may get into a situation where we insert/remove often an then have to recreate the map all the times.

server/gateway.go
kozlovic: Fixing all that...

kozlovic:
@derekcollison I have been reworking all that quite a bit and rebased from main, sorry about that. The main change has been to keep the gw reply mapping in the client object for CLIENT connections, and to account for others. I have extracted the client's gwrm/cgwrt to its own struct so that code can be reused between client/account, etc..

derekcollison (approved):
LGTM

PR #2179, opened 2021-05-04
[FIXED] MQTT retained message in cluster mode may not be delivered

kozlovic:
In cluster mode, a sub connects to server 1, another on server 2.
A publisher connects to server 2 and publishes a retained message.
If both subs restart they would properly receive the retained message.

However, if the publisher sens an empty message that "removes" the
retained message for this topic, and then consumer that connects to
server 1 restarts, it would not receive the retained message as it
should.


derekcollison (approved):
LGTM

PR #2180, opened 2021-05-05
[IMPROVED] Stability when expireMsgs and writeMsg and Compact/Purge concurrent.

derekcollison:
We had issues of instability and incorrect behavior during concurrent operations.
This CL optimizes expiring msgs to be more efficient and hold the lock until completion.
Compact will also now hold the top level lock through completion.

/cc @nats-io/core


kozlovic (approved):
LGTM

PR #2183, opened 2021-05-05
[FIXED] Stream byte limit setting failed on dynamic account, filestore not honoring blksize.

derekcollison:

/cc @nats-io/core

### DIFF
--- a/server/errors.go
+++ b/server/errors.go
@@ -1,4 +1,4 @@
-// Copyright 2012-2020 The NATS Authors
+// Copyright 2012-2021 The NATS Authors
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
 // You may obtain a copy of the License at
@@ -202,6 +202,12 @@ var (
 
 	// ErrJetStreamResourcesExceeded is returned when a call would exceed internal resource limits.
 	ErrJetStreamResourcesExceeded = errors.New("jetstream resources exceeded for server")
+
+	// ErrStorageResourcesExceeded is returned when storage resources would be exceeded.
+	ErrStorageResourcesExceeded = errors.New("insufficient storage resources available")
+
+	// ErrMemeoryResourcesExceeded is returned when storage resources would be exceeded.
+	ErrMemoryResourcesExceeded = errors.New("insufficient memory resources available")
 )
 
 // configErr is a configuration error.
--- a/server/errors.go
+++ b/server/errors.go
@@ -206,7 +206,7 @@ var (
 	// ErrStorageResourcesExceeded is returned when storage resources would be exceeded.
 	ErrStorageResourcesExceeded = errors.New("insufficient storage resources available")
 
-	// ErrMemeoryResourcesExceeded is returned when storage resources would be exceeded.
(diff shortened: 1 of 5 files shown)


server/jetstream.go
```diff
@@ -1421,11 +1440,12 @@ func (js *jetStream) sufficientResources(limits *JetStreamAccountLimits) error {
 	if limits == nil {
 		return nil
 	}
+
 	if js.memReserved+limits.MaxMemory > js.config.MaxMemory {
-		return fmt.Errorf("insufficient memory resources available")
+		return ErrMemoryResourcesExceeded
```
ripienaar, line 1445: dont we have to handle -1 MaxStore and MaxMemory here too?

ripienaar (approved):
LGTM just one question

server/errors.go
```diff
@@ -202,6 +202,12 @@ var (
 
 	// ErrJetStreamResourcesExceeded is returned when a call would exceed internal resource limits.
 	ErrJetStreamResourcesExceeded = errors.New("jetstream resources exceeded for server")
+
+	// ErrStorageResourcesExceeded is returned when storage resources would be exceeded.
+	ErrStorageResourcesExceeded = errors.New("insufficient storage resources available")
+
+	// ErrMemeoryResourcesExceeded is returned when storage resources would be exceeded.
```
kozlovic, line 209: "when memory"

server/filestore.go
```diff
@@ -410,6 +411,12 @@ func (fs *fileStore) recoverMsgBlock(fi os.FileInfo, index uint64) *msgBlock {
 	}
 	defer file.Close()
 
+	if fi, _ := file.Stat(); fi != nil {
+		mb.rbytes = uint64(fi.Size())
+	} else {
+		return nil
```
kozlovic, line 417: So the `file` representing a message block has been opened just above, but here we say if `file.Stat()` fails basically, then we ignore? It may actually be better to continue no?

server/filestore.go
```diff
@@ -99,6 +99,7 @@ type msgBlock struct {
 	liwsz   int64
 	index   uint64
 	bytes   uint64
+	rbytes  uint64
```
kozlovic, line 102: I understand wanted to have short names, but to help reviewers, should we start to document the meaning of some short name variables that are not immediately obvious?

server/jetstream.go
derekcollison: js will have values, and when limits.Max are -1 its a no-op essentially. So not IMO.

server/errors.go
derekcollison: Will fix, and spelling on Memeory

server/filestore.go
derekcollison: Sure, line 1474 gives some ideas.

server/filestore.go
derekcollison: Stat should always succeed, just being overly cautious but if it does fail we should not continue imo.

server/filestore.go
derekcollison: Wrong line, is comment above blkSize() func. I did add some comments as requested.

kozlovic (approved):
LGTM. Thanks for adding the comments.

PR #2185, opened 2021-05-07
Check for bad domain names

derekcollison:

/cc @nats-io/core

### DIFF
--- a/server/jetstream_test.go
+++ b/server/jetstream_test.go
@@ -11208,6 +11208,24 @@ func TestJetStreamOperatorAccounts(t *testing.T) {
 	}
 }
 
+func TestJetStreamServerDomainBadConfig(t *testing.T) {
+	shouldFail := func(domain string) {
+		t.Helper()
+		conf := createConfFile(t, []byte(fmt.Sprintf("jetstream: {domain: \"%s\"}", domain)))
+		defer removeFile(t, conf)
+		if _, err := ProcessConfigFile(conf); err == nil {
+			t.Fatalf("Expected bad domain error, got %v", err)
+		}
+	}
+	shouldFail("HU..B")
+	shouldFail("HU B")
+	shouldFail(" ")
+	shouldFail("\t")
+	shouldFail("CORE.")
+	shouldFail(".CORE")
+	shouldFail("C.*.O. RE")
+}
+
 func TestJetStreamServerDomainConfig(t *testing.T) {
 	conf := createConfFile(t, []byte(`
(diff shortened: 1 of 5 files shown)


server/opts.go
```diff
@@ -1522,6 +1522,9 @@ func parseJetStream(v interface{}, opts *Options, errors *[]error, warnings *[]e
 				opts.JetStreamMaxStore = mv.(int64)
 			case "domain":
 				opts.JetStreamDomain = mv.(string)
+				if subj := fmt.Sprintf(jsDomainAPI, opts.JetStreamDomain); !IsValidSubject(subj) {
```
kozlovic, line 1525: In order to check for both file config and for users embedding the server, I recommend doing validation checks in server.go, `validateOptions()`

derekcollison: Good point..

matthiashanel (commented):
agreeing with Ivan

server/jetstream_test.go
```diff
@@ -11211,12 +11211,13 @@ func TestJetStreamOperatorAccounts(t *testing.T) {
 func TestJetStreamServerDomainBadConfig(t *testing.T) {
 	shouldFail := func(domain string) {
 		t.Helper()
-		conf := createConfFile(t, []byte(fmt.Sprintf("jetstream: {domain: \"%s\"}", domain)))
-		defer removeFile(t, conf)
-		if _, err := ProcessConfigFile(conf); err == nil {
-			t.Fatalf("Expected bad domain error, got %v", err)
+		opts := DefaultTestOptions
+		opts.JetStreamDomain = domain
+		if err := validateOptions(&opts); err == nil {
```
kozlovic, line 11216: This is fine, but to make sure that you are getting the error you are checking for, I usually do:
```
if err := validateOptions(&opts); err == nil || !strings.Contains(err.Error(), "Invalid domain name") {
    t.Fatalf("Expected bad domain error, got %v", err)
}
```

kozlovic (approved):
LGTM, nit on checking for expected error..

kozlovic (changes requested):
Sorry, I did not see that. Yes, when returning error, don't capitalize.

kozlovic (approved):

server/jetstream_test.go
derekcollison: Fixed.

derekcollison:
ok should be good.

kozlovic (approved):
LGTM. Thanks for adding the error check.

PR #2187, opened 2021-05-07
[FIXED] Mirror/Source streams from work queues and a deadlock on interest streams.

derekcollison:

/cc @nats-io/core

### DIFF
--- a/server/stream.go
+++ b/server/stream.go
@@ -139,7 +139,7 @@ type stream struct {
 	outq      *jsOutQ
 	msgs      *inbound
 	store     StreamStore
-	rmch      chan uint64
+	amch      chan uint64
 	lseq      uint64
 	lmsgId    string
 	consumers map[string]*consumer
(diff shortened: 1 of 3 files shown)


server/stream.go
```diff
@@ -318,10 +318,14 @@ func (a *Account) addStreamWithAssignment(config *StreamConfig, fsConfig *FileSt
 		stype:     cfg.Storage,
 		consumers: make(map[string]*consumer),
 		msgs:      &inbound{mch: make(chan struct{}, 1)},
-		rmch:      make(chan uint64, 8192),
 		qch:       make(chan struct{}),
 	}
 
+	// For no-ack consumers when we are interest retention.
+	if cfg.Retention != LimitsPolicy {
```
kozlovic, line 325: Can retention policy be changed/updated? If not good, otherwise, if user starts with different policy, then this channel won't be created and then later if policy is changed, this may cause problems. Unless this function itself would be invoked as part of a stream update?

kozlovic (approved):
LGTM. Question about lazy initialization of the amch channel in context of stream updates.

server/stream.go
derekcollison: RetentionPolicy can not be updated at this time.

derekcollison: I could add a nil check though when we place it, will do that.

kozlovic (approved):
LGTM

PR #2190, opened 2021-05-07
add domain in JS advisories

ripienaar:

/cc @nats-io/core

### DIFF
--- a/server/consumer.go
+++ b/server/consumer.go
@@ -835,6 +835,7 @@ func (o *consumer) sendDeleteAdvisoryLocked() {
 		Stream:   o.stream,
 		Consumer: o.name,
 		Action:   DeleteEvent,
+		Domain:   o.srv.getOpts().JetStreamDomain,
 	}
 
 	j, err := json.Marshal(e)
@@ -859,6 +860,7 @@ func (o *consumer) sendCreateAdvisory() {
 		Stream:   o.stream,
 		Consumer: o.name,
 		Action:   CreateEvent,
+		Domain:   o.srv.getOpts().JetStreamDomain,
 	}
 
 	j, err := json.Marshal(e)
@@ -1303,6 +1305,7 @@ func (o *consumer) processTerm(sseq, dseq, dc uint64) {
 		ConsumerSeq: dseq,
 		StreamSeq:   sseq,
 		Deliveries:  dc,
+		Domain:      o.srv.getOpts().JetStreamDomain,
 	}
 
 	j, err := json.Marshal(e)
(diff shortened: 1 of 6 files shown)


ripienaar:
Tests don't deadlock but a 2nd pair of eyes over locks would be good

server/consumer.go
```diff
@@ -836,6 +836,7 @@ func (o *consumer) sendDeleteAdvisoryLocked() {
 		Stream:   o.stream,
 		Consumer: o.name,
 		Action:   DeleteEvent,
+		Domain:   o.js.config.Domain,
```
derekcollison, line 839: I would pull from server opts. All assets have srv property.

So opts := o.srv.getOpts()
opts.JetStreamDomain.

server/jetstream.go
```diff
@@ -341,6 +341,7 @@ func (s *Server) handleOutOfSpace(stream string) {
 			ServerID: s.ID(),
 			Stream:   stream,
 			Cluster:  s.cachedClusterName(),
+			Domain:   s.js.config.StoreDir,
```
derekcollison, line 344: Same comment as above.

server/jetstream_api.go
```diff
@@ -828,6 +828,30 @@ func (s *Server) getRequestInfo(c *client, raw []byte) (pci *ClientInfo, acc *Ac
 	return &ci, acc, hdr, msg, nil
 }
 
+func (a *Account) jsDomain() string {
```
derekcollison, line 831: Can remove, or simplify as above. Accounts also have srv.

server/jetstream_api.go
```diff
@@ -2638,7 +2662,7 @@ func (s *Server) processStreamRestore(ci *ClientInfo, acc *Account, cfg *StreamC
 	s.Noticef("Starting restore for stream '%s > %s'", acc.Name, streamName)
 
 	start := time.Now().UTC()
-
+	domain := js.config.Domain
```
derekcollison, line 2665: Same here with s.getOpts()

server/jetstream_cluster.go
```diff
@@ -3086,6 +3090,7 @@ func (s *Server) sendConsumerLeaderElectAdvisory(o *consumer) {
 		Consumer: consumer,
 		Leader:   s.serverNameForNode(node.GroupLeader()),
 		Replicas: s.replicas(node),
+		Domain:   acc.jsDomain(),
```
derekcollison, line 3093: just do s.jsDomain() imo.

derekcollison (changes requested):

PR #2191, opened 2021-05-07
[FIXED] Raft groups could continually spin trying to catch up.

derekcollison:
In certain scenarios a follower could get stuck in an endless loop trying to catchup because the leader would not send a snapshot when the log was empty and the requested sequence was exactly the first sequence of the log, which would be lastIndex+1 for a used but empty store.

/cc @nats-io/core

### DIFF
--- a/server/consumer.go
+++ b/server/consumer.go
@@ -1599,9 +1599,10 @@ func (o *consumer) needAck(sseq uint64) bool {
 		}
 		state, err := o.store.State()
 		if err != nil || state == nil {
-			o.mu.RUnlock()
 			// Fall back to what we track internally for now.
-			return sseq > o.asflr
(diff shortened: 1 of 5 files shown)


kozlovic (approved):
LGTM

server/consumer.go
```diff
@@ -1605,7 +1605,8 @@ func (o *consumer) needAck(sseq uint64) bool {
 		state, err := o.store.State()
 		if err != nil || state == nil {
 			o.mu.RUnlock()
-			return false
+			// Fall back to what we track internally for now.
+			return sseq > o.asflr
```
kozlovic, line 1604: Should you capture o.asflr before releasing consumer's lock?

kozlovic (approved):
LGTM

PR #2192, opened 2021-05-07
Don't need to store domain in mqttJSA structure

kozlovic:
At the time where we need the domain to construct the session hash,
we have access to server options. So use that instead of storing
the domain in the internal mqtt structure.


derekcollison (approved):
LGTM

PR #2197, opened 2021-05-08
Create 0006-protocol-naming-conventions

scottf:
Attempt to define naming convention

/cc @nats-io/core


scottf:
Some clients already support unicode (Java, ...) We might need / want to support it

PR #2201, opened 2021-05-10
Fix for failing test

derekcollison:
We needed to account for consumers having filtered subjects configured but not really being filtered.

/cc @nats-io/core

### DIFF
--- a/server/consumer.go
+++ b/server/consumer.go
@@ -1586,6 +1586,25 @@ func (o *consumer) processAckMsg(sseq, dseq, dc uint64, doSample bool) {
 	}
 }
 
+// Determine if this is a truly filtered consumer. Modern clients will place filtered subjects
+// even if the stream only has a single non-wildcard subject designation.
+// Read lock should be held.
+func (o *consumer) isFiltered() bool {
+	if o.cfg.FilterSubject == _EMPTY_ {
+		return false
+	}
+	// If we are here we want to check if the filtered subject is
+	// a direct match for our only listed subject.
+	mset := o.mset
+	if mset == nil {
+		return true
+	}
+	if len(mset.cfg.Subjects) > 1 {
+		return true
+	}
+	return o.cfg.FilterSubject != mset.cfg.Subjects[0]
+}
+
 // Check if we need an ack for this store seq.
 // This is called for interest based retention streams to remove messages.
 func (o *consumer) needAck(sseq uint64) bool {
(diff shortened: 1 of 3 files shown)


server/jetstream_cluster_test.go
```diff
@@ -2145,8 +2142,9 @@ func TestJetStreamClusterMirrorAndSourceInterestPolicyStream(t *testing.T) {
 
 	checkFor(t, 5*time.Second, 250*time.Millisecond, func() error {
 		// This one will be 0 since no other interest exists.
-		if si, _ := js.StreamInfo("WQ22"); si.State.Msgs != 1 {
-			return fmt.Errorf("Expected 1 msg for %q, got %d", "WQ22", si.State.Msgs)
+		if si, _ := js.StreamInfo("IP22"); si.State.Msgs != 1 {
+			fmt.Printf("WQ22 state is %+v\n", si.State)
```
kozlovic, line 2146: Remove?

server/consumer.go
```diff
@@ -1586,6 +1586,27 @@ func (o *consumer) processAckMsg(sseq, dseq, dc uint64, doSample bool) {
...
+// Read lock should be held.
+func (o *consumer) isFiltered() bool {
+	if o.cfg.FilterSubject == _EMPTY_ {
+		return false
+	}
+	// If we are here we want to check if the filtered subject is
+	// a direct match for our only listed subject.
+	mset := o.mset
+	if mset == nil {
+		return true
```
kozlovic, line 1600: is `mset == nil` a normal condition? If so, what does that mean? (so I understand why we would return `true` here).

server/consumer.go
```diff
@@ -1586,6 +1586,27 @@ func (o *consumer) processAckMsg(sseq, dseq, dc uint64, doSample bool) {
...
+	}
+	cfg := mset.cfg
```
kozlovic, line 1602: cfg is a struct, so that makes it a copy, which can be expensive since you call that for every ack'ed message. If you just don't want to repeat mset.cfg, at least do: `cfg := &mset.cfg`

kozlovic (changes requested):

server/jetstream_cluster_test.go
derekcollison: Apologies, I will add tooling for this, my manual checks not doing so well lately.

server/consumer.go
derekcollison: We remove the mset from the consumer when deleted. Will change that up but not in this PR, so its there for safety.

server/consumer.go
derekcollison: Thanks.

server/consumer.go
kozlovic: My question is then: when that is the case, does it makes more sense to return `true` (is filtered) or `false`? Or does it not matter..

derekcollison: In this case we do not want to assume it is not.

kozlovic (approved):
LGTM

PR #2204, opened 2021-05-11
[added] check for mirror stream return error when  de dupe window is set

matthiashanel:

Noticed this in natscli, where I removed the interactive question. 
For an actual check RI preferred doing this here.


derekcollison (approved):
LGTM - Thanks!

PR #2206, opened 2021-05-11
[FIXED] A store directory on disk without 'jetstream' could appear to lose assets on restart.

derekcollison:

/cc @nats-io/core

### DIFF
--- a/server/jetstream.go
+++ b/server/jetstream.go
@@ -164,7 +164,54 @@ func (s *Server) EnableJetStream(config *JetStreamConfig) error {
 		cfg.StoreDir = filepath.Join(os.TempDir(), JetStreamStoreDir)
 	}
 
+	// We will consistently place the 'jetstream' directory under the storedir that was handed to us. Prior to 2.2.3 though
+	// we could have a directory on disk without the 'jetstream' directory. This will check and fix if needed.
+	s.checkStoreDir(&cfg)
+
 	return s.enableJetStream(cfg)
+}
+
+// Check to make sure directory has the jetstream directory.
+// We will have ot properly configured here now regardless, so need to look inside.
+func (s *Server) checkStoreDir(cfg *JetStreamConfig) {
+	fis, _ := ioutil.ReadDir(cfg.StoreDir)
+	// If we have nothing underneath us, could be just starting new, but if we see this we can check.
+	if len(fis) != 0 {
+		return
+	}
+	// Let's check the directory above. If it has us 'jetstream' but also other stuff that we can
+	// identify as accounts then we can fix.
+	fis, _ = ioutil.ReadDir(filepath.Dir(cfg.StoreDir))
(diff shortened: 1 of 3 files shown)


server/jetstream.go
```diff
@@ -164,9 +164,56 @@ func (s *Server) EnableJetStream(config *JetStreamConfig) error {
...
+				if acc, err := s.lookupAccount(accName); err != nil && acc != nil {
+					ok = true
+				}
+			}
+			// If this seems to be an account go ahead and move the directory. This will include all assets
+			// like streams and consumers.
+			if ok {
+				old := filepath.Join(filepath.Dir(cfg.StoreDir), fi.Name())
+				new := filepath.Join(cfg.StoreDir, fi.Name())
+				s.Debugf("JetStream relocated %q to %q", old, new)
```
ripienaar, line 210: This is an important log line, I'd make it an info. 

We have this pattern in a number of places where we think the account server is somehow reliable, its not something thats reliable.  Account lookups can fail etc, being able to determine that a file was moved or not is important.  I'd also log that some directory was not relocated due to account not existing - same reason.

server/jetstream.go
```diff
@@ -164,9 +164,56 @@ func (s *Server) EnableJetStream(config *JetStreamConfig) error {
...
+	fis, _ := ioutil.ReadDir(cfg.StoreDir)
+	// If we have nothing underneath us, could be just starting new, but if we see this we can check.
+	if len(fis) != 0 {
+		return
+	}
+	// Let's check the directory above. If it has us 'jetstream' but also other stuff that we can
+	// identify as accounts then we can fix.
+	fis, _ = ioutil.ReadDir(filepath.Dir(cfg.StoreDir))
+	// If just one that is us 'jetstream' and all is ok.
+	if len(fis) == 1 {
```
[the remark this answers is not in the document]
derekcollison: We know `jetstream` will be there. If other stuff is there we look to see if those directories are accounts. 

There could be OS/Platform specific files in the upper directory. Will take a closer look.

ripienaar (approved):
LGTM

server/jetstream.go
```diff
@@ -164,9 +165,56 @@ func (s *Server) EnableJetStream(config *JetStreamConfig) error {
...
 
+	// We will consistently place the 'jetstream' directory under the storedir that was handed to us. Prior to 2.2.3 though
+	// we could have a directory on disk without the 'jetstream' directory. This will check and fix if needed.
+	s.checkStoreDir(&cfg)
+
 	return s.enableJetStream(cfg)
 }
 
+// Check to make sure directory has the jetstream directory.
+// We will have ot properly configured here now regardless, so need to look inside.
```
kozlovic, line 176: `ot` -> `it`?

derekcollison: Thx, fixed. Force pushed.

PR #2210, opened 2021-05-12
Add initial support for OCSP Stapling

wallyqs:
This adds basic OCSP Stapling support in the server for client connections.  To enable, the server has to start with the following:

```hcl
ocsp: true

tls { ... }
```

When OCSP is enabled this way, in case a certificate has a 'status_request' // Must Staple flag toggled, then it will make an 
HTTP request to the OCSP remote url and fetch staples that will be cached on disk and serve the staples to the clients so that they can confirm the response.

 - [X] Tests added

/cc @nats-io/core


derekcollison:
In general like the account resolver we should place all things under the store directory..

server/ocsp.go
```diff
@@ -0,0 +1,442 @@
...
+		)
+	} else if err == nil && shouldShutdown {
+		// If resp.Status is ocsp.Revoked, ocsp.Unknown, or any other value.
+		s.Errorf("Found existing OCSP certificate status: %s", ocspStatusString(resp.Status))
+		s.Shutdown()
+		return
+	}
+
+	for {
+		time.Sleep(nextRun)
```
derekcollison, line 208: These make me nervous, should be a select and timeout etc.

server/ocsp.go
```diff
@@ -0,0 +1,442 @@
...
+
+		s.mu.Lock()
+		running := s.running
```
derekcollison, line 211: This can be part of the for loop. For s.Running()..

server/ocsp.go
```diff
@@ -0,0 +1,442 @@
...
+		shuttingDown := s.shutdown
+		s.mu.Unlock()
+
+		if !running && !shuttingDown {
```
derekcollison, line 215: Would select on quit chan for server.

server/ocsp.go
```diff
@@ -0,0 +1,442 @@
...
+		if !running && shuttingDown {
+			// Server is shutting down. Exit goroutine.
+			s.Noticef("Shutting down OCSP service...")
+			return
+		}
+
+		_, resp, err := oc.getRemoteStatus()
+		if err != nil {
+			nextRun = oc.getNextRun()
+			s.Errorf("Bad OCSP status update: %s, trying again in %s", err, nextRun)
```
derekcollison, line 229: %v?

server/ocsp.go
```diff
@@ -0,0 +1,442 @@
...
+// getOCSPIssuer returns a CA cert from the given path. If the path is empty,
+// then this checks a given cert chain. If both are empty, then it returns an
+// error.
+func getOCSPIssuer(issuerCert string, chain [][]byte) (*x509.Certificate, error) {
+	var issuer *x509.Certificate
+	var err error
+	switch {
+	case len(chain) == 1 && issuerCert == "":
+		err = fmt.Errorf("require ocsp ca in chain or configuration")
+	case issuerCert != "":
```
derekcollison, line 374: _EMPTY_

server/opts.go
```diff
@@ -261,6 +261,12 @@ type Options struct {
 	// and used as a filter criteria for some system requests
 	Tags jwt.TagList `json:"-"`
 
+	// StateDir is a directory to hold state of the server.
+	StateDir string `json:"-"`
```
derekcollison, line 265: Let's require store_dir opt and tuck under own directory under there like `jetstream`

server/ocsp.go
```diff
@@ -0,0 +1,442 @@
+package server
```
ColinSullivan1, line 1: Need license + copyright, 2020.

derekcollison: 2021 ;)

test/ocsp_test.go
```diff
@@ -0,0 +1,565 @@
+package test
```
ColinSullivan1, line 1: copyright...

server/opts.go
```diff
@@ -828,6 +843,29 @@ func (o *Options) processConfigFileLine(k string, v interface{}, errors *[]error
 		o.TLSTimeout = tc.Timeout
 		o.TLSMap = tc.Map
 
+		// Need to keep track of path of the certs for OCSP Stapling.
+		o.TLSCert = tc.CertFile
+		o.TLSKey = tc.KeyFile
+		o.TLSCaCert = tc.CaFile
+	case "state_dir":
```
matthiashanel, line 850: can we use a different. name here? state is too generic a name.

matthiashanel: will probably need this just in case jetstream is not set

wallyqs: Think maybe for this first pass we will remove this and always fetch staples on boot // avoid caching to address correctly store dir location, permissions, etc...

server/ocsp.go
```diff
@@ -0,0 +1,442 @@
...
+			continue
+		}
+
+		switch n := resp.Status; n {
+		case ocsp.Good:
+			nextRun = oc.getNextRun()
+			t := resp.NextUpdate.Format(time.RFC3339Nano)
+			s.Noticef(
+				"Received OCSP certificate status: good, next update %s, checking again in %s",
+				t, nextRun,
```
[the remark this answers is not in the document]
wallyqs: It is a `time.Duration`

wallyqs:
Removed the caching of staples to a local directory to address in a different PR instead.

server/ocsp.go
wallyqs: Updated to use a select and also using the quitCh so that it does not block.

server/opts.go
wallyqs: Removed caching the files into a local directory for now, will have a different branch to address this part.

test/ocsp_test.go
```diff
@@ -0,0 +1,563 @@
...
+			conf := createConfFile(t, []byte(content))
+			defer removeFile(t, conf)
+			s, opts := RunServerWithConfig(conf)
+			defer s.Shutdown()
+
+			nc, err := nats.Connect(fmt.Sprintf("tls://localhost:%d", opts.Port),
+				test.certs,
+				nats.RootCAs(caCert),
+				nats.ErrorHandler(noOpErrHandler),
+			)
```
wallyqs, line 371: Commit from the nats.go that shows how to verify OCSP responses in the client: https://github.com/nats-io/nats.go/commit/50e25c411c42c6a6998252bee9ce50be89137077#diff-1d2fb9af677137c0182213770500dd5625a51ba7286301cb11d0197b21e0f948R1755

wallyqs:
Some unrelated flappers but seems ok to take another look now.

server/ocsp.go
```diff
@@ -0,0 +1,377 @@
...
+
+func (oc *OCSPMonitor) run() {
+	s := oc.srv
+	s.mu.Lock()
+	quitCh := s.quitCh
+	s.mu.Unlock()
+
+	defer s.grWG.Done()
+	config := s.getOpts().OCSPConfig
+	shouldShutdown := config != nil && (config.Mode == OCSPModeMust || config.Mode == OCSPModeAlways)
```
philpennock, line 163: This doesn't handle the certificate having MustStaple in it.  The Must concept should be coming from the certificate rather than from configuration.  "Always" is for us saying we demand OCSP, but a certificate with the MustStaple field set is telling clients that it's a configuration error to see the certificate without the staple, which is why that's equivalent to us saying Always.

server/ocsp.go
```diff
@@ -0,0 +1,377 @@
...
+		case ocsp.Good:
+			nextRun = oc.getNextRun()
+			t := resp.NextUpdate.Format(time.RFC3339Nano)
+			s.Noticef(
+				"Received OCSP certificate status: good, next update %s, checking again in %s",
+				t, nextRun,
+			)
+			continue
+		default:
+			s.Errorf("Received OCSP certificate status: %s, shutting down", ocspStatusString(n))
```
philpennock, line 204: This should log something about the certificate which we got this response for.  A typical server might have two or three certificates, any of which might have their certificate be revoked and the OCSP response return the non-Good state.

So we should include something about the certificate to identify which one is invalid and needs to be replaced.

server/ocsp.go
```diff
@@ -0,0 +1,377 @@
...
+	cert, err := tls.LoadX509KeyPair(tcOpts.CertFile, tcOpts.KeyFile)
+	if err != nil {
+		return nil, nil, err
+	}
+	cert.Leaf, err = x509.ParseCertificate(cert.Certificate[0])
+	if err != nil {
+		return nil, nil, err
+	}
+	if oc.Mode == OCSPModeAuto && !hasOCSPStatusRequest(cert.Leaf) {
+		// "status_request" flag not set. No need to do anything.
```
philpennock, line 229: Is NewOCSPMonitor() called again on reload when a new certificate is seen, or does this mean that if we start with a certificate without must-staple and later replace it with a must-staple cert then we won't set up the OCSP staples for it?

server/ocsp.go
```diff
@@ -0,0 +1,377 @@
...
+
+	// Callbacks below will be in charge of returning the certificate instead,
+	// so this has to be nil.
+	tc.Certificates = nil
+
+	mon := &OCSPMonitor{
+		srv: srv,
+		hc:  &http.Client{Timeout: 30 * time.Second},
+		// FIXME: Make reloadable aware, otherwise a change in certificates
+		//  could make the staples stale.
```
philpennock, line 248: Note that this being missing means that TLS clients will drop the connection until it's fixed server-side.  Seeing an OCSP staple for the wrong certificate is a hard TLS error, if I recall correctly.

server/ocsp.go
```diff
@@ -0,0 +1,377 @@
...
+		Leaf:   cert.Leaf,
+		Issuer: issuer,
+	}
+
+	// Get the certificate status from the memory, disk, then remote OCSP responder.
+	_, resp, err := mon.getStatus()
```
philpennock, line 254: This is missing handling for the staple in the disk cache being stale at start and needing to be replaced immediately.

server/ocsp.go
```diff
@@ -0,0 +1,377 @@
...
+
+	now := time.Now()
+	if nextUpdate.IsZero() {
+		// If response is missing NextUpdate, we check the day after.
+		// Technically, if NextUpdate is missing, we can try whenever.
+		// https://tools.ietf.org/html/rfc6960#section-4.2.2.1
+		return 24 * time.Hour
+	}
+
+	return nextUpdate.Sub(now) / 2
```
philpennock, line 76: This doesn't handle the staple having already expired and needing to be replaced already, in which case this will return a negative duration, causing the `time.After()` to return immediately.  We'll busy-loop hammering the OCSP responder.  There needs to be a minimum duration here, so that the nats-server instances don't get deny-listed by the CAs.

server/ocsp.go
```diff
@@ -0,0 +1,377 @@
...
+	// GetClientCertificate returns a certificate that's presented to a
+	// server.
+	tc.GetClientCertificate = func(info *tls.CertificateRequestInfo) (*tls.Certificate, error) {
+		return &cert, nil
+	}
+	return tc, mon, nil
+}
+
+func hasOCSPStatusRequest(cert *x509.Certificate) bool {
+	tlsFeatures := asn1.ObjectIdentifier{1, 3, 6, 1, 5, 5, 7, 1, 24}
```
philpennock, line 289: Comment mentioning RFC 7633 as the reference and noting that this corresponds to `id-pe-tlsfeature` ?

server/ocsp.go
```diff
@@ -0,0 +1,377 @@
...
+	// https://tools.ietf.org/html/rfc6066
+
+	for _, ext := range cert.Extensions {
+		if !ext.Id.Equal(tlsFeatures) {
+			continue
+		}
+		if len(ext.Value) != tlsFeaturesLen {
+			continue
+		}
+		return ext.Value[tlsFeaturesLen-1] == statusRequestExt
```
philpennock, line 304: This needs a HUGE FIXME comment on it.
The value of the extension identified by `tlsFeatures` is an ASN.1 sequence of integers.
Here you've checked that the payload has a length matching the byte length if there's just one small integer in the list and checked that the last field is that small integer.

A `tlsFeatures` extension with multiple features in it would cause a parse failure here.

https://play.golang.org/p/kCgFUCg2gYE is an example of how to do this right; specifically:
```
var intSeq []int
rest, err := asn1.Unmarshal(rawBytes, &intSeq)
```

Once you have the `intSeq` then you can iterate it looking for the `statusRequestExt` value.

server/ocsp.go
```diff
@@ -0,0 +1,377 @@
...
+// error.
+func getOCSPIssuer(issuerCert string, chain [][]byte) (*x509.Certificate, error) {
+	var issuer *x509.Certificate
+	var err error
+	switch {
+	case len(chain) == 1 && issuerCert == _EMPTY_:
+		err = fmt.Errorf("require ocsp ca in chain or configuration")
+	case issuerCert != _EMPTY_:
+		issuer, err = parseCertPEM(issuerCert)
+	case len(chain) > 1 && issuerCert == "":
```
philpennock, line 339: curious: why do you write `""` here but `_EMPTY_` above?

server/server.go
```diff
@@ -1451,6 +1451,32 @@ func (s *Server) fetchAccount(name string) (*Account, error) {
 	return acc, nil
 }
 
+func (s *Server) enableOCSP() error {
+	opts := s.getOpts()
+	// Check if OCSP Stapling was enabled/disabled globally.
+	oc := opts.OCSPConfig
+	if oc == nil || oc.Mode == OCSPModeNever {
```
philpennock, line 1458: If `OCSPModeAuto` is the default, then surely `oc == nil` should imply `OCSPModeAuto` and we should not return nil for that scenario?

server/server.go
```diff
@@ -1451,6 +1451,32 @@ func (s *Server) fetchAccount(name string) (*Account, error) {
...
+		}
+		// Check if an OCSP stapling monitor is required for this certificate.
+		if mon != nil {
+			// Override the TLS config with one that follows OCSP.
+			opts.TLSConfig = tc
+			s.startGoRoutine(func() { mon.run() })
+		}
+	}
+	// FIXME: Add support for leafnodes, routes, MQTT, WebSocket
+	s.Noticef("OCSP Stapling enabled")
```
philpennock, line 1475: This could do with identifying which cert it's for (but for now, just mention that requirement in the FIXME comment)

server/server.go
```diff
@@ -1603,6 +1629,17 @@ func (s *Server) Start() {
 		})
 	}
 
+	// Setup OCSP Stapling monitors.
+	// This will abort server from starting if there are no valid staples
+	// and OCSP policy is Always or Must Staple.
+	if oc := opts.OCSPConfig; oc != nil {
```
philpennock, line 1635: If `oc == nil` or if it's not nil and the mode is Auto, and the certificate says MustStaple, then we need to abort for that scenario too.

philpennock (changes requested):
I did not review `test/ocsp_test.go` or decode the PEM test-data to check those for sanity.

server/ocsp.go
variadico: Cool, now documented: https://github.com/nats-io/nats-server/pull/2210/commits/ec5bfb9470aad3a9d8681bbc3d2bb316f2a7e2f9

server/ocsp.go
variadico: Awesome, here's the updated version: https://github.com/nats-io/nats-server/pull/2210/commits/30e741e8af3ea4297a4bcc7cb2c9655bf5cfa5e5

server/ocsp.go
wallyqs: Currently it always fetches from the remote OCSP server, we will address that in the next PR that includes caching to disk.

server/ocsp.go
wallyqs: Yes we don't handle reload in this PR, but that is what should be done.

wallyqs: It is reload aware now, monitors are restarted when TLS certs change.

server/ocsp.go
wallyqs: The OCSP stapling monitors are stopped/resume on reload now.

server/ocsp.go
```diff
@@ -0,0 +1,454 @@
...
+	"net/http"
+	"strings"
+	"sync"
+	"time"
+
+	"golang.org/x/crypto/ocsp"
+)
+
+const (
+	DefaultOCSPCheckInterval = 24 * time.Hour
```
kozlovic, line 33: Late to comment on that, but we need to think harder when exporting this (applies to all other OCSP things here). I think you had to export because you decided to have the tests in the `/test` package, but I am a big proponent of having tests under `/server`, along with the file that is being tested. It also gives you the flexibility to access server internals if needed.

Once things are exported, we can't fix/change them easily without having the risk to break users apps, which we have to realize that a lot of users actually embed the NATS server.

server/server.go
wallyqs: Now it outputs for which port the staples will be fetched: 
```
OCSP Stapling enabled for client connections
```

server/server.go
wallyqs: If cert has MustStaple flag then it activates Must staple mode.

server/ocsp.go
wallyqs: +1 these can be made private

server/ocsp.go
```diff
@@ -0,0 +1,454 @@
...
+			continue
+		}
+
+		switch n := resp.Status; n {
+		case ocsp.Good:
+			nextRun = oc.getNextRun()
+			t := resp.NextUpdate.Format(time.RFC3339Nano)
+			s.Noticef(
+				"Received OCSP status for certificate '%s': good, next update %s, checking again in %s",
+				t, certFile, nextRun,
```
philpennock, line 228: The `t` and `certFile` parameters appear to be reversed here.

philpennock (changes requested):

wallyqs:
Rebased and commit branch with storing staples to disk can be found here: https://github.com/nats-io/nats-server/pull/2240/commits/01c922954d4f4e868b47ef90a36ed702832bfc1d

wallyqs:
More complete version of the feature can be found here so closing this one. OCSP Stapling of client certs, reloading aware, caching to disk etc... https://github.com/nats-io/nats-server/pull/2210

PR #2211, opened 2021-05-12
[FIXED] Websocket: Origin should be checked only if present

### LINKED ISSUE #2207: How to provide origin header when connecting a leafnode server via websocket?
Currently we are testing NATS leafnode connections via websocket. This is a very cool feature and important to us, as it could reduce the infrastructure costs a lot.
I saw the "allowed_origins" parameter in the websocket config block and tried it out. Unfortunately I don't know how to provide the matching value from the leafnode side, there is no config option. 
Our NATS server logs the following error: "websocket handshake error: origin not allowed: origin not provided".
With the restriction commented out the connection works as expected. 
Could you describe, how we have to provide the origin info (client's request Origin header) to the server?

kozlovic:
There are 2 options, same_origin and allowed_origins that should
apply only to webbrowsers that set the Origin http header. If
the header is not present, the server should not fail direct
clients using websocket protocol, or leafnodes.

From spec:
https://datatracker.ietf.org/doc/html/rfc6455#section-1.6

   The WebSocket Protocol uses the origin model used by web browsers to
   restrict which web pages can contact a WebSocket server when the
   WebSocket Protocol is used from a web page.  Naturally, when the
   WebSocket Protocol is used by a dedicated client directly (i.e., not
   from a web page through a web browser), the origin model is not
   useful, as the client can provide any arbitrary origin string.

Resolves #2207

/cc @nats-io/core


server/websocket.go
```diff
@@ -769,11 +769,16 @@ func (w *srvWebsocket) checkOrigin(r *http.Request) error {
 		return nil
 	}
 	origin := r.Header.Get("Origin")
-	if origin == "" {
+	if origin == _EMPTY_ {
 		origin = r.Header.Get("Sec-Websocket-Origin")
 	}
-	if origin == "" {
-		return errors.New("origin not provided")
+	// If theader is not present, we will accept.
```
derekcollison, line 775: spelling

derekcollison (approved):
LGTM - Minor spelling nit.

derekcollison (approved):
LGTM

Issue #2205, opened 2021-05-11
When the leaf node is a cluster, stream create request responses (and others) do not seem to return.


matthiashanel:
I ran a version of the following setup, that only has two JS domains (hub/spoke-1).
When spoke-1 is a single node, everything appears to be working.
When spoke-1 is a cluster, responses seem to get lost.  I'm writing this as I observe the streams get created.

```
+------------------------------------------------------------------------------+
|                                Stream Report                                 |
+--------+---------+-----------+----------+-------+------+---------+-----------+
| Stream | Storage | Consumers | Messages | Bytes | Lost | Deleted | Replicas  |
+--------+---------+-----------+----------+-------+------+---------+-----------+
| test0  | File    | 0         | 0        | 0 B   | 0    | 0       | srv-4252* |
| test1  | File    | 0         | 0        | 0 B   | 0    | 0       | srv-4252* |
| test2  | File    | 0         | 0        | 0 B   | 0    | 0       | srv-4252* |
| test3  | File    | 0         | 0        | 0 B   | 0    | 0       | srv-4242* |
+--------+---------+-----------+----------+-------+------+---------+-----------+
```

only one response was received. (when `--js-domain spoke-1`  is specified, the request goes to `$JS.spoke-1.API.STREAM.CREATE.test5`)
```
> nats --context=hub s add --js-domain spoke-1 --config test.conf --subjects test0 test0
nats: error: could not create Stream: context deadline exceeded
> nats --context=hub s add --js-domain spoke-1 --config test.conf --subjects test1 test1
nats: error: could not create Stream: context deadline exceeded
> nats --context=hub s add --js-domain spoke-1 --config test.conf --subjects test2 test2
^[[Anats: error: could not create Stream: context deadline exceeded
> nats --context=hub s add --js-domain spoke-1 --config test.conf --subjects test3 test3
Stream test3 was created

Information for Stream test3 created 2021-05-10T21:37:34-04:00

Configuration:

             Subjects: test3
     Acknowledgements: true
            Retention: File - Limits
             Replicas: 1
       Discard Policy: Old
     Duplicate Window: 2m0s
     Maximum Messages: unlimited
        Maximum Bytes: unlimited
          Maximum Age: 0.00s
 Maximum Message Size: unlimited
    Maximum Consumers: unlimited


Cluster Information:

                 Name: cluster-spoke-1
               Leader: srv-4242

State:

             Messages: 0
                Bytes: 0 B
             FirstSeq: 0
[...]

PR #2212, opened 2021-05-12
[FIXED] Issue #2205

derekcollison:
When a response was needed from a leafnode cluster back to a hub, we had rules to disallow.
That rule was a bit dated and since we have cluster origin for leafnode clusters and that
is checked before the message is actually sent we could remove the old rule.

Resolves: #2205 

/cc @nats-io/core

### DIFF
--- a/server/client.go
+++ b/server/client.go
@@ -3998,8 +3998,9 @@ func (c *client) processMsgResults(acc *Account, r *SublistResult, msg, deliver,
 			// We handle similarly to routes and use the same data structures.
 			// Leaf node delivery audience is different however.
 			// Also leaf nodes are always no echo, so we make sure we are not
-			// going to send back to ourselves here.
-			if c != sub.client && (c.kind != ROUTER || sub.client.isHubLeafNode()) {
+			// going to send back to ourselves here. For messages from routes we want
+			// to suppress in general unless we know from the hub or its a service reply.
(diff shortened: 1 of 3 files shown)


server/client.go
```diff
@@ -3998,8 +3998,9 @@ func (c *client) processMsgResults(acc *Account, r *SublistResult, msg, deliver,
 			// We handle similarly to routes and use the same data structures.
 			// Leaf node delivery audience is different however.
 			// Also leaf nodes are always no echo, so we make sure we are not
-			// going to send back to ourselves here.
-			if c != sub.client && (c.kind != ROUTER || sub.client.isHubLeafNode()) {
+			// going to send back to ourselves here. For messages from routes we want
+			// to suppress in general unless we know from the hub or its a service reply.
+			if c != sub.client && (c.kind != ROUTER || sub.client.isHubLeafNode() || isServiceReply(c.pa.subject)) {
```
kozlovic, line 4003: To minimize the changes and matches the comment, should we restrict the check for isServiceReply to route clients? Something like `... || (c.kind == ROUTER && isServiceReply(c.pa.subject))) {`

derekcollison: If it is evaluated (last condition) we already know that the kind is ROUTER no? We even know its a route and not a hub..

kozlovic: Right...

kozlovic (approved):
LGTM

Issue #2213, opened 2021-05-12
single server JS does show direct consumer when it shouldn't (clustered JS does not do that)


matthiashanel:
config used (ignore the names saying leaf, this is a single server cluster):
```
port: 4111
server_name: leaf-server-1
jetstream {
	store_dir="./store_leaf_1"
}

cluster {
  name: leaf
}

include ./accounts.conf
```

Steps to reproduce:
1. add a stream
2. add a source stream sourcing first stream
3. first stream now lists extra consumer

```
> nats -s nats://rmt:rmt@localhost:4111 s ls
No Streams defined
> nats -s nats://rmt:rmt@localhost:4111 s add
? Stream Name test
? Subjects to consume test
? Storage backend file
? Retention Policy Limits
? Discard Policy Old
? Stream Messages Limit -1
? Message size limit -1
? Maximum message age limit -1
? Maximum individual message size -1
? Duplicate tracking time window 2m
? Replicas 1
Stream test was created

Information for Stream test created 2021-05-12T00:18:40-04:00

Configuration:

             Subjects: test
     Acknowledgements: true
            Retention: File - Limits
             Replicas: 1
       Discard Policy: Old
     Duplicate Window: 2m0s
     Maximum Messages: unlimited
        Maximum Bytes: unlimited
          Maximum Age: 0.00s
 Maximum Message Size: unlimited
    Maximum Consumers: unlimited


State:

             Messages: 0
                Bytes: 0 B
             FirstSeq: 0
              LastSeq: 0
     Active Consumers: 0
> nats -s nats://rmt:rmt@localhost:4111 s add  --source test
X Sorry, your reply was invalid: Value is required
? Stream Name agg
? Storage backend file
? Retention Policy Limits
? Discard Policy Old
? Stream Messages Limit -1
? Message size limit -1
? Maximum message age limit -1
? Maximum individual message size -1
? Duplicate tracking time window 2m
? Replicas 1
? Adjust source "test" start No
? Import "test" from a different JetStream domain No
? Import "test" from a different account No
Stream agg was created

Information for Stream agg created 2021-05-12T00:18:59-04:00

Configuration:

     Acknowledgements: true
            Retention: File - Limits
             Replicas: 1
       Discard Policy: Old
     Duplicate Window: 2m0s
     Maximum Messages: unlimited
        Maximum Bytes: unlimited
          Maximum Age: 0.00s
 Maximum Message Size: unlimited
    Maximum Consumers: unlimited
[...]

PR #2214, opened 2021-05-12
[FIXED] #2213

derekcollison:
We do not want to report direct consumers that were created for the purpose of sources or mirrors.

Resolves: #2213 

/cc @nats-io/core


matthiashanel (approved):
LGTM

PR #2216, opened 2021-05-12
Error-checking & resilience for JetStream move

philpennock:
Jetstream movement can fail, so return that error and abort start-up if there's
a failure in moving precious data, rather than serve without it.

Create the jetstream directory if needed.

~~Create directories for private data mode 0750 not 0755.~~

This does not handle a directory layout made with 2.2.3, but does support a
2.2.2 to 2.2.4 migration.  The empty directories made under 2.2.3 will still
hinder the renames we do here.

 - Create the jetstream dir if needed instead of assuming it exists
 - Check for errors and return them
 - ~~Use better permissions for jetstream storage~~

/cc @nats-io/core

### DIFF
--- a/server/jetstream.go
+++ b/server/jetstream.go
@@ -167,25 +167,35 @@ func (s *Server) EnableJetStream(config *JetStreamConfig) error {
 
 	// We will consistently place the 'jetstream' directory under the storedir that was handed to us. Prior to 2.2.3 though
 	// we could have a directory on disk without the 'jetstream' directory. This will check and fix if needed.
-	s.checkStoreDir(&cfg)
+	if err := s.checkStoreDir(&cfg); err != nil {
+		return err
+	}
 
 	return s.enableJetStream(cfg)
 }
 
 // Check to make sure directory has the jetstream directory.
 // We will have it properly configured here now regardless, so need to look inside.
-func (s *Server) checkStoreDir(cfg *JetStreamConfig) {
+func (s *Server) checkStoreDir(cfg *JetStreamConfig) error {
 	fis, _ := ioutil.ReadDir(cfg.StoreDir)
 	// If we have nothing underneath us, could be just starting new, but if we see this we can check.
 	if len(fis) != 0 {
-		return
+		return nil
 	}
 	// Let's check the directory above. If it has us 'jetstream' but also other stuff that we can
 	// identify as accounts then we can fix.
 	fis, _ = ioutil.ReadDir(filepath.Dir(cfg.StoreDir))
 	// If just one that is us 'jetstream' and all is ok.
 	if len(fis) == 1 {
-		return
+		return nil
+	}
+
+	haveJetstreamDir := false
+	for _, fi := range fis {
+		if fi.Name() == JetStreamStoreDir {
+			haveJetstreamDir = true
+			break
+		}
 	}
 
 	for _, fi := range fis {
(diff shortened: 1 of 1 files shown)


server/jetstream.go
```diff
@@ -223,7 +244,7 @@ func (s *Server) enableJetStream(cfg JetStreamConfig) error {
 
 	// FIXME(dlc) - Allow memory only operation?
 	if stat, err := os.Stat(cfg.StoreDir); os.IsNotExist(err) {
-		if err := os.MkdirAll(cfg.StoreDir, 0755); err != nil {
+		if err := os.MkdirAll(cfg.StoreDir, 0750); err != nil {
```
derekcollison, line 247: This one looks like a good change but makes me nervous for a dot release.

Interested in what @kozlovic thinks and possibly @ripienaar

derekcollison (commented):
In general LGTM, one comment I think we need to decide on first though.

server/jetstream.go
kozlovic: I really don't know the consequences of this change.

ripienaar: Also agree in principal but feel we need to consider it a bit more and test more especially to see how this interacts on windows.

server/jetstream.go
```diff
@@ -206,13 +216,24 @@ func (s *Server) checkStoreDir(cfg *JetStreamConfig) {
 			// If this seems to be an account go ahead and move the directory. This will include all assets
 			// like streams and consumers.
 			if ok {
+				if !haveJetstreamDir {
```
kozlovic, line 219: With the spirit of bailing out on any error, which I don't disagree with, then what about the lookup error few lines above? We are currently simply logging but move on. Should we bail out?
My only concern is that is the account resolver always known to be fully setup at this point in the startup sequence? Are we going to get to a point with some configs where the server won't be able to restart?

server/jetstream.go
philpennock: Can remove the 0755->0750 change if desired

server/jetstream.go
philpennock: There's a point at which "starting up without data" is far more dangerous than "refusing to start".  In this case, we risk creating a divergent forked view of what state exists which is hard to reconcile, and the safe option is to refuse to start and require administrator intervention, instead of effectively continuing on without precious state.

User-data in jetstream is precious state.  If we can't handle the migration, we should not start.

The main cause of errors is filesystem errors.  Primarily because of detritus from 2.2.3 which is still not handled.  The fixes here improve things to where we will successfully migrate from 2.2.2 to 2.2.4 and where we will not silently fail if 2.2.3 layout exists.  It's better.  It's not perfect.

ripienaar: My review also requested specifically checking the assumption that a `jetstream` dir actually exist rather than trusting len()==1 etc.

server/jetstream.go
ripienaar: Yeah, go does claim to do the right thing - but they are known for doing clever but dumb things rather than clever but right things (http headers, arbitrary x509 restrictions etc) so I'd rather trust but verify that their clever thing is actually the right thing.

Also being that we now made it 755 if we change it we should probably make this a configurable to not burn users who now depend on the past behaviour.

philpennock: Yeah, I wondered about making it configurable.  I think we have the fire of the current issue and the desire to fix the permissions and should avoid compounding issues.  I've undone the permissions change and stuck to the same 0755 used in the old code and the new.

philpennock: https://github.com/nats-io/nats-server/issues/2217 filed to track the permissions change.  Resolving this conversation.

server/jetstream.go
kozlovic: My point was that: should we go a step further and return error when account lookup fails?

ripienaar: Sorry, I missed the details of your question @kozlovic. 

I don't know what the right answer is about the account lookup - but this was also in my initial review since account servers are not reliable or to be trusted.  At the very least we should log loudly in detail and we should ensure that on the next startup we will again try to migrate until nothing is left.

Problem is we have no way of knowing what nothing is left mean since we do not have a clear enumerable way to determine that so I really don't know what the answer is. If migration is a one-shot attempt because we are not capable of picking this up reliably then we need to log loudly manual instructions for moving this data.

Definitely coming around to @philpennock's suggestion that we should have pulled 2.2.3.

kozlovic: Re: pulling 2.2.3.. Not sure we can remove from Docker hub official images.. which is likely where most people get the release from.

kozlovic:
@philpennock @derekcollison I made a change: I don't think we need the fetchAccount() when we detect that this is not a local account (not in config, and we have account resolver) but the name indicates that this is a valid NKEY account public key.

derekcollison (approved):
LGTM

server/jetstream.go
philpennock: If 2.2.4 is imminent then perhaps no longer worth it this time, but perhaps a docker-library/official-images PR which moves `latest` back to a known-good release is the right way to go, even if we can't actually remove a tag.

kozlovic: Here are going to remove "now", so hopefully no need to do that.

PR #2218, opened 2021-05-12
Fix stream source lookup and add in optional External to StreamSource

derekcollison:

/cc @nats-io/core

### DIFF
--- a/server/stream.go
+++ b/server/stream.go
@@ -103,10 +103,11 @@ type PeerInfo struct {
 
 // StreamSourceInfo shows information about an upstream stream source.
 type StreamSourceInfo struct {
-	Name   string        `json:"name"`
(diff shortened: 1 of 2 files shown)


server/stream.go
```diff
@@ -117,7 +118,7 @@ type StreamSource struct {
 	FilterSubject string          `json:"filter_subject,omitempty"`
 	External      *ExternalStream `json:"external,omitempty"`
 
-	// Internal
+	// Internale
```
kozlovic, line 121: "Internale": why the change?

kozlovic (approved):
LGTM (except for typo maybe?)

server/stream.go
derekcollison: Spelling?

matthiashanel (commented):
LGTM

PR #2220, opened 2021-05-13
0006-protocol-naming-conventions

scottf:
Proposal for protocol naming convention

/cc @nats-io/core

### DIFF
--- a/doc/adr/0006-protocol-naming-conventions.md
+++ b/doc/adr/0006-protocol-naming-conventions.md
@@ -44,11 +44,12 @@ term             = (printable except dot, asterisk or gt)+
 prefix           = (printable except dot, asterisk, gt or dollar)+
 filename-safe    = (printable except dot, asterisk, lt, gt, colon, double-quote, fwd-slash, backslash, pipe, question-mark, ampersand)
 
-message-subject  = term (dot term | asterisk)* (dot gt)?
-reply-to         = term (dot term)*
-stream-name      = term
-queue-name       = term
-durable-name     = term
-jetstream-prefix = prefix
-account-name     = (filename-safe)+ maximum 255 characters
+message-subject    = term (dot term | asterisk)* (dot gt)?
+reply-to           = term (dot term)*
+stream-name        = term
+queue-name         = term
+durable-name       = term
+js-internal-prefix = dollar (prefix dot)+
+js-user-prefix     = (prefix dot)+
+account-name       = (filename-safe)+ maximum 255 characters
 ```
--- a/doc/adr/0006-protocol-naming-conventions.md
+++ b/doc/adr/0006-protocol-naming-conventions.md
@@ -1,0 +1,56 @@
+# 6. protocol-naming-conventions
+
+Date: 2021-05-08
+
+## Status
+
+Under Review
+
+## Context
+
+This document describes naming conventions for these protocol components:
+
+* message subject
+* stream name
+* reply to
+* queue name
+* durable name
+* account name
+
+## Prior Work
+
+Currently the NATS Docs regarding [protocol convention](https://docs.nats.io/nats-protocol/nats-protocol#protocol-conventions) says this:
+
+> Subject names, including reply subject (INBOX) names, are case-sensitive and must be non-empty alphanumeric strings with no embedded whitespace. All ascii alphanumeric characters except spaces/tabs and separators which are "." and ">" are allowed. Subject names can be optionally token-delimited using the dot character (.), e.g.:
+A subject is comprised of 1 or more tokens. Tokens are separated by "." and can be any non space ascii alphanumeric character. The full wildcard token ">" is only valid as the last token and matches all tokens past that point. A token wildcard, "*" matches any token in the position it was listed. Wildcard tokens should only be used in a wildcard capacity and not part of a literal token.
+
+> Character Encoding: Subject names should be ascii characters for maximum interoperability. Due to language constraints and performance, some clients may support UTF-8 subject names, as may the server. No guarantees of non-ASCII support are provided.
+
+## Specification
+
+```
+dot              = "."
+asterisk         = "*"
+lt               = "<"
+gt               = ">"
+dollar           = "$"
+colon            = ":"
+double-quote     = ["]
+fwd-slash        = "/"
+backslash        = "\"
+pipe             = "|"
+question-mark    = "?"
+ampersand        = "&"
+printable        = all printable ascii (33 to 126 inclusive)
+term             = (printable except dot, asterisk or gt)+
+prefix           = (printable except dot, asterisk, gt or dollar)+
+filename-safe    = (printable except dot, asterisk, lt, gt, colon, double-quote, fwd-slash, backslash, pipe, question-mark, ampersand)
+
+message-subject  = term (dot term | asterisk)* (dot gt)?
+reply-to         = term (dot term)*
+stream-name      = term
+queue-name       = term
+durable-name     = term
+jetstream-prefix = prefix
+account-name     = (filename-safe)+ maximum 255 characters
+```
--- a/doc/adr/0006-protocol-naming-conventions.md
+++ b/doc/adr/0006-protocol-naming-conventions.md
@@ -1,10 +1,10 @@
 # 6. protocol-naming-conventions
 
-Date: 2021-05-08
+Date: 2021-06-28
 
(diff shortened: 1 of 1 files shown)


ColinSullivan1 (approved):
LGTM.

doc/adr/0006-protocol-naming-conventions.md
```diff
@@ -0,0 +1,56 @@
...
+
+## Status
+
+Under Review
+
+## Context
+
+This document describes naming conventions for these protocol components:
+
+* message subject
```
derekcollison, line 13: Would Cap and make it like..

Subjects (including Reply Subjects)
Stream Names
Consumer Names
Account Names

doc/adr/0006-protocol-naming-conventions.md
```diff
@@ -0,0 +1,56 @@
...
+* queue name
+* durable name
+* account name
+
+## Prior Work
+
+Currently the NATS Docs regarding [protocol convention](https://docs.nats.io/nats-protocol/nats-protocol#protocol-conventions) says this:
+
+> Subject names, including reply subject (INBOX) names, are case-sensitive and must be non-empty alphanumeric strings with no embedded whitespace. All ascii alphanumeric characters except spaces/tabs and separators which are "." and ">" are allowed. Subject names can be optionally token-delimited using the dot character (.), e.g.:
+A subject is comprised of 1 or more tokens. Tokens are separated by "." and can be any non space ascii alphanumeric character. The full wildcard token ">" is only valid as the last token and matches all tokens past that point. A token wildcard, "*" matches any token in the position it was listed. Wildcard tokens should only be used in a wildcard capacity and not part of a literal token.
```
kozlovic, line 25: > should only be used in a wildcard capacity and not part of a literal token

They should but we support them being just characters in a token, that is, as literal as opposed to a wildcard. We had users in the past reporting that we were not treating them properly and had to fix that, and the server (at least) is supposed to recognize "foo*.bar" as a literal subject with 2 tokens, one `foo*` and one `bar`.
But it is understood that in this case `foo*` is just a string and has no wildcard semantic, which some may think that it would "match"  any token that starts with `foo`.

doc/adr/0006-protocol-naming-conventions.md
```diff
@@ -0,0 +1,56 @@
...
+colon            = ":"
+double-quote     = ["]
+fwd-slash        = "/"
+backslash        = "\"
+pipe             = "|"
+question-mark    = "?"
+ampersand        = "&"
+printable        = all printable ascii (33 to 126 inclusive)
+term             = (printable except dot, asterisk or gt)+
+prefix           = (printable except dot, asterisk, gt or dollar)+
```
kozlovic, line 46: Is the `dollar` in the except list? Yet the default JS prefix is `$JS.API.`

doc/adr/0006-protocol-naming-conventions.md
```diff
@@ -0,0 +1,56 @@
...
+filename-safe    = (printable except dot, asterisk, lt, gt, colon, double-quote, fwd-slash, backslash, pipe, question-mark, ampersand)
+
+message-subject  = term (dot term | asterisk)* (dot gt)?
+reply-to         = term (dot term)*
+stream-name      = term
+queue-name       = term
+durable-name     = term
+jetstream-prefix = prefix
```
kozlovic, line 54: See above: unless you are saying that this applies to users and they should not use a `$` sign?

doc/adr/0006-protocol-naming-conventions.md
scottf: changed

doc/adr/0006-protocol-naming-conventions.md
scottf: changed

doc/adr/0006-protocol-naming-conventions.md
scottf: right, `dollar` can be in a server / internal prefix but not a user prefix. (user prefix changed here too). Maybe leave prefix and add this?
```
internal-prefix = dollar (prefix dot)+
user-prefix     = (prefix dot)+
```

doc/adr/0006-protocol-naming-conventions.md
scottf: Is `*` a regular character if it is followed by `.`, `*` or `>` or what would the rules? The prior work is currently the published doc.

doc/adr/0006-protocol-naming-conventions.md
kozlovic: Sure. And to be clear: by prefix you mean the JS API prefix, right? Because stream names allow "$" and MQTT implementation in the server creates streams that start with `$MQTT_`, so wanted to be sure that this is different.

doc/adr/0006-protocol-naming-conventions.md
kozlovic: Both `*` and `>` are regular characters if they are not single character forming a token. So in theory `foo***bar.baz>.bat` is a valid subject.

doc/adr/0006-protocol-naming-conventions.md
scottf: correct the JS api prefix. I will change the
```
jetstream-prefix = prefix
```
to
```
js-internal-prefix = dollar (prefix dot)+
js-user-prefix     = (prefix dot)+
```

doc/adr/0006-protocol-naming-conventions.md
scottf: @derekcollison  / @ColinSullivan1 care to comment? I've been repeatedly told that the docs are the standard. I imagine this is not an issue with newer clients like java because the users are not aware of this. I have no idea what the go client currently supports. I'm guessing C++ supports this more lenient specification.  I wonder what other clients support.
Since this is more lenient than the documentation, it is certainly more adoptable than the other way.
We could note specifically that existing clients that support this behavior will continue to allow this, but interoperability between clients strongly suggest adherence to the strict definition, and that in the future the non-strict standard may be deprecated. We could also just leave the docs as is - do we know of users that are in fact using it and how worried are we that we break them?

doc/adr/0006-protocol-naming-conventions.md
```diff
@@ -0,0 +1,55 @@
...
+filename-safe    = (printable except dot, asterisk, lt, gt, colon, double-quote, fwd-slash, backslash, pipe, question-mark, ampersand)
+
+message-subject    = term (dot term | asterisk)* (dot gt)?
+reply-to           = term (dot term)*
+stream-name        = term
+queue-name         = term
+durable-name       = term
+js-internal-prefix = dollar (prefix dot)+
+js-user-prefix     = (prefix dot)+
+account-name       = (filename-safe)+ maximum 255 characters
```
ripienaar, line 54: Realistically speaking no this can not  be 255 characters, windows allow max 255 for the entire path name+path should not exceed 255 so a 255 character account name would not be usable on windows machines as servers

scottf: Correct, windows has issues. Windows server previous to 2016 maxes at 260 characters, after that it's 1024. Windows 10 defaults to 260, but can be configured to support 32k. I think it's reasonable to limit the account-name to something much shorter, maybe 16, 24 or 32 characters. Any opinions on this length?

doc/adr/0006-protocol-naming-conventions.md
derekcollison: I care quite a bit if we are potentially breaking existing deployments.

doc/adr/0006-protocol-naming-conventions.md
derekcollison: At least length of Nkey which is 57 or so? ~60 IIRC.

ripienaar:
@scottf whats happening with this?

scottf:
@derekcollison @ColinSullivan1 @kozlovic @aricart @wallyqs @variadico 
Last call for comments before I merge.

derekcollison:
I am good if the others are..

doc/adr/0006-protocol-naming-conventions.md
kozlovic: I am fine with above since you mention "should only be used", not "must only be used".

kozlovic (approved):
LGTM

PR #2223, opened 2021-05-18
Bad redelivered values on consumer state should not panic.

derekcollison:

/cc @nats-io/core


matthiashanel (approved):
LGTM - so with the panic gone, should we still look for the cause?

derekcollison:
Of course. But the issue is not that its a corrupt state per se.

PR #2224, opened 2021-05-18
[Fixed] Always initialize httpReqStats

BlizzTom:
 - [ ] Tests added

Resolves #

If you attempt to use the server http handlers it would panic unless you explicity called StartMonitoring

This isn't ideal to have a secondary http server running for those that are embedding nats and only want to host the http handlers on a pre-existing http server

Talked with @kozlovic via Slack about this
 
/cc @nats-io/core


server/server.go
```diff
@@ -2042,6 +2042,16 @@ func (s *Server) StartMonitoring() error {
 	// Snapshot server options.
 	opts := s.getOpts()
 
+	// Used to track HTTP requests
+	s.httpReqStats = map[string]uint64{
```
kozlovic, line 2046: I am confused. StartMonitoring() is calling s.startMonitoring() where this initialization was made. Can you explain to me what different this makes?

BlizzTom: `s.startMonitoring` is only called if an http or https port is configured.

BlizzTom:
Appears CI failure is related to a deadline on jetstream, not sure how this commit caused, is that failure transient? (worked on 1.16, not on 1.15)

server/server.go
kozlovic: So you don't define an HTTP port and you still register the public handlers. So now you want to do the init here so that you call StartMonitoring(), knowing that it won't call startMonitoring(), right?

If I understood correctly, then I am thinking that we could initialize this map when creating the server instance. That would be the common place and ok I guess, even if one does not use monitoring.

BlizzTom: That is the behavior we have in our Fork, I am happy to adjust it that way, just the constructor for server was pretty busy already, but makes sense from an API standpoint.

kozlovic: A simple:
```
s := &Server{
...
    eventIds:     nuid.New(),
    routesToSelf: make(map[string]struct{}),
    httpReqStats: make(map[string]uint64), // Used to track HTTP requests
}
```
would do I think...

BlizzTom: This wouldn't cover the use cases of Starting a server that wasn't called through New, is that ok?

BlizzTom: Updated to constructor init.

kozlovic: > This wouldn't cover the use cases of Starting a server that wasn't called through New, is that ok?

NewServer() is called in all situations. Even New() calls NewServer() but simply ignore the error (legacy API).

kozlovic (approved):
LGTM!

PR #2225, opened 2021-05-19
Under double import scenarios we could possibly map to the wrong subject.

derekcollison:

/cc @nats-io/core


matthiashanel (approved):
LGTM

PR #2227, opened 2021-05-19
Fixed issue that we would stage snapshot restores in temporary storage.

derekcollison:
This did not work when using our default docker image which does not have /tmp by default.

/cc @nats-io/core

### DIFF
--- a/server/filestore.go
+++ b/server/filestore.go
@@ -64,6 +64,12 @@ type FileConsumerInfo struct {
 	ConsumerConfig
 }
 
+// Default file and directory permissions.
+const (
+	defaultDirPerms  = os.FileMode(0750)
+	defaultFilePerms = os.FileMode(0640)
+)
+
 type fileStore struct {
 	mu      sync.RWMutex
 	state   StreamState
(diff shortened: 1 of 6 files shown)


server/jetstream_api.go
```diff
@@ -2629,9 +2630,16 @@ func (s *Server) processStreamRestore(ci *ClientInfo, acc *Account, cfg *StreamC
 
 	var resp = JSApiStreamRestoreResponse{ApiResponse: ApiResponse{Type: JSApiStreamRestoreResponseType}}
 
-	// FIXME(dlc) - Need to close these up if we fail for some reason.
-	// TODO(dlc) - Might need to make configurable or stream direct to storage dir.
-	tfile, err := ioutil.TempFile("", "jetstream-restore-")
+	snapDir := path.Join(js.config.StoreDir, snapStagingDir)
+	if _, err := os.Stat(snapDir); os.IsNotExist(err) {
+		if err := os.MkdirAll(snapDir, 0755); err != nil {
```
ripienaar, line 2635: `0750`, since its a new dir lets start strict rather than clean up later

ripienaar (commented):
LGTM

server/jetstream_api.go
derekcollison: Should I change all of them? I know Phil wants that..

derekcollison: ok made constants and changed all non-test versions.

server/filestore.go
```diff
@@ -64,6 +64,12 @@ type FileConsumerInfo struct {
 	ConsumerConfig
 }
 
+// Default file and directory permissions.
+const (
+	defaultDirPerms  = os.FileMode(0750)
```
kozlovic, line 69: Good to make those a constant so we can change in one place. But if we are going to do a 2.2.5, did we not say for 2.2.4 that we were not comfortable changing from 755 to 750 for an update release?

derekcollison: At last minute for 2.2.4 yes but been thinking through it for awhile now and I think its ok. /cc @philpennock

PR #2231, opened 2021-05-20
Changed remove peer logic to wait for RemovePeer entry to be replicated.

derekcollison:
Also fixed another flapping test.

/cc @nats-io/core


matthiashanel (approved):
LGTM

PR #2233, opened 2021-05-20
[added] pinned_cert option to tls block hex(sha256(spki))

matthiashanel:

To create the hex(sha256(spki)) identifier:
openssl x509 -noout -pubkey -in certfile.pem | openssl pkey -pubin -outform DER | openssl dgst -sha256

### DIFF
--- a/server/auth.go
+++ b/server/auth.go
@@ -14,10 +14,12 @@
 package server
 
 import (
+	"crypto/sha256"
 	"crypto/tls"
 	"crypto/x509/pkix"
 	"encoding/asn1"
 	"encoding/base64"
+	"encoding/hex"
 	"fmt"
 	"net"
 	"net/url"
@@ -185,6 +187,19 @@ func (s *Server) checkAuthforWarnings() {
 		// Warning about using plaintext passwords.
 		s.Warnf("Plaintext passwords detected, use nkeys or bcrypt")
 	}
+	// only warn about client connections others use bidirectional TLS
+	if len(s.opts.TLSPinnedCerts) > 0 && !(s.opts.TLSVerify || s.opts.TLSMap) {
+		s.Warnf("Pinned Certs specified but no verify or verify_and_map that would require presenting a client cert")
+	}
+	if len(s.opts.Websocket.TLSPinnedCerts) > 0 && !(s.opts.Websocket.TLSMap) {
+		s.Warnf("Websocket Pinned Certs specified but no verify_and_map that would require presenting a client cert")
+	}
+	if len(s.opts.MQTT.TLSPinnedCerts) > 0 && !(s.opts.MQTT.TLSMap) {
+		s.Warnf("MQTT Pinned Certs specified but no verify_and_map that would require presenting a client cert")
+	}
+	if len(s.opts.LeafNode.TLSPinnedCerts) > 0 && !(s.opts.LeafNode.TLSMap) {
+		s.Warnf("Leaf Pinned Certs specified but verify_and_map that would require presenting a client cert")
+	}
 }
 
 // If Users or Nkeys options have definitions without an account defined,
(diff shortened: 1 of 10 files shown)


server/opts.go
```diff
@@ -3444,6 +3460,24 @@ func parseTLS(v interface{}, isClientCtx bool) (t *TLSConfigOpts, retErr error)
 				at = mv
 			}
 			tc.Timeout = at
+		case "pinned_certs":
+			ra, ok := mv.([]interface{})
+			if !ok {
+				return nil, &configErr{tk, "error parsing tls config, expected 'pinned_certs' to be a list of sha256(DER encoded SubjectPublicKeyInfo)"}
```
philpennock, line 3466: This could usefully state hex-encoded no-colons

server/opts.go
```diff
@@ -3444,6 +3460,24 @@ func parseTLS(v interface{}, isClientCtx bool) (t *TLSConfigOpts, retErr error)
...
+			}
+			if len(ra) != 0 {
+				wl := PinnedCertSet{}
+				re := regexp.MustCompile("^[A-Fa-f0-9]{64}$")
+				for _, r := range ra {
+					tk, r := unwrapValue(r, &lt)
+					entry := strings.ToLower(r.(string))
+					if !re.MatchString(entry) {
+						return nil, &configErr{tk, fmt.Sprintf("error parsing tls config, 'pinned_certs' key %s does not look like sha256", entry)}
```
philpennock, line 3475: Again, `hex-encoded`.  base64url is used in a few modern standards, so we should give this hint.

server/opts_test.go
```diff
@@ -890,6 +890,94 @@ func TestNkeyUsersConfig(t *testing.T) {
 	}
 }
 
+// Test new nkey users
+func TestTlsWhitelist(t *testing.T) {
```
philpennock, line 894: We should probably get in the habit of writing things like `TestTlsAllowList` for new code.

test/tls_test.go
```diff
@@ -1935,3 +1935,48 @@ func TestTLSClientSVIDAuth(t *testing.T) {
...
+
+	nc, err := nats.Connect(srv.ClientURL(), opts...)
+	if err == nil {
+		nc.Close()
+		t.Fatalf("Expected error trying to connect without a certificate in pinned_certs")
+	}
+
+	ioutil.WriteFile(confFileName, []byte(fmt.Sprintf(tmpl, "bf6f821f09fde09451411ba3b42c0f74727d61a974c69fd3cf5257f39c75f0e9")), 0660)
+	if err := srv.Reload(); err != nil {
+		t.Fatalf("on releaod got %v", err)
```
philpennock, line 1974: s/releaod/reload/

philpennock (approved):
Very minor nits

server/auth.go
```diff
@@ -344,6 +359,25 @@ func (s *Server) isClientAuthorized(c *client) bool {
 	return true
 }
 
+// returns false if the client needs to be disconnected
+func (s *Server) matchesPinnedCert(c *client, tlsPinnedCerts PinnedCertSet) bool {
+	if tlsPinnedCerts == nil {
+		return true
+	}
+	tlsState := c.GetTLSConnectionState()
```
kozlovic, line 367: I wonder if this is correct. Basically you assume that because there is a tlsPinnedCerts map provided (for any of the accept ports that we have, clients, routes, etc..), then the connection has to be TLS, and if not, then it failed this test.
I believe that we have a feature that accepts TLS or non TLS on the same port (see option AllowNonTLS, which applies only to clients at the moment, but that includes MQTT). So if we are here, and this is a CLIENT, and tlsState  is nil, this is probably because of AllowNonTLS, so it should be accepted?

server/opts_test.go
```diff
@@ -890,6 +890,94 @@ func TestNkeyUsersConfig(t *testing.T) {
...
+	tls {
+		cert_file: "./configs/certs/server.pem"
+		key_file: "./configs/certs/key.pem"
+		# Require a client certificate and map user id from certificate
+		verify: true
+		pinned_certs: ["7f83b1657ff1fc53b92dc18148a1d65dfc2d4b1fa3d677284addd200126d9069", 
+			"a8f407340dcc719864214b85ed96f98d16cbffa8f509d9fa4ca237b7bb3f9c32"]
+	}
+	cluster {
+		listen localhost:4223
```
kozlovic, line 905: I know that you are not running the server here, so ports don't really matter, but probably use `-1`. I noticed that you have repeated `4224`.

server/auth.go
matthiashanel: I mean if you specify a map, then you should fail if no certificate is to be had.

server/opts_test.go
matthiashanel: will fix

server/opts.go
matthiashanel: fixed

server/opts.go
matthiashanel: fixed

server/opts_test.go
matthiashanel: name reflects the option tested now

test/tls_test.go
matthiashanel: fixed

server/auth.go
kozlovic: I guess you right. If the user configures this, then it is likely that they want to force the use of a certificate.

kozlovic (approved):
LGTM

PR #2234, opened 2021-05-20
Fix a data race.

derekcollison:

/cc @nats-io/core


server/stream.go
```diff
@@ -1569,7 +1569,7 @@ func (mset *stream) setupMirrorConsumer() error {
 			}
 			mset.setMirrorErr(ccr.Error)
 		case <-time.After(10 * time.Second):
-			mset.unsubscribe(crSub)
+			mset.unsubscribeUnlocked(crSub)
```
ripienaar, line 1572: shouldnt the one on 1502 be the same?

derekcollison: Fixed, thanks.

ripienaar:
LGTM

PR #2236, opened 2021-05-20
[FIXED] MQTT: session fails if the number servers below cluster size

### LINKED ISSUE #2226: enable the MQTT function
I use three test machines to run nats and try to enable the MQTT function, but when simulating a server down (that is, exiting the program and running), it is easy to get the following warning:
JetStream cluster stream'$G> $MQTT_sess_KwEmus3X' has NO quorum, stalled.
And the mqtt client is also easy to be unable to connect to nats all the time.

kozlovic:
Say with a cluster of 3, all MQTT assets are created with a replicas
of 3. However, when a server is shutdown, then any new MQTT client
will fail to connect because we try to create a session stream
with R(3), which leads to insufficient resources.

The longer term solution should be for the server to allow the
creation of an asset with a R() value that is bigger than the
current number of running servers as long as there is quorum.

For now, we will reduce the R() value for the sessions if we get
an "insufficient resources" error.

Note that the other assets still will use the compute R() based
on cluster size. So the first time that a client on a given
account is started, we will still need to have R() == cluster size
(at least for R(3)).

Partially resolves #2226

### DIFF
--- a/server/mqtt.go
+++ b/server/mqtt.go
@@ -1880,12 +1880,21 @@ func (as *mqttAccountSessionManager) createOrRestoreSession(clientID string, opt
 		accName := jsa.c.acc.GetName()
 		return nil, false, fmt.Errorf("%s for account %q, session %q: %v", errTxt, accName, clientID, err)
 	}
+CREATE_STREAM:
 	// Send a request to create the stream for this session.
 	si, err := jsa.createStream(cfg)
-	// If there is an error and not simply "already used" (which means that the
-	// stream already exists) then we fail.
-	if isErrorOtherThan(err, ErrJetStreamStreamAlreadyUsed) {
-		return formatError("create session stream", err)
+	if err != nil {
+		// Check for insufficient resources. If that is the case, and if possible, try
+		// again with a lower replicas value.
+		if cfg.Replicas > 1 && err.Error() == jsInsufficientErr.Description {
+			cfg.Replicas--
+			goto CREATE_STREAM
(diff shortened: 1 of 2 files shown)


server/mqtt.go
```diff
@@ -1880,12 +1880,21 @@ func (as *mqttAccountSessionManager) createOrRestoreSession(clientID string, opt
...
 	si, err := jsa.createStream(cfg)
-	// If there is an error and not simply "already used" (which means that the
-	// stream already exists) then we fail.
-	if isErrorOtherThan(err, ErrJetStreamStreamAlreadyUsed) {
-		return formatError("create session stream", err)
+	if err != nil {
+		// Check for insufficient resources. If that is the case, and if possible, try
+		// again with a lower replicas value.
+		if cfg.Replicas > 1 && err.Error() == jsInsufficientErr.Description {
+			cfg.Replicas--
```
derekcollison, line 1890: Do you have a way to break out?

derekcollison (changes requested):
We should make sure to have a way to break out of that loop.

server/mqtt.go
kozlovic: If cfg.Replicas reaches 1, then we fall back to the next check that says, if the error is not "Stream already used" then we fail.

derekcollison: Apologies.. Yes see that now..

derekcollison (approved):
LGTM

PR #2240, opened 2021-05-22
OCSP Stapling

wallyqs:
Adds basic OCSP Stapling support in the server for client connections.  To enable, the server has to start with the following:

```hcl
ocsp: true

tls { ... }
```

When OCSP is enabled this way, in case a certificate has a 'status_request' // Must Staple flag toggled, then it will make an 
HTTP request to the OCSP remote url and fetch staples that will be cached on disk and serve the staples to the clients so that they can confirm the response.  The TLS certificates can change at anytime and OCSP stapling is reloading aware.

To store OCSP Staples to disk, a store dir can be configured as follows:

```hcl
store_dir = "./store-dir"
```

Or in case there is a store dir for JetStream that will be used instead:

```hcl
jetstream {
  store_dir = "./store-dir"
}
```

The OCSP staples will be stored at a created directory underneath like `$store_dir/ocsp`.

/cc @nats-io/core


server/ocsp.go
```diff
@@ -103,6 +116,39 @@ func (oc *OCSPMonitor) getCacheStatus() ([]byte, *ocsp.Response) {
...
+	opts := oc.srv.getOpts()
+	storeDir := opts.StoreDir
+	if storeDir == _EMPTY_ {
+		return nil, nil, fmt.Errorf("store_dir not set")
+	}
+
+	key := fmt.Sprintf("%x", sha256.Sum256(oc.Leaf.Raw))
+
+	oc.mu.Lock()
+	raw, err := ioutil.ReadFile(filepath.Join(storeDir, key))
```
ripienaar, line 129: missing `defaultOCSPStoreDir` in the join

ripienaar (changes requested):

server/ocsp.go
wallyqs: Thanks, should be fixed now

philpennock (approved):
I confess, I'm a little sad that we're not following the `state.directory` pattern.  Ah well.

server/ocsp.go
```diff
@@ -103,6 +116,39 @@ func (oc *OCSPMonitor) getCacheStatus() ([]byte, *ocsp.Response) {
...
+	resp, err := ocsp.ParseResponse(raw, oc.Issuer)
+	if err != nil {
+		return nil, nil, err
+	}
+	if err := validOCSPResponse(resp); err != nil {
+		return nil, nil, err
+	}
+
+	// Cache the response.
+	oc.mu.Lock()
```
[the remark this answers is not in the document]
wallyqs: There is only one OCSP Stapling monitor per certificate and they are discarded whenever there is a reload so there would not overlap.

server/ocsp.go
```diff
@@ -103,6 +116,39 @@ func (oc *OCSPMonitor) getCacheStatus() ([]byte, *ocsp.Response) {
...
 }
 
+func (oc *OCSPMonitor) getLocalStatus() ([]byte, *ocsp.Response, error) {
+	opts := oc.srv.getOpts()
+	storeDir := opts.StoreDir
+	if storeDir == _EMPTY_ {
+		return nil, nil, fmt.Errorf("store_dir not set")
+	}
+
+	key := fmt.Sprintf("%x", sha256.Sum256(oc.Leaf.Raw))
```
wallyqs, line 126:
```suggestion
	// This key must be based upon the current full certificate, not the public key,
	// so MUST be on the full raw certificate and not an SPKI or other reduced form.
	key := fmt.Sprintf("%x", sha256.Sum256(oc.Leaf.Raw))
```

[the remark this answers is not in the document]
wallyqs: Thanks, ok applied the comment.

ColinSullivan1 (approved):
LGTM.

server/ocsp.go
```diff
@@ -0,0 +1,542 @@
...
+			certFile, t, nextRun,
+		)
+	} else if err == nil && shutdownOnRevoke {
+		// If resp.Status is ocsp.Revoked, ocsp.Unknown, or any other value.
+		s.Errorf("Found existing OCSP status for certificate at '%s': %s", certFile, ocspStatusString(resp.Status))
+		s.Shutdown()
+		return
+	}
+
+	for s.Running() {
```
kozlovic, line 259: Since you select on `quitCh` that is from the server, not sure you need this condition here.

server/server.go
```diff
@@ -1467,6 +1470,48 @@ func (s *Server) fetchAccount(name string) (*Account, error) {
...
+func (s *Server) setupOCSPStapleStoreDir() error {
+	opts := s.getOpts()
+	storeDir := opts.StoreDir
+	if storeDir == _EMPTY_ {
+		s.Warnf("OCSP Stapling disk cache is disabled (missing 'store_dir')")
+		return nil
+	}
+	storeDir = filepath.Join(storeDir, defaultOCSPStoreDir)
+	if stat, err := os.Stat(storeDir); os.IsNotExist(err) {
+		if err := os.MkdirAll(storeDir, os.FileMode(0750)); err != nil {
```
kozlovic, line 1482: I think there is a constant now in the server (for JetStream) to use instead: `defaultDirPerms  = os.FileMode(0750)`. Not sure if you have rebased recently.

kozlovic (changes requested):

server/server.go
wallyqs: ah yes now that I rebased can use this

server/ocsp.go
wallyqs: sounds good, will remove

kozlovic (approved):
LGTM

PR #2241, opened 2021-05-23
Fix for corrupting a message when extending an existing header.

derekcollison:
We had a report of corrupt message payloads when going across leafnodes between streams that were sourced from one another.
We were incorrectly using the underlying buffer when a header already existed.

/cc @nats-io/core


kozlovic (approved):
LGTM

Issue #2242, opened 2021-05-24
expected sequence does not match store when using Nats-Msg-Id header


krisdaniels:
When using a msg id to prevent duplicates, an error is always returned when sending a non duplicate after having received a duplicate ack result  :

```
docker run -ti --rm --network host natsio/nats-box

nats str add ORDERS --subjects "ORDERS.*" --ack --max-msgs=-1 --max-bytes=-1 --max-age=1y --storage file --retention limits --max-msg-size=-1 --discard old --dupe-window="0s" --replicas 2
nats con add ORDERS NEW --filter ORDERS.new --ack explicit --pull --deliver all --sample 100 --max-deliver 20 --replay instant --max-pending 20

nats req -H Nats-Msg-Id:1 ORDERS.new hello1 
nats req -H Nats-Msg-Id:1 ORDERS.new hello2 # duplicate
nats req -H Nats-Msg-Id:2 ORDERS.new hello3 # new msg always fails after duplicate
nats req -H Nats-Msg-Id:2 ORDERS.new hello3 # retry 1 fails
nats req -H Nats-Msg-Id:2 ORDERS.new hello3 # retry 2 succeeds
```
First message + duplicate :
```
# nats req -H Nats-Msg-Id:1 ORDERS.new hello1 

05:16:03 Sending request on "ORDERS.new"
05:16:03 Received on "_INBOX.r0yfSLDniVtFWbJXzrdkyZ.OGFSlLFI" rtt 1.003804ms
{"stream":"ORDERS","seq":1}

# nats req -H Nats-Msg-Id:1 ORDERS.new hello2 # duplicate

05:16:03 Sending request on "ORDERS.new"
05:16:03 Received on "_INBOX.Gu8EQT3FXwoZTn2pXW8oRb.MPEuuDBY" rtt 973.104µs
{"stream":"ORDERS","seq":1,"duplicate": true}
```

First attempt of new message after duplicate
```
# nats req -H Nats-Msg-Id:2 ORDERS.new hello3 # new msg always fails after duplicate

05:16:03 Sending request on "ORDERS.new"
05:16:03 Received on "_INBOX.9MXKkqfTzruzga2f9zztXo.nRG0SrMJ" rtt 1.058504ms
{"error":{"code":503,"description":"expected sequence does not match store"},"stream":"ORDERS","seq":0}
```

Second attempt of new message after duplicate
```
# nats req -H Nats-Msg-Id:2 ORDERS.new hello3 # retry 1 fails

[...]

PR #2245, opened 2021-05-24
[FIXED] #2242.

derekcollison:
When we had a duplicate detected in R>1 mode we set the skip sequence indicator but were not using that when dealing with underlying store.

Resolves: #2242

/cc @nats-io/core


ripienaar (approved):
LGTM

kozlovic (approved):
LGTM, but there is a conflict, likely from the merge of previous PR that need to be addressed.

kozlovic (approved):
LGTM

Issue #2243, opened 2021-05-21
Messages not deleted from `WorkQueuePolicy` stream when replicas > 1 


danpoland:
It may be that I misunderstand the meaning of the `Replicas` configuration for a stream. For demo purposes, I have a NATS cluster with two servers running JetStream. 

First I created a stream called `FOO` using the `Work Queue` retention policy and set the number of replicas to 2. 
```
Stream FOO was created

Information for Stream FOO created 2021-05-21T15:32:16-04:00

Configuration:

             Subjects: FOO.*
     Acknowledgements: true
            Retention: File - WorkQueue
             Replicas: 2
       Discard Policy: Old
     Duplicate Window: 2m0s
     Maximum Messages: unlimited
        Maximum Bytes: unlimited
          Maximum Age: 0.00s
 Maximum Message Size: unlimited
    Maximum Consumers: unlimited


Cluster Information:

                 Name: local
               Leader: nats1
              Replica: nats2, current, seen 0.00s ago

State:

             Messages: 0
                Bytes: 0 B
             FirstSeq: 0
              LastSeq: 0
     Active Consumers: 0
```

I then created a pull-based consumer called TEST, that pulls from the `FOO.test` subject.
```
Information for Consumer FOO > TEST created 2021-05-21T15:36:52-04:00

Configuration:

        Durable Name: TEST
           Pull Mode: true
      Filter Subject: FOO.test
         Deliver All: true
          Ack Policy: Explicit
            Ack Wait: 30s
       Replay Policy: Instant
     Max Ack Pending: 20,000
        Flow Control: false

Cluster Information:

                Name: local
              Leader: nats2
             Replica: nats1, current, seen 0.01s ago

State:

   Last Delivered Message: Consumer sequence: 0 Stream sequence: 0
     Acknowledgment floor: Consumer sequence: 0 Stream sequence: 0
         Outstanding Acks: 0 out of maximum 20000
     Redelivered Messages: 0
     Unprocessed Messages: 0
```

I then published a single message to the FOO.test subject and confirmed it was sitting in the stream and was unprocessed by the consumer.

```
❯ nats s report
Obtaining Stream stats

+----------------------------------------------------------------------------------+
|                                  Stream Report                                   |
+--------+---------+-----------+----------+-------+------+---------+---------------+
| Stream | Storage | Consumers | Messages | Bytes | Lost | Deleted | Replicas      |
+--------+---------+-----------+----------+-------+------+---------+---------------+
[...]

PR #2246, opened 2021-05-24
Fix for #2243. We were not allowing replicated acks processing for work queues.

derekcollison:
Resolves: #2243 

/cc @nats-io/core


ripienaar (approved):
LGTM

kozlovic (approved):
LGTM

PR #2247, opened 2021-05-24
[changed] pinned certs to check the server connected to as well

matthiashanel:
on reload clients with removed pinned certs will be disconnected.
The check happens only on tls handshake now.

### DIFF
--- a/server/auth.go
+++ b/server/auth.go
@@ -187,19 +187,6 @@ func (s *Server) checkAuthforWarnings() {
 		// Warning about using plaintext passwords.
 		s.Warnf("Plaintext passwords detected, use nkeys or bcrypt")
 	}
-	// only warn about client connections others use bidirectional TLS
-	if len(s.opts.TLSPinnedCerts) > 0 && !(s.opts.TLSVerify || s.opts.TLSMap) {
-		s.Warnf("Pinned Certs specified but no verify or verify_and_map that would require presenting a client cert")
-	}
-	if len(s.opts.Websocket.TLSPinnedCerts) > 0 && !(s.opts.Websocket.TLSMap) {
-		s.Warnf("Websocket Pinned Certs specified but no verify_and_map that would require presenting a client cert")
-	}
-	if len(s.opts.MQTT.TLSPinnedCerts) > 0 && !(s.opts.MQTT.TLSMap) {
-		s.Warnf("MQTT Pinned Certs specified but no verify_and_map that would require presenting a client cert")
-	}
-	if len(s.opts.LeafNode.TLSPinnedCerts) > 0 && !(s.opts.LeafNode.TLSMap) {
-		s.Warnf("Leaf Pinned Certs specified but verify_and_map that would require presenting a client cert")
-	}
 }
 
 // If Users or Nkeys options have definitions without an account defined,
@@ -360,7 +347,7 @@ func (s *Server) isClientAuthorized(c *client) bool {
 }
 
 // returns false if the client needs to be disconnected
-func (s *Server) matchesPinnedCert(c *client, tlsPinnedCerts PinnedCertSet) bool {
+func (c *client) matchesPinnedCert(tlsPinnedCerts PinnedCertSet) bool {
 	if tlsPinnedCerts == nil {
 		return true
 	}
@@ -407,14 +394,12 @@ func (s *Server) processClientOrLeafAuthentication(c *client, opts *Options) boo
 		noAuthUser string
 	)
 	tlsMap := opts.TLSMap
-	tlsPinnedCerts := opts.TLSPinnedCerts
 	if c.kind == CLIENT {
 		switch c.clientType() {
 		case MQTT:
 			mo := &opts.MQTT
 			// Always override TLSMap.
 			tlsMap = mo.TLSMap
-			tlsPinnedCerts = mo.TLSPinnedCerts
 			// The rest depends on if there was any auth override in
 			// the mqtt's config.
 			if s.mqtt.authOverride {
@@ -428,7 +413,6 @@ func (s *Server) processClientOrLeafAuthentication(c *client, opts *Options) boo
 			wo := &opts.Websocket
 			// Always override TLSMap.
 			tlsMap = wo.TLSMap
-			tlsPinnedCerts = wo.TLSPinnedCerts
 			// The rest depends on if there was any auth override in
 			// the websocket's config.
 			if s.websocket.authOverride {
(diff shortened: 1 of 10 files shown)


server/auth.go
```diff
@@ -441,11 +425,6 @@ func (s *Server) processClientOrLeafAuthentication(c *client, opts *Options) boo
 		}
 	} else {
 		tlsMap = opts.LeafNode.TLSMap
-		tlsPinnedCerts = opts.LeafNode.TLSPinnedCerts
-	}
-	if !s.matchesPinnedCert(c, tlsPinnedCerts) {
-		s.mu.Unlock()
-		return false
 	}
 	if !authRequired {
```
kozlovic, line 429: Would you mind then moving this back to where it was prior to the first PR you made about pinned_certs?

server/leafnode.go
```diff
@@ -2360,7 +2360,7 @@ func (s *Server) leafNodeResumeConnectProcess(c *client) {
 			rURL := remote.getCurrentURL()
 
 			// Perform the client-side TLS handshake.
-			if resetTLSName, err := c.doTLSClientHandshake("leafnode", rURL, tlsConfig, tlsName, tlsTimeout); err != nil {
+			if resetTLSName, err := c.doTLSClientHandshake("leafnode", rURL, tlsConfig, tlsName, tlsTimeout, c.srv.getOpts().TLSPinnedCerts); err != nil {
```
kozlovic, line 2363: It is "client side", but that simply means that this is the soliciting side. I think that you should still provide the leafnodes pinned certs here, no?

server/reload.go
```diff
@@ -705,6 +705,46 @@ func (s *Server) Reload() error {
...
+			if c.kind == kind && !c.matchesPinnedCert(set) {
+				disconnectClients = append(disconnectClients, c)
+			}
+		}
+	}
+	if !reflect.DeepEqual(newOpts.TLSPinnedCerts, curOpts.TLSPinnedCerts) {
+		checkClients(CLIENT, s.clients, newOpts.TLSPinnedCerts)
+	}
+	if !reflect.DeepEqual(newOpts.MQTT.TLSPinnedCerts, curOpts.MQTT.TLSPinnedCerts) {
+		checkClients(MQTT, s.clients, newOpts.MQTT.TLSPinnedCerts)
```
kozlovic, line 722: That won't work... MQTT and Websocket clients are still c.kind == CLIENT. For c.kind == CLIENT, then you could either check for isWebsocket() or isMqtt(), or invoke c.clientType(). But again, only when c.kind == CLIENT.

kozlovic (changes requested):

server/leafnode.go
matthiashanel: yes, fixed

server/auth.go
matthiashanel: fixed

server/reload.go
```diff
@@ -705,6 +705,46 @@ func (s *Server) Reload() error {
...
+	disconnectClients := []*client{}
+	checkClients := func(kind int, clients map[uint64]*client, set PinnedCertSet) {
+		for _, c := range clients {
+			if (c.kind == kind || (c.kind == CLIENT && c.clientType() == kind)) && !c.matchesPinnedCert(set) {
+				disconnectClients = append(disconnectClients, c)
+			}
+		}
+	}
+	if !reflect.DeepEqual(newOpts.TLSPinnedCerts, curOpts.TLSPinnedCerts) {
+		checkClients(NATS, s.clients, newOpts.TLSPinnedCerts)
```
kozlovic, line 719: For actual CLIENT connections, I am not sure this helper is really nice because we are going to iterate over all clients for as many "type" that we have (NATS, MQTT and WS, even if we have none of clients being actually MQTT and/or WS).

server/reload.go
```diff
@@ -705,6 +705,46 @@ func (s *Server) Reload() error {
 	if err := s.reloadOptions(curOpts, newOpts); err != nil {
 		return err
 	}
+
+	s.mu.Lock()
+	disconnectClients := []*client{}
+	checkClients := func(kind int, clients map[uint64]*client, set PinnedCertSet) {
+		for _, c := range clients {
+			if (c.kind == kind || (c.kind == CLIENT && c.clientType() == kind)) && !c.matchesPinnedCert(set) {
```
kozlovic, line 713: I had a moment when I thought that this was broken because `NATS`, `MQTT`, etc.. are separate constants that would clash would the actual c.kind (CLIENT, ROUTER, etc..). But since you call `checkClients()` with specific maps that contain only the actual kind, it's fine I guess...

And we are lucky that CLIENT is 0 and NATS starts at 1. But this code is a bit risky. I would revisit. Also make it a separate function (especially if we change it to be a bit more safe, likely with more code) instead of embedded here.

kozlovic (changes requested):
Still not happy with `checkClients()`. Although I think it is ok now, it is a bit risky and only works right now because CLIENT happens to be 0 and NATS/MQTT/WS happens to be 1/2/3. Would CLIENT be moved in the definition, that code would break.

server/reload.go
matthiashanel: fixed

server/client.go
```diff
@@ -5006,6 +5006,13 @@ func (c *client) doTLSHandshake(typ string, solicit bool, url *url.URL, tlsConfi
 		return resetTLSName, ErrConnectionClosed
 	}
 
+	if !c.matchesPinnedCert(pCerts) {
```
kozlovic, line 5009: We could have reduce a bit the duplication of code:
```
if err = conn.Handshake(); err != nil {
    ...
    } else {
	c.Errorf("TLS %s handshake error: %v", typ, err)		
}
if err != nil || !c.matchesPinnedCert(pCerts) {
    c.closeConnection(TLSHandshakeError)
    // Grab the lock before returning since the caller was holding the lock on entry
    c.mu.Lock()
    // Returning any error is fine. Since the connection is closed ErrConnectionClosed
    // is appropriate.
    return resetTLSName, ErrConnectionClosed
}
```

server/reload.go
matthiashanel: processing them separately now.
I moved all that code into a separate function.
Keeping the utility closure.

server/client.go
```diff
@@ -4992,6 +4992,11 @@ func (c *client) doTLSHandshake(typ string, solicit bool, url *url.URL, tlsConfi
 				}
 			}
 		}
+	} else if !c.matchesPinnedCert(pCerts) {
+		err = ErrConnectionClosed
+	}
+
+	if err != nil {
 		if kind == CLIENT {
```
kozlovic, line 5000: I would have kept this in the Handshake() error statement, unless you want to have matchedPinnedCert() failure to also be reported as a TLS handshake error, but then, maybe you want to introduce a new error that indicates that it is due to pinned_cert not matching..

kozlovic (changes requested):
Some misunderstanding regarding TLS handshake error report. Would have left this where it was, or then add a new error for failure to match pinned cert?

server/client.go
matthiashanel: have a separate error now

kozlovic (approved):
LGTM

PR #2249, opened 2021-05-24
Fix for varz JetStream usage showing total account usage vs server usage.

derekcollison:

/cc @nats-io/core


matthiashanel (approved):
LGTM

PR #2251, opened 2021-05-26
ocsp: Add OCSP Stapling support for all type of connections

wallyqs:
Adds OCSP Stapling support for all type of connections when the certificate used has a the must staple flag and an OCSP Server URL.

Signed-off-by: Jaime Piña <jaime@synadia.com>


server/ocsp.go
```diff
@@ -400,11 +417,52 @@ func (srv *Server) NewOCSPMonitor(tc *tls.Config) (*tls.Config, *OCSPMonitor, er
...
 		}
 
-		// GetClientCertificate returns a certificate that's presented to a
-		// server.
-		tc.GetClientCertificate = func(info *tls.CertificateRequestInfo) (*tls.Certificate, error) {
-			return &cert, nil
+		// Check whether need to verify staples from a client connection depending on the type.
+		switch kind {
+		case typeStringMap[ROUTER], typeStringMap[GATEWAY], typeStringMap[LEAF]:
+			tc.VerifyConnection = func(s tls.ConnectionState) error {
```
wallyqs, line 423: This is so that the route client connection can verify that the server has a valid OCSP Staple.

server/ocsp.go
```diff
@@ -400,11 +417,52 @@ func (srv *Server) NewOCSPMonitor(tc *tls.Config) (*tls.Config, *OCSPMonitor, er
...
+				return nil
+			}
+
+			// When server makes a client connection, need to also present an OCSP Staple.
+			tc.GetClientCertificate = func(info *tls.CertificateRequestInfo) (*tls.Certificate, error) {
+				raw, _, err := mon.getStatus()
+				if err != nil {
+					return nil, err
+				}
+				cert.OCSPStaple = raw
```
wallyqs, line 455: We have to do this so that when connecting to a server, a client sends its staple which will be verified by the server attempting the connection /cc @philpennock

server/ocsp.go
```diff
@@ -400,11 +417,52 @@ func (srv *Server) NewOCSPMonitor(tc *tls.Config) (*tls.Config, *OCSPMonitor, er
...
-		// server.
-		tc.GetClientCertificate = func(info *tls.CertificateRequestInfo) (*tls.Certificate, error) {
-			return &cert, nil
+		// Check whether need to verify staples from a client connection depending on the type.
+		switch kind {
+		case typeStringMap[ROUTER], typeStringMap[GATEWAY], typeStringMap[LEAF]:
+			tc.VerifyConnection = func(s tls.ConnectionState) error {
+				oresp := s.OCSPResponse
+				if oresp == nil {
+					return fmt.Errorf("missing OCSP Staple")
```
philpennock, line 426: All of these error messages should surely identify which staple is missing, making clear that it's to do with staples from a peer, rather than the local cert itself.

server/reload.go
```diff
@@ -1294,31 +1294,70 @@ func (s *Server) reloadOCSP() error {
...
 	}
 
+	type ocspState struct {
+		config     *tls.Config
+		wasEnabled bool
+	}
+	withOCSP := make(map[string]ocspState)
+
+	// Start OCSP Stapling for client connections.
+	if config := sopts.TLSConfig; config != nil {
```
philpennock, line 1308: These guards look like they're requiring OCSP staples if _any_ options are enabled for TLS for these items, rather than if OCSP is enabled.  Am I misreading?

test/ocsp_test.go
```diff
@@ -964,6 +964,264 @@ func TestOCSPReloadRotateTLSCertEnableMustStaple(t *testing.T) {
...
+	if err != nil {
+		t.Fatal(err)
+	}
+	checkClusterFormed(t, srvA, srvB)
+
+	// Revoke the seed server cluster certificate, following servers will not be able to verify connection.
+	setOCSPStatus(t, addr, "configs/certs/ocsp/server-status-request-url-02-cert.pem", ocsp.Revoked)
+
+	// Original set of servers still can communicate to each other, even though the cert has been revoked.
+	// NOTE: Should we unplug from the cluster in case our server is revoke and OCSP policy is always or must?
```
wallyqs, line 1082: In case a certificate has a staple that has been revoked at some point, currently it is not disconnecting itself from the other routes, should we make this the default maybe? /cc @philpennock

philpennock (commented):
How are the test-data certs generated?  What needs to be done to renew those before they expire?

server/ocsp.go
wallyqs: We can infer some of this from the debug lines from the server, for example here is how it would like when a peer has the bad OCSP staple.

```
[91176] [DBG] 127.0.0.1:52753 - rid:3 - Starting TLS route client handshake
[91176] [ERR] 127.0.0.1:52753 - rid:3 - TLS route handshake error: bad status for OCSP Staple
[91176] [INF] 127.0.0.1:52753 - rid:3 - Router connection closed: TLS Handshake Failure
```
But debug is silent by default so we can do maybe this

```
fmt.Errorf("%s client missing OCSP Staple", kind)
```

So that it reads

```
fmt.Errorf("route client missing OCSP Staple", kind)
```
```

server/reload.go
wallyqs: Good point, we could use a better name for this variable. The check whether OCSP Stapling is enabled or not does not happen until `NewOCSPMonitor` is called and checks each one of the certificates.

wallyqs:
> How are the test-data certs generated? What needs to be done to renew those before they expire?

@variadico has the script to generate the certs for OCSP, maybe we can document or add tools somewhere? The embedded OCSP Server for the tests is also quite useful

philpennock:
> > How are the test-data certs generated? What needs to be done to renew those before they expire?
> 
> @variadico has the script to generate the certs for OCSP, maybe we can document or add tools somewhere? The embedded OCSP Server for the tests is also quite useful

When I needed to replace the certs in the `nats-top` repo, I scripted it and added the script there.  It might be more complicated than needed: I picked cfssl, which I see has also been used in some other places, but I structured it a bit.  I don't much care if my script is used or something else, I just would like to see complicated hierarchies of interdependent test-data certs, with expiration dates, managed with a script so that they can be replaced years from now when everyone has forgotten that OCSP is even a thing we implement.  :)


https://github.com/nats-io/nats-top/blob/master/util/test/make-new-certs.sh

test/ocsp_test.go
philpennock: This should be _consistent_ with what we do when a TLS certificate expires.  If a TLS certificate expires, do we drop the connections?  If yes, then an expired OCSP staple should drop too.  If no, then OCSP should probably not be more strict.

An architecture decision on "NATS: do TLS expiration dates apply only at connection establishment or provide an upper bound on when data can be passed under that certificate" should be made, and then enforced consistently.

I think we should probably use level-enforcement not edge-enforcement; I haven't checked but think we're probably edge-enforcement for TLS cert expiration now; so in that case I think commit the OCSP work as-is, consistent with the TLS enforcement, but separately have a new ADR to document how we think TLS certs should be enforced throughout the NATS ecosystem, including in clients across multiple languages, and then in separate work move to enforce that consistent.

server/ocsp.go
philpennock: 👍

server/ocsp.go
philpennock: LGTM: OCSP staple verification is why `.VerifyConnection` was added as a field member.  See `https://github.com/golang/go/issues/36736` for more.

server/reload.go
philpennock: Okay, as long as it's a naming issue and not a logic flaw, then I'll go ahead and approve.

philpennock (approved):
Make whatever naming changes and logging changes you think appropriate to address the issues noted, and please lets try to get scripted cert updating in sooner rather than later.  I don't want to have to debug all the dependency chains years from now when tests have suddenly broken and try to re-create the various OCSP scenarios correctly for them all.

test/ocsp_test.go
wallyqs: iirc in case a cert expires, established connections are not disconnected but new ones would refuse attempting to connect so this would be similar behavior to that, new cluster routes clients would refuse to connect to the server with the invalid staple.

wallyqs:
Closing this one in favor of the changes in this PR that properly gateway and leafnode remotes: https://github.com/nats-io/nats-server/pull/2263

wallyqs:
Note that since we need `VerifyConnection` now the server can only be built with +Go 1.15.

PR #2253, opened 2021-05-26
[CHANGED] MQTT: do not require bearer JWT

kozlovic:
In operator mode, all connections require a JWT. For MQTT clients,
the JWT is passed as the password in the MQTT connect protocol.

Since MQTT clients do not participate in the NATS protocol: do not
receive an INFO from server and can't sign the nonce, a requirement
was made that MQTT clients use a JWT with bearer boolean set to true,
which then cause the server to skip the signature check.

I propose that we simply bypass the signature if the JWT has bearer
token set to true OR if the client is MQTT.

/cc @nats-io/core

### DIFF
--- a/server/auth.go
+++ b/server/auth.go
@@ -374,6 +374,7 @@ func (s *Server) processClientOrLeafAuthentication(c *client, opts *Options) boo
 		ok   bool
 		err  error
 		ao   bool // auth override
+		mqtt bool
 	)
 	s.mu.Lock()
 	authRequired := s.info.AuthRequired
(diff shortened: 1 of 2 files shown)


derekcollison:
I have not looked at code, but my concern is that JWTs (w/o bearer flag) need the signed nonce to properly authenticate. I feel hesitant on relaxing that just for MQTT clients. Feels like in operator mode we should have a shortcut for nsc to generate a user for MQTT, even maybe embedding the username and make sure that is the same as they are passing in.

kozlovic:
@derekcollison I understand. We can keep the requirement to have Bearer in the JWT for MQTT clients.

PR #2254, opened 2021-05-27
Embeds the errors.json into the server

ripienaar:
This allows 3rd party tools to get at the RAW json format
data that builds the error structures and makes the struct
that describes that data publically available.

We want this for the CLI to make assistive tools to edit the
file and give deeper insight into data than those in ApiErrors

Down the line we'll create support services that will make this
data available via a service so would be needed then too

/cc @nats-io/core


ripienaar:
Ah, the problem with this is go 1.15 doesnt do embed.  Are we happy to drop 1.15?

ripienaar:
I can embed this the old way - just uuencode it into a var, then improve it later on to this once we drop 1.15. wdyt @kozlovic ?

kozlovic:
If that works for you and does not add too much complexity/maintenance, then I am ok. Again, I would personally not object to dropping 1.15, I was jus stating what we were doing in the past and that libs will have to also drop it when they pull the server that uses `embed`.
Let's see what @derekcollison and say @wallyqs think of possibly dropping 1.15?

ripienaar:
02ad7c40277800d2ca76cbb100e59c96e56af4e9 embeds the data as Base64 json data into the server.  Once we're on 1.16 we can move to embed fs.

I added a function to expose that data so we can maintain this API even when moving to embed.

derekcollison:
I do remember the discussion about dynamically being able to discover/download this file based on the version of the server you were talking to. 

How much more size does it add to the binary? Available via http & NATS yes?

ripienaar:
For now I am not exposing it any way other than through the Go api so the CLI can get at it.  Once we have more than just JS errors in here and know a bit more about the topic in general I'll look toward exposing it over various methods. 

For now its 18kb extra data, I can compress it too - though might be moot as we wont ultimate control how embed stores this - we have to trust go devs do the right thing.

ripienaar:
cli does:

```go
source, err := server.ErrorsDataSource()
```

and then just does whatever it likes to do with it.  

```
$ nats errors list consumer
╭────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│                                                                                              NATS Errors                                                                                               │
├───────────┬─────────────────┬──────────────────────────────────────────────────────────────────────────────┬──────────────────────────────────────────────┬────────────────────────────────────────────┤
│ NATS Code │ HTTP Error Code │ Description                                                                  │ Comment                                      │ Go Constant                                │
├───────────┼─────────────────┼──────────────────────────────────────────────────────────────────────────────┼──────────────────────────────────────────────┼────────────────────────────────────────────┤
│ 10012     │ 500             │ {err}                                                                        │ General consumer creation failure string     │ JSConsumerCreateErrF                       │
│ 10013     │ 400             │ consumer name already in use                                                 │                                              │ JSConsumerNameExistErr                     │
│ 10014     │ 404             │ consumer not found                                                           │                                              │ JSConsumerNotFoundErr                      │
.....
```

```
$ nats errors lookup 10013 
NATS Error Code: 10013

        Description: consumer name already in use
          HTTP Code: 400
  Go Index Constant: JSConsumerNameExistErr

No further information available
```

And it has utilities for maintaining the json file, validating it etc.

So today this is not totally amazing as its based on whatever the CLI is compiled on of course.  But as a incremental step I think it's fine until we expose this data over the server via some API

derekcollison:
If that is the case for now have cli directly pull it in but not build it into the server for now..

derekcollison:
I am not convinced totally it should be in the server vs an endpoint on nats.io as mentioned before. 

In that case the cli (or other tool) connects to a server and it gets the version. If it does not already have errors.json for that version it needs to get it and cache it, question now is where from, the server or another endpoint like mentioned above.

Would like to think some more on it myself.

ripienaar:
We cannot expect $enterprise to have access to nats.io just to use basic cli features imo, but fine, I'll move it out.

derekcollison:
Let's just hold on PR for a bit, that is all I am suggesting.

ripienaar:
Don't like PRs hanging around CLI have not had a release for a long time, so rather remove it and move forward

PR #2255, opened 2021-05-28
Minor updates to error data

ripienaar:

/cc @nats-io/core


ripienaar:
some minor fixes that was included in the embed pr

PR #2260, opened 2021-06-07
Remove external dependency on 'ps' for memory and cpu usage.

derekcollison:
This removes the dependency on `ps` and just uses rusage and sampling for percentage of cpu. We do not use virtual memory size in any of the upper layers.

/cc @nats-io/core


kozlovic (approved):
LGTM

PR #2266, opened 2021-06-08
ensure correct error handling in non clustered stream add

ripienaar:
The addStream() can return an ApiErr but we did not handle
that leading to errors like 'stream name already in use (10058)' instead
of just 'stream name already in use' with the correct error code 10058 set

/cc @nats-io/core


derekcollison (approved):
LGTM

PR #2274, opened 2021-06-09
[FIXED] LeafNode config reload failed without any change made

### LINKED ISSUE #2273: reload failed for leafnode configuration
## Defect

(Derek said this is a bug and to file a report, during a debugging meeting for another issue)

* nats-server: `v2.3.0-beta.1`
* client library: whatever @wallyqs was using

Ubuntu 20.04 LTS inside systemd

```
systemctl kill --signal=HUP jsngs@pennock_cheddar.service
journalctl -u jsngs@pennock_cheddar.service -f
```

#### Expected result:

Debugging output enabled, per @derekcollison 

#### Actual result:

```
Jun 09 15:23:47 ip-10-32-164-190 nats-server[42869]: [42869] 2021/06/09 19:23:47.373862 [ERR] Failed to reload server configuration: config reload not supported for LeafNode: old={ 0    [] 0 <nil> 0 false map[]  false 1s [0xc0000fe3c0] <nil> 0 0 <nil>}, new={ 0    [] 0 <nil> 0 false map[]  false 1s [0xc0000fe320] <nil> 0 0 <nil>}
```

kozlovic:
Issuing a configuration reload for a leafnode that has remotes
defined with remotes having more than 1 url could lead to a failure.
This is because we have introduced shuffling of remote urls but
that was done in the server's options object, which then would
cause the DeepEqual when diff'ing options to fail.
We move the suffling to the private list of urls.

The other issue was that the "old" remote option may not have
had a local account and it was not set to "$G", which could make
the DeepEqual fail.

Resolves #2273

/cc @nats-io/core

### DIFF
--- a/server/leafnode.go
+++ b/server/leafnode.go
@@ -1,4 +1,4 @@
-// Copyright 2019-2020 The NATS Authors
+// Copyright 2019-2021 The NATS Authors
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
 // You may obtain a copy of the License at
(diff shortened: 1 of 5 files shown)


derekcollison (approved):
LGTM

wallyqs (approved):
LGTM

PR #2275, opened 2021-06-09
[FIXED] Handling of subject rewrites for subjects to a globally routed subject.

derekcollison:
We were seeing results where subject rewrite for JetStream was not working properly when the reply subject to a leafnode was a globally routed subject.

/cc @nats-io/core

### DIFF
--- a/server/stream.go
+++ b/server/stream.go
@@ -657,7 +657,7 @@ func (mset *stream) sendCreateAdvisory() {
 	}
 
 	subj := JSAdvisoryStreamCreatedPre + "." + name
-	outq.send(&jsPubMsg{subj, subj, _EMPTY_, nil, j, nil, 0, nil})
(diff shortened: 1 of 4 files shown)


server/stream.go
```diff
@@ -1765,16 +1765,17 @@ func (mset *stream) setSourceConsumer(iname string, seq uint64) {
 func (mset *stream) processSourceMsgs(si *sourceInfo) {
 	s := mset.srv
 	defer s.grWG.Done()
+
+	if si == nil {
```
kozlovic, line 1769: I am assuming this is not related to this issue, but required otherwise could crash during the defer.

kozlovic (approved):
LGTM

server/stream.go
derekcollison: yes correct, was a pending change I was going to make so just put it in here.

Want to make sure tests are green so will keep watching and merge after.

PR #2276, opened 2021-06-10
Added active servers to statsz.

derekcollison:
This is generally useful but will also help with cli actions since we know from first response how many total responses are expected.

/cc @nats-io/core

### DIFF
--- a/server/events.go
+++ b/server/events.go
@@ -199,6 +199,7 @@ type ServerStats struct {
 	SlowConsumers    int64          `json:"slow_consumers"`
 	Routes           []*RouteStat   `json:"routes,omitempty"`
 	Gateways         []*GatewayStat `json:"gateways,omitempty"`
+	ActiveServers    int            `json:"active_servers,omitempty"`
 }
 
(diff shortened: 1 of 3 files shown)


server/events.go
```diff
@@ -924,12 +931,6 @@ func (s *Server) remoteServerUpdate(sub *subscription, _ *client, subject, reply
 	}
 	si := ssm.Server
 	node := string(getHash(si.Name))
-	if _, ok := s.nodeToInfo.Load(node); !ok {
```
matthiashanel, line 927: Is this compensated by the addition in processNewServer?

ripienaar:
LGTM, didnt check how this is maintained - like how accurate it is and on what time frame but will help with hints

server/events.go
derekcollison: Yes.

derekcollison:
If the servers shutdown and that message is received the server is removed, otherwise we will remove it when we sweep for lost remotes, iirc 2mins.

matthiashanel (approved):
LGTM

PR #2277, opened 2021-06-10
ocsp: add more config options to customize OCSP

wallyqs:
Allows configuration of OCSP mode and responder url via the config file if not present in the TLS cert:

```hcl
ocsp {
  mode: must # always, never, auto
  url: "http://ocsp.example.net"
}
```

By default, when ocsp is enabled it runs in `auto` mode but there are a few other different OCSP modes that control the behavior on whether OCSP should be enforced and server should shutdown if the cert runs with a revoked staple.

| Mode   | Description                                                           | Server shutdowns when revoked |
| --------- | ---- | ----- |
| auto   | Enables OCSP Stapling when it has the must staple/status_request flag | No                            |
| must   | Enables OCSP Staping when it has the must staple/status_request flag  | Yes                           |
| always | Enables OCSP Stapling for all certificates                            | Yes                           |
| never  | Disables OCSP Stapling even if must staple flag is present            | No                            |

/cc @nats-io/core

### DIFF
--- a/test/ocsp_test.go
+++ b/test/ocsp_test.go
@@ -964,7 +964,7 @@ func TestOCSPReloadRotateTLSCertEnableMustStaple(t *testing.T) {
 	nc.Close()
 }
 
-func TestOCSPClusterReload(t *testing.T) {
+func TestOCSPCluster(t *testing.T) {
 	const (
 		caCert = "configs/certs/ocsp/ca-cert.pem"
 		caKey  = "configs/certs/ocsp/ca-key.pem"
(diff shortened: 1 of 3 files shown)


philpennock (approved):

test/ocsp_test.go
```diff
@@ -1782,6 +1782,480 @@ func TestOCSPGateway(t *testing.T) {
...
+			} else if test.rerr == nil && err != nil {
+				t.Errorf("Expected response")
+			}
+		})
+	}
+}
+
+func TestOCSPCustomConfigReloadDisable(t *testing.T) {
+	doLog = true
+	doDebug = true
```
kozlovic, line 2055: Remove those 2, otherwise it prints server output in test.

test/ocsp_test.go
```diff
@@ -1782,6 +1782,480 @@ func TestOCSPGateway(t *testing.T) {
...
+	)
+	if err != nil {
+		t.Fatal(err)
+	}
+	nc.Close()
+}
+
+func TestOCSPCustomConfigReloadEnable(t *testing.T) {
+	doLog = true
+	doDebug = true
```
kozlovic, line 2159: Same

kozlovic (approved):
LGTM, but remove debug/trace from tests.

kozlovic (approved):

test/ocsp_test.go
wallyqs: thanks, just updated

PR #2279, opened 2021-06-11
[FIXED] Orphaned consumers on sourced or mirrored streams keep trying to create new ones.

derekcollison:
When leader changes happened to streams that mirrored or sourced other streams we could continue to try to create consumers.

This could get excessive on a stream that has sourced many upstream origin streams and had several leader changes.


ripienaar (approved):
LGTM

PR #2280, opened 2021-06-11
[FIX] Avoid spinning when we are in a cacth-up phase but first message matches index but not term.

derekcollison:
This would cause CPU spikes and catch-up cycles that could spin.


ripienaar:
LGTM

PR #2281, opened 2021-06-12
[FIXED] Dynamic account limits would limit based on single server limits.

derekcollison:
When checking limits we would check total ask against the server limits if limits were not set.
We were also dynamically setting account limits based on a single server limit.


ripienaar:
LGTM

PR #2284, opened 2021-06-15
Adds in per subject tracking and limits for streams.

derekcollison:
Streams have always allowed multiple subjects however setting initial pending and selecting starting sequence number needed to be furthered optimized.

We also introduce per subject message limits.

/cc @nats-io/core

### DIFF
--- a/server/norace_test.go
+++ b/server/norace_test.go
@@ -2000,3 +2000,154 @@ func TestNoRaceJetStreamClusterSuperClusterRIPStress(t *testing.T) {
 		time.Sleep(10 * time.Millisecond)
 	}
 }
+
+func TestJetStreamSlowFilteredInititalPendingAndFirstMsg(t *testing.T) {
+	s := RunBasicJetStreamServer()
+	defer s.Shutdown()
+
+	if config := s.JetStreamConfig(); config != nil {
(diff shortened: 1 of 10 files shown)


server/norace_test.go
```diff
@@ -2000,3 +2000,154 @@ func TestNoRaceJetStreamClusterSuperClusterRIPStress(t *testing.T) {
 		time.Sleep(10 * time.Millisecond)
 	}
 }
+
+func TestJetStreamSlowFilteredInititalPendingAndFirstMsg(t *testing.T) {
```
kozlovic, line 2004: Since this is in the `norace_test.go` that we want to run only if not passing `-race` to `go test`, the name of the test has to start with `TestNoRace`. This is because on travis we do:
```
go test -v -run=TestNoRace --failfast -p=1 ./...
```
While at it, can you also change the name of this test: `TestJetStreamClusterMirrorExpirationAndMissingSequences` so it starts with `TestNoRace`

kozlovic (changes requested):

server/norace_test.go
derekcollison: ok will do.

kozlovic (approved):
LGTM

PR #2288, opened 2021-06-17
[fixing] leafnode missing retry and service export interest propagation

matthiashanel:
A missing account on initial connect attempt caused the leaf node
connection to never be established.

An account service import subscription was not propagated along leaf
node connections.

I manually tested that this new behavior works.
I created issue #2289 to modify the unit test later. (discussed with Derek)

### DIFF
--- a/server/leafnode.go
+++ b/server/leafnode.go
@@ -816,7 +816,7 @@ func (s *Server) createLeafNode(conn net.Conn, rURL *url.URL, remote *leafNodeCf
 			// has not yet been pushed, or the request failed for other reasons.
 			// remote needs to be set or retry won't happen
 			c.leaf.remote = remote
-			c.closeConnection(AuthenticationViolation)
+			c.closeConnection(MissingAccount)
 			s.Errorf("No local account %s found for solicited leaf node connection", lacc)
 			return nil
 		}
--- a/server/leafnode.go
+++ b/server/leafnode.go
@@ -817,7 +817,7 @@ func (s *Server) createLeafNode(conn net.Conn, rURL *url.URL, remote *leafNodeCf
 			// remote needs to be set or retry won't happen
 			c.leaf.remote = remote
 			c.closeConnection(MissingAccount)
-			s.Errorf("No local account %s found for solicited leaf node connection", lacc)
+			s.Errorf("Unable to lookup account %s for solicited leafnode connection: %v", lacc, err)
 			return nil
 		}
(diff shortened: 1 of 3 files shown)


server/leafnode.go
```diff
@@ -813,8 +812,12 @@ func (s *Server) createLeafNode(conn net.Conn, rURL *url.URL, remote *leafNodeCf
 		var err error
 		acc, err = s.LookupAccount(lacc)
 		if err != nil {
-			s.Errorf("No local account %q for leafnode: %v", lacc, err)
-			c.closeConnection(MissingAccount)
+			// An account not existing is something that can happen with nats/http account resolver and the account
+			// has not yet been pushed, or the request failed for other reasons.
+			// remote needs to be set or retry won't happen
+			c.leaf.remote = remote
+			c.closeConnection(AuthenticationViolation)
```
kozlovic, line 819: Why not using "MissingAccount" for the reason?

kozlovic (changes requested):
Why changing the closeConnection's reason and removing the error from tracing?

derekcollison (approved):
LGTM - sans Ivan's comment on close reason.

server/leafnode.go
matthiashanel: fixed

server/leafnode.go
```diff
@@ -813,8 +812,12 @@ func (s *Server) createLeafNode(conn net.Conn, rURL *url.URL, remote *leafNodeCf
 		var err error
 		acc, err = s.LookupAccount(lacc)
 		if err != nil {
-			s.Errorf("No local account %q for leafnode: %v", lacc, err)
+			// An account not existing is something that can happen with nats/http account resolver and the account
+			// has not yet been pushed, or the request failed for other reasons.
+			// remote needs to be set or retry won't happen
+			c.leaf.remote = remote
 			c.closeConnection(MissingAccount)
+			s.Errorf("No local account %s found for solicited leaf node connection", lacc)
```
kozlovic, line 820: Looks like my comment about the error disappeared? I commented this:
It is likely that the account lookup failure means that the account does not exist, but there may be value in reporting the actual error. So error could be: `s.Errorf("Unable to lookup account %s for solicited leafnode connection: %v", lacc, err)`.

kozlovic (changes requested):

kozlovic (approved):
LGTM

PR #2292, opened 2021-06-18
remove config !nil check,beacuse the if branch is always not nil

dtest11:
obviously in the else condition config alway not nil ,so no need check
/cc @nats-io/core


derekcollison (approved):
LGTM - Thanks!

Issue #2290, opened 2021-06-18
js: Error description missing in some requests


wallyqs:
Making a bad request results in the following:

```
sub, err = js.PullSubscribe("bar", "")
// err => nats: {err}
```

From the JSON response looks like the templated error is being returned:

```json
{
    "error": {
        "code": 500,
        "description": "{err}",
        "err_code": 10012
    },
    "type": "io.nats.jetstream.api.v1.consumer_create_response"
}
```

PR #2293, opened 2021-06-18
Fix for #2290

derekcollison:

### DIFF
--- a/server/jetstream_api.go
+++ b/server/jetstream_api.go
@@ -935,7 +935,7 @@ func (s *Server) jsTemplateCreateRequest(sub *subscription, c *client, subject, 
 
 	t, err := acc.addStreamTemplate(&cfg)
 	if err != nil {
-		resp.Error = ApiErrors[JSStreamTemplateCreateErrF].ErrOr(err)
+		resp.Error = ApiErrors[JSStreamTemplateCreateErrF].ErrOrNewT(err, "{err}", err)
 		s.sendAPIErrResponse(ci, acc, subject, reply, string(msg), s.jsonResponse(&resp))
 		return
 	}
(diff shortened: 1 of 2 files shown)


ripienaar (approved):
LGTM

server/jetstream_api.go
```diff
@@ -1351,13 +1351,13 @@ func (s *Server) jsStreamUpdateRequest(sub *subscription, c *client, subject, re
 
 	mset, err := acc.lookupStream(streamName)
 	if err != nil {
-		resp.Error = ApiErrors[JSStreamNotFoundErr].ErrOr(err)
+		resp.Error = ApiErrors[JSStreamNotFoundErr].ErrOrNewT(err, "{err}", err)
```
kozlovic, line 1354: The JSStreamNotFoundErr does not have a `{err}`, so I don't think ErrOrNewT will do anything. Basically, the way I understand it, unless the err constant ends with `F`, there is no need to call ErrOrNewT. (this comment applies to other changes in this PR).

ripienaar: You’re right. I messed up will send a follow up. 

I wonder if we should give the 2 types of error different types to help detect this mistake at compile time because it’s a pain as is.

kozlovic: However, what I am not clear is: should still be a ErrOrNewT() but without replacement. Because lookupStream() returns a normal `error`, but here we still want to assign to resp.Error a APIError. So at one point there should be some conversion no? Or that happens during serialization where an `error` is converted to an APIError and APIError is an interface that implements `error` properties.

ripienaar: It should be harmless as is I think - but I should change to ErrOrNew()

what I will experiment with is that the tagged variables are just not acceptable to the non T functions and reverse. Then we can catch it at compile time. As is it’s really prone to human error

ripienaar: Yes confirming using a non tagged err in ErrOrNewT() is harmless - its a noop.  So this is fine, but I'll clean up and look into detecting these 2 easy mistakes at compile time

ripienaar: https://github.com/nats-io/nats-server/pull/2294

PR #2295, opened 2021-06-18
further tagged error confusion cleanups

ripienaar:

/cc @nats-io/core


derekcollison (approved):
LGTM - But not too familiar.

ripienaar:
will take another manual run through them to ensure there's no more 'morrow, while thinking about ways the compiler can help us avoid ths.

PR #2296, opened 2021-06-18
Allow for more advanced purge operations.

derekcollison:
This will allow purge by subject and the ability to specify the sequence or number of messages to keep.

### DIFF
--- a/server/filestore.go
+++ b/server/filestore.go
@@ -936,13 +936,7 @@ func (fs *fileStore) FilteredState(sseq uint64, subj string) SimpleState {
 		}
 	} else {
 		// Fallback to linear scan.
-		var eq func(string, string) bool
-		if wc {
-			eq = subjectIsSubsetMatch
-		} else {
-			eq = func(a, b string) bool { return a == b }
-		}
-
+		eq := compareFn(subj)
 		for seq := sseq; seq <= lseq; seq++ {
 			if sm, _ := fs.msgForSeq(seq); sm != nil && eq(sm.subj, subj) {
 				ss.Msgs++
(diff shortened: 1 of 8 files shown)


server/filestore.go
```diff
@@ -3109,6 +3103,57 @@ func (fs *fileStore) dmapEntries() int {
...
+	if subjectHasWildcard(subject) {
+		return subjectIsSubsetMatch
+	}
+	return subjectsEqual
+}
+
+// PurgeEx will remove messages based on subject filters, sequence and number of messages to keep.
+// Will return the number of purged messages.
+func (fs *fileStore) PurgeEx(subject string, sequence, keep uint64) (purged uint64, err error) {
+	if subject == _EMPTY_ || subject == ">" && keep == 0 {
```
ripienaar, line 3128: Could also check if the subject is `*` and it the ingest subject is single token

ripienaar (approved):
LGTM

server/filestore.go
derekcollison: I thought that as well at first but I don't know the true makeup of the diversity of subjects in the stream since they could be delivered via source or mirror etc.

server/jetstream_api.go
```diff
@@ -352,6 +352,16 @@ type JSApiStreamListResponse struct {
 
 const JSApiStreamListResponseType = "io.nats.jetstream.api.v1.stream_list_response"
 
+// JSApiStreamPurgeRequest is optional request information to the purge API.
+type JSApiStreamPurgeRequest struct {
+	// Purge up to but not including sequence.
+	Sequence uint64 `json:"seq,omitempty"`
+	// Subject to match against messages for the purge command.
+	Subject string `json:"filter,omitempty"`
+	// Number of messages to keep.
```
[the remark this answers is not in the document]
derekcollison: Will add a better comment, good point.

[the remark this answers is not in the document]
derekcollison: Updated.

server/filestore.go
ripienaar: Oh yeah good point didn’t think about that

ripienaar:
LGTM

server/memstore.go
```diff
@@ -366,6 +386,40 @@ func (ms *memStore) expireMsgs() {
...
+	if ss := ms.FilteredState(1, subject); ss.Msgs > 0 {
+		if keep > 0 {
+			if keep > ss.Msgs {
+				return 0, nil
+			}
+			ss.Msgs -= keep
+		}
+		last := ss.Last
+		if sequence > 0 {
+			last = sequence + 1
```
kozlovic, line 405: You have:
```
// Purge up to but not including sequence.
Sequence uint64 `json:"seq,omitempty"`
```
So I have a stream with subject "foo" and "bar", each with 5 messages. I sent request to purge "foo" with Sequence: 4, I expect this subject to have only messages 1, 2, 3 removed, so total messages left in the stream should be 2 + 5 = 7. But instead, the 5 messages from "foo" are removed.
Question is: is the comment for the option wrong or should it be that `last` is equal to `sequence-1`?

derekcollison: That code changed in a subsequent commit.

Sequences are not tracked per subject, so when using sequence in this regard it is absolute to the whole stream.

PR #2297, opened 2021-06-20
Fixes to PurgeEx

derekcollison:
Fix for purge and keep when keep == total msgs.
Fix for purge with keep but no filtered subject.


ripienaar:
LGTM

PR #2299, opened 2021-06-21
test to show slow purges

ripienaar:

/cc @nats-io/core

### DIFF
--- a/server/norace_test.go
+++ b/server/norace_test.go
@@ -17,6 +17,7 @@ package server
 
 import (
 	"bufio"
+	"context"
 	"encoding/json"
 	"fmt"
 	"math/rand"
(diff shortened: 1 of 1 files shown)


server/norace_test.go
```diff
@@ -1756,6 +1757,89 @@ func TestNoRaceJetStreamClusterSourcesMuxd(t *testing.T) {
...
+		}
+	}
+
+	s := RunBasicJetStreamServer()
+	defer s.Shutdown()
+
+	nc, js := jsClientConnect(t, s)
+	defer nc.Close()
+
+	cfg := StreamConfig{
```
derekcollison, line 1776: Can be normal js.AddStream. Check jetstream_cluster_test.go

server/norace_test.go
```diff
@@ -1756,6 +1757,89 @@ func TestNoRaceJetStreamClusterSourcesMuxd(t *testing.T) {
...
+	cerr(t, err)
+
+	_, err = nc.Request(fmt.Sprintf(JSApiStreamCreateT, cfg.Name), req, time.Second)
+	cerr(t, err)
+
+	si, err := js.StreamInfo("KV")
+	cerr(t, err)
+
+	// 10kb messages spread over 1000 subjects
+	body := make([]byte, 101024)
```
derekcollison, line 1791: This supposed to be 10*1024?

server/norace_test.go
```diff
@@ -1756,6 +1757,89 @@ func TestNoRaceJetStreamClusterSourcesMuxd(t *testing.T) {
...
+		err := nc.Publish(fmt.Sprintf("kv.%d", i%1000), body)
+		cerr(t, err)
+	}
+	si, err = js.StreamInfo("KV")
+	cerr(t, err)
+	if si == nil || si.Config.Name != "KV" {
+		t.Fatalf("StreamInfo is not correct %+v", si)
+	}
+
+	func() {
```
derekcollison, line 1802: Can use checkFor, see jetstream_cluster_test.go for some examples.

derekcollison (changes requested):

server/norace_test.go
ripienaar: ah, I initially had a per subject bit in there so did it this way but removed it.

server/norace_test.go
ripienaar: something like that :P I probably just fiddled numbers till it broke, should be 100kb

derekcollison (approved):
LGTM

PR #2300, opened 2021-06-21
[FIXED] Websocket (de)compression issues with continuation frames

### LINKED ISSUE #2287: Websocket failed to deliver big json obj when compression = true
Hi,
_First of all please let me say thank you for such awesome open source project, it's simply brilliant!_

- **Large JSON Silently fails to transfer.**
Lately, when testing nats directly from browser (using `nats.ws`), when I tried to send an JSON object that's a bit large (>3000bytes, consists of random numbers), the other party (browser) yields `invalid frame header` error & disconnects.

- **No visible error on server log**
By observing nats log, the initial message (PUB) is successfully delivered to server, but then something went wrong, & there's no visible 'error' in server side.

- **`ws-to-tcp` and other transparent `ws-to-tcp` proxy works**
Upon further testing: when used `ws-to-tcp` (npm i ws-to-tcp -g) instead of the built in `ws`, with same code, same nats-server, only difference is the ws is now supplied by 3rd party proxy (connect directly to 4222), everything works. 

- **compression = false fixes the problem**
Thus I came to suspect something might be improperly handled inside `ws` related logic of nats-server.
Finally: I've tried messed around the conf, & found out when ***compression = false***, the problem disappears.

My config & log are attached.

### _Logs_

[compression-off.log](https://github.com/nats-io/nats-server/files/6668274/compression-off.log)
[compression-on.log](https://github.com/nats-io/nats-server/files/6668275/compression-on.log)

### _Snapshot when one side fails_

snap

### _Config_

```
listen: "0.0.0.0:4222"
accounts: {
    base: { }
}
system_account: base

websocket {
    port: 8000
    no_tls: true
    same_origin: false
    compression: false #or true
}
```

nats-server: v2.2.6
nats.ws: ^1.1.4

macOS Catalina 10.15.7
Darwin Kernel Version 19.6.0: Thu May  6 00:48:39 PDT 2021;

Attached above.

#### Expected result:
Large JSON (<max_payload_size) should deliver via ws when `compression = true`

#### Actual result:
Fails to deliver when `compression = true`

kozlovic:
For compression, continuation frames had the compress bit set, which is
wrong since only the first frame should.
For decompression, continuation frames were decompressed individually
instead of assembling the full payload and then decompressing.

Resolves #2287

### DIFF
--- a/server/websocket.go
+++ b/server/websocket.go
@@ -75,7 +75,6 @@ const (
 	wsFirstFrame        = true
 	wsContFrame         = false
 	wsFinalFrame        = true
-	wsCompressedFrame   = true
 	wsUncompressedFrame = false
 
 	wsSchemePrefix    = "ws"
@@ -92,6 +91,7 @@ const (
 )
 
 var decompressorPool sync.Pool
+var compressLastBlock = []byte{0x00, 0x00, 0xff, 0xff, 0x01, 0x00, 0x00, 0xff, 0xff}
 
 // From https://tools.ietf.org/html/rfc6455#section-1.3
 var wsGUID = []byte("258EAFA5-E914-47DA-95CA-C5AB0DC85B11")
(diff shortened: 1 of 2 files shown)


derekcollison (approved):
LGTM

PR #2301, opened 2021-06-21
[FIXED] Clients disconnected on reload when only $SYS account configured

### LINKED ISSUE #2282: reloading server with $SYS account disconnects clients connected to $G
this is the config 
```
accounts {
    $SYS {
        users = [
            {
                user: "system",
                pass: "password"
            }
        ]
    }
}
```
Then I start a server like this
```
nats-server -c x.cfg
```
I subscribe 
```
nats --server localhost:4222   sub foo
```
Then I reload the server
```
 kill -HUP 12588
```
as a result of this, the client connected to $G (which is not listed) gets disconnected probably because $G is not explicitly listed.
Meaning I would either not have expected the client to connect in the first place OR it not be forced into reconnect.
```
nats --server localhost:4222   sub foo
17:50:18 Subscribing on foo
17:50:28 Disconnected due to: EOF, will attempt reconnect
17:50:30 Reconnected [nats://localhost:4222]
```

kozlovic:
This was introduced by PR#1900 by detecting that a single system
account was configured. A "fake" user name was created and added
to the list of users and set as a no_auth_user.
However, on config reload, a new fake user was created, which
would cause existing connected clients to be disconnected.

Resolves #2282

### DIFF
--- a/server/reload_test.go
+++ b/server/reload_test.go
@@ -4522,3 +4522,39 @@ func TestConfigReloadAccountMappings(t *testing.T) {
 	checkPending(fsub, 1)
 	checkPending(sub, 0)
 }
+
+func TestConfigReloadWithSysAccountOnly(t *testing.T) {
+	conf := createConfFile(t, []byte(`
+		listen: "127.0.0.1:-1"
+		accounts {
+			$SYS {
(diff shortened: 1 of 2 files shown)


derekcollison (approved):
LGTM

matthiashanel (approved):
LGTM

PR #2302, opened 2021-06-22
JetStream Encryption at Rest

derekcollison:
This adds in encryption at rest for JetStream. Supports XChaChaPoly1305 and ChaCha20 for message blocks since we use highway hash and sequence numbers already for authentication.

/cc @nats-io/core


kozlovic (approved):
LGTM.

server/jetstream.go
```diff
@@ -1011,12 +1082,23 @@ func (a *Account) EnableJetStream(limits *JetStreamAccountLimits) error {
...
 			if _, err := os.Stat(metasum); os.IsNotExist(err) {
-				s.Warnf("    Missing Consumer checksum for %q", metasum)
+				s.Warnf("    Missing consumer checksum for %q", metasum)
 				continue
 			}
+
+			// Check if we are encrypted.
+			if key, err := ioutil.ReadFile(path.Join(odir, ofi.Name(), JetStreamMetaFileKey)); err == nil {
+				s.Debugf("  Consumer metafile is encrypted, reading encrypted keyfile")
+				// Decode the buffer before proceeding.
```
ripienaar, line 1092: curious, what data is in consumer state thats sensitive? I guess subjects?

derekcollison: Consumer meta is config for consumer. State is delivered, ack'd stuff etc. So both I consider private and need to be encrypted.

PR #2304, opened 2021-06-22
[CHANGED] Reduce print for an account subs limit to every 2 sec

kozlovic:
We could make it for all limits by having a map of error types
instead of applying just to max subs.

/cc @nats-io/core


derekcollison (approved):
LGTM - Thanks

PR #2305, opened 2021-06-22
[FIXED] Failed route TLS handshake would leave failed conn's lock, locked

kozlovic:
This is a regression introduced in v2.2.6.

/cc @nats-io/core


derekcollison (approved):
LGTM

PR #2306, opened 2021-06-22
Improvements to filtered purge and general memory usage for filestore.

derekcollison:
We optimized the filtered purge to skip msgBlks that are not in play.
Also optimized msgBlock buffer usage by using two sync.Pools to enhance reuse.

/cc @nats-io/core


kozlovic (approved):
LGTM

kozlovic (approved):
LGTM

PR #2308, opened 2021-06-23
rename copy to other with param …

dtest11:
1. the keyword copy is builtin func ,rename copy to other with param as prefix
    2. fix some func note

/cc @nats-io/core


server/accounts.go
```diff
@@ -2843,7 +2843,7 @@ func (a *Account) isClaimAccount() bool {
 	return a.claimJWT != ""
 }
 
-// updateAccountClaims will update an existing account with new claims.
+// UpdateAccountClaims will update an existing account with new claims.
```
dtest11, line 2846: error

PR #2311, opened 2021-06-24
make func doc note as Godoc suggested

dtest11:
I'm learning nats, I will continue to modify godoc, 
if you don't think it is necessary, please let me know, 
I am very sorry for the interference. 
smile

 -

/cc @nats-io/core


derekcollison:
All good, please let us know how we can help! And keep the ideas and PRs coming.

PR #2313, opened 2021-06-24
Added in Stream get last message by subject.

derekcollison:
This is to aid in K/V overlay for simple Get ops vs creating a full blown consumer.

/cc @nats-io/core


kozlovic (approved):
LGTM

ripienaar (approved):
so nice

PR #2315, opened 2021-06-25
Update OCSP CA

variadico:
Currently, the CA used in OCSP tests is expired as of `Jun 24 22:45:27 2021 GMT`. I forgot to set an explicit expiration and the default is 30 days.

This change regenerates the OCSP certs with a CA that expires in 2029.


wallyqs (approved):

derekcollison (approved):
LGTM

PR #2317, opened 2021-06-25
FIXED Clustered streams with seq mismatch state on store failures.

derekcollison:
We did account for high level failure properly but were not properly accounting for low level store failures. When in clustered mode we need to account for the failure properly, and we use clfs for that, just needed updating on low level store failure.

/cc @nats-io/core


ripienaar (approved):
LGTM

PR #2318, opened 2021-06-27
Add hardened systemd service

c0deaddict:
 - [ ] Tests added

Resolves #

I'm working on adding a [NATS service to NixOS](https://github.com/NixOS/nixpkgs/pull/128189). During this process I found that the recommended systemd service definition does not contain any systemd hardening options. In NixOS more and more services have hardening options applied (where possible). Giving services the least amount of privileges makes sense to me.

This PR adds a second systemd service file that has hardening options applied. Tested this on my machine with basic pub/sub and with (JetStream) streams (to verify files can be accessed by NATS). Reason for adding a second file is to give users a choice between a bare-bones service definition or a hardened definition. Having the hardening options in this repo could also serve as documentation about the capabilities/privileges that NATS needs.

I would like to hear your opinion on this. Via this PR I would also like to get feedback about the used systemd hardening options. I'm unsure if I tightened the capabilities/privileges too much.

/cc @nats-io/core

### DIFF
--- a/util/nats-server-hardened.service
+++ b/util/nats-server-hardened.service
@@ -1,0 +1,61 @@
+[Unit]
+Description=NATS Server
+After=network.target ntp.service
+
+[Service]
+Type=simple
+ExecStart=/usr/sbin/nats-server -c /etc/nats-server.conf
+ExecReload=/bin/kill -s HUP $MAINPID
+ExecStop=/bin/kill -s SIGINT $MAINPID
+User=nats
+Group=nats
+
+# Hardening
+CapabilityBoundingSet=
+LimitNOFILE=800000 # JetStream requires 2 FDs open per stream.
+LockPersonality=true
+MemoryDenyWriteExecute=true
+NoNewPrivileges=true
+PrivateDevices=true
+PrivateTmp=true
+PrivateUsers=true
+ProcSubset=pid
+ProtectClock=true
+ProtectControlGroups=true
+ProtectHome=true
+ProtectHostname=true
+ProtectKernelLogs=true
+ProtectKernelModules=true
+ProtectKernelTunables=true
+ProtectSystem=strict
+ReadOnlyPaths=
+RestrictAddressFamilies=AF_INET AF_INET6
+RestrictNamespaces=true
+RestrictRealtime=true
+RestrictSUIDSGID=true
+SystemCallFilter=@system-service ~@privileged ~@resources
+UMask=0077
+
+# If you have systemd >= 247
+ProtectProc=invisible
+
+# If you have systemd >= 248
+PrivateIPC=true
+
+# Optional: writable directory for JetStream.
+ReadWritePaths=/var/lib/nats
+
+# Optional: resource control.
+# Replace weights by values that make sense for your situation.
+# For a list of all options see:
+# https://www.freedesktop.org/software/systemd/man/systemd.resource-control.html
+#CPUAccounting=true
+#CPUWeight=100 # of 10000
+#IOAccounting=true
+#IOWeight=100 # of 10000
+#MemoryAccounting=true
+#MemoryMax=1GB
+#IPAccounting=true
+
+[Install]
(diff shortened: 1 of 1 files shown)


derekcollison:
This looks awesome! Excited about the NIX integration as well.
Will have @philpennock take a look.

util/nats-server-hardened.service
```diff
@@ -0,0 +1,43 @@
...
+PrivateIPC=true
+ProcSubset=pid
+ProtectClock=true
+ProtectControlGroups=true
+ProtectHome=true
+ProtectHostname=true
+ProtectKernelLogs=true
+ProtectKernelModules=true
+ProtectKernelTunables=true
+ProtectProc=invisible
```
philpennock, line 29: This one I'm not comfortable with, unless added to a section which calls out that it's for very new systemd?  This isn't in `systemd 245` as supplied by Ubuntu 20.04 LTS.

NixOS being able to benefit from shipping the latest and greatest is good, but this PR would be even better if we could call out those which won't work in the latest LTS from Ubuntu, as that will reduce the number of times we're asked for support by people who see "hardened" and try it, without understanding or the ability to trade off.

philpennock (approved):
We run some nats-servers with a similar expanded set of hardening settings.

I don't see `NoNewPrivileges=true` in your list; deliberate omission?

Other non-hardening directives worth considering:

```
# JetStream requires 2 FDs open per stream
LimitNOFILE=800000

CPUAccounting=yes
IOAccounting=yes
MemoryAccounting=yes
IPAccounting=yes
```

util/nats-server-hardened.service
c0deaddict: Thanks for your feedback!

I left out `NoNewPrivileges=true` because it's implied by other flags. But now that you mention it, having it in the file explicitly makes sense.

I've added the `LimitNOFILE` to the set. Also mentioned the resource control options. I won't put the resource control options in the NixOS NATS module configuration, as the weight values depend on the user's use of the module (usage, system load, capacity, etc.). In NixOS it's fairly easy for admins to extend the NATS systemd service: `systemd.services.nats.serviceConfig.CPUWeight = 1000;`

`ProtectProc` and `PrivateIPC` have a comment now with the minimum required systemd version. I did some testing, and it appears that systemd happily ignores options that it doesn't know yet. Even my unstable NixOS only has systemd `247.6`, which doesn't support `PrivateIPC`, yet I don't see any warning in `systemctl status nats` or `journalctl -u nats`.

c0deaddict: I searched in the documentation on what systemd does when it encounters an unknown option. For the `[Unit]` section the [docs](https://www.freedesktop.org/software/systemd/man/systemd.unit.html) say:

> Unit files may contain additional options on top of those listed here. If systemd encounters an unknown option, it will write a warning log message but continue loading the unit. If an option or section name is prefixed with X-, it is ignored completely by systemd. Options within an ignored section do not need the prefix. Applications may use this to include additional information in the unit files.

However, the [docs for the `[Service]` section](https://www.freedesktop.org/software/systemd/man/systemd.service.html)  doesn't mention how unknown options are dealt with.

So, a comment makes sense. Then people know what to expect.

c0deaddict: I've commented out the resource control options because I'm afraid that some users will copy it without reading the comments, and thus have their NATS running at 1% CPU limit.

c0deaddict: Can this PR be merged @philpennock?

derekcollison:
@philpennock Can you let us know if this can be merged?

philpennock (approved):
This is great, thanks!

util/nats-server-hardened.service
philpennock: I am very sorry, you deserved a more timely response than this. Poor GitHub notifications management on my part.

Thank you for the contribution and for addressing my feedback, I'm very happy with the result and have given my GH 👍 approval.

PR #2322, opened 2021-06-28
Added ability to match based on last expected sequence per subject.

derekcollison:

/cc @nats-io/core


ripienaar:
LGTM

PR #2323, opened 2021-06-28
Add OCSP cert generator script

variadico:
This adds a script to generate TLS keys and certificates used in OCSP tests.


wallyqs (approved):
LGTM, /cc @philpennock

philpennock:
I'm surprised openssl likes all the leading whitespace, and do we want Derek's email address in there?

I think my biggest gotcha concern is that the hash algorithm seems to be left up to chance and the version of openssl installed, so might be SHA1 which would cause problems.  Might want some `default_md = sha256` in there.

And you could probably simplify a lot, if you wanted to, by using `SUBJECTALTNAME = email:copy` and then setting `$SUBJECTALTNAME` in environ first, instead of having to write out a new config each time.  Eg,

```sh
env SUBJECTALTNAME='IP:127.0.0.1, IP:::1, DNS:localhost, DNS:client.localhost' openssl ca ....
```

Otherwise looks functionally correct ... did you really intend to keep the CA private key around in-repo too?  For the stuff I did, using cfssl, I dropped the CA private key so that if we had to regen, it would be with a new CA, and there was no risk of the tests CA being abused.

If you wanted it to look _good_, I suggest writing a shell function to emit the huge repeated blocks and using here-docs to construct it, which would let you cut out a lot of the repetition and let the code thus draw attention to what's actually different each time.  But that's perhaps too picky: what you have works, so LGTM.

philpennock (approved):

variadico:
> do we want Derek's email address in there?

I just copied that from other certs, but we probably don't need it. Removed! ✅ 

>  Might want some default_md = sha256 in there.

Added! ✅ 

> did you really intend to keep the CA private key around in-repo

Yeah, we need the CA private key in the tests to sign OCSP responses. I also noticed some keys in `test/configs/certs/tlsauth`. Maybe there's a better way to handle this? But not sure how much we want it.

> I suggest writing a shell function to emit the huge repeated blocks and using here-docs to construct it,

Yeah, I agree it's not super DRY. On the other hand, I wasn't sure if I should get too fancy with Bash because it may make it harder for people to read. This way you can read the config from top to bottom and not have to jump around or lookup syntax. Buuuuttt... also agree it makes it harder to notices the differences between the configs. Maybe someone who is actually good at Bash can do this later? 😅

wallyqs (approved):
LGTM, thanks @variadico !

ripienaar:
Is this something we should track for general implementation in all clients?

wallyqs:
@ripienaar maybe, right now in Go verification is done opt-in by setting it via [nats.Secure option passing the TLS config](https://github.com/nats-io/nats-server/blob/main/test/ocsp_test.go#L86-L98). I think what we can do in the clients is to try to honor OCSP Must Staple if present automatically and verify connections.

ripienaar:
Ok. If you two review https://github.com/nats-io/nats-architecture-and-design/pull/3 I can merge that and we can open something. Be good to have a ADR also explaining all this. There is a  bunch of tribal knowledge :)

PR #2324, opened 2021-06-29
Improvements to large numbers of JetStream R1 consumers per stream.

derekcollison:
1. We were holding open FDs longer than we should for consumers causing issues with open FD limits. We now do not hold them open and cap updates a bit better.

2. When doing a stream delete, consumer delete was repeating alot of work that was not necessary, causing longer delays. This has been optimized a bit, still more improvements to be made.

3. We cover all JS under a single export, but that was also trapping GetNext for pull based consumers, and since this was a no-op (is handled at user account level) we were creating alot of garbage service import responses and reverse map entries that had to be garbage collected. We have a fix in to avoid this but still looking for a better one.

4. Still had some lingering references to all exports vs single JS export.

/cc @nats-io/core


ripienaar (approved):
seems like good improvements, but I dont know 1/5th of this code

derekcollison:
Understood, was more FYI, will wait on @kozlovic

server/jetstream_api.go
```diff
@@ -633,6 +608,7 @@ func (js *jetStream) apiDispatch(sub *subscription, c *client, subject, reply st
 
 	// Shortcircuit.
 	if len(rr.psubs)+len(rr.qsubs) == 0 {
+		debug.PrintStack()
```
kozlovic, line 611: Remove

kozlovic (approved):
LGTM except for left over debug.PrintStack()

server/jetstream_api.go
derekcollison: Fixed.

PR #2330, opened 2021-06-30
adr documenting the idle_heartbeat feature for consumers

aricart:
/cc @nats-io/core

### DIFF
--- a/doc/adr/0009-js-idle-heartbeat.md
+++ b/doc/adr/0009-js-idle-heartbeat.md
@@ -47,6 +47,6 @@ recovered elsewhere.
 
 ## Consequences
 
-In the future, clients will use this information to set client-side timers that
-track how many heartbeats have been missed and perhaps take some action such as
-re-create a subscription to resume messages.
+Clients can use this information to set client-side timers that track how many
+heartbeats have been missed and perhaps take some action such as re-create a
+subscription to resume messages.
--- a/doc/adr/0009-js-idle-heartbeat.md
+++ b/doc/adr/0009-js-idle-heartbeat.md
@@ -1,0 +1,52 @@
+# 9. js-idle-heartbeat
+
+Date: 2021-06-30
+
+## Status
+
+Accepted
+
+## Context
+
+The JetStream ConsumerConfig option `idle_heartbeat` enables server-side
+heartbeats to be sent to the client. To enable the option on the consumer simply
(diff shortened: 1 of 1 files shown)


doc/adr/0009-js-idle-heartbeat.md
```diff
@@ -0,0 +1,52 @@
...
+specify it with a value representing the number of nanoseconds that the server
+should use as notification interval.
+
+The server will only notify after the specified interval has elapsed and no new
+messages have been delivered to the consumer. Delivering a message to the
+consumer resets the interval.
+
+The idle heartbeats notifications are sent to the consumer's subscription as a
+regular NATS message. The message will have a `code` of `100` with a
+`description` of `Idle Heartbeat`. The message will contain additional headers
```
derekcollison, line 22: I think these are just part of the reply subject meta IIRC.

derekcollison: The sequence numbers etc I mean.

doc/adr/0009-js-idle-heartbeat.md
```diff
@@ -0,0 +1,52 @@
...
+```
+
+This feature is intended as an aid to clients to detect when they have been
+disconnected. Without it the consumer's subscription may sit idly waiting for
+messages, without knowing that the server might have simply gone away and
+recovered elsewhere.
+
+## Consequences
+
+In the future, clients will use this information to set client-side timers that
```
derekcollison, line 50: This already working for Go client, so maybe

Clients can choose to implement auto handling of a situation where the heartbeat detects we lost the last N messages etc.

derekcollison (approved):

doc/adr/0009-js-idle-heartbeat.md
ripienaar, line 22: might be worth calling out that the `description` isnt actually a promise or part of the API, the code is of course.

ripienaar (commented):
> This already working for Go client, so maybe ...

The go client does not set a timer though, the go client uses the heartbeats to track sequences and reset in case messages were lost.

It does not yet use these heartbeats to recover from failed ephemerals which is what the `Consequences` here is refering to I think?

Essentially without handling this we recreated the bug that led to https://github.com/nats-io/stan.go#connection-status

PR #2331, opened 2021-06-30
adr documenting purge and options

aricart:
/cc @nats-io/core

### DIFF
--- a/doc/adr/0010-js-purge.md
+++ b/doc/adr/0010-js-purge.md
@@ -1,0 +1,67 @@
+# 10. js-purge
+
+Date: 2021-06-30
+
+## Status
+
+Accepted
+
+## Context
+
+JetStream provides the ability to purge streams by sending a request message to:
+`$JS.API.STREAM.PURGE.<streamName>`. The request will return a new message with
+the following JSON:
+
+```typescript
+{
+    type: "io.nats.jetstream.api.v1.stream_purge_response", 
+    error?: {
+        code: number,
+        err_code: number,
+        description: string
+    },
+    success: boolean,
+    purged: number
+}
+```
+
+The `error` field in the response will only be set if there was an error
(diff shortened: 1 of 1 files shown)


doc/adr/0010-js-purge.md
```diff
@@ -0,0 +1,67 @@
...
+
+Accepted
+
+## Context
+
+JetStream provides the ability to purge streams by sending a request message to:
+`$JS.API.STREAM.PURGE.<streamName>`. The request will return a new message with
+the following JSON:
+
+```typescript
```
ripienaar, line 15: be better to show json here

doc/adr/0010-js-purge.md
```diff
@@ -0,0 +1,67 @@
...
+    },
+    success: boolean,
+    purged: number
+}
+```
+
+The `error` field in the response will only be set if there was an error
+processing the request. The `success` field will be set to `true` if the request
+succeeded. The `purged` field will be set to the number of messages that were
+purged from the stream.
```
ripienaar, line 31: not sure every ADR has to explain this, since its how every single API response works and (in theory) covered by the overall api design ADR?

doc/adr/0010-js-purge.md
```diff
@@ -0,0 +1,67 @@
...
+
+```typescript
+{
+    seq?: number,
+    keep?: number,
+    filter?: string
+}
+```
+
+- `seq` is the optional upper-bound sequence for messages to be deleted
```
ripienaar, line 46: maybe call out specfically that it can be combined with `filter`

doc/adr/0010-js-purge.md
aricart: The syntax showing things like: `{foo?: string, bar: number}` is fairly intuitive, and conveys that field `foo` is optional, while `bar` is not - the `typescript` directive colors it properly.

PR #2333, opened 2021-06-30
Check not really needed here, so removed

derekcollison:
Was showing a data race.

/cc @nats-io/core


ripienaar (approved):
LGTM

PR #2336, opened 2021-07-01
Cleanup references to 'master' and 'gnatsd'

philpennock:
Cleanup references to 'master' and 'gnatsd'
    
 * We've renamed our default branch from 'master' to 'main'
 * We switched some time ago from 'gnatsd' to 'nats-server' and some temporary redirects no longer work, so fix those too


derekcollison (approved):
LGTM

PR #2337, opened 2021-07-01
Should not flush in place for stream loop.

derekcollison:

/cc @nats-io/core

### DIFF
--- a/server/stream.go
+++ b/server/stream.go
@@ -2991,7 +2991,7 @@ func (mset *stream) internalLoop() {
 					pm.o.didNotDeliver(pm.seq)
 				}
 			}
-			c.flushClients(10 * time.Millisecond)
+			c.flushClients(0)
 		case <-mch:
 			for im := mset.pending(mset.msgs); im != nil; im = im.next {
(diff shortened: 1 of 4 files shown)


server/stream.go
```diff
@@ -2991,7 +2991,7 @@ func (mset *stream) internalLoop() {
 					pm.o.didNotDeliver(pm.seq)
 				}
 			}
-			c.flushClients(10 * time.Millisecond)
+			c.flushClients(0)
```
kozlovic, line 2994: I think we should also look at the sendq here: https://github.com/nats-io/nats-server/blob/bd5d5d0ff89516c8702c9a0e31c004b8da315c63/server/sendq.go#L87

kozlovic (approved):
LGTM but may want to check the other place where we do that.

server/stream.go
derekcollison: Good call. Agree.

derekcollison: ok done.

derekcollison:
Should I squash?

kozlovic (approved):
LGTM

PR #2338, opened 2021-07-01
[fixed] header handling in system services

matthiashanel:
On export/import a header was inserted which broke parsing of the
message.
Fixed unit test broken by .beta in version


kozlovic (approved):
LGTM

PR #2341, opened 2021-07-04
[ADDED] ReloadOptions() for embedded cases

### LINKED ISSUE #2272: Reload programmatically defined options
## Feature Request

Provide a way to dynamically load a new configuration provided programmatically.

#### Use Case:

Same use case a for Server.Reload, but programmatically (without a configuration file).

#### Proposed Change:

Either by providing a Server.Load(Option) or by providing public versions of the options (implementing the hot-swap interface)

In case of a Server.Load(Option), we may have the need of a function to get the full options, as Option must have all its properties set it would be difficult to change only a single option without that.

The option.Reload(Server) version is more specific and therefore probably less error-prone. It also makes sense as when configuration is given programmatically, granularity would generally be higher.

nats-server users that don't use a configuration file

taigrr:
 - [x] Tests added (TBD - contingent on core team weigh-in)

Resolves #2272

- Provide a public interface to Reload() the server from an Options struct instead of a changed config file. This is useful when embedding.

Testing to be completed (and commits squashed) once appropriately defined by other contributors
This is a minor addition

/cc @nats-io/core

### DIFF
--- a/server/reload.go
+++ b/server/reload.go
@@ -701,10 +701,29 @@ func (s *Server) recheckPinnedCerts(curOpts *Options, newOpts *Options) {
 	}
 }
 
-// Reload reads the current configuration file and applies any supported
-// changes. This returns an error if the server was not started with a config
-// file or an option which doesn't support hot-swapping was changed.
+// Reload reads the current configuration file and calls out to ReloadOptions
+// to apply the changes. This returns an error if the server was not started
+// with a config file or an option which doesn't support hot-swapping was changed.
 func (s *Server) Reload() error {
+	s.mu.Lock()
+	configFile := s.configFile
+	s.mu.Unlock()
+	if configFile == "" {
+		return errors.New("can only reload config when a file is provided using -c or --config")
+	}
+
+	newOpts, err := ProcessConfigFile(configFile)
+	if err != nil {
+		// TODO: Dump previous good config to a .bak file?
+		return err
+	}
+	return s.ReloadOptions(newOpts)
+}
+
+// ReloadOptions applies any supported options from the provided Option
+// type. This returns an error if an option which doesn't support
+// hot-swapping was changed.
+func (s *Server) ReloadOptions(newOpts *Options) error {
 	s.mu.Lock()
 
 	s.reloading = true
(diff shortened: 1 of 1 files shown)


taigrr:
The Travis failure appears unrelated to this PR.

server/reload.go
```diff
@@ -788,6 +788,80 @@ func (s *Server) Reload() error {
 	return nil
 }
 
+// ReloadOptions applies any supported options from the provided Option
+// type. This returns an error if an option which doesn't
+// support hot-swapping was changed.
+func (s *Server) ReloadOptions(newOpts *Options) error {
```
kozlovic, line 794: This is a lot of repeat with Reload(), so what I would suggest is that this one could be as-is (assuming it does the same than Reload()) and change Reload() to first load the configuration file and then invoke this new ReloadOptions() function. I think that processing config file could be done outside of server lock, something like:
```
func (s *Server) Reload() error {
	s.mu.Lock()
	configFile := s.configFile
	s.mu.Unlock()
	if configFile == "" {
		return errors.New("can only reload config when a file is provided using -c or --config")
	}

	newOpts, err := ProcessConfigFile(configFile)
	if err != nil {
		return err
	}

	return s.ReloadOptions(newOpts)
}
```

kozlovic (changes requested):

server/reload.go
taigrr: Perfect! I agree completely. Doing things this way would also help answer my question on test coverage (as current tests on Reload() would pass through). The only reason I didn't do it this way the first time was my hesitancy to move things outside of the mutex, since things are usually bounded by mutexes for a reason (even though I couldn't immediately see why). 

I will update and squash this PR soon.

taigrr: @kozlovic I've made the changes suggested.

taigrr:
Whoops. Needed another rebase. Ready now!

kozlovic (approved):
LGTM

kozlovic:
Will let @ripienaar chime in, but I think this is ok and ready to merge.

taigrr:
@kozlovic do you want me to rebase again to rewrite the commit message to match the new title?

kozlovic:
@taigrr No need.

ripienaar (approved):
LGTM

PR #2344, opened 2021-07-06
[FIXED] Use of sync.Pool for filestore could cause msg corruption.

derekcollison:

/cc @nats-io/core

### DIFF
--- a/server/norace_test.go
+++ b/server/norace_test.go
@@ -2216,3 +2216,83 @@ func TestNoRaceJetStreamSlowFilteredInititalPendingAndFirstMsg(t *testing.T) {
 	testConsumerCreate("foo.100000", 1, 1)
 	testConsumerCreate("foo", 1, 100_000-2)
 }
+
+func TestNoRaceJetStreamFileStoreBufferReuse(t *testing.T) {
+	// Uncomment to run. Needs to be on a big machine.
+	skip(t)
+
+	s := RunBasicJetStreamServer()
+	defer s.Shutdown()
+
+	if config := s.JetStreamConfig(); config != nil {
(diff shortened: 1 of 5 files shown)


server/norace_test.go
```diff
@@ -2216,3 +2216,83 @@ func TestNoRaceJetStreamSlowFilteredInititalPendingAndFirstMsg(t *testing.T) {
 	testConsumerCreate("foo.100000", 1, 1)
 	testConsumerCreate("foo", 1, 100_000-2)
 }
+
+func TestNoRaceJetStreamFileStoreBufferReuse(t *testing.T) {
+	// Uncomment to run. Needs to be on a big machine.
+	//skip(t)
```
kozlovic, line 2222: Should we skip by default?

server/events.go
```diff
@@ -721,37 +721,37 @@ func (s *Server) initEventTracking() {
 	}
 	monSrvc := map[string]msgHandler{
 		"STATSZ": s.statszReq,
-		"VARZ": func(sub *subscription, _ *client, subject, reply string, msg []byte) {
+		"VARZ": func(sub *subscription, c *client, subject, reply string, msg []byte) {
```
kozlovic, line 724: These are changes that were done in https://github.com/nats-io/nats-server/pull/2338. Should should have rebased instead so that this PR shows only your changes? (it's ok if you don't)

kozlovic (approved):
LGTM, but wonder about the new test being ON by default.

server/norace_test.go
derekcollison: Yes will turn off by default I think. Thanks.

kozlovic (approved):
LGTM

PR #2345, opened 2021-07-06
Add error codes for all consumer creation errors

ripienaar:
I wanted to supress some logging of consumer create
errors that just isn't needed and would be really
annoying on large networks, so I added many constants
and updated all errors.

I think only JSConsumerStoreFailedErrF is worth logging
on large networks else there would be quite a lot of
logs generated that one just cannot act on

/cc @nats-io/core


derekcollison (approved):
LGTM

PR #2346, opened 2021-07-06
[FIXED] #2332 Peer info reports that had large last active values.

### LINKED ISSUE #2332: PeerInfo Lag contains exceptionally large uint64 value
## Defect

THIS ISSUE IS AWAITING MORE INFO

Replica.PeerInfo.Lag value is `18446744073709551614` (`fffffffffffffffe`) which seems like an overflow?

From a user:
> This issue occurs especially when we purge some data in stream and starting sequence number is other than 1 for ephemeral consumer with deliver policy as all.

Server:
Java Client: 2.11.4

derekcollison:
Also put in safety for lag going upside down as well.

Resolves #2332 

/cc @nats-io/core

### DIFF
--- a/server/jetstream_cluster.go
+++ b/server/jetstream_cluster.go
@@ -4665,9 +4665,9 @@ func (js *jetStream) clusterInfo(rg *raftGroup) *ClusterInfo {
 
 	for _, rp := range peers {
 		if rp.ID != id && rg.isMember(rp.ID) {
-			lastSeen := now.Sub(rp.Last)
-			if lastSeen < 0 {
-				lastSeen = 1
+			var lastSeen time.Duration
+			if now.After(rp.Last) && rp.Last.Unix() != 0 {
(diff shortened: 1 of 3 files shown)


server/jetstream_cluster.go
```diff
@@ -4665,9 +4665,9 @@ func (js *jetStream) clusterInfo(rg *raftGroup) *ClusterInfo {
 
 	for _, rp := range peers {
 		if rp.ID != id && rg.isMember(rp.ID) {
-			lastSeen := now.Sub(rp.Last)
-			if lastSeen < 0 {
-				lastSeen = 1
+			lastSeen := time.Microsecond // Just sane small non-zero default.
```
ripienaar, line 4668: so when not seen at all this will now give some microseconds? Probably ok just chekcing

kozlovic (approved):
LGTM

PR #2347, opened 2021-07-06
Make requirement for interest on creation of ephemerals soft vs hard.

derekcollison:
Will start the cleanup timer if no interest but allow creation.

/cc @nats-io/core


ripienaar (approved):
LGTM

starting to think HBs on ephemerals should have been required, too late now.

kozlovic (approved):
LGTM

PR #2354, opened 2021-07-08
Fix for multiple concurrent ephemeral consumer requests with max consumers set.

derekcollison:
Bug that did not limit ephemeral consumers when in clustered mode and with multiple concurrent requests.

/cc @nats-io/core

### DIFF
--- a/server/jetstream_cluster_test.go
+++ b/server/jetstream_cluster_test.go
@@ -7640,6 +7640,59 @@ func TestJetStreamClusterMaxConsumers(t *testing.T) {
 	}
 }
 
+func TestJetStreamClusterMaxConsumersMultipleConcurrentRequests(t *testing.T) {
+	c := createJetStreamClusterExplicit(t, "JSC", 3)
+	defer c.shutdown()
+
+	nc, js := jsClientConnect(t, c.randomServer())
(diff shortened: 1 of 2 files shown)


server/jetstream_cluster_test.go
```diff
@@ -7640,6 +7640,59 @@ func TestJetStreamClusterMaxConsumers(t *testing.T) {
...
+	var wg sync.WaitGroup
+
+	for n := 0; n < 10; n++ {
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			nc, js := jsClientConnect(t, c.randomServer())
+			defer nc.Close()
+			<-startCh
+			js.SubscribeSync("in.maxcc.foo")
```
ripienaar, line 7679: have to be careful that this doesnt unsubscrie (nc.Close()) before the other go routine is going on, my sample test had a wait on a channel there to prevent the defers from firing

derekcollison: Did not affect my tests, was getting failure everytime.

ripienaar: ok :)

ripienaar:
LGTM

PR #2357, opened 2021-07-10
[FIXED] Consumer NumPending bug

derekcollison:
Fix for a consumer's num pending being stuck at 1.

We were trying to protect the sgap uint64 from wrapping, but in some cases the consumers is eager and can get a message before we sgap++.

Instead of slowing things down and synchronizing ++ then --, we allow it to wrap temporarily and have and adjustedPending() func that will set to zero for reporting.

/cc @nats-io/core


ripienaar (approved):
LGTM

PR #2360, opened 2021-07-13
[ADDED] In-process connections

neilalexander:
This PR adds three things:
1.  `InProcessConn()` function to `Server` which builds a `net.Pipe` to get a connection to the NATS server without using TCP sockets
2. `DontListen` option which tells the NATS server not to listen on the usual TCP listener
3. `startupComplete` channel, which is closed right before we start `AcceptLoop`, and `readyForConnections` will wait for it

The main motivation for this is that we have an application that can run either in a monolith (single-process) mode or a polylith (multi-process) mode. We'd like to be able to use NATS for both modes for simplicity, but the monolith mode has to be able to cater for a variety of platforms where opening socket connections either doesn't make sense (mobile) or just isn't possible (WASM). These changes will allow us to use NATS entirely in-process instead.

An accompanying PR nats-io/nats.go#774 adds support to the client side.

This is my first PR to this project so apologies in advance if I've missed anything obvious anywhere.

/cc @nats-io/core

### DIFF
--- a/server/opts.go
+++ b/server/opts.go
@@ -199,6 +199,7 @@ type Options struct {
 	ServerName            string        `json:"server_name"`
 	Host                  string        `json:"addr"`
 	Port                  int           `json:"port"`
+	DontListen            bool          `json:"dont_listen"`
 	ClientAdvertise       string        `json:"-"`
 	Trace                 bool          `json:"-"`
 	Debug                 bool          `json:"-"`
@@ -4283,6 +4284,9 @@ func MergeOptions(fileOpts, flagOpts *Options) *Options {
 	}
 	if flagOpts.Host != "" {
 		opts.Host = flagOpts.Host
+	}
+	if flagOpts.DontListen {
+		opts.DontListen = flagOpts.DontListen
 	}
 	if flagOpts.ClientAdvertise != "" {
 		opts.ClientAdvertise = flagOpts.ClientAdvertise
--- a/server/server.go
+++ b/server/server.go
@@ -2948,6 +2948,8 @@ func (s *Server) ProfilerAddr() *net.TCPAddr {
 }
 
 func (s *Server) readyForConnections(d time.Duration) error {
+	end := time.Now().Add(d)
+
 	// Wait for startup. At this point AcceptLoop will only just be
 	// starting, so we'll still check for the presence of listeners
 	// after this.
@@ -2968,7 +2970,6 @@ func (s *Server) readyForConnections(d time.Duration) error {
 	}
 	chk := make(map[string]info)
 
-	end := time.Now().Add(d)
 	for time.Now().Before(end) {
 		s.mu.Lock()
 		chk["server"] = info{ok: s.listener != nil || opts.DontListen, err: s.listenerErr}
--- a/server/server.go
+++ b/server/server.go
@@ -2119,6 +2119,7 @@ func (s *Server) InProcessConn() (net.Conn, error) {
 	pl, pr := net.Pipe()
 	if !s.startGoRoutine(func() {
 		s.createClient(pl)
+		s.grWG.Done()
 	}) {
 		pl.Close()
 		pr.Close()
--- a/server/server.go
+++ b/server/server.go
@@ -167,6 +167,7 @@ type Server struct {
 	leafNodeEnabled    bool
 
 	quitCh           chan struct{}
+	startupComplete  chan struct{}
 	shutdownComplete chan struct{}
 
 	// Tracking Go routines
@@ -495,6 +496,10 @@ func NewServer(opts *Options) (*Server, error) {
 
 	// For tracking leaf nodes.
 	s.leafs = make(map[uint64]*client)
+
+	// Closed when startup is complete. Allows WaitForStartup() to block
+	// waiting until NATS has started.
+	s.startupComplete = make(chan struct{})
 
 	// Used to kick out all go routines possibly waiting on server
 	// to shutdown.
@@ -1613,6 +1618,7 @@ func (s *Server) Start() {
 	}
 	s.Noticef("  ID:       %s", s.info.ID)
 
+	defer close(s.startupComplete)
 	defer s.Noticef("Server is ready")
 
 	// Check for insecure configurations.
@@ -1849,7 +1855,9 @@ func (s *Server) Start() {
 	}
 
 	// Wait for clients.
-	s.AcceptLoop(clientListenReady)
+	if !opts.DontListen {
+		s.AcceptLoop(clientListenReady)
+	}
 }
 
 // Shutdown will shutdown the server instance by kicking out the AcceptLoop
@@ -2021,6 +2029,11 @@ func (s *Server) Shutdown() {
 	close(s.shutdownComplete)
 }
 
+// WaitForStartup will block until the server has been fully started.
+func (s *Server) WaitForStartup() {
+	<-s.startupComplete
+}
+
 // WaitForShutdown will block until the server has been fully shutdown.
 func (s *Server) WaitForShutdown() {
 	<-s.shutdownComplete
@@ -2098,6 +2111,27 @@ func (s *Server) AcceptLoop(clr chan struct{}) {
 	// Let the caller know that we are ready
 	close(clr)
 	clr = nil
+}
+
+// InProcessConn returns a connection to the server which can
+// be used in-process when DontListen is set.
+func (s *Server) InProcessConn() (net.Conn, error) {
+	pl, pr := net.Pipe()
+	created := true
+	if !s.startGoRoutine(func() {
+		if c := s.createClient(pl); c == nil {
+			created = false
+		}
+	}) {
+		created = false
+	}
+	if !created {
+		s.grWG.Done()
+		_ = pl.Close()
+		_ = pr.Close()
+		return nil, fmt.Errorf("failed to make connection")
+	}
+	return pr, nil
 }
 
 func (s *Server) acceptConnections(l net.Listener, acceptName string, createFunc func(conn net.Conn), errFunc func(err error) bool) {
--- a/server/server.go
+++ b/server/server.go
@@ -167,7 +167,6 @@ type Server struct {
 	leafNodeEnabled    bool
 
 	quitCh           chan struct{}
-	startupComplete  chan struct{}
 	shutdownComplete chan struct{}
 
 	// Tracking Go routines
@@ -496,10 +495,6 @@ func NewServer(opts *Options) (*Server, error) {
 
 	// For tracking leaf nodes.
 	s.leafs = make(map[uint64]*client)
-
-	// Closed when startup is complete. Allows WaitForStartup() to block
-	// waiting until NATS has started.
-	s.startupComplete = make(chan struct{})
 
 	// Used to kick out all go routines possibly waiting on server
 	// to shutdown.
@@ -1618,7 +1613,6 @@ func (s *Server) Start() {
 	}
 	s.Noticef("  ID:       %s", s.info.ID)
 
-	defer close(s.startupComplete)
 	defer s.Noticef("Server is ready")
 
 	// Check for insecure configurations.
@@ -2029,11 +2023,6 @@ func (s *Server) Shutdown() {
 	close(s.shutdownComplete)
 }
 
-// WaitForStartup will block until the server has been fully started.
-func (s *Server) WaitForStartup() {
-	<-s.startupComplete
-}
-
 // WaitForShutdown will block until the server has been fully shutdown.
 func (s *Server) WaitForShutdown() {
 	<-s.shutdownComplete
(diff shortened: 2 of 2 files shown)


server/server.go
```diff
@@ -1960,6 +1973,27 @@ func (s *Server) AcceptLoop(clr chan struct{}) {
 	clr = nil
 }
 
+// InProcessConn returns a connection to the server which can
+// be used in-process when DontListen is set.
```
ripienaar, line 1977: This seem to be usable even when DontListen is not set though right, that's good I'd be keen on this feature also but I want a mix or listening sockets and inprocess connections. So maybe comment needs an update?

ripienaar (commented):
love this, been keen to have something like this for quite some time :)

server/server.go
neilalexander: It is indeed usable either way, so I'll update the comment!

kozlovic (changes requested):
Thank you for the contribution!

server/opts.go
```diff
@@ -184,6 +184,7 @@ type Options struct {
 	ServerName            string        `json:"server_name"`
 	Host                  string        `json:"addr"`
 	Port                  int           `json:"port"`
+	DontListen            bool          `json:"dont_listen"`
```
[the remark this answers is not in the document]
neilalexander: I suppose this could work, although it does create the risk of the API shape having more "magic numbers". I'll be happy to accept direction on this from whoever has strong enough opinions on it.

server/server.go
```diff
@@ -1960,6 +1973,29 @@ func (s *Server) AcceptLoop(clr chan struct{}) {
...
+	created := true
+	if !s.startGoRoutine(func() {
+		if c := s.createClient(pl); c == nil {
+			created = false
+		}
+	}) {
+		created = false
+	}
+	if !created {
+		s.grWG.Done()
```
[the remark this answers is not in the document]
neilalexander: I've obviously misunderstood the purpose of this wait group so will remove this.

server/server.go
```diff
@@ -1881,6 +1889,11 @@ func (s *Server) Shutdown() {
 	close(s.shutdownComplete)
 }
 
+// WaitForStartup will block until the server has been fully started.
+func (s *Server) WaitForStartup() {
```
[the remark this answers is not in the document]
neilalexander: OK, made the change to this instead.

server/server.go
```diff
@@ -1960,6 +1973,29 @@ func (s *Server) AcceptLoop(clr chan struct{}) {
...
+// InProcessConn returns an in-process connection to the server,
+// avoiding the need to use a TCP listener for local connectivity
+// within the same process. This can be used regardless of the
+// state of the DontListen option.
+func (s *Server) InProcessConn() (net.Conn, error) {
+	pl, pr := net.Pipe()
+	created := true
+	if !s.startGoRoutine(func() {
+		if c := s.createClient(pl); c == nil {
+			created = false
```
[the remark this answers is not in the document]
neilalexander: Alright, I've updated it. The semantics around `createClient` weren't terribly clear.

kozlovic (approved):

kozlovic (commented):
The changes look good to me. Let's see if other maintainers have an opinion on the DontListen vs special port number.

ripienaar:
The magic ports seem a bit unclear and hard to discover to me, otoh we have too many settings already but I think a documented setting might be better.

neilalexander:
Hm, ultimately it doesn't appear as if the `ReadyForConnections` solution works, I guess because none of the `chk` cases fire in a simple setup and therefore we return after the first 25ms pause, regardless of whether NATS is set up or not by that point.

It might make sense to re-add the `startupComplete` channel, close it right before the `AcceptLoop` would start and then wait for that in `ReadyForConnections` before checking the sockets.

neilalexander:
I've gone ahead and done that. There's no new API surface this time — `readyForConnections` will instead wait for the startup channel to close, which will happen right before `AcceptLoop` will be called. I hope that's OK.

server/server.go
```diff
@@ -1960,6 +1971,22 @@ func (s *Server) AcceptLoop(clr chan struct{}) {
...
 }
 
+// InProcessConn returns an in-process connection to the server,
+// avoiding the need to use a TCP listener for local connectivity
+// within the same process. This can be used regardless of the
+// state of the DontListen option.
+func (s *Server) InProcessConn() (net.Conn, error) {
+	pl, pr := net.Pipe()
+	if !s.startGoRoutine(func() {
+		s.createClient(pl)
```
kozlovic, line 1981: It should be this:
```
s.createClient(pl)
s.grWG.Done()
```
I mentioned that the call of `s.grWG.Done()` when startGoRoutine() failed needed to be removed, but failed to mention that it has to be done once the function started in startGoRoutine() finishes.

server/server.go
```diff
@@ -2734,6 +2761,17 @@ func (s *Server) ProfilerAddr() *net.TCPAddr {
 }
 
 func (s *Server) readyForConnections(d time.Duration) error {
+	// Wait for startup. At this point AcceptLoop will only just be
+	// starting, so we'll still check for the presence of listeners
+	// after this.
+	select {
```
kozlovic, line 2767: Not sure if this is that important, but maybe we should "deduct" from `d` the time spent in this select. Again, not sure if this is worth it.

kozlovic (changes requested):
I had a go with this change and bringing the changes in the nats.go to create a quick test in the server package and my main concern is the behavior of the net.Pipe when one side closes the connection.

It appears that it results in internal pipe write timeout, which is not that good. For instance, I tried to connect with wrong user/password and the server would report a WriteDeadline exceeded (when trying to write the authorization violation error).

Without more testing (which is going to be a bit annoying since it would require changes from the client that I would not want to commit if we end-up rejecting the whole idea), I think we can't merge this PR.
If you could have some tests and figure out what the issue is with those time outs, then we could reconsider.

neilalexander:
> It appears that it results in internal pipe write timeout, which is not that good. For instance, I tried to connect with wrong user/password and the server would report a WriteDeadline exceeded (when trying to write the authorization violation error).

Looking at the source for `net.Pipe`, the only way it can return a `os.ErrDeadlineExceeded` is if a deadline has been set and not cleared or re-set afterwards. Otherwise it should return an `io.ErrClosedPipe`.

https://github.com/golang/go/blob/d8f348a589b3df4bb48636023be4ccf9ac96d307/src/net/pipe.go#L183-L188

I haven't made any changes to deadlines anywhere so wondering if there is some existing deadline being set elsewhere?

> If you could have some tests and figure out what the issue is with those time outs, then we could reconsider.

Have you got an example or a specific test that I can look at? We've been experimenting with in-process conns using these commits in our codebase for the last couple of days and we have yet to see this.

server/server.go
neilalexander: I suppose moving the `end := time.Now().Add(d)` declaration to before the `select` should have this effect.

kozlovic: Probably.

server/server.go
neilalexander: Have added this in.

server/server.go
neilalexander: Have done this.

server/server.go
kozlovic: I don't see this in the commit that says you did (just the move of `end := ..`

neilalexander: Oh, what. VS Code apparently decided it wanted to block on the language server and not save changes, so that's incredibly helpful.

kozlovic:
> Have you got an example or a specific test that I can look at? 

I was trying to connect with wrong username/password, so what I see is that the server blocks on flushOutbound() writing `-ERR 'Authorization Violation'\r\n` message (up to write deadline which default is 10 seconds) and the client blocks in writing the CONNECT + PING protocols in nats.go sendConnect() at this place: `if err := nc.bw.writeDirect(cProto, pingProto); err != nil {`.

So the server reads the CONNECT (not the PING yet), fails the client by trying to write `-ERR ..` to it and that blocks because the client is not reading. The connect process in the client is synchronous, that is, there is no read loop started yet.

So if I have the client sends only CONNECT and then PING later, the issue is the same, this time the client blocks writing PING because the server is also blocked writing `-ERR` (since client is not reading). Having buffered pipes may help, but to the point where buffer will be full. At least that may help pass the phase of CONNECT...

variadico:
> Hm, ultimately it doesn't appear as if the ReadyForConnections solution works

Just curious, what is failing? Is there a test that consistently fails? Sometimes tests can be flaky, so you may need to rerun them a few times.

kozlovic:
@variadico No, it was just that in the case of "DontListen" (part of this PR), ReadyForConnections() should ignore the client port, so I asked to add `ok: s.listener != nil || opts.DontListen`, but the issue was that user calling ReadyForConnections() in that case would return immediately with "ok", while the user wanted at least all of Server.Start() (minus AcceptLoop) to be done, hence the use of the startupComplete channel that is just closed prior to starting the AcceptLoop() (which is blocking).

kozlovic:
@neilalexander We will discuss this PR internally and come up with a decision maybe later next week. Thanks again!

kozlovic (approved):
LGTM. Switching to Draft until we decide internally (so it does not get merged by mistake).

neilalexander:
Thanks @kozlovic, will wait to hear back on a way forward. Hoping we can come up with something suitable — NATS is shaping up very well for us so far, we just need to be able to get it to work in obscure places like WASM as well as cross-process!

neilalexander:
Is there any update on whether this was discussed further? We're optimistically hoping for good news. 😄

derekcollison:
@kozlovic and I plan to discuss this week IIRC.

kozlovic:
@neilalexander Sorry for the delay. We just had a discussion today about this, and we feel it is important and part of future kind of deployments, but we would need a bit more information. The wasm limitation may be temporary, are you following [wasi](https://wasi.dev/) developments?
What wasm environment are you planning on deploying this?

Finally, would you have 15/20 minutes to speak with @derekcollison? He would like to have a short discussion with you. Thanks!

derekcollison:
We are starting to see more interest in this and we ourselves are very interested in the WASM/WASI ecosystem. Possible to get 20 minutes of your time to chat live via Zoom or GH?

neilalexander:
Hey @kozlovic, @derekcollison — more than happy to have a chat. I'm in GMT but please feel free to drop me a mail at `neilalexander` at `matrix.org` (or, if you are Matrix natives, the same MXID) to arrange a time.

At a high level, our project is [Dendrite](https://github.com/matrix-org/dendrite), which is a Matrix homeserver implementation. We basically need to be able to run everywhere — within the web browser, on mobile devices and on servers alike. We see the in-process connections as a way of getting around the limitations with not being able to open arbitrary sockets with WASM (and this probably still holds true in WASI as of today, not sure?) and also not opening unnecessary sockets on mobile devices either. I can explain more certainly about what we're hoping to do on the call.

derekcollison:
Neil,

Have not forgotten about you, promise.

neilalexander:
Hey Derek,

Is there any update? Cheers!

derekcollison:
Still very much on our list, just prioritizing KV and ObjectStore abstractions atm.

neilalexander:
Hi all, is there any update on this?

derekcollison:
Unfortunately nothing has changed atm.

Still a nice to have I think for us but not a high priority.

derekcollison:
@neilalexander if you want to update this to the main branch (rebase etc). We can see if we can get this one in..

neilalexander:
@derekcollison Have rebased against `main`!

derekcollison:
Thanks! We still have the hang issues that were mentioned yes?

neilalexander:
I’m not entirely sure *what* issues it might introduce, but we’ve been using this in Dendrite for months without any particular issues. Admittedly we don’t ever *stop* NATS or close the connection until the point that the entire process shuts down. We’ve also seen one or two cases of `slow consumer` but I suspect that’s more to do with our message sizes as we’ve also seen it when running NATS in a separate process via TCP.

server/server.go
```diff
@@ -2930,10 +2971,9 @@ func (s *Server) readyForConnections(d time.Duration) error {
 	}
 	chk := make(map[string]info)
 
-	end := time.Now().Add(d)
 	for time.Now().Before(end) {
-		s.mu.RLock()
-		chk["server"] = info{ok: s.listener != nil, err: s.listenerErr}
+		s.mu.Lock()
```
kozlovic, line 2975: You need to change that to an RLock(). This has changed in main so that is causing issue now with your PR.

server/opts.go
```diff
@@ -4284,6 +4285,9 @@ func MergeOptions(fileOpts, flagOpts *Options) *Options {
 	if flagOpts.Host != "" {
 		opts.Host = flagOpts.Host
 	}
+	if flagOpts.DontListen {
+		opts.DontListen = flagOpts.DontListen
```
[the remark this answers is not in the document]
kozlovic: No, that's ok.

kozlovic (changes requested):

server/server.go
neilalexander: Fixed!

kozlovic (approved):
LGTM

PR #2363, opened 2021-07-14
Stubs for WebAssembly

neilalexander:
This should allow NATS Server to build under the `js/wasm` target.

/cc @nats-io/core

### DIFF
--- a/server/disk_avail_wasm.go
+++ b/server/disk_avail_wasm.go
@@ -1,4 +1,4 @@
-// Copyright 2020 The NATS Authors
+// Copyright 2022 The NATS Authors
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
 // You may obtain a copy of the License at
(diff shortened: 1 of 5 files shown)


derekcollison:
Should we be tracking this one still @neilalexander?

neilalexander:
@derekcollison If you don't mind — this PR pretty much got NATS to build for WASM but I am not 100% sure of the consequences of stubbing some of these values.

gedw99:
> 
@derekcollison @neilalexander 
So then needs built tags, so that the code can work in browser and not in browser right ? 
Then at compile time just use the flags appropriately. 

Here is an example of what i mean: https://github.com/gioui/gio-x/tree/main/explorer

keeps everything reusable for different environments.

kozlovic (changes requested):
I am not knowledgeable on WASM to really comment, but there are conflicts that would need to be resolved before being able to merge.
For Go 1.17 support, please *add* (not replace) build directive as such:
```
//go:build <directive>
```
So suppose you want to have:
```
// +build wasm
```
it should become:
```
//go:build wasm
// +build wasm
```
etc.. Alternatively, if you are using Go 1.17, do a "go fmt" that should add those for you.

gedw99:
+1 Any movements on this ??

derekcollison:
Since the other PR landed with direct connect clients etc, @neilalexander do you think you could rebase on main and update copyright to 2022? Will see if we can merge for 2.9.

neilalexander:
@derekcollison Absolutely, I’ll do it first thing in the morning!

kozlovic (approved):
LGTM

unicomp23:
what's holding up the merge?

derekcollison:
Should be ok to merge, will wait on @kozlovic

PR #2364, opened 2021-07-19
[fixed] tls timeout now accepts units

matthiashanel:

ran into trying to set with units this the other day.
didn't work and didn't cause an error, so made it work


kozlovic (changes requested):
I understand that this is not related to your PR, but staticcheck has reported an error just above your change regarding to the user of strings.TrimSuffix instead of manual HasSuffix/trim.

kozlovic (approved):
LGTM

PR #2365, opened 2021-07-19
[changed] default account fetch timeout to be smaller as client timeout

matthiashanel:

changes this (account for creds has not been pushed):
```bash
nats --server localhost:4222 --creds ~/test/jetstream-leaf-nodes-demo/keys/creds/OP/NOPUSH/nopush.creds sub foo
nats: error: read tcp 127.0.0.1:49279->127.0.0.1:4222: i/o timeout, try --help
```

to that:
```bash
nats --server localhost:4222 --creds ~/test/jetstream-leaf-nodes-demo/keys/creds/OP/NOPUSH/nopush.creds sub foo
nats: error: nats: Authorization Violation, try --help
```

### DIFF
--- a/server/const.go
+++ b/server/const.go
@@ -202,5 +202,5 @@ const (
 	DEFAULT_GLOBAL_ACCOUNT = "$G"
 
 	// DEFAULT_FETCH_TIMEOUT is the default time that the system will wait for an account fetch to return.
-	DEFAULT_ACCOUNT_FETCH_TIMEOUT = 2 * time.Second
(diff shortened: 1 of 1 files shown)


kozlovic (approved):
LGTM. I assume we had not tests that expected/checked the default fetch timeout to be exactly 2secs.

matthiashanel:
Yeah, no tests for that

kozlovic:
To be clear, was not mentioning that you did not add a test for that, but instead that no existing test would start to fail due to this change?

matthiashanel:
I ran it locally without issues

PR #2369, opened 2021-07-22
[fixed] subscription on wildcard import that is not a subset

### LINKED ISSUE #2361: Subscriptions to streams which are imported with a wildcard subject and a prefix do not work correctly
## Defect
In Accounts configuration, if I import a stream from another account, with a subject containing a wildcard and any prefix, subscribers that subscribe to a subject respecting the prefix combined with the wildcard, do not receive any message.
In particular, given an import like the following `{stream: {account: ACCOUNT_X, subject: > }, prefix: myprefix}`, subscribers using a subject which is constructed with `*` instead of `myprefix` and any string instead of `>` (e.g. `*.test`) seem to not receive any message.

nats-server: v2.3.2

CentOS Linux release 7.9.2009 (Core)

1. Use this nats-server accounts configuration 
```
accounts: {
  ACCOUNT_X: {
    users: [
      {user: publisher}
    ]
    exports: [
      {stream: >}
    ]
  },
  ACCOUNT_Y: {
    users: [
      {user: subscriber}
    ]
    imports: [
      {stream: {account: ACCOUNT_X, subject: > }, prefix: myprefix}
    ]
  }
}
```
2. Configure a nats client with username `subscriber` to subscribe to `*.testsubject`
3. With a nats client with username `publisher`, publish a message on the subject `testsubject`

#### Expected result:
I expect that `subscriber` receives the event published by `publisher`

#### Actual result:
The subscriber does not receive any message

matthiashanel:
fixes #2361
The subject used was not a subset of the import. Nor the other way
around. Instead it is an overlap that needs to be dynamically computed.

I added more unit tests for transforms and discovered another test involving ">".
Took care to not iterate over subj more often than necessary. 
Part of this was to support better re-use of tokenized subjects.

### DIFF
--- a/server/client.go
+++ b/server/client.go
@@ -2509,9 +2509,10 @@ func (c *client) addShadowSubscriptions(acc *Account, sub *subscription) error {
 		for _, tk := range tokens {
 			if tk == pwcs {
 				hasWC = true
+				break
 			}
 		}
-		if tokens[len(tokens)-1] == fwcs {
+		if !hasWC && tokens[len(tokens)-1] == fwcs {
 			hasWC = true
 		}
 	}
@@ -2524,7 +2525,7 @@ func (c *client) addShadowSubscriptions(acc *Account, sub *subscription) error {
 			continue
 		}
 		if subj == im.to {
-			ims = append(ims, ime{im, "", false})
+			ims = append(ims, ime{im, _EMPTY_, false})
 			continue
 		}
 		if tokensModified {
@@ -2537,10 +2538,10 @@ func (c *client) addShadowSubscriptions(acc *Account, sub *subscription) error {
 		tokenizeSubjectIntoSlice(im.to, &imTokens)
 
 		if isSubsetMatchTokenized(tokens, imTokens) {
-			ims = append(ims, ime{im, "", true})
+			ims = append(ims, ime{im, _EMPTY_, true})
 		} else if hasWC {
 			if isSubsetMatchTokenized(imTokens, tokens) {
-				ims = append(ims, ime{im, "", false})
+				ims = append(ims, ime{im, _EMPTY_, false})
 			} else {
 				imTokensLen := len(imTokens)
 				for i, t := range tokens {
--- a/server/client.go
+++ b/server/client.go
@@ -2478,8 +2478,9 @@ func (c *client) processSubEx(subject, queue, bsid []byte, cb msgHandler, noForw
 
 // Used to pass stream import matches to addShadowSub
 type ime struct {
-	im  *streamImport
-	dyn bool
+	im          *streamImport
+	overlapSubj string
+	dyn         bool
 }
 
 // If the client's account has stream imports and there are matches for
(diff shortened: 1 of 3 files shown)


derekcollison (commented):
In general looks good, will let Ivan chime in and give official LGTM. Thanks.

server/client.go
```diff
@@ -2491,47 +2492,86 @@ func (c *client) addShadowSubscriptions(acc *Account, sub *subscription) error {
...
 	acc.mu.RLock()
-	// Loop over the import subjects. We have 3 scenarios. If we have an
+	subj := string(sub.subject)
+	if len(acc.imports.streams) > 0 {
+		tokens = tsa[:0]
+		tokenizeSubjectIntoSlice(subj, &tokens)
+		for _, tk := range tokens {
+			if tk == pwcs {
+				hasWC = true
+			}
```
kozlovic, line 2512: I think that a `break` makes sense here: as soon as you discover a `*` token, you know there is wildcard.

server/client.go
```diff
@@ -2491,47 +2492,86 @@ func (c *client) addShadowSubscriptions(acc *Account, sub *subscription) error {
...
+		}
+		if tokens[len(tokens)-1] == fwcs {
```
kozlovic, line 2514: Could even be: `if !hasWC && tokens[len(tokens)-1] == fwcs {`...

server/client.go
```diff
@@ -2491,47 +2492,86 @@ func (c *client) addShadowSubscriptions(acc *Account, sub *subscription) error {
...
+	// the import. If we are a subset or overlap, we have to dynamically calculate
+	// the subject. On overlap, ime requires the overlap subject.
 	for _, im := range acc.imports.streams {
 		if im.invalid {
 			continue
 		}
-		subj := string(sub.subject)
 		if subj == im.to {
-			ims = append(ims, ime{im, false})
+			ims = append(ims, ime{im, "", false})
```
kozlovic, line 2527: I know it is a bit silly, but for consistency sake, replace `""` with `_EMPTY_`.

server/client.go
```diff
@@ -2491,47 +2492,86 @@ func (c *client) addShadowSubscriptions(acc *Account, sub *subscription) error {
...
-					tokens = append(tokens, subj[start:i])
-					start = i + 1
+			tokenizeSubjectIntoSlice(subj, &tokens)
+			tokensModified = false
+		}
+		imTokens := imTsa[:0]
+		tokenizeSubjectIntoSlice(im.to, &imTokens)
+
+		if isSubsetMatchTokenized(tokens, imTokens) {
+			ims = append(ims, ime{im, "", true})
```
kozlovic, line 2540: `_EMPTY_`

server/client.go
```diff
@@ -2491,47 +2492,86 @@ func (c *client) addShadowSubscriptions(acc *Account, sub *subscription) error {
...
+		} else if hasWC {
+			if isSubsetMatchTokenized(imTokens, tokens) {
+				ims = append(ims, ime{im, "", false})
```
kozlovic, line 2543: `_EMPTY_`

kozlovic (approved):
Maybe can break when first wildcard is discovered, and change `""` to `_EMPTY_`, otherwise LGTM.

server/client.go
matthiashanel: agreed

server/client.go
matthiashanel: Yes

server/client.go
matthiashanel: done

server/client.go
matthiashanel: done

server/client.go
matthiashanel: done

kozlovic (approved):
LGTM

kozlovic (approved):
LGTM

derekcollison (approved):
LGTM

PR #2371, opened 2021-07-23
wip consumer descriptions

ripienaar:
@derekcollison this is the WIP work I had for descriptions, hacky and incomplete no doubt but was enough for me to identify which source or mirror was constantly creating things

### DIFF
--- a/server/consumer.go
+++ b/server/consumer.go
@@ -48,6 +48,7 @@ type ConsumerInfo struct {
 
 type ConsumerConfig struct {
 	Durable         string        `json:"durable_name,omitempty"`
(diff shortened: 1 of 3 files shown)


derekcollison (commented):
This in general looks fine to me as is, not sure I have to do a redo. 

Question though, why did you stamp inside of source info? Why not just look it up and have StreamInfo return it as part of config etc?

ripienaar:
Dont recall to be honest. Just hacked it up to debug a case where a source kept being recreated, rather than do what’s best :)

derekcollison:
I think this can be closed with the meta fields for streams and consumers getting close to landing.

PR #2373, opened 2021-07-25
Remove the swapping of accounts during processing of service imports.

derekcollison:
When processing service imports we would swap out the accounts during processing.
With the addition of internal subscriptions and internal clients publishing in JetStream we had an issue with the wrong account being used.

This was specific to delayed pull subscribers trying to unsubscribe due to max of 1 while other JetStream API calls were running concurrently.

/cc @nats-io/core

### DIFF
--- a/server/norace_test.go
+++ b/server/norace_test.go
@@ -1102,6 +1102,89 @@ func TestNoRaceJetStreamDeleteStreamManyConsumers(t *testing.T) {
 	mset.delete()
 }
 
+// We used to swap accounts on an inbound message when processing service imports.
+// Until JetStream this was kinda ok, but with JetStream we can have pull consumers
+// trying to access the clients account in another Go routine now which causes issues.
+// This is not limited to the case above, its just the one that exposed it.
+// This test is to show that issue and that the fix works, meaning we no longer swap c.acc.
+func TestNoRaceJetStreamServiceImportAccountSwapIssue(t *testing.T) {
+	s := RunBasicJetStreamServer()
+	defer s.Shutdown()
+
+	if config := s.JetStreamConfig(); config != nil {
+		defer removeDir(t, config.StoreDir)
+	}
+
+	// Client based API
+	nc, js := jsClientConnect(t, s)
+	defer nc.Close()
+
+	_, err := js.AddStream(&nats.StreamConfig{
+		Name:     "TEST",
+		Subjects: []string{"foo", "bar"},
+	})
+	if err != nil {
+		t.Fatalf("Unexpected error: %v", err)
+	}
+
+	sub, err := js.PullSubscribe("foo", "dlc")
+	if err != nil {
+		t.Fatalf("Unexpected error: %v", err)
+	}
+
+	beforeSubs := s.NumSubscriptions()
+
+	// How long we want both sides to run.
+	timeout := time.Now().Add(3 * time.Second)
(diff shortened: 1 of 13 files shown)


server/norace_test.go
```diff
@@ -1102,6 +1102,89 @@ func TestNoRaceJetStreamDeleteStreamManyConsumers(t *testing.T) {
...
+	}
+
+	beforeSubs := s.NumSubscriptions()
+
+	// How long we want both sides to run.
+	timeout := time.Now().Add(3 * time.Second)
+	errs := make(chan error, 1)
+
+	// Publishing side, which will signal the consumer that is waiting and which will access c.acc. If publish
+	// operation runs concurrently we will catch c.acc being $SYS sometimes from the
```
wallyqs, line 1142:
```suggestion
	// operation runs concurrently we will catch c.acc being $SYS sometimes.
```

wallyqs (approved):
LGTM! Did some local testing too and confirmed issue is resolved.

server/norace_test.go
derekcollison: Will fix thx!

server/norace_test.go
```diff
@@ -1102,6 +1102,89 @@ func TestNoRaceJetStreamDeleteStreamManyConsumers(t *testing.T) {
...
+
+	// Publishing side, which will signal the consumer that is waiting and which will access c.acc. If publish
+	// operation runs concurrently we will catch c.acc being $SYS some of the time.
+	go func() {
+		time.Sleep(100 * time.Millisecond)
+		for time.Now().Before(timeout) {
+			// This will signal the delivery of the pull messages.
+			js.Publish("foo", []byte("Hello"))
+			// This will swap the account because of JetStream service import.
+			// We can get an error here with the bug or not.
```
kozlovic, line 1149: Are you saying that without the bug we could get legitimate error here, but does that mean that the test will flap, or just that with the bug you would get an error here, but without the bug, if there is an error something else is really wrong and we want the test to fail?

server/norace_test.go
```diff
@@ -1102,6 +1102,89 @@ func TestNoRaceJetStreamDeleteStreamManyConsumers(t *testing.T) {
...
+	// Double check all received.
+	si, err := js.StreamInfo("TEST")
+	if err != nil {
+		t.Fatalf("Unexpected error: %v", err)
+	}
+	if int(si.State.Msgs) != received {
+		t.Fatalf("Expected to receive %d msgs, only got %d", si.State.Msgs, received)
+	}
+	// Now check for leaked subs from the fetch call above. That is what we first saw from the bug.
+	if afterSubs := s.NumSubscriptions(); afterSubs != beforeSubs {
```
kozlovic, line 1183: Is that going to flap? (can there be some subscriptions that are setup once after some first event and those subs stay for a while).
I have run this new test and it passes, I am just wondering if there is a risk of flapping because of this strict comparison.

kozlovic (approved):
LGTM, just some comments about the test.

server/norace_test.go
derekcollison: No with bug present could fail here sometimes as well as below. Neither should fail with the fix.

server/norace_test.go
derekcollison: I ran it for 100 times straight over the weekend so I think its stable.

PR #2374, opened 2021-07-27
fix error replacements in consumer validation

ripienaar:

/cc @nats-io/core

### DIFF
--- a/server/consumer.go
+++ b/server/consumer.go
@@ -374,21 +374,21 @@ func (mset *stream) addConsumerWithAssignment(config *ConsumerConfig, oname stri
 			return nil, ApiErrors[JSConsumerInvalidPolicyErrF].NewT("{err}", "consumer delivery policy is deliver new, but optional start sequence is also set")
 		}
 		if config.OptStartTime != nil {
-			return nil, ApiErrors[JSConsumerInvalidPolicyErrF].NewT("consumer delivery policy is deliver new, but optional start time is also set")
+			return nil, ApiErrors[JSConsumerInvalidPolicyErrF].NewT("{err}", "consumer delivery policy is deliver new, but optional start time is also set")
 		}
 	case DeliverByStartSequence:
 		if config.OptStartSeq == 0 {
-			return nil, ApiErrors[JSConsumerInvalidPolicyErrF].NewT("consumer delivery policy is deliver by start sequence, but optional start sequence is not set")
+			return nil, ApiErrors[JSConsumerInvalidPolicyErrF].NewT("{err}", "consumer delivery policy is deliver by start sequence, but optional start sequence is not set")
 		}
 		if config.OptStartTime != nil {
-			return nil, ApiErrors[JSConsumerInvalidPolicyErrF].NewT("consumer delivery policy is deliver by start sequence, but optional start time is also set")
+			return nil, ApiErrors[JSConsumerInvalidPolicyErrF].NewT("{err}", "consumer delivery policy is deliver by start sequence, but optional start time is also set")
 		}
 	case DeliverByStartTime:
 		if config.OptStartTime == nil {
-			return nil, ApiErrors[JSConsumerInvalidPolicyErrF].NewT("consumer delivery policy is deliver by start time, but optional start time is not set")
+			return nil, ApiErrors[JSConsumerInvalidPolicyErrF].NewT("{err}", "consumer delivery policy is deliver by start time, but optional start time is not set")
 		}
 		if config.OptStartSeq != 0 {
-			return nil, ApiErrors[JSConsumerInvalidPolicyErrF].NewT("consumer delivery policy is deliver by start time, but optional start sequence is also set")
+			return nil, ApiErrors[JSConsumerInvalidPolicyErrF].NewT("{err}", "consumer delivery policy is deliver by start time, but optional start sequence is also set")
 		}
 	}
 


wallyqs (approved):
LGTM

ripienaar:
When I am back i will look to remove the panic - go stdlib also have a pnic in this situation where i copied the idea from, but I think we can remove it if we make some generic error

I left the panic on the assumption that test coverage would find the problem - valid enough assumption for new code but when retrofitting maybe not so valid.

Should also see if there's a linter we can extend perhaps

derekcollison:
Let's add a test though..

ripienaar:
> Let's add a test though..

Rather than a test, due to this kind of problem I dont think testing them is particularly useful, can I add a linter that identifies this style of problem?

If ran against the code pre this commit I got:

```
$ ./lint github.com/nats-io/nats-server/v2/server
/Users/rip/go/src/github.com/nats-io/nats-server/server/consumer.go:377:16: invalid number of arguments given
/Users/rip/go/src/github.com/nats-io/nats-server/server/consumer.go:381:16: invalid number of arguments given
/Users/rip/go/src/github.com/nats-io/nats-server/server/consumer.go:384:16: invalid number of arguments given
/Users/rip/go/src/github.com/nats-io/nats-server/server/consumer.go:388:16: invalid number of arguments given
/Users/rip/go/src/github.com/nats-io/nats-server/server/consumer.go:391:16: invalid number of arguments given
```

derekcollison:
If we need to run linters to avoid panics etc maybe we need to revisit how we want this implemented to make it a bit more stable.

ripienaar:
Even the go stdlib behaves this way. Happy to take suggestions but I think linkers are pretty good. It’s not just to avoid the panic it also checks for a general well formedness of these strings and we could also catch some other best practises compilers would never catch as it relates to them. 

So even if we could turn the core problem into a compile error linters would add value

ripienaar:
As an example. Essentially this replaces “{err}” and other tags with some text. 

No way can the compiler or most humans not mess this up and even down the line as we add new tags to errors we would make mistakes in consolidating that, or just typos etc. 

This is not something we can resolve at compile time no matter how we change things and we will always be subject to human error - I can do it in a linter though and warn immediately.

derekcollison:
You are probably right by why can't it be done at compile time? Seems like it should be possible.

ripienaar:
func x(args …string) {}

How does the compiler enforce args is an even number of arguments. A common way is to do like Pair{“{err}”, err} and then accept …Pair but this is very unwieldy and annoying in use with many errors like we have. So I made it like strings.Replacer() that has the same issue. 

And then still doesn’t answer the question if “{err}” is valid nor does it enforce that all tokens are passed in. 

Linter for 100% all above in all cases would also be hard. But we have few patterns of use so should be doable.

derekcollison:
That can easily be checked at runtime, I just do not want the server to crash of course but also want to keep the number of steps needed to compile and test the server to a known set.

ripienaar:
Yeah - we do check this at runtime and I will do something to remove the panic for sure (as per previous comment)

however for us only time we will catch this is if in test - automated or manual- we exercise every path leading to an error - a very tall order. 

I would lint in CI, you would do what you do today. Generally it would just be fine since these errors while easy to make seems not too common based on looking how we have been using it last while

ripienaar:
There is one option we can adopt and that is when generating the ApiErrors struct we can also generate functions like:

func JSConsumerInvalidPolicyErrF(err string) error { … }

This would define all the tokens that can go into the string and we might add some additional variant for the NewOrT() usage. But the point is then all supported tokens are turned into arguments and the compiler will catch (most) issues for us. 

I can show proper examples of this if yous think it makes sense - I think it would be easier to use than what we have atm and less error prone

derekcollison:
I am open to any suggestions that make it less error prone.

ripienaar:
Will tinker in the week and show a small POC

ripienaar:
So basically I'd generate a bunch of functions like this, for each error with no replacements:

```go
// NewJSAccountResourcesExceededErr resource limits exceeded for account
func NewJSAccountResourcesExceededErr(opts ...Option) *ApiError {
	eopts := parseOpts(opts)

	if ae, ok := eopts.err.(*ApiError); ok {
		return ae
	}

	return ApiErrors[JSAccountResourcesExceededErr]
}
```

And for each one with replacements, specifically calling out each replacement, arguments in alpha order:

```go
// NewJSRestoreSubscribeFailedErrF JetStream unable to subscribe to restore snapshot {subject}: {err}
func NewJSRestoreSubscribeFailedErrF(err error, subject string, opts ...Option) *ApiError {
	eopts := parseOpts(opts)
	if ae, ok := eopts.err.(*ApiError); ok {
		return ae
	}

	return ApiErrors[JSRestoreSubscribeFailedErrF].NewT([]interface{}{"{err}", err, "{subject}", subject}...)
}
```

This means:

```go
err := ApiErrors[JSAccountResourcesExceededErr]
err = ApiErrors[JSAccountResourcesExceededErr].ErrOr(err)
err = ApiErrors[JSRestoreSubscribeFailedErrF].NewT("{err}", err, "{subject}", "foo")
err = ApiErrors[JSRestoreSubscribeFailedErrF].ErrOrNewT(err, "{err}", err, "{subject}", "foo")

```

becomes:

```go
err := NewJSAccountResourcesExceededErr() // just returns the static error
err = NewJSAccountResourcesExceededErr(UnlessApiErr(err)) // returns err if its a ApiError else the static error
err = NewJSRestoreSubscribeFailedErrF(err, "foo") // new instance of error with substitutions
err = NewJSRestoreSubscribeFailedErrF(err, "foo", UnlessApiErr(err)) // returns err if its a ApiErr else new instance of error with substitutions
```

derekcollison:
Interested in understanding why this has to be so complex..

ripienaar:
> Interested in understanding why this has to be so complex..

maybe best to zoom.

But we need:

 * basic token replacement
 * ability to have a generic or specific error 
 * easy way to generate all these errors

thats all these functions allow and the proposed approach is pretty much spot on and removes most complexity/annoyance

PR #2375, opened 2021-07-27
[fixed] updated jwt library to avoid an error when counting tokens

matthiashanel:


kozlovic (approved):
LGTM

PR #2377, opened 2021-07-27
Add descriptions to JetStream streams and consumers.

derekcollison:

/cc @nats-io/core

### DIFF
--- a/server/consumer.go
+++ b/server/consumer.go
@@ -48,6 +48,7 @@ type ConsumerInfo struct {
 
 type ConsumerConfig struct {
 	Durable         string        `json:"durable_name,omitempty"`
+	Description     string        `json:"description,omitempty"`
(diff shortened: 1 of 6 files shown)


kozlovic (approved):
LGTM

ripienaar:
LGTM but should we at the same time add them to the consumers for sources and mirrors?

derekcollison:
You mean auto-add them to any direct consumer we place on a mirror or consumer? What name should we use?

ripienaar:
Yes, when I see a mirror or source not working I also tend to see 1000s of failed API requests. Then I use nats event or just a sub to see what is going on. So in my original WIP PR I added descriptions to the consumer create requests indicating what stream is making those requests

derekcollison:
Let's sync up next week and we can do a separate PR. Want to understand a bit better.

PR #2378, opened 2021-07-27
[fixed] jetstream unique server name requirement across domains

matthiashanel:
including domain in server info
adding check for cluster name in duplicate leaf node connection check

This does not address non unique domains in the same domain, say within super cluster.

---

By using the hash of the server name to compute a shortened identifier, jetstream introduced the need for unique server names (say across an entire super cluster)
```
node := string(getHash(si.Name))
s.nodeToInfo.Store(node, nodeInfo{si.Name, si.Cluster, si.ID, false, si.JetStream})
```

Short of changing how this key is computed, say by using a truncated server Id or the hash of domain+cluster+server. server names have to be unique. This change only addresses the need for unique server names across domains.

When this change is accepted, I'll add domain to jsm server info as well.

### DIFF
--- a/server/events.go
+++ b/server/events.go
@@ -157,6 +157,7 @@ type ServerInfo struct {
 	Host      string    `json:"host"`
 	ID        string    `json:"id"`
 	Cluster   string    `json:"cluster,omitempty"`
+	Domain    string    `json:"domain,omitempty"`
 	Version   string    `json:"ver"`
 	Seq       uint64    `json:"seq"`
 	JetStream bool      `json:"jetstream"`
@@ -270,6 +271,7 @@ RESET:
 	id := s.info.ID
 	host := s.info.Host
 	servername := s.info.Name
+	domain := s.info.Domain
 	seqp := &s.sys.seq
 	js := s.info.JetStream
 	cluster := s.info.Cluster
@@ -294,6 +296,7 @@ RESET:
 		case pm := <-sendq:
 			if pm.si != nil {
 				pm.si.Name = servername
+				pm.si.Domain = domain
 				pm.si.Host = host
 				pm.si.Cluster = cluster
 				pm.si.ID = id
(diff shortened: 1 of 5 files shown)


derekcollison:
Just to note the requirement for unique server names has always been in JetStream. But want to make sure we do not require that across distinct JS domains. Like NGS.

server/events.go
```diff
@@ -946,8 +948,15 @@ func (s *Server) remoteServerShutdown(sub *subscription, _ *client, _ *Account,
 		return
 	}
 	// Additional processing here.
-	node := string(getHash(si.Name))
-	s.nodeToInfo.Store(node, nodeInfo{si.Name, si.Cluster, si.ID, true, true})
+	if s.sameDomain(si.Domain) {
```
derekcollison, line 951: if !same just return?

derekcollison (approved):
LGTM

server/events.go
matthiashanel: fixed

matthiashanel:
> Just to note the requirement for unique server names has always been in JetStream. But want to make sure we do not require that across distinct JS domains. Like NGS.

Will make a note of that in the docs

server/leafnode.go
```diff
@@ -1198,7 +1198,7 @@ func (s *Server) setLeafNodeInfoHostPortAndIP() error {
 // (this solves the stale connection situation). An error is returned to help the
 // remote detect the misconfiguration when the duplicate is the result of that
 // misconfiguration.
-func (s *Server) addLeafNodeConnection(c *client, srvName string, checkForDup bool) {
+func (s *Server) addLeafNodeConnection(c *client, srvName string, clusterName string, checkForDup bool) {
```
kozlovic, line 1201: or `.., srvName, clusterName string, ..`

server/events.go
```diff
@@ -958,8 +968,10 @@ func (s *Server) remoteServerUpdate(sub *subscription, _ *client, _ *Account, su
 		return
 	}
 	si := ssm.Server
-	node := string(getHash(si.Name))
-	s.nodeToInfo.Store(node, nodeInfo{si.Name, si.Cluster, si.ID, false, si.JetStream})
+	if s.sameDomain(si.Domain) {
```
kozlovic, line 971: Since there is nothing else after this "if" statement, then could do the same than Derek mentione:
```
if !s.sameDomain(si.Domain) {
    return
}
node := ...
```

kozlovic (approved):
LGTM, only some little remarks.
I assume, though, that this is "backward compatible", meaning that 2 servers at different version will work ok, in that the "old" server won't have the Domain sent as part of the INFO protocol.

matthiashanel:
@kozlovic 
> I assume, though, that this is "backward compatible", meaning that 2 servers at different version will work ok, in that the "old" server won't have the Domain sent as part of the INFO protocol.

No. That information is required. But it will  be so broken as to be unusable. (stream's can't be created)

PR #2381, opened 2021-07-28
Allow delivery policy of DeliverLastPerSubject, which is helpful for K/V on JetStream

derekcollison:

/cc @nats-io/core

### DIFF
--- a/server/consumer.go
+++ b/server/consumer.go
@@ -86,8 +86,10 @@ const (
 	DeliverNew
 	// DeliverByStartSequence will look for a defined starting sequence to start.
 	DeliverByStartSequence
-	// DeliverByStartTime will select the first messsage with a timestamp >= to StartTime
+	// DeliverByStartTime will select the first messsage with a timestamp >= to StartTime.
 	DeliverByStartTime
+	// DeliverLastPerSubject will start the consumer with the last message for all subjects received.
+	DeliverLastPerSubject
 )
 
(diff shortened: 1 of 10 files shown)


server/consumer.go
```diff
@@ -2953,18 +3031,28 @@ func (o *consumer) setInitialPendingAndStart() {
 		}
 	} else {
 		// Here we are filtered.
-		ss := mset.store.FilteredState(o.sseq, o.cfg.FilterSubject)
-		if ss.Msgs > 0 {
+		dp := o.cfg.DeliverPolicy
+		if dp == DeliverLastPerSubject && o.hasSkipListPending() && o.sseq < o.lss.resume {
+			if o.lss != nil {
```
ripienaar, line 3036: i tink `o.hasSkipListPending()` check above implies o.lss will be non nil

derekcollison: Upper clause has an additional conditional so I need to check again.

```go
if dp == DeliverLastPerSubject && o.hasSkipListPending() && o.sseq < o.lss.resume {
```

ripienaar (approved):
LGTM

aricart:
@derekcollison - the next ask is probably to do last_per_subject starting at a known sequence. This would allow a client to store a 'stop' location in the stream, but only have to review the latest changes to some of the data.

derekcollison:
Good idea to discuss between you, R.I. and others.

PR #2382, opened 2021-07-28
[FIXED] A stream with interest retention had a clustered consumer could cause server panic.

derekcollison:
A clustered consumer on an interest retention policy it could cause the server to panic when the consumer was being deleted.

/cc @nats-io/core

### DIFF
--- a/server/consumer.go
+++ b/server/consumer.go
@@ -2802,7 +2802,6 @@ func (o *consumer) stopWithFlags(dflag, sdflag, doSignal, advisory bool) error {
 	// non-leader consumers will need to restore state first.
 	if dflag && rp == InterestPolicy {
 		stop := mset.lastSeq()
-
 		o.mu.Lock()
 		if !o.isLeader() {
 			o.readStoredState()
@@ -2810,8 +2809,7 @@ func (o *consumer) stopWithFlags(dflag, sdflag, doSignal, advisory bool) error {
 		start := o.asflr
 		o.mu.Unlock()
 
-		rmseqs := make([]uint64, 0, stop-start+1)
-
+		var rmseqs []uint64
 		mset.mu.RLock()
 		for seq := start; seq <= stop; seq++ {
 			if !mset.checkInterest(seq, o) {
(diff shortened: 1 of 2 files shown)


kozlovic (approved):
LGTM

matthiashanel (commented):
LGTM - I assume stop was off?

derekcollison:
Was negative regardless but I think it was being read in from the consumer side replicas that could have caused it.

sloveridge:
This is a straight forward fix for the panic but in the situation where start > stop+2 (the trigger for the cap out of bounds panic) messages will not be cleaned within stopWithFlags as they should be when using interest policy

ie; the loop to append rmseqs will not be entered
```
start := o.asflr
o.mu.Unlock()

var rmseqs []uint64
mset.mu.RLock()
for seq := start; seq <= stop; seq++ {
	if !mset.checkInterest(seq, o) {
		rmseqs = append(rmseqs, seq)
	}
}
mset.mu.RUnlock()

for _, seq := range rmseqs {
	mset.store.RemoveMsg(seq)
}
```


This can be verified by altering the test case to:
```
func TestJetStreamConsumerCleanupWithRetentionPolicy(t *testing.T) {

...

        for i := 0; i < 5; i++ {
                m, err := sub.NextMsg(time.Second)
                if err != nil {
                        t.Fatalf("Unexpected error: %v", err)
                }
                m.Ack()
        }

...

        if si.State.Msgs != 0 {
                t.Fatalf("%d messages not cleaned up after unsub", si.State.Msgs)
        }
}
```

derekcollison:
In the test we artificially make the number larger to trigger the panic. The floor will represent usually the messages already acked and hence would have been already cleaned up when in interest policy mode.

Maybe I am not understanding correctly but I think this is correct.

sloveridge:
I guess it depends on how stream ack floor ended up > stream last msg. You are likely correct though that it wouldn't matter as long as the increment still caused the message removal.

My concern was that as the cause of floor being higher wasn't directly addressed that the issue might be somewhat hidden with a symptom of a build up of old messages in the stream over time.

derekcollison:
The ack floor will never be artificially high (unless there is a bug). But you could imagine a consumer in a clustered setup being asked to auto-delete due to lack of interest. Note that the consumer leader != stream leader, so it could be that the stream state is behind where the consumer state is on that server if say the stream is very busy etc. That was my theory on how it could trigger.

PR #2383, opened 2021-07-28
Remove '-s' flag from ldflags from the build

### LINKED ISSUE #2366: Include a copy of nats server with symbols in each release
## Feature Request
In each release, include an additional nats-server executable that includes symbols.

#### Use Case:
During debugging sessions, it would be useful to have symbols for breaking point.

#### Proposed Change:
In addition to the stripped executable, include another executable with symbols.

Operators of nats server.

#### Alternative Approaches
Someone build their own nats-server with symbols.

yzhao1012:
This retains the symbols (but still removes the dwarf info).

This increases the executable size by ~8% (test on my laptop):

$ go build -ldflags '-w' -o nats-server
-rwxrwxr-x 1 yzhao yzhao 11911273 Jul 28 15:52 nats-server

$ go build -ldflags '-s -w' -o nats-server
-rwxrwxr-x 1 yzhao yzhao 11128832 Jul 28 15:52 nats-server

Symbols are very useful during debugging. It would be great that if the
NATS team consider including symbols.

Resolves #2366

/cc @nats-io/core

### DIFF
--- a/.goreleaser-nightly.yml
+++ b/.goreleaser-nightly.yml
@@ -5,7 +5,7 @@ builds:
     id: nats-server
     binary: nats-server
     ldflags:
-      - -s -w -X github.com/nats-io/nats-server/server.gitCommit={{.ShortCommit}}
+      - -w -X github.com/nats-io/nats-server/server.gitCommit={{.ShortCommit}}
     env:
       - GO111MODULE=off
       - CGO_ENABLED=0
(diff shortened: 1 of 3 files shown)


kozlovic:
@derekcollison I don't see a problem including symbols, so would be ok to accept this PR. If any objection, then undo the changes for actual release (.goreleaser.yml) but keep changes for the nightly?

derekcollison:
I am fine with it as well, just wanted your opinion. When we were ~5M image size it mattered but at this point it might be better to have them in there.

kozlovic (approved):
LGTM, thank you for your contribution!

PR #2387, opened 2021-07-30
[IMPROVED] Server restart time with many expired JetStream messages.

derekcollison:
When a server was restarting, if a stream had a MaxAge configured and there were a very large amount of messages to expire, this process would take too long and delay the server reaching a running state.

This PR introduces an optimized expiration tailored for startup.

/cc @nats-io/core

### DIFF
--- a/server/filestore.go
+++ b/server/filestore.go
@@ -863,12 +863,120 @@ func (fs *fileStore) recoverMsgs() error {
 	fs.enforceBytesLimit()
 
 	// Do age checks too, make sure to call in place.
-	if fs.cfg.MaxAge != 0 && fs.state.Msgs > 0 {
+	if fs.cfg.MaxAge != 0 {
+		fs.expireMsgsOnRecover()
 		fs.startAgeChk()
-		fs.expireMsgsLocked()
 	}
 
 	return nil
+}
+
+// Will expire msgs that have aged out on restart.
+// We will treat this differently in case we have a recovery
+// that will expire alot of messages on startup. Should only be called
+// on startup. Lock should be held.
+func (fs *fileStore) expireMsgsOnRecover() {
+	if fs.state.Msgs == 0 {
+		return
+	}
+
+	var minAge = time.Now().UnixNano() - int64(fs.cfg.MaxAge)
+	var purged, bytes uint64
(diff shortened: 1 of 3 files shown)


server/filestore.go
```diff
@@ -863,14 +863,121 @@ func (fs *fileStore) recoverMsgs() error {
...
+			mb.dirtyCloseWithRemove(true)
+			mb.mu.Unlock()
+			// Update fs first here as well.
+			fs.state.FirstSeq = mb.last.seq + 1
+			fs.state.FirstTime = time.Time{}
+			deleted++
+			continue
+		}
+
+		// If we are here we have to process the interior messages of this blk.
```
matthiashanel, line 906: For simplicity, would it have sufficed to expire entire blocks and then call fs.expireMsgsLocked() instead of doing the rest here?
Also what call causes the consumer to be updated?

server/filestore.go
```diff
@@ -863,14 +863,121 @@ func (fs *fileStore) recoverMsgs() error {
...
+			break
+		}
+		// Can we remove whole block here?
+		if mb.last.ts <= minAge {
+			purged += mb.msgs
+			bytes += mb.bytes
+			mb.dirtyCloseWithRemove(true)
+			mb.mu.Unlock()
+			// Update fs first here as well.
+			fs.state.FirstSeq = mb.last.seq + 1
```
kozlovic, line 900: Would move this (and the why not the next line) before releasing the lock.

kozlovic (approved):
LGTM except for the `mb.last.seq+1` outside of the mb's lock.

server/filestore.go
derekcollison: Possibly but this seemed more appropriate since we can do different things since we know we are starting up etc.

No consumer updates here, on startup its not registered so I was incorrect on slowdown, was just expireMsgs() with alot of messages.

server/filestore.go
derekcollison: Will fix.

server/filestore.go
matthiashanel: ok

matthiashanel (commented):
LGTM

PR #2390, opened 2021-07-30
Off by one bug (surprise) that would skip first new message on deliver last per subject.

derekcollison:

/cc @nats-io/core


kozlovic (commented):
LGTM

PR #2393, opened 2021-08-01
[FIX] #2388

### LINKED ISSUE #2388: JetStream API deny permissions along leaf node connections are inserted to easily.
A leafnode will always insert deny clauses in `leafnode.go` function `processLeafnodeInfo`.
This is irrespective of the domain name used.

```txt
[42059] [INF] Adding deny export of "$JS.API.>" for leafnode configuration on "JSY" that bridges system account
[42059] [INF] Adding deny import of "$JS.API.>" for leafnode configuration on "JSY" that bridges system account
```

Specifically this should only be added when the domain names on either end of the leaf node connection are not empty and do not differ. Thus line `1012` should add a condition for `!s.sameDomain(info.Domain)` to only execute the code below then.

In the specific use case, jetstream is not even enabled on the leaf node.
Thus, we could just check for `opts.JetStream` instead. Meaning don't add the denies when JS is not enabled.

But consider such a daisy chained setup:
s1 (JetStream) <-ln- s2 (no JetStream) <-ln- s3 (JetStream)

Both approaches will break down.

The check for `!s.sameDomain` suffers from the domain not being settable without jetstream.

I'd suggest we make domain settable without jetsttream and only insert the denies when `!s.sameDomain(info.Domain)`

derekcollison:
Allow non-JS leafnode(s) to access a HUB transparently. Also tests for mixed mode.

Mixed mode should always configure a JS block for the server and specify domain and disable the servers they do not want to do JS.
e.g. jetstream: {domain: "SPOKE", enabled: false}

Resolves #2388

 -
 -
 -

/cc @nats-io/core

### DIFF
--- a/server/events.go
+++ b/server/events.go
@@ -632,6 +632,12 @@ func (s *Server) heartbeatStatsz() {
 	s.sendStatsz(fmt.Sprintf(serverStatsSubj, s.info.ID))
 }
 
+func (s *Server) sendStatszUpdate() {
+	s.mu.Lock()
+	s.sendStatsz(fmt.Sprintf(serverStatsSubj, s.info.ID))
+	s.mu.Unlock()
+}
+
 // This should be wrapChk() to setup common locking.
 func (s *Server) startStatszTimer() {
 	// We will start by sending out more of these and trail off to the statsz being the max.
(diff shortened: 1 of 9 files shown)


matthiashanel (approved):
Lgtm

PR #2394, opened 2021-08-02
Fix for #2392. We were not resetting the pending timer properly during a stream purge.

### LINKED ISSUE #2392: Not Acking a message before Purging stream prevents redeliver of subsequent Non acked message for consumer
## Defect
Following issue was reported on slack:

> whenever I purge the stream and publish a message on the stream then a consumer gets a message only once even if I did not send ack signal. Before I purge, stream and consumer act as I expected. But whenever I do as following, a consumer cannot get the same message more than once. 
pub message --> sub (no ack) --> purge stream --> pub message --> sub(no ack) --> wait for ack wait time --> sub(no ack) << cannot get message
When I check the stream there is still 1 message but consumer just can’t get that message

Go code to reproduce this behaviour below.

 - [x ] Included `nats-server -DV` output
 - [x ] Included a [Minimal, Complete, and Verifiable example] (https://stackoverflow.com/help/mcve)

Version:  2.3.3-beta
Git:      [19bfef59]
Go build: go1.16.6

Tested with github.com/nats-io/nats.go v1.11.1-0.20210714175522-2b2bb8f326df

Using synadia/nats-server:nightly

```
package main

import (
	"fmt"
	"github.com/nats-io/nats.go"
	"time"
)

func main() {
	nc, _ := nats.Connect(nats.DefaultURL)
	defer nc.Drain()
	js, _ := nc.JetStream()
	fmt.Println("\n-- Ack Before Purge --")
	test(js, true)
	fmt.Println("\n-- No Ack Before Purge --")
	test(js, false)
}

func test(js nats.JetStreamContext, ack bool) {
	js.AddStream(&nats.StreamConfig{
		Name:     "foo",
		Subjects: []string{"foo.*"},
	})
	defer js.DeleteStream("foo")
	js.Publish("foo.a", []byte("show once"))
	sub, _ := js.SubscribeSync("foo.*", nats.AckWait(2*time.Second), nats.DeliverAll(), nats.AckExplicit())
	defer sub.Unsubscribe()
	next(sub, ack) // No ack / nack here is the issue
	js.PurgeStream("foo")
	fmt.Println("*Stream Purged*")
	js.Publish("foo.a", []byte("show twice"))
	next(sub, false)
	next(sub, false)
}


func next(sub *nats.Subscription, ack bool) {
	m, err := sub.NextMsg(5 * time.Second)
	ci, _ := sub.ConsumerInfo()
	if err == nats.ErrTimeout {
		fmt.Printf("should not see (Ack Pending: %d)\n", ci.NumAckPending)
	} else {
		if ack {
			m.Ack()
		}
		fmt.Printf("%s (Ack Pending: %d)\n", m.Data, ci.NumAckPending)
	}
}
```

#### Expected result:
Message published after purge should be redelivered after ack wait when not acknowledged

-- Ack Before Purge --
show once (Ack Pending: 1)
*Stream Purged*
show twice (Ack Pending: 1)
show twice (Ack Pending: 1)

-- No Ack Before Purge --
show once (Ack Pending: 1)
*Stream Purged*
show twice (Ack Pending: 1)
show twice (Ack Pending: 1)

#### Actual result:
Message published after purge **NOT** being redelivered after ack wait when not acknowledged

-- Ack Before Purge --
show once (Ack Pending: 1)
*Stream Purged*
show twice (Ack Pending: 1)
show twice (Ack Pending: 1)

-- No Ack Before Purge --
show once (Ack Pending: 1)
*Stream Purged*
show twice (Ack Pending: 1)
should not see (Ack Pending: 1)

derekcollison:
Now we just let it fire as normal and it will be canceled or not as needed.

Resolves #2392 

/cc @nats-io/core

### DIFF
--- a/server/consumer.go
+++ b/server/consumer.go
@@ -2757,20 +2757,19 @@ func (o *consumer) hasNoLocalInterest() bool {
 }
 
 // This is when the underlying stream has been purged.
+// sseq is the new first seq for the stream after purge.
 func (o *consumer) purge(sseq uint64) {
+	if sseq == 0 {
+		return
+	}
+
 	o.mu.Lock()
 	o.sseq = sseq
 	o.asflr = sseq - 1
 	o.adflr = o.dseq - 1
 	o.sgap = 0
-	if len(o.pending) > 0 {
-		o.pending = nil
-		if o.ptmr != nil {
-			o.ptmr.Stop()
-			// Do not nil this out here. This allows checkPending to fire
-			// and still be ok and not panic.
-		}
-	}
+	o.pending = nil
+
 	// We need to remove all those being queued for redelivery under o.rdq
 	if len(o.rdq) > 0 {
 		rdq := o.rdq
--- a/server/jetstream_test.go
+++ b/server/jetstream_test.go
@@ -12382,6 +12382,58 @@ func TestJetStreamConsumerCleanupWithRetentionPolicy(t *testing.T) {
 	}
 }
 
+// Issue #2392
+func TestJetStreamPurgeEffectsConsumerDelivery(t *testing.T) {
+	s := RunBasicJetStreamServer()
+	defer s.Shutdown()
+
+	if config := s.JetStreamConfig(); config != nil {
+		defer removeDir(t, config.StoreDir)
+	}
+
+	// Client for API requests.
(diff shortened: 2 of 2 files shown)


matthiashanel (approved):
Lgtm

PR #2406, opened 2021-08-04
Various bug fixes and improvements to clustered filestore consumer stores.

derekcollison:
/cc @nats-io/core

### DIFF
--- a/server/filestore.go
+++ b/server/filestore.go
@@ -4607,6 +4607,11 @@ func (o *consumerFileStore) UpdateDelivered(dseq, sseq, dc uint64, ts int64) err
 		return ErrNoAckPolicy
 	}
 
+	// On restarts the old leader may get a replay from the raft logs that are old.
+	if dseq <= o.state.Delivered.Consumer {
+		return nil
+	}
+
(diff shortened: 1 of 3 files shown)


server/filestore.go
```diff
@@ -4800,6 +4830,13 @@ func (o *consumerFileStore) Update(state *ConsumerState) error {
 
 	// Replace our state.
 	o.mu.Lock()
+
+	// Check to see if this is an outdated update.
+	if state.Delivered.Consumer < o.state.Delivered.Consumer {
+		o.mu.Unlock()
+		return fmt.Errorf("old update ignored")
```
kozlovic, line 4837: Just checking: is an outdated really bad, indicative of a bug etc.. or is that "normal" in some situations? If normal, then we should return nil here otherwise upper-layer may fail.

kozlovic (approved):
LGTM but a question about outdated update.

server/filestore.go
derekcollison: Was a bug, removed the bug but kept the check. So that does not get triggered now running full test suite.

PR #2409, opened 2021-08-04
attempt to improve UX of the error system

ripienaar:
Previously we had a few confusing functions like NewT
and similar that were quite fragile to use due to minimal
validation and a panic in go stdlib string Replacer.

Now we generate helper methods for every string, these
are used to access errors, fill in templates and conditional
returns of error type using the new Unless() option

We now get compile time errors for some common mistakes
and have better IDE helpers for arguments etc

/cc @nats-io/core

### DIFF
--- a/server/jetstream.go
+++ b/server/jetstream.go
@@ -867,7 +867,7 @@ func (s *Server) JetStreamNumAccounts() int {
 func (s *Server) JetStreamReservedResources() (int64, int64, error) {
 	js := s.getJetStream()
 	if js == nil {
-		return -1, -1, ApiErrors[JSNotEnabledForAccountErr]
+		return -1, -1, NewJSNotEnabledForAccountError()
 	}
 	js.mu.RLock()
 	defer js.mu.RUnlock()
@@ -911,7 +911,7 @@ func (a *Account) EnableJetStream(limits *JetStreamAccountLimits) error {
 	js := s.getJetStream()
 	if js == nil {
 		a.assignJetStreamLimits(limits)
-		return ApiErrors[JSNotEnabledErr]
+		return NewJSNotEnabledError()
 	}
 
 	if s.SystemAccount() == a {
@@ -1227,14 +1227,14 @@ func (a *Account) lookupStream(name string) (*stream, error) {
 	a.mu.RUnlock()
 
 	if jsa == nil {
-		return nil, ApiErrors[JSNotEnabledForAccountErr]
+		return nil, NewJSNotEnabledForAccountError()
 	}
 	jsa.mu.Lock()
 	defer jsa.mu.Unlock()
 
 	mset, ok := jsa.streams[name]
 	if !ok {
-		return nil, ApiErrors[JSStreamNotFoundErr]
+		return nil, NewJSStreamNotFoundError()
 	}
 	return mset, nil
 }
@@ -1251,10 +1251,10 @@ func (a *Account) UpdateJetStreamLimits(limits *JetStreamAccountLimits) error {
 	}
 	js := s.getJetStream()
 	if js == nil {
-		return ApiErrors[JSNotEnabledErr]
+		return NewJSNotEnabledError()
 	}
 	if jsa == nil {
-		return ApiErrors[JSNotEnabledForAccountErr]
+		return NewJSNotEnabledForAccountError()
 	}
 
 	if limits == nil {
@@ -1348,7 +1348,7 @@ func (a *Account) removeJetStream() error {
 
 	js := s.getJetStream()
 	if js == nil {
-		return ApiErrors[JSNotEnabledForAccountErr]
+		return NewJSNotEnabledForAccountError()
 	}
 
 	return js.disableJetStream(js.lookupAccount(a))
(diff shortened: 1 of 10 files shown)


ripienaar:
I marked this as draft as I want to carefully double check some stuff, feedback welcome. I think this greatly simplify the usage of these things.

To get even greater control we can adjust our tags from like `{foo}` to `{foo:int64}` and I can then add stricted type control on the helper functions, this would be a huge win, but I didnt do that now want to hear feedback first.

ripienaar:
To summarize this:

 * Replace all `e := ApiErrors[Foo]` with `e := NewFooError()`
 * Replace all `ApiErrors[Foo].ErrOrNewT(err, "{err}", err, "{foo}", foo)` with `NewFooError(err, foo, Unless(err))`
 * Replace all `ApiErrors[Foo].ErrOr(err)` with `NewFooError(Unless(err))`

and some other little things, one function, one pattern, all the same, much nicer.

ripienaar:
Changing `e := ApiErrors[Foo]` to `e := NewFooError()` seems redundant - its functionally identical. It improve consistency though its only ever one way.

Also I was chatting with @matthiashanel around tracing work, if via work he is thinking I put my account into trace mode and I give it a trace subject then I think I want every time an `NewXError()` is called a trace emitted to that subject from within the account with some context - line number and file etc - so using only the functions everywhere means I can bring that in with a code change only to the generated code nothing else.

server/jetstream.go
```diff
@@ -1522,24 +1522,24 @@ func (jsa *jsAccount) checkBytesLimits(addBytes int64, storage StorageType, repl
...
-				return ApiErrors[JSMemoryResourcesExceededErr]
+				return NewJSMemoryResourcesExceededError()
 			}
 		}
 	case FileStorage:
 		// Account limits defined.
 		if jsa.limits.MaxStore > 0 {
 			if jsa.storeReserved+totalBytes > jsa.limits.MaxStore {
-				return ApiErrors[JSStorageResourcesExceededErr]
+				return NewJSMemoryResourcesExceededError()
```
kozlovic, line 1537: Should it be NewJSStorageResourcesExceededError here? (and the one below)

This would solve this test failure:
```
=== RUN   TestJetStreamClusterMaxBytesForStream
    jetstream_cluster_test.go:964: Expected "insufficient storage resources" error, got "insufficient memory resources available"
--- FAIL: TestJetStreamClusterMaxBytesForStream (1.70s)
```

kozlovic (commented):
Not against the idea. Just a comment that 2 tests seem to be failing. One can be fixed by using the proper NewJS error function for file store, but this one:
```
=== RUN   TestJetStreamClusterStreamLimitWithAccountDefaults
    jetstream_cluster_test.go:143: Expected JSStorageResourcesExceededErr but got insufficient memory resources available
--- FAIL: TestJetStreamClusterStreamLimitWithAccountDefaults (0.43s)
```
I am not sure. Is it because this test uses `if !IsNatsErr(err, JSStorageResourcesExceededErr) {` and should be replaced with something else?

server/jetstream.go
ripienaar: Yes - I am sure there's a few of these, this is what I meant with carefully `carefully double check some stuff` :)

ripienaar:
> Not against the idea. Just a comment that 2 tests seem to be failing. One can be fixed by using the proper NewJS error function for file store, but this one:
> 
> ```
> === RUN   TestJetStreamClusterStreamLimitWithAccountDefaults
>     jetstream_cluster_test.go:143: Expected JSStorageResourcesExceededErr but got insufficient memory resources available
> --- FAIL: TestJetStreamClusterStreamLimitWithAccountDefaults (0.43s)
> ```
> 
> I am not sure. Is it because this test uses `if !IsNatsErr(err, JSStorageResourcesExceededErr) {` and should be replaced with something else?

I have not looked at the test failures at all yet - but `IsNatsErr` should fully support these errors as is, so somethings going on, I will investigate.

For now just seeing feedback on general approach before I go too deep on fixing all the knock on stuff

kozlovic:
> For now just seeing feedback on general approach before I go too deep on fixing all the knock on stuff

I think the approach is good.

ripienaar:
> Not against the idea. Just a comment that 2 tests seem to be failing. One can be fixed by using the proper NewJS error function for file store, but this one:
> 
> ```
> === RUN   TestJetStreamClusterStreamLimitWithAccountDefaults
>     jetstream_cluster_test.go:143: Expected JSStorageResourcesExceededErr but got insufficient memory resources available
> --- FAIL: TestJetStreamClusterStreamLimitWithAccountDefaults (0.43s)
> ```
> 
> I am not sure. Is it because this test uses `if !IsNatsErr(err, JSStorageResourcesExceededErr) {` and should be replaced with something else?

Ah this one fails because its via the go client and it doesnt support new errors yet

ripienaar:
I think this one is good for review

ripienaar:
@derekcollison you good with this? I would merge after the LGTM from Ivan but you have some big PRs in flight so dont want to cause a bunch of rebases for you, let me know whe you're ready for me to merge this

derekcollison:
If you both are good that is good for me. Merge at your discretion.

ripienaar:
Merged this with one failing test, GHA passed 1.16, CircleCI just wouldnt.  Pressed rerun for over 3 hours, its just not worth it.

PR #2410, opened 2021-08-04
[FIXED] #2401

### LINKED ISSUE #2401: When two server are connected via leaf node and both have JS enabled and domain set to the same value, JS is currently not extended
It would be expected that they behave as if domain where not set at all.
Meaning they should be part of the same meta group.

derekcollison:
Leafnodes with same domain and shared system account should behave like flat jetstream network.

Resolves #2401 

/cc @nats-io/core

### DIFF
--- a/server/jetstream_cluster_test.go
+++ b/server/jetstream_cluster_test.go
@@ -6615,6 +6615,21 @@ func TestJetStreamClusterLeafNodesWithoutJS(t *testing.T) {
 	testJS(sl, "HUB", true)
 }
 
+func TestJetStreamClusterLeafNodesWithSameDomainNames(t *testing.T) {
+	tmpl := strings.Replace(jsClusterAccountsTempl, "store_dir:", "domain: HUB, store_dir:", 1)
+	c := createJetStreamCluster(t, tmpl, "HUB", _EMPTY_, 3, 11233, true)
+	defer c.shutdown()
+
+	tmpl = strings.Replace(jsClusterTemplWithLeafNode, "store_dir:", "domain: HUB, store_dir:", 1)
+	lnc := c.createLeafNodesWithTemplateAndStartPort(tmpl, "SPOKE", 3, 11311)
+	defer lnc.shutdown()
+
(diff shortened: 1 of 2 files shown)


matthiashanel (approved):
LGTM

server/jetstream_cluster_test.go
```diff
@@ -6615,6 +6615,21 @@ func TestJetStreamClusterLeafNodesWithoutJS(t *testing.T) {
...
+func TestJetStreamClusterLeafNodesWithSameDomainNames(t *testing.T) {
+	tmpl := strings.Replace(jsClusterAccountsTempl, "store_dir:", "domain: HUB, store_dir:", 1)
+	c := createJetStreamCluster(t, tmpl, "HUB", _EMPTY_, 3, 11233, true)
+	defer c.shutdown()
+
+	tmpl = strings.Replace(jsClusterTemplWithLeafNode, "store_dir:", "domain: HUB, store_dir:", 1)
+	lnc := c.createLeafNodesWithTemplateAndStartPort(tmpl, "SPOKE", 3, 11311)
+	defer lnc.shutdown()
+
+	//lncm := c.createLeafNodesWithTemplateMixedMode(tmpl, "SPOKE", 3, 2, true)
```
kozlovic, line 6627: Remove?

server/leafnode.go
```diff
@@ -1057,7 +1058,7 @@ func (c *client) processLeafnodeInfo(info *Info) {
 						c.Debugf("Error adding JetStream domain mapping: %v", err)
 					}
 				}
-			} else if hasJSDomain {
+			} else if hasJSDomain && opts.JetStreamDomain != info.Domain {
```
kozlovic, line 1061: Do we need to treat `info.Domain` possibly empty (old server?) differently here or not?

kozlovic (approved):
LGTM but not sure if info.Domain empty should be treated differently.

server/jetstream_cluster_test.go
derekcollison: Yes will update.

server/leafnode.go
derekcollison: No for this one we want both empty or exact match.

PR #2411, opened 2021-08-04
[FIXED] #2402

### LINKED ISSUE #2402: When dealing with accounts or domains, a single ACK can acknowledge multiple messages.
across accounts as well as domains, stream and durable names can be duplicated. 
Furthermore $JS.ACK.> is not bound to a domain/api prefix. 
Therefore the information provided when acking is not sufficient. 
Furthermore this may cause excess traffic.

Ideally this would be moved under $JS.API.>

Alternatives:
1) we add the denies to be `$JS.>` instead of `$JS.API.>` and add a mapping `$JS`->`$JS.<domain>`. This would effectively freeze API as part of the prefix. It would break installations where ppl altered the prefix to not end in API. 
This is a guess, but so far I wouldn't expect this to happen often. 

2) Or add more denies and mappings similar to what `$JS.API` does to `$JS.ACK`. just for `$JS.ACK` -> `$JS.<domain>.ACK.>`
I assume this needs to be done for `$JS.FC` as well as `$JS.SNAPSHOT.` 
For `$JS.EVENT.` we may need a deny only, but not mapping as this is a stream and not a service.

In either case clients would have to start using the api prefix for acks/fc as well.

derekcollison:
Tighten deny clauses for JetStream over leafnodes when separating domains.

Resolves #2402 (for now)

/cc @nats-io/core

### DIFF
--- a/server/leafnode.go
+++ b/server/leafnode.go
@@ -116,19 +116,42 @@ func (s *Server) addInJSDeny(r *leafNodeCfg) {
 	s.addInJSDenyImport(r)
 }
 
+// To deny all JS traffic.
+var jsDenyAll = []string{jsAllAPI, jsAckAPI, jsFlowControlAPI, jsSnapshotAPI}
+
+// Lock should be held.
+func (r *leafNodeCfg) containsDenyExport(deny string) bool {
+	for _, dsubj := range r.DenyExports {
+		if dsubj == deny {
+			return true
+		}
+	}
+	return false
+}
+
+// Lock should be held.
+func (r *leafNodeCfg) containsDenyImport(deny string) bool {
+	for _, dsubj := range r.DenyImports {
+		if dsubj == deny {
+			return true
+		}
+	}
+	return false
+}
+
 // Will add in the deny export for JetStream on solicited connections if we
 // detect we have multiple JetStream domains and we know our local account
 // is JetStream enabled.
 // r lock should be held.
 func (s *Server) addInJSDenyExport(r *leafNodeCfg) {
-	for _, dsubj := range r.DenyExports {
-		if dsubj == jsAllAPI {
-			return
+	for _, deny := range jsDenyAll {
+		if r.containsDenyExport(deny) {
+			continue
 		}
+		// Not present so add here.
+		s.Noticef("Adding deny export of %q for leafnode configuration on %q that bridges system account", deny, r.LocalAccount)
+		r.DenyExports = append(r.DenyExports, deny)
 	}
-
-	s.Noticef("Adding deny export of %q for leafnode configuration on %q that bridges system account", jsAllAPI, r.LocalAccount)
-	r.DenyExports = append(r.DenyExports, jsAllAPI)
 
 	// We added in some deny clauses here so need to regenerate the permissions etc.
 	perms := &Permissions{}
@@ -144,14 +167,14 @@ func (s *Server) addInJSDenyExport(r *leafNodeCfg) {
 // is JetStream enabled.
 // r lock should be held.
 func (s *Server) addInJSDenyImport(r *leafNodeCfg) {
-	for _, dsubj := range r.DenyImports {
-		if dsubj == jsAllAPI {
-			return
+	for _, deny := range jsDenyAll {
+		if r.containsDenyImport(deny) {
+			continue
 		}
-	}
 
-	s.Noticef("Adding deny import of %q for leafnode configuration on %q that bridges system account", jsAllAPI, r.LocalAccount)
-	r.DenyImports = append(r.DenyImports, jsAllAPI)
+		s.Noticef("Adding deny import of %q for leafnode configuration on %q that bridges system account", deny, r.LocalAccount)
+		r.DenyImports = append(r.DenyImports, deny)
+	}
 
 	// We added in some deny clauses here so need to regenerate the permissions etc.
 	perms := &Permissions{}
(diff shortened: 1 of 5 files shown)


server/leafnode.go
```diff
@@ -116,19 +116,22 @@ func (s *Server) addInJSDeny(r *leafNodeCfg) {
...
-		if dsubj == jsAllAPI {
+		if dsubj == jsDenyAll {
 			return
 		}
 	}
 
-	s.Noticef("Adding deny export of %q for leafnode configuration on %q that bridges system account", jsAllAPI, r.LocalAccount)
-	r.DenyExports = append(r.DenyExports, jsAllAPI)
+	s.Noticef("Adding deny export of %q for leafnode configuration on %q that bridges system account", jsDenyAll, r.LocalAccount)
+	r.DenyExports = append(r.DenyExports, jsDenyAll)
```
matthiashanel, line 134: Because `$JS.>` covers the result of our mapping, I would not expected this to work.
My assumption is that because we deny `$JS.API.>` here, the result of the mapping `$JS.<domain>.API...` would not be denied.
Now we deny `$JS.>` but our mapping from $JS.<domain>.. is covered by it.

 
My expectation would be to see something like this instead. 
r.DenyExports = append(r.DenyExports, jsAllAPI)
r.DenyExports = append(r.DenyExports, jsAckAPI)
r.DenyExports = append(r.DenyExports, jsFlowControlAPI)

server/leafnode.go
```diff
@@ -145,13 +148,13 @@ func (s *Server) addInJSDenyExport(r *leafNodeCfg) {
...
-		if dsubj == jsAllAPI {
+		if dsubj == jsDenyAll {
 			return
 		}
 	}
 
-	s.Noticef("Adding deny import of %q for leafnode configuration on %q that bridges system account", jsAllAPI, r.LocalAccount)
-	r.DenyImports = append(r.DenyImports, jsAllAPI)
+	s.Noticef("Adding deny import of %q for leafnode configuration on %q that bridges system account", jsDenyAll, r.LocalAccount)
+	r.DenyImports = append(r.DenyImports, jsDenyAll)
```
matthiashanel, line 157: Same as earlier comment

server/leafnode.go
```diff
@@ -165,7 +168,7 @@ func (s *Server) addInJSDenyImport(r *leafNodeCfg) {
 // Used for $SYS accounts when sharing but using separate JS domains.
 // r lock should be held.
 func (s *Server) addInJSDenyAll(r *leafNodeCfg) {
-	denyAll := []string{jscAllSubj, raftAllSubj, jsAllAPI}
+	denyAll := []string{jscAllSubj, raftAllSubj, jsDenyAll}
```
matthiashanel, line 171: denyAll := []string{jscAllSubj, raftAllSubj, jsAllAPI, jsAckAPI, jsFlowControlAPI}

server/leafnode.go
```diff
@@ -1362,7 +1365,7 @@ func (c *client) processLeafNodeConnect(s *Server, arg []byte, lang string) erro
 	// If we have JS enabled and the other side does as well we need to add in an import deny clause.
 	if jsConfigured && proto.JetStream {
 		// We should never have existing perms here, if that changes this needs to be reworked.
-		c.setPermissions(&Permissions{Publish: &SubjectPermission{Deny: []string{jsAllAPI}}})
+		c.setPermissions(&Permissions{Publish: &SubjectPermission{Deny: []string{jsDenyAll}}})
```
matthiashanel, line 1368: []string{jsAllAPI, jsAckAPI, jsFlowControlAPI}

kozlovic (commented):
Not sure, but although this test passes for me locally, it has failed on both GA and Travis, so either test needs tweaking or that shows an issue:
```
=== RUN   TestJetStreamClusterSingleLeafNodeWithoutSharedSystemAccount
    jetstream_cluster_test.go:6073: Unexpected error: nats: jetstream not enabled
--- FAIL: TestJetStreamClusterSingleLeafNodeWithoutSharedSystemAccount (1.38s)
```

server/leafnode.go
```diff
@@ -1053,7 +1056,7 @@ func (c *client) processLeafnodeInfo(info *Info) {
 				// allow access cross domain for each non-system account.
 				if hasJSDomain && acc.jetStreamConfigured() {
 					src := fmt.Sprintf(jsDomainAPI, opts.JetStreamDomain)
-					if err := acc.AddMapping(src, jsAllAPI); err != nil {
+					if err := acc.AddMapping(src, jsDenyAll); err != nil {
```
matthiashanel, line 1059: Here, my expectation would be to see:
```
src := fmt.Sprintf("$JS.%s.>", opts.JetStreamDomain)
```
which would result in a mapping from `$JS.<domain>.>` to `$JS.>`, essentially stripping the domain.
Because we had more detailed denies earlier, messages sent with the prefix `$JS.<domain>` would not be denied.

Thus messages sent to anything with the prefix `$JS.<domain>.ACK` would not be blocked and see the domain stripped to be the prefix`$JS.ACK` instead.

For what is currently under API, it'd be `$JS.<domain>.API` -> `$JS.API`

Same for FC and SNAPSHOT

server/leafnode.go
```diff
@@ -1393,7 +1396,7 @@ func (c *client) processLeafNodeConnect(s *Server, arg []byte, lang string) erro
 	// allow access cross domain for each non-system account.
 	if doDomainMappings {
 		src := fmt.Sprintf(jsDomainAPI, opts.JetStreamDomain)
-		if err := acc.AddMapping(src, jsAllAPI); err != nil {
+		if err := acc.AddMapping(src, jsDenyAll); err != nil {
```
matthiashanel, line 1399: same as other mapping comment

matthiashanel (commented):
I would have made changes at the same locations, but would have picked different denies and mappings.

matthiashanel:
In addition I'd make the domain names `ACK`, `API`, `FC` illegal.
Alternatively, disallow all upper case domain names.

derekcollison:
ok made some updates but also made sure we could ack cross domain and we can't yet. I know Matthias has some comments but short version is we will wait on this and think a bit more. Will convert to draft.

derekcollison:
This is better handled with @matthiashanel work.

PR #2413, opened 2021-08-04
[CHANGED] Check that max_payload is not greater than max_pending

kozlovic:
This is related to PR #2407. Since the 64MB pending size is actually
configurable, we should fail only if max_payload is greater than
the configured max_pending. This is done in validateOptions() which
covers both config file and direct options in embedded cases.
The check in opts.go is reverted to max int32 since at this point
we don't know if/what max_pending will be, so we simply check
that it is not more than a int32.

For the next minor release, we could have another change that
imposes a lower limit to max_payload (regardless if max_pending
is higher).

/cc @nats-io/core

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -1549,7 +1549,8 @@ func (s *Server) Start() {
 		s.Warnf("Trusted Operators should utilize a System Account")
 	}
 	if opts.MaxPayload > MAX_PAYLOAD_MAX_SIZE {
-		s.Warnf("Maximum payloads over 8MB are generally discouraged and could lead to poor performance")
+		s.Warnf("Maximum payloads over %v are generally discouraged and could lead to poor performance",
+			friendlyBytes(int64(MAX_PAYLOAD_MAX_SIZE)))
 	}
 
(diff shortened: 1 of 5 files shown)


server/server.go
```diff
@@ -599,6 +599,10 @@ func validateOptions(o *Options) error {
 		return fmt.Errorf("lame duck grace period (%v) should be strictly lower than lame duck duration (%v)",
 			o.LameDuckGracePeriod, o.LameDuckDuration)
 	}
+	if int64(o.MaxPayload) > o.MaxPending {
```
derekcollison, line 602: Should we add a warning for anything over say 8M? Just suggesting that is not recommended?

derekcollison (approved):
LGTM - Small comment.

server/server.go
kozlovic: Added in next commit: https://github.com/nats-io/nats-server/pull/2413/commits/d1365b74120c63e68c4bfcda336e97ad35a8991f

kozlovic: Actually pushed other commit to use the constant instead of hardcoding in warning message.

PR #2418, opened 2021-08-06
[FIXED] #2417

### LINKED ISSUE #2417: Stream delete can fail for non empty directory
During concurrent consumer create/delete and stream create/delete it can happen that a stream directory during delete is not empty 

```
stream delete failed: unlinkat /var/folders/2p/lsd279293x7618xcjfl2bc380000gn/T/jstest632941891/jetstream/$G/streams/ELECTION: directory not empty (10050)
```

derekcollison:

Resolves #2417 

/cc @nats-io/core

### DIFF
--- a/server/filestore.go
+++ b/server/filestore.go
@@ -4145,7 +4145,19 @@ func (fs *fileStore) Delete() error {
 	if err := fs.Stop(); err != nil {
 		return err
 	}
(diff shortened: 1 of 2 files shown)


ripienaar (approved):
LGTM I guess, seems hacky but also cant think of a better way :)

PR #2421, opened 2021-08-08
[FIXED] Subscription leaks when auto-unsub used after delivered > max.

derekcollison:
This has been here a very long time but until JetStream I don't think it was being triggered out in the wild.

/cc @nats-io/core


ripienaar:
LGTM

kozlovic (commented):
LGTM

PR #2425, opened 2021-08-10
Simplify flow control for consumers for mirrors and sources.

derekcollison:
R.I. had reported stalls with mirrors that I believe was tied to flow control.
Simplified the flow control, especially for clustered streams. Also tries to avoid stalls due to msg loss of flow control replies.

/cc @nats-io/core


ripienaar:
LGTM from functionality wise, dont know the code in question

ripienaar (approved):
LGTM

PR #2426, opened 2021-08-10
Fix for #2420

### LINKED ISSUE #2420: consumer info max_msgs_per_subject default to 0 should be -1
## Defect

If `max_msgs_per_subject` is not provided during stream create, it's value becomes 0, not -1. This is different behavior than
`max_consumers`, `max_msgs`, `max_bytes` and `max_msg_size`

nats-server: v2.3.5-beta

Windows

Execute `$JS.API.STREAM.CREATE.foo` with a minimal payload, i.e. no `max_` values are supplied.

```
{"name":"foo","subjects":["bar"],"retention":"limits","storage":"memory","num_replicas":1,"no_ack":false,"discard":"old"}
```

The response comes back with 

```
{
	"name": "foo",
	"subjects": ["bar"],
	"retention": "limits",
	"max_consumers": -1,
	"max_msgs": -1,
	"max_bytes": -1,
	"max_age": 0,
	"max_msgs_per_subject": 0,
	"max_msg_size": -1,
	"discard": "old",
	"storage": "memory",
	"num_replicas": 1,
	"duplicate_window": 120000000000
}
```

#### Expected result:

max_msgs_per_subject is -1

#### Actual result:

max_msgs_per_subject is -0

derekcollison:

Resolves #2420

/cc @nats-io/core


ripienaar (commented):
we probably need to update schemas and a few other things too as this is essnetially a breaking change.  Configs that was previously put in would now not be comparible equal anymore and so idempotent stream creations will now fail.  This will also break the terraform code.

Is it not possible to just treat 0 as unlimited?

derekcollison:
Needs to be consistent, was an oversight. However I am not following the issue with creating streams etc?

ripienaar:
> Needs to be consistent, was an oversight. However I am not following the issue with creating streams etc?

systems like terraform works on the basis of putting a configuration in and then later comparing the configuration retrieved with what was put in, and then tries to reconcile any differences.

This change will not break that basic contract where what went in wont match what came out so unexpected changes will happen.

Same would happen with anyone using GH ACtions to configure JS, k8s CRDs (which are luckily very out of date) and even the CLI.

In reality its fine as I have not done a terraform release since adding this property - but in general these types of changes should be considered breaking changes.

derekcollison:
We can put code in to make sure updates work etc, will take some work but I can do it if you think we need that beahvior.

ripienaar:
> We can put code in to make sure updates work etc, will take some work but I can do it if you think we need that behavior.

Think its new enough that we can get away with it

derekcollison:
ok will wait on LGTM

ripienaar (approved):
LGTM

PR #2430, opened 2021-08-11
[fixed] issue where js overwrote leafnode remotes permissions from creds

### LINKED ISSUE #2415: User authorization in leafnode connection doesn't work in v2.3.4
## Defect
+  user authorization doesnt work in leafnode connection when account(created by nsc) in central cluster has jetstream enabled.
+  after I changed my image from `nats:2.3.4-alpine` to `nats:2.2.6-alpine3.13` ,it backs to work

docker image : nats:2.3.4-alpine
natscli : the latest dev version 

Linux mpc-ubuntu-server 5.4.0-80-generic #90-Ubuntu SMP Fri Jul 9 22:49:44 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux
EMU`D04KL{GA9Q16W})P DF

1. use nsc to create 3 accounts as follow ,and each account has 4 users.
```
a: {
   jetstream:enable
   users:[
   {a1,deny-pubsub:test.a1},
   {a2,deny-pubsub:test.a2},
   {a3,deny-pubsub:test.a3},
   {a4,deny-pubsub:test.a4},
   ]
}

b: {
   users:[
   {b1,deny-pubsub:test.b1},
   {b2,deny-pubsub:test.b2},
   {b3,deny-pubsub:test.b3},
   {b4,deny-pubsub:test.b4},
   ]
}

c: {
   jetstream:enable
   users:[
   {c1,deny-pubsub:test.c1},
   {c2,deny-pubsub:test.c2},
   {c3,deny-pubsub:test.c3},
   {c4,deny-pubsub:test.c4},
   ]
}
```

2. establish a central cluster of 3 nodes.each node uses nats-based-resolver and preload the accounts above.
3. deploy 4 leafnodes(leaf1~leaf4) connecting to the central cluster . each node has three account named A,B,C 
```
leafnode connection :
leaf1 
A->a1->a
B->b1->b
C->c1->c

leaf2
A->a2->a
B->b2->b
C->c2->c
....
```
4. use client1 to sub test.a1 in leaf1(account :A) while client2 to pub msg to test.a1 in leaf2(account:A)
5. use client3 to sub test.b1 in leaf1(account :B) while client4 to pub msg to test.b1 in leaf2(account:B)
6. use client5 to sub test.c1 in leaf1(account :C) while client6 to pub msg to test.c1 in leaf2(account:C)

#### Expected result:
client1,3,5 can't receive msg sent by client2,4,6

#### Actual result:
client 2,4,6 successed  to pub msg 
client1,5 received the msg sent by client 2,6
client 3 cant receive msgs sent by client 4

matthiashanel:

Fixes #2415. We did a set instead of merge.
changes in `jwt_test.go` are to make the `createUserWithLimit` usable by my new test.

### DIFF
--- a/server/leafnode.go
+++ b/server/leafnode.go
@@ -1362,8 +1362,7 @@ func (c *client) processLeafNodeConnect(s *Server, arg []byte, lang string) erro
 
 	// If we have JS enabled and the other side does as well we need to add in an import deny clause.
 	if jsConfigured && proto.JetStream {
-		// We should never have existing perms here, if that changes this needs to be reworked.
-		c.setPermissions(&Permissions{Publish: &SubjectPermission{Deny: []string{jsAllAPI}}})
+		c.mergePubDenyPermissions([]string{jsAllAPI})
 	}
 
 	// Set the Ping timer
--- a/server/jwt_test.go
+++ b/server/jwt_test.go
@@ -3864,7 +3864,7 @@ func newTimeRange(start time.Time, dur time.Duration) jwt.TimeRange {
 	return jwt.TimeRange{Start: start.Format("15:04:05"), End: start.Add(dur).Format("15:04:05")}
 }
 
-func createUserWithLimit(t *testing.T, accKp nkeys.KeyPair, expiration time.Time, limits func(*jwt.Limits)) string {
+func createUserWithLimit(t *testing.T, accKp nkeys.KeyPair, expiration time.Time, limits func(*jwt.UserPermissionLimits)) string {
 	t.Helper()
 	ukp, _ := nkeys.CreateUser()
 	seed, _ := ukp.Seed()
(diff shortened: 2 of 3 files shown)


server/leafnode.go
```diff
@@ -1362,8 +1362,7 @@ func (c *client) processLeafNodeConnect(s *Server, arg []byte, lang string) erro
 
 	// If we have JS enabled and the other side does as well we need to add in an import deny clause.
 	if jsConfigured && proto.JetStream {
-		// We should never have existing perms here, if that changes this needs to be reworked.
-		c.setPermissions(&Permissions{Publish: &SubjectPermission{Deny: []string{jsAllAPI}}})
+		c.mergePubDenyPermissions([]string{jsAllAPI})
```
kozlovic, line 1365: Based on the removed comment above, this is should not happen, so either comment was correct at one point of time but no longer the case, was never correct, or is still is and the merge approach is wrong. I have no idea which is which. I have ran tests and they all pass, so I would think the change is ok (and fixes the user report).

kozlovic (approved):
LGTM, but read my comment.

server/leafnode.go
kozlovic: Understood.

PR #2431, opened 2021-08-11
fixed unit test to require fc export as well

matthiashanel:

Realized that FC needed to be exported as well...


kozlovic (approved):
LGTM

PR #2432, opened 2021-08-11
JetStream additions

derekcollison:
This adds two items.

1. Domain information for PubAcks.
2. Active state for push consumers that details active subject and optionally queue group.

/cc @nats-io/core

### DIFF
--- a/server/consumer.go
+++ b/server/consumer.go
@@ -43,6 +43,7 @@ type ConsumerInfo struct {
 	NumRedelivered int             `json:"num_redelivered"`
 	NumWaiting     int             `json:"num_waiting"`
 	NumPending     uint64          `json:"num_pending"`
+	Active         *PushActive     `json:"active,omitempty"`
 	Cluster        *ClusterInfo    `json:"cluster,omitempty"`
 }
 
@@ -67,6 +68,11 @@ type ConsumerConfig struct {
 
 	// Don't add to general clients.
 	Direct bool `json:"direct,omitempty"`
+}
+
+type PushActive struct {
+	Subject string `json:"subject"`
+	Queue   string `json:"queue,omitempty"`
 }
 
 type CreateConsumerRequest struct {
@@ -190,6 +196,7 @@ type consumer struct {
 	asflr             uint64
 	sgap              uint64
 	dsubj             string
+	qgroup            string
 	lss               *lastSeqSkipList
 	rlimit            *rate.Limiter
 	reqSub            *subscription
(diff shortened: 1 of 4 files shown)


server/consumer.go
```diff
@@ -69,6 +70,11 @@ type ConsumerConfig struct {
 	Direct bool `json:"direct,omitempty"`
 }
 
+type PushActive struct {
+	Subject string
+	Queue   string
```
kozlovic, line 75: Should this have json tag and specifically Queue be omitempty?

derekcollison: lol yes. One sec.

derekcollison: ok pushed.

server/consumer.go
```diff
@@ -69,6 +70,11 @@ type ConsumerConfig struct {
 	Direct bool `json:"direct,omitempty"`
 }
 
+type PushActive struct {
+	Subject string `json:"subject,omitempty"`
```
kozlovic, line 74: I think subject will always be present, but that's ok.

kozlovic (approved):
LGTM (at least the PushActive part)

server/consumer.go
derekcollison: Yes agree, will update and if local passes will merge.

kozlovic (approved):
LGTM

server/consumer.go
```diff
@@ -69,6 +70,11 @@ type ConsumerConfig struct {
 	Direct bool `json:"direct,omitempty"`
 }
 
+type PushActive struct {
+	Subject string `json:"subject"`
+	Queue   string `json:"queue,omitempty"`
```
ripienaar, line 75: but there could be multiple queue groups on the same subject - and some unqueued?

server/stream.go
```diff
@@ -382,10 +382,11 @@ func (a *Account) addStreamWithAssignment(config *StreamConfig, fsConfig *FileSt
...
 	// Create our pubAck template here. Better than json marshal each time on success.
-	b, _ := json.Marshal(&JSPubAckResponse{PubAck: &PubAck{Stream: cfg.Name, Sequence: math.MaxUint64}})
-	end := bytes.Index(b, []byte(strconv.FormatUint(math.MaxUint64, 10)))
-	// We need to force cap here to make sure this is a copy when sending a response.
-	mset.pubAck = b[:end:end]
+	if domain := s.getOpts().JetStreamDomain; domain != _EMPTY_ {
+		mset.pubAck = []byte(fmt.Sprintf("{%q:%q, %q:%q, %q:", "stream", cfg.Name, "domain", domain, "seq"))
+	} else {
+		mset.pubAck = []byte(fmt.Sprintf("{%q:%q, %q:", "stream", cfg.Name, "seq"))
+	}
```
ripienaar, line 389: After this pubacks are broken.

I am seeing acks like:

```
{"stream":"GOVERNOR_TEST", "seq":10
```

no closing `}`

server/consumer.go
derekcollison: No, once a consumer is bound to a queue group, meaning we know you want to do scale out on this consumer, its bound to that queue group.

server/stream.go
derekcollison: I am looking at a race at the moment in that section, will have PR shortly.

server/consumer.go
ripienaar: No, 2 nats clients can have queue groups on this subject, 2 different queue groups. And there could also be subscribes on it that has no queue group - while others does.

You specifically take only the first, but thats just not deterministic at all because if there are 2 clients each with its own queue group one server will report them one way and another another way depending on how and when splits happened.

derekcollison: Just like two apps having different delivery subjects (e.g. _INBOX.>) only one is bound to the server side consumer. This is similar in that only one queue group should be bound to the delivery subject for scale out binding to a single consumer.

If someone binds that shouldn't, the app that *should* be bound will fail to bind with a queue mismatch error.

server/stream.go
derekcollison: https://github.com/nats-io/nats-server/pull/2434

PR #2437, opened 2021-08-13
Added in ability for normal accounts to access scoped connz info.

derekcollison:
Added in client kind and sub type for clients.
Added in ability to filter connections based on matching subject interest.

/cc @nats-io/core


server/events.go
```diff
@@ -57,6 +57,7 @@ const (
 	leafNodeConnectEventSubj = "$SYS.ACCOUNT.%s.LEAFNODE.CONNECT" // for internal use only
 	remoteLatencyEventSubj   = "$SYS.LATENCY.M2.%s"
 	inboxRespSubj            = "$SYS._INBOX.%s.%s"
+	accConnzReqSubj          = "$SYS.REQ.ACCOUNT.PING.CONNZ"
```
matthiashanel, line 60: I am a bit confused about PING here. 
Ping is implied and it overlaps with "$SYS.REQ.ACCOUNT.%s.%s" where the last token is what is requested (CONNZ) and the one before is the requested account id.

server/events.go
```diff
@@ -1329,6 +1370,29 @@ func (s *Server) remoteConnsUpdate(sub *subscription, _ *client, _ *Account, sub
...
+	mappedSubj := fmt.Sprintf(accReqSubj, a.Name, "CONNZ")
+
+	// Add in this to the account in 2 places.
+	// "$SYS.REQ.SERVER.PING.CONNZ" and "$SYS.REQ.ACCOUNT.CONNZ"
+	if err := a.AddServiceImport(sacc, connzSubj, mappedSubj); err != nil {
+		s.Errorf("Error setting up system service imports for account: %v", err)
+	}
+	if err := a.AddServiceImport(sacc, accConnzReqSubj, mappedSubj); err != nil {
+		s.Errorf("Error setting up system service imports for account: %v", err)
+	}
```
matthiashanel, line 1393: Not sure about this one. the service is mapped to the same subject ("$SYS.REQ.ACCOUNT.%s.%s", with a.Name and CONNZ).
I think this is redundant. 
Specifically because we don't really export accConnzReqSubj. To me this seems to work because it has the same layout than accReqSubj

server/monitor.go
```diff
@@ -309,14 +369,28 @@ func (s *Server) Connz(opts *ConnzOptions) (*Connz, error) {
...
 	s.mu.Unlock()
 
+	// Filter by subject now if needed. We do this outside of server lock.
+	if filter != _EMPTY_ {
+		var oc []*client
+		for _, c := range openClients {
+			c.mu.Lock()
+			for _, sub := range c.subs {
+				if SubjectsCollide(filter, string(sub.subject)) {
+					oc = append(oc, c)
```
matthiashanel, line 379: once we find one subject, break and skip the others?

matthiashanel (commented):
I have a few questions/suggestions

server/events.go
```diff
@@ -801,6 +801,14 @@ func (s *Server) initEventTracking() {
 				if acc, err := extractAccount(subject); err != nil {
 					return nil, err
 				} else {
+					if ci, _, _, _, err := c.srv.getRequestInfo(c, msg); err == nil {
+						// Make sure the accounts match.
+						if ci.Account != acc {
+							// Do not leak too much here.
+							return nil, fmt.Errorf("bad request")
+						}
```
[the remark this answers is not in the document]
derekcollison: extractAccount pulls from the subject space, this is checking if the request came through a NATS normal account through the service import.

server/events.go
derekcollison: This is what is mapped into a normal account that they send the request to. This gets changed into correct account scoped subject before being invoked at the system account layer.

server/events.go
derekcollison: See above. But let me know if there are additional questions.

server/monitor.go
derekcollison: True, will update.

matthiashanel (approved):
I can live with the PR as is. LGTM, just a naming suggestion and perhaps move code?

About that name: "$SYS.REQ.ACCOUNT.PING.CONNZ"
We also have serverPingReqSubj = "$SYS.REQ.SERVER.PING.%s".
There, `PING` means every server responds. 
`PING` here does not mean every account responds, only the own account responds.
Would it make sense to change this to: "$SYS.REQ.ACCOUNT.SELF.CONNZ" and make it clearer that you are talking to the account the request originated in?

derekcollison:
The PING is consistent above in that it signals you may get multiple responses since every applicable server will respond. Was trying to mimic the server version.

PR #2439, opened 2021-08-14
Fixed a leak of internal clients for JetStream consumers.

derekcollison:
We were not properly unregistering from the account on cleanup.

/cc @nats-io/core


kozlovic (approved):
LGTM

PR #2442, opened 2021-08-15
[FIXED] LeafNode with "wss://.." url was not always initiating TLS

kozlovic:
If the remote did not have any TLS configuration, the URL scheme
"wss://" was not used as the indicating that the connection should
be attempted as a TLS connection, causing "invalid websocket connection"
in the server attempting to create the remote leafnode connection.


derekcollison (approved):
LGTM

ripienaar (approved):
LGTM

PR #2449, opened 2021-08-17
[IMPROVED] Make error actionable when adding operator+leafnodes

alsuren:
There are many examples in the documentation for one half of this configuration or the other,
but none which configure a leafnode remote on an operator-authenticated cluster.

The error "operator mode requires account nkeys in remotes." is not very clear or actionable.
It was only chance googling that led us to https://github.com/nats-io/k8s/pull/286 and gave us a hint about what we needed to do.

/cc @nats-io/core

### DIFF
--- a/server/leafnode.go
+++ b/server/leafnode.go
@@ -276,7 +276,10 @@ func validateLeafNode(o *Options) error {
 		}
 		for _, r := range o.LeafNode.Remotes {
 			if !nkeys.IsValidPublicAccountKey(r.LocalAccount) {
-				return fmt.Errorf("operator mode requires account nkeys in remotes")
+				return fmt.Errorf(
+					"operator mode requires account nkeys in remotes. " +
+						"Please add an `account` key to each remote in your `leafnodes` section, to assign it to an account. " +
+						"Each account value should be a 56 character public key, starting with the letter 'A'")
 			}
 		}
 		if o.LeafNode.Port != 0 && o.LeafNode.Account != "" && !nkeys.IsValidPublicAccountKey(o.LeafNode.Account) {
--- a/server/leafnode_test.go
+++ b/server/leafnode_test.go
@@ -2483,7 +2483,9 @@ func TestLeafNodeOperatorBadCfg(t *testing.T) {
 			authorization {
 				account: notankey
 			}`,
-		`operator mode requires account nkeys in remotes`: `remotes: [{url: u}]`,
+		("operator mode requires account nkeys in remotes. " +
+			"Please add an `account` key to each remote in your `leafnodes` section, to assign it to an account. " +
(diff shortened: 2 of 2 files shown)


kozlovic (changes requested):
Reporting what staticcheck says:
```
server/leafnode.go:279:12: error strings should not end with punctuation or a newline (ST1005)
```
But more importantly, I wonder if instead of doing that, we should not make use of what @ripienaar has been introducing with errors. Should we start (in cases like that) have codes for non JetStream errors and keep adding to it when users report less than descriptive errors?
If so, R.I., would you mind showing us what it would take here to add a new error, etc..?

ripienaar:
You're right this is a good case where if we had those errors outside of JetStream would have been useful, but we are not really ready for that yet - the current work is very JS orientated and I think we're still gathering feedback on how they are to use before we consider moving them into the more senstive parts of things.

kozlovic:
@alsuren Ok, so based on RI response, I would suggest that you remove the terminal punctuation, otherwise staticcheck would reject this code change.

alsuren:
Hi. Sorry for the delay. Pushed.

```
=== RUN   TestJetStreamMaxMsgsPerAndDiscardNew/Memory
    jetstream_test.go:11193: Unexpected error: context deadline exceeded
--- FAIL: TestJetStreamMaxMsgsPerAndDiscardNew (11.98s)
    --- PASS: TestJetStreamMaxMsgsPerAndDiscardNew/File (0.68s)
    --- FAIL: TestJetStreamMaxMsgsPerAndDiscardNew/Memory (11.30s)
```
This feels unrelated, and passes locally. Is this test known to be flakey?

kozlovic:
That one yes, but maybe not this one. Please check if this test is not checking for the error that you have now changed.
```
=== RUN   TestLeafNodeOperatorBadCfg/operator_mode_requires_account_nkeys_in_remotes
    leafnode_test.go:2512: Expected error operator mode requires account nkeys in remotes but got operator mode requires account nkeys in remotes. Please add an `account` key to each remote in your `leafnodes` section, to assign it to an account. Each account value should be a 56 character public key, starting with the letter 'A'
--- FAIL: TestLeafNodeOperatorBadCfg (0.00s)
    --- PASS: TestLeafNodeOperatorBadCfg/operator_mode_does_not_allow_specifying_user_in_leafnode_config (0.00s)
    --- PASS: TestLeafNodeOperatorBadCfg/operator_mode_and_non_account_nkeys_are_incompatible (0.00s)
    --- FAIL: TestLeafNodeOperatorBadCfg/operator_mode_requires_account_nkeys_in_remotes (0.00s)
```

kozlovic (approved):
LGTM

PR #2450, opened 2021-08-17
Make sure jwt claim update does not wipe system imports

derekcollison:

/cc @nats-io/core


ripienaar (approved):
LGTM

kozlovic (approved):
LGTM (hopefully we are done with replacing "" with _EMPTY_ ;-) )

PR #2451, opened 2021-08-17
Added test for system account update as well

derekcollison:
Updating the system account could wipe the account connz export.

/cc @nats-io/core


ripienaar (approved):
LGTM

PR #1480, opened 2020-06-16
Properly handle leafnode spoke permissions.

derekcollison:
When a leafnode would connect with credentials that had permissions the spoke did not have a way of knowing what those were. This could lead to being disconnected when sending subscriptions or messages to the hub which were not allowed.

Signed-off-by: Derek Collison <derek@nats.io>

/cc @nats-io/core

PR #2455, opened 2021-08-19
[FIXED] LeafNode: wrong permission check prevented message flow

### LINKED ISSUE #2454: Leaf Node credentials require publish permissions to subscribe
## Defect

Using Credentials to control access for Leaf Nodes has changed between 2.1.9 and 2.2 and above. It appears it is now necessary to provide the Leaf Node credentials with permission to Publish and Subscribe in order to Subscribe to a subject. This presents an issue for us because we don't want to distribute privileged credentials to the Leaf Nodes.

nats-server version 2.2.0 and above (verified up to 2.3.4)

Ubuntu 20.04

See attached repro.txt for sample script to reproduce. [repro.txt](https://github.com/nats-io/nats-server/files/7014969/repro.txt)

Server with a Leaf Node both in Operator mode using the same Operator and Account.
Leaf Node uses the LEAF credentials in its leafnodes->remotes section.

User SRV with no restrictions.
User LEAF with permissions to Publish only on "foo" and Subscribe only on "bar"

User SRV performs a nats-sub on subject "bar" on the Server Node (as a baseline)
User LEAF performs a nats-sub on subject "bar" on the Leaf Node (demonstrates missing message)

User SRV performs a nats-pub on subject "bar" on the Server Node.

#### Expected result:

Both Server Node and Leaf Node should receive the "bar" message.

#### Actual result:

Only Server Node receives the "bar" message.

Log message in Server.log indication the Leaf Node requires Publish permission
`[DBG] 127.0.0.1:54526 - lid:1 - Not permitted to publish to "bar"`

Also confirmed giving the LEAF credentials Publish permission on "bar" allows it to receive the message. This is not ideal though as it would mean credentials exist on the Leaf Node allowing Publish to subjects it shouldn't be able to.

I found with these changes I had the results I expected but I'm not sure of the larger ramifications of this:
```
index 9fc10586..b43d6ce4 100644
--- a/server/client.go
+++ b/server/client.go
@@ -3056,9 +3056,9 @@ func (c *client) deliverMsg(sub *subscription, acc *Account, subject, reply, mh,

        // Check if we are a leafnode and have perms to check.
        if client.kind == LEAF && client.perms != nil {
-               if !client.pubAllowedFullCheck(string(subject), true, true) {
+               if !client.canSubscribe(string(subject)) {
                        client.mu.Unlock()
-                       client.Debugf("Not permitted to publish to %q", subject)
+                       client.Debugf("Not permitted to subscribe to %q", subject)
                        return false
                }
        }

```

kozlovic:
The PR #1480 was added in v2.2.0, so that matches user report that suggest that things changed compared to pre-2.2.0 releases.
The check for "pubAllowed" when client is a LEAF was added in that PR. Only the PUB permissions were checked regardless of the type of leaf connection (hub or spoke). In PR 1480, the check was done only for spoke but then was changed to perform pub check as long as client is LEAF: https://github.com/nats-io/nats-server/pull/1480/commits/3729552a37ba6141b8b5fba7d9fc3c1b1d654f2a

I think that we need to distinguish if the connection is hub, check subscribe permissions, if spoke, pub permissions. I have added a test that reproduces the user report `TestLeafNodeOperatorAndPermissions` that seem to be fixed with the proposed change.

Resolves #2454

### DIFF
--- a/server/client.go
+++ b/server/client.go
@@ -3056,9 +3056,23 @@ func (c *client) deliverMsg(sub *subscription, acc *Account, subject, reply, mh,
 
 	// Check if we are a leafnode and have perms to check.
 	if client.kind == LEAF && client.perms != nil {
-		if !client.pubAllowedFullCheck(string(subject), true, true) {
+		if client.isSpokeLeafNode() {
+			// `client` connection is considered a spoke, that is, it is a
+			// "remote" to the other server. We check if it is allowed to
+			// publish.
+			if !client.pubAllowedFullCheck(string(subject), true, true) {
+				client.mu.Unlock()
+				client.Debugf("Not permitted to publish to %q", subject)
+				return false
+			}
+		} else if !client.canSubscribe(string(subject)) {
+			// `client` connection is considered the hub, that is, it accepted
+			// the connection from the server that created a "remote" connection
+			// to this server. Here, we want to check if the other side can
+			// receive this message, so is it allowed to subscribe to this subject.
+
 			client.mu.Unlock()
-			client.Debugf("Not permitted to publish to %q", subject)
+			client.Debugf("Not permitted to subscribe to %q", subject)
 			return false
 		}
 	}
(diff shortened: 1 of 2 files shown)


matthiashanel (approved):
LGTM

derekcollison (approved):
LGTM

PR #2456, opened 2021-08-20
Improve file utilization when using a JetStream stream as a KV.

derekcollison:

/cc @nats-io/core

### DIFF
--- a/server/filestore.go
+++ b/server/filestore.go
@@ -1831,12 +1831,11 @@ func (fs *fileStore) removeMsg(seq uint64, secure, needFSLock bool) (bool, error
 				mb.dmap = make(map[uint64]struct{})
 			}
 			mb.dmap[seq] = struct{}{}
+			// Check if <50% utilization and minimum size met.
+			if mb.rbytes > compactMinimum && mb.rbytes>>1 > mb.bytes {
+				mb.compact()
+			}
 		}
-	}
-
-	// Check if <50% utilization and minimum size met.
-	if mb.rbytes > compactMinimum && mb.rbytes>>1 > mb.bytes {
-		mb.compact()
 	}
 
 	var qch, fch chan struct{}
--- a/server/filestore.go
+++ b/server/filestore.go
@@ -178,6 +178,8 @@ const (
 	purgeDir = "__msgs__"
 	// used to scan blk file names.
 	blkScan = "%d.blk"
+	// used for compacted blocks that are staged.
+	newScan = "%d.new"
 	// used to scan index file names.
 	indexScan = "%d.idx"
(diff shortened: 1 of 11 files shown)


server/filestore.go
```diff
@@ -216,8 +218,12 @@ const (
 	defaultStreamBlockSize = 16 * 1024 * 1024 // 16MB
 	// Default for workqueue or interest based.
 	defaultOtherBlockSize = 8 * 1024 * 1024 // 8MB
+	// Default for KV based
+	defaultKVBlockSize = 8 * 1024 * 1024 // 8MB
```
matthiashanel, line 222: Should this be used somewhere?

server/filestore.go
```diff
@@ -791,7 +800,6 @@ func (mb *msgBlock) rebuildState() (*LostStreamData, error) {
 
 			mb.msgs++
 			mb.bytes += uint64(rl)
-			mb.rbytes += uint64(rl)
```
matthiashanel, line 794: This was essentially done by line 670?
mb.rbytes = uint64(len(buf))

server/filestore.go
derekcollison: It is in stream.go

server/filestore.go
derekcollison: Yes that was a bug, could not figure out why my compaction and rebuild state was saying same for rbytes ;)

matthiashanel (approved):
LGTM

server/filestore.go
```diff
@@ -1870,6 +1889,93 @@ func (fs *fileStore) removeMsg(seq uint64, secure, needFSLock bool) (bool, error
...
+
+	// We will write to a new file and mv/rename it in case of failure.
+	mfn := path.Join(path.Join(mb.fs.fcfg.StoreDir, msgDir), fmt.Sprintf(newScan, mb.index))
+	defer os.Remove(mfn)
+	if err := ioutil.WriteFile(mfn, nbuf, defaultFilePerms); err != nil {
+		return
+	}
+	os.Rename(mfn, mb.mfn)
+
+	// Close cache and open FDs and index file.
```
kozlovic, line 1966: Should not closing be done before the rename? For instance, I don't think if this would work on Windows.

kozlovic (approved):
LGTM, but wonder if we should not be closing fds before the rewrite/rename, with possibly reopening in failed to rename, etc..

server/filestore.go
derekcollison: I will move it up.

ripienaar (approved):
LGTM

ripienaar (approved):
LGTM

PR #2461, opened 2021-08-23
[added] operator option to ensure user are signed by certain accounts

matthiashanel:
option name: resolver_pinned_accounts
Contains a list of public account nkeys.
Connecting user of leaf nodes need to be signed by this.
The system account will always be able to connect.

### DIFF
--- a/server/auth.go
+++ b/server/auth.go
@@ -25,6 +25,7 @@ import (
 	"net/url"
 	"regexp"
 	"strings"
+	"sync/atomic"
 	"time"
 
 	"github.com/nats-io/jwt/v2"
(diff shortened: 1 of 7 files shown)


server/auth.go
```diff
@@ -584,9 +586,15 @@ func (s *Server) processClientOrLeafAuthentication(c *client, opts *Options) boo
 			return false
 		}
 		issuer := juc.Issuer
-		if juc.IssuerAccount != "" {
+		if juc.IssuerAccount != _EMPTY_ {
 			issuer = juc.IssuerAccount
 		}
+		if pinnedAcounts != nil {
+			if _, ok := pinnedAcounts[issuer]; !ok {
+				c.Debugf("Account not listed as operator pinned account")
```
philpennock, line 594: Do we have varz metrics we can bump up here?  I'd like it if "tried to authenticate as an unpinned account" bumped up the counter.

Also, what is our threat-model around enumeration of "which accounts are allowed to connect to this server", given public operators with public account lists?  Should we move the rejection to _after_ otherwise-successful authentication, to avoid timing-attacks based on differences in timing?

philpennock (approved):
Approving because I think this works, but the comments on the first point are something I'd like to be at least considered prior to merging.

derekcollison (approved):
LGTM

server/auth.go
matthiashanel: added the counter

philpennock (approved):

PR #2462, opened 2021-08-23
[CHANGED] ConsumerInfo's SequencePair replaced with SequenceInfo

kozlovic:
This change was made in a previous PR wit this commit:
https://github.com/nats-io/nats-server/pull/2438/commits/9405b77e464dccf3be051b8c750a2354a088c5df

After some discussions, we agreed that the original approach
is best, so using a dedicated object SequenceInfo for ConsumerInfo.

### DIFF
--- a/server/consumer.go
+++ b/server/consumer.go
@@ -37,8 +37,8 @@ type ConsumerInfo struct {
 	Name           string          `json:"name"`
 	Created        time.Time       `json:"created"`
 	Config         *ConsumerConfig `json:"config,omitempty"`
-	Delivered      SequencePair    `json:"delivered"`
-	AckFloor       SequencePair    `json:"ack_floor"`
+	Delivered      SequenceInfo    `json:"delivered"`
+	AckFloor       SequenceInfo    `json:"ack_floor"`
 	NumAckPending  int             `json:"num_ack_pending"`
 	NumRedelivered int             `json:"num_redelivered"`
 	NumWaiting     int             `json:"num_waiting"`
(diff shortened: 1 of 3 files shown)


derekcollison (approved):
LGTM

ripienaar:
@kozlovic do we need a general client update issue for this? Seems to me SequencePair is now nowhere in the public api right?

kozlovic:
@ripienaar I believe that you had mentioned the addition of the "Last" field in https://github.com/nats-io/nats-architecture-and-design/issues/43. Maybe it needs to be updated to mention that the server has ConsumerInfo's Delivered and AckFloor field a SequenceInfo instead of SequencePair, but really the JSON tags of Delivered and AckFloor remain the same, so it does not matter for the clients.

In some clients, the "SequencePair" object is used in message metadata, but agreed that at this point I *think* that the server is not sending anything JS related with a sequence pair (in the strict term, stream/consumer sequence) to the client. It still uses SequencePair for the consumer state that is stored on disk.

ripienaar:
OK, well I changed the cli side i find it much easier to remain in sync if we use the same type names etc - but as clients wish so I guess its fine

kozlovic:
> OK, well I changed the cli side i find it much easier to remain in sync if we use the same type names etc - but as clients wish so I guess its fine

I changed it too in the Go and C client, but I think you were asking about SequencePair. I mentioned that some library use such object to represent a stream and consumer sequence, as in message metadata object, but agreed that at the moment the objects returned by the server (such as consumer info, etc..) do not have any SequencePair object.

PR #2463, opened 2021-08-23
Add in optional compression schemes for Accept-Encoding on server API.

derekcollison:

/cc @nats-io/core

### DIFF
--- a/server/events.go
+++ b/server/events.go
@@ -15,6 +15,7 @@ package server
 
 import (
 	"bytes"
+	"compress/gzip"
 	"encoding/json"
 	"errors"
 	"fmt"
@@ -26,6 +27,7 @@ import (
 	"sync/atomic"
 	"time"
 
+	"github.com/klauspost/compress/s2"
 	"github.com/nats-io/jwt/v2"
 	"github.com/nats-io/nats-server/v2/server/pse"
 )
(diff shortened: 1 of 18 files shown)


server/events.go
```diff
@@ -331,16 +334,44 @@ RESET:
 			c.mu.Lock()
 
 			// Prep internal structures needed to send message.
-			c.pa.subject = []byte(pm.sub)
-			c.pa.size = len(b)
-			c.pa.szb = []byte(strconv.FormatInt(int64(len(b)), 10))
-			c.pa.reply = []byte(pm.rply)
+			c.pa.subject, c.pa.reply = []byte(pm.sub), []byte(pm.rply)
+			c.pa.size, c.pa.szb = len(b), []byte(strconv.FormatInt(int64(len(b)), 10))
```
kozlovic, line 338: No big deal, but looking down, it seems that if there is compression, c.setHeader() will set some of the c.pa fields based on the new header content, etc.. So this line here will be undone if there is compression and header added. Not asking to change, but making sure I understand this.

kozlovic (approved):
LGTM, just have a question about the send loop and building of msg/hdr size.

server/events.go
derekcollison: Yes that is correct. It needs a properly formed message with /r/n and any existing header info in the c.pa before we call into setHeader, hence the ordering.

ripienaar (commented):
LGTM

PR #2467, opened 2021-08-24
Do not use crypto rand for nonce generation.

derekcollison:
Crypto rand is not needed for nonce generation and could drain entropy.

/cc @nats-io/core

### DIFF
--- a/server/filestore.go
+++ b/server/filestore.go
@@ -35,6 +35,8 @@ import (
 	"sync/atomic"
 	"time"
 
+	mrand "math/rand"
+
 	"github.com/klauspost/compress/s2"
 	"github.com/minio/highwayhash"
 	"golang.org/x/crypto/chacha20"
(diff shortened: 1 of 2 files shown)


ripienaar:
more details on slack but calls to:

```
        rbek.XORKeyStream(buf, buf)
``` 

takes around 40ms each

The code here, which includes a call to XorKeyStream(), takes about 60 to 70ms

https://github.com/nats-io/nats-server/blob/be17afb0a50463780f542f27bbf8b928d718bb1c/server/filestore.go#L538-L560

This has knock on effects that deleteFirstMsg() takes about 90ms per message, this is called for every message that needs to be expired - in my test I ensure I have 500 message that needs expiring to trigger these timings.

derekcollison:
ok the expiration I guess was the missing piece to get to see the results you were seeing.

PR #2479, opened 2021-08-30
[FIXED] JetStream: possible deadlock due to lock inversion

kozlovic:
The locking is jetStream->Server, not the otherway around. There
was few places where lock inversion could have caused deadlock.

Also, a change made recently to solve a deadlock was causing
a race that is demonstrated with TestJetStreamRaceOnRAFTCreate.

### DIFF
--- a/server/monitor.go
+++ b/server/monitor.go
@@ -1262,6 +1262,8 @@ func (s *Server) HandleRoot(w http.ResponseWriter, r *http.Request) {
 func (s *Server) updateJszVarz(js *jetStream, v *JetStreamVarz, doConfig bool) {
 	if doConfig {
 		js.mu.RLock()
+		// We want to snapshot the config since it will then be available outside
+		// of the js lock. So make a copy first, then point to this copy.
 		cfg := js.config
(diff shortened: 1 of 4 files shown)


server/monitor.go
```diff
@@ -1259,6 +1259,19 @@ func (s *Server) HandleRoot(w http.ResponseWriter, r *http.Request) {
 	)
 }
 
+func (s *Server) updateJszVarz(js *jetStream, v *JetStreamVarz, doConfig bool) {
+	if doConfig {
+		js.mu.RLock()
+		cfg := js.config
```
derekcollison, line 1265: Maybe mention here that we do it this way to make a copy.

derekcollison (approved):
LGTM

server/monitor.go
kozlovic: Done in this new commit: https://github.com/nats-io/nats-server/pull/2479/commits/80ebf2d7b253730ffece28018705c3e0b4247bb2

derekcollison (approved):
LGTM

Issue #2397, opened 2021-08-03
jetstream could not pull message after nats-server restart


carr123:
i was testing jetstream on nats-server v2.3.2.  one sender and one receiver program are running  for quite a long time.

this is what my stream look like :

```go
	_, err = js.AddStream(&nats.StreamConfig{
		Name:      streamName,
		Subjects:  []string{streamSubjects},
		Storage:   nats.FileStorage,
		Replicas:  3,
		Retention: nats.WorkQueuePolicy,
		Discard:   nats.DiscardNew,
		MaxMsgs:   -1,
		MaxAge:    time.Hour * 24 * 365,
	})
```

this is how i create the consumer:

```go
	if _, err := js.AddConsumer(streamName, &nats.ConsumerConfig{
		Durable:       durableName,
		DeliverPolicy: nats.DeliverAllPolicy,
		AckPolicy:     nats.AckExplicitPolicy,
		ReplayPolicy:  nats.ReplayInstantPolicy,
		FilterSubject: subjectName,
		AckWait:       time.Second * 30,
		MaxDeliver:    -1,
		MaxAckPending: 1000,
	}); err != nil && !strings.Contains(err.Error(), "already in use") {
		log.Println("AddConsumer fail")
		return
	}
```

this is what the subscriber look like:

```go
	sub, err := js.PullSubscribe("ORDERS.created", durableName, nats.Bind("ORDERS", durableName))
	if err != nil {
		fmt.Println(" PullSubscribe:", err)
		return
	}
       msgs, err := sub.Fetch(1000, nats.MaxWait(10*time.Second))
```      

when i restart my nats-server cluster nodes(upgrade to nats-server 2.3.3), 
the consumer can no longer pull messages even if i restart my consumer program. 
the Fetch call just return : "nats: timeout", but i'm sure there are lots of message in the working queue.
only if i delete the consumer by calling js.DeleteConsumer(streamName, durableName),  recreate it, my program can 
resume fetching messages.
actually, every time i restart nats-server nodes, my consumer program encouter the same problem.

there is another issue, after i restart nats-server nodes,  restart my program, it sometimes reports : "PullSubscribe: nats: JetStream system temporarily unavailable"

I expect nats-server nodes restarting action not impacting jetstream clients fetching messages.

carr123:
reproduce this issue:
1. start a client program fetching messages in a loop
2. make sure there are NO messages in the jetsteam
2. restart each nats-server cluster node one by one.
[...]

PR #2480, opened 2021-09-01
[FIXED] Issue #2397

derekcollison:
When we had partial state due to server failure or being shutdown ungracefully we could enter into a stream reset state.
The stream reset state is harsh but worked, however there was a bug that would not restart consumers that were attached.
Also if no state exists, or state was truncated, we can detect that and not go through a full reset.

/cc @nats-io/core

### DIFF
--- a/server/consumer.go
+++ b/server/consumer.go
@@ -271,6 +271,26 @@ const (
 	JsDefaultMaxAckPending = 20_000
 )
 
+// Helper function to set consumer config defaults from above.
+func setConsumerConfigDefaults(config *ConsumerConfig) {
+	// Set to default if not specified.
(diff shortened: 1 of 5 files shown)


ripienaar (approved):
LGTM, though bunch of code that I dont know

kozlovic (approved):
LGTM, but new test introduced in previous PR is finding data races that will need to be addressed, otherwise travis will now fail a lot again.

derekcollison:
I added a fix for the race condition.

derekcollison:
I will redo the race fix in a bit after another mtg.

wallyqs (approved):
LGTM

kozlovic (approved):
LGTM

PR #2482, opened 2021-09-01
[FIXED] Pull requests: don't send 408 when request expires

kozlovic:
When expiring requests, the server would send 408 if interest was
still present, which can happen for pull subscribe implementations
that maintain interest for the duration of the pull subscription.

Let's keep the 408 for when a request is "force expired", that
is, a request was removed from the queue because it queue was
full but interest is still found.

### DIFF
--- a/server/jetstream_test.go
+++ b/server/jetstream_test.go
@@ -1843,24 +1843,24 @@ func TestJetStreamWorkQueueRequest(t *testing.T) {
 			// Now do expiration
 			req.Batch = 1
 			req.NoWait = false
-			req.Expires = 10 * time.Millisecond
+			req.Expires = 100 * time.Millisecond
 			jreq, _ = json.Marshal(req)
 
 			nc.PublishRequest(getSubj, reply, jreq)
 			// Let it expire
-			time.Sleep(20 * time.Millisecond)
+			time.Sleep(200 * time.Millisecond)
 
 			// Send a few more messages. These should not be delivered to the sub.
 			sendStreamMsg(t, nc, "foo", "Hello World!")
 			sendStreamMsg(t, nc, "bar", "Hello World!")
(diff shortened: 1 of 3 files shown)


server/jetstream_test.go
```diff
@@ -1853,13 +1853,15 @@ func TestJetStreamWorkQueueRequest(t *testing.T) {
 			// Send a few more messages. These should not be delivered to the sub.
 			sendStreamMsg(t, nc, "foo", "Hello World!")
 			sendStreamMsg(t, nc, "bar", "Hello World!")
-			// We will have an alert here.
+			time.Sleep(10 * time.Millisecond)
```
derekcollison, line 1856: Would make at least 100 these days.

derekcollison (approved):
LGTM - Might bump one sleep.

server/jetstream_test.go
kozlovic: We sleep above past the expiration time. This sleep was to make sure that we would not start delivering, but I hear you. Will increase.

kozlovic: Increased the expiration timeout anyway (but still "small") and bumped server version.

derekcollison (approved):
LGTM

ripienaar:
Very glad to see this removed LGTM

kozlovic:
@ripienaar It is still sent (as of now) if the request is "force expired", that is, the server could not remove (due to expiration) a single request and therefore evicts the first from the queue. In that case, if there is interest, server will still send 408, which I would argue is wrong "status/description", since in this case it is not a timeout but an eviction or some other error, but that would definitively break clients if we introduce a new code there.

ripienaar:
Yes, my moan about them was the ones that came at some random future time and not at the point of expiration timer expiring (the lazy ones).  So happy to see those specific ones go away.

PR #2484, opened 2021-09-02
Set defaultServerOptions port to -1 for random

kozlovic:
Updated some tests based on this change but also missing defer
connection close or server shutdown.

Fixed how the OCSP run go routine would shutdown, which would
never complete because grWG was not decremented by this go routine
prior to invoking s.Shutdown()


server/ocsp.go
```diff
@@ -233,7 +233,15 @@ func (oc *OCSPMonitor) run() {
 	quitCh := s.quitCh
 	s.mu.Unlock()
 
-	defer s.grWG.Done()
+	var doShutdown bool
```
kozlovic, line 236: @wallyqs @variadico Could you please review this part: what was happening is that when run() decides to call s.Shutdown(), this go routine would be blocked because Shutdown() would be waiting on grWG.Wait() that would never be done because this go routine itself had not yet called s.grWG.Done().
Can you think of if this would have been user-visible? And if so, what would be the symptoms? If not user-visible, this is just a fix that will help make sure that there are not left-over go routines when a test completes.

wallyqs: Thanks @kozlovic for the fix, this would have happened then when a user sets the ocsp policy to `always`, which means that the server should shutdown when a cert is revoked.  This ought to have been covered by this test: https://github.com/nats-io/nats-server/blob/7ec4ce986a9c720a3bb81062b7b84f9376c3ac14/test/ocsp_test.go#L118-L132

wallyqs (approved):
LGTM

server/ocsp.go
kozlovic: Yes, so server was shutdown. Looking at where we wait in Shutdown(), the only thing that would not be executed was deleting the ports file and closing the logger (for closable ones).

variadico: Ah, maybe we should be checking `srv.WaitForShutdown()` to make sure the last line of `Shutdown` is actually called?
```go
// Notify that the shutdown is complete
close(s.shutdownComplete)
```

kozlovic: Excellent idea, and I did add that here: https://github.com/nats-io/nats-server/pull/2484/commits/674299a184851b308b8010b676f5c12166c22ffa

I have verified that without the fix, this call would have blocked.

variadico: Woot! Awesome!

variadico (approved):

derekcollison (approved):
LGTM

Issue #2488, opened 2021-09-04
Consumers stops receiving messages


anjmao:
#### Versions of `nats-server` and affected client libraries used:

Nats server version
```
[83] 2021/09/04 18:51:12.239432 [INF] Starting nats-server
[83] 2021/09/04 18:51:12.239488 [INF]   Version:  2.4.0
[83] 2021/09/04 18:51:12.239494 [INF]   Git:      [219a7c98]
[83] 2021/09/04 18:51:12.239496 [DBG]   Go build: go1.16.7
[83] 2021/09/04 18:51:12.239517 [INF]   Name:     NBVE7O7DMRAZ63STC7Z644KHF5HJ6QQUGLZVGDIKEG32CFL2J6O2456M
[83] 2021/09/04 18:51:12.239533 [INF]   ID:       NBVE7O7DMRAZ63STC7Z644KHF5HJ6QQUGLZVGDIKEG32CFL2J6O2456M
[83] 2021/09/04 18:51:12.239605 [DBG] Created system account: "$SYS"
```

Go client version: `v1.12.0`

#### OS/Container environment:

GKE Kubernetes. Running nats js HA cluster. Deployed via nats helm chart.

#### Steps or code to reproduce the issue:

Stream configuration:
```yaml
apiVersion: jetstream.nats.io/v1beta1
kind: Stream
metadata:
  name: agent
spec:
  name: agent
  subjects: ["data.*"]
  storage: file
  maxAge: 1h
  replicas: 3
  retention: interest
```

There are two consumers to this stream. Each runs as queue subscriber in two services with 2 pod replicas. Note that I don't care if message is not processed, this is why ack none is set.
```go

// 2 pods for service A.
js.QueueSubscribe(
	"data.received",
	"service1_queue",
	func(msg *nats.Msg) {},
	nats.DeliverNew(),
	nats.AckNone(),
)

// 2 pods for service B.
s.js.QueueSubscribe(
	"data.received",
	"service2_queue",
	func(msg *nats.Msg) {},
	nats.DeliverNew(),
	nats.AckNone(),
)
```

#### Expected result:

Consumer receives messages.

#### Actual result:

Stream stats after few days:

```
agent                  │ File    │ 3         │ 28,258   │ 18 MiB  │ 0    │ 84      │ nats-js-0, nats-js-1*, nats-js-2
```

Consumers stats:

```
[...]

PR #2490, opened 2021-09-05
Fixed a bug that could lead to perceived message loss under JetStream.

derekcollison:
Under load and pressure from concurrent publishing and consuming with multiple consumers the filestore would return a partial or no cache error to the upper layers. For consumers this could result in us skipping a stream sequence when we should not.

 This change stabilizes the filestore and removes the flush state for msg blocks. I also found some bugs that did not track last sequence properly after snapshots / restore.

Also increased timeout under Travis for the Go tests. We were bumping up against the 10m default.

/cc @nats-io/core

### DIFF
--- a/server/filestore.go
+++ b/server/filestore.go
@@ -3094,13 +3094,7 @@ func (mb *msgBlock) fetchMsg(seq uint64) (*fileStoredMsg, error) {
 	if err := mb.loadMsgsWithLock(); err != nil {
 		return nil, err
 	}
-
-	sm, err = mb.cacheLookup(seq)
-	if err == errNoCache {
-		fmt.Printf("NO CACHE [%d] FOR %d!!\n\n", mb.index, seq)
-		fmt.Printf("first %d, last %d\n", mb.first.seq, mb.last.seq)
-	}
-	return sm, err
+	return mb.cacheLookup(seq)
 }
 
 var (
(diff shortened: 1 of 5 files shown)


anjmao:
Hi @derekcollison 

Looks like this bug https://github.com/nats-io/nats-server/issues/2488  is related to your fixes.

derekcollison:
Unclear if they are related tbh.

tigrato:
> Unclear if they are related tbh.

Not sure if it helps but my JetStream was having the same issues described in #2488 and once I switched from file storage to memory storage it improved a lot and no issues until now.

server/filestore.go
```diff
@@ -3126,17 +3082,25 @@ checkCache:
...
 	// We have a cache miss here.
-	if err := mb.loadMsgs(); err != nil {
+	if err := mb.loadMsgsWithLock(); err != nil {
 		return nil, err
 	}
 
-	return mb.cacheLookup(seq)
+	sm, err = mb.cacheLookup(seq)
+	if err == errNoCache {
+		fmt.Printf("NO CACHE [%d] FOR %d!!\n\n", mb.index, seq)
```
kozlovic, line 3100: Remove this print, or if needed for debug, make it a log/debug statement.

kozlovic (changes requested):
LGTM, but I believe that there are left over debug print statements.

server/filestore.go
derekcollison: Thanks fixed.

kozlovic (approved):
LGTM

PR #2493, opened 2021-09-07
During peer removal, try to remap any stream or consumer assets.

derekcollison:
Also if we do not have room trap add peer and process there.
Also fixed a bug that would treat ephemerals the same as durables during remapping after peer removal.

/cc @nats-io/core

### DIFF
--- a/server/jetstream_cluster.go
+++ b/server/jetstream_cluster.go
@@ -1012,10 +1012,60 @@ func (sa *streamAssignment) copyGroup() *streamAssignment {
 	return &csa
 }
 
+// Lock should be held.
+func (sa *streamAssignment) missingPeers() bool {
+	return len(sa.Group.Peers) < sa.Config.Replicas
+}
+
+// Called when we detect a new peer. Only the leader will process checking
+// for any streams, and consequently any consumers.
+func (js *jetStream) processAddPeer(peer string) {
+	js.mu.Lock()
+	defer js.mu.Unlock()
+
+	s, cc := js.srv, js.cluster
+	isLeader := cc.isLeader()
+
+	// Now check if we are meta-leader. We will check for any re-assignments.
+	if !isLeader {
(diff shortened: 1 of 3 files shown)


server/jetstream_cluster.go
```diff
@@ -1012,10 +1012,60 @@ func (sa *streamAssignment) copyGroup() *streamAssignment {
...
+			if sa.missingPeers() {
+				// Make sure the right cluster etc.
+				if si.cluster != sa.Client.Cluster {
+					continue
+				}
+				// If we are here we can add in this peer.
+				csa := sa.copyGroup()
+				csa.Group.Peers = append(csa.Group.Peers, peer)
+				// Send our proposal for this csa. Also use same group definition for all the consumers as well.
+				cc.meta.Propose(encodeAddStreamAssignment(csa))
```
ripienaar, line 1051: I dont know the flows of this code much - but this Propose can block, and with enough streams this can take a very long time especially if the propose channel fills, is that ok? We could even lose leadership during this?

Mainly just curious how the code work, sure its fine

derekcollison: Yes all of those things can happen. The way we mitigate is to only stay in place for a client, and run outside of main route loops and GWs.

This code can block, but is in its own Go routine already processing the raft layer.

ripienaar (approved):
LGTM

server/jetstream_cluster.go
kozlovic: Just want to point out that we have the js lock at the top of the function, so by blocking I figure you mean waiting for some other nodes to do something. If it is waiting on something in the same process, then we may have an issue since we are holding the js lock.

derekcollison: Only thing we wait on would be underlying raft layer to drain propc chan.

PR #2499, opened 2021-09-08
[FIXED] - Message cleanup for interest stream and no-ack consumers in clustered mode.

### LINKED ISSUE #2488: Consumers stops receiving messages

derekcollison:
Fixed a bug when an interest retention stream with noack consumers is in clustered mode.
We were not properly propagating the ack state and proper cleanup of the stream messages.

Resolves #2488 (Related)

/cc @nats-io/core

### DIFF
--- a/server/consumer.go
+++ b/server/consumer.go
@@ -2340,11 +2340,6 @@ func (o *consumer) deliverMsg(dsubj, subj string, hdr, msg []byte, seq, dc uint6
 	// Send message.
 	o.outq.send(pmsg)
 
-	// If we are ack none and mset is interest only we should make sure stream removes interest.
-	if ap == AckNone && mset.cfg.Retention != LimitsPolicy && mset.amch != nil {
-		mset.amch <- seq
-	}
-
 	if ap == AckExplicit || ap == AckAll {
 		o.trackPending(seq, dseq)
 	} else if ap == AckNone {
@@ -2359,6 +2354,15 @@ func (o *consumer) deliverMsg(dsubj, subj string, hdr, msg []byte, seq, dc uint6
 
 	// FIXME(dlc) - Capture errors?
 	o.updateDelivered(dseq, seq, dc, ts)
+
+	// If we are ack none and mset is interest only we should make sure stream removes interest.
+	if ap == AckNone && mset.cfg.Retention != LimitsPolicy {
+		if o.node == nil || o.cfg.Direct {
+			mset.amch <- seq
+		} else {
+			o.updateAcks(dseq, seq)
+		}
+	}
 }
 
 func (o *consumer) needFlowControl() bool {
--- a/server/jetstream_cluster.go
+++ b/server/jetstream_cluster.go
@@ -1690,6 +1690,7 @@ func (js *jetStream) applyStreamEntries(mset *stream, ce *CommittedEntry, isReco
 					mset.setLastSeq(mset.store.SkipMsg())
 					continue
 				}
+
 				// Process the actual message here.
 				if err := mset.processJetStreamMsg(subject, reply, hdr, msg, lseq, ts); err != nil {
 					if !isRecovering {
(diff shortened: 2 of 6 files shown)


server/consumer.go
```diff
@@ -2359,6 +2354,15 @@ func (o *consumer) deliverMsg(dsubj, subj string, hdr, msg []byte, seq, dc uint6
 
 	// FIXME(dlc) - Capture errors?
 	o.updateDelivered(dseq, seq, dc, ts)
+
+	// If we are ack none and mset is interest only we should make sure stream removes interest.
+	if ap == AckNone && mset.cfg.Retention != LimitsPolicy {
+		if o.node == nil || o.cfg.Direct {
+			mset.amch <- seq
```
kozlovic, line 2361: You used to check for `amch != nil`, not sure if this was necessary or not, but you don't do that anymore.

kozlovic (approved):
LGTM, but small comment about mset.amch

server/consumer.go
derekcollison: In stream.go we always create when cfg.Retention != LimitsPolicy, and since we capture that condition above in line 2359 it has to be present.

kozlovic: As long as it is not set to nil (which I just checked it is not), then indeed it is ok.

PR #2501, opened 2021-09-08
[IMPROVED] MQTT stream per session replaced with single stream

kozlovic:
With the availability of a "max message per subject" for a given
stream, it is possible to replace individual streams that were
created per session with a single stream that gets all sessions
as a single message per subject, which subject is composed of
the session client ID hash.

The first time the new stream is created for a given account,
all existing MQTT session streams will be transferred to the
new mux'ed MQTT session stream.

### DIFF
--- a/server/mqtt.go
+++ b/server/mqtt.go
@@ -922,9 +922,8 @@ func (s *Server) mqttCreateAccountSessionManager(acc *Account, quitCh chan struc
 		},
 	}
 	// We need to include the domain in the subject prefix used to store sessions in the $MQTT_sess stream.
-	// If no domain is present, use "_" so that the token always exists.
 	if d := s.getOpts().JetStreamDomain; d != _EMPTY_ {
-		as.domainTk = string(getHash(d)) + "."
+		as.domainTk = d + "."
 	}
 
 	var subs []*subscription
@@ -1136,7 +1135,7 @@ func (jsa *mqttJSA) newRequestEx(kind, subject string, hdr int, msg []byte, time
 	var sb strings.Builder
 	sb.WriteString(jsa.rplyr)
 	sb.WriteString(kind)
-	sb.WriteByte('.')
+	sb.WriteByte(btsep)
 	sb.WriteString(jsa.nuid.Next())
 	reply := sb.String()
 	jsa.mu.Unlock()
--- a/server/mqtt.go
+++ b/server/mqtt.go
@@ -113,6 +113,10 @@ const (
 	mqttRetainedMsgsStreamName    = mqttStreamNamePrefix + "rmsgs"
 	mqttRetainedMsgsStreamSubject = "$MQTT.rmsgs"
 
+	// Stream name for MQTT sessions on a given account
+	mqttSessStreamName          = mqttStreamNamePrefix + "sess"
+	mqttSessStreamSubjectPrefix = "$MQTT.sess."
+
 	// Stream name prefix for MQTT sessions on a given account
 	mqttSessionsStreamNamePrefix = mqttStreamNamePrefix + "sess_"
 
@@ -148,6 +152,7 @@ const (
 	mqttJSAMsgLoad        = "ML"
 	mqttJSASessPersist    = "SP"
 	mqttJSARetainedMsgDel = "RD"
+	mqttJSAStreamNames    = "SN"
 
 	// Name of the header key added to NATS message to carry mqtt PUBLISH information
 	mqttNatsHeader = "Nmqtt-Pub"
@@ -159,6 +164,9 @@ const (
 
 	// This is how frequently the timer to cleanup the sessions flappers map is firing.
 	mqttSessFlappingCleanupInterval = 5 * time.Second
+
+	// Default retry delay if transfer of old session streams to new one fails
+	mqttDefaultTransferRetry = 5 * time.Second
 )
 
 var (
@@ -217,6 +225,20 @@ type mqttAccountSessionManager struct {
 	replicas   int
 	rrmLastSeq uint64        // Restore retained messages expected last sequence
 	rrmDoneCh  chan struct{} // To notify the caller that all retained messages have been loaded
+	sp         sessPersist   // Used for cluster-wide processing of session records being persisted
+	domainTk   string        // Domain (with trailing "."), or possibly empty. This is added to session subject.
+}
+
+type sessPersist struct {
+	mu   sync.Mutex
+	ch   chan struct{}
+	head *sessPersistRecord
+	tail *sessPersistRecord
+}
+
+type sessPersistRecord struct {
+	seq  uint64
+	next *sessPersistRecord
 }
 
 type mqttJSA struct {
@@ -257,6 +279,7 @@ type mqttSession struct {
 	maxp     uint16
 	tmaxack  int
 	clean    bool
+	domainTk string
 }
 
 type mqttPersistedSession struct {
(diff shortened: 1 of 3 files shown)


server/test_test.go
```diff
@@ -200,6 +201,7 @@ func (c *cluster) shutdown() {
 			os.Remove(cf)
 		}
 		if sd != _EMPTY_ {
+			sd = strings.TrimSuffix(sd, JetStreamStoreDir)
```
kozlovic, line 204: @derekcollison I noticed that the temporary directory `nats-server` would still be filled with empty directory, such as `jetstream12345678`. Since it seems that under this directory, the server would actually create `jetstream` directory, this remove all would remove `jetstream` but leave the empty parent directory.
This change helps cleaning up, and seems that JetStream pass, but keep that in mind if later when writing a test in case this change has an effect (for instance if you were expecting only `jetstream` to be removed but not the parent dir, etc..).

server/mqtt.go
```diff
@@ -1947,6 +2045,66 @@ func (as *mqttAccountSessionManager) notifyRetainedMsgDeleted(subject string, se
...
 }
 
+func (as *mqttAccountSessionManager) transferUniqueSessStreamsToMuxed(log *Server) {
+	// Set retry to true, will be set to false on success.
+	retry := true
+	defer func() {
+		if retry {
+			next := mqttDefaultTransferRetry
+			log.Warnf("Failed to transfer all MQTT session streams, will try again in %v", next)
+			time.AfterFunc(next, func() { as.transferUniqueSessStreamsToMuxed(log) })
```
kozlovic, line 2055: @derekcollison I wonder if I should cap the retry to once or twice. My concern is that if simply getting the list of streams (say there are thousands of them) fails (line 2060 below), then server would go in this loop of trying to do this, etc..

server/mqtt.go
```diff
@@ -159,6 +164,9 @@ const (
 
 	// This is how frequently the timer to cleanup the sessions flappers map is firing.
 	mqttSessFlappingCleanupInterval = 5 * time.Second
+
+	// Default retry delay if transfer of old session streams to new one fails
+	mqttDefaultTransferRetry = 5 * time.Second
```
derekcollison, line 169: Might take longer.. If you try multiple times no worries.

server/mqtt.go
```diff
@@ -217,6 +225,20 @@ type mqttAccountSessionManager struct {
 	replicas   int
 	rrmLastSeq uint64        // Restore retained messages expected last sequence
 	rrmDoneCh  chan struct{} // To notify the caller that all retained messages have been loaded
+	sp         sessPersist   // Used for cluster-wide processing of session records being persisted
+	domainTk   string        // Domain (with trailing "."), or possibly empty. This is added to session subject.
```
derekcollison, line 229: If empty should it be "_."?

server/mqtt.go
```diff
@@ -893,6 +917,14 @@ func (s *Server) mqttCreateAccountSessionManager(acc *Account, quitCh chan struc
 			nuid:   nuid.New(),
 			quitCh: quitCh,
 		},
+		sp: sessPersist{
+			ch: make(chan struct{}, 1),
+		},
+	}
+	// We need to include the domain in the subject prefix used to store sessions in the $MQTT_sess stream.
+	// If no domain is present, use "_" so that the token always exists.
```
derekcollison, line 925: Where do we do the "_" part?

server/mqtt.go
```diff
@@ -1080,7 +1133,12 @@ func (jsa *mqttJSA) newRequestEx(kind, subject string, hdr int, msg []byte, time
 	// Either we use nuid.Next() which uses a global lock, or our own nuid object, but
 	// then it needs to be "write" protected. This approach will reduce across account
 	// contention since we won't use the global nuid's lock.
-	reply := jsa.rplyr + kind + "." + jsa.nuid.Next()
+	var sb strings.Builder
+	sb.WriteString(jsa.rplyr)
+	sb.WriteString(kind)
+	sb.WriteByte('.')
```
derekcollison, line 1139: I believe we have const for that.. pwc IIRC.

server/mqtt.go
```diff
@@ -1170,6 +1228,20 @@ func (jsa *mqttJSA) deleteStream(name string) (bool, error) {
 	return sdr.Success, sdr.ToError()
 }
 
+func (jsa *mqttJSA) loadLastMsgFor(streamName string, subject string) (*StoredMsg, error) {
+	mreq := &JSApiMsgGetRequest{LastFor: subject}
+	req, err := json.Marshal(mreq)
+	if err != nil {
+		return nil, err
+	}
+	lmri, err := jsa.newRequest(mqttJSAMsgLoad, fmt.Sprintf(JSApiMsgGetT, streamName), 0, req)
```
derekcollison, line 1237: ok to block in place here waiting on response?

server/mqtt.go
```diff
@@ -1378,6 +1486,32 @@ func (as *mqttAccountSessionManager) processSessionPersist(_ *subscription, pc *
 	sess.mu.Unlock()
 }
 
+func (as *mqttAccountSessionManager) sessPersistProcessing(closeCh chan struct{}) {
+	as.mu.RLock()
+	sp := &as.sp
+	quitCh := as.jsa.quitCh
```
derekcollison, line 1492: Should we guard jsa being nil?

server/mqtt.go
derekcollison: On the list of streams? Should be fine, I would keep trying but maybe space things out in case we are blocked as a cluster etc.

server/mqtt.go
kozlovic: This is the interval between retries (but does not count the time of a transfer itself). Not sure what you are asking me to change here? Make the interval longer?

server/mqtt.go
kozlovic: I am not sure I need to make a special token for empty. What I do when needed is simply to trim the prefix `$MQTT.sess.` + this value, which if empty is fine.

In other words, I don't think I need the `$MQTT.sess.` subject prefix to include something special if there is no domain.

server/mqtt.go
kozlovic: That is left over comment before I decided to not go with "_.".
Also, I realize that below I am using `getHash()` for the domain, and I am not sure I should do that. It may be ok (or even a good thing) to see what the domain is in this subject.

server/mqtt.go
kozlovic: Ok, will replace..

server/mqtt.go
kozlovic: Yes, all those helpers do that. They are invoked only in places where it is ok to wait.

server/mqtt.go
kozlovic: Never set to nil after being created. That's my MO..

server/mqtt.go
kozlovic: You mean use a backoff here? Save the next time and double or something (with a cap)?

server/mqtt.go
kozlovic: You meant `btsep`.

server/mqtt.go
derekcollison: ok sounds reasonable.

server/mqtt.go
derekcollison: Gotcha

server/mqtt.go
derekcollison: Just space them out in some form on error. No need for backoff IMO.

server/mqtt.go
derekcollison: Might make sense to do 10s, but where it is now is good too.

server/mqtt.go
kozlovic: But they are, every 5 seconds (next := mqtDefaultTransferRetry). I can make this constant 10s instead?

derekcollison (approved):
LGTM

PR #2506, opened 2021-09-09
Added ability to check if no message exists as a test to store the message.

### LINKED ISSUE #2494: Ability to publish to a kv style topic only if empty
#### Use Case:

Today we have `Nats-Expected-Last-Subject-Sequence` that limits publishing to a stream if the previous message had a specific sequence

Along the same lines it would be useful to ensure only the first client can write a message when the subject is empty.

We discussed maybe (ab)using `Nats-Expected-Last-Subject-Sequence:0` but decided not to, so maybe `Nats-Expected-Empty-Subject:1`

A sample use case could be for example registering a schema that relates to a specific topic, the first publisher to this topic should register the subject but that's true only ever once.

This setting should probably only be accepted in `max_msgs_per_subject > 0` stream

derekcollison:
Feature request.

Resolves #2494

/cc @nats-io/core

### DIFF
--- a/server/jetstream_test.go
+++ b/server/jetstream_test.go
@@ -12973,6 +12973,55 @@ func TestJetStreamPerSubjectPending(t *testing.T) {
 	}
 }
 
+func TestJetStreamPublishExpectNoMsg(t *testing.T) {
+	s := RunBasicJetStreamServer()
+	defer s.Shutdown()
+
+	config := s.JetStreamConfig()
+	if config != nil {
+		defer removeDir(t, config.StoreDir)
+	}
+
+	nc, js := jsClientConnect(t, s)
+	defer nc.Close()
+
+	_, err := js.AddStream(&nats.StreamConfig{
+		Name:              "KV",
+		Subjects:          []string{"KV.>"},
+		MaxMsgsPerSubject: 5,
+	})
+	if err != nil {
+		t.Fatalf("add stream failed: %s", err)
+	}
+
+	if _, err = js.Publish("KV.22", []byte("hello world")); err != nil {
+		t.Fatalf("Unexpected error: %v", err)
+	}
+
+	// This should succeed.
+	m := nats.NewMsg("KV.33")
+	m.Header.Set(JSExpectedLastSubjSeq, "0")
+	if _, err := js.PublishMsg(m); err != nil {
+		t.Fatalf("Unexpected error: %v", err)
+	}
+
+	// This should fail.
+	m = nats.NewMsg("KV.22")
+	m.Header.Set(JSExpectedLastSubjSeq, "0")
+	if _, err := js.PublishMsg(m); err == nil {
+		t.Fatalf("Expected error: %v", err)
+	}
+
+	if err := js.PurgeStream("KV"); err != nil {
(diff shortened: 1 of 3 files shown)


server/jetstream_test.go
```diff
@@ -12973,6 +12973,46 @@ func TestJetStreamPerSubjectPending(t *testing.T) {
...
+
+	// This should succeed.
+	m := nats.NewMsg("KV.33")
+	m.Header.Set(JSExpectedLastSubjSeq, "0")
+	if _, err := js.PublishMsg(m); err != nil {
+		t.Fatalf("Unexpected error: %v", err)
+	}
+
+	// This should fail.
+	m = nats.NewMsg("KV.22")
```
ripienaar, line 13009: Does it again work after purging the subject?

ripienaar (commented):
Question about behaviour

server/jetstream_test.go
derekcollison: Yes, just updated the test as well.

ripienaar (approved):
LGTM

server/stream.go
```diff
@@ -2686,9 +2686,13 @@ func (mset *stream) processJetStreamMsg(subject, reply string, hdr, msg []byte,
 			return fmt.Errorf("last msgid mismatch: %q vs %q", lmsgId, last)
 		}
 		// Expected last sequence per subject.
-		if seq := getExpectedLastSeqPerSubject(hdr); seq > 0 {
+		if seq, exists := getExpectedLastSeqPerSubject(hdr); exists {
 			// TODO(dlc) - We could make a new store func that does this all in one.
 			_, lseq, _, _, _, err := mset.store.LoadLastMsg(subject)
+			// If seq passed in is zero that signals we expect no msg to be present.
```
kozlovic, line 2692: I think that this change may break MQTT code that uses this header:
```
=== RUN   TestMQTTCluster
=== RUN   TestMQTTCluster/first_start
=== RUN   TestMQTTCluster/first_start/qos_0
    mqtt_test.go:2755: Error reading: EOF
--- FAIL: TestMQTTCluster (7.10s)
    --- FAIL: TestMQTTCluster/first_start (6.68s)
        --- FAIL: TestMQTTCluster/first_start/qos_0 (1.13s)
```
I set this header and initially seq is 0, however, it does not mean that I was not expecting any message to be set. Of course, we can fix the MQTT code and not set the header if know seq is 0.

kozlovic (changes requested):

ripienaar:
Yeah looks good to me apart from what Ivan says

server/stream.go
kozlovic: Actually, no, I don't think MQTT code should change. If I don't set the expected to 0, then I have the issue where 2 or more may be updating at the same time, and this is wrong.

So I think overloading the meaning of this header should be rethought.

kozlovic: Humm... I guess original code would have ignored it anyway, so for sure we could fix the MQTT code and/or mqtt test..

derekcollison: We paired and fixed.

kozlovic (approved):
LGTM

PR #2507, opened 2021-09-09
[ADDED] Monitoring: ClientID (for MQTT clients) on connection events

### LINKED ISSUE #2270: MQTT connect event
hello, i recently deployed an mqtt broker based on nats-server. thanks to nats2.2, that's a great work.
is there any way i can get notified when an mqtt client is connected, disconnected ?
thanks.

kozlovic:
ClientID has been added to various monitoring objects. Also, added
the ability to filter connections on `client_id`.

On auth violation, the proper code was not invoked, which meant
that no disconnect event (with auth reason) would be published.

Resolves #2270

### DIFF
--- a/server/client.go
+++ b/server/client.go
@@ -1895,7 +1895,11 @@ func (c *client) authViolation() {
 	} else {
 		c.Errorf(ErrAuthentication.Error())
 	}
-	c.sendErr("Authorization Violation")
+	if c.isMqtt() {
+		c.mqttEnqueueConnAck(mqttConnAckRCNotAuthorized, false)
+	} else {
+		c.sendErr("Authorization Violation")
+	}
 	c.closeConnection(AuthenticationViolation)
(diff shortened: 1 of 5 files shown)


derekcollison (commented):
In general looks good. If this is MQTT specific should we name it more geared towards that verus generic?

kozlovic:
Was thinking that we may make use of `ClientID` for other connection types in the future, so this is why I did not want to make it specific in the field name, although the comment says that as of now this is for MQTT clients only. But happy to change to MQTTClientID if you think it is better.

matthiashanel (approved):
Lgtm

derekcollison:
If we think we will really open it up to others maybe..

derekcollison:
I think MQTTClient is good.

derekcollison (approved):
LGTM

PR #2518, opened 2021-09-13
[FIXED] Report config error if "token" defined for cluster{} and gateway{}

kozlovic:
Authentication token has never been supported for cluster and
gateway (and leafnode). There is not even a Token option in
ClusterOpts or GatewayOpts.

However, the parsing of the configuration was not rejecting this
misconfiguration, making users believe that token was used for
authentication.

Documentation will also be fixed since it is reported there that
token is supported, which again, has never been the case.


derekcollison (approved):
LGTM

PR #2522, opened 2021-09-13
During compaction the cache may not be loaded completely

derekcollison:
This could lead to the filtered subject state being incorrect.

/cc @nats-io/core

### DIFF
--- a/server/filestore.go
+++ b/server/filestore.go
@@ -783,9 +783,7 @@ func (mb *msgBlock) rebuildStateLocked() (*LostStreamData, error) {
 
 		var deleted bool
 		if mb.dmap != nil {
-			if _, ok := mb.dmap[seq]; ok {
-				deleted = true
-			}
+			_, deleted = mb.dmap[seq]
 		}
 
 		// Always set last.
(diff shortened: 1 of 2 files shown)


server/filestore.go
```diff
@@ -4073,6 +4071,14 @@ func (mb *msgBlock) removeSeqPerSubject(subj string, seq uint64) {
 	if seq != ss.First {
 		return
 	}
+
+	// Here what we are removing is the first message.
+	// If we only have one message left we can simply assign it to last.
+	if ss.Msgs == 1 {
+		ss.First = ss.Last
```
kozlovic, line 4078: Just checking: is it that or ss.Last+1? Because if I remember, when we expire/remove all, then First is set to Last+1 to indicate that there are no message left.

kozlovic (approved):
LGTM but check my comment about first = last (to me if first == last, then it would mean that there is 1 message of that sequence, while if first = last+1 it can be indication that the first avail seq is "first" but there are currently no message. I think this is how it is in NATS Streaming).

server/filestore.go
derekcollison: So we removed the candidate msg above, since we are here we know that the one removed was the first, and we only have 1 msg left so first should equal last.

derekcollison: If no msgs left we remove a few lines above the complete entry.

kozlovic: Oh right, ss.Msgs == 1 is *after* the remove. My bad.

kozlovic (approved):
LGTM

PR #2523, opened 2021-09-13
js: Force expire pending fetch(n > 1) requests when above limit

wallyqs:
This prevents batch requests of n size, causing blocking issues when reaching the limit of max number of waiting pull requests.

 - [x] Tests added

/cc @nats-io/core

### DIFF
--- a/server/consumer.go
+++ b/server/consumer.go
@@ -1842,15 +1842,20 @@ func (wq *waitQueue) pop() *waitingRequest {
 	if wr != nil {
 		wr.n--
 		if wr.n <= 0 {
-			wq.reqs[wq.rp] = nil
-			wq.rp = (wq.rp + 1) % cap(wq.reqs)
-			// Check if we are empty.
-			if wq.rp == wq.wp {
-				wq.rp, wq.wp = -1, 0
-			}
+			wq.done()
 		}
 	}
 	return wr
+}
+
+// done will remove a next request from the queue.
+func (wq *waitQueue) done() {
+	wq.reqs[wq.rp] = nil
+	wq.rp = (wq.rp + 1) % cap(wq.reqs)
+	// Check if we are empty.
+	if wq.rp == wq.wp {
+		wq.rp, wq.wp = -1, 0
+	}
 }
 
 // processNextMsgReq will process a request for the next message available. A nil message payload means deliver
@@ -2070,8 +2075,10 @@ func (o *consumer) getNextMsg() (subj string, hdr, msg []byte, seq uint64, dc ui
 // Lock should be held.
 func (o *consumer) forceExpireFirstWaiting() *waitingRequest {
 	// FIXME(dlc) - Should we do advisory here as well?
-	wr := o.waiting.pop()
-	if wr == nil {
+	wr := o.waiting.peek()
+	if wr != nil {
+		o.waiting.done()
+	} else {
 		return wr
 	}
 	// If we are expiring this and we think there is still interest, alert.
--- a/server/jetstream_test.go
+++ b/server/jetstream_test.go
@@ -10392,6 +10392,362 @@ func TestJetStreamPullConsumerMaxAckPendingRedeliveries(t *testing.T) {
 	}
 }
 
+func TestJetStreamConsumerPullMaxWaiting(t *testing.T) {
+	s := RunBasicJetStreamServer()
+	defer s.Shutdown()
+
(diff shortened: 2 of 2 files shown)


server/consumer.go
```diff
@@ -2070,8 +2075,10 @@ func (o *consumer) getNextMsg() (subj string, hdr, msg []byte, seq uint64, dc ui
 // Lock should be held.
 func (o *consumer) forceExpireFirstWaiting() *waitingRequest {
 	// FIXME(dlc) - Should we do advisory here as well?
-	wr := o.waiting.pop()
-	if wr == nil {
+	wr := o.waiting.peek()
+	if wr != nil {
+		o.waiting.done()
```
kozlovic, line 2080: I don't feel confident to review this change, but what strikes me is that here you are by-passing the "pop()" which does reduce wr.n, and only remove when count is <= 0. Not sure what this means really, so again, I am not the best to review this.

server/consumer.go
```diff
@@ -2070,8 +2075,10 @@ func (o *consumer) getNextMsg() (subj string, hdr, msg []byte, seq uint64, dc ui
 // Lock should be held.
 func (o *consumer) forceExpireFirstWaiting() *waitingRequest {
 	// FIXME(dlc) - Should we do advisory here as well?
-	wr := o.waiting.pop()
-	if wr == nil {
+	wr := o.waiting.peek()
```
derekcollison, line 2078: This seems at first glance to be same logic as pop, so what is different here?

wallyqs: Here we are removing the fetch request forcefully regardless of the batch size, while `pop()` keeps the request valid until the batch request is fulfilled or if there is still interest. This is to avoid filling the queue with stale requests for example if the queue is filled with requests like `sub.Fetch(1000, nats.MaxWait(1*time.Hour)`, then no new fetch requests can be made for the consumer until the 1000 elements are received or one of the request expires.
With this change, in case of reaching the num waiting limit of inflight fetch requests, the older fetch requests are force expired to give room to the newer fetch requests.

server/jetstream_test.go
```diff
@@ -10392,6 +10392,362 @@ func TestJetStreamPullConsumerMaxAckPendingRedeliveries(t *testing.T) {
...
+		<-ctx.Done()
+
+		info, _ = subB.ConsumerInfo()
+		if info.NumWaiting != 0 {
+			t.Errorf("Expected: %v, got: %v", 0, info.NumWaiting)
+		}
+		if len(errs) == 0 {
+			t.Errorf("Expected at least an error, got: %v", len(errs))
+		}
+		if len(msgs) != 1 {
```
wallyqs, line 10632: Here unless the older sub.Fetch requests are force expired when reaching the num waiting limit, no messages will be able to be received until any of the sub.Fetch calls expire (in this example I set the deadline to 30 seconds).

server/jetstream_test.go
```diff
@@ -10392,6 +10392,362 @@ func TestJetStreamPullConsumerMaxAckPendingRedeliveries(t *testing.T) {
...
+		defer cancelFetch()
+
+		for i := 0; i < max; i++ {
+			expectedMap[fmt.Sprintf("quux:%d", i)] = false
+			go func() {
+				// Can only have at most 5 inflight fetch requests so
+				// so any fetch will timeout until these are canceled.
+				// These may receive some messages but they are not going
+				// to return until the context cancels the request forcefully later.
+				msgs, err := sub.Fetch(1000, nats.Context(ctx))
```
wallyqs, line 10550: These requests will block for 30 seconds any activity on the consumer, so new Fetch requests will not work and return `nats: timeout` until either 30 seconds pass or 1000 messages are received.

server/consumer.go
derekcollison: We have code that is supposed to deal with requests becoming invalid or expiring, why does that code not work here?

server/jetstream_test.go
derekcollison: I believe that is the correct behavior assuming the server still sees interest for the reply subject for the pending request.

server/jetstream_test.go
derekcollison: Might be worth discussing a bit, maybe we can all jump on a short call tomorrow.

wallyqs:
Closing for now since users less likely to run into this after this fix: #2538

PR #2526, opened 2021-09-14
[FIXED] #2525

### LINKED ISSUE #2525: Report of large message headers causing JetStream issues.

derekcollison:
We use u16 to encode header len when replicating JetStream messages.
Make sure to error if we exceed that limit.

Resolves #2525 

/cc @nats-io/core

### DIFF
--- a/server/jetstream_cluster.go
+++ b/server/jetstream_cluster.go
@@ -19,6 +19,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"math"
 	"math/rand"
 	"path"
 	"reflect"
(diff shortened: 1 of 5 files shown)


server/jetstream_cluster.go
```diff
@@ -4420,6 +4421,20 @@ func (mset *stream) processClusteredInboundMsg(subject, reply string, hdr, msg [
 		return err
 	}
 
+	// Since we encode header len as u16 make sure we do not exceed.
+	// Again this works if it goes through but better to be pre-emptive.
+	if len(hdr) > math.MaxUint16 {
+		err := fmt.Errorf("JetStream header size exceeds limits for '%s > %s'", jsa.acc().Name, mset.cfg.Name)
```
ripienaar, line 4427: on the assumption someone might be trying to trigger overflows or memory exchaustion attacks should we maybe log the IP of the client or some info about him?

derekcollison: If the message violates total payload limits or control line limits the error is logged and the client cutoff.

ripienaar: Yes, but I am mainly interested in having a way to know who is the bad client - i guess events will be raised so its probably fine. LGTM..

ripienaar (approved):
LGTM

PR #2529, opened 2021-09-14
Fixed a condition where JetStream assets could be created in multiple leafnode domains.

derekcollison:
Also added in optional Domain to StreamInfo.

/cc @nats-io/core

### DIFF
--- a/server/leafnode.go
+++ b/server/leafnode.go
@@ -1076,11 +1076,15 @@ func (c *client) processLeafnodeInfo(info *Info) {
 					// account also has JetStream enabled.
 					if accHasJS {
 						s.addInJSDenyExport(remote)
+						// If we specified a domain do not import by default.
+						if hasJSDomain {
+							s.addInJSDenyImport(remote)
+						}
 					}
 				}
 				// If we have a specified JetStream domain we will want to add a mapping to
 				// allow access cross domain for each non-system account.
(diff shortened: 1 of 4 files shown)


server/leafnode.go
```diff
@@ -1076,11 +1076,15 @@ func (c *client) processLeafnodeInfo(info *Info) {
 					// account also has JetStream enabled.
 					if accHasJS {
 						s.addInJSDenyExport(remote)
+						// If we specified a domain do not import by default.
+						if hasJSDomain {
+							s.addInJSDenyImport(remote)
+						}
```
matthiashanel, line 1082: This works, but I have one question about this. Why is it that this is guarded by `accHasJs`? 

If a downstream server (say through another leaf node hop) has no JS enabled, adding a deny should not matter as there is no interest on that end.
If such an account does have JS enabled, then there, presumably we would add the deny. Meaning a deny earlier on shouldn't matter either. 

What am I missing?

matthiashanel (approved):
LGTM, but one question.

server/leafnode.go
derekcollison: We don't need the deny unless the account has JS enabled (not the server), so just trying to keep it small (the deny stuff).

Not sure that answers your question though..

kozlovic (commented):
Will defer to Matthias that knows way more than me on that subject...

PR #2531, opened 2021-09-15
fix error code for JSStreamHeaderExceedsMaximumErr

ripienaar:
Fixes an error code that was added that left a big gap in numbers, this has
not been released so should be safe.

`nats errors add server/errors.json` will assist in picking the next number 
when adding errors

`go generate` will now fail on gaps

/cc @nats-io/core

### DIFF
--- a/server/errors.json
+++ b/server/errors.json
@@ -190,16 +190,6 @@
     "deprecates": ""
   },
   {
-    "constant": "JSStreamHeaderExceedsMaximumErr",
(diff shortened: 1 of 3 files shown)


derekcollison (approved):
LGTM - I do think we should review dependency on nats cli to change error codes and build the server though.

ripienaar:
It's not needed to build the server. It just helps construct the JSON.

PR #2532, opened 2021-09-15
MQTT add test to check for idempotent session stream creation.

kozlovic:
Creating a stream will become idempotent, so assuming that we
should try to transfer the old session streams only on success
will no longer work.

Added a test that checks that "stream" list is queried only once
which means transfer was attempted only once after the second
cluster restart and new connection.


derekcollison (approved):
LGTM

PR #2533, opened 2021-09-15
[CHANGED] JetStream: flow control requires heartbeats

### LINKED ISSUE #2520: Jetstream: Require Heartbeat when Flow Control is Set or provide alternative
## Feature Request

1A. Require `idle_heartbeat` when `flow_control` is requested when creating a consumer; server returns an error when the consumer is not created correctly.
-or - 
1B. Default `idle_heartbeat` when `flow_control` is requested but `idle_heartbeat` is not supplied when creating a consumer.
-or-
2. Provide an api call to manually request a heartbeat status message

#### Use Case:

Currently if a consumer is created with flow control is set, but heartbeat is not, if a flow control message is missed, there is no way to recover without restarting the server. The consumer simply stops getting messages.

#### Proposed Change:

Implement option (1A or 1B) and/or 2

Consumers can recover from missed flow control.

kozlovic:
The server will now reject the creation of a push consumer with
flow control if no heartbeat is set.

Resolves #2520

### DIFF
--- a/server/errors.json
+++ b/server/errors.json
@@ -1058,5 +1058,15 @@
     "help": "",
     "url": "",
     "deprecates": ""
+  },
+  {
+    "constant": "JSConsumerWithFlowControlNeedsHeartbeats",
+    "code": 400,
+    "error_code": 10108,
+    "description": "consumer with flow control also needs heartbeats",
+    "comment": "",
+    "help": "",
+    "url": "",
+    "deprecates": ""
   }
 ]
--- a/server/jetstream_errors_generated.go
+++ b/server/jetstream_errors_generated.go
@@ -152,6 +152,9 @@ const (
 	// JSConsumerWQRequiresExplicitAckErr workqueue stream requires explicit ack
 	JSConsumerWQRequiresExplicitAckErr ErrorIdentifier = 10098
 
+	// JSConsumerWithFlowControlNeedsHeartbeats consumer with flow control also needs heartbeats
+	JSConsumerWithFlowControlNeedsHeartbeats ErrorIdentifier = 10108
+
 	// JSInsufficientResourcesErr insufficient resources
 	JSInsufficientResourcesErr ErrorIdentifier = 10023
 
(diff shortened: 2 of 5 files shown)


derekcollison (approved):
LGTM - Did you check with R.I. on introducing the new error? Apparently there is a process.

kozlovic:
> Did you check with R.I. on introducing the new error? Apparently there is a process.

Yes, which I followed: `nats errors add server/errors.json` from the root repo. Then add my constant name, code and the error code is picked for me, etc..
Then `go generate` and that's it.

server/errors.json
```diff
@@ -1058,5 +1058,15 @@
...
+    "code": 400,
+    "error_code": 10108,
+    "description": "consumer with flow control also needs heartbeats",
+    "comment": "",
+    "help": "",
+    "url": "",
+    "deprecates": ""
   }
-]
+]
```
scottf, line 1072: don't know if the team is picky about the blank line at the end of file

server/jetstream_errors_generated.go
```diff
@@ -152,6 +152,9 @@ const (
 	// JSConsumerWQRequiresExplicitAckErr workqueue stream requires explicit ack
 	JSConsumerWQRequiresExplicitAckErr ErrorIdentifier = 10098
 
+	// JSConsumerWithFlowControlNeedsHeartbeats consumer with flow control also needs heartbeats
+	JSConsumerWithFlowControlNeedsHeartbeats ErrorIdentifier = 10108
```
scottf, line 156: Is there a scheme to error ids? Ask RI I suppose

scottf (approved):
Not directly related, but I'm just wondering if there is a minimum heartbeat that makes sense. Like a heartbeat every 100 ms seems like overkill, but maybe there is a use case for it. I mean getting a hearbeat every 100ms seems like a lot of unnecessary traffic, Maybe there should be a minimum?

ripienaar:
Heartbeats also useful in some lesser election scenarios, and, at least in tests, very short intervals help 

no real schema for constants just make them obvious

ripienaar (approved):
LGTM

derekcollison:
Agree we should have a sane minimum. Default for directs for sources and mirrors is 2s, I think OrderedConsumer is 5s.

I might offer 500ms as lowest, but maybe even 1s.

server/errors.json
kozlovic: This is done by the `nats errors add` and in general git will report if there is no newline after the last character, so that's a good thing.

PR #2535, opened 2021-09-16
Allow clustered JetStream to allow duplicate stream creation calls.

### LINKED ISSUE #2528: Adding an already defined stream behaves inconsistently depending if connected to a single server or a cluster
Currently if you call js.AddStream for a stream that has already been defined (and the attributes of the stream that you are trying to create with js.AddStream match the attributes of the already defined stream) behave inconsistently:

- If the client application is connected to a single nats-server then the call behaves as if it is idempotent (i.e. returns an error only if the stream you are trying to add has a mismatch of attribute(s) with the already defined stream)
- If the client application is connected to a cluster of nats-servers the call behaves as if it is _not_ idempotent (i.e. always returns an error if there's a stream with the same name already defined regardless of whether the attributes match or not)

derekcollison:
This makes the behavior consistent with single server mode.

Resolves #2528

/cc @nats-io/core

### DIFF
--- a/server/events.go
+++ b/server/events.go
@@ -240,8 +240,10 @@ type pubMsg struct {
 	sub  string
 	rply string
 	si   *ServerInfo
+	hdr  map[string]string
 	msg  interface{}
 	oct  compressionType
+	echo bool
 	last bool
 }
 
@@ -359,7 +361,6 @@ RESET:
 					b = bb.Bytes()
 					contentHeader = "snappy"
 				case unsupportedCompression:
-					b = c.setHeader(contentEncodingHeader, "identity", b)
 					contentHeader = "identity"
 				}
 			}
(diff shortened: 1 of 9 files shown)


ripienaar (approved):
LGTM

server/events.go
```diff
@@ -415,11 +426,16 @@ func (s *Server) sendShutdownEvent() {
 	s.mu.Unlock()
 	// Send to the internal queue and mark as last.
 	si := &ServerInfo{}
-	sendq <- &pubMsg{nil, subj, _EMPTY_, si, si, noCompression, true}
+	sendq <- &pubMsg{nil, subj, _EMPTY_, si, nil, si, noCompression, false, true}
```
kozlovic, line 429: If fields would be named, only the calls where we actually pass the non default values would need to be changed, but I understand that this was done this way so that adding a new field would fail compilation, forcing you to review all placed where pubMsg is initialized.

server/jetstream_cluster.go
```diff
@@ -1890,6 +1891,7 @@ func (js *jetStream) processStreamLeaderChange(mset *stream, isLeader bool) {
 			Sources: mset.sourcesInfo(),
 			Mirror:  mset.mirrorInfo(),
 		}
+		resp.DidCreate = true
```
kozlovic, line 1894: It is a bit confusing because this function (leader change) sounds like a node could be elected to be a leader for an existing stream. Say the node that is currently the leader dies, some other node may become the leader for a stream (assuming R>1). Yet, we want "DidCreate" to be true really only once, when the stream is created.

kozlovic (approved):
LGTM unless the "processStreamLeaderChange" is invoked more than once for a given stream.

derekcollison:
processStreamLeaderChange can be called multiple times, but we remember if we have responded etc. What is the actual concern?

server/jetstream_cluster.go
derekcollison: If you look above we bail on this function before considering reporting if we are *not* the leader or have already responded.

kozlovic: I see, ok then!

kozlovic (approved):
LGTM

PR #2537, opened 2021-09-16
[FIXED] ConsumerInfo in clustered mode during setup returned NotFound.

derekcollison:
This change returns an info in those states now to better support client logic when binding consumers.

/cc @nats-io/core


kozlovic (approved):
LGTM

kozlovic (approved):
LGTM

PR #2539, opened 2021-09-16
Exposing reserved memory in jsz/varz

matthiashanel:

@wallyqs this probably needs to make it into prometheus

### DIFF
--- a/server/jetstream.go
+++ b/server/jetstream.go
@@ -49,10 +49,12 @@ type JetStreamConfig struct {
 }
 
 type JetStreamStats struct {
-	Memory   uint64            `json:"memory"`
-	Store    uint64            `json:"storage"`
-	Accounts int               `json:"accounts,omitempty"`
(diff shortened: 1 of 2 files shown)


wallyqs (approved):
LGTM

ripienaar (approved):
LGTM

derekcollison (approved):
LGTM

server/jetstream.go
```diff
@@ -1645,6 +1647,8 @@ func (js *jetStream) usageStats() *JetStreamStats {
 	var stats JetStreamStats
 	js.mu.RLock()
 	stats.Accounts = len(js.accounts)
+	stats.ReservedMemory = (uint64)(js.memReserved)
+	stats.ReservedStore = (uint64)(js.storeReserved)
```
ripienaar, line 1651: these are zero when dynamic limits are used rather than fixed ones from config, would be nice if they adjust to the dynamically reserved level rather than return 0

PR #2542, opened 2021-09-18
Remove flag that strips dwarf symbols

yzhao1012:
Dwarf symbols are used by debugging tools to get the layout of the data structure, and many other runtime information.
For example gdb needs the calling convention information to figure out how to invoke a function.

The increase in executable size is high though, current size of the dwarf symbols built from the mainline is ~3MB.
Or the executable file built with `go build` is ~15MB, and `go build -ldflags "-w"` is ~12MB.

Let me know if this is acceptable.

### DIFF
--- a/.goreleaser-nightly.yml
+++ b/.goreleaser-nightly.yml
@@ -5,7 +5,7 @@ builds:
     id: nats-server
     binary: nats-server
     ldflags:
-      - -w -X github.com/nats-io/nats-server/server.gitCommit={{.ShortCommit}}
+      - -X github.com/nats-io/nats-server/server.gitCommit={{.ShortCommit}}
     env:
       - GO111MODULE=off
       - CGO_ENABLED=0
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -14,7 +14,7 @@ builds:
 - main: ./main.go
   binary: nats-server
   ldflags:
-    - -w -X github.com/nats-io/nats-server/server.gitCommit={{.ShortCommit}}
+    - -X github.com/nats-io/nats-server/server.gitCommit={{.ShortCommit}}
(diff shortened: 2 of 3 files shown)


kozlovic:
@yzhao1012 As you know we already not stripping symbols anymore (since you are the one that submitted PR earlier). I don't think having the dwarf symbols in a release product is necessary. It is unlikely that one will do a "gdb" session on a live production server. The product is open-sourced and anyone can build it the way they want. If some users have the need to debug the server, they can build a version without the "-w" flag.

alexbozhenko:
Go's own [doc on gdb](https://go.dev/doc/gdb) says :
> Note that [Delve](https://github.com/go-delve/delve) is a better alternative to GDB when debugging Go programs built with the standard toolchain. It understands the Go runtime, data structures, and expressions better than GDB. Delve currently supports Linux, OSX, and Windows on amd64. For the most up-to-date list of supported platforms, please see[ the Delve documentation](https://github.com/go-delve/delve/tree/master/Documentation/installation).

But Delve too complains about missing debug symbols:
```
go install github.com/go-delve/delve/cmd/dlv@latest
# dlv attach $(pidof nats-server)
Warning: no debug info found, some functionality will be missing such as stack traces and variable evaluation.
Type 'help' for list of commands.
```

Currently, the difference between the binary `-w` flag is 15MB, and without the flag: 20MB

alexbozhenko:
related discussion: https://github.com/golang/go/issues/26074

neilalexander:
At the moment I would prefer to keep flags as-is, production systems won't be running under delve and development systems can build from source if needed.

alexbozhenko:
for reference, the flag is documented here: https://pkg.go.dev/cmd/link

PR #2546, opened 2021-09-20
protect against negative dupe window via negative max age

ripienaar:

/cc @nats-io/core

### DIFF
--- a/server/jetstream_test.go
+++ b/server/jetstream_test.go
@@ -13094,6 +13094,39 @@ func TestJetStreamPullLargeBatchExpired(t *testing.T) {
 	}
 }
 
+func TestNegativeDupeWindow(t *testing.T) {
+	s := RunBasicJetStreamServer()
+	defer s.Shutdown()
+
+	config := s.JetStreamConfig()
+	if config != nil {
+		defer removeDir(t, config.StoreDir)
+	}
+
+	nc, js := jsClientConnect(t, s)
+	defer nc.Close()
+
+	// we incorrectly set MaxAge to -1 which then as a side effect sets dupe window to -1 which should fail
+	_, err := js.AddStream(&nats.StreamConfig{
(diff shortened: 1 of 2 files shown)


derekcollison (approved):
LGTM

ripienaar:
I considered doing this, which if we go this way would probably remove over 1KLOC of boiler plate, thoughts @derekcollison ?

```go
type testServer struct {
	s  *Server
	nc *nats.Conn
	js nats.JetStreamContext
}

func withBasicJetStreamServer(t *testing.T, f func(*testing.T, *testServer)) {
	t.Helper()

	defer s.Shutdown()

	config := s.JetStreamConfig()
	if config != nil {
		defer removeDir(t, config.StoreDir)
	}

	defer nc.Close()

	f(t, &testServer{s, nc, js})
}

func TestNegativeDupeWindow(t *testing.T) {
	withBasicJetStreamServer(t, func(t *testing.T, srv *testServer) {
		// we incorrectly set MaxAge to -1 which then as a side effect sets dupe window to -1 which should fail
		_, err := srv.js.AddStream(&nats.StreamConfig{
			Name:              "TEST",
			Subjects:          nil,
			Retention:         nats.WorkQueuePolicy,
			MaxConsumers:      1,
			MaxMsgs:           -1,
			MaxBytes:          -1,
			Discard:           nats.DiscardNew,
			MaxAge:            -1,
			MaxMsgsPerSubject: -1,
			MaxMsgSize:        -1,
			Storage:           nats.FileStorage,
			Replicas:          1,
			NoAck:             false,
		})
		if err == nil || err.Error() != "duplicates window can not be negative" {
			t.Fatalf("Expected dupe window error got: %v", err)
		}
	})
}
```

derekcollison:
Great idea! Let's get 2.5.1 out but I like the approach.

PR #2549, opened 2021-09-21
Fix RAFT WAL repair.

derekcollison:
When we stored a message in the raft layer in a wrong position (state corrupt), we would panic, leaving the message there.
On restart we would truncate the WAL and try to repair, but we truncated to the wrong index of the bad entry.

This change also includes additional changes to truncateWAL and also reduces the conditional for panic on storeMsg.

/cc @nats-io/core

### DIFF
--- a/server/raft.go
+++ b/server/raft.go
@@ -235,20 +235,22 @@ type RaftConfig struct {
 }
 
 var (
-	errProposalFailed  = errors.New("raft: proposal failed")
-	errNotLeader       = errors.New("raft: not leader")
-	errAlreadyLeader   = errors.New("raft: already leader")
-	errNilCfg          = errors.New("raft: no config given")
-	errCorruptPeers    = errors.New("raft: corrupt peer state")
-	errStepdownFailed  = errors.New("raft: stepdown failed")
-	errEntryLoadFailed = errors.New("raft: could not load entry from WAL")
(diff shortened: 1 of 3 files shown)


wallyqs (approved):
LGTM

server/raft.go
```diff
@@ -418,12 +420,17 @@ func (s *Server) startRaftNode(cfg *RaftConfig) (RaftNode, error) {
 		for index := state.FirstSeq; index <= state.LastSeq; index++ {
 			ae, err := n.loadEntry(index)
 			if err != nil {
-				n.warn("Could not load %d from WAL [%+v] with error: %v", index, state, err)
-				continue
+				n.warn("Could not load %d from WAL [%+v]: %v", index, state, err)
```
ripienaar, line 423: These are pretty scary warnings to users, can we add some indication about what they can do about them? Or mention if the server will recover on its own? 

Seeing just these in the server log and no indication of what action to take is not a good experience.  For myself even, I have no idea what to do about these log lines

wallyqs (approved):
LGTM

server/raft.go
ripienaar: OK, but like this log says "ask Derek" :) if that's the intended outcome, LGTM :)

kozlovic (approved):
LGTM

PR #2554, opened 2021-09-21
[FIXED] #2551

### LINKED ISSUE #2551: JetStream: Invalid consumer filter subject on server restart in some cases
## Defect

A consumer C1 is created and functions normally, but upon server restart fails to be recovered with following warning:

```text
[184164] 2021/09/20 21:12:07.346160 [INF]   Recovering 1 consumers for stream - "Q1"
[184164] 2021/09/20 21:12:07.346963 [WRN]     Error adding consumer: consumer filter subject is not a valid subset of the interest subjects (10093)
```
C1 can be created successfully again, but above error will always reproduce on next server restart.

The issue occurs with C1 on replica stream Q1 (which sources upstream S1).  Creating C2 with identical filter subject but on non-replica stream S1 works and does not have restart issue.

The consumer filter subject `message.>` is in fact a possible superset of upstream S1 subject `message.s1.*`.

[nats-server -DV output](https://github.com/nats-io/nats-server/files/7200175/output.txt)

* nats-server 2.5.0 and 2.5.1-beta.1 both reproduce.
* nats-cli 0.0.26 (nats con add)

Linux, non-Dockerized nats-server

With streams S1, Q1 (sourcing S1) and consumer C1 (on Q1) as [tbeets/nats-replica](https://github.com/tbeets/nats-replica), C1 creates and functions normally.  On server restart, C1 fails to be restored with:

```text
[184164] 2021/09/20 21:12:07.346160 [INF]   Recovering 1 consumers for stream - "Q1"
[184164] 2021/09/20 21:12:07.346963 [WRN]     Error adding consumer: consumer filter subject is not a valid subset of the interest subjects (10093)
```
#### Expected result:

Consumer creation with a filter subject that is not a valid subset of stream subjects should consistently fail or succeed.

#### Actual result:

In C1 case (consumer on a replica stream), initial creation succeeds, but server restore on restart fails.
In C2 case (consumer on non-replica stream), initial creation and server restore both succeed.

derekcollison:
When a mirror would be processed before the origin stream we would not recover the consumers due to failure on looking up source's subjects.

This change processes all streams first then does all consumers.

Resolves #2551 

/cc @nats-io/core

### DIFF
--- a/server/jetstream.go
+++ b/server/jetstream.go
@@ -1034,6 +1034,13 @@ func (a *Account) EnableJetStream(limits *JetStreamAccountLimits) error {
 		}
 	}
 
+	// Collect consumers, do after all streams.
+	type ce struct {
+		mset *stream
+		odir string
+	}
+	var consumers []*ce
+
 	// Now recover the streams.
 	fis, _ := ioutil.ReadDir(sdir)
 	for _, fi := range fis {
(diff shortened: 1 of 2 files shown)


kozlovic (approved):
LGTM

PR #2561, opened 2021-09-22
[FIXED] Deadlock with stream mirrors

derekcollison:
Fix for deadlock with stream mirrors or sources where origin is interest or workqueue policy.
This would cause an interaction with channels and locks that could cause a deadlock.

/cc @nats-io/core


kozlovic (approved):
LGTM

PR #2563, opened 2021-09-22
[FIXED] Storage limit of 0, which means disabled, was not always enforced

matthiashanel:

The attached unit test shows that a mixed enabled/disabled use case does neither work in jwt nor in config.
For json parsing reasons, `0` has to be the disabled state.
in `opts.go` line `1562` I can see this code, indicating that `0` is the disabled state.

```go
var dynamicJSAccountLimits = &JetStreamAccountLimits{-1, -1, -1, -1}
```


philpennock:
I support this change in theory, but am not familiar enough with those parts of the code-base to perform a quality review.

matthiashanel:
> I support this change in theory, but am not familiar enough with those parts of the code-base to perform a quality review.

This is not a fix. This is to demonstrate that it's not working. no review needed.

derekcollison:
ok I pushed a CL that fixes.

derekcollison:
@kozlovic can you and @matthiashanel take a quick look?

server/jwt_test.go
```diff
@@ -4430,17 +4430,7 @@ func TestJWTJetStreamDisabledLimitsEnforcement(t *testing.T) {
...
-	si, err := js.StreamInfo("disk")
-	require_NoError(t, err)
-
-	require_True(t, si.State.Msgs == 0)
-
-	// TODO consider changing from off to on and on to off (delete directory) on update
+	require_Error(t, err)
 }
 
 func TestJetStreamDisabledLimitsEnforcement(t *testing.T) {
```
matthiashanel, line 4436: Maybe move that test into the corresponding jetstream file.

matthiashanel (commented):
LGTM, perhaps move that one test, but only a suggestion.
But can't approve myself as I am the original author.

server/jwt_test.go
derekcollison: Feel free to move things around etc. The fix is there.

kozlovic:
@matthiashanel I pushed a commit that fixes a MQTT test that was enabling JS for an account but was not setting MaxStore. So to be clear, if no value is set (default is 0 based on type), then it means disabled. Is that considered a change or was more of a bug (was meant to be disabled but was not)?
Change here: https://github.com/nats-io/nats-server/pull/2563/commits/b0d487fe68100b0e03bd91ebec19e6dafa94ae38

matthiashanel:
> n it means disabled

bug

kozlovic (approved):
LGTM

PR #2564, opened 2021-09-22
Direct consumers used for mirroring should not be affected by max consumers limits

derekcollison:

/cc @nats-io/core

### DIFF
--- a/server/stream.go
+++ b/server/stream.go
@@ -3303,8 +3303,8 @@ func (mset *stream) getMsg(seq uint64) (*StoredMsg, error) {
 
 // getConsumers will return all the current consumers for this stream.
 func (mset *stream) getConsumers() []*consumer {
-	mset.mu.Lock()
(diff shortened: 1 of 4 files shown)


server/stream.go
```diff
@@ -3313,10 +3313,21 @@ func (mset *stream) getConsumers() []*consumer {
 	return obs
 }
 
+// Lock should be held for this one.
+func (mset *stream) numPublicConsumers() int {
```
kozlovic, line 3317: Looks like you are keeping track of the directs in setConsumer() and remoteConsumer(), so would not it be simpler here to do `len(mset.consumers) - mset.direct`?

kozlovic (approved):
LGTM, but I feel like you are not taking advantage of the `direct` count that is updated on add/remove...

kozlovic (approved):
LGTM

PR #2565, opened 2021-09-22
Make sure wp is > len(cache.buf)

derekcollison:
Reported on startup by JNM.

/cc @nats-io/core


kozlovic (approved):
LGTM

PR #2568, opened 2021-09-23
[FIXED] Internal connections were counted toward account's max connections (client or leaf)

matthiashanel:
Question, what's the expected influence of JS on conn count?

### DIFF
--- a/server/events.go
+++ b/server/events.go
@@ -49,7 +49,6 @@ const (
 	accUpdateEventSubjNew    = "$SYS.REQ.ACCOUNT.%s.CLAIMS.UPDATE"
 	connsRespSubj            = "$SYS._INBOX_.%s"
 	accConnsEventSubjNew     = "$SYS.ACCOUNT.%s.SERVER.CONNS"
-	accConnsEventSubjOld     = "$SYS.SERVER.ACCOUNT.%s.CONNS" // kept for backward compatibility
 	shutdownEventSubj        = "$SYS.SERVER.%s.SHUTDOWN"
 	authErrorEventSubj       = "$SYS.SERVER.%s.CLIENT.AUTH.ERR"
 	serverStatsSubj          = "$SYS.SERVER.%s.STATSZ"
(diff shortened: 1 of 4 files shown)


server/events.go
```diff
@@ -1613,7 +1613,7 @@ func (s *Server) accConnsUpdate(a *Account) {
 	if !s.eventsEnabled() || a == nil {
 		return
 	}
-	s.sendAccConnsUpdate(a, fmt.Sprintf(accConnsEventSubjOld, a.Name), fmt.Sprintf(accConnsEventSubjNew, a.Name))
+	s.sendAccConnsUpdate(a, fmt.Sprintf(accConnsEventSubjNew, a.Name))
```
kozlovic, line 1616: Why remove "old" subject? Would that possibly break apps listening to old subject? Don't mind dropping, but maybe not in 2.6.1...

kozlovic (changes requested):
Don't want to drop "old" subject in this update release, I think..

matthiashanel:
It would

server/events.go
matthiashanel: It would

matthiashanel (commented):
Fine with changes. But would not remove the backwards compatibility subject

server/events.go
derekcollison: Possibly if apps were listening to those updates, but was messing up the accounting logic even beyond the counting JETSTREAM and ACCOUNT clients internally.

derekcollison: We could send to both if you both feel strongly, but we can't process both as subs..

derekcollison: I personally have no problem removing all together, but will trust your judgement here. But we can't process both for sure.

matthiashanel: we can still send them all we have to do is not subscribe in events.go line 780.

given this comment:
```
// accConnsUpdate is called whenever there is a change to the account's
// number of active connections, or during a heartbeat.
func (s *Server) accConnsUpdate(a *Account) {
```
how is processing the message twice different from a change in account followed by a heartbeat?

kozlovic (approved):
LGTM

PR #2573, opened 2021-09-27
[ADDED] TLS connection rate limiter

julius-welink:
 - [x] Tests added

TLS handshake negotiation is very CPU intensive, especially if RSA is used. For this reason I see that nats has "Lame Duck Mode". Unfortunately, during a network outage, it is still possible for many clients to try to connect at the same time. This rationale behind tls.connection_rate_limit is that we prevent CPU from being overwhelmed by a surge of clients trying to connect at the same time, by dropping the connection before TLS is initialized, thus effectively throttling TLS connection rate.

The good thing about this change is that is does not affect current nats functionality, unless enabled. Also, I will soon be testing this in a real project with more than 500 clients.

Other solutions that I have considered:
1. SYN rate limit on iptables. Unfortunately it's not possible to configure them within a docker container and I use docker.
2. Sidecar haproxy. This could work, but I believe rate limiter directly implemented in nats-server is a more straight forward solution. Haproxy would be an additional point of failure or misconfiguration.

The is no related issue afaik. I am making this change for myself. PR should be a good way to get the tests running, since TravisCI does not cooperate with me :) I am willing to spend more time on this PR if there is a chance it could be accepted to the main nats-server repo. Please let me know what else I need to do to get this approved.

I'll wait for some feedback before squashing commits. Do I need to update docs? If yes, how?

/cc @nats-io/core


julius-welink:
I'm seeing some test failures unrelated to my PR.

server/server.go
```diff
@@ -362,6 +364,10 @@ func NewServer(opts *Options) (*Server, error) {
 		httpReqStats: make(map[string]uint64), // Used to track HTTP requests
 	}
 
+	if opts.tlsConfigOpts != nil && opts.tlsConfigOpts.RateLimit > 0 {
```
kozlovic, line 367: When you bring it to Options, it will become:
```
if opts.TLSRateLimit > 0 {
    s.connRateCounter = newRateCounter(opts.TLSRateLimit)
}
```

kozlovic (changes requested):
I have made some comments. I think the idea is good, but I have mainly commented about the code, I would like other, say @philpennock or @derekcollison to comment on the approach itself.

julius-welink:
Thanks for the review! Will address the above comments soon. I have some more considerations:

[ ] Make sure that throttling only affects clients and not server -> server connections
[ ] I have some special clients that I wish were not throttled, but this would probably require some protocol changes 🤔

server/rate_counter.go
```diff
@@ -0,0 +1,38 @@
+package server
```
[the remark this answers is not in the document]
julius-welink: Added copyright.

Utils mostly contains functions, where rate counter is "a class". (sry I'm an OOP guy :) 

Still move to utils ?

server/server.go
```diff
@@ -258,6 +258,8 @@ type Server struct {
 	rerrMu   sync.Mutex
 	rerrLast time.Time
 
+	connRateCounter *rateCounter
```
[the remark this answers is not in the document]
julius-welink: created in `server.go:368`

```
s.connRateCounter = newRateCounter(opts.tlsConfigOpts.RateLimit)
```

server/rate_counter_test.go
```diff
@@ -0,0 +1,27 @@
...
+		if !counter.allow() {
+			break
+		}
+	}
+
+	if i != 10 {
+		t.Errorf("Expected i = 10, got %d", i)
+	}
+
+	time.Sleep(1100 * time.Millisecond)
```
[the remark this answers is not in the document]
julius-welink: I disagree on this, because that would not properly test `if now.After(r.end) {` condition. The next best thing I can think of is to inject time in to allow:
```
func (r *rateCounter) allow(t time.Time) bool {
```

server/opts.go
```diff
@@ -497,6 +497,7 @@ type TLSConfigOpts struct {
 	Map               bool
 	TLSCheckKnownURLs bool
 	Timeout           float64
+	RateLimit         int64
```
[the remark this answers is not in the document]
julius-welink: Do we need  `json:"tls_connection_rate_limit"` ?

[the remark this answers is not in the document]
kozlovic: No.

server/rate_counter.go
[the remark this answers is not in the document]
kozlovic: If you keep this file, which is fine, change copyright date to simply 2021. We have used (maybe incorrectly?) the logic that the date range if based on the file creation/last modification. So since you add this file now, just put 2021.

server/rate_counter_test.go
```diff
@@ -0,0 +1,27 @@
+package server
```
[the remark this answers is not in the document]
kozlovic: Same comment than for the other file (date: 2021)

server/rate_counter_test.go
```diff
@@ -0,0 +1,41 @@
...
+
+import (
+	"testing"
+	"time"
+)
+
+func TestRateCounter(t *testing.T) {
+	counter := newRateCounter(10)
+	counter.interval = 10 * time.Millisecond
```
kozlovic, line 23: To avoid flapping, I would make this bigger, say 100ms.

server/server.go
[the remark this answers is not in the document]
kozlovic: Oh sorry, github diff tricked me here and I missed that it was in the server struct definition, saw it as inside NewServer. So all good.

test/tls_test.go
```diff
@@ -1819,6 +1819,42 @@ func TestTLSPinnedCertsClient(t *testing.T) {
...
+func TestTLSConnectionRate(t *testing.T) {
+	config := `
+        host: localhost
+	    port: -1
+		
+		tls {
+		  cert_file: "./configs/certs/server-cert.pem"
+		  key_file:  "./configs/certs/server-key.pem"
+		  connection_rate_limit: 3
+		}
```
kozlovic, line 1831: Would try to fix spacing here... Also recommend replacing host/port with:
```
listen: "127.0.0.1:-1"
```

kozlovic (changes requested):
Just some more little changes. Thanks!

kozlovic (approved):
LGTM. From a code perspective, this looks good to me. Want to let @derekcollison decide if we merge this or if the rate limit approach is not warranted at this time.

julius-welink:
> LGTM. From a code perspective, this looks good to me. Want to let @derekcollison decide if we merge this or if the rate limit approach is not warranted at this time.

server/server.go
```diff
@@ -2409,6 +2415,15 @@ func (s *Server) createClient(conn net.Conn) *client {
 
 	// Check for TLS
 	if !isClosed && tlsRequired {
+		if s.connRateCounter != nil && !s.connRateCounter.allow() {
+			c.Warnf("Rejecting connection due to TLS rate limiting")
```
philpennock, line 2419: I think this is a change in the log volume surfacing as a result of client actions, right?  We will log freely for peers within a cluster, or across gateways, but tend to be cautious about log messages which scale with end-user connections.  (Particularly if the client IP is decorated into the log, as that can be PII).  We should definitely log this, but I think (regrettably) it might need to be at debug or trace level.  @derekcollison can better say what the policy is here.

We should probably add new fields to the `type Varz` in `server/monitor.go`; something like `ThrottledConnections` ?

derekcollison:
I am ok with it, but still feel mostly the same as my original comment.

server/server.go
julius-welink: What would you say if I aggregated logs within 1s window?

`
Rejected 100 connections due to TLS rate limiting
`

philpennock: Seems sane enough, but keeping track and doing the conditional logging seems like more work than adding a counter to the stats tracking and incrementing it. :)

julius-welink: You're right, quite a bit of code:
https://github.com/nats-io/nats-server/pull/2573/commits/5e5eb63a07c5dd32d0c520d881dfb61ef0c5bc18

if you think that's fine, I could rebase the whole PR on the latest nats codebase. Possibly open a new PR with changed squished to a single commit. Let me know if this is up for merging or not.

server/rate_counter.go
```diff
@@ -0,0 +1,65 @@
+// Copyright 2021 The NATS Authors
```
kozlovic, line 1: Change to 2021-2022 now :-)

server/rate_counter_test.go
```diff
@@ -0,0 +1,52 @@
+// Copyright 2021 The NATS Authors
```
kozlovic, line 1: Same here..

server/server.go
```diff
@@ -362,6 +364,26 @@ func NewServer(opts *Options) (*Server, error) {
 		httpReqStats: make(map[string]uint64), // Used to track HTTP requests
 	}
 
+	if opts.TLSRateLimit > 0 {
+		s.connRateCounter = newRateCounter(opts.tlsConfigOpts.RateLimit)
+		s.startGoRoutine(func() {
```
kozlovic, line 369: This would not work since the server has not been started yet. `startGoRoutine()` will not start the go routine if `s.grRunning` is not set to true, which happens only in `Server.Start()`.
Since this routine just does something based on a ticker, we could arguably change that to a timer func, but it would need to be stopped on Shutdown().
If you keep this go routine, still move the s.startGoRoutine() to Start(), or even AcceptLoop() since this applies only to client (when TLS is enabled).
(also, make it a separate function, so that in AcceptLoop we have `s.startGoRoutine(s.logRejectedTLSConns)` or something like that..)

test/tls_test.go
```diff
@@ -1819,6 +1819,40 @@ func TestTLSPinnedCertsClient(t *testing.T) {
...
+			connection_rate_limit: 3
+		}
+	`
+
+	confFileName := createConfFile(t, []byte(config))
+	defer removeFile(t, confFileName)
+
+	srv, _ := RunServerWithConfig(confFileName)
+	defer srv.Shutdown()
+
```
kozlovic, line 1837: Here you could add a debug logger that captures warnings, which I then think that would have shown that the logger would not work (because of startGoRoutine() invoked in NewServer()).

Something like this:
```
l := &captureWarnLogger{warn: make(chan string, 100)}
s.SetLogger(l, false, false)
```

Then at the end of this test, check that you got some of the warnings you expect:
```
for {
	select {
	case w := <-l.warn:
		if strings.Contains(w, "connections due to TLS rate limiting") {
			return
		}
	case <-time.After(time.Second):
		t.Fatal("Did not get warning")
	}
}
```

kozlovic (changes requested):

kozlovic:
@julius-welink Also forgot to mention that you have conflicts, so you may want to rebase from main branch..

kozlovic (approved):
LGTM. @derekcollison I think there is still value and I am ok with this PR, should we merge?

julius-welink:
@derekcollison 
> I am ok with it, but still feel mostly the same as my original comment.
> cluster itself was not sized properly.

image

This spike here is when client reconnect. So to properly size the cluster I'd need to spend 20x more on hosting fees, just to account for these reconnect spikes. If I don't, NATS would simply stop responding due to high load and not let in any clients. It has happened. In the case of mass reconnect, I'd rather put clients on queue.

> large RSA certs

I'm playing around with EC (Elliptic Curve) certs. Haven't yet tested in production, but I hope those will reduce CPU usage.


Something else on my mind. This is out problem statement: "NATS would simply stop responding due to high load". It should be technically possible to stop accepting new connections when NATS detects high CPU usage. I've seen you do measurements of "event loop" delay. Under a high load NATS was logging warnings. This perhaps seems a bit less reliable, how would you even test that? Nevertheless, it could be an alternative to this MR. Do you think it's feasible to go to this direction instead?

derekcollison:
I understand your concerns on costs. What I was trying to highlight is that in a total system design, you need to account and size for client rollover during server upgrades, network outages or even server failures themselves. When using large RSA keys that becomes something that needs to be tested.

One answer is of course ECC keys which can help. We also do jitter in the clients and if you control the clients can help out here by increasing this value, etc.

kozlovic:
@derekcollison Still, do you think that this PR as-is as merit? Code wise, I approved it, and want to know if we merge it.

julius-welink:
@derekcollison 
The problem with only using Jitter reconnect is that clients do not know if server is busy or not. Using a small jitter does not solve the issue, using large jitter makes it more difficult for client to reconnect under normal server load.

> client rollover during server upgrades
I do account for controlled client rollover. Not sure how you can reasonably account for a total network outage though.

julius-welink:
@kozlovic @philpennock thanks for the reviews!

PR #2579, opened 2021-09-27
Adjust cluster size on cold start in mixed mode

derekcollison:

/cc @nats-io/core

### DIFF
--- a/server/jetstream_cluster.go
+++ b/server/jetstream_cluster.go
@@ -821,6 +821,7 @@ func (js *jetStream) monitorCluster() {
 			// Check to see if we can adjust our cluster size down iff we are in mixed mode and we have
 			// seen a total that is what our original estimate was.
 			if js, total := s.trackedJetStreamServers(); js < total && total >= n.ClusterSize() {
+				s.Noticef("Adjusting JetStream expected peer set size to %d from original %d", js, n.ClusterSize())
 				n.AdjustBootClusterSize(js)
 			}
 		}
--- a/server/jetstream_cluster.go
+++ b/server/jetstream_cluster.go
@@ -153,6 +153,24 @@ const (
 	defaultMetaFSBlkSize = 1024 * 1024
 )
 
+// Returns information useful in mixed mode.
+func (s *Server) trackedJetStreamServers() (js, total int) {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	if !s.running || !s.eventsEnabled() {
+		return -1, -1
+	}
+	s.nodeToInfo.Range(func(k, v interface{}) bool {
+		si := v.(nodeInfo)
+		if si.js {
+			js++
+		}
+		total++
+		return true
+	})
+	return js, total
+}
+
 func (s *Server) getJetStreamCluster() (*jetStream, *jetStreamCluster) {
 	s.mu.Lock()
 	shutdown := s.shutdown
@@ -726,6 +744,11 @@ func (js *jetStream) monitorCluster() {
 	t := time.NewTicker(compactInterval)
 	defer t.Stop()
 
+	// Used to check cold boot cluster when possibly in mixed mode.
+	const leaderCheckInterval = time.Second
+	lt := time.NewTicker(leaderCheckInterval)
+	defer lt.Stop()
+
 	var (
 		isLeader     bool
 		lastSnap     []byte
@@ -784,6 +807,20 @@ func (js *jetStream) monitorCluster() {
 			// Periodically check the cluster size.
 			if n.Leader() {
 				js.checkClusterSize()
+			}
+		case <-lt.C:
+			s.Debugf("Checking JetStream cluster state")
+			// If we have a current leader or had one in the past we can cancel this here since the metaleader
+			// will be in charge of all peer state changes.
+			// For cold boot only.
+			if n.GroupLeader() != _EMPTY_ || n.HadPreviousLeader() {
+				lt.Stop()
+				continue
+			}
+			// If we are here we do not have a leader and we did not have a previous one, so cold start.
+			// Check to see if we can adjust our cluster size.
+			if js, total := s.trackedJetStreamServers(); js < total {
+				n.AdjustBootClusterSize(js)
 			}
 		}
 	}
--- a/server/jetstream_cluster.go
+++ b/server/jetstream_cluster.go
@@ -818,8 +818,9 @@ func (js *jetStream) monitorCluster() {
 				continue
 			}
 			// If we are here we do not have a leader and we did not have a previous one, so cold start.
-			// Check to see if we can adjust our cluster size.
-			if js, total := s.trackedJetStreamServers(); js < total {
+			// Check to see if we can adjust our cluster size down iff we are in mixed mode and we have
+			// seen a total that is what our original estimate was.
+			if js, total := s.trackedJetStreamServers(); js < total && total >= n.ClusterSize() {
 				n.AdjustBootClusterSize(js)
 			}
 		}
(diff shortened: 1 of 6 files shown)


server/jetstream_cluster.go
```diff
@@ -785,6 +808,20 @@ func (js *jetStream) monitorCluster() {
...
 				js.checkClusterSize()
 			}
+		case <-lt.C:
+			s.Debugf("Checking JetStream cluster state")
+			// If we have a current leader or had one in the past we can cancel this here since the metaleader
+			// will be in charge of all peer state changes.
+			// For cold boot only.
+			if n.GroupLeader() != _EMPTY_ || n.HadPreviousLeader() {
+				lt.Stop()
+				return
```
kozlovic, line 818: Really? This will exit monitorCluster() that is used during mirroring, so will break it. Case in point test TestNoRaceJetStreamClusterSuperClusterMirrors failing locally..

derekcollison:
I see the failure here, will take a look in a bit..

TestJetStreamClusterMetaSnapshotsAndCatchup

server/jetstream_cluster.go
derekcollison: Yep should have been continue, on it.

server/monitor.go
```diff
@@ -1142,7 +1142,7 @@ type Varz struct {
 type JetStreamVarz struct {
 	Config *JetStreamConfig `json:"config,omitempty"`
 	Stats  *JetStreamStats  `json:"stats,omitempty"`
-	Meta   *ClusterInfo     `json:"meta,omitempty"`
+	Meta   *MetaClusterInfo `json:"meta,omitempty"`
```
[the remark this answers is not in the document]
derekcollison: All remote access is fine, but direct access from the server where you pull out that exact variable and assigned to typed variable (not :=) would have an issue.

If you do have a server and pull it directly and say `jsz.Meta.Name` that still works fine.

server/monitor.go
```diff
@@ -2391,19 +2392,28 @@ type AccountDetail struct {
 	Streams []StreamDetail `json:"stream_detail,omitempty"`
 }
 
-// LeafInfo has detailed information on each remote leafnode connection.
+// MetaClusterInfo shows information about the meta group.
+type MetaClusterInfo struct {
+	Name     string      `json:"name,omitempty"`
```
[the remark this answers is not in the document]
derekcollison: Really only makes sense for meta group where unknown peers can be present.
Streams, and consequently consumers always have known peer sets, so their cluster info is always correct by default.

Felt like a waster there, hence the approach I tried to take.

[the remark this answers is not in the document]
kozlovic: I was not suggesting to add Size to ClusterInfo, but make ClusterInfo part of MetaClusterInfo instead.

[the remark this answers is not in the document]
derekcollison: So I unrolled it.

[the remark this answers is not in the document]
kozlovic: Ok.. Getting this locally:
```
--- FAIL: TestJetStreamOperatorAccounts (0.13s)
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
[signal SIGSEGV: segmentation violation code=0x1 addr=0x40 pc=0x185480d]

goroutine 57627 [running]:
testing.tRunner.func1.2(0x1ffc620, 0x27c6e80)
	/usr/local/go/src/testing/testing.go:1143 +0x49f
testing.tRunner.func1(0xc007737980)
	/usr/local/go/src/testing/testing.go:1146 +0x695
panic(0x1ffc620, 0x27c6e80)
	/usr/local/go/src/runtime/panic.go:971 +0x499
github.com/nats-io/nats-server/server.(*Server).Jsz(0xc0070cd800, 0xc00013b9e8, 0xc0070cd800, 0xc00026f000, 0x2)
	/Users/ivan/dev/go/src/github.com/nats-io/nats-server/server/monitor.go:2569 +0x6cd
github.com/nats-io/nats-server/server.TestJetStreamOperatorAccounts(0xc007737980)
	/Users/ivan/dev/go/src/github.com/nats-io/nats-server/server/jetstream_test.go:11754 +0x553
testing.tRunner(0xc007737980, 0x2169488)
	/usr/local/go/src/testing/testing.go:1193 +0x203
created by testing.(*T).Run
	/usr/local/go/src/testing/testing.go:1238 +0x5d8
FAIL	github.com/nats-io/nats-server/server	283.625s
```

kozlovic (commented):
Ok with the approach, but some tests need fixing.

server/jetstream_cluster.go
```diff
@@ -153,6 +153,24 @@ const (
...
+	}
+	s.nodeToInfo.Range(func(k, v interface{}) bool {
+		si := v.(nodeInfo)
+		if si.js {
+			js++
+		}
+		total++
+		return true
+	})
+	return js, total
```
matthiashanel, line 171: Does this contain nodes we have seen only?
meaning total may not be the value started with as some servers have not been observed?

derekcollison: Correct, only nodes we have seen and update for are counted here so we can not go down based on split brain scenario on cold boot.

server/monitor.go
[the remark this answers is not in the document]
derekcollison: I think this is fine but interested in your thoughts.

server/monitor.go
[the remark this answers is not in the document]
derekcollison: All tests pass now.

matthiashanel (approved):
This would break down with seed server, but that is as far as as I understand it not supported right now either.
LGTM

derekcollison:
I have done a seed cluster, and that would work ok with this if you made that mixed mode on cold boot.

server/jetstream_cluster.go
```diff
@@ -785,6 +808,21 @@ func (js *jetStream) monitorCluster() {
...
+			// will be in charge of all peer state changes.
+			// For cold boot only.
+			if n.GroupLeader() != _EMPTY_ || n.HadPreviousLeader() {
+				lt.Stop()
+				continue
+			}
+			// If we are here we do not have a leader and we did not have a previous one, so cold start.
+			// Check to see if we can adjust our cluster size down iff we are in mixed mode and we have
+			// seen a total that is what our original estimate was.
+			if js, total := s.trackedJetStreamServers(); js < total && total >= n.ClusterSize() {
```
ripienaar, line 823: I think given that this usually is a weird situation, and boot only, we need to be logging here at a level visible by default

kozlovic (approved):
LGTM

server/jetstream_cluster.go
derekcollison: Agree, which level though? Just info?

ripienaar: warn if we have such a level dont recall now

derekcollison: I might suggest INFO level, since its probably a mixed mode setup by design. WDYT?

ripienaar: Doesn’t it essentially only happen if the system is badly balanced? Potentially badly deployed? Not a normal situation?

but whatever, any log wil do given how little we have so whatever you wish.

derekcollison: Originally I suggested that but if you recall we had a few folks chime in that that may be a normal setup where storage is only in a couple of clusters vs everywhere, so that could be a valid setup.

derekcollison: ok pushed, pls check real quick.

kozlovic (approved):
LGTM

matthiashanel (approved):
LGTM

PR #2580, opened 2021-09-28
[fixed] missing support for header in some request handler

matthiashanel:

this is what already happens in `zreq`
Basically the message contains header, and they need to be striped before unmarshal.


kozlovic (approved):
LGTM

PR #2581, opened 2021-09-28
added unit test to test stream source/mirror cross different domains

matthiashanel:

Added this unit test to test something and it worked.
Committing it as we don't have a test where source/mirror are not neighbors.


kozlovic (approved):
LGTM

PR #2582, opened 2021-09-28
Update nats-io/jwt dependency

wallyqs:
Updates go modules and vendoring of the nats-server to avoid referencing nats-io/jwt/v1. 

Procedure:

```sh
go get github.com/nats-io/jwt/v2@v2.1.0
go mod tidy
go mod vendor
```

/cc @nats-io/core


kozlovic (approved):
LGTM

PR #2583, opened 2021-09-29
fixes the msg get audit event

ripienaar:
We resused msg variable for both the request
and later the data loaded from the stream.

The data thus replaces the request and the audit
had the data as both request and response

/cc @nats-io/core


derekcollison (approved):
LGTM

PR #2585, opened 2021-09-30
Rollup functionality.

derekcollison:
Messages published with the rollup header will place the new message onto the stream and purge all others based on subject, or the complete stream.

/cc @nats-io/core

### DIFF
--- a/server/stream.go
+++ b/server/stream.go
@@ -221,6 +221,13 @@ const (
 	JSLastConsumerSeq     = "Nats-Last-Consumer"
 	JSLastStreamSeq       = "Nats-Last-Stream"
 	JSConsumerStalled     = "Nats-Consumer-Stalled"
+	JSMsgRollup           = "Nats-Rollup"
+)
+
+// Rollups, can be subject only or all messages.
+const (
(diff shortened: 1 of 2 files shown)


ripienaar (approved):
very cool, LGTM

server/stream.go
```diff
@@ -2931,6 +2960,11 @@ func (mset *stream) processJetStreamMsg(subject, reply string, hdr, msg []byte,
 		if msgId != _EMPTY_ {
 			mset.storeMsgId(&ddentry{msgId, seq, ts})
 		}
+		if rollupSub {
+			mset.purge(&JSApiStreamPurgeRequest{Subject: subject, Keep: 1})
+		} else if rollupAll {
+			mset.purge(&JSApiStreamPurgeRequest{Keep: 1})
```
ripienaar, line 2966: should we error for rollup subject with unsupported value?

derekcollison: We do above.

ripienaar: doh, missed that!

tpihl:
Looks very interesting

PR #2595, opened 2021-10-06
In addition to sealed we add in other stream perms to control purge, msg deletes and rollups.

derekcollison:

/cc @nats-io/core


ripienaar (approved):
LGTM

PR #2623, opened 2021-10-14
[FIXED] #2622

### LINKED ISSUE #2622: Double put of large objects results in loss metadata
With the new object store preview a put of a large object followed by another put of the same name will incorrectly discard the metadata from the first put.

derekcollison:
We were not escaping the top level iterator across message blocks when calculating when to break due to keep > 0.

Resolves #2622

/cc @nats-io/core


ripienaar:
Hah missed that. All good.

PR #2624, opened 2021-10-14
Adding tracing facility

matthiashanel:
Adds msg_trace (trace rule name to trace) to operator and accounts.
    enables tracing for operator and accounts.
Adds ping_probes option (array of probe names) to accounts,
    enabling header based tracing.
Adds Nats-Ping header and ping_probes option (array of probe names) to
    accounts, enabling tracing by Nats-Ping header.

Nats-Ping header can be empty (default) or contain a list of probe names
When set, the message will not be handed to subscriber/jetstream/other
accounts.
Optionally the message body can contain a list of server ids to filter
responding servers on.

Still missing: handling responses, jwt integration & dynamic tracing
(settable via system account API). Settle on what to use as trigger.

This basically works by having a single trace context per receive loop.
With this context we call tracing functions for pub/sub/store.
After handling of an incoming message is done we call complete to send
traces.

### DIFF
--- a/server/accounts.go
+++ b/server/accounts.go
@@ -90,6 +90,10 @@ type Account struct {
 	tags         jwt.TagList
 	nameTag      string
 	lastLimErr   int64
+
+	// tracing policies
+	traces     map[string]Trace
+	pingProbes *probes
 }
 
 // Account based limits.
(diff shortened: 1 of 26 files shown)


bwerthmann:
Fixes nats-io/nats-general#72

bruth:
@derekcollison @aricart Not sure if this was discussed offline as desirable, but this PR has quite a number of conflicts, so closing now. We can revisit if the general concept should be revisited.

PR #2660, opened 2021-11-01
[IMPROVED] Websocket: Add client IP in websocket upgrade failures

kozlovic:
The error message would now look like this:
```
[8672] 2021/11/01 10:56:50.251985 [ERR] [::1]:59279 - websocket handshake error: invalid value for header 'Upgrade'
```

(without this change the part `[::1]:59279 - ` would not be present)


derekcollison (approved):
LGTM

PR #2665, opened 2021-11-02
[FIXED] Messages not expiring properly after server restart #2662.

### LINKED ISSUE #2662: JetStream messages that have large expirations appear to not expire at the correct time after server restart.
Server restart sets the expire timer for each stream after messages have been expired to the default.

derekcollison:
Upon server restart a server would set the check expiration to the configured amount vs delta of next to expire.

Resolves #2662 

/cc @nats-io/core

### DIFF
--- a/server/filestore.go
+++ b/server/filestore.go
@@ -910,8 +910,9 @@ func (fs *fileStore) recoverMsgs() error {
 
 // Will expire msgs that have aged out on restart.
 // We will treat this differently in case we have a recovery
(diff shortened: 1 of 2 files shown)


derekcollison:
I believe this was introduced by me here: https://github.com/nats-io/nats-server/commit/4e92b0ed6eb095c8b5c61f5fcba5a8856f874981

kozlovic (approved):
LGTM, and thanks for the PR's title, will be helpful when writing the release notes!

PR #2671, opened 2021-11-03
Allow system account to respond with jetstream not enabled.

derekcollison:
Makes interactions with nats cli a bit better.

/cc @nats-io/core


ripienaar (approved):
LGTM

PR #2672, opened 2021-11-03
Altered jwt tests to not use localhost and prefixed tests with TestJWT

matthiashanel:


kozlovic (approved):
LGTM

PR #2675, opened 2021-11-04
Add in max_sub_tokens support

derekcollison:

/cc @nats-io/core

### DIFF
--- a/server/client.go
+++ b/server/client.go
@@ -2418,6 +2418,14 @@ func (c *client) processSubEx(subject, queue, bsid []byte, cb msgHandler, noForw
 			c.subPermissionViolation(sub)
 			return nil, ErrSubscribePermissionViolation
 		}
+
+		if opts := srv.getOpts(); opts != nil && opts.MaxSubTokens > 0 {
+			if len(bytes.Split(sub.subject, []byte(tsep))) > int(opts.MaxSubTokens) {
+				c.mu.Unlock()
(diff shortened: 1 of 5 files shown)


server/client.go
```diff
@@ -2418,6 +2418,14 @@ func (c *client) processSubEx(subject, queue, bsid []byte, cb msgHandler, noForw
 			c.subPermissionViolation(sub)
 			return nil, ErrSubscribePermissionViolation
 		}
+
+		if maxSubTokens := int(c.srv.getOpts().MaxSubTokens); maxSubTokens > 0 {
```
kozlovic, line 2422: `srv` has already been captured as `c.srv`, but no biggy.

kozlovic (changes requested):
I saw fix for reload, but that's not really complete: if we support, then we need to add new object and report that it has been changed during reload, and decide what we do with existing subs? Or fail the reload if the value is changed?

server/client.go
derekcollison: fixed

derekcollison:
I think on reload it just enforces going forward, no changes imo. The ones that failed before of course aren't there and the ones that succeeded but may have failed with the new restriction I think we leave them.

kozlovic (approved):
LGTM - But it means that the server is not going to print that it has reloaded (and accepted) the new value, but that is not crucial I guess.

kozlovic:
That's fine, we can add later.

PR #2683, opened 2021-11-09
fixed unit test where after a server shutdown no quorum existed

matthiashanel:

Ran this test 80 times locally. passed every single time


derekcollison (approved):
LGTM

PR #2684, opened 2021-11-10
[FIXED] A slow consumer could cause the publisher to block

### LINKED ISSUE #2679: A slow/disconnected subscriber can cause publishing to become blocked
## Defect

With 1 publisher and multiple subscribers I have observed that under certain circumstances a subscriber 
which is slow/has a poor connection can block publishing (so preventing other subscribers from receiving data).

### Reproduction

It is possible to reproduce this with a local `nats-server` and `nats-cli`.

 - Run `nats-server` (default config/no arguments)
 - Start publishing messages ever 20ms of ~20kb: `./nats --trace pub test --count=-10000000 --sleep=20ms "{{Random 20000 20000}}"`
 - Start subscribing (just print out the message number to make the drops obvious) `./nats sub test | grep Received`
 - Start a second subscriber, `./nats sub test | grep Received`. Cause this subscriber to simulate a disconnect by suspending the process (i.e. pressing `CTRL+Z` in the terminal.)

After a few seconds, at the same time, the first subscriber stops receiving messages, the publisher stops sending (message count stops increasing). This lasts for  about 10 seconds.

I also see the following in the nats console logs.

```
[129148] 2021/11/05 11:45:56.973798 [INF] 127.0.0.1:50642 - cid:12 - "v1.12.0:go:NATS CLI Version 0.0.26" - Slow Consumer Detected: WriteDeadline of 10s exceeded with 1 chunks of 20020 total bytes.
[129148] 2021/11/05 11:45:56.973825 [WRN] 127.0.0.1:50616 - cid:10 - "v1.12.0:go:NATS CLI Version 0.0.26" - Readloop processing time: 10.000846205s
```

I've tested this on Ubuntu 20.04 with `nats-server` 2.6.4 and `nats-cli` 0.0.26.

The largish message size is required to trigger the issue.

Suspending the nats-cli is a crude way to simulate a subscriber with a connection issue  - but I have seen this happens in more realistic/reasonable situations (standard slow subscribers/poor network conditions).

### Investigation/Potential Fix

I am not familiar with golang or the nat-server codebase- but here's my attempt at figure out what is going on.

Using a shorter write deadline option makes the drop/blockage shorter - but there are still gaps of write deadline length when the subscriber disconnects.

I think the client `readLoop` for the publisher calling `flushClientsWithCheck` is causing the issue. When flushing messages to the subscriber the publisher `readLoop` ends up being blocked for the duration of `WriteDeadline`. The attempt to avoid this by using a time budget does not work since the heuristic of using the `Last flush time for Write` to guess how long the flush will take fails under the disconnect condition.

I verified this by changing the `budget` that `flushClientsWithCheck` is called with to zero in `client.go` - so clients are only notified, not flushed. i.e: 

```
@@ -1204,9 +1204,9 @@ func (c *client) readLoop(pre []byte) {
                // if applicable. Routes and Gateways will never
                // spend time flushing outbound in place.
                var budget time.Duration
-               if c.kind == CLIENT {
-                       budget = time.Millisecond
-               }
+               //if c.kind == CLIENT {
+               //      budget = time.Millisecond
+               //}
 
                // Flush, or signal to writeLoop to flush to socket.
                last := c.flushClientsWithCheck(budget, true)
```
(This might have performance implications but is the simplest way to fix it - assuming my reasoning here is correct.)

After I made this change I cannot reproduce the issue. When suspending a subscriber we see the `Slow consumer` log - but no slow `Readloop processing` warning.

kozlovic:
The server reads data from a client from a go routine. When receiving
messages, it checks for matching subscriptions, and if found, would
send those messages from the producer's readLoop.
A notion of "budget" was used to make sure the server does not spend
too much time sending to clients from the producer's readLoop, however,
regardless of how small the budget was, if one of the subscription's
connection TCP buffer was full, a TCP write would block for as long
as the defined write_deadline (which is now 10 seconds).

We are removing this behavior and therefore clients (like it was the
case for other type of connections) will now always notify the
subscriber's writeLoop that data is ready to be sent, but the send
will not occur in the producer's writeLoop.

Resolves #2679

### DIFF
--- a/server/client.go
+++ b/server/client.go
@@ -1026,17 +1026,6 @@ func (c *client) writeLoop() {
 // sent to during processing. We pass in a budget as a time.Duration
 // for how much time to spend in place flushing for this client.
 func (c *client) flushClients(budget time.Duration) time.Time {
-	return c.flushClientsWithCheck(budget, false)
-}
-
-// flushClientsWithCheck will make sure to flush any clients we may have
-// sent to during processing. We pass in a budget as a time.Duration
-// for how much time to spend in place flushing for this client.
-// The 'clientsKindOnly' boolean indicates whether to check kind of client
-// and pending client to run flushOutbound in flushClientsWithCheck.
-// flushOutbound() could block the caller up to the write deadline when
-// the receiving client cannot drain data from the socket fast enough.
-func (c *client) flushClientsWithCheck(budget time.Duration, clientsKindOnly bool) time.Time {
 	last := time.Now().UTC()
 
 	// Check pending clients for flush.
@@ -1057,7 +1046,7 @@ func (c *client) flushClientsWithCheck(budget time.Duration, clientsKindOnly boo
 			continue
 		}
 
-		if budget > 0 && (!clientsKindOnly || c.kind == CLIENT && cp.kind == CLIENT) && cp.out.lft < 2*budget && cp.flushOutbound() {
+		if budget > 0 && cp.out.lft < 2*budget && cp.flushOutbound() {
 			budget -= cp.out.lft
 		} else {
 			cp.flushSignal()
@@ -1199,17 +1188,8 @@ func (c *client) readLoop(pre []byte) {
 			atomic.AddInt64(&s.inBytes, int64(c.in.bytes))
 		}
 
-		// Budget to spend in place flushing outbound data.
-		// Client will be checked on several fronts to see
-		// if applicable. Routes and Gateways will never
-		// spend time flushing outbound in place.
-		var budget time.Duration
-		if c.kind == CLIENT {
-			budget = time.Millisecond
-		}
-
-		// Flush, or signal to writeLoop to flush to socket.
-		last := c.flushClientsWithCheck(budget, true)
+		// Signal to writeLoop to flush to socket.
+		last := c.flushClients(0)
 
 		// Update activity, check read buffer size.
 		c.mu.Lock()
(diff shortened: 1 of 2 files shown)


kozlovic:
@derekcollison I did not completely remove the concept of "budget" since we have a case in the event loop where we want to flush in place (but this is in a "send loop") the last message posted.

Also, although benchmark did not reveal adverse effects, I do believe that this may - in some cases - increase the number of slow consumers by never slowing down the producer (that is, not reading producer's messages and therefore applying back pressure). We will see...

server/client.go
```diff
@@ -1057,7 +1046,7 @@ func (c *client) flushClientsWithCheck(budget time.Duration, clientsKindOnly boo
 			continue
 		}
 
-		if budget > 0 && (!clientsKindOnly || c.kind == CLIENT && cp.kind == CLIENT) && cp.out.lft < 2*budget && cp.flushOutbound() {
+		if budget > 0 && cp.out.lft < 2*budget && cp.flushOutbound() {
```
derekcollison, line 1049: Should we remove budget arg all together here?

kozlovic: Not sure you saw my earlier comment here: https://github.com/nats-io/nats-server/pull/2684#issuecomment-964668827.
There is a case in the server that we would want to flush in place if possible: https://github.com/nats-io/nats-server/blob/91d6531b8bdbcee0fdc1d2a4810712cdf6f76a7b/server/events.go#L411

derekcollison (approved):
LGTM

PR #2688, opened 2021-11-13
Fix for latency tracking bug.

derekcollison:
The bug occurs when latency tracking is on, a requestor and responder are not connected to the same server, and the responder sends two responses for a single request.

/cc @nats-io/core


matthiashanel (approved):
LGTM

PR #2690, opened 2021-11-16
Travis updates

kozlovic:
- Add Go 1.17
- Fix go fmt from Go 1.17 (build directives)
- Download version of misspell and staticcheck instead of doing
"go get" since current staticcheck would be broken without go.mod

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,8 +17,8 @@ env:
 - GO111MODULE=off
 go_import_path: github.com/nats-io/nats-server
 install:
-- GO111MODULE=on go get honnef.co/go/tools/cmd/staticcheck@2021.1.2
-- GO111MODULE=on go get github.com/client9/misspell/cmd/misspell@v0.3.4
+- GO111MODULE=on go install honnef.co/go/tools/cmd/staticcheck@latest
+- GO111MODULE=on go install github.com/client9/misspell/cmd/misspell@latest
 before_script:
 - GO_LIST=$(go list ./...)
 - go build
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,8 +7,8 @@ vm:
 
 language: go
 go:
+- 1.17.x
 - 1.16.x
-- 1.15.x
 addons:
   apt:
     packages:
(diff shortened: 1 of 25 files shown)


.travis.yml
```diff
@@ -17,11 +17,8 @@ env:
 - GO111MODULE=off
 go_import_path: github.com/nats-io/nats-server
 install:
-- go get github.com/nats-io/nats.go/
-- go get github.com/nats-io/nkeys
-- go get github.com/nats-io/jwt
-- go get -u honnef.co/go/tools/cmd/staticcheck
-- go get -u github.com/client9/misspell/cmd/misspell
+- GO111MODULE=on go get honnef.co/go/tools/cmd/staticcheck@2021.1.2
```
derekcollison, line 20: Why do we need pinned versions of these again?

wallyqs (approved):
LGTM

server/disk_avail.go
```diff
@@ -11,8 +11,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// +build !windows
-// +build !openbsd
+//go:build !windows && !openbsd
+// +build !windows,!openbsd
```
derekcollison, line 15: We need both?

wallyqs: These only apply to [Go 1.17](https://golang.org/doc/go1.17),  we will need both to avoid `go fmt` failing on Go 1.17 and to support Go 1.16 which does not recognize the new syntax.

.travis.yml
kozlovic: So at first, when we were doing `go get -u`, staticcheck would fail because one of its dependency has a breaking change, so that made the whole thing fail with error like that:
```
# honnef.co/go/tools/go/loader
../../../honnef.co/go/tools/go/loader/loader.go:330:48: err.Line undefined (type config.ParseError has no field or method Line)
```
So I was looking at pinning to a version, but after thinking about it, if we do the `GO111MODULE on go get`, then it should be fine to not pin to a version, since this is using go.mod and by definition using pinned dependencies (assuming that each tool's main is not simply broken). So I will update the PR to remove the version.

Note, staticcheck's dependency resolved the issue yesterday (https://github.com/BurntSushi/toml/commit/dcab1c6c8e42cfd8d4ebc085554815678d7b1a2b), but still, using `GO111MODULE=on go get` seem to be the way to go, and allow us to pin to a version if one day something break in their main branch, so we don't have to wait for them to fix.

kozlovic:
@derekcollison I have removed the pin to a specific version, instead, I use `go install` but provide `@latest` as instructed by Go 1.17: https://github.com/nats-io/nats-server/pull/2690/commits/aaf6d82c10bff2c0048e9020f1399da5fcda30b9.

.travis.yml
derekcollison: With 1.17 we can remove vendor directory yes?

derekcollison (approved):
LGTM

.travis.yml
kozlovic: Goal is at one point to do it. But need a bit more experimentation. I recall an issue with pulling tool deps (staticcheck/coverall, etc..) that would prevent running the project/tests with go mods. Will do that in another branch.

PR #2692, opened 2021-11-16
Add support for designated OCSP responders

variadico:
Currently, OCSP responses that aren't signed by the root CA will fail
with a verification error. This change allows intermediates, or
designated responders, to sign responses on behalf of the CA.


wallyqs (approved):

philpennock (approved):

server/ocsp.go
```diff
@@ -446,13 +446,33 @@ func (srv *Server) NewOCSPMonitor(config *tlsConfigKind) (*tls.Config, *OCSPMoni
...
+				leaf := chain[0]
+				parent := issuers[len(issuers)-1]
+
+				resp, err := ocsp.ParseResponseForCert(oresp, leaf, parent)
 				if err != nil {
 					return fmt.Errorf("failed to parse OCSP response from %s client: %w", kind, err)
 				}
-				if err := resp.CheckSignatureFrom(issuer); err != nil {
-					return err
+				if resp.Certificate == nil {
```
wallyqs, line 458: Reference on why need to check here that it is `nil`: https://github.com/nats-io/nats-server/blob/91d6531b8bdbcee0fdc1d2a4810712cdf6f76a7b/vendor/golang.org/x/crypto/ocsp/ocsp.go#L401-L406
```
// CheckSignatureFrom checks that the signature in resp is a valid signature
// from issuer. This should only be used if resp.Certificate is nil. Otherwise,
// the OCSP response contained an intermediate certificate that created the
// signature. That signature is checked by ParseResponse and only
// resp.Certificate remains to be validated.
```

kozlovic (approved):
LGTM

PR #2693, opened 2021-11-17
Js leaf deny

matthiashanel:
From now on JS API traffic will be denied across leaf nodes, unless the system account is shared and the availability domain is identical. 
The main reason we start to enforce this is that enabling jetstream in an availability domain would alter the api traffic flow.
For example, a leaf node without jetstream was able to communicate with a JS in the hub without using it's domain name.
If for an unrelated reason jetstream had to be enabled in a the leaf node, this default traffic would all of a sudden go some place else (remain local) where before it'd go to a hub.

To avoid this from the get go we suggest setting the domain in client applications as soon as you need to communicate with a JS other than the local one. 
If domain names are already used in such a way, there is nothing to do.

We do also recommend setting domain names for each availability zone. 
(Availability zones can only change when crossing leaf nodes. If the leaf node share the system account and specify the same domain name, they are considered within the same availability zone. When this is done, meta group leadership is pinned to server that accept leaf node connections. This is to avoid extra hops.)

If you are currently not using domain names and rely on the old behavior, the server contains a new option `default_js_domain` that contains a map from account name to default domain name. 
```
default_js_domain:{
 ACC1:domain,
 ACC2:domain,
}
```
In case the JetStream you want to reach does not have a domain name set just specify `""` as domain name.

Please be aware that if you are currently not using domain names, and have to specify the empty domain name, in order to connect accross leaf nodes both ends have to be configured in such a way. This is because with this change, each server is going to make it's individual decission about which traffic to deny. For example. If you have a hub without jetstream and a leaf node that offers JS. 
Until you enabled a domain name in the leaf node and change your clients to provide this name, you have to set 
`default_js_domain:{your-account-in-hub:""}` in the hub and `default_js_domain:{your-account-in-leaf:""}`
This will instruct both server to not deny default api traffic.

Alternatively you could also restart your leaf node and provide a domain name. Then, in the hub set `default_js_domain:{your-account-in-hub:"leaf-domain-name"}`

In mixed mode setups (rare use case) you will have to set the domain name in the non JS enabled server as well.
Basically telling them, to which availability domain they belong.

```
jetstream: {
	enabled: false
	domain: domain-this-server-belongs
}
```

In the rare case where a JS availability domain is extended via a non clustered leaf node, you need to explicitly instruct the server to be in clustered mode with this new JetStream config option `extension_hint: will_extend`.

There are two more situations where the extension_hint can be used, but the server will print when this option is applicable. Primarily they are there to assist eh server during initial startup until it was able to decide if it's leaf node connection was part of the same or a different availability domain.


ripienaar:
This ```default_js_domain:{
 ACC1:domain,
 ACC2:domain,
}``` seems really weird, we have nowhere this kind of structure in the configuration, seems weird to introduce a new style of flag.

why can default_js_domain not be set on the account?

server/mqtt_test.go
```diff
@@ -216,16 +216,14 @@ func testMQTTDefaultOptions() *Options {
 
 func testMQTTRunServer(t testing.TB, o *Options) *Server {
 	t.Helper()
-	o.NoLog = false
+	o.NoLog = true
```
kozlovic, line 219: This was on purpose that I would set logging (NoLog=false) and a dummy logger below. The idea is to have better code coverage and test of debug/trace statements. Any reason you have removed this?

server/mqtt.go
```diff
@@ -243,14 +243,16 @@ type sessPersistRecord struct {
...
-	quitCh  chan struct{}
+	mu        sync.Mutex
+	id        string
+	c         *client
+	sendq     chan *mqttJSPubMsg
+	rplyr     string
+	replies   sync.Map
+	nuid      *nuid.NUID
+	quitCh    chan struct{}
+	domain    string // Domain or possibly empty. This is added to session subject.
```
kozlovic, line 254: Granted it would probably be more work, but I thought we discussed that we would actually store a prefix here (either `$JS.API` or actual `$JS.<domain>.API`) and add this prefix to all API subjects. We would have to replace existing references to the constants such as `JSApiStreamCreateT` to something like `mqttAPIStreamCreateT` that would be defined as the same than JSApiStreamCreateT but without the prefix, in this case `%s.STREAM.CREATE.%s`.  So in place where we currently have:
```
jsa.newRequest(mqttJSAStreamCreate, fmt.Sprintf(JSApiStreamCreateT, cfg.Name), 0, cfgb)
```
we would do:
```
jsa.newRequest(mqttJSAStreamCreate, fmt.Sprintf(mqttAPIStreamCreateT, jsa.apiPfx, cfg.Name), 0, cfgb)
```
or something similar.

I agree that the places impacted currently should be "low volume", but still, that makes use of more sprintf/memory allocs if you trim the prefix and create the actual prefix in each requests (in newRequestEx).

We could go with your approach first and improve later if you think it is best to minimize changes in this PR.

kozlovic (commented):
Just some comments on the MQTT code/tests for now.

server/mqtt_test.go
matthiashanel: Since I find myself changing this value every time I need to debug a unit test, I instinctively changed that. Will undo this change.

server/mqtt.go
matthiashanel: I will submit a separate PR once this one is oked.

server/server.go
```diff
@@ -1319,7 +1319,24 @@ func (s *Server) registerAccountNoLock(acc *Account) *Account {
 	}
 	acc.srv = s
 	acc.updated = time.Now().UTC()
+	accName := acc.Name
+	jsEnabled := acc.jsLimits != nil
 	acc.mu.Unlock()
+
+	if s.opts != nil && len(s.opts.JsAccDefaultDomain) > 0 {
```
kozlovic, line 1326: capture `opts:=s.getOpts()` and use `opts` instead of `s.opts`.

server/mqtt_test.go
```diff
@@ -3115,88 +3115,149 @@ func TestMQTTClusterPlacement(t *testing.T) {
...
+		// Use RoutesFromStr() to make an array of urls
+		urls := RoutesFromStr(fmt.Sprintf("nats://127.0.0.1:%d,nats://127.0.0.1:%d,nats://127.0.0.1:%d",
+			o1.LeafNode.Port, o2.LeafNode.Port, o3.LeafNode.Port))
+		lno.LeafNode.Remotes = []*RemoteLeafOpts{{URLs: urls}}
 
-	// Connect from one server in the cluster check it works from there too.
-	connectAndPublish(o3)
-	testMQTTCheckPubMsg(t, mc, rc, "foo", mqttPubQos1, []byte("msg"))
+		ln := RunServer(lno)
+		defer ln.Shutdown()
```
kozlovic, line 3202: Replace with `defer testMQTTShutdownServer(ln)`

kozlovic (changes requested):

server/leafnode_test.go
```diff
@@ -4072,3 +4099,942 @@ func TestLeafNodeInterestPropagationDaisychain(t *testing.T) {
...
+			strmCfg := func(name, placementCluster string) *nats.StreamConfig {
+				if placementCluster == "" {
+					return &nats.StreamConfig{Name: name, Replicas: 1, Subjects: []string{name}}
+				}
+				return &nats.StreamConfig{Name: name, Replicas: 1, Subjects: []string{name},
+					Placement: &nats.Placement{Cluster: placementCluster}}
+			}
+			// Only after the system account is fully connected can streams be placed anywhere.
+			testJSFunctions := func(pass bool) {
+				ncA := natsConnect(t, fmt.Sprintf("nats://a1:a1@127.0.0.1:%d", sA.opts.Port))
```
kozlovic, line 4313: I think `defer ncA.Close()` is missing here (after this call).

server/leafnode_test.go
```diff
@@ -4072,3 +4099,942 @@ func TestLeafNodeInterestPropagationDaisychain(t *testing.T) {
...
+
+			if testCase.leafFunctionPreJoin {
+				cl := cluster{t: t, servers: []*Server{sLA, sLB}}
+				cl.waitOnLeader()
+				cl.waitOnPeerCount(2)
+				testJSFunctions(false)
+			} else {
+				// In cases where the leaf nodes have to wait for the system account to connect,
+				// JetStream should not be operational during that time
+				ncA := natsConnect(t, fmt.Sprintf("nats://a1:a1@127.0.0.1:%d", sLA.opts.Port))
```
kozlovic, line 4367: I think `defer ncA.Close()` is missing here.

server/leafnode_test.go
```diff
@@ -4072,3 +4099,942 @@ func TestLeafNodeInterestPropagationDaisychain(t *testing.T) {
...
+				cnt++
+				if placementCluster == "" {
+					return &nats.StreamConfig{Name: name, Replicas: 1, Subjects: []string{name}}
+				}
+				return &nats.StreamConfig{Name: name, Replicas: 1, Subjects: []string{name},
+					Placement: &nats.Placement{Cluster: placementCluster}}
+			}
+
+			test := func(port int, expectedDefPlacement string) {
+				ncA := natsConnect(t, fmt.Sprintf("nats://a1:a1@127.0.0.1:%d", port))
```
kozlovic, line 4566: I think `defer ncA.Close()` is missing here.

server/leafnode_test.go
```diff
@@ -4072,3 +4099,942 @@ func TestLeafNodeInterestPropagationDaisychain(t *testing.T) {
...
+
+		lnCnt := 1
+		if sysShared {
+			lnCnt++
+		}
+
+		checkLeafNodeConnectedCount(t, sHub, lnCnt)
+		checkLeafNodeConnectedCount(t, sLeaf, lnCnt)
+
+		ncA := natsConnect(t, fmt.Sprintf("nats://a1:a1@127.0.0.1:%d", sHub.opts.Port))
```
kozlovic, line 4719: I think `defer ncA.Close()` is missing here.

server/leafnode_test.go
```diff
@@ -4072,3 +4099,942 @@ func TestLeafNodeInterestPropagationDaisychain(t *testing.T) {
...
+			sysCreds,
+			noDomainFix)))
+		defer removeFile(t, confL)
+		sLeaf, _ := RunServerWithConfig(confL)
+		defer sLeaf.Shutdown()
+
+		checkLeafNodeConnectedCount(t, sHub, 2)
+		checkLeafNodeConnectedCount(t, sLeaf, 2)
+
+		ncA := natsConnect(t, fmt.Sprintf("nats://127.0.0.1:%d", sHub.opts.Port), createUserCreds(t, nil, aKp))
```
kozlovic, line 4879: I think `defer ncA.Close()` is missing here.

server/leafnode_test.go
```diff
@@ -4072,3 +4099,942 @@ func TestLeafNodeInterestPropagationDaisychain(t *testing.T) {
...
+
+	checkClusterFormed(t, sLeaf1, sLeaf2)
+	c2 := cluster{t: t, servers: []*Server{sLeaf1, sLeaf2}}
+	c2.waitOnPeerCount(2)
+
+	checkLeafNodeConnectedCount(t, sHub1, 4)
+	checkLeafNodeConnectedCount(t, sLeaf1, 2)
+	checkLeafNodeConnectedCount(t, sLeaf2, 2)
+
+	ncB := natsConnect(t, fmt.Sprintf("nats://b1:b1@127.0.0.1:%d", sLeaf1.getOpts().Port))
```
kozlovic, line 5027: I think `defer ncB.Close()` is missing here.

server/leafnode_test.go
matthiashanel: fixed (added for ncL below as well)

server/leafnode_test.go
matthiashanel: fixed

server/leafnode_test.go
matthiashanel: fixed

server/leafnode_test.go
matthiashanel: fixed

server/leafnode_test.go
matthiashanel: fixed

server/leafnode_test.go
matthiashanel: fixed

kozlovic (approved):
LGTM

matthiashanel:
Early releases of the Jetstream feature attempted to "do the right thing" in a variety of circumstances, but the logic used proved less resilient to certain changes, allowing some actions-at-a-distance to break previously functioning deployments. We're updating the NATS server logic to provide a more resilient model.

Our advice for developers is that:
 1. You should always set a Jetstream Domain in the server
 2. If you only ever want to talk to Jetstream in the same server/cluster to which you connect, then the clients do not need to specify a Jetstream Domain
 3. In all other cases, the client should set the Jetstream Domain explicitly

Note in particular that a deployment model with a backbone to which leafnodes or leafnode clusters connect needs particular care in enabling Jetstream in the backbone: if clients are not setting a Jetstream Domain and expecting to talk to the one leafnode they have in their account, such an upgrade can break working configurations.  Administrators can use `default_js_domain`/`mqtt.js_domain` to provide mappings to assist in keeping such clients functioning across a transition. However, we recommend that this be considered a transient upgrade assistance mode, not a long term configuration.

We added `js_domain` to permanently support mqtt running on the non JetStream enabled end of a leaf node configuration.
As you upgrade these servers, we recommend specifying the domain for mqtt to use, absent a local one.
Regular nats clients need to specify the nats.Domain option. 
For `mqtt`, `js_domain` is the way to configure that.

For example:
```
mqtt: {
    js_domain: target-domain
}
```

Updating clients with a new domain setting may not be as quick or easy as updating nats-server.
That is why we added the `default_js_domain`. It is a map from account name (or account nkey) to domain (or "" when none was set) to use for client's that don't set a domain.
To avoid outages we recommend first upgrading and re-configuring the server in which accounts and clusters do not have JetStream enabled.
In ngs, if a leaf node enabled JetStream for clients connected to ngs, the domain name is now required.
Existing ngs customers are not affected, and only have to care about upgrading their leaf nodes and connected clients.

For example:
```
default_js_domain:{
 ACC1:other-domain-name,
 ACC2:my-default-js-traffic-goes-to-this-domain,
 ACC3:""
}
```

You should only add to `default_js_domain` accounts that do not have Jetstream enabled.
Once you have changed all your clients to either use the local JetStream or provide the appropriate Domain name and mqtt has `js_domain` set where needed, you can remove `default_js_domain` from your configurations. 

In the rare case where a JetStream availability domain is extended via a NON-clustered leaf node, you need to explicitly instruct the server to be in clustered mode with the new extension_hint JetStream configuration option. 
All potential usages of extension_hint are related to NON-clustered server. Any other potential usages to speed up the startup process will be printed on server startup.
Possible values to set are: will_extend, no_extend

For example:
```
jetstream {
    extension_hint: will_extend
}
```

If your cluster consists of a mix of servers with and without JetStream enabled, servers without JetStream enabled need to set the domain name to the same value as servers with JetStream enabled.

```
jetstream: {
	enabled: false
	domain: name
}
```

PR #2695, opened 2021-11-17
allow streams api to be filtered like list api

ripienaar:

/cc @nats-io/core

### DIFF
--- a/server/jetstream_cluster.go
+++ b/server/jetstream_cluster.go
@@ -3836,7 +3836,7 @@ func (s *Server) allPeersOffline(rg *raftGroup) bool {
 
 // This will do a scatter and gather operation for all streams for this account. This is only called from metadata leader.
 // This will be running in a separate Go routine.
-func (s *Server) jsClusteredStreamListRequest(acc *Account, ci *ClientInfo, offset int, subject, reply string, rmsg []byte) {
+func (s *Server) jsClusteredStreamListRequest(acc *Account, ci *ClientInfo, filter string, offset int, subject, reply string, rmsg []byte) {
 	defer s.grWG.Done()
 
 	js, cc := s.getJetStreamCluster()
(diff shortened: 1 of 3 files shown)


server/jetstream_cluster_test.go
```diff
@@ -9442,6 +9442,52 @@ func TestJetStreamClusterAccountInfoForSystemAccount(t *testing.T) {
 	}
 }
 
+func TestJetClusterStreamListFilter(t *testing.T) {
```
derekcollison, line 9445: Can be made into one test that tests single server and cluster, I have a few examples of that.

server/jetstream_cluster.go
```diff
@@ -3848,8 +3848,29 @@ func (s *Server) jsClusteredStreamListRequest(acc *Account, ci *ClientInfo, offs
...
 	for _, sa := range cc.streams[acc.Name] {
-		streams = append(streams, sa)
+		if IsNatsErr(sa.err, JSClusterNotAssignedErr) {
+			continue
+		}
+
+		if filter != _EMPTY_ {
+			// These could not have subjects auto-filled in since they are raw and unprocessed.
+			if len(sa.Config.Subjects) == 0 {
+				if SubjectsCollide(filter, sa.Config.Name) {
```
derekcollison, line 3858: Should it be collide or do we want the filter to be equal or a subset of the subjects attached to the stream?

server/jetstream_cluster_test.go
ripienaar: Ok

server/jetstream_cluster.go
ripienaar: Don’t know tbh. I copied logic from stream names they should behave the same. So this is whatever names does :)

derekcollison: ok so this is from names filter code, so agree using same thing makes sense here.

ripienaar:
Updated to one test

derekcollison (approved):
LGTM

PR #2696, opened 2021-11-17
support always presenting a nonce to clients

ripienaar:
The nonce feature is useful to custom authentication plugins
but at present there is no way to enable a nonce to be presented
other than by setting nkey accounts etc.

This enables the nonce to always be presented in those situations.
Since its primarily useful to embedded scenarios there is no corresponding
configuration file behavior for this flag.

/cc @nats-io/core


kozlovic (approved):
LGTM

derekcollison (approved):
LGTM

PR #2698, opened 2021-11-17
Bump tests code coverage timeout

kozlovic:
Default of 10 minute (per package) seem to no longer be enough,
so bumping to 20 minutes, as it is on non code coverage runs.


derekcollison (approved):
LGTM

PR #2699, opened 2021-11-18
expose the nonce to custom authentication

ripienaar:

/cc @nats-io/core

### DIFF
--- a/server/server_test.go
+++ b/server/server_test.go
@@ -603,32 +603,46 @@ func TestNilMonitoringPort(t *testing.T) {
 	}
 }
 
-type DummyAuth struct{}
+type DummyAuth struct {
+	t         *testing.T
+	needNonce bool
+}
 
 func (d *DummyAuth) Check(c ClientAuthentication) bool {
+	if d.needNonce && len(c.GetNonce()) == 0 {
+		d.t.Fatalf("Expected a nonce but received none")
+	} else if !d.needNonce && len(c.GetNonce()) > 0 {
+		d.t.Fatalf("Received a nonce when none was expected")
+	}
+
 	return c.GetOpts().Username == "valid"
 }
 
 func TestCustomClientAuthentication(t *testing.T) {
-	var clientAuth DummyAuth
+	testAuth := func(t *testing.T, nonce bool) {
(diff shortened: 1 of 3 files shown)


kozlovic (approved):
LGTM, but we have a test called `TestCustomClientAuthentication` that you could extend/copy form to actually verify that a nonce is specified?

ripienaar:
> LGTM, but we have a test called `TestCustomClientAuthentication` that you could extend/copy form to actually verify that a nonce is specified?

OK, added a bit to that test

server/server_test.go
```diff
@@ -629,6 +638,14 @@ func TestCustomClientAuthentication(t *testing.T) {
 	if _, err := nats.Connect(addr, nats.UserInfo("invalid", "")); err == nil {
 		t.Fatal("Expected client to fail to connect")
 	}
+
+	s.opts.AlwaysEnableNonce = true
```
kozlovic, line 642: I am getting races when running your test, so need to revisit that.

kozlovic (changes requested):

server/server_test.go
ripienaar: OK, I was less lazy now, think it probably fixed now

server/server_test.go
```diff
@@ -603,32 +603,46 @@ func TestNilMonitoringPort(t *testing.T) {
...
 }
 
-type DummyAuth struct{}
+type DummyAuth struct {
+	t         *testing.T
+	needNonce bool
+}
 
 func (d *DummyAuth) Check(c ClientAuthentication) bool {
+	if d.needNonce && len(c.GetNonce()) == 0 {
```
kozlovic, line 612: It's ok since tests are not supposed to fail, but if it fails, it will produce a DATA race since t.Fatal will be invoked from different go routine than the test itself. Again, no worry since if we get the DATA race it means that the test failed anyway.

kozlovic (approved):
LGTM

server/server_test.go
ripienaar: yeah fair, but its exceptional as you say so if its fails its failing anyway :) I'll leave as is and merge on green

PR #2701, opened 2021-11-18
Remove TestNoRace tests from code coverage

kozlovic:
The "TestNoRace" can be very expensive memory/cpu wise.

When running test coverage, we do it with `-covermode=atomic` which is a bit similar to `-race` in that the runtime adds tracking that causes memory overhead. So we will remove those from code coverage.
It saves about 2min of running the code coverage tests by excluding those, but server package code coverage falls by a bit more than 1%.


derekcollison (approved):
LGTM

PR #2702, opened 2021-11-18
Better handling for stream mismatch scenarios.

derekcollison:
1. When a snapshot did not yield actionable data, we were not setting new last sequence if we have to readjust based on snapshot. This could lead to spinning on stream reset for followers.

2. When a stream has lots of failures by design, like KV abstraction, if we cleared the clfs state we would endlessly spin trying to reset the stream.

3. Also have better logging when we fail to decode a low level packet from the raft layer.

/cc @nats-io/core


kozlovic (approved):
LGTM

PR #2704, opened 2021-11-19
[FIXED] Bug in memstore when setting max msgs per subject to 1.

derekcollison:

/cc @nats-io/core


kozlovic (approved):
LGTM

PR #2707, opened 2021-11-22
[FIXED/CHANGED] Add leafnode websocket connection type

kozlovic:
This was missing since WEBSOCKET allowed connection type is really
used for client connections.
If one wants to limit a configured user to leafnode connections,
including if the connection is over websocket, but does not
want an application to connect over websocket using this user,
this would have been impossible to configure.

The JWT library has been updated to add LEAFNODE_WS and MQTT_WS for
future work.


matthiashanel (approved):
LGTM

aricart (approved):

PR #2711, opened 2021-11-29
[FIXED #2706] - Only utilize full state with deleted details when necessary.

### LINKED ISSUE #2706: Stream state with large interior deletes slow
When using Max Messages Per Subject style streams interior deletes are the norm, streams can easily build up millions or even 10s of millions of interior deleted messages.

In those cases Stream State API is very slow primarily because we duplicate and sort the list of these interior deletes - even when detail is not requested.

In memory store its not too bad, still we duplicate potentially 10s of millions of uint64s and then sort them - this can take minutes.

https://github.com/nats-io/nats-server/blob/ea48105526db36b566b3afb36ddfacccf3f97f3a/server/memstore.go#L750-L766

In filestore we seem to also do some block level maintenance in the middle of state gathering:

https://github.com/nats-io/nats-server/blob/ea48105526db36b566b3afb36ddfacccf3f97f3a/server/filestore.go#L3386-L3398

In the typical stream state we just discard all these numbers:

https://github.com/nats-io/nats-server/blob/ea48105526db36b566b3afb36ddfacccf3f97f3a/server/stream.go#L3525-L3538

For memory stream its quite easy to optimise this down to only giving or even calculating this list but for disk things are a bit more complex.  We'd need to extend the Store and WAL interfaces State() with a boolean and then carefully go through places where deleted is not needed, or a new form of State() should be added - but I think the design of the current State() functions need a rethink.

derekcollison:

Resolves #2706

/cc @nats-io/core

### DIFF
--- a/server/const.go
+++ b/server/const.go
@@ -41,7 +41,7 @@ var (
 
 const (
 	// VERSION is the current version for the server.
-	VERSION = "2.6.5"
+	VERSION = "2.6.6-beta"
 
 	// PROTO is the currently supported protocol.
 	// 0 was the original
--- a/server/consumer.go
+++ b/server/consumer.go
@@ -2905,12 +2905,13 @@ func (o *consumer) selectStartingSeqNo() {
 	if o.mset == nil || o.mset.store == nil {
 		o.sseq = 1
 	} else {
-		stats := o.mset.store.State()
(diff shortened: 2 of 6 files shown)


kozlovic (approved):
LGTM

PR #2714, opened 2021-11-30
[FIXED] Monitoring: tls configuration not updated on reload

kozlovic:
When creating the http server, we need to provide a TLS configuration.
After a config reload, the new TLS config would not be reflected.

We had the same issue with Websocket and was fixed with the use
of tls.Config.GetConfigForClient API, which makes the TLS handshake
to ask for a TLS config. That fix for websocket was simply not applied
to the HTTPs monitoring case.

I have also fixed some flappers due to the use of localhost instead
of 127.0.0.1 (connections possibly would resolve to some IPv6 address
that the server would not accept, etc..)

### DIFF
--- a/server/events_test.go
+++ b/server/events_test.go
@@ -2257,7 +2257,7 @@ func TestServerEventsFilteredByTag(t *testing.T) {
 			listen: -1
 			no_advertise: true
 			routes [
-				nats-route://localhost:%d
+				nats-route://127.0.0.1:%d
 			]
 		}
 		system_account: SYS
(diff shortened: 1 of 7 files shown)


kozlovic:
@matthiashanel I have fixed some tests that tended to flap because of the use of "localhost" since it can resolve to IPv6 (::1 or some other IPv6 local ips) which server may not accept. You will remember that I showed you that once on a google meet because you had a flapper and that was the reason.

matthiashanel (approved):
LGTM

philpennock (approved):

PR #2717, opened 2021-12-01
Stream valid subjects

ripienaar:
/cc @nats-io/core


ripienaar:
This could be a breaking change and prevent streams from recovering at startup.  Not sure what to do about that.

But this is quite a big deal especially on streams - it caused leafnode protocol errors for example.

derekcollison (approved):
LGTM

PR #2718, opened 2021-12-01
prevent stream update to add subjects to mirrors

ripienaar:

/cc @nats-io/core


ripienaar:
@derekcollison same concerns about startup as the other one - so if you're updating configs in your follow up PR maybe also do this one

derekcollison (approved):
LGTM

PR #2719, opened 2021-12-01
[FIXED] Corrupted headers receiving from consumer with meta-only

kozlovic:
When a consumer is configured with "meta-only" option, and the
stream was backed by a memory store, a memory corruption could
happen causing the application to receive corrupted headers.

Also replaced most of usage of `append(a[:0:0], a...)` to make
copies. This was based on this wiki:
https://github.com/go101/go101/wiki/How-to-efficiently-clone-a-slice%3F

But since Go 1.15, it is actually faster to call make+copy instead.


derekcollison (approved):
LGTM - Thanks!

PR #2721, opened 2021-12-01
There were situations where invalid subjects could be assigned to streams.

derekcollison:
This will patch them on the fly during recovery. Specifically subjects with leading or trailing spaces and mirror streams with any subjects at all.

/cc @nats-io/core

### DIFF
--- a/server/jetstream.go
+++ b/server/jetstream.go
@@ -1115,6 +1115,33 @@ func (a *Account) EnableJetStream(limits *JetStreamAccountLimits) error {
 				s.Warnf("  Error adding stream %q to template %q: %v", cfg.Name, cfg.Template, err)
 			}
 		}
+
+		// We had a bug that could allow subjects in that had prefix or suffix spaces. We check for that here
+		// and will patch them on the fly for now. We will warn about them.
+		var hadSubjErr bool
+		for i, subj := range cfg.StreamConfig.Subjects {
+			if !IsValidSubject(subj) {
+				s.Warnf("  Detected bad subject %q while adding stream %q, will attempt to repair", subj, cfg.Name)
+				if nsubj := strings.TrimSpace(subj); IsValidSubject(nsubj) {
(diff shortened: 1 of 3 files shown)


server/jetstream.go
```diff
@@ -1115,6 +1115,33 @@ func (a *Account) EnableJetStream(limits *JetStreamAccountLimits) error {
...
+				}
+			}
+		}
+		if hadSubjErr {
+			continue
+		}
+
+		// The other possible bug is assigning subjects to mirrors, so check for that and patch as well.
+		if cfg.StreamConfig.Mirror != nil && len(cfg.StreamConfig.Subjects) > 0 {
+			cfg.StreamConfig.Subjects = nil
```
matthiashanel, line 1141: Maybe a warning trace?

matthiashanel (commented):
PR looks good, I'd add a trace

server/jetstream.go
kozlovic: I second that. Add a s.Warnf here like you do for subject fixing..

kozlovic (approved):
Overall LGTM, but would add a warning for subjects in mirrors..

server/jetstream.go
derekcollison: ok will do, one sec..

kozlovic (approved):
LGTM, but if I recall RI's PR was checking for deliver subject. Does not seem to be covered here (would be more for consumers, not streams).

derekcollison:
I confirmed with R.I. that the consumer thing was not reported, he just added it. We could add a test, but he does check now so did not think it would be worth the recovery fixup since never reported.

ripienaar:
A consumer in that state will never work. So users probsbly already fixed it themselves. I just wanted to have a better UX. 

a stream in this situation could have one of n bad subjects and would have worked and we need to avoid the unrecoverable situation where a bad stream has data and then becomes unusable.

matthiashanel (approved):
LGTM

PR #2725, opened 2021-12-02
Set incomplete error when cluster list fails

matthiashanel:

This is what is done in `jsClusteredStreamListRequest` as well


kozlovic:
Adding RI as reviewer to make sure that this is the proper way to set the error..

ripienaar (approved):
LGTM

PR #2727, opened 2021-12-02
[UPDATED] golang.org/x/crypto dependency

kozlovic:
They just released some fix that is not affecting the NATS Server
but could cause some security vulnerability reports.

/cc @nats-io/core


derekcollison (approved):
LGTM

PR #2743, opened 2021-12-13
fixes a nil panic on jsz

ripienaar:
Appears what happens is that the getPublicConsumers()
is called which produces a list of consumers and that
between the time the list is made and the Info() is
called the ephemeral was removed.

/cc @nats-io/core


ripienaar:
I am not sure how I would write a test for this concurrency issue though.

wallyqs (approved):

derekcollison (approved):
LGTM

PR #2747, opened 2021-12-15
[FIXED] TLS map: panic for existing user but conn type not allowed

kozlovic:
For TLS configuration with `verify_and_map` set to true, if a
connection connects and has a certificate with ID that matches
a user, but that user's `allowed_connection_types` is specified
and does not have the connection type in its list, then the
server will panic.


wallyqs (approved):
LGTM

derekcollison (approved):
LGTM

PR #2749, opened 2021-12-16
Adding jwt request callout to auth

matthiashanel:
This calls out to a signup service and informs the client of the jwt.
If the client provides an account id, that is where the request will go.
If non is provided, try the system account as global default.

In order to signal an account's willingness to partake in this, the pressence of this tag is needed in the account jwt:
nats-onboard-jwt

This takes the value of token and an optional nkey and calls the service with it.
A client will have to add the account option.
It will have to obtain the jwt from the info block, so that reconnects do not require obtaining a new jwt
Other than that the only thing needed is to forget the jwt once you receive ErrAuthExpired.

I tested this out end to end without issues.

Changes to my local nats.go, are small, but not complete.
What is missing is restoring state on ErrAuthExpired

```
diff --git a/nats.go b/nats.go
index 6f9067c..c1c9204 100644
--- a/nats.go
+++ b/nats.go
@@ -439,6 +439,9 @@ type Options struct {

        // InboxPrefix allows the default _INBOX prefix to be customized
        InboxPrefix string
+
+       // Account name to send jwt requests to
+       Account string
 }

 const (
@@ -666,6 +669,7 @@ type serverInfo struct {
        MaxPayload   int64    `json:"max_payload"`
        CID          uint64   `json:"client_id,omitempty"`
        ClientIP     string   `json:"client_ip,omitempty"`
+       JWT          string   `json:"client_jwt,omitempty"`
        Nonce        string   `json:"nonce,omitempty"`
        Cluster      string   `json:"cluster,omitempty"`
        ConnectURLs  []string `json:"connect_urls,omitempty"`
@@ -698,6 +702,7 @@ type connectInfo struct {
        Echo         bool   `json:"echo"`
        Headers      bool   `json:"headers"`
        NoResponders bool   `json:"no_responders"`
+       Account      string `json:"account"`
 }

 // MsgHandler is a callback function that processes messages delivered to
@@ -1124,6 +1129,17 @@ func CustomInboxPrefix(p string) Option {
        }
 }

+// DrainTimeout is an Option to set the timeout for draining a connection.
+func Account(a string) Option {
+       return func(o *Options) error {
+               if !nkeys.IsValidPublicAccountKey(a) {
+                       return fmt.Errorf("Invalid Account Id")
+               }
+               o.Account = a
+               return nil
+       }
+}
+
 // Handler processing

 // SetDisconnectHandler will set the disconnect event handler.
@@ -2073,7 +2089,7 @@ func (nc *Conn) connectProto() (string, error) {
        // If our server does not support headers then we can't do them or no responders.
        hdrs := nc.info.Headers
        cinfo := connectInfo{o.Verbose, o.Pedantic, ujwt, nkey, sig, user, pass, token,
-               o.Secure, o.Name, LangString, Version, clientProtoInfo, !o.NoEcho, hdrs, hdrs}
+               o.Secure, o.Name, LangString, Version, clientProtoInfo, !o.NoEcho, hdrs, hdrs, o.Account}

        b, err := json.Marshal(cinfo)
        if err != nil {
@@ -2953,6 +2969,16 @@ func (nc *Conn) processInfo(info string) error {
                return err
        }

+       // bootstrap into jwt environment.
+       if nc.Opts.UserJWT == nil && ncInfo.JWT != "" {
+               j := ncInfo.JWT
+               nc.Opts.UserJWT = func() (string, error) {
+                       return j, nil
+               }
+               // We allow to set an nkey and retrieve the matching jwt. Thus, clear nkey.
+               nc.Opts.Nkey = ""
+       }
+
        // Copy content into connection's info structure.
        nc.info = ncInfo
        // The array could be empty/not present on initial connect,
```

### DIFF
--- a/server/auth.go
+++ b/server/auth.go
@@ -403,14 +403,14 @@ func (s *Server) onboardingCallout(c *client, opts *Options) bool {
 	msgChan := make(chan string)
 	defer close(msgChan)
 	ib := s.newRespInbox()
-	sub, err := s.sysSubscribe(ib, func(_ *subscription, _ *client, acc *Account, subject, reply string, rmsg []byte) {
+	sub, err := acc.subscribeInternal(ib, func(_ *subscription, _ *client, acc *Account, subject, reply string, rmsg []byte) {
 		msgChan <- string(rmsg)
 	})
 	if err != nil {
 		c.Debugf("Authentication requires a user JWT - Setup failure %s", signupName)
 		return false
 	}
-	defer s.sysUnsubscribe(sub)
+	defer sub.client.processUnsub(sub.sid)
 	var hdr map[string]string
 	ci := c.getClientInfo(true)
 	// TODO clear out account in ci? the value of $G may be confusing
--- a/server/auth.go
+++ b/server/auth.go
@@ -371,6 +371,89 @@ func (c *client) matchesPinnedCert(tlsPinnedCerts PinnedCertSet) bool {
 
 // JWT Tag value for Account JWT Tag, enabling on-boarding services
 const featureTagOnboard = "nats-onboard-jwt"
+const onboardSubjFmt = "$jwt.request.%s"
+
+func (s *Server) onboardingCallout(c *client, opts *Options) bool {
+	var acc *Account
+	if c.opts.Account == _EMPTY_ {
+		// check default (system account) callout
+		acc = s.SystemAccount()
+		if acc == nil {
+			c.Debugf("Authentication requires a user JWT - Token Request failed - Account Missing")
+			return false
+		}
+	} else {
+		// check account listed
+		var err error
+		acc, err = s.fetchAccount(c.opts.Account)
+		if err != nil {
+			c.Debugf("Authentication requires a user JWT - Token Request Account %q lookup failed: %s", c.opts.Account, err)
+			return false
+		}
+	}
+	acc.mu.RLock()
+	signupName := fmt.Sprintf("Account: %s/%s", acc.Name, acc.nameTag)
+	featOnboard := acc.tags.Contains(featureTagOnboard)
+	acc.mu.RUnlock()
+	if !featOnboard {
+		c.Debugf("Authentication requires a user JWT")
+		return false
+	}
+	// Perform the actual JWT Request
+	msgChan := make(chan string)
+	defer close(msgChan)
+	ib := s.newRespInbox()
+	sub, err := s.sysSubscribe(ib, func(_ *subscription, _ *client, acc *Account, subject, reply string, rmsg []byte) {
+		msgChan <- string(copyBytes(rmsg))
+	})
+	if err != nil {
+		c.Debugf("Authentication requires a user JWT - Setup failure %s", signupName)
+		return false
+	}
+	defer s.sysUnsubscribe(sub)
+	var hdr map[string]string
+	ci := c.getClientInfo(true)
+	// TODO clear out account in ci? the value of $G may be confusing
+	if b, _ := json.Marshal(ci); b != nil {
+		hdr = map[string]string{ClientInfoHdr: string(b)}
+	}
+	// Always look up with an nkey, even if none is provided.
+	// If none is provided by the client, the service has to return a bearer token
+	var dest string
+	if c.opts.Nkey == _EMPTY_ {
+		if kp, err := nkeys.CreateUser(); err != nil {
+			c.Debugf("Authentication requires a user JWT - %s Setup failure: %s", signupName, err)
+			return false
+		} else if pub, err := kp.PublicKey(); err != nil {
+			c.Debugf("Authentication requires a user JWT - Setup failure %s", signupName, err)
+			return false
+		} else {
+			dest = pub
+		}
+	} else {
+		dest = c.opts.Nkey
+	}
+	to := secondsToDuration(opts.AuthTimeout) - time.Now().Sub(c.start)
+	if to <= 0 {
+		c.Debugf("Authentication requires a user JWT - Token Request Timed Out Immediately: %s", signupName)
+		return false
+	}
+	s.sendInternalAccountMsgWithReply(acc, fmt.Sprintf(onboardSubjFmt, dest), ib, hdr, c.opts.Token, true)
+	select {
+	case <-time.After(to):
+		c.Debugf("Authentication requires a user JWT - Token Request Timed Out: %s", signupName)
+		return false
+	case m := <-msgChan:
+		if m == _EMPTY_ {
+			c.Debugf("Authentication requires a user JWT - Empty Token Request Response")
+			return false
+		}
+		c.opts.JWT = m
+	}
+	c.flags.set(jwtRetrieved)
+	c.Warnf("Obtained JWT using: %s", signupName)
+	return true
+}
 
 func (s *Server) processClientOrLeafAuthentication(c *client, opts *Options) bool {
 	var (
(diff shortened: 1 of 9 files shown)


server/auth.go
```diff
@@ -447,8 +451,80 @@ func (s *Server) processClientOrLeafAuthentication(c *client, opts *Options) boo
...
+			sub, err := s.sysSubscribe(ib, func(_ *subscription, _ *client, acc *Account, subject, reply string, rmsg []byte) {
+				msgChan <- copyBytes(rmsg)
+			})
+			if err != nil {
+				c.Debugf("Authentication requires a user JWT - Setup failure %s", signupName)
+				return false
+			}
+			defer s.sysUnsubscribe(sub)
+			var hdr map[string]string
+			ci := c.getClientInfo(true)
```
ripienaar, line 492: What will our guidance be on what part of client info will be used to securely identify a new client as being authorized to request access? Would they have something like a JWT set that the callback can inspect?

server/auth.go
```diff
@@ -447,8 +451,80 @@ func (s *Server) processClientOrLeafAuthentication(c *client, opts *Options) boo
 	if s.trustedKeys != nil {
 		if c.opts.JWT == _EMPTY_ {
 			s.mu.Unlock()
-			c.Debugf("Authentication requires a user JWT")
-			return false
+
+			var acc *Account
```
kozlovic, line 455: If we could move this code into its own function (if possible, that is, don't need a lot of variables defined in processClientOrLeafAuthentication), that would be great. The function processClientOrLeafAuthentication is quite long already..

server/auth.go
```diff
@@ -447,8 +451,80 @@ func (s *Server) processClientOrLeafAuthentication(c *client, opts *Options) boo
...
+			if !featOnboard {
+				c.Debugf("Authentication requires a user JWT")
+				return false
+			}
+			// Perform the actual JWT Request
+			msgChan := make(chan []byte)
+			defer close(msgChan)
+			ib := s.newRespInbox()
+			sub, err := s.sysSubscribe(ib, func(_ *subscription, _ *client, acc *Account, subject, reply string, rmsg []byte) {
+				msgChan <- copyBytes(rmsg)
```
kozlovic, line 484: I see that later you convert to a string anyway, so I would have made `msgChan` a chan string instead and here have `msgChan <- string(rmsg)`.

server/auth.go
```diff
@@ -447,8 +451,80 @@ func (s *Server) processClientOrLeafAuthentication(c *client, opts *Options) boo
...
+				} else if pub, err := kp.PublicKey(); err != nil {
+					c.Debugf("Authentication requires a user JWT - Setup failure %s", signupName, err)
+					return false
+				} else {
+					dest = pub
+				}
+			} else {
+				dest = c.opts.Nkey
+			}
+			s.sendInternalAccountMsgWithReply(acc, fmt.Sprintf("$jwt.request.%s", dest), ib, hdr, c.opts.Token, true)
```
kozlovic, line 513: So `$jwt.request.*` is a subject that some outside application is listening to, right?

server/auth.go
```diff
@@ -447,8 +451,80 @@ func (s *Server) processClientOrLeafAuthentication(c *client, opts *Options) boo
...
+			select {
+			case <-time.After(secondsToDuration(opts.AuthTimeout)):
+				c.Debugf("Authentication requires a user JWT - Token Request Timed Out: %s", signupName)
+				return false
+			case m := <-msgChan:
+				if len(m) == 0 {
+					c.Debugf("Authentication requires a user JWT - Empty Token Request Response")
+					return false
+				}
+				c.opts.JWT = string(m)
```
kozlovic, line 523: Would not need that if you made `msgChan` a `chan string` as discussed above.

server/jwt_test.go
```diff
@@ -5844,6 +5844,116 @@ func TestJWTNoSystemAccountButNatsResolver(t *testing.T) {
...
+		uJwt, err := jwt.IssueUserJWT(signer, to, uPub, "", 0)
+		if err != nil {
+			errChan <- err
+			return
+		}
+		if err := msg.Respond([]byte(uJwt)); err != nil {
+			errChan <- err
+			return
+		}
+		if ri, ok := msg.Header["Nats-Request-Info"]; !ok {
```
kozlovic, line 5918: You can use the constant `ClientInfoHdr` here instead..

server/auth.go
```diff
@@ -447,8 +451,80 @@ func (s *Server) processClientOrLeafAuthentication(c *client, opts *Options) boo
...
+					return false
+				} else {
+					dest = pub
+				}
+			} else {
+				dest = c.opts.Nkey
+			}
+			s.sendInternalAccountMsgWithReply(acc, fmt.Sprintf("$jwt.request.%s", dest), ib, hdr, c.opts.Token, true)
+			select {
+			case <-time.After(secondsToDuration(opts.AuthTimeout)):
```
kozlovic, line 515: I am not sure if opts.AuthTimeout is guaranteed to not be unlimited. I know we have a default if set to 0, so I guess there is no "unlimited" value, but may want to check.
Also, we could get fancy and before the select, deduct the time already elapsed since the TCP connection was accepted from the AuthTimeout. To do so, you can use "time.Now().UTC() - c.start".

kozlovic (changes requested):
The review is more at the low-level code, not the feature itself..

server/auth.go
matthiashanel: I basically re-use the token value in the jwt context to do just that.
TestJWTSignupFromSysAcc provides nats.Token("shared-secret-bearer") in line 5928 to the connection which is checked by the jwt issuing service in line 5893.

Right now, I'd allow for an empty token, but can be convinced to require one

matthiashanel: The client info I threw in because it exists. 
(Once a customer wanted to know the ip, a client connected from, in order to issue jwt)

server/auth.go
matthiashanel: fixed

server/auth.go
matthiashanel: yes, but I will move this string into a constant.

server/auth.go
matthiashanel: fixed

server/auth.go
matthiashanel: fixed

server/auth.go
matthiashanel: I checked, it is not unlimited.
Also, got fancy and subtract elapsed time

server/auth.go
```diff
@@ -368,6 +369,92 @@ func (c *client) matchesPinnedCert(tlsPinnedCerts PinnedCertSet) bool {
...
+	if !featOnboard {
+		c.Debugf("Authentication requires a user JWT")
+		return false
+	}
+	// Perform the actual JWT Request
+	msgChan := make(chan string)
+	defer close(msgChan)
+	ib := s.newRespInbox()
+	sub, err := s.sysSubscribe(ib, func(_ *subscription, _ *client, acc *Account, subject, reply string, rmsg []byte) {
+		msgChan <- string(copyBytes(rmsg))
```
kozlovic, line 407: You don't need `copyBytes()` here, `string()` is already making a copy of the byte slice.

kozlovic (changes requested):
Remove unnecessary copyBytes(). Other than that, I wonder if it is not a bit too costly to create sys sub here in processing of a user's CONNECT. Those are "real" subscriptions that need to be propagate to the cluster/super-cluster.

server/auth.go
matthiashanel: fixed

derekcollison:
I will take a look over the weekend but this one feels (without reading all comments) that we might have wanted a bit more time to think through the edge cases.

server/auth.go
```diff
@@ -368,6 +369,93 @@ func (c *client) matchesPinnedCert(tlsPinnedCerts PinnedCertSet) bool {
...
+			c.Debugf("Authentication requires a user JWT - Token Request Account %q lookup failed: %s", c.opts.Account, err)
+			return false
+		}
+	}
+	acc.mu.RLock()
+	signupName := fmt.Sprintf("Account: %s/%s", acc.Name, acc.nameTag)
+	featOnboard := acc.tags.Contains(featureTagOnboard)
+	acc.mu.RUnlock()
+	if !featOnboard {
+		c.Debugf("Authentication requires a user JWT")
```
tbeets, line 399: Shouldn't server debug report that Account blah-blah not enabled for external onboarding service?  Otherwise, not helpful to discriminate and fix.

PR #2750, opened 2021-12-17
[FIXED] JetStream: interest across gateways

### LINKED ISSUE #2745: QueueSub... is not normal in a super-cluster, when using different cluster to subscribe without providing durable name.
## Defect
 - [x] nats-server -DV
 ```shell
[82] 2021/12/15 04:47:13.935429 [INF] Starting nats-server
[82] 2021/12/15 04:47:13.935484 [INF]   Version:  2.6.4
[82] 2021/12/15 04:47:13.935487 [INF]   Git:      [a27de5a]
[82] 2021/12/15 04:47:13.935490 [DBG]   Go build: go1.16.10
[82] 2021/12/15 04:47:13.935499 [INF]   Name:     NCONUHPM3NLBQJ4QCSQWCYTEINVWAAHCBSVXJY6N3JYZ77CLOY2UTURJ
[82] 2021/12/15 04:47:13.935502 [INF]   ID:       NCONUHPM3NLBQJ4QCSQWCYTEINVWAAHCBSVXJY6N3JYZ77CLOY2UTURJ
[82] 2021/12/15 04:47:13.935533 [DBG] Created system account: "$SYS"
```

- client libraries: nats.go v1.13.1

- nats:2.6.4-alpine3.14 on k8s
1. Supposed we have a super-cluster, which contains C1 cluster and C2 cluster2, details shown as below:

|ClusterName|Jetstream| k8s svc|accounts|
|---|---|----|---|
|c1|enabled| nats-c1:4222|```"PUB":{user: "pub1", password: "pub1"}```|
|c2|enabled| nats-c2:4222|```"PUB":{user: "pub1", password: "pub1"}```|

2. first, we run the test code as below, which uses the ```nats-c1:4222``` as the serverURL.
using nats-cli tool to see the consumer info and stream info, we can get the stream and consumer are both created on c1 cluster. 

```go
const serverURL = ""
const streamName = "demo"
const subject = "demo.test"
const queue = "test"

connect, err := nats.Connect(serverURL, nats.UserInfo("pub1", "pub1"))
defer connect.Close()
if err != nil {
	log.Fatal(err)
}
js, err := connect.JetStream()
if err != nil {
	log.Fatal(err)
}

_, err = js.StreamInfo(streamName)
if err != nil {
	if err == nats.ErrStreamNotFound {
		_, err := js.AddStream(&nats.StreamConfig{
			Name:     streamName,
			Subjects: []string{streamName + ".>"},
			Discard:  0,
			Storage:  nats.FileStorage,
			Replicas: 3,
		})
		if err != nil {
			return
		}
	} else {
		return
	}
}

ch1 := make(chan *nats.Msg)
ch2 := make(chan *nats.Msg)
_, err = js.ChanQueueSubscribe(subject, queue, ch1, nats.BindStream(streamName))
if err != nil {
	return
}
_, err = js.ChanQueueSubscribe(subject, queue, ch2, nats.BindStream(streamName))
if err != nil {
	return
}

go func() {
	for i := 0; i < 100000; i++ {
		msg := nats.NewMsg(subject)
		msg.Data = []byte(fmt.Sprintf("MSG %d", i))
		pubAck, err := js.PublishMsg(msg)
		if err != nil {
			return
		}
		fmt.Printf("%s %s %d\n", pubAck.Domain, pubAck.Stream, pubAck.Sequence)
	}
}()

for {
	select {
	case msg := <-ch1:
		fmt.Printf("ch1: %s\n", string(msg.Data[:]))
		err := msg.Ack()
		if err != nil {
			return
		}
	case msg := <-ch2:
		fmt.Printf("ch2: %s\n", string(msg.Data[:]))
		err := msg.Ack()
		if err != nil {
			return
		}
	}
}
```

3. second, we run the code again, but changing the serverURL to ```nats-c2:4222```.

#### Expected result:

In queue subscribe model, two case should have the same result. All sending msg should be received from ch1 and ch2.

#### Actual result:

Only in the first case, the result is as expected. in the second case, only sending go-routine is normal, the queue-sub can't get any msg. Using nats-cli to get the consumer info, we get a error consumer state: ``` Active Interest: No interest```.
But when we call the 

```go 
_, err = js.ChanQueueSubscribe(subject, queue, ch1, nats.BindStream(streamName), nats.Durable(consumerName))
_, err = js.ChanQueueSubscribe(subject, queue, ch2, nats.BindStream(streamName), nats.Durable(consumerName))
```
with additional subOpts. (consumer is not required to exist), everying is normal.

kozlovic:
If the interest existed prior to the initial creation of the
consumer, the gateway "watcher" would not be started, which means
that interest moving across the super-cluster after that would
not be detected.

The watcher runs every second and not sure if this is costly or
not, so we may want to go a different approach of having a separate
interest change channel that would be specific to gateways. But this
means adding a new sublist where the interest would be registered
and that sublist would need to be updated when processing GW RSub
and RUnsub?

Resolves #2745 

### DIFF
--- a/server/consumer.go
+++ b/server/consumer.go
@@ -786,15 +786,16 @@ func (o *consumer) setLeader(isLeader bool) {
 		if o.isPushMode() {
 			o.inch = make(chan bool, 8)
 			o.acc.sl.registerNotification(o.cfg.DeliverSubject, o.cfg.DeliverGroup, o.inch)
-			if o.active = <-o.inch; !o.active {
-				// Check gateways in case they are enabled.
-				if s.gateway.enabled {
-					o.active = s.hasGatewayInterest(o.acc.Name, o.cfg.DeliverSubject)
-					stopAndClearTimer(&o.gwdtmr)
-					o.gwdtmr = time.AfterFunc(time.Second, func() { o.watchGWinterest() })
-				}
-			} else {
+			if o.active = <-o.inch; o.active {
 				o.checkQueueInterest()
+			}
+			// Check gateways in case they are enabled.
+			if s.gateway.enabled {
+				if !o.active {
+					o.active = s.hasGatewayInterest(o.acc.Name, o.cfg.DeliverSubject)
+				}
+				stopAndClearTimer(&o.gwdtmr)
+				o.gwdtmr = time.AfterFunc(time.Second, func() { o.watchGWinterest() })
 			}
 		}
 
@@ -1277,7 +1278,7 @@ func (o *consumer) updateDeliverSubjectLocked(newDeliver string) {
 		o.forceExpirePending()
 	}
 
-	o.acc.sl.ClearNotification(o.dsubj, o.inch)
+	o.acc.sl.clearNotification(o.dsubj, o.cfg.DeliverGroup, o.inch)
 	o.dsubj, o.cfg.DeliverSubject = newDeliver, newDeliver
 	// When we register new one it will deliver to update state loop.
 	o.acc.sl.registerNotification(newDeliver, o.cfg.DeliverGroup, o.inch)
@@ -3133,6 +3134,7 @@ func (o *consumer) stopWithFlags(dflag, sdflag, doSignal, advisory bool) error {
 		o.signalNewMessages()
 	}
 	n := o.node
+	qgroup := o.cfg.DeliverGroup
 	o.mu.Unlock()
 
 	if c != nil {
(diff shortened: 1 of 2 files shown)


derekcollison (approved):
LGTM

tbeets:
Could this scheme at a server be inverted (eval on SUB/UNSUB at the server) to solve the general case of alerting on foreign-account interest changes?

tbeets:
On SUB/UNSUB can we follow the current export/import maps in reverse?

PR #2751, opened 2021-12-20
[FIXED] JetStream: stream blocked recovering snapshot

kozlovic:
If a node fell behind, when catching up with the rest of the
cluster, it is possible that a lot of append entries accumulate
and the server would print warnings such as:
```
[WRN] RAFT [jZ6RvVRH - S-R3F-CQw2ImK6] <some number> append entries pending
```
It would then continously print the following warning:
```
AppendEntry failed to be placed on internal channel
```
When that happens, this node would always be shown with be running the
same number of operations behind (using `nats s info`) if there are
no new messages added to the stream, or an increasing number of
operations if there is still activity.

### DIFF
--- a/server/raft.go
+++ b/server/raft.go
@@ -263,6 +263,7 @@ var (
 	errLeaderLen         = fmt.Errorf("raft: leader should be exactly %d bytes", idLen)
 	errTooManyEntries    = errors.New("raft: append entry can contain a max of 64k entries")
 	errBadAppendEntry    = errors.New("raft: append entry corrupt")
+	errChannelFull       = errors.New("raft: entry failed to be placed on internal channel")
 )
 
 // This will bootstrap a raftNode by writing its config into the store directory.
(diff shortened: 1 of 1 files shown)


server/raft.go
```diff
@@ -263,6 +263,7 @@ var (
 	errLeaderLen         = fmt.Errorf("raft: leader should be exactly %d bytes", idLen)
 	errTooManyEntries    = errors.New("raft: append entry can contain a max of 64k entries")
 	errBadAppendEntry    = errors.New("raft: append entry corrupt")
+	errChannelFull       = errors.New("raft: entry failed to be placed on internal channel")
```
derekcollison, line 266: Are there other places we could use this? I think this code has a few select blocks on channels IIRC.

derekcollison (approved):
LGTM - One comment on other places we might be able to use this.

server/raft.go
kozlovic: In this file, we either use a different error, such as `errProposalFailed`, or simply n.error()/n.warn()/n.debug() in the "default:" case. Following our discussion about possibly removing use of channels, I think this is a stop gap and this error will disappear in the future.

derekcollison (approved):
LGTM

PR #2752, opened 2021-12-21
Memstore tracking of interior deletes improved.

derekcollison:
Memory store tracked interior deletes for stream state, but under KV semantics this could be a very large state to maintain.

In looking into memory usage we discovered it was actually faster to not track at all and generate on the fly. 
And this also saves quite a bit of memory in edge cases.

/cc @nats-io/core


kozlovic (approved):
LGTM

PR #2755, opened 2021-12-22
Added in ability to have account limits configured in server config.

derekcollison:

/cc @nats-io/core

### DIFF
--- a/server/accounts.go
+++ b/server/accounts.go
@@ -278,6 +278,8 @@ func (a *Account) shallowCopy() *Account {
 	}
 	// JetStream
 	na.jsLimits = a.jsLimits
+	// Server config account limits.
+	na.limits = a.limits
 
 	return na
 }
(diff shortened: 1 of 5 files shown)


wallyqs (approved):
LGTM

server/accounts.go
```diff
@@ -278,6 +278,8 @@ func (a *Account) shallowCopy() *Account {
 	}
 	// JetStream
 	na.jsLimits = a.jsLimits
+	// Server config account limits.
+	na.limits = a.limits
```
matthiashanel, line 282: Did we miss that for jwt before?

matthiashanel (approved):
Lgtm ,with a question to confirm my suspicion

server/accounts.go
derekcollison: No JWT fills in when it process the claim, it pulls apart the limits from the JWTs and places them into the account.

https://github.com/nats-io/nats-server/blob/acc_config_limits/server/accounts.go#L3168

PR #2756, opened 2021-12-22
Added test to show cross account interest for push consumers works.

derekcollison:
This was me trying to double check that interest across stream imports worked correctly.

/cc @nats-io/core


matthiashanel (approved):
Lgtm

PR #2757, opened 2021-12-22
Fixed system account issue where the wrong struct got updated

matthiashanel:
s.fetchAccount should not be used for the system account as it creates a new struct.
Changed to update in place in one location.
The other one I changed to lookup.  (In this location the system account will be fixed up a few lines further down.


derekcollison (approved):
LGTM

PR #2764, opened 2021-12-29
Large number of ephemeral consumers could exhaust Go runtime's max OS threads.

### LINKED ISSUE #2742: JetStream `Consumer assignment not cleaned up, retrying`, degraded cluster performance
## Defect

When using a lot of ephemeral consumers, a cluster with significant load eventually starts spamming a sample like below:

```
[1] 2021/12/13 02:44:15.208575 [WRN] Consumer assignment not cleaned up, retrying
[1] 2021/12/13 02:44:15.230045 [WRN] Consumer assignment not cleaned up, retrying
[1] 2021/12/13 02:44:15.241153 [INF] JetStream cluster new consumer leader for 'XXXX > YYYY > ZZZZ'
[1] 2021/12/13 02:44:15.289728 [WRN] Consumer assignment not cleaned up, retrying
[1] 2021/12/13 02:44:15.289972 [WRN] Consumer assignment not cleaned up, retrying
[1] 2021/12/13 02:44:15.300970 [WRN] Consumer assignment not cleaned up, retrying
[1] 2021/12/13 02:44:15.308939 [INF] JetStream cluster new consumer leader for 'XXXX > YYYY > ZZZZ'
[1] 2021/12/13 02:44:15.321413 [WRN] Consumer assignment not cleaned up, retrying
[1] 2021/12/13 02:44:15.359668 [WRN] Consumer assignment not cleaned up, retrying
```

I'm unsure if the leader migration for ephemeral consumers is normal even if none of the servers have rebooted.

`nats str info` of an affected stream:

```
Information for Stream XXXX created 2021-12-10T15:32:39Z

Configuration:

             Subjects: XXXX.>
     Acknowledgements: true
            Retention: File - Limits
             Replicas: 3
       Discard Policy: Old
     Duplicate Window: 2m0s
     Maximum Messages: unlimited
        Maximum Bytes: unlimited
          Maximum Age: 0.00s
    Maximum Consumers: unlimited


Cluster Information:

                 Name: XXXX
               Leader: prod-nats-js-2
              Replica: prod-nats-js-0, current, seen 0.36s ago
              Replica: prod-nats-js-1, current, seen 0.36s ago

State:

             Messages: 308,916
                Bytes: 63 MiB
             FirstSeq: 37,439 @ 2021-12-11T03:57:36 UTC
              LastSeq: 346,354 @ 2021-12-13T03:23:29 UTC
     Active Consumers: 57
```

`nats con info` of an affected ephemeral consumer:

```
Information for Consumer XXXX > 07ix4IRz created 2021-12-13T03:24:25Z

Configuration:

    Delivery Subject: XXXX.deliver.ephemeral.XXXX-34af714a-aa41-4434-8189-a70c80d85af4.01605c32-a7d7-452d-9287-db2b3aa85924
      Filter Subject: XXXX.34af714a-aa41-4434-8189-a70c80d85af4
      Deliver Policy: All
          Ack Policy: None
       Replay Policy: Instant
        Flow Control: false

Cluster Information:

                Name: XXXX
              Leader: XXXX
State:

   Last Delivered Message: Consumer sequence: 0 Stream sequence: 347268
     Unprocessed Messages: 0
          Active Interest: Active
````

 - [ ] Included `nats-server -DV` output (can't reproduce in staging)

Server: Docker `nats:2.6.6`

docker-compose
Ubuntu 20.04

1. Start a 3-node JetStream-enabled cluster
2. Create a stream with replication set to 3
3. Create a large amount of ephemeral consumers, many of which never have any messages delivered

#### Expected result:

Server CPU should remain low. Should not see warnings in console. Ephemeral consumers should be removed and not migrated.

#### Actual result:

CPU load steadily increases. CPU sits at effectively 0% until a steady stream of messages (~70 mps) are sent through the cluster. The two screenshots below are of the same 6-hour window.

image

image

### PR DESCRIPTION
**Pull request by derekcollison**
Under certain situations large number of consumers that are racing to update state or delete their stores during a delete
would start taking up OS threads due to blocking disk IO. When this happened and their were a bunch of Go routines becoming runnable the Go runtime would create extra OS threads to fill in the runnable pool and would exhaust the max thread setting. Even if it did not exhaust it you could see a quick run up in memory and unstable systems.

This code places a channel as a simple semaphore to limit the number of disk IO blocking OS threads.

Resolves #2742 

/cc @nats-io/core

### DIFF
--- a/server/consumer.go
+++ b/server/consumer.go
@@ -635,7 +635,8 @@ func (mset *stream) addConsumerWithAssignment(config *ConsumerConfig, oname stri
 	o.nextMsgSubj = fmt.Sprintf(JSApiRequestNextT, mn, o.name)
 
 	if o.isPushMode() {
-		o.dthresh = JsDeleteWaitTimeDefault
+		// Add in 1 sec of jitter above and beyond the default of 5s.
+		o.dthresh = JsDeleteWaitTimeDefault + time.Duration(rand.Int63n(1000))*time.Millisecond
 		if !o.isDurable() {
 			// Check if we are not durable that the delivery subject has interest.
 			// Check in place here for interest. Will setup properly in setLeader.
@@ -1026,8 +1027,8 @@ func (o *consumer) updateDeliveryInterest(localInterest bool) bool {
 }
 
 func (o *consumer) deleteNotActive() {
-	// Need to check again if there is not an interest now that the timer fires.
-	if !o.hasNoLocalInterest() {
+	// If we have local interest simply return.
+	if o.hasLocalInterest() {
 		return
 	}
 	o.mu.RLock()
@@ -1055,6 +1056,7 @@ func (o *consumer) deleteNotActive() {
 			// Check to make sure we went away.
 			// Don't think this needs to be a monitored go routine.
 			go func() {
+				var fs bool
 				ticker := time.NewTicker(time.Second)
 				defer ticker.Stop()
 				for range ticker.C {
@@ -1062,8 +1064,11 @@ func (o *consumer) deleteNotActive() {
 					ca := js.consumerAssignment(acc, stream, name)
 					js.mu.RUnlock()
 					if ca != nil {
-						s.Warnf("Consumer assignment not cleaned up, retrying")
-						meta.ForwardProposal(removeEntry)
+						if fs {
+							s.Warnf("Consumer assignment not cleaned up, retrying")
+							meta.ForwardProposal(removeEntry)
+						}
+						fs = true
 					} else {
 						return
 					}
@@ -3014,6 +3019,11 @@ func (o *consumer) isActive() bool {
 	return active
 }
 
+// hasLocalInterest returns if we have local interest.
+func (o *consumer) hasLocalInterest() bool {
+	return !o.hasNoLocalInterest()
+}
+
 // hasNoLocalInterest return true if we have no local interest.
 func (o *consumer) hasNoLocalInterest() bool {
 	o.mu.RLock()
@@ -3239,7 +3249,7 @@ func validFilteredSubject(filteredSubject string, subjects []string) bool {
 	return false
 }
 
-// SetInActiveDeleteThreshold sets the delete threshold for how long to wait
+// setInActiveDeleteThreshold sets the delete threshold for how long to wait
 // before deleting an inactive ephemeral consumer.
 func (o *consumer) setInActiveDeleteThreshold(dthresh time.Duration) error {
 	o.mu.Lock()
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: server/consumer.go [hunk 1]
```diff
@@ -1055,15 +1056,19 @@ func (o *consumer) deleteNotActive() {
...
 					ca := js.consumerAssignment(acc, stream, name)
 					js.mu.RUnlock()
-					if ca != nil {
-						s.Warnf("Consumer assignment not cleaned up, retrying")
-						meta.ForwardProposal(removeEntry)
+					if ca != nil && fs {
+						if fs {
+							s.Warnf("Consumer assignment not cleaned up, retrying")
+							meta.ForwardProposal(removeEntry)
+						}
```
ripienaar, line 1070: the if on 1066 already checks if `fs` is true, then 1067 again checks if `fs` is true and 1071 in a block where `fs` is always `true` sets it to `true`

server/filestore.go
```diff
@@ -5129,6 +5130,26 @@ func (o *consumerFileStore) encryptState(buf []byte) []byte {
...
+// Used to limit number of disk IO calls in flight since they could all be blocking an OS thread.
+// https://github.com/nats-io/nats-server/issues/2742
+var dios chan struct{}
+
+// Used to setup our simplistic counting semaphore using buffered channels.
+// golang.org's semaphore seemed a bit heavy.
+func init() {
+	// Minimum for blocking disk IO calls.
+	const minNIO = 4
+	nIO := runtime.GOMAXPROCS(0)
```
ripienaar, line 5142: I think we recommend a value of around 4 for this dont we? Is that really enough under heavy load? 

I am not sure how the raft layer behaves exactly but this will presumably not slow down handling of the general keep alive of raft membership right? But could slow down significantly committing data there I presume. Would that increase leader elections under very stressed situations?

server/consumer.go
derekcollison: Yea that one got messed up after several revisions, will fix.

derekcollison: ok fixed.

server/filestore.go
derekcollison: This is usually set to num cores/cpus, so on my machine its 20 (hyper-threads), but on smaller machines we can use up to 4 instead of setting to 1 or 2. I played with increasing this in some tests but makes no difference since the machine can only be really doing N things in parallel.

This is only effecting the consumer stores, not stream stores which is what is used for nrg/raft layer.

Under a loaded test I had yesterday, performance was much better when lots of ephemeral consumers were trying to do those things all at the same time, and of course the crazy number of threads took its toll on the OS. Did both darwin and linux, so feel pretty good after spending yesterday on it perf wise.

ripienaar: Yes, I know defaults but we often say nats server works best on 4ish cores and recommend setting this to 4, was thinking about that

ripienaar (approved):
LGTM

server/consumer.go
derekcollison: Actually redid this again, sorry. I had it mostly right before. Please take one more look, apologies.

ripienaar (approved):
LGTM

PR #2765, opened 2021-12-29
Avoid race condition

derekcollison:
Have seen this one a few times trigger race detector.

/cc @nats-io/core


ripienaar (approved):
LGTM

derekcollison:
I have a deadlock, will check this out in a few. We had this before, its coming back to me now ;)

derekcollison:
I kept second CL separate for now to see what was deadlocking that code.

ripienaar (approved):
LGTN

PR #2766, opened 2021-12-30
FIX: Fix broken link to monitoring documentation

kfabryczny:


derekcollison (approved):
Both work, but I like the simpler version. Thanks.

PR #2734, opened 2021-12-06
[IMPROVED] Websocket: added client IP from X-Forwarded-For header

kozlovic:
This is for cases when there is a proxy (Nginx, HAProxy, etc..)
between the client and the NATS Server. If this header is present,
the first IP is the one of the originating client and will be
used as the host/IP in server's representation of the client host.

Resolves #2514

Signed-off-by: Ivan Kozlovic <ivan@synadia.com>

PR #2769, opened 2022-01-03
Add X-Forwarded-For IP to the client's remote address

### LINKED ISSUE #2767: Forwarding headers should not be trusted by default
## Defect

**SECURITY PROBLEM** in _unreleased_ code.

_nb: I was asked for review on the original but didn't see the request until a sweep just now, so this is partly on me for missing it, sorry._

Issue #2734 adds support for `X-Forwarded-For:` headers.  It does this always.  This means that anyone wishing to hide their tracks just needs to include a bogus `X-Forwarded-For:` header.  This is a well-known technique for getting around ACL/logging systems which are too permissive.

UNRELEASED, but anything including #2734 

`git tag --contains 67c345270cd2b9ba8b62afc2eebee473199b07e2` confirms this is still not in a released version.

#### Expected result:

There should be a configuration block enabling parsing of `X-Forwarded-For:` (XFF) headers, providing a set of CIDR netblocks which as "trusted".  Each purported proxy in turn appends their own idea of the origin IP.  If Mallory includes "X-Forwarded-For: 192.0.2.42" in the original request, then that will still be there when it reaches the target server.

We have to start at the very end (our idea of the origin IP without the header) and check that, then check the header working backwards, checking that it's allowed to tell us of a different origin IP.  And yes, people do chain proxies, particularly once CDNs come into play with ingress and egress XFF stamping.

So the correct algorithm is, in pseudocode:

```
def trusted_forwarder(ip):
  for block in trusted_cidrs:
    if block.family != ip.family:
      continue
    if block.contains(ip):
      return True
  return False

def set_source_ip(request):
  request.source_addr = request.network_source_addr
  if not trusted_forwarder(request.network_source_addr):
    return
  xff = request.headers.find('x-forwarded-for')  # case-insensitive, and consider multiple headers, check spec
  if not xff:
    return
  purported_ips = reverse(httphdr_commalist_split(xff))
  while len(purported_ips) > 1 and trusted_forwarder(purported_ips[0]):
    purported_ips.drop_first()
  request.source_addr = purported_ips[0]
```

A CIDR netblock such as `192.0.2.0/24` is the sanest unit of configuration for a trusted IP, and you can consider treating bare IPs as as /32, (and then there's IPv6 too).  Usually there's a _short_ list of netblocks, at most 3 entries, so it's probably not worth using a trie or anything for matching: just iterate the very short list of configured netblocks.

#### Actual result:

The `X-Forwarded-For:` header is trusted always.

kozlovic:
Instead of replacing connection's host with value specified by
this header, we will simply add the address to the logging only.
So instead of having something like:
```
192.168.1.1:5678 - wid:10 - Client connection created
```
we could have:
```
1.2.3.4/192.168.1.1:5678 - wid:10 - Client connection created
```
As seen above, this PR simply prefixes the connection's remote address
with the header's value (if a valid IP).

Related to #2734
Resolves #2767

### DIFF
--- a/server/client.go
+++ b/server/client.go
@@ -610,12 +610,15 @@ func (c *client) initClient() {
 			if conn = addr.String(); conn != _EMPTY_ {
 				host, port, _ := net.SplitHostPort(conn)
 				iPort, _ := strconv.Atoi(port)
-				c.port = uint16(iPort)
+				c.host, c.port = host, uint16(iPort)
 				if c.isWebsocket() && c.ws.clientIP != _EMPTY_ {
-					c.host = c.ws.clientIP
-					conn = net.JoinHostPort(c.host, port)
-				} else {
-					c.host = host
+					cip := c.ws.clientIP
+					// Surround IPv6 addresses with square brackets, as
+					// net.JoinHostPort would do...
+					if strings.Contains(cip, ":") {
+						cip = "[" + cip + "]"
+					}
+					conn = fmt.Sprintf("%s/%s", cip, conn)
 				}
 				// Now that we have extracted host and port, escape
 				// the string because it is going to be used in Sprintf
--- a/server/websocket_test.go
+++ b/server/websocket_test.go
@@ -3989,12 +3989,12 @@ func TestWSXForwardedFor(t *testing.T) {
 			}
 			select {
 			case d := <-l.ch:
-				ipAndColumn := fmt.Sprintf("%s:", test.expectedValue)
+				ipAndSlash := fmt.Sprintf("%s/", test.expectedValue)
 				if test.useHdrValue {
-					if !strings.HasPrefix(d, ipAndColumn) {
-						t.Fatalf("Expected debug statement to start with: %q, got %q", ipAndColumn, d)
+					if !strings.HasPrefix(d, ipAndSlash) {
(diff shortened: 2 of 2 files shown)


server/client.go
```diff
@@ -610,12 +610,15 @@ func (c *client) initClient() {
...
-					conn = net.JoinHostPort(c.host, port)
-				} else {
-					c.host = host
+					cip := c.ws.clientIP
+					// Surround IPv6 addresses with square brackets, as
+					// net.JoinHostPort would do...
+					if strings.Contains(cip, ":") {
+						cip = "[" + cip + "]"
+					}
+					conn = fmt.Sprintf("%s/%s", cip, conn)
```
philpennock, line 621: Do we normally have reflection bounces inside the client connection execution path?

philpennock (approved):

server/client.go
kozlovic: Sorry, I am not sure what you are asking here. Is it about the use of fmt.Sprintf() (which we use also down below in original code)? The intent of the code above is to surround an IPv6 address with square brackets. The net.JoinHostAndPort() does exactly that.

philpennock: Yes.  I approved the PR, it's fine to merge, I was just pointing at this in case it's an issue so you could adjust it if desired.

PR #2771, opened 2022-01-06
Implement overflow placement for JetStream streams.

derekcollison:
This allows stream placement to overflow to adjacent clusters.
We also do more balanced placement based on resources (store or mem). We can continue to expand this as well.
We also introduce an account requirement that stream configs contain a MaxBytes value.

We now track account limits and server limits more distinctly, and do not reserver server resources based on account limits themselves.

/cc @nats-io/core

### DIFF
--- a/server/events.go
+++ b/server/events.go
@@ -1,4 +1,4 @@
-// Copyright 2018-2021 The NATS Authors
+// Copyright 2018-2022 The NATS Authors
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
 // You may obtain a copy of the License at
@@ -181,6 +181,7 @@ type ClientInfo struct {
 	RTT        time.Duration `json:"rtt,omitempty"`
 	Server     string        `json:"server,omitempty"`
 	Cluster    string        `json:"cluster,omitempty"`
+	Alternates []string      `json:"alts,omitempty"`
 	Stop       *time.Time    `json:"stop,omitempty"`
 	Jwt        string        `json:"jwt,omitempty"`
 	IssuerKey  string        `json:"issuer_key,omitempty"`
@@ -669,6 +670,14 @@ func (s *Server) sendStatsz(subj string) {
 		jStat.Config = &c
 		js.mu.RUnlock()
 		jStat.Stats = js.usageStats()
+		// Update our own usage since we do not echo so we will not hear ourselves.
+		ourNode := string(getHash(s.serverName()))
+		if v, ok := s.nodeToInfo.Load(ourNode); ok && v != nil {
+			ni := v.(nodeInfo)
+			ni.stats = jStat.Stats
+			s.nodeToInfo.Store(ourNode, ni)
+		}
+		// Metagroup info.
 		if mg := js.getMetaGroup(); mg != nil {
 			if mg.Leader() {
 				if ci := s.raftNodeToClusterInfo(mg); ci != nil {
@@ -695,9 +704,11 @@ func (s *Server) sendStatsz(subj string) {
 func (s *Server) heartbeatStatsz() {
 	if s.sys.stmr != nil {
 		// Increase after startup to our max.
-		s.sys.cstatsz *= 4
-		if s.sys.cstatsz > s.sys.statsz {
-			s.sys.cstatsz = s.sys.statsz
+		if s.sys.cstatsz < s.sys.statsz {
+			s.sys.cstatsz *= 2
+			if s.sys.cstatsz > s.sys.statsz {
+				s.sys.cstatsz = s.sys.statsz
+			}
 		}
 		s.sys.stmr.Reset(s.sys.cstatsz)
 	}
@@ -714,7 +725,7 @@ func (s *Server) sendStatszUpdate() {
 func (s *Server) startStatszTimer() {
 	// We will start by sending out more of these and trail off to the statsz being the max.
 	s.sys.cstatsz = 250 * time.Millisecond
-	// Send out the first one after 250ms.
+	// Send out the first one quickly, we will slowly back off.
 	s.sys.stmr = time.AfterFunc(s.sys.cstatsz, s.wrapChk(s.heartbeatStatsz))
 }
 
@@ -1031,10 +1042,10 @@ func (s *Server) processRemoteServerShutdown(sid string) {
 	})
 	// Update any state in nodeInfo.
 	s.nodeToInfo.Range(func(k, v interface{}) bool {
-		si := v.(nodeInfo)
-		if si.id == sid {
-			si.offline = true
-			s.nodeToInfo.Store(k, si)
+		ni := v.(nodeInfo)
+		if ni.id == sid {
+			ni.offline = true
+			s.nodeToInfo.Store(k, ni)
 			return false
 		}
 		return true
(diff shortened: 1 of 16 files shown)


server/jetstream_cluster.go
```diff
@@ -3375,6 +3385,30 @@ func (js *jetStream) processStreamAssignmentResults(sub *subscription, c *client
...
+		canDelete := !result.Update && time.Since(sa.Created) < 5*time.Second
+
+		// See if we should retry in case this cluster is full but there are others.
+		if cfg, ci := sa.Config, sa.Client; cfg != nil && ci != nil && isInsufficientResourcesErr(result.Response.Error) && canDelete {
+			// If cluster is defined we can not retry.
+			if cfg.Placement == nil || cfg.Placement.Cluster == _EMPTY_ {
+				// If we have additional clusters to try we can retry.
+				if ci != nil && len(ci.Alternates) > 0 {
+					if rg := js.createGroupForStream(ci, cfg); rg != nil {
+						s.Warnf("Retrying cluster placement for stream '%s > %s'", result.Account, result.Stream)
```
ripienaar, line 3397: we should probably increment some stat counter when this happens and maybe send an advisory, and maybe clarify why this being retried here rather than just saying its being retried.  And include where the placement is being done instead in the log (and advisory)

server/jetstream_cluster.go
```diff
@@ -3609,22 +3686,33 @@ func groupName(prefix string, peers []string, storage StorageType) string {
...
+		clusters = append(clusters, ci.Alternates...)
+	}
+
 	// Need to create a group here.
-	// TODO(dlc) - Can be way smarter here.
-	peers := cc.selectPeerGroup(replicas, cluster)
-	if len(peers) == 0 {
-		return nil
+	for _, cn := range clusters {
+		peers := cc.selectPeerGroup(replicas, cn, cfg)
```
ripienaar, line 3709: Fine for now I guess but in real world I think we'd need a lot more control here - probably once tag placement is in. 

But like even now it would be pretty bad if the next nearest cluster with available space is lets say Asia while I am connected to USA?

server/events.go
```diff
@@ -181,6 +181,7 @@ type ClientInfo struct {
 	RTT        time.Duration `json:"rtt,omitempty"`
 	Server     string        `json:"server,omitempty"`
 	Cluster    string        `json:"cluster,omitempty"`
+	Alternates []string      `json:"alts,omitempty"`
```
ripienaar, line 184: `alts` is a bit unclear, hard to guess the meaning for someone who dont know the code

server/jetstream.go
```diff
@@ -48,22 +48,22 @@ type JetStreamConfig struct {
...
-	ReservedMemoryUsed uint64            `json:"reserved_memory_used,omitempty"`
-	ReserveStoreUsed   uint64            `json:"reserved_storage_used,omitempty"`
-	Accounts           int               `json:"accounts,omitempty"`
-	API                JetStreamAPIStats `json:"api"`
-	ReservedMemory     uint64            `json:"reserved_memory,omitempty"`
-	ReservedStore      uint64            `json:"reserved_storage,omitempty"`
+	Memory         uint64            `json:"memory"`
+	Store          uint64            `json:"storage"`
+	ReservedMemory uint64            `json:"reserved_memory"`
+	ReservedStore  uint64            `json:"reserved_storage"`
```
ripienaar, line 56: Can these changes really be done at this point given semver?

server/jetstream_cluster.go
derekcollison: We have plans on continuing to extend the resource utilization used in the decision, right now this is a step up for random placement.

Also note that the system will not overflow if you provide placement guidance, so you can opt-out and have the stream create request fail.

server/events.go
derekcollison: I think you will need sample code or docs for this. I tried other names to make it a bit more clear but that did not seem to help much tbh.

derekcollison: ClientInfo has not been documented formally yet either. We have been watching and adapting its use for JetStream, even though it is a general construct across service export boundaries.

server/jetstream.go
derekcollison: We discussed yesterday, not totally sure yet, but the reserved portions were incorrect and needed to be fixed. The one issue identified would be the nats cli most likely.

server/jetstream_cluster.go
ripienaar: ok, maybe just some extra info in the log then to indicate its this, a user who come across this for the first time would not have any idea what it relates to

derekcollison: Possibly if they were doing no monitoring at all of their system to see it ran out of resources. What would you add to the description that you think would be helpful in a system like NGS? Or any super cluster tbh.

ripienaar: ```Retrying cluster placement for stream 'X > Y'` due to insufficient resources in cluster C``?

derekcollison: That works, thanks.

derekcollison: Note though that this is the not happy path where the meta-leader guesses wrong around placement, either due to concurrent requests, dropped updates around reserved or real usage etc. 

In the happy path we have no logging at all atm.

server/jetstream.go
kozlovic: Is it deliberate that you removed the `,omitempty`? (for both ReservedMemory and ReservedStore)

server/jetstream_api.go
```diff
@@ -817,7 +817,9 @@ func (a *Account) trackAPI() {
 		jsa.usage.api++
 		jsa.apiTotal++
 		jsa.sendClusterUsageUpdate()
+		js := jsa.js
 		jsa.mu.Unlock()
+		atomic.AddInt64(&js.apiTotal, 1)
```
kozlovic, line 822: Since this is atomic op, you don't need to capture `js := jsa.js` and do this here outside of `jsa`'s lock. So I would move this to after the line `jsa.apiTotal++` (before sendClusterUsageUpdate())

server/jetstream_api.go
```diff
@@ -834,6 +836,7 @@ func (a *Account) trackAPIErr() {
 		jsa.sendClusterUsageUpdate()
 		js := jsa.js
 		jsa.mu.Unlock()
+		atomic.AddInt64(&js.apiTotal, 1)
```
kozlovic, line 839: Same for those two atomic ops. I don't think that they need to be done outside of jsa's lock.

server/jetstream_cluster.go
```diff
@@ -3562,31 +3598,77 @@ func (cc *jetStreamCluster) remapStreamAssignment(sa *streamAssignment, removePe
...
-	var nodes []string
-	peers := cc.meta.Peers()
-	s := cc.s
+func (cc *jetStreamCluster) selectPeerGroup(r int, cluster string, cfg *StreamConfig) []string {
+	if cluster == _EMPTY_ || cfg == nil {
+		return nil
+	}
+
+	var maxBytes uint64
+	if cfg != nil && cfg.MaxBytes > 0 {
```
kozlovic, line 3607: You already checked for `cfg == nil` at the top of the function, so no need here.

kozlovic (changes requested):

server/jetstream_api.go
derekcollison: Fixed.

server/jetstream.go
kozlovic: Could you comment on this question: "Is it deliberate that you removed the ,omitempty? (for both ReservedMemory and ReservedStore)"

derekcollison: Yes it was deliberate based on feedback and my usage from varz/jsz and statusz stuff.

kozlovic: Ok

kozlovic (approved):
LGTM

kozlovic (approved):
LGTM

PR #2775, opened 2022-01-10
Replace use of go channels with queues

kozlovic:


server/events.go
```diff
@@ -248,6 +248,22 @@ type pubMsg struct {
...
+	msg interface{}, oct compressionType, echo, last bool) *pubMsg {
+
+	var m *pubMsg
+	pm := pubMsgPool.Get()
+	if pm != nil {
+		m = pm.(*pubMsg)
+	} else {
+		m = &pubMsg{}
+	}
+	(*m) = pubMsg{c, sub, rply, si, hdr, msg, oct, echo, last}
```
derekcollison, line 263: Interesting approach with the temp object. Cool.

server/events.go
```diff
@@ -404,6 +420,7 @@ RESET:
 					c.flushClients(time.Second)
 					return
 				}
+				pubMsgPool.Put(pmi)
```
derekcollison, line 423: If there are interior items that are on the heap and could affect the GC I might set those to nil.
I do this in the current PR I have up as well.

derekcollison (approved):
LGTM

server/events.go
kozlovic: I should have also put a comment that I do this way because setting all fields of an object when using pool is CRITICAL, and doing this way we would get a compiler error if a field is added to the pubMsg struct without updating this function.

server/events.go
kozlovic: I was thinking about this and may then do a pubMsg.done() that will reset some (all?) fields before putting back to the pull.

derekcollison: Only the ones that are externally stored. Yes.

kozlovic: Done. Did it for strings too, not sure if required...

derekcollison: Yes on strings IMO, so that is good.

derekcollison (approved):
LGTM

PR #2776, opened 2022-01-12
Pull Consumer Updates

derekcollison:
This change cleans up alot of the inner workings of pull consumers to be more resilient, better at waiting request management, and the ability to determine and watch interest cross account.

We also introduce new consumer config items to limit next msg requests' batch size and maximum wait times as well as indicate inactive thresholds.

We also now allow pull consumers that are ephemeral (they will auto-cleanup) and can have ack-all, not just ack explicit.

/cc @nats-io/core


matthiashanel (approved):
LGTM
Change looks ok. But not super familiar, and a second set of eyes would be good

server/consumer.go
```diff
@@ -1698,6 +1737,7 @@ func (o *consumer) info() *ConsumerInfo {
 
 	// If we are a pull mode consumer, report on number of waiting requests.
 	if o.isPullMode() {
+		o.expireWaiting()
```
wallyqs, line 1740: Does this mean that doing `sub.ConsumerInfo()` will expire the pending requests?

derekcollison: Yes it will reclaim old requests here, and when we move to a wait state in the loop and gather, and we check them when we pop to deliver a message.

server/consumer.go
```diff
@@ -68,6 +65,17 @@ type ConsumerConfig struct {
 	FlowControl     bool          `json:"flow_control,omitempty"`
 	HeadersOnly     bool          `json:"headers_only,omitempty"`
 
+	// Pull based options.
+	MaxRequestBatch   int           `json:"max_batch,omitempty"`
+	MaxRequestExpires time.Duration `json:"max_expires,omitempty"`
```
ripienaar, line 70: what are your thoughts about defaults for these? and maximums? Definitely think these need sane defaults, eventhough clients are supposed to set defaults here and not allow uncapped I think its a bad default to allow uncapped as per recent mails

server/consumer.go
```diff
@@ -634,9 +644,17 @@ func (mset *stream) addConsumerWithAssignment(config *ConsumerConfig, oname stri
 	o.ackSubj = fmt.Sprintf("%s.*.*.*.*.*", pre)
 	o.nextMsgSubj = fmt.Sprintf(JSApiRequestNextT, mn, o.name)
 
+	// If not durable determine the inactive threshold.
+	if !o.isDurable() {
+		if o.cfg.InactiveThreshold != 0 {
+			o.dthresh = o.cfg.InactiveThreshold
```
ripienaar, line 650: Would it make sense to constrain minimum inactive threshold around max ack time? They should expire some time after the last ack is expected, important for WQs etc

server/consumer.go
```diff
@@ -2007,31 +2078,65 @@ func (wq *waitQueue) pop() *waitingRequest {
...
+	// Check if we are empty.
+	if wq.rp == wq.wp {
+		wq.rp, wq.wp = -1, 0
+	}
+}
+
+// Return next waiting request. This will check for expirations but not noWait or interest.
+// That will be handled by expireWaiting.
+// Lock should be held.
+func (o *consumer) nextWaiting() *waitingRequest {
```
ripienaar, line 2103: guessing the fairer distribution of messages will come in follow up PRs right

server/consumer.go
```diff
@@ -1027,26 +1051,41 @@ func (o *consumer) updateDeliveryInterest(localInterest bool) bool {
...
+	o.mu.Lock()
 	if o.mset == nil {
-		o.mu.RUnlock()
+		o.mu.Unlock()
 		return
 	}
+	// Push mode just look at active.
+	if o.isPushMode() {
+		// If weare active simply return.
+		if o.active {
```
ripienaar, line 1062: I might read wrong but active isnt yet cross account aware?

server/consumer.go
tpihl: If this ack can be extended, that activity should also extend the inactivity timer(?)

server/consumer.go
derekcollison: If the request still remains it is legit, so not sure about default limits.

The other issue of fairness is still a consideration.

server/consumer.go
derekcollison: Pull consumer logic is different, if we have seen a request in that window we are good, if not attempt to expire all invalid requests and if empty, then we consider inactive.

server/consumer.go
derekcollison: Yes needs more thought.

server/consumer.go
derekcollison: Correct, for push we can not use same logic since its a stream.

server/consumer.go
ripienaar: OK, well we can park it till fairness is clearer I suppose as it relates to that ultimately

server/consumer.go
wallyqs: I think this is good approach, it was needed some way to trigger the expiration of accumulated pending requests that have just timed out without a client issuing a 'get next' request. Now when the state is confirmed by using `nats consumer info` for example, the expired requests would no longer be accounted in the `NumWaiting` field.

derekcollison: We used to expire pending requests sporadically and not on a regular time interval. I did not opt for a time interval per se but added a bunch of spots to where we do collect now.

1. Namely when we pop a request we now check inline, so no delivering to a request that we invalidate.
2. When the consumer's loopAndGather goes into a waiting state we check the pending list then as well.
3. When someone asks for consumer info on a pull consumer.
4.

server/consumer.go
derekcollison: We do not follow that philosophy with ephemerals today. E.g. push consumers if the delivery subject loses interest for some period of time (5s default IIRC) we reclaim the consumer, regardless of pending acks.

This stays consistent but we could consider adding this, but I feel pretty strong we would not want to do that.

One app sends ack that gets lost and then disconnects, we will never clean up an ephemeral pull consumer if the logic requires the acks.

server/consumer.go
```diff
@@ -1027,26 +1051,41 @@ func (o *consumer) updateDeliveryInterest(localInterest bool) bool {
...
-	o.mu.RLock()
+	o.mu.Lock()
 	if o.mset == nil {
-		o.mu.RUnlock()
+		o.mu.Unlock()
 		return
 	}
+	// Push mode just look at active.
+	if o.isPushMode() {
+		// If weare active simply return.
```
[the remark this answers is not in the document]
derekcollison: Thx will fix and force push this one.

server/consumer.go
ripienaar: Not “never” just up to max ack waiting time.

derekcollison: Possibly but that would be inconsistent with how we do that today for push. So prefer to keep them the same for now and add in that logic if needed to a later date.

Also, if I am thinking about this correctly, if we only wait an additional max wait but take no other action, this feels like just reclaiming the consumer since we don't wait for non acked messages. IMO..

ripienaar: Yeah - my thinking is if you are processing messages from a WQ removing the consumer too early while there are still messages in flight could result in processed messages not being removed from tbe stream. 

Agree it’s something we can park for a later time but seems important.

derekcollison: Good point but if its WQ and the consumer goes away that releases all of the messages anyway that were being held for that consumer.

kozlovic (approved):
LGTM

PR #2777, opened 2022-01-12
[ADDED] Ability to express the max_file_store and max_memory_store as a string (100M, etc..)

### LINKED ISSUE #2754: The "max_file_store" option cannot be quoted
## Defect

The server does not process a quoted string of the `max_file_store` option.  It'd be great it if could.

```bash
$ cat quoted.conf 
jetstream {
  max_file_store: "200M"
}

$ nats-server -c quoted.conf
nats-server: quoted.conf:1:0: interface conversion: interface {} is string, not int64
```

Others like this should be checked as well...

hooksie1:
Resolves #2754

 - Adds a function to check if max_file_store is a string or int64. Possibly could be reused again.

/cc @nats-io/core

### DIFF
--- a/server/opts.go
+++ b/server/opts.go
@@ -1680,26 +1680,22 @@ func parseJetStreamForAccount(v interface{}, acc *Account, errors *[]error, warn
 	return nil
 }
 
-func getInterfaceValue(v interface{}) (int64, error) {
+// takes in a storage size as either an int or a string and returns an int64 value based on the input.
+func getStorageSize(v interface{}) (int64, error) {
 	_, ok := v.(int64)
 	if ok {
 		return v.(int64), nil
 	}
 
-	_, ok = v.(string)
+	s, ok := v.(string)
 	if !ok {
 		return 0, fmt.Errorf("must be int64 or string")
 	}
 
-	num, err := stringStorageSize(v.(string))
-	if err != nil {
-		return 0, err
+	if s == "" {
+		return 0, nil
 	}
 
-	return int64(num), nil
-}
-
-func stringStorageSize(s string) (int, error) {
 	suffix := strings.ToUpper(s[len(s)-1:])
 	prefix := s[:len(s)-1]
 	var num int
@@ -1710,13 +1706,13 @@ func stringStorageSize(s string) (int, error) {
 		num = num * 1024
 	case "M":
 		num, err = strconv.Atoi(prefix)
-		num = num * 1048576
+		num = num * 1024 * 1024
 	case "G":
 		num, err = strconv.Atoi(prefix)
-		num = num * 1073741824
+		num = num * 1024 * 1024 * 1024
 	case "T":
 		num, err = strconv.Atoi(prefix)
-		num = num * 1099511627776
+		num = num * 1024 * 1024 * 1024 * 1024
 	default:
 		return 0, fmt.Errorf("sizes defined as strings must end in K, M, G, T")
 	}
@@ -1724,7 +1720,7 @@ func stringStorageSize(s string) (int, error) {
 		return 0, err
 	}
 
-	return num, nil
+	return int64(num), nil
 }
 
 // Parse enablement of jetstream for a server.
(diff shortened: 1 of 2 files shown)


server/opts.go
```diff
@@ -1680,6 +1680,53 @@ func parseJetStreamForAccount(v interface{}, acc *Account, errors *[]error, warn
...
+	num, err := stringStorageSize(v.(string))
+	if err != nil {
+		return 0, err
+	}
+
+	return int64(num), nil
+}
+
+func stringStorageSize(s string) (int, error) {
+	suffix := strings.ToUpper(s[len(s)-1:])
```
kozlovic, line 1703: This would panic if string `s` is empty. You can reproduce with:
```
jetstream {
  max_file_store: ""
}
```
You may want to add to the test.

server/opts.go
```diff
@@ -1680,6 +1680,53 @@ func parseJetStreamForAccount(v interface{}, acc *Account, errors *[]error, warn
 	return nil
 }
 
+func getInterfaceValue(v interface{}) (int64, error) {
```
kozlovic, line 1683: I feel that this function name is too generic and does not give any indication that we are going to get the value for a int64 or a "storage size". Can we have a better name?
Possibly you could "collapse" the two functions into one and be something like `getByteSize()` or `getStorageSize()` (with a line of doc that says what it does)

server/opts.go
```diff
@@ -1680,6 +1680,53 @@ func parseJetStreamForAccount(v interface{}, acc *Account, errors *[]error, warn
...
+	prefix := s[:len(s)-1]
+	var num int
+	var err error
+	switch suffix {
+	case "K":
+		num, err = strconv.Atoi(prefix)
+		num = num * 1024
+	case "M":
+		num, err = strconv.Atoi(prefix)
+		num = num * 1048576
```
kozlovic, line 1713: I would have done `1024*1024` (same for others). The compiler will do the operation once, but it is easier to read for a dev.

server/opts.go
```diff
@@ -1714,7 +1761,12 @@ func parseJetStream(v interface{}, opts *Options, errors *[]error, warnings *[]e
 				opts.JetStreamMaxMemory = mv.(int64)
 				opts.maxMemSet = true
 			case "max_file_store", "max_file":
-				opts.JetStreamMaxStore = mv.(int64)
+				// need to define error to avoid non-name on left side of :=
```
kozlovic, line 1764: Yes, or you could do (assuming the function is renamed)
```
sz, err := getStorageSize(mv)
if err != nil {
 ...
}
opts.JetStreamMaxStore = sz
..
```

server/opts_test.go
```diff
@@ -3256,3 +3256,31 @@ func TestMaxSubTokens(t *testing.T) {
...
+	tt := []struct {
+		input string
+		want  int64
+		err   bool
+	}{
+		{"1K", 1024, false},
+		{"1M", 1048576, false},
+		{"1G", 1073741824, false},
+		{"1T", 1099511627776, false},
+		{"1L", 0, true},
```
kozlovic, line 3270: May want to add `{"", 0, false}` to the list?

server/opts.go
```diff
@@ -1714,7 +1761,12 @@ func parseJetStream(v interface{}, opts *Options, errors *[]error, warnings *[]e
...
+				var err error
```
kozlovic, line 1765: Also, I understand that you said that we could then use that other places, we could start with the `max_memory_store` just above, at least this configuration block would be consistent in ability to parse string byte sizes.

kozlovic (changes requested):

server/opts.go
hooksie1: yeah good catch

server/opts.go
hooksie1: good point, I'll change the name to reflect.

server/opts.go
hooksie1: yeah, I wasn't sure which was better and guessed wrong 😛 I"ll change it to the other way.

server/opts.go
hooksie1: yeah switched to this method.

server/opts_test.go
hooksie1: added.

server/opts.go
hooksie1: sounds good. I didn't want to change everything and this not be an approach that was wanted. I'll change that one as well.

server/opts.go
```diff
@@ -1680,6 +1680,49 @@ func parseJetStreamForAccount(v interface{}, acc *Account, errors *[]error, warn
...
+		return 0, fmt.Errorf("must be int64 or string")
+	}
+
+	if s == "" {
+		return 0, nil
+	}
+
+	suffix := strings.ToUpper(s[len(s)-1:])
+	prefix := s[:len(s)-1]
+	var num int
```
kozlovic, line 1701: I would actually rewrite from here to the end something like that:
```
num, err := strconv.ParseInt("1", 10, 64)
if err != nil {
    return 0, err
}
suffixMap := map[string]int64{"K": 10, "M": 20, "G": 30, "T": 40}
mult, ok := suffixMap[suffix]
if !ok {
    return 0, fmt.Errorf("sizes defined as strings must end in K, M, G, T")
}
num *= 1 << mult
return num, nil
```
At any rate, even if we were to keep the switch, there is no need to parse the prefix in each "case" line, and also, we want to use the 64bit version so that `num` is not a 32bit if server is built as 32bit and would overflow with some of the units.

kozlovic (changes requested):
Sorry for the requests to change again, but I did not notice it during the first review.

server/opts.go
hooksie1: Ah great points. Yeah let me redo that.

hooksie1:
No problem at all, that's much cleaner than what I had initially.  Thanks for the really helpful input!

kozlovic (approved):
LGTM

kozlovic:
@hooksie1 Thank you for the contribution!

PR #2779, opened 2022-01-13
[Adding] support for JS MaxBytesRequired

matthiashanel:
updates jwt library to pick up last commit on main.
also test if value is set.


derekcollison (approved):
LGTM

PR #2780, opened 2022-01-13
[IMPROVED] JetStream clustering with lots of streams/consumers

kozlovic:
Some operations could cause the route to block due to lock being
held during store operations. On macOS, having lots of streams/consumers
and restarting the cluster would cause lots of concurrent IO that
would cause lock to be held for too long, causing head-of-line
blocking in processing of messages from a route.


derekcollison (approved):
LGTM

PR #2781, opened 2022-01-13
When rebuilding the complete filestore state we need to do this in a go routine.

derekcollison:

/cc @nats-io/core


wallyqs (approved):
LGTM

PR #2782, opened 2022-01-13
Start monitoring before JetStream during startup.

wallyqs:
Make the http monitoring endpoint available before starting JetStream to be able to monitor whether it is available during startup and recovery of streams.

/cc @nats-io/core


derekcollison (approved):
LGTM

PR #2787, opened 2022-01-13
Prepare for release v2.7.0, start with v2.7.0-rc1

kozlovic:
Since we have never released with goreleaser and GO111MODULE=on
and have removed vendor directory, we are going to push a
v2.7.0-rc1 tag. If the goreleaser process works fine and submit
the draft release correctly, we will not do an actual v2.7.0-rc1
release. Instead, we will follow up immediately with the v2.7.0
tag and the proper release.


derekcollison (approved):
LGTM

PR #2791, opened 2022-01-15
[ADDED] Replication lag to stream monitoring

### LINKED ISSUE #2778: Metrics for replication lag
## Feature Request
I would like metrics for replication lag.

#### Use Case:
There are issues we've seen where source replication lag spikes when looking at the output of `nats stream report`. We'd like to be able to alert on these in our monitoring infrastructure, but that's really difficult without getting the lag data itself. We've summed the stream message counts, but that can be a really sensitive alert.

#### Proposed Change:
Add replication lag metrics.

Anyone who needs to monitor jetstream replication.

#### Alternative Approaches
We can sum total messages for streams and compare it to the stream that is being replicated into, but this seems really clunky. It works for mirroring just fine, so that's not as much of a concern right now.

mattstep:
 - [x] Tests added

[Resolves #2778 ](https://github.com/nats-io/nats-server/issues/2778)

- Adds ReplicationLagInfo list to StreamDetail
- Adds option to populate ReplicationLagInfo

/cc @nats-io/core

### DIFF
--- a/server/monitor.go
+++ b/server/monitor.go
@@ -2388,11 +2388,13 @@ type JSzOptions struct {
 }
 
 type StreamDetail struct {
-	Name     string          `json:"name"`
-	Cluster  *ClusterInfo    `json:"cluster,omitempty"`
-	Config   *StreamConfig   `json:"config,omitempty"`
-	State    StreamState     `json:"state,omitempty"`
-	Consumer []*ConsumerInfo `json:"consumer_detail,omitempty"`
+	Name     string              `json:"name"`
+	Cluster  *ClusterInfo        `json:"cluster,omitempty"`
+	Config   *StreamConfig       `json:"config,omitempty"`
+	State    StreamState         `json:"state,omitempty"`
+	Consumer []*ConsumerInfo     `json:"consumer_detail,omitempty"`
+	Mirror   *StreamSourceInfo   `json:"mirror,omitempty"`
+	Sources  []*StreamSourceInfo `json:"sources,omitempty"`
 }
 
 type AccountDetail struct {
@@ -2469,6 +2471,8 @@ func (s *Server) accountDetail(jsa *jsAccount, optStreams, optConsumers, optCfg 
 				State:   stream.state(),
 				Cluster: ci,
 				Config:  cfg,
+				Mirror:  stream.mirrorInfo(),
+				Sources: stream.sourcesInfo(),
 			}
 			if optConsumers {
 				for _, consumer := range stream.getPublicConsumers() {
--- a/server/monitor_test.go
+++ b/server/monitor_test.go
@@ -4056,6 +4056,14 @@ func TestMonitorJsz(t *testing.T) {
 		Replicas: 1,
 	})
 	require_NoError(t, err)
+	_, err = js.AddStream(&nats.StreamConfig{
+		Name:     "my-stream-mirror",
+		Replicas: 2,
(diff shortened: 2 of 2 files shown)


mattstep:
I'll open a related PR against `prometheus-nats-exporter` if this looks good.

ripienaar:
Maybe better to ensure these fields from stream info is populated for the monioring results

https://github.com/nats-io/nats-server/blob/297b44394f92e990da450178cffaa7adecc88837/server/stream.go#L105-L106

That way the structure between stream info and this data doesnt change and data parsing can be used on both. wdyt @derekcollison ?

mattstep:
> Maybe better to ensure these fields from stream info is populated for the monioring results
> 
> https://github.com/nats-io/nats-server/blob/297b44394f92e990da450178cffaa7adecc88837/server/stream.go#L105-L106
> 
> That way the structure between stream info and this data doesnt change and data parsing can be used on both. wdyt @derekcollison ?

I'm still new to the codebase, but it just looks to me like the config and stats are returned separately for monitoring.

server/monitor.go
```diff
@@ -2377,22 +2377,30 @@ func (s *Server) accountInfo(accName string) (*AccountInfo, error) {
...
-	Cluster  *ClusterInfo    `json:"cluster,omitempty"`
-	Config   *StreamConfig   `json:"config,omitempty"`
-	State    StreamState     `json:"state,omitempty"`
-	Consumer []*ConsumerInfo `json:"consumer_detail,omitempty"`
+	Name           string                `json:"name"`
+	Cluster        *ClusterInfo          `json:"cluster,omitempty"`
+	Config         *StreamConfig         `json:"config,omitempty"`
+	State          StreamState           `json:"state,omitempty"`
+	Consumer       []*ConsumerInfo       `json:"consumer_detail,omitempty"`
+	ReplicationLag []*ReplicationLagInfo `json:"replication_lag"`
```
wallyqs, line 2397: Maybe add omitempty like the others just in case?

```suggestion
	ReplicationLag []*ReplicationLagInfo `json:"replication_lag,omitempty"`
```

mattstep: Done! I also added it to the fields of ReplicationLagInfo, though they really shouldn't be empty, I just saw that was the common way.

server/monitor.go
```diff
@@ -2377,22 +2377,25 @@ func (s *Server) accountInfo(accName string) (*AccountInfo, error) {
...
-	Limit      int    `json:"limit,omitempty"`
+	Account     string `json:"account,omitempty"`
+	Accounts    bool   `json:"accounts,omitempty"`
+	Streams     bool   `json:"streams,omitempty"`
+	Consumer    bool   `json:"consumer,omitempty"`
+	Config      bool   `json:"config,omitempty"`
+	LeaderOnly  bool   `json:"leader_only,omitempty"`
+	Offset      int    `json:"offset,omitempty"`
+	Limit       int    `json:"limit,omitempty"`
+	Replication bool   `json:"replication,omitempty"`
```
ripienaar, line 2388: We can probably just assume giving replication details are needed if stream details were requested I think.

This looks a lot nicer and smaller than before well done.

mattstep: Sounds good, I was just flagging it because it might be a performance issue, if not, that's easier! Updated.

ripienaar (approved):
LGTM

kozlovic (approved):
LGTM

PR #2803, opened 2022-01-20
[FIXED] Possible deadlock

kozlovic:
This is due to a re-entrant RLock(). It works sometimes, but if there
is a go routine requesting the write lock, then the second RLock()
will not be granted which will lead to a deadlock.
In summary: one should never make re-entrant RLock calls.


derekcollison (approved):
LGTM

PR #2811, opened 2022-01-21
Do not require CONNECT when no_auth_user defined, don't complain about password for internal clients.

### LINKED ISSUE #2810: server configuration using accounts and nkeys for authentication emits plaintext password warning
When a configuration as the following is used, the server emits a plain text warning

```
system_account: SYS
accounts: {
  SYS: {
    users: [
  	  { nkey: UBDLZ6YYFQVINKWTYJFF5BGJ6I3JW2WO6WHZ4EVBAEQG3G7W3STGLIMZ }
  	]
  }
}
```

> [92763] 2022/01/21 10:35:57.803364 [INF] Starting nats-server
[92763] 2022/01/21 10:35:57.803526 [INF]   Version:  2.6.6-beta
[92763] 2022/01/21 10:35:57.803529 [INF]   Git:      [not set]
[92763] 2022/01/21 10:35:57.803539 [INF]   Name:     NBGJ36EW3TMD4QQ2TXQQQUWIL6TY2XD2RZU2ZH6XZWUVY4DY6NHPJRNR
[92763] 2022/01/21 10:35:57.803541 [INF]   ID:       NBGJ36EW3TMD4QQ2TXQQQUWIL6TY2XD2RZU2ZH6XZWUVY4DY6NHPJRNR
__[92763] 2022/01/21 10:35:57.803546 [WRN] Plaintext passwords detected, use nkeys or bcrypt__

When not using accounts no such message is emitted:
```
authorization: {
	users: [
	  { nkey: UDXU4RCSJNZOIQHZNWXHXORDPRTGNJAHAHFRGZNEEJCPQTT2M7NLCNF4 }
	]
}
```

> [92820] 2022/01/21 10:40:11.444052 [INF] Starting nats-server
[92820] 2022/01/21 10:40:11.444150 [INF]   Version:  2.6.6-beta
[92820] 2022/01/21 10:40:11.444153 [INF]   Git:      [not set]
[92820] 2022/01/21 10:40:11.444156 [INF]   Name:     NBRYFWGB6NPYV5R3MYPJSUPQ5B5O6355PR4ENGANQJEDFLS7AFMK3WKP
[92820] 2022/01/21 10:40:11.444158 [INF]   ID:       NBRYFWGB6NPYV5R3MYPJSUPQ5B5O6355PR4ENGANQJEDFLS7AFMK3WKP
[92820] 2022/01/21 10:40:11.444162 [INF] Using configuration file: /tmp/ns.conf
[92820] 2022/01/21 10:40:11.444804 [INF] Listening for client connections on 0.0.0.0:4222
[92820] 2022/01/21 10:40:11.445002 [INF] Server is ready

derekcollison:
Resolves #2810 and also allows telnet demo for protocol to demo.nats.io to work again.
 
/cc @nats-io/core

### DIFF
--- a/server/auth_test.go
+++ b/server/auth_test.go
@@ -282,7 +282,7 @@ func TestNoAuthUserNoConnectProto(t *testing.T) {
 		accounts {
 			A { users [{user: "foo", password: "pwd"}] }
 		}
-		authorization { timeout: 1s }
+		authorization { timeout: 1 }
 		no_auth_user: "foo"
 	`))
 	defer os.Remove(conf)
--- a/server/auth_test.go
+++ b/server/auth_test.go
@@ -15,11 +15,13 @@ package server
 
 import (
 	"fmt"
+	"net"
 	"net/url"
 	"os"
 	"reflect"
 	"strings"
 	"testing"
+	"time"
 
 	"github.com/nats-io/jwt/v2"
 	"github.com/nats-io/nats.go"
@@ -272,4 +274,47 @@ func TestNoAuthUser(t *testing.T) {
 			}
 		})
 	}
+}
+
+func TestNoAuthUserNoConnectProto(t *testing.T) {
+	conf := createConfFile(t, []byte(`
+		listen: "127.0.0.1:-1"
+		accounts {
+			A { users [{user: "foo", password: "pwd"}] }
+		}
+		authorization { timeout: 1s }
+		no_auth_user: "foo"
+	`))
+	defer os.Remove(conf)
+	s, o := RunServerWithConfig(conf)
+	defer s.Shutdown()
+
+	checkClients := func(n int) {
+		t.Helper()
+		time.Sleep(100 * time.Millisecond)
+		if nc := s.NumClients(); nc != n {
+			t.Fatalf("Expected %d clients, got %d", n, nc)
+		}
+	}
+
+	conn, err := net.Dial("tcp", fmt.Sprintf("%s:%d", o.Host, o.Port))
+	require_NoError(t, err)
+	defer conn.Close()
+	checkClientsCount(t, s, 1)
+
+	// With no auth user we should not require a CONNECT.
+	// Make sure we are good on not sending CONN first.
+	_, err = conn.Write([]byte("PUB foo 2\r\nok\r\n"))
+	require_NoError(t, err)
+	checkClients(1)
+	conn.Close()
+
+	// Now make sure we still do get timed out though.
+	conn, err = net.Dial("tcp", fmt.Sprintf("%s:%d", o.Host, o.Port))
+	require_NoError(t, err)
+	defer conn.Close()
+	checkClientsCount(t, s, 1)
+
+	time.Sleep(1200 * time.Millisecond)
+	checkClientsCount(t, s, 0)
 }
--- a/server/parser.go
+++ b/server/parser.go
@@ -147,8 +147,7 @@ func (c *client) parse(buf []byte) error {
 	// proper CONNECT if needed.
 	authSet := c.awaitingAuth()
 	// Snapshot max control line as well.
-	mcl := c.mcl
-	trace := c.trace
+	s, mcl, trace := c.srv, c.mcl, c.trace
 	c.mu.Unlock()
 
 	// Move to loop instead of range syntax to allow jumping of i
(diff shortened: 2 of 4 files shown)


server/parser.go
```diff
@@ -160,7 +159,24 @@ func (c *client) parse(buf []byte) error {
...
+					var ok bool
+					// Check here for NoAuthUser. If this is set allow non CONNECT protos as our first.
+					// E.g. telnet proto demos.
+					if noAuthUser := s.getOpts().NoAuthUser; noAuthUser != _EMPTY_ {
+						s.mu.Lock()
+						user, exists := s.users[noAuthUser]
+						s.mu.Unlock()
+						if exists {
+							c.RegisterUser(user)
+							ok = true
```
kozlovic, line 174: For clients that do never send CONNECT, we will be in this code for every protocol message, since we don't clear the `authSet`. This will cause multiple registrations, no?

I am not convinced that we should allow clients without a CONNECT. The no_auth_user simply means that if no user is specified in the CONNECT, use the one configured, but it does not mean that clients should not send a CONNECT.

Also, your new test checks that the client gets disconnected anyway after the auth timeout. So what is the point of all that? Why let a client do pub/sub for the time between it TCP connects and the auth timeout?

kozlovic (changes requested):

server/parser.go
derekcollison: For the protocol demo and doing telnet. Agree I should clear the authSet though. Will update.

server/auth_test.go
```diff
@@ -273,3 +275,46 @@ func TestNoAuthUser(t *testing.T) {
...
 	}
 }
+
+func TestNoAuthUserNoConnectProto(t *testing.T) {
+	conf := createConfFile(t, []byte(`
+		listen: "127.0.0.1:-1"
+		accounts {
+			A { users [{user: "foo", password: "pwd"}] }
+		}
+		authorization { timeout: 1s }
```
kozlovic, line 285: Actually, duration is not supported here, so it has to be a float (or simply 1).

kozlovic (changes requested):

server/auth_test.go
derekcollison: Did not complain, and properly set it to 1s vs 2s. So I think it worked.

kozlovic: It does not, just tried with `10s` in a manual test and got kicked out quickly. Probably defaulted to 1 sec

kozlovic: https://github.com/nats-io/nats-server/blob/7aba8a8e9ee7f44302c1965d61e11e2491c5a43c/server/opts.go#L3220

server/parser.go
```diff
@@ -160,7 +159,24 @@ func (c *client) parse(buf []byte) error {
...
+					var ok bool
+					// Check here for NoAuthUser. If this is set allow non CONNECT protos as our first.
+					// E.g. telnet proto demos.
+					if noAuthUser := s.getOpts().NoAuthUser; noAuthUser != _EMPTY_ {
+						s.mu.Lock()
+						user, exists := s.users[noAuthUser]
+						s.mu.Unlock()
+						if exists {
+							c.RegisterUser(user)
+							authSet, ok = false, true
```
kozlovic, line 174: But again, this seem wrong. As long as there is no_auth_user, even if a client plans to send CONNECT, then it will always be registered as a no_auth_user...

kozlovic (changes requested):

server/parser.go
kozlovic: Nevermind, if a client sends CONNECT, we will not be there of course...

server/parser.go
derekcollison: I feel no_auth setting should behave same as simple server. This was only thing it was very different on.

server/auth_test.go
derekcollison: Ah, I see.

matthiashanel (approved):
Lgtm.

server/parser.go
```diff
@@ -171,6 +171,8 @@ func (c *client) parse(buf []byte) error {
 						s.mu.Unlock()
 						if exists {
 							c.RegisterUser(user)
+							c.clearAuthTimer()
```
kozlovic, line 174: Please surround this and the call below with `c.mu.Lock()/c.mu.Unlock()`, since those two require locking.

kozlovic (changes requested):

kozlovic (approved):
LGTM

PR #2813, opened 2022-01-24
Updates to Pull Consumers

derekcollison:
This change fixes a break in behavior for one-shot pull batch requests and formalizes pull requests and their associated behaviors. We also now can expire out of order and we will expire on a leader change for the consumer as that state is shared for pending requests between leaders and followers, only if >= 2.7.1

/cc @nats-io/core

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -6,7 +6,7 @@ require (
 	github.com/klauspost/compress v1.13.4
 	github.com/minio/highwayhash v1.0.1
 	github.com/nats-io/jwt/v2 v2.2.1-0.20220113022732-58e87895b296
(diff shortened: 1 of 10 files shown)


tbeets:
In 2.7.1+ the in-flight pull batch status of subscribers is synchronized in the consumer RG?  Does that imply a lot of new synchronization traffic since previously the current leader only tracked?

derekcollison:
We track addition and deletion of reply subjects for the requests now, so that is an addition to delivered and ack updates.

derekcollison:
The new consumer leader will alert that the request assigned to that reply subject is no longer valid and that the user/app should adjust accordingly.

kozlovic (approved):
LGTM

kozlovic (approved):
LGTM

PR #2814, opened 2022-01-24
If we detect negative deleted, adjust to 0.

derekcollison:
This is a bandaid for now. This could possibly happen but unclear why.

/cc @nats-io/core


kozlovic (approved):
LGTM

PR #2815, opened 2022-01-25
Added in /healthz endpoint.

derekcollison:
Used for health and liveness probes for environments like k8s.

Currently this code returns a 200 and `{ "status": "ok" }` iff all configured ports are open
and if JetStream is configured and we have contact with the meta leader and its current and all streams are up to date.

If we are not current we return a 503 at first error and a `{ "status": "unavailable", "error": "DESCRIPTION" }` body.

We can do much more but this feels like a decent start to test with upgrading servers in a k8s environment that is running JetStream.

/cc @nats-io/core

### DIFF
--- a/server/monitor.go
+++ b/server/monitor.go
@@ -1260,6 +1260,7 @@ func (s *Server) HandleRoot(w http.ResponseWriter, r *http.Request) {
 	<a href=.%s>subsz</a><br/>
 	<a href=.%s>accountz</a><br/>
 	<a href=.%s>jsz</a><br/>
+	<a href=.%s>healthz</a><br/>
     <br/>
     <a href=https://docs.nats.io/nats-server/configuration/monitoring>help</a>
   </body>
@@ -1272,6 +1273,7 @@ func (s *Server) HandleRoot(w http.ResponseWriter, r *http.Request) {
 		s.basePath(SubszPath),
 		s.basePath(AccountzPath),
 		s.basePath(JszPath),
+		s.basePath(HealthzPath),
 	)
 }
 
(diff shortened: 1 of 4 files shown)


server/monitor.go
```diff
@@ -2686,9 +2688,98 @@ func (s *Server) HandleJsz(w http.ResponseWriter, r *http.Request) {
...
 	ResponseHandler(w, r, b)
 }
+
+type HealthStatus struct {
+	Status string `json:"status"`
+	Error  string `json:"error,omitempty"`
+}
+
+// https://tools.ietf.org/id/draft-inadarei-api-health-check-05.html
+func (s *Server) HandleHealthz(w http.ResponseWriter, r *http.Request) {
```
matthiashanel, line 2704: I assume this needs exposing via the server status requests in events.go line 859 and become part of the table monSrvc.

Provided that, this probably has to be split into two functions and have returns that align with similar calls.

matthiashanel (changes requested):

server/monitor.go
derekcollison: I don't think so, this one is really only http and for the environments like k8s and liveness probes etc.
I could be wrong though and we could then put it there but does not feel like same level as the others IMO.

tbeets:
I think there are some chicken-egg.   K8s makes some assumptions that "livenessProbe" means the application container is running and did not bomb out in some basic way.  "readinessProbe" means that K8s will route service traffic (network) of said application container.  "startupProbe" I think is really the legacy "liveness" when K8s only had one probe.

Since JS RAFT sync runs over NATS messaging, I'm not seeing that there is yet a distinction between NATS Core readiness and separately or inclusively JS readiness that K8s could grok.

I think @kozlovic made some changes to move the HTTP monitor availability ahead of JS catch-up/self-repair to decouple K8s from prematurely killing the server on startup?

wallyqs:
The startup probe got introduced in k8s 1.16 (around a couple of years), and is the one that runs first now when it is enabled. After it succeeds, then the readiness / liveness probes start running.  The implementation of `healthz` in this PR would be useful as a startup probe since it would help make the container invisible to a k8s `Service` until it is marked as ready for connections so clients would be unable to connect to it.

server/monitor.go
```diff
@@ -2686,9 +2688,98 @@ func (s *Server) HandleJsz(w http.ResponseWriter, r *http.Request) {
...
+		w.WriteHeader(http.StatusServiceUnavailable)
+	} else if js := s.getJetStream(); js != nil {
+		// Check JetStream status here.
+		js.mu.RLock()
+		clustered, cc := !js.standAlone, js.cluster
+		js.mu.RUnlock()
+		if clustered {
+			// We do more checking for clustered mode to allow for proper rolling updates.
+			// We will make sure that we have seen the meta leader and that we are current with all assets.
+			node := js.getMetaGroup()
```
wallyqs, line 2723: I think this part is very useful for when the container is starting up (before being considered to be able to accept connections from clients // startup probe), but for the liveness probe that checks periodically that the server is healthy after starting successfully we may have to skip this part otherwise the server may get restarted by not being caught up after running for awhile?

Maybe we could define some special http headers for the endpoint with the preconditions that we want to assert, during startup we want to ensure that JS is caught up so can make the k8s probe send a header like `Check-JS-Cluster-Current: true`, but then during liveness check it is only needed to check that JetStream has not been auto disabled so then a different header is set like `Check-JS-Enabled: true`.

Jarema (commented):
I think we need a clear distinction between readiness, startup, and liveness checks in context of NATS.

**1. Readiness probe**
OK: Server is ready, caught up regarding all assets, meta, etc.
No distinction between JS and Core (k8s doesn't care. It's ready or not).
**2. Liveness**
Ok if NATS Server is alive. It though may not have caught up yet or it might just got disconnected and is recovering after that. Or temporarily lost meta leader. Then **readiness fails**, but liveness doesn't.
IMHO liveness probe should fail **only** if server is in **not recoverable state**, deadlock etc. It basically says "kill me if I'm in that state. I will not recover on my own".

3. Startup probe:
That's basically a more graceful readiness that takes into account longer "warmup" of initial data load of fresh instance that in the past forced users to make readiness probe more relaxed or hack around with initial delay etc (which could lead to another issues and less responsive k8s deployments).

Distinction can be made by headers as @wallyqs suggested.
We can use the same checks for startup and readiness, just with different configs in k8s regarding threshold, etc.
For liveness, we should really check only the non-recoverable conditions of NATS Server.

Regarding headers (with suggested names) - one con: If we want our users to have NATS Clusters installations without going deep into its implementation details, we should expose some "opinionated" way to tell "this is for readiness" and "this is for liveness". Otherwise, people may encounter unexpected restarts in scaling setups, which could be an unpleasant surprise.
That "opinionated" part could be a path in `healthz` endpoint (`healthz/liveness` though sounds non-orthodox) or k8s-specific header `K8s-Liveness` `K8s-Readiness`. Regardless of the implementation, those would just be an aliased config of non-vendor-specific (vendor being k8s) NATS settings for convenience and respecting principle of least surprise. Alternative would be to just have clear info in docs (not everybody uses our Helm charts).

server/monitor.go
derekcollison: This is more for startup for sure. For liveness, how should we define when we would ever want k8s to stop the server? I am struggling to think of one, so for liveness its varz maybe and if it returns a 200?

kozlovic (approved):
From a code perspective, LGTM

tbeets:
Is there danger of k8s cutting off container network connectivity for a "Live" but not "Ready" JS-enabled node thus creating a self-fulfilling prophesy? (i.e. node cut off from other nodes by k8s thus cannot complete JS activities to become in ready state)

I am assuming k8s is pretty dumb about "readiness for service traffic" and can't distinguish server message traffic from client message traffic container wise...

derekcollison:
This should not be used for liveness, just ready probe IMO.

server/monitor.go
ColinSullivan1: Exposing as a NATS endpoint would be useful for future NATS tooling around server maintenance.

server/monitor.go
wallyqs: For liveness, in case JS is enabled and it becomes auto disabled due to an I/O error, instead of leaving the server running without JS then maybe make the healthcheck fail and let it be restarted to try to auto-recover? For non JS deployments, just checking `/`/ or `/varz` as we are right now would suffice.

server/monitor.go
derekcollison: Longer discussion but there is no value there IMO.

derekcollison:
Going to merge. @wallyqs please start kicking the tires on using this as a ready probe for upgrading a JetStream enabled cluster in k8s etc.

wallyqs:
Thanks, trying it now on this branch (https://github.com/nats-io/k8s/pull/410).  First we'd start using it as a startup probe.  Btw following image is helpful to put the probes into context:

image

source: https://andrewlock.net/deploying-asp-net-core-applications-to-kubernetes-part-6-adding-health-checks-with-liveness-readiness-and-startup-probes/#the-three-kinds-of-probe-liveness-readiness-and-startup-probes

derekcollison:
Diagram is helpful. I think this is *only* a readiness probe, the /healthz endpoint I mean.

tbeets:
The way the k8s docs read, as soon as you define a ReadinessProbe (defaults to implicit Success if not defined), then k8s won't open your container up to network until ReadinessProbe returns success.  If I am interpreting right (proof is in test of course) it feels like our /healthz sense of  "readiness" (JS in communication with meta leader and caught up which requires container networking) is in conflict?

tbeets:
As long as we document /healthz in terms of "JetStream health" not k8s readiness, I think we're okay.  But dual k8s/NATS Jetstream community could get confused.

wallyqs:
@tbeets yes that's right, this would be useful for upgrades when new nodes are added but during the startup phase then the metadata leader check prevents the cluster from being formed since there cannot be a quorum at the beginning so it cannot start.  What we are looking at right now is that the probe will become activated/deactivated depending on the context, so for example for an upgrade, the generated yaml will have the probe but for a fresh cluster then the probe will be disabled so that  there can be a quorum.

PR #2816, opened 2022-01-25
Fix jwt based user/activation token revocation and granularity

matthiashanel:
user and activation token did not honor the jwt value for all * on
connect.

activation token where not re evaluated when the export revoked a key.
In part this is a consistency measure so servers that already have an
account and servers that don't behave the same way.

in jwt activation token revocations are stored per export.
The server stored them per account, thus effectively merging
revocations. Now they are stored per export inside the server too.

fixes nats-io/nsc/issues/442

### DIFF
--- a/server/monitor.go
+++ b/server/monitor.go
@@ -2261,6 +2261,9 @@ func (s *Server) accountInfo(accName string) (*AccountInfo, error) {
 		}
 	}
 	collectRevocations := func(revocations map[string]int64) map[string]time.Time {
+		if len(revocations) == 0 {
+			return nil
+		}
 		rev := map[string]time.Time{}
 		for k, v := range a.usersRevoked {
 			rev[k] = time.Unix(v, 0)
--- a/server/monitor.go
+++ b/server/monitor.go
@@ -2261,11 +2261,12 @@ func (s *Server) accountInfo(accName string) (*AccountInfo, error) {
 		}
 	}
 	collectRevocations := func(revocations map[string]int64) map[string]time.Time {
-		if len(revocations) == 0 {
+		l := len(revocations)
+		if l == 0 {
 			return nil
 		}
-		rev := map[string]time.Time{}
-		for k, v := range a.usersRevoked {
+		rev := make(map[string]time.Time, l)
+		for k, v := range revocations {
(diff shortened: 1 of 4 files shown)


server/monitor.go
```diff
@@ -2260,6 +2260,13 @@ func (s *Server) accountInfo(accName string) (*AccountInfo, error) {
 			vrIssues[i] = ExtVrIssues{v.Description, v.Blocking, v.TimeCheck}
 		}
 	}
+	collectRevocations := func(revocations map[string]int64) map[string]time.Time {
+		rev := map[string]time.Time{}
```
kozlovic, line 2264: That will likely cause that collectRevocations will always return a non-nil map, which then means that `ExtExport.RevokedAct` will never be nil (and so omitempty does not help). Should you declare as `var rev map[string]time.Time` and create only in the for-loop if nil?

server/monitor.go
```diff
@@ -2162,7 +2163,6 @@ type AccountInfo struct {
 	Claim       *jwt.AccountClaims   `json:"decoded_jwt,omitempty"`
 	Vr          []ExtVrIssues        `json:"validation_result_jwt,omitempty"`
 	RevokedUser map[string]time.Time `json:"revoked_user,omitempty"`
-	RevokedAct  map[string]time.Time `json:"revoked_activations,omitempty"`
```
kozlovic, line 2165: As a 2.7.1, I don't really like that we "break" this by removing an exported field, but since it was a bug, I guess that is fine. But some applications will fail to compile when updating their dependencies to v2.7.1...

server/monitor.go
matthiashanel: fixed

server/monitor.go
matthiashanel: This is part of an account debug call.
I poked at the jsm library and it appears to not be used there.

Should I leave the entry in there and comment it as obsolete?

kozlovic: No, that would not really help. The application would be "broken" anyway without a change to go get the information from the exports. Thanks for checking.

kozlovic (approved):
LGTM

PR #2826, opened 2022-01-28
supporting max_ha, an account limit for streams+consumer > 1 peer

matthiashanel:
As a reminder a stream of R=3 and one of R=5 would be allowed to be created when the limit is 2. (This is not a limit of 8)


server/accounts.go
```diff
@@ -3258,6 +3258,7 @@ func (s *Server) updateAccountClaimsWithRefresh(a *Account, ac *jwt.AccountClaim
 			MaxStore:         ac.Limits.JetStreamLimits.DiskStorage,
 			MaxStreams:       int(ac.Limits.JetStreamLimits.Streams),
 			MaxConsumers:     int(ac.Limits.JetStreamLimits.Consumer),
+			MaxHaResources:   0, //TODO read from claim JS Limits
```
derekcollison: MaxHAResources

server/jetstream_cluster.go
```diff
@@ -3759,6 +3759,31 @@ func (js *jetStream) createGroupForStream(ci *ClientInfo, cfg *StreamConfig) *ra
...
 }
 
+func (js *jetStream) haExceeded(acc string, limit int) bool {
+	if limit <= 0 {
+		return false
+	}
+	js.mu.RLock()
+	defer js.mu.RUnlock()
+	if assignment, ok := js.cluster.streams[acc]; ok {
+		cnt := 0
```
derekcollison, line 3771: count or n, not a fan of cnt ;)

server/jetstream_cluster.go
```diff
@@ -4499,11 +4535,24 @@ func (s *Server) jsClusteredConsumerRequest(ci *ClientInfo, acc *Account, subjec
 		return
 	}
 
-	js.mu.Lock()
-	defer js.mu.Unlock()
+	// Grab our jetstream account info.
+	acc.mu.RLock()
+	jsa := acc.js
```
derekcollison, line 4542: Put all on one line?

server/jetstream_cluster.go
```diff
@@ -4499,11 +4535,24 @@ func (s *Server) jsClusteredConsumerRequest(ci *ClientInfo, acc *Account, subjec
...
+	accName := acc.Name
+	accHaLim := jsa.limits.MaxHaResources
+	acc.mu.RUnlock()
 
 	var resp = JSApiConsumerCreateResponse{ApiResponse: ApiResponse{Type: JSApiConsumerCreateResponseType}}
 
+	if cfg.Durable != _EMPTY_ && js.haExceeded(accName, accHaLim) {
```
derekcollison, line 4549: We should check more when we select the peer group and check that? Not sure cfg.Durable will always be the only pace to check. We can have R>1 ephemerals if attached to a WQ/Interest stream.

kozlovic (commented):
Nothing to add to Derek's comments. Likely need to change all occurrences of "Ha" with "HA", but that's just details. The more important thing, which I don't have an answer about, is regarding when to check for HA exceeds, which right now is based on durable name set.

server/jetstream_cluster.go
matthiashanel: When in this code R>1. Altered this check to be 
`(cfg.Durable != _EMPTY_ || sa.Config.Retention == WorkQueuePolicy) && js.haExceeded(accName, accHaLim)`

matthiashanel:
> Nothing to add to Derek's comments. Likely need to change all occurrences of "Ha" with "HA", but that's just details. The more important thing, which I don't have an answer about, is regarding when to check for HA exceeds, which right now is based on durable name set.

In this code R will be > 1. 
Altered the check to be :
`(cfg.Durable != _EMPTY_ || sa.Config.Retention == WorkQueuePolicy) && js.haExceeded(accName, accHaLim)`

PR #2829, opened 2022-01-28
Updating a push consumer to be pull should fail.

derekcollison:
This CL will disallow that upgrade. We had a check in place for pull to push, but not the reverse.

/cc @nats-io/core


jnmoyne (approved):
Test scenario now returns
`nats: error: Consumer creation failed: can not update push consumer to pull based (10012)` when attempting to create the consumer again.

LGTM!

kozlovic (approved):
LGTM

PR #2833, opened 2022-01-31
Added in ability to get per subject details via StreamInfo.

derekcollison:
Note that we went with a simpler approach vs message per result or a paged api. We will place the subject details (just msg count) into the stream info state. Since NATS as a system can limit payloads of any msg entering the system, but the system itself can send any size message to any client, we allow this to proceed. We artificially limit it to 100k, and for items like KV we can fall back to the watcher/ordered consumer pattern as needed.

/cc @nats-io/core

### DIFF
--- a/server/jetstream_test.go
+++ b/server/jetstream_test.go
@@ -1,4 +1,4 @@
-// Copyright 2019-2021 The NATS Authors
+// Copyright 2019-2022 The NATS Authors
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
 // You may obtain a copy of the License at
@@ -8420,7 +8420,7 @@ func TestJetStreamDeleteMsg(t *testing.T) {
 				t.Fatalf("Expected to get the stream back")
 			}
 
-			expected := StreamState{Msgs: 6, Bytes: 6 * bytesPerMsg, FirstSeq: 12, LastSeq: 20}
+			expected := StreamState{Msgs: 6, Bytes: 6 * bytesPerMsg, FirstSeq: 12, LastSeq: 20, NumSubjects: 1}
 			state = mset.state()
 			state.FirstTime, state.LastTime, state.Deleted, state.NumDeleted = time.Time{}, time.Time{}, nil, 0
 
@@ -14656,7 +14656,7 @@ func TestJetStreamStreamInfoSubjectsDetails(t *testing.T) {
 	nc, js := jsClientConnect(t, s)
 	defer nc.Close()
 
-	getInfo := func(filter string) *StreamInfo {
+	getInfo := func(t *testing.T, filter string) *StreamInfo {
 		t.Helper()
 		// Need to grab StreamInfo by hand for now.
 		req, err := json.Marshal(&JSApiStreamInfoRequest{SubjectsFilter: filter})
@@ -14666,6 +14666,9 @@ func TestJetStreamStreamInfoSubjectsDetails(t *testing.T) {
 		var si StreamInfo
 		err = json.Unmarshal(resp.Data, &si)
 		require_NoError(t, err)
+		if si.State.NumSubjects != 3 {
+			t.Fatalf("Expected NumSubjects to be 3, but got %d", si.State.NumSubjects)
+		}
 		return &si
 	}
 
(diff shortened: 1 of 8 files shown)


server/jetstream_test.go
```diff
@@ -14642,6 +14642,112 @@ func TestJetStreamMaxMsgsPerSubjectWithDiscardNew(t *testing.T) {
...
+			Name:     "TEST",
+			Subjects: []string{"*"},
+			Storage:  st,
+		})
+		require_NoError(t, err)
+		defer js.DeleteStream("TEST")
+
+		msg := []byte("ok")
+		// Now place 60 msgs. 20x each of foo, bar and baz.
+		for i := 0; i < 20; i++ {
```
kozlovic, line 14725: I would have picked a different number per subject to verify that we don't accidentally put the same number of messages per subject to all subjects...

kozlovic (changes requested):
Would make sure that the test has different number of messages per subject...

server/jetstream_test.go
derekcollison: Will fix.

kozlovic (approved):
LGTM, but the test "TestNoRaceJetStreamSlowFilteredInititalPendingAndFirstMsg" is failing in most of the runs for this PR, but passes locally (-count 10) on my machine...

derekcollison:
Yep, removed default NumSubjects for StreamInfo since it was also present in FastState which was taking too long and was used to create consumers etc.

server/store.go
```diff
@@ -119,16 +119,17 @@ const (
...
-	Deleted    []uint64        `json:"deleted,omitempty"`
-	Lost       *LostStreamData `json:"lost,omitempty"`
-	Consumers  int             `json:"consumer_count"`
+	Msgs       uint64            `json:"messages"`
+	Bytes      uint64            `json:"bytes"`
+	FirstSeq   uint64            `json:"first_seq"`
+	FirstTime  time.Time         `json:"first_ts"`
+	LastSeq    uint64            `json:"last_seq"`
+	LastTime   time.Time         `json:"last_ts"`
+	Subjects   map[string]uint64 `json:"subjects,omitempty"`
```
ripienaar, line 128: Maybe time to revisit compressed API responses and acceptible response content negotiation like we have for server reqs.

As beore when we spoke I am not a huge fan of this approach as its inherently limited, think we'll sooner or later need to consider some streaming api responses anyway

server/norace_test.go
```diff
@@ -4108,3 +4108,89 @@ func TestNoRaceJetStreamClusterHealthz(t *testing.T) {
...
+	require_NoError(t, err)
+
+	req, err := json.Marshal(&JSApiStreamInfoRequest{SubjectsFilter: nats.AllKeys})
+	require_NoError(t, err)
+	resp, err := nc.Request(fmt.Sprintf(JSApiStreamInfoT, "TEST"), req, 5*time.Second)
+	require_NoError(t, err)
+	var sir JSApiStreamInfoResponse
+	err = json.Unmarshal(resp.Data, &sir)
+	require_NoError(t, err)
+	if sir.Error == nil || sir.Error.Code != 500 || sir.Error.Description != "subject details would exceed maximum allowed" {
```
ripienaar, line 4193: ```if !IsNatsErr(sir.Error, JSStreamInfoMaxSubjectsErr) { .... }```

ripienaar (approved):
LGTM

server/store.go
aricart: On KV for example - if a KV uses a nkey for the base of a subject, this effectively makes a subject 80+ chars in length. If the KV organizes several values by key, that means that for each datum, we have 2+ values, so assuming 1000 entries, we are looking at at least 2000 subjects or 160K payload for this request. With 1M entries (2M subjects) we are looking at 160MB.

derekcollison: This is opt-in, by default you do not get the subject details (histogram).

We have an artificial limit of 100k subjects that if we have more than that we will error and the user can fall back to the watcher impl like we have now.

Also note that maximum payloads are not in play here since its a system generated message.

server/norace_test.go
derekcollison: Thanks will make that change.

derekcollison:
I added back in the NumSubjects part of StreamInfo.

I think this PR is good, please re-review when you have a moment.

kozlovic (approved):
LGTM, but it looks like there are changes in this PR that have already been submitted in different PRs and likely in main already. Also, seems like you have a conflict, so you may want to rebase.

server/store.go
scottf: What about just making getting subject a JS API call similar to getting consumers or stream names? This way it could be paged.

ripienaar: Indeed, thats what I also suggested when this was first mooted. However paging a dynamic data set like subjects is bound to be unreliable, hence a seperate API call and streaming results would yield a better outcome.

derekcollison: I considered that quite a bit, but decided to go with similar approach to deleted details for a stream info.

aricart:
I have documented the client changes here:

https://github.com/nats-io/nats-architecture-and-design/issues/91

server/store.go
derekcollison: We debated this quite a bit and this is what I think is best solution given all things considered.

PR #2834, opened 2022-02-01
[FIXED] JetStream: BackOff redeliveries would always use first in list

kozlovic:
If the consumer's sequence was not the same than the stream's sequence,
then the redelivery would always use the first duration from the
BackOff list.


derekcollison (approved):
LGTM

PR #2835, opened 2022-02-01
Consumer pending gets stuck.

derekcollison:
Under certain scenarios the pending for a consumer could appear to get stuck.

Under the covers we were calculating pending per msg block incorrectly when a single message existed beyond the requested sequence.

/cc @nats-io/core


kozlovic (approved):
LGTM

PR #2837, opened 2022-02-01
Fix flapping test - TestNoRaceJetStreamClusterCorruptWAL

derekcollison:
This test was using fetch and failing if the complete batch was not filled.
This has nothing to do with the test, we just want to make sure the leader steps down and there were no low level errors on the fetch.

/cc @nats-io/core


kozlovic (approved):
LGTM

PR #2840, opened 2022-02-04
Remove dynamic account behaviors.

derekcollison:
We used these in tests and for experimenting with sandboxed environments like the demo network.

/cc @nats-io/core


server/route.go
```diff
@@ -1037,17 +1037,16 @@ func (c *client) processRemoteSub(argo []byte, hasOrigin bool) (err error) {
...
-			expire = true
+		// if the option of retrieving accounts later exists, create an expired one.
+		// When a client comes along, expiration will prevent it from being used,
+		// cause a fetch and update the account to what is should be.
+		if staticResolver {
+			c.Errorf("Unknown account %q for remote subject %q", accountName, sub.subject)
+			return
 		}
+		c.Debugf("Unknown account %q for remote subject %q", accountName, sub.subject)
+		expire = true
```
kozlovic, line 1048: No big deal, but expire is initialized to false but now unconditionally set to true, so we could remove `expire` all together (including in the test just below)

kozlovic (approved):
LGTM, although code could be simplified a bit by removing `expire`...

server/route.go
derekcollison: Will do.

derekcollison: Fixed.

kozlovic (approved):
LGTM

PR #2847, opened 2022-02-06
Fixed flow control stall under specific conditions of message size.

derekcollison:

/cc @nats-io/core


kozlovic (approved):
LGTM

PR #2848, opened 2022-02-08
Improved sparse consumers replay time.

derekcollison:
When a stream has multiple subjects and a consumer filters the stream to a small and spread out list of messages the logic would do a linear scan looking for the next message for the filtered consumer.
This CL allows the store layer to utilize the per subject info to improve the times.

/cc @nats-io/core

### DIFF
--- a/server/consumer.go
+++ b/server/consumer.go
@@ -2476,23 +2476,22 @@ var (
 // Get next available message from underlying store.
 // Is partition aware and redeliver aware.
 // Lock should be held.
-func (o *consumer) getNextMsg() (subj string, hdr, msg []byte, seq uint64, dc uint64, ts int64, err error) {
+func (o *consumer) getNextMsg() (subj string, hdr, msg []byte, sseq uint64, dc uint64, ts int64, err error) {
 	if o.mset == nil || o.mset.store == nil {
 		return _EMPTY_, nil, nil, 0, 0, 0, errBadConsumer
 	}
-	for {
-		seq, dc := o.sseq, uint64(1)
-		if o.hasSkipListPending() {
-			seq = o.lss.seqs[0]
-			if len(o.lss.seqs) == 1 {
-				o.sseq = o.lss.resume
-				o.lss = nil
-				o.updateSkipped()
-			} else {
-				o.lss.seqs = o.lss.seqs[1:]
-			}
-		} else if o.hasRedeliveries() {
-			seq = o.getNextToRedeliver()
+	seq, dc := o.sseq, uint64(1)
+	if o.hasSkipListPending() {
+		seq = o.lss.seqs[0]
+		if len(o.lss.seqs) == 1 {
+			o.sseq = o.lss.resume
+			o.lss = nil
+			o.updateSkipped()
+		} else {
+			o.lss.seqs = o.lss.seqs[1:]
+		}
+	} else if o.hasRedeliveries() {
+		for seq = o.getNextToRedeliver(); seq > 0; seq = o.getNextToRedeliver() {
 			dc = o.incDeliveryCount(seq)
 			if o.maxdc > 0 && dc > o.maxdc {
 				// Only send once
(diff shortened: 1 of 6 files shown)


server/norace_test.go
```diff
@@ -4196,3 +4196,82 @@ func TestNoRaceJetStreamStreamInfoSubjectDetailsLimits(t *testing.T) {
 		t.Fatalf("Did not get correct error response: %+v", sir.Error)
 	}
 }
+
+func TestJetStreamSparseConsumers(t *testing.T) {
```
kozlovic, line 4200: Prefix should be `TestNoRace...` if you want it to be run on Travis (since we run the "no race" tests with `-run=TestNoRace ..`

server/memstore.go
```diff
@@ -628,6 +628,61 @@ func (ms *memStore) LoadLastMsg(subject string) (subj string, seq uint64, hdr, m
 	return sm.subj, sm.seq, sm.hdr, sm.msg, sm.ts, nil
 }
 
+// LoadNextMsg will find the next message matching the filter subject starting at the start sequence.
+// The filter subject can be a wildcard.
+func (ms *memStore) LoadNextMsg(filter string, wc bool, start uint64) (subj string, seq uint64, hdr, msg []byte, ts int64, err error) {
+	ms.mu.RLock()
+	defer ms.mu.RUnlock()
+
```
kozlovic, line 636: I see that in case of filestore, you check for "closed" state. Do you have the same for memstore?

kozlovic (changes requested):

server/norace_test.go
derekcollison: Your right, will fix.

server/memstore.go
derekcollison: We do not check for closed state as much in memstore, just storing a new message.

matthiashanel: Will that cause unexpected and untested differences in behaviour?

server/consumer.go
```diff
@@ -2476,23 +2476,22 @@ var (
...
+	if o.hasSkipListPending() {
+		seq = o.lss.seqs[0]
+		if len(o.lss.seqs) == 1 {
+			o.sseq = o.lss.resume
+			o.lss = nil
+			o.updateSkipped()
+		} else {
+			o.lss.seqs = o.lss.seqs[1:]
+		}
+	} else if o.hasRedeliveries() {
```
matthiashanel, line 2493: are skipListPending and redeliveries mutually exclusive or does skipListPending include re deliveries?
Otherwise is it possible that redeliveries are never checked?

server/memstore.go
derekcollison: I don't think so since the upper layers are controlling access to the stores.

server/consumer.go
derekcollison: SkipLists are only for when the consumer first starts when you are asking for the last per subject. So they clear pretty quickly but regardless we want then processed first then any redeliveries.

kozlovic (approved):
LGTM

derekcollison:
Thanks, I have a test that flapped on me and am investigating that one a bit more before merging.

server/filestore.go
```diff
@@ -3462,6 +3521,35 @@ func (fs *fileStore) LoadLastMsg(subject string) (subj string, seq uint64, hdr,
...
+	if start < fs.state.FirstSeq {
+		start = fs.state.FirstSeq
+	}
+
+	for _, mb := range fs.blks {
+		// Skip blocks that are less than our starting sequence.
+		if start > atomic.LoadUint64(&mb.last.seq) {
+			continue
+		}
+		if sm, expireOk, err := mb.firstMatching(filter, wc, start); err == nil {
```
kozlovic, line 3542: So if `err != nil`, we don't care really what the error is, we will fall through and return `ErrStoreEOF`, right?

kozlovic (approved):

server/filestore.go
derekcollison: Correct, start will select the mb, but we may not match anything in there. If so we simply move on. Should we return an error if not MsgNotFound?

kozlovic: Depends on the caller, but really MsgNotFound could be equivalent on what you return now, which is EOF because technically, it is as if we were going through the end of the file to look for it and we didn't find it.

derekcollison: Correct, what I meant was a non msg not found error should we return that, like corrupt msg etc?

kozlovic: Yes, it would help debug if someday there is an error reading the file, etc..

derekcollison: ok fixed.

kozlovic (approved):
LGTM

PR #2857, opened 2022-02-09
A stream name is tied to its identity and can not be changed on a restore.

derekcollison:

/cc @nats-io/core


kozlovic (approved):
LGTM

ripienaar (approved):
LGTM

PR #2858, opened 2022-02-09
Make sure we snapshot initial consumer info during consumer creation.

derekcollison:
In certain circumstances, mostly with direct push based consumers, the consumer info returned by the js.Subscribe() calls could have a consumer info that already reflects delivering some, possibly all the original pending messages.

This change makes the initial pending state correct in those situations.

/cc @nats-io/core

### DIFF
--- a/server/consumer.go
+++ b/server/consumer.go
@@ -239,6 +239,7 @@ type consumer struct {
 	maxdc             uint64
 	waiting           *waitQueue
 	cfg               ConsumerConfig
+	ici               *ConsumerInfo
 	store             ConsumerStore
 	active            bool
 	replay            bool
@@ -848,6 +849,9 @@ func (o *consumer) setLeader(isLeader bool) {
 			o.pch = make(chan struct{}, 1)
 		}
 		o.mu.Unlock()
+
+		// Snapshot initial info.
+		o.infoWithSnap(true)
 
 		// Now start up Go routine to deliver msgs.
 		go o.loopAndGatherMsgs(qch)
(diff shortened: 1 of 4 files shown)


server/consumer.go
```diff
@@ -849,6 +850,9 @@ func (o *consumer) setLeader(isLeader bool) {
 		}
 		o.mu.Unlock()
 
+		// Snapshot initial info.
+		o.ici = o.info()
```
kozlovic, line 854: But o.ici is captured outside of o's lock (note that initialInfo() and clearInitialInfo() do acquire lock to get o.ici).
The way I see it, o.ici should be set inside o.info() (while still under lock - see my note below in initialInfo()), or have a infoLocked() version that is assumed to run under lock (and does not have to check for o.mset), however, there may be the issue with js.clusterInfo() that requires JS lock...

server/consumer.go
```diff
@@ -1830,6 +1834,27 @@ func (o *consumer) writeStoreStateUnlocked() error {
...
 
+// Returns an initial info. Only applicable for non-clustered consumers.
+// We will clear after we return it, so one shot.
+func (o *consumer) initialInfo() *ConsumerInfo {
+	o.mu.Lock()
+	ici := o.ici
+	o.ici = nil // gc friendly
+	o.mu.Unlock()
+	if ici == nil {
+		ici = o.info()
```
kozlovic, line 1845: If we set o.ici inside o.info() *and* we would not want it to be set in that specific condition, then maybe we need a version of o.info() (since this is called many times) where we pass a boolean to indicate if we want to set it or not. So for instance, in setLeader() we would pass true so that it is set, but everywhere else it would not.

kozlovic (changes requested):

server/consumer.go
derekcollison: Sorry yes that was supposed to be different code path there, will fix.

Can't do infoLocked, clustering.

derekcollison: Will fix.

kozlovic (approved):
LGTM!

PR #2859, opened 2022-02-10
Remove fss files from a snapshot when a block is removed.

derekcollison:
During a filestore snapshot we generate the fss files for the snapshot but were not cleaning them up if the block was deleted before a server restart.

https://gist.github.com/nekufa/010185dfb59261f222a0042d3a7d2a1c

/cc @nats-io/core

### DIFF
--- a/server/filestore.go
+++ b/server/filestore.go
@@ -30,6 +30,7 @@ import (
 	"net"
 	"os"
 	"path"
+	"path/filepath"
 	"runtime"
 	"sort"
 	"sync"
@@ -187,6 +188,8 @@ const (
 	indexScan = "%d.idx"
 	// used to load per subject meta information.
 	fssScan = "%d.fss"
+	// to look for orphans
+	fssScanAll = "*.fss"
 	// used to store our block encryption key.
 	keyScan = "%d.key"
 	// This is where we keep state on consumers.
@@ -890,7 +893,6 @@ func (fs *fileStore) recoverMsgs() error {
 						fs.psmc[subj] += ss.Msgs
 					}
 				}
-
 			} else {
 				return err
 			}
@@ -907,6 +909,14 @@ func (fs *fileStore) recoverMsgs() error {
 
 	if err != nil {
 		return err
+	}
+
+	// We had a bug that would leave fss files around during a snapshot.
+	// Clean them up here if we see them.
+	if fms, err := filepath.Glob(path.Join(mdir, fssScanAll)); err == nil && len(fms) > 0 {
+		for _, fn := range fms {
+			os.Remove(fn)
+		}
 	}
 
 	// Limits checks and enforcement.
@@ -4325,6 +4335,10 @@ func (mb *msgBlock) dirtyCloseWithRemove(remove bool) {
 		if mb.mfn != _EMPTY_ {
 			os.Remove(mb.mfn)
 			mb.mfn = _EMPTY_
+		}
+		if mb.sfn != _EMPTY_ {
+			os.Remove(mb.sfn)
+			mb.sfn = _EMPTY_
 		}
 	}
 }
--- a/server/norace_test.go
+++ b/server/norace_test.go
@@ -29,6 +29,8 @@ import (
 	"net"
 	"net/http"
 	"net/url"
+	"path"
+	"path/filepath"
 	"runtime"
 	"runtime/debug"
 	"strconv"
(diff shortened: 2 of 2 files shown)


server/filestore.go
```diff
@@ -909,6 +911,15 @@ func (fs *fileStore) recoverMsgs() error {
 		return err
 	}
 
+	// We had a bug that would leave fss files around during a snapshot.
+	// Clean them up here if we see them.
+	fms, err := filepath.Glob(path.Join(mdir, fssScanAll))
```
kozlovic, line 916: Is the intent to return the error if filepath.Glob() produces an error?
Because you don't use fms outside and you WILL return that error as part of recoverMsgs(). So if you don't want to return an error due to Glob(), then I would recommend: `if fms, err := ...; err == nil && len(fms) > 0 {`

server/filestore_test.go
```diff
@@ -3539,3 +3539,50 @@ func TestFileStoreFetchPerf(t *testing.T) {
 	}
 	fmt.Printf("Elapsed to load all messages is %v\n", time.Since(now))
 }
+
+func TestFileStoreSubjectInfoWithSnapshotCleanup(t *testing.T) {
```
kozlovic, line 3543: Ran that test on -race mode and it takes about 150MB (only 29MB without -race). I think it is ok, but just letting you know (I noticed the 10,000 msgs of 1K each).

server/filestore.go
kozlovic: Reading on filepath.Glob(), it says "Glob ignores file system errors such as I/O errors reading directories. The only possible returned error is ErrBadPattern, when pattern is malformed", so I guess it is fine the way you have it?

derekcollison: I don't care about any errors here per so, correct.

server/filestore_test.go
derekcollison: Should I put under norace?

kozlovic: Maybe move.

server/filestore.go
kozlovic: So that means you are going to make it a "one liner", right?

kozlovic (approved):
LGTM

server/filestore.go
derekcollison: Fixed.

server/filestore_test.go
derekcollison: Done.

nekufa:
@derekcollison by the way, is this fix included in docker image synadia/nats-server:nightly-20220217?
even we have started from scratch - disk infinitely grows. i try to reproduce it locally, but has no success..

derekcollison:
Should be, let me double check that nightlies being built correctly etc.

derekcollison:
Yes should be, what are you seeing?

nekufa:
i see that pvc is infinitly used
image

and caling nats stream reports shows me all streams have zero messages..
when i run nats stream purge X disk is magically freeing.

i have reduce cluster configuration to 1 instance (we use official nats helm), but looks like configuration is not the same as simply running official nats image in docker locally. i debug messages on client side - all mesages are fetched and acked correctly. nats-server log does not contain any issues or errors. we've try to set debug to true, but there are many messages about client connections and closed. looks like, it's not usefull in this case.

i understand that it's hard to fix that is not reproducable, but i can't write the reproducer, locally everything works fine and the load is not very high
image

derekcollison:
With nightly and the fix referenced, on a server restart all `*.fss` files from `/data/jetstream/$G/streams/anomaly/msgs` should be cleaned up, we have a test for that. 

So I would suggest doing a server restart and specifically checking `/data/jetstream/$G/streams/anomaly/msgs` directory, if fss files are still there that is not expected.

exename:
i cant find *.fss files, our problem -  many *.key files. 
nats/jetstream  not clean it  and  they fill all inodes on disk

derekcollison:
My apologies, are you using encrypted file storage?

exename:
seems yes, its our options in helm values file https://github.com/nats-io/k8s/tree/main/helm/charts/nats
```
nats:
  image: synadia/nats-server:nightly-20220217

  logging:
    debug: true
    trace: false
    logtime: true
    connectErrorReports:
    reconnectErrorReports:
  jetstream:
    enabled: true
    domain:
    encryption:
      key: Xaxe8eith2joseiX
    memStorage:
      enabled: true
      size: 1Gi
    fileStorage:
      enabled: true
      storageDirectory: /data
      size: 2Gi
      accessModes:
        - ReadWriteOnce
      annotations:

cluster:
  enabled: false
  replicas: 1

natsbox:
  enabled: true
  image: natsio/nats-box:0.7.0
  pullPolicy: IfNotPresent
 
reloader:
  enabled: true
  image: natsio/nats-server-config-reloader:0.6.2

exporter:
  enabled: true
  image: natsio/prometheus-nats-exporter:0.9.0
```
result nats.conf 
```
  nats.conf: |
    # PID file shared with configuration reloader.
    pid_file: "/var/run/nats/nats.pid"

    ###############
    #             #
    # Monitoring  #
    #             #
    ###############
    http: 8222
    server_name:$POD_NAME
    ###################################
    #                                 #
    # NATS JetStream                  #
    #                                 #
    ###################################
    jetstream {
      key: "Xaxe8eith2joseiX"
      max_mem: 1Gi
      store_dir: /data

      max_file:2Gi
    }
    debug: true
    logtime: true
    lame_duck_duration: 120s
  ```

exename:
thanks for hint about encryption, without it we cant see key files and all good.
We will wait for a fix and probably tomorrow we will check with encryption

derekcollison:
https://github.com/nats-io/nats-server/pull/2878

derekcollison:
ok this has been merged and is also being built for nightly as we speak.

nekufa:
@derekcollison with latest build everything works with encrypted storage :) thanks!

PR #2863, opened 2022-02-11
Fixed consumer dlv count and num pending wrong due to redeliveries

kozlovic:
Introduced by #2848, so should not have impacted existing releases.


derekcollison (approved):
LGTM - Thanks.

PR #2865, opened 2022-02-11
Allow pull requests to specify a heartbeat

derekcollison:

/cc @nats-io/core

### DIFF
--- a/server/jetstream_test.go
+++ b/server/jetstream_test.go
@@ -15045,6 +15045,131 @@ func TestJetStreamConsumerPendingCountWithRedeliveries(t *testing.T) {
 	}
 }
 
+func TestJetStreamPullConsumerHeartBeats(t *testing.T) {
+	s := RunBasicJetStreamServer()
+	defer s.Shutdown()
+
+	if config := s.JetStreamConfig(); config != nil {
+		defer removeDir(t, config.StoreDir)
+	}
+
+	nc, js := jsClientConnect(t, s)
+	defer nc.Close()
+
+	_, err := js.AddStream(&nats.StreamConfig{Name: "T", Storage: nats.MemoryStorage})
+	require_NoError(t, err)
+
+	_, err = js.AddConsumer("T", &nats.ConsumerConfig{Durable: "dlc", AckPolicy: nats.AckExplicitPolicy})
+	require_NoError(t, err)
+
+	rsubj := fmt.Sprintf(JSApiRequestNextT, "T", "dlc")
+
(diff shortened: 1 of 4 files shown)


server/jetstream_test.go
```diff
@@ -15045,6 +15045,131 @@ func TestJetStreamConsumerPendingCountWithRedeliveries(t *testing.T) {
...
+	}
+
+	// These should work. Test idle first.
+	start, msgs := time.Now(), doReq(1, 50*time.Millisecond, 250*time.Millisecond, 5)
+	expectHBs(start, msgs, 5, 50*time.Millisecond)
+
+	// Now test that we do not send heartbeats while we receive traffic.
+	go func() {
+		for i := 0; i < 5; i++ {
+			time.Sleep(50 * time.Millisecond)
```
kozlovic, line 15154: Is that not too close from HB? Will the test flap where sometimes the HB will go in-between messages?

server/jetstream_test.go
```diff
@@ -15045,6 +15045,131 @@ func TestJetStreamConsumerPendingCountWithRedeliveries(t *testing.T) {
...
+			js.Publish("T", nil)
+		}
+	}()
+
+	start, msgs = time.Now(), doReq(10, 75*time.Millisecond, 350*time.Millisecond, 6)
+	// The first 5 should be msgs, no HBs.
+	for i := 0; i < 5; i++ {
+		if m := msgs[i].msg; len(m.Header) > 0 {
+			t.Fatalf("Got a potential heartbeat msg when we should not have: %+v", m.Header)
```
kozlovic, line 15163: See above, you send after 50ms, and that is quite close to time HB is supposed to be sent, no?

kozlovic (commented):
Concerned that a test may flap?

server/jetstream_test.go
derekcollison: The HB for this test comes at line 15159 and is 75ms with request timeout of 350ms.

server/jetstream_test.go
derekcollison: HB is 75ms.

server/jetstream_test.go
kozlovic: Right, was looking at the one above, but you start here the publishing in go routine, but req is below at 75ms indeed (still close numbers, so if it start to flap will have to adjust I guess, but good for now).

kozlovic (approved):
LGTM

PR #2867, opened 2022-02-14
Allow replica updates during stream update.

derekcollison:

/cc @nats-io/core

### DIFF
--- a/server/jetstream_cluster.go
+++ b/server/jetstream_cluster.go
@@ -94,6 +94,7 @@ type raftGroup struct {
 	Name      string      `json:"name"`
 	Peers     []string    `json:"peers"`
 	Storage   StorageType `json:"store"`
+	Cluster   string      `json:"cluster,omitempty"`
 	Preferred string      `json:"preferred,omitempty"`
 	// Internal
 	node RaftNode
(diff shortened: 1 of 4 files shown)


server/jetstream_cluster.go
```diff
@@ -2282,15 +2292,26 @@ func (js *jetStream) processClusterUpdateStream(acc *Account, osa, sa *streamAss
...
 	if err == nil && mset != nil {
-		if !alreadyRunning {
+		if !alreadyRunning && numReplicas > 1 {
+			if needsNode {
+				js.createRaftGroup(rg, storage)
+			}
 			s.startGoRoutine(func() { js.monitorStream(mset, sa) })
+		} else if numReplicas == 1 && alreadyRunning {
+			// We downgraded to R1. Make sure we cleanup the raft node and the stream monitor.
+			mset.removeNode()
```
kozlovic, line 2311: I assume that this will take care of shutting down the existing "monitorStream" go routine, right?

kozlovic (approved):
LGTM assuming confirmation to my assumption...

server/jetstream_cluster.go
derekcollison: Yes correct, I tested that..

PR #2875, opened 2022-02-17
[FIXED] Interest policy and staggered filtered consumers could fail to remove messages.

### LINKED ISSUE #2873: Interest policy: nats server won't delete acked messages if consumers are not keeping up
## Defect
when using Interest policy, nats server won't delete acked messages if consumers are not keeping up

 - [x] Included `nats-server -DV` output - unfortunately I have no access atm, please let me know if you need it

nats-server 2.7.2
python nats client 2.0.0
docker container nats 2.7.2 alpine

1) create jetstream with following config
```
Configuration:

             Subjects: messages.*
     Acknowledgements: true
            Retention: File - Interest
             Replicas: 1
       Discard Policy: New
     Duplicate Window: 2m0s
    Allows Msg Delete: true
         Allows Purge: true
       Allows Rollups: false
     Maximum Messages: unlimited
        Maximum Bytes: unlimited
          Maximum Age: unlimited
    Maximum Consumers: unlimited
```
2. create two polling consumers, one for messages.msg1 and another for messages.msg2

```
Configuration:

        Durable Name: MSG_XXX_CONSUMER
           Pull Mode: true
      Filter Subject: messages.message1
      Deliver Policy: All
          Ack Policy: Explicit
            Ack Wait: 30s
       Replay Policy: Instant
     Max Ack Pending: 20,000
   Max Waiting Pulls: 512
```

3. consumers are slower than inbound producer but ack messages normally within configured timeframe
4. messages get stacked up inside nats server filesystem, waiting for consumer to pick them up
5. stop producer

#### Expected result:
messages should be eventually all consumed, acked and removed from nats file system

#### Actual result:
messages are all consumed, acked, but stay on nats file system.
when we delete durable consumers from server and reregister consumers without producing any messages on stream, messages eventually disappear from nats server

Screenshot 2022-02-16 at 13 21 48

derekcollison:
On filtered subjects when consumers were staggered we need to disqualify a filtered consumer as not applicable before doing any sequence and ack floor determinations are made to see if we can delete a message from a stream.

Resolves #2873

/cc @nats-io/core

### DIFF
--- a/server/consumer.go
+++ b/server/consumer.go
@@ -2085,6 +2085,16 @@ func (o *consumer) needAck(sseq uint64) bool {
 	var asflr, osseq uint64
 	var pending map[uint64]*Pending
 	o.mu.RLock()
+
+	// Check first if we are filtered, and if so check if this is even applicable to us.
+	if o.isFiltered() && o.mset != nil {
+		subj, _, _, _, err := o.mset.store.LoadMsg(sseq)
+		if err != nil || !o.isFilteredMatch(subj) {
+			o.mu.RUnlock()
+			return false
+		}
+	}
+
 	if o.isLeader() {
 		asflr, osseq = o.asflr, o.sseq
 		pending = o.pending
--- a/server/jetstream_cluster_test.go
+++ b/server/jetstream_cluster_test.go
@@ -10667,6 +10667,75 @@ func TestJetStreamClusterStreamTagPlacement(t *testing.T) {
 	placeErr("C1", []string{"cloud:DO"})
 }
 
+func TestJetStreamClusterInterestRetentionWithFilteredConsumersExtra(t *testing.T) {
+	c := createJetStreamClusterExplicit(t, "R3S", 3)
+	defer c.shutdown()
+
+	subjectNameZero := "foo.bar"
+	subjectNameOne := "foo.baz"
+
+	// Client based API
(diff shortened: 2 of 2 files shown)


server/consumer.go
```diff
@@ -2085,6 +2085,16 @@ func (o *consumer) needAck(sseq uint64) bool {
 	var asflr, osseq uint64
 	var pending map[uint64]*Pending
 	o.mu.RLock()
+
+	// Check first if we are filtered, and if so check if this is even applicable to us.
+	if o.isFiltered() && o.mset != nil {
+		subj, _, _, _, err := o.mset.store.LoadMsg(sseq)
+		if err != nil || !o.isFilteredMatch(subj) {
+			o.mu.RUnlock()
```
kozlovic, line 2093: Oops, I see that you already have unlock the consumer line 2087. So move o.mu.RLock() from line 2087 down after the outer if, assuming that o.isFiltered() or o.isFilteredMatch() are not grabbing the lock.

kozlovic (changes requested):

server/consumer.go
derekcollison: I think its ok.

kozlovic: Sorry, I read as an RUnlock() above...

kozlovic (approved):
LGTM

PR #2877, opened 2022-02-17
Small improvements to send performance to a full stream.

derekcollison:
Cleaned up some locking and if fifo make index updates lazy like writeMsgRecord.

/cc @nats-io/core


kozlovic (approved):
LGTM

PR #2879, opened 2022-02-18
Various changes

kozlovic:
- Limit IPQueue logging
- Add a rand per raft node. This is because in some situations when
running 3 servers at the same time, we would end-up with identical
inboxes for different subjects on the different nodes which would
cause panics
- Move the creation of internal subscriptions after the tracking
of the node and its peers.

### DIFF
--- a/server/raft.go
+++ b/server/raft.go
@@ -197,6 +197,9 @@ type raft struct {
 	stepdown *ipQueue // of string
 	leadc    chan bool
 	quit     chan struct{}
+
+	// Random generator, used to generate inboxes for instance
+	prand *rand.Rand
 }
 
 // cacthupState structure that holds our subscription, and catchup term and index
@@ -340,6 +343,7 @@ func (s *Server) startRaftNode(cfg *RaftConfig) (RaftNode, error) {
 	sq := s.sys.sq
 	sacc := s.sys.account
 	hash := s.sys.shash
+	pub := s.info.ID
 	s.mu.Unlock()
 
 	ps, err := readPeerState(cfg.Store)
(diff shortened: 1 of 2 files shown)


server/raft.go
```diff
@@ -351,6 +355,12 @@ func (s *Server) startRaftNode(cfg *RaftConfig) (RaftNode, error) {
 	}
 
 	qpfx := fmt.Sprintf("RAFT [%s - %s] ", hash[:idLen], cfg.Name)
+	rsrc := time.Now().UnixNano()
+	if len(pub) >= 32 {
```
derekcollison, line 359: Why 32?

server/stream.go
```diff
@@ -3218,9 +3218,8 @@ func (mset *stream) internalLoop() {
 					pm.o.didNotDeliver(pm.seq)
 				}
 				pm.returnToPool()
+				c.flushClients(0)
```
derekcollison, line 3221: Does this have an impact on perf for simple / normal setups?

derekcollison (commented):
In general looks good, a few questions.

server/raft.go
kozlovic: This is because `highwayhash.New64()` requires a key that is 32 bytes long.

server/stream.go
kozlovic: I have undone this and the stall client changes. I would rather introduce these kind of changes in a later release.

kozlovic:
@derekcollison I have removed the changes to stall clients and flushClients() for now. Just keep the ipqueue logging changes and the raft's rand to create inboxes. Will revisit the other changes for a later release.

derekcollison:
Sounds good.

PR #2881, opened 2022-02-23
[FIXED] Gateway: connect could fail due to PING sent before CONNECT

kozlovic:
When a gateway connection was created (either accepted or initiated)
the timer to fire the first PING was started at that time, which
means that for an outbound connection, if the INFO coming from
the other side was delayed, it was possible for the outbound to
send the PING protocol before the CONNECT, which would cause
the accepting side to close the connection due to a "parse" error
(since the CONNECT for an inbound is supposed to be the very
first protocol).

Also noticed that we were not setting the auth timer like we do
for the other type of connections. If authorization{timeout:<n>}
is not set, the default is 2 seconds.


derekcollison (approved):
LGTM

wallyqs (approved):
LGTM 👍

PR #2887, opened 2022-02-25
[FIXED] JetStream: Some stream advisories missing

### LINKED ISSUE #2886: Some JetStream Asset CRUD events aren't being published
## Defect

A NATS user reported that:

_Hello, quick question about Jetstream Advisories. Looking at the [Monitoring Jetstream](https://docs.nats.io/running-a-nats-service/nats_admin/monitoring/monitoring_jetstream) documentation, there is an avisory which you title “Stream CRUD operations”. That suggests all operations on streams (create, update, delete) would fire events. However, we don’t seem to get events for other than CREATED events? The subject you have on that table suggests that as well as it’s $JS.EVENT.ADVISORY.STREAM.CREATED.<STREAM>, rather than $JS.EVENT.ADVISORY.STREAM.<STREAM_OPERATION>.<STREAM> . Is this on purpose?_

I verified this locally. After discussion on NATS slack, @kozlovic discovered that the stream closes its “internalLoop” responsible to send messages before sending the advisory.

It's possible other events are affected as well.

kozlovic:
The "deleted" advisory was missing because the stream's send loop
was closed before the advisory was pushed to the queue to be sent.

Added tests, both for single and clustered mode to test all stream
advisories.

Resolves #2886

### DIFF
--- a/server/jetstream_cluster_test.go
+++ b/server/jetstream_cluster_test.go
@@ -10849,6 +10849,162 @@ func TestJetStreamClusterMirrorOrSourceNotActiveReporting(t *testing.T) {
 	}
 }
 
+func TestJetStreamStreamAdvisories(t *testing.T) {
+	s := RunBasicJetStreamServer()
+	defer s.Shutdown()
+
+	if config := s.JetStreamConfig(); config != nil {
+		defer removeDir(t, config.StoreDir)
+	}
+
+	c := createJetStreamClusterExplicit(t, "JSC", 3)
+	defer c.shutdown()
+
+	checkAdv := func(t *testing.T, sub *nats.Subscription, expectedPrefixes ...string) {
+		t.Helper()
+		seen := make([]bool, len(expectedPrefixes))
+		for i := 0; i < len(expectedPrefixes); i++ {
+			msg := natsNexMsg(t, sub, time.Second)
+			var gotOne bool
+			for j, pfx := range expectedPrefixes {
+				if !seen[j] && strings.HasPrefix(msg.Subject, pfx) {
+					seen[j] = true
+					gotOne = true
+					break
+				}
+			}
+			if !gotOne {
+				t.Fatalf("Expected one of prefixes %q, got %q", expectedPrefixes, msg.Subject)
+			}
+		}
+	}
+
+	// Used to keep stream names pseudo unique. t.Name() has slashes in it which caused problems.
+	var testN int
+
+	checkAdvisories := func(t *testing.T, s *Server, replicas int) {
+
+		nc, js := jsClientConnect(t, s)
+		defer nc.Close()
+
+		testN++
+		streamName := "TEST_ADVISORIES_" + fmt.Sprintf("%d", testN)
+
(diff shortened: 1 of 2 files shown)


server/jetstream_cluster_test.go
```diff
@@ -10736,6 +10736,143 @@ func TestJetStreamClusterInterestRetentionWithFilteredConsumersExtra(t *testing.
...
+		// or after the restore_complete advisory because they are sent
+		// using different "send queues". That is, the restore uses the
+		// server's event queue while the stream_created is sent from
+		// the stream's own send queue.
+		rmsg = natsNexMsg(t, sub, time.Second)
+		if strings.HasPrefix(rmsg.Subject, JSAdvisoryStreamRestoreCompletePre) {
+			checkAdv(t, sub, JSAdvisoryStreamCreatedPre)
+		} else if strings.HasPrefix(rmsg.Subject, JSAdvisoryStreamCreatedPre) {
+			checkAdv(t, sub, JSAdvisoryStreamRestoreCompletePre)
+		} else {
```
kozlovic, line 10867: @derekcollison In clustered mode, the stream "created" event in this condition (after a restore) is never sent. The reason is in stream.go, addStreamWithAssignment has this code:
```
if isLeader {
    // Send advisory.
    var suppress bool
    if !s.standAloneMode() && sa == nil {
        suppress = true
    } else if sa != nil {
        suppress = sa.responded
    }
    if !suppress {
        mset.sendCreateAdvisory()
    }
}
```
and from restore, this function is invoked without stream assignment (sa is nil), so the advisory is suppressed. No idea what needs to be done here.

kozlovic:
@derekcollison I have created this PR with a test that checks for all stream advisory. The "deleted" advisory missing was due to a close of the channel before we scheduled the advisory (even this current fix may not be 100% reliable, since I am not certain that Go's select{} with case <-channel would consume always in the order they are listed. We know there will be a message pushed in the queue, but the close of the channel may be picked-up first?

Anyway, the other issue is as described in one review-comment: not sure how to make the advisory be sent in case of the restore of the stream in clustered mode.

derekcollison:
ok so next steps for this one are what?

kozlovic:
> ok so next steps for this one are what?

That you figure out how to have the server send the advisory, in cluster mode, when doing a stream recover. As I pointed out, the code suppress the advisory because the server is leader, clustered, but `sa` is nil.

derekcollison:
ok and the delete advisory is fixed to your satisfaction?

Should I simply extend this PR or create new one?

kozlovic:
You should extend, since there is the new test already (which obviously fails because of the missing "created" advisory on stream restore).
As for the delete, moving the close of channel seem to do the trick, but not sure if this will work 100%.

kozlovic (commented):
LGTM. I can't "approve" this PR since I am considered the author (GitHub won't let me select "Approve"). But I can confirm that the changes you made allow the update advisory to be seen in clustered/R3 and the create advisory after a snapshot restore.

ripienaar (approved):
LGTM

PR #2888, opened 2022-02-26
Adds deterministic bucketing (i.e. 'partitioning') to subject mapping

jnmoyne:
Add the ability to deterministically map one or more subject tokens to a 'bucket' (or 'partition') number (uses a hashing to map the value of the token(s) to a bucket number. Works everywhere subject mapping is used (i.e. account mappings and account imports).

Mapping syntax: `$<token number(s)>%<number of buckets>`

So `$1%10` maps to the value of the bucket number (i.e. a number between `0` and `9`) for the value of $1.

You can map the values of more than one subject token using `+`, e.g.: mapping `foo.*.* : foo.$1.$2.$1+2%10` means map the combination of $1 and $2 over 10 buckets. This example will translate subject `foo.a.1` to `foo.a.1.5` (bucket number is 5).

 - [X] Tests added
/cc @nats-io/core

### DIFF
--- a/server/accounts.go
+++ b/server/accounts.go
@@ -4125,19 +4125,20 @@ type transform struct {
 // Helper to pull raw place holder index. Returns -1 if not a place holder.
 func placeHolderIndex(token string) ([]int, int32) {
 	if len(token) > 1 {
-		var tp int
-		var tphnb int32
+		//var tp int
 		if token[0] == '$' {
-			if n, err := fmt.Sscanf(token, "$%d", &tp); err == nil && n == 1 {
-				return []int{tp}, -1
+
+			bucketingParts := strings.Split(token[1:], "%")
+			if len(bucketingParts) == 1 { // simple non-bucket mapping
+				tp, err := strconv.Atoi(bucketingParts[0])
+				if err == nil {
+					return []int{tp}, -1
+				}
 			}
-		}
-		if token[0] == '#' {
-			colunmSeparatedParts := strings.Split(token[1:], ":")
-			if len(colunmSeparatedParts) == 2 {
-				n, err := fmt.Sscanf(colunmSeparatedParts[1], "%d", &tphnb) // get the number of bucket
-				if err == nil && n == 1 {
-					tokenIndexes := strings.Split(colunmSeparatedParts[0], "+")
+			if len(bucketingParts) == 2 { // mapping to a bucket
+				tphnb, err := strconv.Atoi(bucketingParts[1])
+				if err == nil {
+					tokenIndexes := strings.Split(bucketingParts[0], "+")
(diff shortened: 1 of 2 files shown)


matthiashanel (changes requested):
I find `$1%10` or `$1+2%10` a pretty cryptic and not easily extensible notation.
Furthermore, new token notations retroactively constricts what subjects can look like.  

In previous discussions template tokens of the form `{{...}}` seem to be palatable and can easily be extended. For example: `{{hash(10,1}}`, `{{hash(10,1,2}}` 
If different bucketing schemes are asked for they have a place to go: `{{wasm:your-function}}`, `{{persistent-sharding(10,1)}}`
Independent of the configuration context, a user sees `{{...}}` and knows this is a substitution the server performs. 

I do not expect mappings to change often, so parsing could be done with a regex.

I do not see the unit test for stream import with local subject mapping.
Neither is there a test where multiple tokens are hashed.

jnmoyne:
Not sure what you mean by "Furthermore, new token notations retroactively constricts what subjects can look like."?

jnmoyne:
Also not sure how to do creating two accounts adding imports, creating credentials and then using those credentials to make two connections to test mapping during an import (in a unit test, I did the validate it by testing it manually and found that the same exact mapping code is used regardless of it being due to a mapping or an import). So I assumed there was no reason to repeat the same test once, once as a mapping and another one as an import, was that incorrect?

jnmoyne:
So looks like there were _no_ unit tests on subject mapping. I'm created two tests one for account mapping and another similar one for import that will test current the $-style mapping format both at the account level and import between accounts.

derekcollison:
They are tests for subject mappings for sure..

I agree with the templating idea as well, more work I know, but I think it will be worth it as @matthiashanel suggests.

derekcollison:
Also, unless very needed not to, could we squash all of these commits down to 1?

jnmoyne:
> They are tests for subject mappings for sure..

I couldn't find any... there are tests for account import that implicitly do a little bit of subject mapping, but only because they add a prefix (i.e. no mapping using token wildcards), maybe I'm not looking in the right place or missing something but as far as I can tell there were no tests for mapping token wildcards either at the account level or during an import...

> Also, unless very needed not to, could we squash all of these commits down to 1?

Appologies, it seems I managed to have a commit that has 2 parents (looks like when I did the git pull -rebase more than once maybe?) and Git will not let me squash commits because of having 2 parents...

derekcollison:
ok, check with Ivan or Phil who have much better `git fu` than I do.. ;)

jnmoyne:
Yeah, will create a new branch with just one commit which will include my upcoming changes so it's cleaner.

PR #2890, opened 2022-03-02
[ADDED] deterministic subject tokens to partition mapping

jnmoyne:
introduces 'Moustache' style subject mapping format (e.g. foo.*.* -> foo.{{wildcard(1)}}.{{wildcard(2)}}.{{partition(10,1,2)}})

 - [X] Tests added

/cc @nats-io/core

### DIFF
--- a/server/accounts.go
+++ b/server/accounts.go
@@ -18,6 +18,7 @@ import (
 	"encoding/hex"
 	"errors"
 	"fmt"
+	"hash/fnv"
 	"hash/maphash"
 	"io/ioutil"
 	"math"
@@ -25,6 +26,7 @@ import (
 	"net/http"
 	"net/textproto"
 	"reflect"
+	"regexp"
 	"sort"
 	"strconv"
 	"strings"
@@ -4114,18 +4116,73 @@ type transform struct {
 	src, dest string
 	dtoks     []string
 	stoks     []string
-	dtpi      []int8
+	dtpi      [][]int // destination token position indexes
+	dtpinp    []int32 // destination token position index number of partitions
 }
 
-// Helper to pull raw place holder index. Returns -1 if not a place holder.
-func placeHolderIndex(token string) int {
-	if len(token) > 1 && token[0] == '$' {
-		var tp int
-		if n, err := fmt.Sscanf(token, "$%d", &tp); err == nil && n == 1 {
-			return tp
+func getMappingFunctionArgs(functionName string, token string) ([]string, error) {
+	regEx := `{{` + functionName + ` *\((.*)\)}}`
+	r1, err := regexp.Compile(regEx)
+	if err != nil {
+		return nil, err
+	}
+	commandStrings := r1.FindStringSubmatch(token)
+	if len(commandStrings) > 1 {
+		ra, err := regexp.Compile(`,\s*`)
+		if err != nil {
+			return nil, err
+		}
+		args := ra.Split(commandStrings[1], -1)
+		return args, nil
+	}
+	return nil, nil
+}
+
+// Helper to pull raw place holder indexes and number of partitions. Returns -1 if not a place holder.
+func placeHolderIndex(token string) ([]int, int32) {
+	if len(token) > 1 {
+		// old $1, $2, etc... mapping format still supported to maintain backwards compatibility
+		if token[0] == '$' { // simple non-partition mapping
+			tp, err := strconv.Atoi(token[1:])
+			if err == nil {
+				return []int{tp}, -1
+			}
+		}
+
+		// New 'moustache' style mapping
+		// wildcard(wildcard token index) (equivalent to $)
+		args, err := getMappingFunctionArgs("wildcard", token)
+		if err == nil && args != nil {
+			if len(args) == 1 {
+				tp, err := strconv.Atoi(strings.Trim(args[0], " "))
+				if err == nil {
+					return []int{tp}, -1
+				}
+			}
+		}
+
+		// partition(number of partitions, token1, token2, ...)
+		args, err = getMappingFunctionArgs("partition", token)
+		if err == nil && args != nil {
+			if len(args) >= 2 {
+				tphnp, err := strconv.Atoi(strings.Trim(args[0], " "))
+				if err == nil {
+					var numPositions = len(args[1:])
+					tps := make([]int, numPositions)
+					for ti, t := range args[1:] {
+						i, err := strconv.Atoi(strings.Trim(t, " "))
+						if err == nil {
+							tps[ti] = i
+						} else {
+							return []int{-1}, -1
+						}
+					}
+					return tps, int32(tphnp)
+				}
+			}
 		}
 	}
-	return -1
+	return []int{-1}, -1
 }
 
 // newTransform will create a new transform checking the src and dest subjects for accuracy.
@@ -4139,7 +4196,8 @@ func newTransform(src, dest string) (*transform, error) {
 		return nil, ErrBadSubject
 	}
 
-	var dtpi []int8
+	var dtpi [][]int
+	var dtpinb []int32
 
 	// If the src has partial wildcards then the dest needs to have the token place markers.
 	if npwcs > 0 || hasFwc {
@@ -4153,25 +4211,30 @@ func newTransform(src, dest string) (*transform, error) {
 
 		nphs := 0
 		for _, token := range dtokens {
-			tp := placeHolderIndex(token)
-			if tp >= 0 {
-				if tp > npwcs {
-					return nil, ErrBadSubject
-				}
+			tp, nb := placeHolderIndex(token)
+			if tp[0] >= 0 {
 				nphs++
 				// Now build up our runtime mapping from dest to source tokens.
-				dtpi = append(dtpi, int8(sti[tp]))
+				var stis []int
+				for _, position := range tp {
+					if position > npwcs {
+						return nil, ErrBadSubject
+					}
+					stis = append(stis, sti[position])
+				}
+				dtpi = append(dtpi, stis)
+				dtpinb = append(dtpinb, nb)
 			} else {
-				dtpi = append(dtpi, -1)
+				dtpi = append(dtpi, []int{-1})
+				dtpinb = append(dtpinb, -1)
 			}
 		}
-
-		if nphs != npwcs {
+		if nphs < npwcs {
 			return nil, ErrBadSubject
 		}
 	}
 
-	return &transform{src: src, dest: dest, dtoks: dtokens, stoks: stokens, dtpi: dtpi}, nil
+	return &transform{src: src, dest: dest, dtoks: dtokens, stoks: stokens, dtpi: dtpi, dtpinp: dtpinb}, nil
 }
 
 // match will take a literal published subject that is associated with a client and will match and transform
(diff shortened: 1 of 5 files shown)


server/accounts.go
```diff
@@ -4114,18 +4116,73 @@ type transform struct {
...
 }
 
-// Helper to pull raw place holder index. Returns -1 if not a place holder.
-func placeHolderIndex(token string) int {
-	if len(token) > 1 && token[0] == '$' {
-		var tp int
-		if n, err := fmt.Sscanf(token, "$%d", &tp); err == nil && n == 1 {
-			return tp
+func getMappingFunctionArgs(functionName string, token string) ([]string, error) {
+	regEx := `{{` + functionName + ` *\((.*)\)}}`
```
kozlovic, line 4124: I would create 2 regex'es (as var at top of this file) and pass those regex as the func argument (instead of name). The compile should be done at a minimum. I checked the calling stack of this and it is not only called during processing of config, it can happen when processing a sub, etc.. so I would try to reduce the cost here.

server/accounts.go
```diff
@@ -4215,6 +4278,16 @@ func (tr *transform) transformSubject(subject string) (string, error) {
 	return tr.transform(tts)
 }
 
+func (tr *transform) getHashBucket(key string, numBuckets int) string {
+	h := fnv.New32a()
+	_, err := h.Write([]byte(key))
+	if err != nil {
+		return fmt.Sprintf("error: %v", err)
```
kozlovic, line 4285: For this hash, Write() cannot return any error, so would simply call `h.Write()` here... but if it did return an error, I think then that getHashBucket() would have to return (string, error) because sounds like bad things would happen if we were to use "error: <some error>" as a bucket name :-)

server/accounts.go
```diff
@@ -4114,18 +4116,73 @@ type transform struct {
...
-			return tp
+func getMappingFunctionArgs(functionName string, token string) ([]string, error) {
+	regEx := `{{` + functionName + ` *\((.*)\)}}`
+	r1, err := regexp.Compile(regEx)
+	if err != nil {
+		return nil, err
+	}
+	commandStrings := r1.FindStringSubmatch(token)
+	if len(commandStrings) > 1 {
+		ra, err := regexp.Compile(`,\s*`)
```
kozlovic, line 4131: Same thing here. Make this regex a "constant" by defining it at the top of the file.

kozlovic (changes requested):
First pass at the review.

server/accounts.go
jnmoyne: I checked the fnv source code and indeed it only ever will return nil as the err (I thought of passing that error back as the partition/bucket name as it's the only kind of signaling mechanism available that there was a problem with the hashing (though I didn't think a problem with the hash would ever happen since it's just a hash function :)). The simplification is implemented in https://github.com/nats-io/nats-server/pull/2890/commits/ceb47f0298443758546029c469dc29336c515fdb

server/accounts.go
jnmoyne: Optimization done in https://github.com/nats-io/nats-server/pull/2890/commits/ceb47f0298443758546029c469dc29336c515fdb

server/accounts.go
```diff
@@ -4114,18 +4120,64 @@ type transform struct {
...
-	dtpi      []int8
+	dtpi      [][]int // destination token position indexes
+	dtpinp    []int32 // destination token position index number of partitions
+}
+
+func getMappingFunctionArgs(functionRegEx *regexp.Regexp, token string) ([]string, error) {
+	commandStrings := functionRegEx.FindStringSubmatch(token)
+	if len(commandStrings) > 1 {
+		args := commaSeparatorRegEx.Split(commandStrings[1], -1)
+		return args, nil
```
matthiashanel, line 4131: return commaSeparatorRegEx.Split(), nil ?

server/accounts.go
```diff
@@ -4230,15 +4295,23 @@ func (tr *transform) transform(tokens []string) (string, error) {
...
 			if len(token) == 1 && token[0] == fwc {
 				break
 			}
 		} else {
 			// >= 0 means use source map index to figure out which source token to pull.
-			token = tokens[index]
+			if tr.dtpinp[i] > 0 { // there is a valid (i.e. not -1) value for number of partitions, this is a partition transform token
+				var keyForHashing string
+				for _, sourceToken := range tr.dtpi[i] {
+					keyForHashing = keyForHashing + tokens[sourceToken]
```
matthiashanel, line 4309: I'd assume most ppl only ever have one token to hash. 

However, because strings are immutable many objects are created and immediately discarded.
This is a nifty trick from our code base:
basically create a slice with an array as backing buffer.
append to it. (if the backing buffer is exceeded a bigger one will be allocated by append)

```
var (
  _ims           [64]byte
  ims            = _ims[:0]
)
for _, sourceToken := range tr.dtpi[i] {
    ims = append(ims, []byte(tokens[sourceToken]))
}
token = tr.getHashPartition(ims, int(tr.dtpinp[i]))
```

I assume getHashPartition signature changed to use []byte (which it internally casts to anyway)

server/accounts.go
```diff
@@ -4215,6 +4273,13 @@ func (tr *transform) transformSubject(subject string) (string, error) {
 	return tr.transform(tts)
 }
 
+func (tr *transform) getHashPartition(key string, numBuckets int) string {
+	h := fnv.New32a()
+	h.Write([]byte(key))
+
+	return fmt.Sprintf("%d", h.Sum32()%uint32(numBuckets))
```
matthiashanel, line 4280: I'd imagine, not having to read the format string helps.
`strconv.Itoa(h.Sum32()%uint32(numBuckets))`

matthiashanel (changes requested):

server/accounts.go
jnmoyne: You are correct that this being one of the few places actually in the code path for every message being published on a matching subject every little optimization does add up changed to `strconv.Itoa(h.Sum32()%uint32(numBuckets))`

server/accounts.go
jnmoyne: Thanks forgot to clean that up as part of how I wrote different iterations of the code, probably didn't change it because the compiler takes care of optimizing that anyways but it doesn't add anything to source code readability so should and will be cleaned up.

server/accounts.go
```diff
@@ -4114,18 +4120,64 @@ type transform struct {
 	src, dest string
 	dtoks     []string
 	stoks     []string
-	dtpi      []int8
+	dtpi      [][]int // destination token position indexes
+	dtpinp    []int32 // destination token position index number of partitions
+}
+
+func getMappingFunctionArgs(functionRegEx *regexp.Regexp, token string) ([]string, error) {
```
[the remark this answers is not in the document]
jnmoyne: Yeah, that was a leftover from before Ivan's suggestion to move the regex compilation at initialization time (and therefore with a regex per function name, which means I could then use mustCompile (doesn't return an err) rather than compile (which does)), and therefore now that err is indeed superfluous and can be removed from the signature, will clean up.

server/accounts.go
jnmoyne: Thanks for the example code, will do that - some pretty advanced optimizing here 😮 but since in the function that gets call for every subject being transformed even small improvements do add up!

server/accounts.go
```diff
@@ -4114,18 +4120,64 @@ type transform struct {
...
 }
 
-// Helper to pull raw place holder index. Returns -1 if not a place holder.
-func placeHolderIndex(token string) int {
-	if len(token) > 1 && token[0] == '$' {
-		var tp int
-		if n, err := fmt.Sscanf(token, "$%d", &tp); err == nil && n == 1 {
-			return tp
+// Helper to pull raw place holder indexes and number of partitions. Returns -1 if not a place holder.
+func placeHolderIndex(token string) ([]int, int32) {
```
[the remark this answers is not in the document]
jnmoyne: Right now it errors out using an ErrBadSubject which is not really right because the problem is not really with the subject (that you match on) but with interpreting the destination mapping (not really a subject per se). So I think I will add a new Error that will better help.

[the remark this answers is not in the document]
jnmoyne: Will also improve the error message that ends up being logged.

For example, used to be `Error adding mapping for "foo.*.*": invalid subject` and is now `Error adding mapping for "foo.*.*" to "foo.$1.{{wildcard(2)}}.{{partition(10,1,2,a)}}" : invalid subject mapping destination`

jnmoyne:
All requested changes implemented by latest commit

server/accounts.go
```diff
@@ -161,6 +163,10 @@ const (
 	Chunked
 )
 
+var commaSeparatorRegEx = regexp.MustCompile(`,\s*`)
+var partitionMappingFunctionRegEx = regexp.MustCompile(`{{partition *\((.*)\)}}`)
+var wildcardMappingFunctionRegEx = regexp.MustCompile(`{{wildcard *\((.*)\)}}`)
```
ripienaar, line 168: maybe `{{\s*wildcard\s*\((.*)\)\s*}}` to allow `{{ wildcard(foo) }}` also, that seems more likely than `{{wildcard (foo)}}` but since we support one we might as well do both ways allowing space? People who used moustache things will expect that to be allowed

server/accounts.go
```diff
@@ -4114,18 +4120,65 @@ type transform struct {
 	src, dest string
 	dtoks     []string
 	stoks     []string
-	dtpi      []int8
+	dtpi      [][]int // destination token position indexes
+	dtpinp    []int32 // destination token position index number of partitions
```
ripienaar, line 4124: we need these comments on all these 100s of short named variables, nice one :)

server/accounts.go
```diff
@@ -4114,18 +4120,65 @@ type transform struct {
...
-// Helper to pull raw place holder index. Returns -1 if not a place holder.
-func placeHolderIndex(token string) int {
-	if len(token) > 1 && token[0] == '$' {
-		var tp int
-		if n, err := fmt.Sscanf(token, "$%d", &tp); err == nil && n == 1 {
-			return tp
+func getMappingFunctionArgs(functionRegEx *regexp.Regexp, token string) []string {
+	commandStrings := functionRegEx.FindStringSubmatch(token)
+	if len(commandStrings) > 1 {
+		return commaSeparatorRegEx.Split(commandStrings[1], -1)
```
ripienaar, line 4130: could maybe be made into a single regex call using `FindAllStringSubmatch()` but need to do some experimentation to be sure :)

server/accounts_test.go
```diff
@@ -46,6 +47,30 @@ func simpleAccountServer(t *testing.T) (*Server, *Account, *Account) {
...
 
+func TestPlaceHolderIndex(t *testing.T) {
+	testString := "$1"
+	indexes, nbPartitions, err := placeHolderIndex(testString)
+
+	if err != nil || len(indexes) != 1 || indexes[0] != 1 || nbPartitions != -1 {
+		t.Fatalf("Error parsing %s", testString)
+	}
+
+	testString = "{{partition(10,1,2,3)}}"
```
ripienaar, line 58: test the variants with spaces in them too

ripienaar (changes requested):
looking good, some small comments

server/accounts.go
jnmoyne: Good suggestion, implemented by the latest commit

server/accounts.go
jnmoyne: Thanks. Yes all those short named variables don't make it easy to get up to speed on what the code does, I kept renaming 'dpi' to a long variable name as I was figuring out one by one what each letter in the short variable name stood for, and having those long variable names made it much easier to read and understand the source code. Then at the end I renamed them back to short names but did add the comments as it will greatly help (IMHO) others that may want to read the code for those functions and understand it.

One could probably spend quite a bit of time going over all of the server code and adding comments like these :)

server/accounts.go
jnmoyne: Maybe, but IMHO probably not much to gain from that, at least right now I re-use that commaSeparatorRegEx for all the commands. At this point I would leave it how it is and later when I have more time I could do some experimentation and eventually create another PR if there's an improvement to be had.

ripienaar (approved):
LGTM, but you'll definitely have to get ivan|derek|mh to approve also

kozlovic (approved):
LGTM

derekcollison (approved):
LGTM

PR #2893, opened 2022-03-03
[FIXED] Added "Nats-Msg-Id" header to prevent receiving duplicate messages with MQTT QoS 1

JheSue:
MQTT Client (qos=1) always re-publish messages when the connection is lost. It may casuse JetStream received duplicate messages.

This patch added "Nats-Msg-Id" header to messages coming from MQTT to solve this duplicate messages problem.

/cc @nats-io/core

### DIFF
--- a/server/mqtt.go
+++ b/server/mqtt.go
@@ -156,7 +156,8 @@ const (
 	mqttJSAStreamNames    = "SN"
 
 	// Name of the header key added to NATS message to carry mqtt PUBLISH information
-	mqttNatsHeader = "Nmqtt-Pub"
+	mqttNatsHeader      = "Nmqtt-Pub"
+	mqttNatsMsgIdHeader = "Nats-Msg-Id"
 
 	// This is how long to keep a client in the flappers map before closing the
 	// connection. This prevent quick reconnect from those clients that keep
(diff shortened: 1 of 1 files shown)


server/mqtt.go
```diff
@@ -2894,6 +2896,10 @@ func (s *Server) mqttProcessPub(c *client, pp *mqttPublish) error {
 	bb.WriteByte(':')
 	bb.WriteByte('0' + mqttGetQoS(pp.flags))
 	bb.WriteString(_CRLF_)
+	bb.Write(mqttNatsMsgIdHeaderB)
+	bb.WriteByte(':')
+	bb.WriteString(fmt.Sprintf("%x", pp.pi))
```
kozlovic, line 2901: Not sure what you meant by "always re-publish messages when the connection is lost", but I can't accept this PR. If you run a simple test (and notice that our test suite failed for the MQTT tests), you will see that valid messages will be deleted as duplicates, when they are not.
This is because the client library will likely reuse the packet identifier as soon as it gets a PUBACK. If packet identifiers were never reused, and since they are encoded as 16 bits, you would only be able send on a given topic only 65536 messages.. total, for ever. So that approach won't work.

I am not closing this PR directly since i am still curious what you meant by "always re-publish messages when the connection is lost".

kozlovic (changes requested):

server/mqtt.go
cfsghost: @kozlovic Client library get no PUBACK back that doesn't mean server receive nothing from client. Server probably got a message and sent PUBACK back but client doesn't get it since network issues. In this case, client will re-send message with duplicate flag cause server will receive duplicate message.

@JheSue It's "at least once" which is correct behavior for QoS 1. I think you want QoS 2. (Not supported by JetStream)

A solution is to modified you MQTT client to generate unique id for your message and put it to payload. You can  create a subscriber to forward messages to another stream. When you publish it to next stream, you can use unique id as "Nats-Msg-Id" to solve duplicate message problem as what you want.

kozlovic: @cfsghost You wrote:
```
Client library get no PUBACK back that doesn't mean server receive nothing from client. Server probably got a message and sent PUBACK back but client doesn't get it since network issues. In this case, client will re-send message with duplicate flag cause server will receive duplicate message.
```
Right, and that would be the "best-case" scenario where the change from this PR may have worked, but it doesn't work as I explained because if the client receives the PUBACK, it will recycle the packet identifier and therefore the next "new" message would be rejected as a duplicate.
Moreover, any other session having the same packet-id would then result in duplicates.

It is clear that this change won't work. I am going to close the PR now and will see if @JheSue comes back with some comments.

PR #2896, opened 2022-03-03
[FIXED] JetStream stream info consumers count in clustered mode

### LINKED ISSUE #2895: JetStream: In clustering mode, consumers count in Stream info/list may be wrong
In clustering mode, the number of consumers in stream info may be wrong in presence of non durable consumers. Ephemeral are handled by specific nodes. The StreamInfo response would contain only the consumer count that the stream leader is handling.

kozlovic:
In clustering mode, the number of consumers in stream info may be
wrong in presence of non durable consumers. Ephemeral are handled
by specific nodes. The StreamInfo response would contain only the
consumer count that the stream leader is handling.

This fix overrides the stream's state consumers count with the
number of consumers from the stream assignment record.

Resolves #2895

### DIFF
--- a/server/jetstream_api.go
+++ b/server/jetstream_api.go
@@ -1645,6 +1645,8 @@ func (s *Server) jsStreamInfoRequest(sub *subscription, c *client, a *Account, s
 		resp.ApiResponse.Type = JSApiStreamCreateResponseType
 	}
 
+	var clusterWideConsCount int
+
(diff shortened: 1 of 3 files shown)


matthiashanel (approved):
LGTM

derekcollison (approved):
LGTM - Awesome catch - thx

Issue #2885, opened 2022-02-25
JetStream $SYS folder keep growing


nayanparegi:
When we create a consumer with filter subject and that subject never got published the consumer sequence remain to 0 and raft folder size for that consumer will grow indefinitely.

derekcollison:
What server version please?
Can you provide more information on the system setup, topology and stream that the consumer is attached to?

nayanparegi:
@derekcollison  we are using 2.7.2.

We are running NATS inside the Kubernetes cluster.

The scenario is like below:
Stream Name: StreamA

Two different services running inside the Kubernetes cluster. Service A is subscribing to MessageA and Service B is subscribing to MessageB using durable subscription with filter subject StreamA.MessageA and StreamB.MessageB.

The different service is publishing MessageA and MessageB. However, frequency of MessageB is very rare, sometime once a week or two. Frequency of MessageA is around 10000 messages per minute.

Now when I look at the storage it is continuously growing unless I receive the any message for subject "MessageB"

When I tried to look at the storage, I found that RAFT group created for consumer for MessageB is growing in storage until it receives the message.

derekcollison:
Can you send us `nats stream info` for that stream and `nats consumer info` for all consumers as well?

Also, there is a fix for staggered but filtered consumers on retention policy streams that was released with 2.7.3, so might be worth considering upgrading if your environment allows.

nayanparegi:
@derekcollison  thanks for your reply. Below is the detail:

Connection Information:

               Client ID: 16718
               Client IP: 127.0.0.1
                     RTT: 152.6145ms
       Headers Supported: true
         Maximum Payload: 2.0 MiB
       Connected Cluster: nats
           Connected URL: nats://127.0.0.1:4222
       Connected Address: 127.0.0.1:4222
     Connected Server ID: NB3A3UAAEU26SGWXD464LT6OKPOQGBTU3PSHTSCN64NFANYPVC5EK4JL
   Connected Server Name: nats-2

JetStream Account Information:

           Memory: 0 B of 1.0 GiB
          Storage: 1.1 GiB of 102 GiB
          Streams: 2 of 10
        Consumers: 1,962 of Unlimited


And below is consumer which is having an issue:

Configuration:

        Durable Name: ********
         Description: ********
    Delivery Subject: _INBOX.WRFfN0AJ4FUjGYhwiQd9I_
      Filter Subject: ********
      Deliver Policy: All
 Deliver Queue Group: ********
[...]

PR #2897, opened 2022-03-03
Better startup logging to help debug RAFT to streams/consumers.

derekcollison:
In the process of working on #2885 with a user, I was struggling to map $SYS directories to consumer names.

This change allows a bit better logging on startup to more easily map a RAFT log directory etc to the stream/consumer.

/cc @nats-io/core


kozlovic (approved):
LGTM

aricart:
LGTM.

PR #2898, opened 2022-03-03
[CHANGED] JetStream: Redeliveries may be delayed if necessary

kozlovic:
We have seen situations where when a lot of pending messages accumulate,
there is a contention between the processing of the ACKs and the
checking of the pending map.

Decision is made to abort checking of pending list if processing of
ack(s) would be delayed because of that. The result is that a
redelivery may be post-poned.

Internally, the ACKs are also now using a queue to prevent processing
of them from the network handler, which could cause head-of-line
blocking, especially bad for routes.

### DIFF
--- a/server/consumer.go
+++ b/server/consumer.go
@@ -25,6 +25,7 @@ import (
 	"strconv"
 	"strings"
 	"sync"
+	"sync/atomic"
 	"time"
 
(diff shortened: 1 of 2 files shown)


derekcollison (approved):
LGTM!

PR #2899, opened 2022-03-04
[FIXED] Never used clustered and filtered consumers consume too much disk storage under $SYS

### LINKED ISSUE #2885: JetStream $SYS folder keep growing

derekcollison:
When a filtered consumer who has no state, meaning no messages are being processed, it still will receive updates to properly track the delivered sequence as it relates to the entire stream.
Since we did not have state we were inadvertently skipping the compaction logic for the raft store.

Resolves #2885

/cc @nats-io/core


kozlovic (approved):
LGTM

PR #2900, opened 2022-03-04
Can't recover stream on restart with deleted msg error.

derekcollison:
Fixed a bug that would not recover a stream properly after a non-clean shutdown with deleted messages.

/cc @nats-io/core

### DIFF
--- a/server/filestore.go
+++ b/server/filestore.go
@@ -4426,6 +4426,10 @@ func (mb *msgBlock) generatePerSubjectInfo() error {
 	for seq := fseq; seq <= lseq; seq++ {
 		sm, err := mb.cacheLookup(seq)
 		if err != nil {
+			// Since we are walking by sequence we can ignore some errors that are benign to rebuilding our state.
+			if err == ErrStoreMsgNotFound || err == errDeletedMsg {
+				continue
+			}
 			return err
 		}
 		if sm != nil && len(sm.subj) > 0 {
--- a/server/filestore.go
+++ b/server/filestore.go
@@ -315,17 +315,17 @@ func newFileStoreWithCreated(fcfg FileStoreConfig, cfg StreamConfig, created tim
 	// Always track per subject information.
 	fs.tms = true
 
+	// Recover our message state.
+	if err := fs.recoverMsgs(); err != nil {
+		return nil, err
+	}
+
 	// Write our meta data iff does not exist.
 	meta := path.Join(fcfg.StoreDir, JetStreamMetaFile)
 	if _, err := os.Stat(meta); err != nil && os.IsNotExist(err) {
 		if err := fs.writeStreamMeta(); err != nil {
 			return nil, err
 		}
-	}
-
-	// Recover our message state.
-	if err := fs.recoverMsgs(); err != nil {
-		return nil, err
 	}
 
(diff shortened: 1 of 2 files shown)


server/filestore.go
```diff
@@ -4426,6 +4426,10 @@ func (mb *msgBlock) generatePerSubjectInfo() error {
 	for seq := fseq; seq <= lseq; seq++ {
 		sm, err := mb.cacheLookup(seq)
 		if err != nil {
+			// Since we are walking by sequence we can ignore some errors that are benign to rebuilding our state.
```
kozlovic, line 4429: Since this seem to have been introduced by https://github.com/nats-io/nats-server/pull/2857, are the changes done there safe, or should the same type of error check be done?
https://github.com/nats-io/nats-server/pull/2857/files#diff-384c189826934c9a6fc3554dafc63dab2076245010e3d6fce5c71a93e15e9877R606
https://github.com/nats-io/nats-server/pull/2857/files#diff-384c189826934c9a6fc3554dafc63dab2076245010e3d6fce5c71a93e15e9877R4374

derekcollison: Good idea, will take a closer look.

derekcollison: I believe they are correct, however I have seen a report that I think is from this move so I may "undo" that one..

https://github.com/nats-io/nats-server/pull/2857/files#diff-384c189826934c9a6fc3554dafc63dab2076245010e3d6fce5c71a93e15e9877R319

derekcollison: Was hoping "no message cache" thing was similar to this one but I don't think its is per se. Still investigating.

derekcollison: FYI let me do the flip back on this PR as well. Will post new CL then squash after.

server/jetstream_test.go
```diff
@@ -15209,6 +15209,51 @@ func TestJetStreamPullConsumerHeartBeats(t *testing.T) {
...
+	// Restart.
+	s = RunJetStreamServerOnPort(-1, sd)
+	defer s.Shutdown()
+
+	nc, js = jsClientConnect(t, s)
+	defer nc.Close()
+
+	// Make sure we recovered our stream
+	_, err = js.StreamInfo("T")
+	require_NoError(t, err)
```
matthiashanel, line 15254: check for 99 messages and that `js.GetMsg("T", 22)` fails?

Am I right to assume that that upon recovery the missing fss file should exist again?

derekcollison: We have hundreds of tests that already do this. I understand where you are coming from and actually started to write that and was like, the stream info message count deleted messages have tests everywhere for over a year. So I say no.

derekcollison: Meaning I do not believe it buys us anything over the current tests that are in place.

derekcollison: The bug was in rebuilding the fss state. That file does not actually exists during the server running, only on exit as a short cut on restart.

derekcollison: Also just as an fyi, I always write the test to fail first and it did with "stream not found", which is what was reported, so as long as the fix had the stream being found I think we are good, IMO.

kozlovic (approved):
lgtm

PR #2901, opened 2022-03-05
[FIXED] LeafNode: queue sub interest not properly sent to new LN

kozlovic:
In complex situations, queue members count across various servers
may not be properly accounted for when sent to a new leafnode
connection.

The new test TestLeafNodeQueueGroupWithLateLNJoin has a drawing
of such setup, when after LN1 joined, and then queue members
were removed with 1 left, LN1 was told that there was no
more interest, so message published to LN1 would not reach
the remaining queue sub connected to LN2.


derekcollison (approved):
LGTM

matthiashanel (approved):
LGTM

PR #2902, opened 2022-03-05
Gateways: data race when setting first ping timer

kozlovic:
This was introduced when fixing #2881. The call to setFirstPingTimer
needed to be done under the client's lock.

Moved setFirstPingTimer from a server receiver to a client receiver.
The only reason it was a server receiver is because we need the
server options, but c.srv is always set when invoking this function,
so we will get the server from c.srv in that function now.

Related to #2881


derekcollison (approved):
LGTM

PR #2903, opened 2022-03-05
Set active for Sources and Mirrors to -1 on no contact

derekcollison:
If we had not heard from a source or mirror we would still calculated active.
This would wrap and create a large number which overflowed JSON's 2^53 limit.
This would cause schema violations in the NATS cli.

/cc @nats-io/core


ripienaar (approved):
LGTM guess I have to see what the CLI does with this new behaviour but for sure better than past behaviour

kozlovic (approved):
LGTM

PR #2904, opened 2022-03-06
[FIXED] Inconsistent durable consumer state after stream peer removal

derekcollison:
When removing a stream peer with an attached durable consumer, the consumer could become inconsistent.

/cc @nats-io/core

### DIFF
--- a/server/jetstream_cluster.go
+++ b/server/jetstream_cluster.go
@@ -1102,13 +1102,22 @@ func (js *jetStream) setConsumerAssignmentRecovering(ca *consumerAssignment) {
 	}
 }
 
-// Just copied over and changes out the group so it can be encoded.
+// Just copies over and changes out the group so it can be encoded.
 // Lock should be held.
 func (sa *streamAssignment) copyGroup() *streamAssignment {
 	csa, cg := *sa, *sa.Group
 	csa.Group = &cg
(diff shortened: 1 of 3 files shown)


server/jetstream_cluster.go
```diff
@@ -3106,7 +3127,7 @@ func (js *jetStream) monitorConsumer(o *consumer, ca *consumerAssignment) {
 			}
 			aq.recycle(&ces)
 		case isLeader = <-lch:
-			if !isLeader && n.GroupLeader() != noLeader {
+			if recovering && !isLeader {
```
kozlovic, line 3130: I have some doubts about how this work. Since above we do a aq.pop(), we get everything that is in the list at that time and process everything in the list before getting back to a select{}. The question is: is there cases where we would have something like: notify lch channel (say isLeader), then aq.push() x 10, then notify lch channel (say ! leader), then aq.push() x 20. If that is possible, then it is wrong because aq.pop() could get the 30 at once.. but regardless, if the notion of leader change matters in term of processing elements from the apply channel, then we can't have 2 different channels to notify.. would have to be in the same "queue".

kozlovic (approved):
lgtm, but a more general concern that is unlikely unfounded.

kozlovic (approved):
LGTM

PR #2905, opened 2022-03-07
Made improvements to scale up and down for streams and consumers.

derekcollison:

/cc @nats-io/core

### DIFF
--- a/server/consumer.go
+++ b/server/consumer.go
@@ -749,6 +749,16 @@ func (o *consumer) checkQueueInterest() {
 	}
 }
 
+// clears our node if we have one. When we scale down to 1.
+func (o *consumer) clearNode() {
+	o.mu.Lock()
+	defer o.mu.Unlock()
+	if o.node != nil {
+		o.node.Delete()
+		o.node = nil
+	}
+}
+
 // Lock should be held.
 func (o *consumer) isLeader() bool {
 	if o.node != nil {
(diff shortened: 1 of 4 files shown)


server/consumer.go
```diff
@@ -749,6 +749,16 @@ func (o *consumer) checkQueueInterest() {
...
 }
 
+// clears our node if we have one. When we scale down to 1.
+func (o *consumer) clearNode() {
+	o.mu.Lock()
+	if o.node != nil {
+		o.node.Delete()
+		o.node = nil
+	}
+	defer o.mu.Unlock()
```
kozlovic, line 759: Weird to have the defer here. Either you have this after lock, or remove defer.

kozlovic: Note also that although that we often check for o.node != nil, I did not find that we were setting it to nil, except now.

server/stream.go
```diff
@@ -3241,6 +3240,11 @@ func (mset *stream) internalLoop() {
 			c.flushClients(0)
 			outq.recycle(&pms)
 		case <-msgs.ch:
+			// This can possible change now so needs to be checked here.
```
kozlovic, line 3243: Typo? "This can possible change"

kozlovic (changes requested):

server/consumer.go
derekcollison: Will fix.

We set it in setConsumerAssignment.

kozlovic (approved):
lgtm

Issue #2882, opened 2022-02-24
ReservedStore and ReservedMemory report 18EB


variadico:
## Description
Here's an excerpt from a `/varz` endpoint. It shows that 18 exabytes have been reserved for JetStream storage and memory.
```
    "jetstream": {
      "config": {
        "max_memory": 25255962624,  // 25GB
        "max_storage": 15415142400, // 15GB
      },
      "stats": {
        "memory": 0,
        "storage": 0,
        "reserved_memory": 18446744073709552000,  // 18EB (1EB = 1000PB)
        "reserved_storage": 18446744073709552000, // 18EB (1EB = 1000PB)
        "api": {
          "total": 14,
          "errors": 0
        }
      },
```

I have a suspicion that 18EB is coming from these lines: https://github.com/nats-io/nats-server/blob/main/server/jetstream.go#L1740-L1741

```go
type JetStreamStats struct {
	ReservedMemory uint64            `json:"reserved_memory"`
	ReservedStore  uint64            `json:"reserved_storage"`
}
type jetStream struct {
	memReserved   int64
	storeReserved int64
}

// Convert reserved int64 to uint64
stats.ReservedMemory = (uint64)(js.memReserved)
stats.ReservedStore = (uint64)(js.storeReserved)
```

If `js.memReserved` or `js.storeReserved` is negative, then we could be assigning a number close to the number seen in the `/varz` info.

```go
var a int64 = -1
fmt.Println(uint64(a)) // prints 18446744073709551615, which is 18EB
```

@bwerthmann originally discovered this while editing existing streams to set max bytes (Inf => 1024), then deleting the streams.

## Steps to Reproduce
This issue can be reliably reproduced with the following commands.
```
nats-server-2.7.2 -m 8222 -js
```
```
nats stream create str1 --subjects=subj1
nats stream edit str1 --max-bytes=1234
nats stream rm str1
```
Then see the `reserved_storage` value in http://localhost:8222/varz.

## Fixes

It seems like the main issue is that we don't update the reserved bytes during a stream update. Below is a history of `js.storeReserved` as different operations are carried out.

1. Create stream with no MaxBytes: `js.storeReserved` → 0
2. Update stream with MaxBytes=1234: `js.storeReserved` → 0
3. Delete stream: `js.storeReserved` → -1234

Here are two ways we can prevent this issue:
[...]

PR #2907, opened 2022-03-07
Prevent reserved bytes underflow

variadico:
Currently, there is a bug where `js.storeReserved` and `js.memReserved` can underflow, causing `/varz` metrics to be incorrect. Below is a history of the state changes.

1. Create stream with no MaxBytes: `js.storeReserved` → 0
2. Update stream with MaxBytes=1234: `js.storeReserved` → 0 (should have reserved 1234)
3. Delete stream: `js.storeReserved` → -1234

**Bug:**`js.storeReserved=-1234` is an `int` that later gets converted to a `uint64`, which causes an underflow.

This change releases the old MaxBytes and reserves the new MaxBytes during the stream update step. This way, when a stream is deleted, these values don't become negative.

Resolves https://github.com/nats-io/nats-server/issues/2882 (see for additional info)

### DIFF
--- a/server/jetstream_test.go
+++ b/server/jetstream_test.go
@@ -7134,7 +7134,13 @@ gateway {
 
 	var servers []*Server
 	for _, name := range []string{"a0", "a1", "a2", "b0", "b1", "b2"} {
-		servers = append(servers, getServer(t, name))
+		s := getServer(t, name)
+		if config := s.JetStreamConfig(); config != nil {
+			defer removeDir(t, config.StoreDir)
+		}
+		defer s.Shutdown()
+
+		servers = append(servers, s)
 	}
 
 	checkClusterFormed(t, servers[:3]...)
--- a/server/jetstream_test.go
+++ b/server/jetstream_test.go
@@ -6827,7 +6827,7 @@ port: %d
 
 jetstream: {
   enabled: true
-  store_dir: %s
+  store_dir: '%s'
   max_mem: %d
   max_file: %d
 }
@@ -7088,7 +7088,7 @@ port: %d
 
 jetstream: {
   enabled: true
-  store_dir: %s
+  store_dir: '%s'
   max_mem: %d
   max_file: %d
 }
--- a/server/jetstream_test.go
+++ b/server/jetstream_test.go
@@ -6913,7 +6913,8 @@ cluster {
 		}
 		return fmt.Errorf("large server is not leader")
 	})
-	_, js := jsClientConnect(t, largeSrv)
+	nc, js := jsClientConnect(t, largeSrv)
+	defer nc.Close()
 
 	cases := []struct {
 		name           string
@@ -7187,7 +7188,8 @@ gateway {
 		}
 		return streams
 	}
-	_, js := jsClientConnect(t, servers[0])
+	nc, js := jsClientConnect(t, servers[0])
+	defer nc.Close()
 
 	cases := []struct {
 		name           string
--- a/server/jetstream_test.go
+++ b/server/jetstream_test.go
@@ -6802,6 +6802,222 @@ func TestJetStreamSystemLimits(t *testing.T) {
 	}
 }
 
+func TestJetStreamSystemLimitsPlacement(t *testing.T) {
+	const smallSystemLimit = 128
+	const mediumSystemLimit = smallSystemLimit * 2
+	const largeSystemLimit = smallSystemLimit * 3
+
+	getServer := func(t *testing.T, serverName string) *Server {
+		storeDir, err := os.MkdirTemp(tempRoot, "jstests-storedir-")
+		require_NoError(t, err)
+
+		natsPort, clusterPort := 24000, 26000
+		systemLimit := smallSystemLimit
+		if serverName == "medium" {
+			natsPort, clusterPort = 24001, 26001
+			systemLimit = mediumSystemLimit
+		} else if serverName == "large" {
+			natsPort, clusterPort = 24002, 26002
+			systemLimit = largeSystemLimit
+		}
+
+		s, _ := RunServerWithConfig(createConfFile(t, []byte(fmt.Sprintf(`
+server_name: %s
+port: %d
+
+jetstream: {
+  enabled: true
+  store_dir: %s
+  max_mem: %d
+  max_file: %d
+}
+
+server_tags: [%s]
+
+cluster {
+  name: cluster-a
+  port: %d
+  routes: [
+    nats-route://127.0.0.1:26000
+    nats-route://127.0.0.1:26001
+    nats-route://127.0.0.1:26002
+  ]
+}
+	`,
+			serverName,
+			natsPort,
+			storeDir,
+			systemLimit,
+			systemLimit,
+			serverName,
+			clusterPort,
+		))))
+
+		return s
+	}
+
+	smallSrv := getServer(t, "small")
+	defer smallSrv.Shutdown()
+	if config := smallSrv.JetStreamConfig(); config != nil {
+		defer removeDir(t, config.StoreDir)
+	}
+
+	mediumSrv := getServer(t, "medium")
+	defer mediumSrv.Shutdown()
+	if config := mediumSrv.JetStreamConfig(); config != nil {
+		defer removeDir(t, config.StoreDir)
+	}
+
+	largeSrv := getServer(t, "large")
+	defer largeSrv.Shutdown()
+	if config := largeSrv.JetStreamConfig(); config != nil {
+		defer removeDir(t, config.StoreDir)
+	}
+
+	checkClusterFormed(t, smallSrv, mediumSrv, largeSrv)
+
+	requestLeaderStepDown := func(clientURL string) error {
+		nc, err := nats.Connect(clientURL)
+		if err != nil {
+			return err
+		}
+		defer nc.Close()
+
+		ncResp, err := nc.Request(JSApiLeaderStepDown, nil, time.Second)
+		if err != nil {
+			return err
+		}
+
+		var resp JSApiLeaderStepDownResponse
+		if err := json.Unmarshal(ncResp.Data, &resp); err != nil {
+			return err
+		}
+		if resp.Error != nil {
+			return resp.Error
+		}
+		if !resp.Success {
+			return fmt.Errorf("leader step down request not successful")
+		}
+
+		return nil
+	}
+
+	// Force large server to be leader
+	checkFor(t, 10*time.Second, 500*time.Millisecond, func() error {
+		if largeSrv.JetStreamIsLeader() {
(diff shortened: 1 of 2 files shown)


server/stream.go
```diff
@@ -1047,6 +1047,9 @@ func (mset *stream) update(config *StreamConfig) error {
 		}
 	}
 
+	mset.js.releaseStreamResources(&ocfg)
+	mset.js.reserveStreamResources(cfg)
```
matthiashanel, line 1051: Have you checked the lock ordering here?  This change wants the jetstream lock while holding the mset lock.
In the create/destroy case that happens outside mset.mu.

matthiashanel (commented):
Would you mind adding the following tests?
1) Reserve 2/3rd of the total account limit, then update to one extra byte.  
2) Reserve 2/3rd of the total server limit, then update to one extra byte. 

I believe we may have an accounting issue there as well. 

Probably two scenarios apply to going down by a byte as well.

server/stream.go
```diff
@@ -970,7 +970,40 @@ func (jsa *jsAccount) configUpdateCheck(old, new *StreamConfig) (*StreamConfig,
...
 	// Check limits.
-	if err := jsa.checkAllLimits(&cfg); err != nil {
+	if old.MaxBytes == cfg.MaxBytes {
+		// If MaxBytes isn't being changed, then we can run the normal limit
+		// checking code.
+		if err := jsa.checkAllLimits(&cfg); err != nil {
+			return nil, err
+		}
+		return &cfg, nil
+	}
```
matthiashanel, line 980: is this section really needed? wouldn't it basically be the same issue, just for nothing has changed?
no adjustment has been made, where I'd expect a value of 0.

If I understand this right, the code below this should produce just that

matthiashanel (changes requested):

server/stream.go
```diff
@@ -1055,6 +1090,11 @@ func (mset *stream) update(config *StreamConfig) error {
 	}
 	mset.mu.Unlock()
 
+	if js != nil {
+		js.releaseStreamResources(&ocfg)
+		js.reserveStreamResources(cfg)
```
matthiashanel, line 1095: In addition, the same diff is needed here.
diff := oldMaxBytes - newMaxBytes 
diff > 0 -> reserve diff
diff == 0 -> do nothing
diff < 0 -> release by -diff

matthiashanel (changes requested):

variadico:
The newer commits add the following new tests. They caught some bugs, so the original handling of MaxBytes was also updated.
```
--- PASS: TestStorageReservedBytes (0.06s)
    --- PASS: TestStorageReservedBytes/file_reserve_66%_of_system_limit (0.01s)
    --- PASS: TestStorageReservedBytes/memory_reserve_66%_of_system_limit (0.00s)

    --- PASS: TestStorageReservedBytes/file_update_past_system_limit (0.00s)
    --- PASS: TestStorageReservedBytes/memory_update_past_system_limit (0.00s)

    --- PASS: TestStorageReservedBytes/file_update_to_system_limit (0.00s)
    --- PASS: TestStorageReservedBytes/memory_update_to_system_limit (0.00s)

    --- PASS: TestStorageReservedBytes/file_reserve_66%_of_account_limit (0.00s)
    --- PASS: TestStorageReservedBytes/memory_reserve_66%_of_account_limit (0.00s)

    --- PASS: TestStorageReservedBytes/file_update_to_account_limit (0.00s)
    --- PASS: TestStorageReservedBytes/memory_update_to_account_limit (0.00s)
```

variadico:
Added more tests to verify that the leader's system limits don't override non-leader server's limits.
```
--- PASS: TestJetStreamSystemLimitsPlacement (2.11s)
    --- PASS: TestJetStreamSystemLimitsPlacement/file_create_large_stream_on_small_server (0.00s)
    --- PASS: TestJetStreamSystemLimitsPlacement/memory_create_large_stream_on_small_server (0.00s)

    --- PASS: TestJetStreamSystemLimitsPlacement/file_create_large_stream_on_medium_server (0.00s)
    --- PASS: TestJetStreamSystemLimitsPlacement/memory_create_large_stream_on_medium_server (0.00s)

    --- PASS: TestJetStreamSystemLimitsPlacement/file_create_large_stream_on_large_server (0.01s)
    --- PASS: TestJetStreamSystemLimitsPlacement/memory_create_large_stream_on_large_server (0.00s)
```

matthiashanel (approved):
LGTM

server/jetstream_test.go
```diff
@@ -6801,6 +6802,532 @@ func TestJetStreamSystemLimits(t *testing.T) {
...
+			systemLimit = largeSystemLimit
+		}
+
+		s, _ := RunServerWithConfig(createConfFile(t, []byte(fmt.Sprintf(`
+server_name: %s
+port: %d
+
+jetstream: {
+  enabled: true
+  store_dir: %s
```
kozlovic, line 6830: Use `'%s'` for any file storage in config files, so that can run on Windows.

server/jetstream_test.go
```diff
@@ -6801,6 +6802,532 @@ func TestJetStreamSystemLimits(t *testing.T) {
...
+			systemLimit,
+			serverName,
+			clusterPort,
+		))))
+
+		return s
+	}
+
+	smallSrv := getServer(t, "small")
+	defer smallSrv.Shutdown()
```
kozlovic, line 6860: Move this after the `defer removeDir()` so that works on Windows.

server/jetstream_test.go
```diff
@@ -6801,6 +6802,532 @@ func TestJetStreamSystemLimits(t *testing.T) {
...
+	if config := smallSrv.JetStreamConfig(); config != nil {
+		defer removeDir(t, config.StoreDir)
+	}
+
+	mediumSrv := getServer(t, "medium")
+	defer mediumSrv.Shutdown()
```
kozlovic, line 6866: Move this after the `defer removeDir()` so that works on Windows.

server/jetstream_test.go
```diff
@@ -6801,6 +6802,532 @@ func TestJetStreamSystemLimits(t *testing.T) {
...
+	if config := mediumSrv.JetStreamConfig(); config != nil {
+		defer removeDir(t, config.StoreDir)
+	}
+
+	largeSrv := getServer(t, "large")
+	defer largeSrv.Shutdown()
```
kozlovic, line 6872: Move this after the `defer removeDir()` so that works on Windows.

server/jetstream_test.go
```diff
@@ -6801,6 +6802,532 @@ func TestJetStreamSystemLimits(t *testing.T) {
...
+		if largeSrv.JetStreamIsLeader() {
+			return nil
+		}
+
+		if err := requestLeaderStepDown(largeSrv.ClientURL()); err != nil {
+			return err
+		}
+		return fmt.Errorf("large server is not leader")
+	})
+	_, js := jsClientConnect(t, largeSrv)
```
kozlovic, line 6916: You should get the nats connection here (can name what you want) so that you can do a `defer <conn name>.Close()` so that this connection does not try to reconnect at the end of the test.

server/jetstream_test.go
```diff
@@ -6801,6 +6802,532 @@ func TestJetStreamSystemLimits(t *testing.T) {
...
+			t.Fatalf("unknown server name: %s", serverName)
+		}
+
+		s, _ := RunServerWithConfig(createConfFile(t, []byte(fmt.Sprintf(`
+server_name: %s
+port: %d
+
+jetstream: {
+  enabled: true
+  store_dir: %s
```
kozlovic, line 7090: Use `'%s'`

variadico:
Added new tests where we create a supercluster and ensure we can't create a large MaxBytes stream on a smaller cluster.

```
--- PASS: TestJetStreamSuperClusterSystemLimitsPlacement (5.22s)
    --- PASS: TestJetStreamSuperClusterSystemLimitsPlacement/file_create_large_stream_on_small_cluster_b0 (0.01s)
    --- PASS: TestJetStreamSuperClusterSystemLimitsPlacement/memory_create_large_stream_on_small_cluster_b0 (0.00s)
    --- PASS: TestJetStreamSuperClusterSystemLimitsPlacement/file_create_large_stream_on_small_cluster_b1 (0.00s)
    --- PASS: TestJetStreamSuperClusterSystemLimitsPlacement/memory_create_large_stream_on_small_cluster_b1 (0.00s)
    --- PASS: TestJetStreamSuperClusterSystemLimitsPlacement/file_create_large_stream_on_small_cluster_b2 (0.00s)
    --- PASS: TestJetStreamSuperClusterSystemLimitsPlacement/memory_create_large_stream_on_small_cluster_b2 (0.00s)
    --- PASS: TestJetStreamSuperClusterSystemLimitsPlacement/file_create_large_stream_on_large_cluster_a0 (0.01s)
    --- PASS: TestJetStreamSuperClusterSystemLimitsPlacement/memory_create_large_stream_on_large_cluster_a0 (0.01s)
    --- PASS: TestJetStreamSuperClusterSystemLimitsPlacement/file_create_large_stream_on_large_cluster_a1 (0.01s)
    --- PASS: TestJetStreamSuperClusterSystemLimitsPlacement/memory_create_large_stream_on_large_cluster_a1 (0.00s)
    --- PASS: TestJetStreamSuperClusterSystemLimitsPlacement/file_create_large_stream_on_large_cluster_a2 (0.00s)
    --- PASS: TestJetStreamSuperClusterSystemLimitsPlacement/memory_create_large_stream_on_large_cluster_a2 (0.00s)
```

server/jetstream_test.go
```diff
@@ -6801,6 +6802,532 @@ func TestJetStreamSystemLimits(t *testing.T) {
...
+		return fmt.Errorf("a0 server is not leader")
+	})
+	getStreams := func(jsm nats.JetStreamManager) []string {
+		var streams []string
+		for s := range jsm.StreamNames() {
+			streams = append(streams, s)
+		}
+		return streams
+	}
+	_, js := jsClientConnect(t, servers[0])
```
kozlovic, line 7190: Same comment than before

server/jetstream_test.go
```diff
@@ -6801,6 +6802,532 @@ func TestJetStreamSystemLimits(t *testing.T) {
...
+			clusterName,
+			gatewayPort,
+		))))
+
+		return s
+	}
+
+	var servers []*Server
+	for _, name := range []string{"a0", "a1", "a2", "b0", "b1", "b2"} {
+		servers = append(servers, getServer(t, name))
```
kozlovic, line 7135: You should defer shutdown of those servers, no?

kozlovic (changes requested):

kozlovic (approved):
LGTM

PR #2911, opened 2022-03-09
[FIXED] Some lock inversions

kozlovic:
The established ordering is client -> Account, so fixed few places
where we had Account -> client.

Added a new file, locksordering.txt with the list of known ordering
for some of the objects.


matthiashanel (approved):
LGTM

I assume the deleted assignments where highlighted by staticcheck or compiler?

kozlovic:
>  assume the deleted assignments where highlighted by staticcheck or compiler?

Yes, staticcheck, but an updated local version, would not be found on Travis yet since we use their latest release.

PR #2914, opened 2022-03-09
[FIXED] Consumer with no activity can lose quorum

### LINKED ISSUE #2913: Unused JetStream consumers go out of sync

Nats: 2.7.4 nightly (24067d7374e278385f8784040f3af454756b46b9)

Linux / k8s

* Have a stream with pull consumers on different subjects
* Publish a few hundred thousand messages to only some of the subjects, leaving some subjects completely unused

#### Expected result:

* All works. Nothing bad happens. Business as usual.

#### Actual result:

The unused consumers that never receive work go out of sync.
The log contains a lot of warnings like `JetStream cluster consumer '$G > [stream] > [consumer]' has NO quorum, stalled.` and errors like `RAFT [EEUOgTxU - C-R3F-IsTtTeSi] Error sending snapshot to follower [VpUo9efM]: raft: no snapshot available` and `Received an error looking up message for consumer: store is closed`.

This started to happen after the update from 2.7.2 to nightly, since I ran into the issue #2885. This also seems to be related, because it apparently affects the same consumers.

derekcollison:
When a consumer has no state we are now compacting the log, but were not snapshotting.

This caused issues on leader change and losing quorum.

Resolves #2913 

/cc @nats-io/core

### DIFF
--- a/server/consumer.go
+++ b/server/consumer.go
@@ -1760,7 +1760,7 @@ func (o *consumer) readStoredState() error {
 		return nil
 	}
 	state, err := o.store.State()
-	if err == nil && state != nil {
+	if err == nil && state != nil && state.Delivered.Consumer != 0 {
 		o.applyState(state)
 		if len(o.rdc) > 0 {
 			o.checkRedelivered()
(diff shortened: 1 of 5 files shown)


kozlovic (approved):
LGTM

matthiashanel (approved):
LGTM

PR #2917, opened 2022-03-09
Ensure file path is correct during stream restore

kozlovic:
Also had to change all references from `path.` to `filepath.` when
dealing with files, so that it works properly on Windows.

Fixed also lots of tests to defer the shutdown of the server
after the removal of the storage, and fixed some config files
directories to use the single quote `'` to surround the file path,
again to work on Windows.


matthiashanel (approved):
LGTM

philpennock (approved):

PR #2923, opened 2022-03-15
[CHANGED] Gateway: Detect duplicate names between clusters

kozlovic:
Gateway connection will be closed and error reported if a remote
has a name that is a duplicate of the local cluster.

### DIFF
--- a/server/client.go
+++ b/server/client.go
@@ -201,6 +201,7 @@ const (
 	ClusterNameConflict
 	DuplicateRemoteLeafnodeConnection
 	DuplicateClientID
+	DuplicateServerName
 )
 
 // Some flags passed to processMsgResults
--- a/server/errors.go
+++ b/server/errors.go
@@ -181,6 +181,10 @@ var (
 
 	// ErrCertNotPinned is returned when pinned certs are set and the certificate is not in it
 	ErrCertNotPinned = errors.New("certificate not pinned")
+
+	// ErrDuplicateServerName is returned when processing a server remote connection and
+	// the server reports that this server name is already used in the cluster.
+	ErrDuplicateServerName = errors.New("duplicate server name")
 )
 
 // configErr is a configuration error.
--- a/server/gateway.go
+++ b/server/gateway.go
@@ -1013,6 +1013,13 @@ func (c *client) processGatewayInfo(info *Info) {
 			return
 		}
 
+		// Check for duplicate server name with servers in our cluster
+		if s.isDuplicateServerName(info.Name) {
+			c.Errorf("Remote server has a duplicate name: %q", info.Name)
+			c.closeConnection(DuplicateServerName)
+			return
+		}
+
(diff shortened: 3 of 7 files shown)


kozlovic:
@derekcollison Also in this PR, I have added the cluster name (if present) at the top of the banner as we discussed, and did not remove the log statement that is added when the routing go routine is started (this is why it can appear much further down in the log with debug/trace enabled and state recovered). So with the change, it would look like this:
```
[95622] 2022/03/15 14:46:58.816611 [INF] Starting nats-server
[95622] 2022/03/15 14:46:58.816686 [INF]   Version:  2.7.5-beta
[95622] 2022/03/15 14:46:58.816688 [INF]   Git:      [not set]
[95622] 2022/03/15 14:46:58.816689 [INF]   Cluster:  local
[95622] 2022/03/15 14:46:58.816694 [INF]   Name:     NDRJVEA2WCMV4TS2MZCFQXUQ5IS34NAGMTXJYLNQJQPJKUL7JCLMVATD
[95622] 2022/03/15 14:46:58.816695 [INF]   ID:       NDRJVEA2WCMV4TS2MZCFQXUQ5IS34NAGMTXJYLNQJQPJKUL7JCLMVATD
[95622] 2022/03/15 14:46:58.817104 [INF] Listening for client connections on 0.0.0.0:4222
[95622] 2022/03/15 14:46:58.817293 [INF] Server is ready
[95622] 2022/03/15 14:46:58.817305 [INF] Cluster name is local
[95622] 2022/03/15 14:46:58.818017 [INF] Listening for route connections on localhost:6222
```
Notice the new `Cluster:  local` at the top and then later the existing `Cluster name is local`.

derekcollison (commented):
In general LGTM. Couple of questions.

Will the servers keep retrying?
Should we consider shutting the server down?

kozlovic:
@derekcollison 

> Will the servers keep retrying?

Yes, since connection is closed, it will be retried...

> Should we consider shutting the server down?

I thought about it, and indeed, closing the connection will result in retry so would be better to do a Fatalf() and be done. I also wonder now if this should be driven by a new config option (as it is now or replace with fatal) because you understand that if users have a deployment with same server names and they upgrade, then system will stop working (hence why I put the [CHANGED] title).

kozlovic:
Also, multiple servers may end-up shutting down. This is because the routing mesh may happen before the GW connections are initiated, so say I have a cluster "east" with names "east-0", "east-1" and "west-2" (yes, the last one is a mistake), with the other cluster "west" being "west-0", "west-1" and "west-2", if one of the "west-2" server connects to any server, say "east-0", then they will both exit with reporting there is a duplicate name "west-2", and that is because "east-0" will find the remote gateway server name "west-2" being a duplicate of its cluster list ("east-0", "east-1" and "west-2").

Maybe that's ok, but want to point that out.

derekcollison (approved):
LGTM

PR #2924, opened 2022-03-15
Test ack metrics

derekcollison:
Had a question about them and noticed we did not really have a test.

/cc @nats-io/core


kozlovic (approved):
LGTM

PR #2925, opened 2022-03-16
Offline streams behavior during list and delete improved.

derekcollison:
When a stream or consumer was offline we would not properly respond to a delete call.
We also would hang if no stream info requests were sent during a stream list due to the asset being offline.

/cc @nats-io/core

### DIFF
--- a/server/jetstream_cluster.go
+++ b/server/jetstream_cluster.go
@@ -2617,36 +2617,37 @@ func (js *jetStream) processClusterDeleteStream(sa *streamAssignment, isMember, 
 	js.mu.RLock()
 	s := js.srv
 	hadLeader := sa.Group.node == nil || sa.Group.node.GroupLeader() != noLeader
+	offline := s.allPeersOffline(sa.Group)
+	var isMetaLeader bool
+	if cc := js.cluster; cc != nil {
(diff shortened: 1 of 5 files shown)


server/jetstream_cluster.go
```diff
@@ -4336,14 +4339,14 @@ func (s *Server) jsClusteredStreamListRequest(acc *Account, ci *ClientInfo, filt
...
-	js.mu.Unlock()
+	js.mu.RUnlock()
 
 	const timeout = 4 * time.Second
 	notActive := time.NewTimer(timeout)
 	defer notActive.Stop()
 
 LOOP:
-	for {
+	for len(sent) > 0 {
```
kozlovic, line 4349: Well, it looks like we should then do the range streams loop before setting up the internal sub, because if all are offline, then this change here means that we will never wait for responses that we would have gotten in the msg callback.

kozlovic (commented):
Not sure if a change is needed or not...

server/jetstream_cluster.go
derekcollison: Not following, maybe quick f2f?

kozlovic: Got it.

PR #2927, opened 2022-03-17
Improve consumer state handling.

derekcollison:
 
/cc @nats-io/core

### DIFF
--- a/server/jetstream_cluster_test.go
+++ b/server/jetstream_cluster_test.go
@@ -11206,6 +11206,71 @@ func TestJetStreamRemovedPeersAndStreamsListAndDelete(t *testing.T) {
 	require_Error(t, err, nats.ErrStreamNotFound)
 }
 
+func TestJetStreamConsumerDeliverNewBug(t *testing.T) {
+	sc := createJetStreamSuperCluster(t, 3, 3)
+	defer sc.shutdown()
+
+	pcn := "C2"
+	sc.waitOnLeader()
+	ml := sc.leader()
+	if ml.ClusterName() == pcn {
+		pcn = "C1"
+	}
+
+	// Client based API
+	nc, js := jsClientConnect(t, ml)
+	defer nc.Close()
+
+	_, err := js.AddStream(&nats.StreamConfig{
+		Name:      "T",
+		Replicas:  3,
+		Placement: &nats.Placement{Cluster: pcn},
+	})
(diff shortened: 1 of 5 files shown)


server/jetstream_cluster_test.go
```diff
@@ -11206,6 +11206,71 @@ func TestJetStreamRemovedPeersAndStreamsListAndDelete(t *testing.T) {
...
+	c := sc.clusterForName(pcn)
+	for _, s := range c.servers {
+		sd := s.JetStreamConfig().StoreDir
+		s.Shutdown()
+		removeDir(t, sd)
+		s = c.restartServer(s)
+		c.waitOnServerHealthz(s)
+	}
+
+	c.waitOnConsumerLeader("$G", "T", "d")
```
matthiashanel, line 11262: the way I read your email, don't these two lines mask the issue?
The issue happened before the consumer leader was ready, but didn't happen after.
So shouldn't one instance of the test avoid these and a much shorter test test that once `c.waitOnServerHealthz(z)` returned the consumer has a leader (vs wait for one)

matthiashanel (commented):
Change itself makes sense. 
I just have one question about the test.

server/monitor.go
```diff
@@ -2772,21 +2783,23 @@ func (s *Server) HandleHealthz(w http.ResponseWriter, r *http.Request) {
 						if !cc.isStreamCurrent(accName, sname) {
 							health.Status = "unavailable"
 							health.Error = fmt.Sprintf("JetStream stream %q for account %q is not current", sname, accName)
-							w.WriteHeader(http.StatusServiceUnavailable)
 							js.mu.RUnlock()
 							break Err
 						}
+						// Now do consumers.
+						for _, o := range stream.getConsumers() {
```
kozlovic, line 2790: I wonder if there was not a reason you did not add consumers to the probe when you did the original PR... but I don't recall.

kozlovic (approved):
LGTM

server/monitor.go
derekcollison: In any case was an oversight, good catch by @matthiashanel

server/jetstream_cluster_test.go
derekcollison: Will merge but we can add on additional torture tests for state being wiped or corrupt for sure.

PR #2928, opened 2022-03-17
Show version on main monitoring page with link to source

derekcollison:
On main page for monitoring I put a version with a link that is gitCommit aware directly to the source.

image

/cc @nats-io/core


kozlovic (approved):
LGTM

PR #2929, opened 2022-03-17
Add stream limit update test

variadico:
This adds a test to see if we can update a stream when the stream limit is 1. Currently, this test fails, so we're skipping it. This test will be enabled in a future PR.


matthiashanel (approved):
LGTM

PR #2930, opened 2022-03-17
Detect exact duplicates for URLs for routes, gateways or leafnodes.

derekcollison:
If misconfigured could prevent the JetStream system from electing a leader. We will skip the entry and warn to the log.

/cc @nats-io/core

### DIFF
--- a/server/opts.go
+++ b/server/opts.go
@@ -1432,7 +1432,7 @@ func parseCluster(v interface{}, opts *Options, errors *[]error, warnings *[]err
 			}
 		case "routes":
 			ra := mv.([]interface{})
-			routes, errs := parseURLs(ra, "route")
+			routes, errs := parseURLs(ra, "route", warnings)
 			if errs != nil {
 				*errors = append(*errors, errs...)
 				continue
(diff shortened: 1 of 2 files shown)


derekcollison:
@wallyqs could you double check my warnings logic, not as familiar with that as you are.

kozlovic (approved):
LGTM

derekcollison:
Thanks.

Had a side effect that was making another test fail, so fixed that and want @wallyqs to eyeball the warning construction.

server/opts.go
```diff
@@ -1487,14 +1487,28 @@ func parseCluster(v interface{}, opts *Options, errors *[]error, warnings *[]err
...
+
 	for _, u := range a {
 		tk, u := unwrapValue(u, &lt)
 		sURL := u.(string)
+		if dd[sURL] {
+			err := &configWarningErr{
+				field: sURL,
+				configErr: configErr{
+					token:  tk,
+					reason: "Duplicate entry detected",
```
wallyqs, line 1505: How about adding the type of connection to the reason? Then it would read:
```suggestion
					reason: fmt.Sprintf("Duplicate %s entry detected", typ),
```

```
Duplicate route entry detected
```

kozlovic (approved):
LGTM

wallyqs (approved):
LGTM.

Currently it reads like this:

```
examples/js-cluster.conf:9:6: invalid use of field "nats://127.0.0.1:6224": Duplicate route entry detected
examples/js-cluster.conf:11:6: invalid use of field "nats://127.0.0.1:6225": Duplicate route entry detected
```

The lines already point to the location so that is enough I think, but we could make the warning clearer by changing the `field` from the warning error so that it reads something like this:

```
examples/js-cluster.conf:9:6: invalid use of field "routes": Duplicate route "nats://127.0.0.1:6224" entry detected
examples/js-cluster.conf:11:6: invalid use of field "routes": Duplicate route "nats://127.0.0.1:6225" entry detected
```

Have a commit here with the change showing how it could look like: https://github.com/nats-io/nats-server/commit/c7e357eb846f33df126ae5f447a578abf4c895f6

derekcollison:
I committed your suggestion already, maybe pull and see?

lmk

derekcollison:
Actually will merge and we can iterate on warning as needed.

PR #2931, opened 2022-03-18
Changes to IPQueues

kozlovic:
Removed the warnings, instead have a sync.Map where they are
registered/unregistered and can be inspected with an undocumented
monitor page.
Added the notion of "in progress" which is the number of messages
that have been pop()'ed. When recycle() is invoked this count
goes down.

### DIFF
--- a/server/consumer.go
+++ b/server/consumer.go
@@ -492,8 +492,12 @@ func (mset *stream) addConsumerWithAssignment(config *ConsumerConfig, oname stri
 	if c == nil {
 		return nil, NewJSStreamInvalidError()
 	}
+	var accName string
 	c.mu.Lock()
 	s, a := c.srv, c.acc
+	if a != nil {
+		accName = a.Name
+	}
 	c.mu.Unlock()
 
 	// Hold mset lock here.
@@ -570,7 +574,6 @@ func (mset *stream) addConsumerWithAssignment(config *ConsumerConfig, oname stri
 		maxdc:   uint64(config.MaxDeliver),
 		maxp:    config.MaxAckPending,
 		created: time.Now().UTC(),
-		ackMsgs: newIPQueue(),
 	}
 
 	// Bind internal client to the user account.
@@ -595,6 +598,9 @@ func (mset *stream) addConsumerWithAssignment(config *ConsumerConfig, oname stri
 			}
 		}
 	}
+	// Create ackMsgs queue now that we have a consumer name
+	o.ackMsgs = s.newIPQueue(fmt.Sprintf("[ACC:%s] consumer '%s' on stream '%s' ackMsgs", accName, o.name, mset.cfg.Name))
+
 	// Create our request waiting queue.
 	if o.isPullMode() {
 		o.waiting = newWaitQueue(config.MaxWaiting)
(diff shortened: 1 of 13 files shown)


server/consumer.go
```diff
@@ -595,6 +598,9 @@ func (mset *stream) addConsumerWithAssignment(config *ConsumerConfig, oname stri
 			}
 		}
 	}
+	// Create ackMsgs queue now that we have a consumer name
+	o.ackMsgs = s.newIPQueue(fmt.Sprintf("[ACC:%s] consumer '%s' on stream '%s' ackMsgs", accName, o.name, mset.cfg.Name))
```
derekcollison, line 602: This is totally fine, but for external logging I have been doing "'%s' > '%s' > '%s"" which is acc -> stream -> consumer

server/raft.go
```diff
@@ -197,6 +197,9 @@ type raft struct {
 	leadc    chan bool
 	quit     chan struct{}
 
+	// Account name of the asset this raft group is for
+	accName string
```
derekcollison, line 201: Maybe just make part of RaftConfig struct?

server/raft.go
```diff
@@ -353,7 +356,7 @@ func (s *Server) startRaftNode(cfg *RaftConfig) (RaftNode, error) {
 		return nil, errNoPeerState
 	}
 
-	qpfx := fmt.Sprintf("RAFT [%s - %s] ", hash[:idLen], cfg.Name)
+	qpfx := fmt.Sprintf("[ACC:%s] RAFT '%s' ", accName, cfg.Name)
```
derekcollison, line 359: Whatever we decide on format probably make a utility function so we only change once?

server/consumer.go
kozlovic: The reason I chose this format, is that in the monitoring page the ">" is escaped and so does not look good. For a stream, it would be something like this:
```
"stream '$G' \u003e 'TEST' messages": {
      "pending": 123
}
```
where as with my version it is:
```
"[ACC:$G] stream 'TEST' messages": {
      "pending": 123
}
```
The queues' name is only reported in the monitoring page, so how it looks there is important.

Also, with current format, you can easily select queues for a given account, something like:
```
http://localhost:8223/ipqueuesz?queues=ACC:<account name here>
```
The `queues` is a filter that I check with `strings.Contains()` (so technically, even with `'acc' > 'stream'` that would work since I would find the account name in the string..

server/raft.go
kozlovic: Not all queues are associated with an account, such as the system queue, etc.. but we could have an util function for the queue names that do have an account for sure.

server/raft.go
kozlovic: No, because the place I need it (https://github.com/nats-io/nats-server/pull/2931/files#diff-0a3d4e9b87bb3880bdb735fa489ef618b571de3370dae31d8f509b26e2fb852eR2128), I have only `*raft` and I don't see we keep reference of the config.

server/consumer.go
derekcollison: Sounds good.

derekcollison (approved):
LGTM

PR #2932, opened 2022-03-18
ensures the cluster info in jsz is sent from the leader only

ripienaar:
The data from other nodes are usually wrong, this can be quite
confusing for users so we now only send it when we are the leader

/cc @nats-io/core


server/monitor.go
```diff
@@ -2641,11 +2642,14 @@ func (s *Server) Jsz(opts *JSzOptions) (*JSInfo, error) {
 	}
 	s.js.mu.RUnlock()
 
-	if mg := s.js.getMetaGroup(); mg != nil {
-		if ci := s.raftNodeToClusterInfo(mg); ci != nil {
-			jsi.Meta = &MetaClusterInfo{Name: ci.Name, Leader: ci.Leader, Replicas: ci.Replicas, Size: mg.ClusterSize()}
+	if isLeader {
```
ripienaar, line 2645: I am a bit concerned here that if we just not set it like here users will get nil and code that was working before - they assumed cluster info was always present - would now encounter nil pointer exceptions.

So wdyt? I can make this just `jsi.Meta=&MetaClusterInfo{Leader: ci.Leader}` in that case - where we have a cluster but we arent a leader.  I think this will lead to fewer breakages as a result of this change, seems a better option than what we first thought.

server/monitor.go
```diff
@@ -1341,7 +1341,9 @@ func (s *Server) updateJszVarz(js *jetStream, v *JetStreamVarz, doConfig bool) {
 	v.Stats = js.usageStats()
 	if mg := js.getMetaGroup(); mg != nil {
 		if ci := s.raftNodeToClusterInfo(mg); ci != nil {
-			v.Meta = &MetaClusterInfo{Name: ci.Name, Leader: ci.Leader, Replicas: ci.Replicas, Size: mg.ClusterSize()}
+			if ci.Leader == s.info.Name {
```
ripienaar, line 1344: same comment here about not making nils for non leaders

derekcollison: Agree.

server/monitor.go
ripienaar: allright, I did that

server/js_test.go
```diff
@@ -0,0 +1,122 @@
+package server
```
kozlovic, line 1: Was this file added by error? Not sure the purpose of it (and staticcheck complains that withJetStream is never used):
```
server/js_test.go:11:6: func withJetStream is unused (U1000)
```

server/monitor.go
```diff
@@ -1341,7 +1341,11 @@ func (s *Server) updateJszVarz(js *jetStream, v *JetStreamVarz, doConfig bool) {
 	v.Stats = js.usageStats()
 	if mg := js.getMetaGroup(); mg != nil {
 		if ci := s.raftNodeToClusterInfo(mg); ci != nil {
-			v.Meta = &MetaClusterInfo{Name: ci.Name, Leader: ci.Leader, Replicas: ci.Replicas, Size: mg.ClusterSize()}
```
kozlovic, line 1344: Personal preference, but I would have done something like:
```
v.Meta = &MetaClusterInfo{Name: ci.Name, Leader: ci.Leader, Size: mg.ClusterSize()}
if ci.Leader == s.info.Name {
    v.Meta.Replicas = ci.Replicas
}
```
basically adding to the common setting when this is the leader.

server/monitor.go
```diff
@@ -2643,9 +2646,14 @@ func (s *Server) Jsz(opts *JSzOptions) (*JSInfo, error) {
 
 	if mg := s.js.getMetaGroup(); mg != nil {
 		if ci := s.raftNodeToClusterInfo(mg); ci != nil {
-			jsi.Meta = &MetaClusterInfo{Name: ci.Name, Leader: ci.Leader, Replicas: ci.Replicas, Size: mg.ClusterSize()}
+			if isLeader {
```
kozlovic, line 2649: Same comment than before

kozlovic (changes requested):
Suspicious server/js_test.go...

server/js_test.go
ripienaar: yes, doh. Will remove

server/monitor.go
ripienaar: yeah that seems better, i will fix.

ripienaar:
I did a force push to ensure that errant file isnt in history

server/monitor.go
```diff
@@ -2606,22 +2609,21 @@ func (s *Server) Jsz(opts *JSzOptions) (*JSInfo, error) {
 		opts.Accounts = true
 	}
 
-	// Check if we want a response from the leader only.
-	if opts.LeaderOnly {
```
kozlovic, line 2610: I am not familiar with `/jsz` endpoint, but I now realize that there was already an option to have just the result from the leader, no? So are we making this Jsz option deprecated?

kozlovic: That being said, at least from http monitoring, this option would never be set, so always false..

kozlovic (approved):
LGTM, but a comment about LeaderOnly option...

server/monitor.go
ripienaar: Indeed, the thing this tries to improve is that users a) dont know of this setting b) gets confused when non leaders have different information from the leader - only the leader has accurate information.

To avoid the confusion we'll rather not show the information where we know its wrong.

ripienaar: And now I wonder if the settings I am setting - leader - is known to be correct always even on non leader? @derekcollison ?

kozlovic: Agreed. Few days ago we were doing a Zoom with Derek and were checking the /jsz page and thought there was a problem, until Derek remembered that the info was correct only from the leader.

kozlovic (changes requested):
Actually, seems like test always failing, could you have a look?
```
=== RUN   TestServerEventsPingMonitorz/JSZ-24
    events_test.go:2101: Error receiving msg: nats: timeout
--- FAIL: TestServerEventsPingMonitorz (1.07s)
```

server/monitor.go
```diff
@@ -2606,22 +2609,21 @@ func (s *Server) Jsz(opts *JSzOptions) (*JSInfo, error) {
...
-		// So if we have JS but no clustering, we are the leader so allow.
-		if cc != nil {
-			js.mu.RLock()
-			isLeader := cc.isLeader()
-			js.mu.RUnlock()
-			if !isLeader {
-				return nil, fmt.Errorf("%w: not leader", errSkipZreq)
-			}
-		}
+	js, cc := s.getJetStreamCluster()
```
kozlovic, line 2612: Maybe move this block after line 2637? What I am not sure is, if `!s.JetStreamEnabled()` condition down there would cover this check that you did here or not.. if so, the the new error that you send here is not needed.

kozlovic: Or, to avoid "races", I would replace the call to `if !s.JetStreamEnabled() {` with:
```
	jsi := &JSInfo{
		ID:  s.ID(),
		Now: time.Now().UTC(),
	}
	js, cc := s.getJetStreamCluster()
	if js == nil {
		jsi.Disabled = true
		return jsi, nil
	}

	// if we have JS but no clustering, we are the leader so allow by default
	isLeader := true
	if cc != nil {
...
```

ripienaar: Yeah seems so, lets see if they pass now for sure a bug in how i refactored here

ripienaar: Oh, didnt see your comment about the race before I pushed - but I guess it had that race before also - will check.

kozlovic (changes requested):
Well, the test I reported now passes, but the one you modified (I think) now fails, even without those changes:
```
=== RUN   TestMonitorJsz/cluster-info
    monitor_test.go:4241: received cluster info from non leader: leader server_5500, server: server_7500
```

server/monitor.go
kozlovic: And by race I meant that if we have `if !s.JetStreamEnabled()` and then few lines later call `s.getJetStreamCluster()`, then at that time, js may be returned as nil (if it was disabled in the meantime), so then you would still have to check again if js is nil or not. So to save on that, I would have replaced s.JetStreamEnabled() with above code: two birds in one stone.

ripienaar: Pushed a change for the race thing - we already called getJetStreamCluster() earlier so just chekced if its result was nil or not

kozlovic: Yes, just my suggestion was to move your code change past the place where we initialize JSInfo, etc.. but yeah, it is equivalent except that if JS was to be disabled and one day we actually use LeaderOnly, the current code would return "not leader" error instead of the jsi.Disabled result.
I think it would be more clear and consistent to have the code checking for cluster/leader after the test for js being enabled or not.

server/monitor.go
```diff
@@ -2606,32 +2609,30 @@ func (s *Server) Jsz(opts *JSzOptions) (*JSInfo, error) {
...
-	}
-
 	jsi := &JSInfo{
 		ID:  s.ID(),
 		Now: time.Now().UTC(),
 	}
-	if !s.JetStreamEnabled() {
+
+	js := s.getJetStream()
+	if js == nil {
```
kozlovic, line 2618: To be consistent with original check `if !s.JetStreamEnabled()`, please add this to this test:
```
if js == nil || !js.isEnabled() {
..
```

kozlovic (changes requested):

kozlovic (changes requested):

derekcollison:
Where are we on this one?

kozlovic:
@derekcollison I suggested @ripienaar to make some simplifications/modifications, which I would not mind doing, but not sure I can from his fork.

kozlovic:
@derekcollison @ripienaar Sorry, I can't do that from his branch.. RI, if you can't/don't want to do the changes we discussed last week, we could merge like this and then I can have a quick pass to make the changes. Just let me know your preference.

ripienaar:
@kozlovic I did the requested change, but GH having actions issues again so tests not triggering atm, but have a look and we can try get tests going in a bit

server/monitor.go
```diff
@@ -2606,33 +2609,26 @@ func (s *Server) Jsz(opts *JSzOptions) (*JSInfo, error) {
...
+	isLeader := js.cluster == nil || js.cluster.isLeader()
+	js.mu.RUnlock()
+
+	if opts.LeaderOnly && !isLeader {
+		return nil, fmt.Errorf("%w: not leader", errSkipZreq)
+	}
+
+	var accounts []*jsAccount
 
 	s.js.mu.RLock()
```
ripienaar, line 2633: we could surely do these things using `js` and not `s.js` thus avoiding an extra lock?

kozlovic: That would be nice indeed. Starts here and apparently until line 2649 with `*s.js.usageStats()`.

kozlovic (changes requested):
Good idea replacing `s.js.` with `js.` that we have already captured under lock.

kozlovic (approved):
LGTM

PR #2937, opened 2022-03-21
Disable some supercluster limit placement tests

variadico:
This disables some flaky placement tests. I need to keep debugging, but in the meantime I don't want to impact anyone else.


kozlovic (approved):
LGTM - yes, this one is failing on my local machine quite often for me.

PR #2940, opened 2022-03-21
Use random high port in placement test

variadico:
This updates TestJetStreamSuperClusterSystemLimitsPlacement to use random high ports.


derekcollison (commented):
I also fixed this in my current branch, I think the update to Docker new FS took port 700. I just shifted to 7001 and that worked.

kozlovic (approved):
LGTM

Issue #2941, opened 2022-03-22
JS Consumer's sampling events are not sent when sampling option is enabled using update configuration


boris-ilijic:
#### Versions of `nats-server` and affected client libraries used:
- nats-server: 2.7.4
- nats.go: 1.13.0
- natscli: 0.0.30

#### OS/Container environment:
- docker
- k8s
- ngs

#### Steps or code to reproduce the issue:
1. Create JS Consumer without specifying sampling option
2. Update existing JS Consumer using configuration which contains sampling option (e.g. `--sample=100`)

#### Expected result:
- Sampling events are sent on the subject: `$JS.EVENT.METRIC.>`

#### Actual result:
- Events are not sent on the subject: `$JS.EVENT.METRIC.>`
- Nats `consumer info` command reports that sampling is active
- See failing test [here](https://github.com/nats-io/nats-server/pull/2942)

PR #2942, opened 2022-03-22
Add failing test for updating JS Consumer with sampling option

boris-ilijic:
Test related to #2941


derekcollison:
Thanks, will put it on my list.

PR #2943, opened 2022-03-22
[CHANGED] Duplicates in authorization{} and accounts{} now detected

### LINKED ISSUE #2926: Client occasionally incorrectly assigned to global account
## Defect

Occasionally a client which is supposed to be in a specific account is added to the global `$G` account instead.

This is visible in the trace log because when the client is correctly added to the account (`lagoonRemote` in this example) you can see the subscription propagated to connected nodes with a trace like this after the client connects:

```
->> [RS+ lagoonRemote lagoon.sshportal.api sshportalapi 1]
```

However sometimes you see a trace like this:

```
->> [RS+ $G lagoon.sshportal.api sshportalapi 1]
```

Because this seems to be some kind of race condition, I haven't nailed down the exact scenario when it happens. However I've got an environment where it is reliably reproducible. See the link below to a repository containing a `docker-compose` environment which you can run locally to reproduce the problem. That repository contains the full `nats.conf`.

 - [x] Included a [Minimal, Complete, and Verifiable example](https://stackoverflow.com/help/mcve)

Server:

```
mre-nats-0-1  | [1] 2022/03/16 15:58:07.153560 [INF] Starting nats-server
mre-nats-0-1  | [1] 2022/03/16 15:58:07.153594 [INF]   Version:  2.7.4
mre-nats-0-1  | [1] 2022/03/16 15:58:07.153597 [INF]   Git:      [a86b84a]
mre-nats-0-1  | [1] 2022/03/16 15:58:07.153601 [DBG]   Go build: go1.17.8
```

Client:
```
~ # nats --version
0.0.28
```

Docker container `nats:2.7.4-alpine3.15` (server).
Docker container `natsio/nats-box:0.8.1` (client)

1. Check out this repository: https://github.com/smlx/nats-mre
2. Run `docker-compose up -d; ./find.bug.sh`

#### Expected result:

Client is always added to the `lagoonRemote` account.

#### Actual result:

Client is occasionally added to the global `$G` account.

kozlovic:
If accounts{} block is specified, authorization{} should not have
any user/password/token or users array defined.

The reason is that users parsed in accounts{} are associated with
their respective account but users parsed in authorization{} are
associated with the global account. If the same user name is
in both, and since internally the parsing of those 2 blocks is
completely random (even if layed out in the config in a specific
order), the outcome may be that a user is either associated with
an account or the default global account.

To minimize breaking changes, but still avoid this unexpected
outcome, the server will now detect if there are duplicate users
(or nkeys) inside authorization{} block itself, but also between
this block and accounts{}.
The check will also detect if accounts{} has any user/nkey, then
the authorization{} block should not have any user/password/token,
making this test similar to the check we had in authorization{}
block itself.

Resolves #2926

### DIFF
--- a/server/config_check_test.go
+++ b/server/config_check_test.go
@@ -917,13 +917,13 @@ func TestConfigCheck(t *testing.T) {
 			name: "when user authorization config has both token and users",
 			config: `
 		authorization = {
-                 token = "s3cr3t"
-		  users = [
-		    {
-		      user = "foo"
-		      pass = "bar"
-		    }
-		  ]
+			token = "s3cr3t"
+			users = [
+				{
+					user = "foo"
+					pass = "bar"
+				}
+			]
 		}
 		`,
 			err:       errors.New(`Can not have a token and a users array`),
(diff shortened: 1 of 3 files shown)


derekcollison (approved):
LGTM

PR #2947, opened 2022-03-23
[IMPROVED] Stream recovery and Memory Utilization

derekcollison:
Under certain circumstances a stream that was still trying to catch up at the upper layers above NRG could be elected leader staling consumers.

Improved memory usage through JetStream storage layer.

Previously we would rely more heavily on Go's garbage collector since when we loaded a block for an underlying stream we would pass references upward to avoid copies.

Now we always copy when passing back to the upper layers which allows us to not only expire our cache blocks but pool and reuse them.

The upper layers also had changes made to allow the pooling layer at that level to interoperate with the storage layer.

/cc @nats-io/core

### DIFF
--- a/server/consumer.go
+++ b/server/consumer.go
@@ -2149,8 +2149,9 @@ func (o *consumer) needAck(sseq uint64) bool {
 
 	// Check first if we are filtered, and if so check if this is even applicable to us.
 	if o.isFiltered() && o.mset != nil {
-		subj, _, _, _, err := o.mset.store.LoadMsg(sseq)
-		if err != nil || !o.isFilteredMatch(subj) {
+		var svp StoreMsg
+		sm, err := o.mset.store.LoadMsg(sseq, &svp)
+		if err != nil || !o.isFilteredMatch(sm.subj) {
 			o.mu.RUnlock()
 			return false
 		}
@@ -2587,9 +2588,9 @@ var (
 // Get next available message from underlying store.
 // Is partition aware and redeliver aware.
 // Lock should be held.
-func (o *consumer) getNextMsg() (subj string, hdr, msg []byte, sseq uint64, dc uint64, ts int64, err error) {
+func (o *consumer) getNextMsg() (*jsPubMsg, uint64, error) {
 	if o.mset == nil || o.mset.store == nil {
-		return _EMPTY_, nil, nil, 0, 0, 0, errBadConsumer
+		return nil, 0, errBadConsumer
 	}
 	seq, dc := o.sseq, uint64(1)
 	if o.hasSkipListPending() {
@@ -2614,8 +2615,13 @@ func (o *consumer) getNextMsg() (subj string, hdr, msg []byte, sseq uint64, dc u
 				continue
 			}
 			if seq > 0 {
-				subj, hdr, msg, ts, err = o.mset.store.LoadMsg(seq)
-				return subj, hdr, msg, seq, dc, ts, err
+				pmsg := getJSPubMsgFromPool()
+				sm, err := o.mset.store.LoadMsg(seq, &pmsg.StoreMsg)
+				if sm == nil || err != nil {
+					pmsg.returnToPool()
+					pmsg, dc = nil, 0
+				}
+				return pmsg, dc, err
 			}
 		}
 		// Fallback if all redeliveries are gone.
@@ -2626,11 +2632,12 @@ func (o *consumer) getNextMsg() (subj string, hdr, msg []byte, sseq uint64, dc u
 	if o.maxp > 0 && len(o.pending) >= o.maxp {
 		// maxp only set when ack policy != AckNone and user set MaxAckPending
 		// Stall if we have hit max pending.
-		return _EMPTY_, nil, nil, 0, 0, 0, errMaxAckPending
+		return nil, 0, errMaxAckPending
 	}
 
 	// Grab next message applicable to us.
-	subj, sseq, hdr, msg, ts, err = o.mset.store.LoadNextMsg(o.cfg.FilterSubject, o.filterWC, seq)
+	pmsg := getJSPubMsgFromPool()
+	sm, sseq, err := o.mset.store.LoadNextMsg(o.cfg.FilterSubject, o.filterWC, seq, &pmsg.StoreMsg)
 
 	if sseq >= o.sseq {
 		o.sseq = sseq + 1
@@ -2639,7 +2646,12 @@ func (o *consumer) getNextMsg() (subj string, hdr, msg []byte, sseq uint64, dc u
 		}
 	}
 
-	return subj, hdr, msg, sseq, dc, ts, err
+	if sm == nil {
+		pmsg.returnToPool()
+		return nil, 0, err
+	}
+
+	return pmsg, dc, err
 }
 
 // forceExpireFirstWaiting will force expire the first waiting.
@@ -2786,13 +2798,11 @@ func (o *consumer) loopAndGatherMsgs(qch chan struct{}) {
 	// Deliver all the msgs we have now, once done or on a condition, we wait for new ones.
 	for {
 		var (
-			seq, dc     uint64
-			subj, dsubj string
-			hdr         []byte
-			msg         []byte
-			err         error
-			ts          int64
-			delay       time.Duration
+			pmsg  *jsPubMsg
+			dc    uint64
+			dsubj string
+			err   error
+			delay time.Duration
 		)
 
 		o.mu.Lock()
@@ -2812,10 +2822,11 @@ func (o *consumer) loopAndGatherMsgs(qch chan struct{}) {
 			goto waitForMsgs
 		}
 
-		subj, hdr, msg, seq, dc, ts, err = o.getNextMsg()
+		// Grab our next msg.
+		pmsg, dc, err = o.getNextMsg()
 
 		// On error either wait or return.
-		if err != nil {
+		if err != nil || pmsg == nil {
 			if err == ErrStoreMsgNotFound || err == ErrStoreEOF || err == errMaxAckPending || err == errPartialCache {
 				goto waitForMsgs
 			} else {
@@ -2836,15 +2847,17 @@ func (o *consumer) loopAndGatherMsgs(qch chan struct{}) {
 		} else {
 			// We will redo this one.
 			o.sseq--
+			pmsg.returnToPool()
 			goto waitForMsgs
 		}
 
 		// If we are in a replay scenario and have not caught up check if we need to delay here.
 		if o.replay && lts > 0 {
-			if delay = time.Duration(ts - lts); delay > time.Millisecond {
+			if delay = time.Duration(pmsg.ts - lts); delay > time.Millisecond {
 				o.mu.Unlock()
 				select {
 				case <-qch:
+					pmsg.returnToPool()
 					return
 				case <-time.After(delay):
 				}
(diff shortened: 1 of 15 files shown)


server/stream.go
```diff
@@ -2880,21 +2887,26 @@ func (mset *stream) processJetStreamMsg(subject, reply string, hdr, msg []byte,
 		// Expected last sequence per subject.
 		if seq, exists := getExpectedLastSeqPerSubject(hdr); exists {
 			// TODO(dlc) - We could make a new store func that does this all in one.
-			_, lseq, _, _, _, err := mset.store.LoadLastMsg(subject)
+			var smv StoreMsg
+			var fseq uint64
+			sm, err := mset.store.LoadLastMsg(subject, &smv)
+			if sm != nil {
+				fseq = sm.seq
```
kozlovic, line 2894: So clearly here, it looks like you have LoadLastMsg() that can return an error and a non nil sm..

server/stream.go
```diff
@@ -3115,45 +3127,57 @@ func (mset *stream) processJetStreamMsg(subject, reply string, hdr, msg []byte,
...
+	}
 	jsPubMsgPool.Put(pm)
 }
 
 func (pm *jsPubMsg) size() int {
 	if pm == nil {
 		return 0
 	}
-	return len(pm.subj) + len(pm.reply) + len(pm.hdr) + len(pm.msg)
+	return len(pm.dsubj) + len(pm.reply) + len(pm.hdr) + len(pm.msg)
```
kozlovic, line 3180: This seem like a huge change, meaning that we could have gotten the size completely wrong. What was the impact of that?

server/stream.go
```diff
@@ -3253,37 +3277,51 @@ func (mset *stream) internalLoop() {
...
 
 	for {
 		select {
 		case <-outq.ch:
 			pms := outq.pop()
 			for _, pmi := range pms {
 				pm := pmi.(*jsPubMsg)
-				c.pa.subject = []byte(pm.subj)
-				c.pa.deliver = []byte(pm.dsubj)
+				c.pa.subject = []byte(pm.dsubj)
```
kozlovic, line 3289: Again, this looks like a big difference, knowing that newJSPubMsg() still sets subj param to subj field. So unless all calls to newJSPubMsg() have had the passed param changed, wonder how this all worked?

server/filestore.go
```diff
@@ -3412,37 +3487,42 @@ func (mb *msgBlock) cacheLookup(seq uint64) (*fileStoredMsg, error) {
 	}
 
 	// Parse from the raw buffer.
-	subj, hdr, msg, mseq, ts, err := msgFromBuf(buf, hh)
-	if err != nil {
+
+	//	subj, hdr, msg, mseq, ts, err := msgFromBuf(buf, sm, hh)
```
kozlovic, line 3491: Remove?

server/filestore.go
```diff
@@ -3509,63 +3589,78 @@ func msgFromBuf(buf []byte, hh hash.Hash64) (string, []byte, []byte, uint64, int
...
+	} else {
+		// Allow append below.
+		sm.buf = sm.buf[:0]
+	}
+	// To recycle the large blocks we can never pass back a reference, so need to copy for the upper
+	// layers and for us to be safe to expire, and recycle, the large msgBlocks.
 	end := dlen - 8
-	var mhdr, msg []byte
+
+	//	var mhdr, msg []byte
```
kozlovic, line 3612: remove?

server/filestore.go
```diff
@@ -3509,63 +3589,78 @@ func msgFromBuf(buf []byte, hh hash.Hash64) (string, []byte, []byte, uint64, int
...
+		li, end = li-bi, end-bi
+		sm.hdr = sm.buf[0:li:li]
+		sm.msg = sm.buf[li:end]
+		//mhdr = data[bi:li:li]
+		//msg = data[li:end:end]
 	} else {
-		msg = data[slen:end:end]
+		sm.buf = append(sm.buf, data[slen:end]...)
+		sm.msg = sm.buf[0 : end-slen]
+		//msg = data[slen:end:end]
```
kozlovic, line 3627: remove

server/filestore.go
```diff
@@ -3509,63 +3589,78 @@ func msgFromBuf(buf []byte, hh hash.Hash64) (string, []byte, []byte, uint64, int
...
 }
 
 // LoadMsg will lookup the message by sequence number and return it if found.
-func (fs *fileStore) LoadMsg(seq uint64) (string, []byte, []byte, int64, error) {
-	sm, err := fs.msgForSeq(seq)
-	if sm != nil {
-		return sm.subj, sm.hdr, sm.msg, sm.ts, nil
+func (fs *fileStore) LoadMsg(seq uint64, sm *StoreMsg) (*StoreMsg, error) {
+	fsm, err := fs.msgForSeq(seq, sm)
+	if fsm != nil {
```
kozlovic, line 3636: I checked and it seems like msgForSeq only returns fsm != nil if error is nil, and conversely, if err is != nil, fsm will be nil, so let's keep it clear here.

server/consumer.go
```diff
@@ -2783,16 +2795,16 @@ func (o *consumer) loopAndGatherMsgs(qch chan struct{}) {
 	inch := o.inch
 	o.mu.Unlock()
 
+	//	var smv StoreMsg
```
kozlovic, line 2798: remove?

kozlovic (changes requested):
Initial review, will need a second (or third) look ;-)

server/stream.go
```diff
@@ -670,13 +674,14 @@ func (mset *stream) rebuildDedupe() {
 		return
 	}
 
+	var smv StoreMsg
 	state := mset.store.State()
 	for seq := sseq; seq <= state.LastSeq; seq++ {
-		_, hdr, _, ts, err := mset.store.LoadMsg(seq)
+		sm, err := mset.store.LoadMsg(seq, &smv)
```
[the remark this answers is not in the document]
derekcollison: For memstore we use sm.copy() which does the right thing. For filestore I will add a clear method that properly wipes all fields before using.

server/stream.go
```diff
@@ -2241,12 +2246,13 @@ func (mset *stream) setStartingSequenceForSource(sname string) {
 		return
 	}
 
+	var smv StoreMsg
 	for seq := state.LastSeq; seq >= state.FirstSeq; seq-- {
-		_, hdr, _, _, err := mset.store.LoadMsg(seq)
-		if err != nil || len(hdr) == 0 {
+		sm, err := mset.store.LoadMsg(seq, &smv)
+		if err != nil || sm == nil || len(sm.hdr) == 0 {
```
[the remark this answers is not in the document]
derekcollison: This is me being pedantic and more conservative these days. It does do the right thing. Will remove.

server/stream.go
derekcollison: No, it is doing the correct thing, would you prefer this to read err == nil?

server/stream.go
derekcollison: dsubj = "_INBOX.XX", subj = "foo", so could have been lower in our tests, but maybe higher in real use cases depending on length origin sub.

server/stream.go
derekcollison: The embedded StoreMsg now has the origin subject, dsubj is what jsPubMsg needs to set c.pa.subject to. Looks weird since it got flipped but I believe is correct and consistent.

In newJSPubMsg I left the arg names the same but will flip to be consistent with struct naming.

server/stream.go
kozlovic: The problem is that after this check you check for errors but in the form of `if err != nil || fseq != seq` so I guess you have a situation where you check for both error or sequence mismatch. So I guess it's fine like that..

server/raft.go
```diff
@@ -2134,11 +2161,12 @@ func (n *raft) catchupFollower(ar *appendEntryResponse) {
...
 
 func (n *raft) loadEntry(index uint64) (*appendEntry, error) {
-	_, _, msg, _, err := n.wal.LoadMsg(index)
+	var smp StoreMsg
+	sm, err := n.wal.LoadMsg(index, &smp)
 	if err != nil {
 		return nil, err
 	}
-	return n.decodeAppendEntry(msg, nil, _EMPTY_)
+	return n.decodeAppendEntry(sm.msg, nil, _EMPTY_)
```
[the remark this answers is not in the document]
derekcollison: I don't agree the runtime will hold reference to StoreMsg, it will hold one to the interior of the []byte from sm.msg.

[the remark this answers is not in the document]
derekcollison: That is ok to hold reference to though IMO.

server/memstore.go
```diff
@@ -603,15 +603,20 @@ func (ms *memStore) LoadMsg(seq uint64) (string, []byte, []byte, int64, error) {
...
 		}
-		return _EMPTY_, nil, nil, 0, err
+		return nil, err
 	}
-	return sm.subj, sm.hdr, sm.msg, sm.ts, nil
+
+	if smp == nil {
+		smp = new(StoreMsg)
+	}
+	sm.copy(smp)
```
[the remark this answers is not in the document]
derekcollison: Yes we always return a copy from a Store impl (MemStore and FileStore). That was needed to reuse blk buffers etc.

server/jetstream_cluster.go
```diff
@@ -1858,7 +1860,7 @@ func (js *jetStream) applyStreamEntries(mset *stream, ce *CommittedEntry, isReco
 				}
 
 				if err != nil && !isRecovering {
-					s.Debugf("JetStream cluster failed to delete msg %d from stream %q for account %q: %v",
+					s.Debugf("JetStream cluster failed to delete stream msg %d from '%s > %s': %v",
```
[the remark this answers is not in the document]
derekcollison: Yes will fix, thx!

server/raft.go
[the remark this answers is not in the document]
kozlovic: I would think that since sm.msg is not cleared and decodeAppendEntry references sm.msg, then sm cannot be reclaimed, but ok...

server/jetstream_cluster.go
```diff
@@ -5368,8 +5374,11 @@ RETRY:
 		case <-qch:
 			return nil
 		case isLeader := <-lch:
-			js.processStreamLeaderChange(mset, isLeader)
-			return nil
+			if isLeader {
+				n.StepDown()
+				notActive.Reset(activityInterval)
+				goto RETRY
+			}
```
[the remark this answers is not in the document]
derekcollison: We will bail out in due course when we do the retry and are not being helped. This will allow proper delays vs fast retry so opted for that.

server/filestore.go
derekcollison: Done in previous update. I saw it as I was edited filestore.go. Thx.

server/filestore.go
derekcollison: So only check err here is what you are saying correct?

derekcollison: We can just direct return, fixed.

kozlovic: Yes, regular error checking knowing that if err != nil, then fsm will be nil.

server/stream.go
```diff
@@ -670,13 +674,14 @@ func (mset *stream) rebuildDedupe() {
...
 		var msgId string
-		if err == nil && len(hdr) > 0 {
-			if msgId = getMsgId(hdr); msgId != _EMPTY_ {
-				mset.storeMsgIdLocked(&ddentry{msgId, seq, ts})
+		if err == nil && len(sm.hdr) > 0 {
+			if msgId = getMsgId(sm.hdr); msgId != _EMPTY_ {
+				mset.storeMsgIdLocked(&ddentry{msgId, sm.seq, sm.ts})
 			}
 		}
 		if seq == state.LastSeq {
```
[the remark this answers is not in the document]
derekcollison: Reworked a bit.

server/stream.go
kozlovic: Since you changed, does this need to be reverted?

server/memstore.go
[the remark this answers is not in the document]
kozlovic: I would recommend that you put a comment on LoadMsg (and the others) to say that the caller then "owns" the data (or that it is copied).

server/stream.go
kozlovic: Ok, so I checked and we always call newJSPubMsg() with a dsubj (the now first param) and empty subj (the second). So not sure why we have the second in newJSPubMsg(), unless we think we may need it later.

derekcollison: We used it for consumer msgs, but now they are constructed in a different way so could change that func.

kozlovic (approved):
LGTM

PR #2948, opened 2022-03-23
JS: BackOff list caused too frequent checkPending() calls

kozlovic:
Since the "next" timer value is set to the AckWait value, which
is the first element in the BackOff list if present, the check
would possibly happen at this interval, even when we were past
the first redelivery and the backoff interval had increased.

The end-user would still see the redelivery be done at the durations
indicated by the BackOff list, but internally, we would be checking
at the initial BackOff's ack wait.

I added a test that uses the store's interface to detect how many
times the checkPending() function is invoked. For this test it
should have been invoked twice, but without the fix it was invoked
15 times.

Also fixed an unrelated test that could possibly deadlock causing
tests to be aborted due to inactivity on Travis.

### DIFF
--- a/server/consumer.go
+++ b/server/consumer.go
@@ -3244,6 +3244,11 @@ func (o *consumer) checkPending() {
 	now := time.Now().UnixNano()
 	ttl := int64(o.cfg.AckWait)
 	next := int64(o.ackWait(0))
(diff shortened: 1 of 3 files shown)


derekcollison (approved):
LGTM!

PR #2950, opened 2022-03-24
[ADDED] server limits (max ack pending/dedupe window) to js config

matthiashanel:
Also shifting consumer config check to jsConsumerCreate as in clustered
mode this was enforced in the wrong place

 - [ ] Tests added

Resolves #

 -
 -
 -

/cc @nats-io/core

### DIFF
--- a/server/consumer.go
+++ b/server/consumer.go
@@ -328,7 +328,7 @@ func (mset *stream) addConsumer(config *ConsumerConfig) (*consumer, error) {
 
 func (mset *stream) addConsumerWithAssignment(config *ConsumerConfig, oname string, ca *consumerAssignment) (*consumer, error) {
 	mset.mu.RLock()
-	s, jsa := mset.srv, mset.jsa
+	s, jsa, tierName := mset.srv, mset.jsa, mset.tier
 	mset.mu.RUnlock()
 
 	// If we do not have the consumer currently assigned to us in cluster mode we will proceed but warn.
@@ -500,6 +500,13 @@ func (mset *stream) addConsumerWithAssignment(config *ConsumerConfig, oname stri
 	}
 	c.mu.Unlock()
 
+	jsa.mu.RLock()
+	jsaLimits, limitsFound := jsa.limits[tierName]
+	jsa.mu.RUnlock()
+	if !limitsFound {
+		return nil, NewJSNoLimitsError()
+	}
+
 	// Hold mset lock here.
 	mset.mu.Lock()
 	if mset.client == nil || mset.store == nil {
@@ -524,8 +531,8 @@ func (mset *stream) addConsumerWithAssignment(config *ConsumerConfig, oname stri
 	// than stream config we prefer the account limits to handle cases where account limits are
 	// updated during the lifecycle of the stream
 	maxc := mset.cfg.MaxConsumers
-	if maxc <= 0 || (mset.jsa.limits.MaxConsumers > 0 && mset.jsa.limits.MaxConsumers < maxc) {
-		maxc = mset.jsa.limits.MaxConsumers
+	if maxc <= 0 || (jsaLimits.MaxConsumers > 0 && jsaLimits.MaxConsumers < maxc) {
+		maxc = jsaLimits.MaxConsumers
 	}
 	if maxc > 0 && mset.numPublicConsumers() >= maxc {
 		mset.mu.Unlock()
--- a/server/consumer.go
+++ b/server/consumer.go
@@ -299,7 +299,7 @@ const (
 )
 
 // Helper function to set consumer config defaults from above.
-func setConsumerConfigDefaults(config *ConsumerConfig) {
+func setConsumerConfigDefaults(config *ConsumerConfig, lim *JSLimitOpts) {
 	// Set to default if not specified.
 	if config.DeliverSubject == _EMPTY_ && config.MaxWaiting == 0 {
 		config.MaxWaiting = JSWaitQueueDefaultMax
@@ -318,7 +318,11 @@ func setConsumerConfigDefaults(config *ConsumerConfig) {
 	}
 	// Set proper default for max ack pending if we are ack explicit and none has been set.
 	if (config.AckPolicy == AckExplicit || config.AckPolicy == AckAll) && config.MaxAckPending == 0 {
-		config.MaxAckPending = JsDefaultMaxAckPending
+		if lim.MaxAckPending > 0 && lim.MaxAckPending < JsDefaultMaxAckPending {
+			config.MaxAckPending = lim.MaxAckPending
+		} else {
+			config.MaxAckPending = JsDefaultMaxAckPending
+		}
 	}
 }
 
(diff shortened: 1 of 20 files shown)


server/stream.go
```diff
@@ -891,6 +895,9 @@ func checkStreamCfg(config *StreamConfig) (StreamConfig, error) {
 	if cfg.MaxAge != 0 && cfg.Duplicates > cfg.MaxAge {
 		return StreamConfig{}, fmt.Errorf("duplicates window can not be larger then max age")
 	}
+	if lim.Duplicates > 0 && cfg.Duplicates > lim.Duplicates {
+		return StreamConfig{}, fmt.Errorf("duplicates window can not be larger then server limit of %s", lim.Duplicates.String())
```
kozlovic, line 899: Detail, but since lim.Duplicates is a time.Duration, using `%v` for the format would do the right thing of printing as say `2m0s`.

server/opts.go
```diff
@@ -177,6 +177,11 @@ type RemoteLeafOpts struct {
 	tlsConfigOpts *TLSConfigOpts
 }
 
+type JSLimitOpts struct {
+	MaxAckPending int           `json:"-"`
```
kozlovic, line 181: Don't think we need the json tag for new additions (since Options are not json marshalled)

server/opts.go
```diff
@@ -226,6 +231,7 @@ type Options struct {
 	JetStreamDomain       string            `json:"-"`
 	JetStreamExtHint      string            `json:"-"`
 	JetStreamKey          string            `json:"-"`
+	JetStreamLimits       JSLimitOpts       `json:"-"`
```
kozlovic, line 234: Would not have added json tags with newer additions. I added a note long time ago at the top of Options:
```
// NOTE: This structure is no longer used for monitoring endpoints
// and json tags are deprecated and may be removed in the future.
```

kozlovic (changes requested):

server/consumer.go
```diff
@@ -318,106 +318,91 @@ func setConsumerConfigDefaults(config *ConsumerConfig) {
...
-	}
-
-	if config == nil {
-		return nil, NewJSConsumerConfigRequiredError()
-	}
-
-	// Make sure we have sane defaults.
-	setConsumerConfigDefaults(config)
-
+func checkConsumerCfg(config *ConsumerConfig, lim *JSLimitOpts, cfg *StreamConfig, acc *Account) (x *ApiError) {
```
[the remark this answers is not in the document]
matthiashanel: testing left over, removed

server/consumer.go
```diff
@@ -433,55 +418,94 @@ func (mset *stream) addConsumerWithAssignment(config *ConsumerConfig, oname stri
...
+	setConsumerConfigDefaults(config, lim)
+
+	if err := checkConsumerCfg(config, lim, &cfg, acc); err != nil {
+		return nil, err
+	}
+
+	sampleFreq := 0
+	if config.SampleFrequency != _EMPTY_ {
+		var err error
+		sampleFreq, err = strconv.Atoi(strings.TrimSuffix(config.SampleFrequency, "%"))
```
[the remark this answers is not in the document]
matthiashanel: had it the other way as well and switched back, but since you bring it up, I'll remove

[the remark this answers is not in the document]
kozlovic: You can have a one limit comment that says that it has been checked in checkConsumerConfig to be correct, so ignoring error here. Something like that.

server/jetstream_api.go
```diff
@@ -3311,6 +3302,17 @@ func (s *Server) jsConsumerCreate(sub *subscription, c *client, a *Account, subj
 	}
 
 	if s.JetStreamIsClustered() && !req.Config.Direct {
+		streamCfg, ok := js.clusterStreamConfig(acc.Name, streamName)
+		if !ok {
+			resp.Error = NewJSStreamNotFoundError(Unless(err))
+			s.sendAPIErrResponse(ci, acc, subject, reply, string(msg), s.jsonResponse(&resp))
+			return
+		}
+		if err := checkConsumerCfg(&req.Config, lim, &streamCfg, acc); err != nil {
```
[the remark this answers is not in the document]
matthiashanel: This is the call hierarchy for the new function.
You see that it gets called via jsConsumerCreate (the non replicated case) as well.
I went with the same pattern found for stream. (there however it's a wee bit less obvious)
Basically the non clustered case relies on the check in addstream and the clustered version calls it before replicating. 

(tbh, add stream/consumer should never check nor fail. but that would have made that change a lot more complex)

```
checkConsumerCfg
	*stream.addConsumerWithAssignment
		*jetStream.processClusterCreateConsumer
			*jetStream.restartClustered
			*jetStream.processConsumerAssignment
		*stream.addConsumer
			*Server.jsConsumerCreate
	*Server.jsConsumerCreate
		*Server.jsConsumerCreateRequest
		*Server.jsDurableCreateRequest
```

server/stream.go
matthiashanel: changed

server/opts.go
matthiashanel: removed

server/opts.go
matthiashanel: removed

server/jetstream_api.go
```diff
@@ -3225,22 +3225,12 @@ func (s *Server) jsConsumerCreate(sub *subscription, c *client, a *Account, subj
 		return
 	}
 
-	// We reject if flow control is set without heartbeats.
```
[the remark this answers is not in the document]
matthiashanel: basically, see next comment for more detail

server/opts.go
```diff
@@ -178,60 +178,60 @@ type RemoteLeafOpts struct {
...
-	Cluster               ClusterOpts       `json:"cluster,omitempty"`
-	Gateway               GatewayOpts       `json:"gateway,omitempty"`
-	LeafNode              LeafNodeOpts      `json:"leaf,omitempty"`
-	JetStream             bool              `json:"jetstream"`
-	JetStreamMaxMemory    int64             `json:"-"`
-	JetStreamMaxStore     int64             `json:"-"`
-	JetStreamDomain       string            `json:"-"`
-	JetStreamExtHint      string            `json:"-"`
-	JetStreamKey          string            `json:"-"`
-	JetStreamLimits       JSLimitOpts       `json:"-"`
```
kozlovic: Arg... too bad that this causes so much "difference" in the diff because of the formatting. But ultimately, we should remove those tags anyway.

server/consumer.go
```diff
@@ -504,11 +504,7 @@ func (mset *stream) addConsumerWithAssignment(config *ConsumerConfig, oname stri
 
 	sampleFreq := 0
 	if config.SampleFrequency != _EMPTY_ {
-		var err error
-		sampleFreq, err = strconv.Atoi(strings.TrimSuffix(config.SampleFrequency, "%"))
-		if err != nil {
-			return nil, NewJSConsumerInvalidSamplingError(err)
-		}
+		sampleFreq, _ = strconv.Atoi(strings.TrimSuffix(config.SampleFrequency, "%"))
```
kozlovic, line 507: Little comment above to explain that you know that it can't fail?

kozlovic (approved):
LGTM, but may add a comment

server/opts.go
matthiashanel: that's what I thought too, moving this field to the end seems worse though.

server/consumer.go
matthiashanel: fixed

kozlovic (approved):
LGTM

PR #2952, opened 2022-03-24
[FIXED] update of R1 Consumer in clustered setup.

matthiashanel:
missing reply caused timeout

 - [ ] Tests added

Resolves #

 -
 -
 -

/cc @nats-io/core

### DIFF
--- a/server/jetstream_cluster.go
+++ b/server/jetstream_cluster.go
@@ -2920,6 +2920,10 @@ func (js *jetStream) processClusterCreateConsumer(ca *consumerAssignment, state 
 		// Start our monitoring routine.
 		if rg.node == nil {
 			// Single replica consumer, process manually here.
+			js.mu.Lock()
+			// to force response in case we think we have responded before.
+			ca.responded = false
+			js.mu.Unlock()
 			js.processConsumerLeaderChange(o, true)
 		} else {
 			if !alreadyRunning {
--- a/server/jetstream_test.go
+++ b/server/jetstream_test.go
@@ -16565,3 +16565,64 @@ func Benchmark_JetStream10x1kWorker(b *testing.B) {
 func Benchmark_JetStream4x512Worker(b *testing.B) {
 	benchJetStreamWorkersAndBatch(b, 4, 512)
 }
+
+func TestJetStreamConsumerStreamUpdate(t *testing.T) {
+	test := func(t *testing.T, s *Server, replica int) {
+		nc := natsConnect(t, s.ClientURL())
+		defer nc.Close()
+		js, err := nc.JetStream()
+		require_NoError(t, err)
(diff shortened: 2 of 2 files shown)


derekcollison (approved):
LGTM! Thanks.

server/jetstream_cluster.go
```diff
@@ -2915,25 +2915,30 @@ func (js *jetStream) processClusterCreateConsumer(ca *consumerAssignment, state
 			}
 		}
 
+		update := false
 		// Start our monitoring routine.
 		if rg.node == nil {
 			// Single replica consumer, process manually here.
 			js.processConsumerLeaderChange(o, true)
+			update = true
```
kozlovic, line 2923: But processConsumerLeaderChange() will send something out (see bottom of that function). So it feels like you are going to send 2 responses in some cases?

kozlovic (commented):
Not sure if this is correct...

server/jetstream_cluster.go
derekcollison: Good catch.. hmm..

derekcollison: Need to clear ca.responded but that has to be under js lock.

https://github.com/nats-io/nats-server/blob/c36405ffbc05680d44e8bd75f3a480f1fe2a25ef/server/jetstream_cluster.go#L3345

derekcollison: Meaning clear it here before calling into processConsumerLeaderChange..

matthiashanel: basically ca.responded is never set to false, which is why it is not sent

matthiashanel: what's the use of this value even? shouldn't that be part of the stack, calling context?

matthiashanel: Implemented it as Derek suggested. but really, not sure why this variable needs to exist in the first place. 
Shouldn't we always respond when called via `applyMetaEntries` and never when not?

derekcollison: Its purpose is to not produce responses after we have already responded.

server/jetstream_cluster.go
```diff
@@ -2918,6 +2918,9 @@ func (js *jetStream) processClusterCreateConsumer(ca *consumerAssignment, state
 		// Start our monitoring routine.
 		if rg.node == nil {
 			// Single replica consumer, process manually here.
+			js.mu.Lock()
+			ca.responded = false
```
derekcollison, line 2922: Maybe add quick comment, // to force response in case we think we have responded before.

derekcollison (approved):
LGTM, small ask for comment.

derekcollison:
Merge when ready. Thx

PR #2958, opened 2022-03-26
[FIXED] Scaling up an R1 stream would not replicate existing messages.

### LINKED ISSUE #2955: upgrading a stream's replication factor is not working
```go
func TestJetStreamStreamReplicationUpgrade(t *testing.T) {
	c := createJetStreamClusterExplicit(t, "cluster", 3)
	defer c.shutdown()

	ncSys := natsConnect(t, c.randomServer().ClientURL(), nats.UserInfo("admin", "s3cr3t!"))
	defer ncSys.Close()

	nc := natsConnect(t, c.randomServer().ClientURL())
	defer nc.Close()

	js, err := nc.JetStream()
	require_NoError(t, err)
	cfg := &nats.StreamConfig{Name: "foo", Subjects: []string{"foo"}, Replicas: 1, Retention: nats.LimitsPolicy, MaxMsgs: 10}
	si, err := js.AddStream(cfg)
	require_NoError(t, err)
	require_True(t, si.Config.Replicas == 1)

	msg := [512]byte{}
	for i := 0; i < 10; i++ {
		js.Publish("foo", msg[:])
	}

	cfg.Replicas = 3
	si, err = js.UpdateStream(cfg)
	require_NoError(t, err)
	require_True(t, si.Config.Replicas == 3)
	require_True(t, si.State.Msgs == 10)

	time.Sleep(time.Minute)

	// INSTEAD: one server contains: s.Jsz().Store == blkfi.Size() == 5450 (this is what all server should look like)
	// all other server contain:     s.Jsz().Store == blkfi.Size() == 0
	// indicating that no catchup has happened

	refInfoSize := uint64(0)
	for _, s := range c.servers {
		jsi, err := s.Jsz(nil)
		require_NoError(t, err)
		require_True(t, jsi.Streams == 1)
		refInfoSize = jsi.Store
		fmt.Printf("storesize %d jsinfo %+v\n", refInfoSize, jsi)

		// s.Shutdown() tried shutdown to rule out some weird flush issue, but then 1.blk file is gone
		// (probably because it's empty)
	}

	time.Sleep(time.Minute)

	for _, s := range c.servers {
		blkfi, err := os.Stat(filepath.Join(s.StoreDir(), "$G/streams/foo/msgs/1.blk"))
		require_NoError(t, err)
		fmt.Printf("%d\n", blkfi.Size())
		require_True(t, refInfoSize == uint64(blkfi.Size()))
	}
}


```

derekcollison:
Also fixed a bug where we were incorrectly not spinning up the monitoring loop for a stream when going from 3->1->3.

Resolves #2955 

/cc @nats-io/core

### DIFF
--- a/server/jetstream_cluster.go
+++ b/server/jetstream_cluster.go
@@ -1479,7 +1479,7 @@ func (mset *stream) removeNode() {
 }
 
 // Monitor our stream node for this stream.
-func (js *jetStream) monitorStream(mset *stream, sa *streamAssignment) {
+func (js *jetStream) monitorStream(mset *stream, sa *streamAssignment, sendSnapshot bool) {
 	s, cc, n := js.server(), js.cluster, sa.Group.node
 	defer s.grWG.Done()
 
@@ -1543,6 +1543,14 @@ func (js *jetStream) monitorStream(mset *stream, sa *streamAssignment) {
 	// we replace with the restore chan.
 	restoreDoneCh := make(<-chan error)
 	isRecovering := true
+
+	// This is triggered during a scale up from 1 to clustered mode. We need the new followers to catchup,
+	// similar to how we trigger the catchup mechanism post a backup/restore. It's ok to do here and preferred
+	// over waiting to be elected, this just queues it up for the new members to see first and trigger the above
+	// RAFT layer catchup mechanism.
+	if sendSnapshot && mset != nil && n != nil {
+		n.SendSnapshot(mset.stateSnapshot())
+	}
 
 	for {
 		select {
@@ -1617,8 +1625,7 @@ func (js *jetStream) monitorStream(mset *stream, sa *streamAssignment) {
 			sa.Restore = nil
 			// If we were successful lookup up our stream now.
 			if err == nil {
-				mset, err = acc.lookupStream(sa.Config.Name)
-				if mset != nil {
+				if mset, err = acc.lookupStream(sa.Config.Name); mset != nil {
 					mset.setStreamAssignment(sa)
 				}
 			}
@@ -1647,7 +1654,7 @@ func (js *jetStream) monitorStream(mset *stream, sa *streamAssignment) {
 				panic("Finished restore but not leader")
 			}
 			// Trigger the stream followers to catchup.
-			if n := mset.raftNode(); n != nil {
+			if n = mset.raftNode(); n != nil {
 				n.SendSnapshot(mset.stateSnapshot())
 			}
 			js.processStreamLeaderChange(mset, isLeader)
@@ -1930,8 +1937,10 @@ func (js *jetStream) applyStreamEntries(mset *stream, ce *CommittedEntry, isReco
 				if err := json.Unmarshal(e.Data, &snap); err != nil {
 					return err
 				}
-				if err := mset.processSnapshot(&snap); err != nil {
-					return err
+				if !mset.IsLeader() {
+					if err := mset.processSnapshot(&snap); err != nil {
+						return err
+					}
 				}
 			}
 		} else if e.Type == EntryRemovePeer {
@@ -2272,6 +2281,11 @@ func (js *jetStream) processUpdateStreamAssignment(sa *streamAssignment) {
 	sa.consumers = osa.consumers
 	sa.err = osa.err
 
+	// If we detect we are scaling down to 1, non-clustered, and we had a previous node, clear it here.
+	if sa.Config.Replicas == 1 && sa.Group.node != nil {
+		sa.Group.node = nil
+	}
+
 	// Update our state.
 	accStreams[stream] = sa
 	cc.streams[acc.Name] = accStreams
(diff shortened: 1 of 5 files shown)


server/jetstream_cluster.go
```diff
@@ -2272,6 +2281,11 @@ func (js *jetStream) processUpdateStreamAssignment(sa *streamAssignment) {
 	sa.consumers = osa.consumers
 	sa.err = osa.err
 
+	// If we detect we are scaling down to 1, non-clustered, and we had a previous node, clear it here.
+	if sa.Config.Replicas == 1 && sa.Group.node != nil {
+		sa.Group.node = nil
```
kozlovic, line 2286: If this was a R>1 and now R1, did we stop monitoring of that stream? Is that done somewhere else?

server/jetstream_cluster_test.go
```diff
@@ -10674,6 +10674,23 @@ func TestJetStreamClusterStreamReplicaUpdateFunctionCheck(t *testing.T) {
 	}
 
 	checkSubsPending(t, sub, 3*numMsgs)
+
+	// Make sure cluster replicas are current.
+	for _, r := range si.Cluster.Replicas {
+		if !r.Current {
```
kozlovic, line 10680: Does that need to be protected with a "checkFor()"?

server/jetstream_cluster_test.go
```diff
@@ -10674,6 +10674,23 @@ func TestJetStreamClusterStreamReplicaUpdateFunctionCheck(t *testing.T) {
...
+			t.Fatalf("Expected replica to be current: %+v", r)
+		}
+	}
+
+	// Now check each indidvidual stream on each server to make sure replication occurred.
+	for _, s := range c.servers {
+		mset, err := s.GlobalAccount().lookupStream("TEST")
+		require_NoError(t, err)
+		state := mset.state()
+		if state.Msgs != uint64(3*numMsgs) || state.FirstSeq != 1 || state.LastSeq != 30 || state.Bytes != 1320 {
```
kozlovic, line 10690: Same thing, will that create a flapper? Should both checks be cover under a unique checkFor?

server/jetstream_cluster_test.go
derekcollison: Do you see this one flapping?

server/jetstream_cluster_test.go
derekcollison: Probably a good idea, will make that change.

server/jetstream_cluster_test.go
kozlovic: No, I did not. Was just wondering if the state (Current and number msgs in check below) are guaranteed to be up-to-date at this place in the test. If not, then it may cause flapping on Travis.

derekcollison: I could not make it fail after 50 runs with race and non-race but made the change anyway, good call by you.

server/jetstream_cluster_test.go
```diff
@@ -10677,19 +10677,30 @@ func TestJetStreamClusterStreamReplicaUpdateFunctionCheck(t *testing.T) {
 
 	// Make sure cluster replicas are current.
 	for _, r := range si.Cluster.Replicas {
-		if !r.Current {
-			t.Fatalf("Expected replica to be current: %+v", r)
-		}
+		checkFor(t, 5*time.Second, 100*time.Millisecond, func() error {
+			if !r.Current {
```
kozlovic, line 10681: Oh, that's for nothing, the state of si.Cluster.Replicas is not going to change, so r.Current would always be the same. It would have to include the StreamInfo request itself. Since I have not been able to have it flap either, you can remove for now.

kozlovic (changes requested):

server/jetstream_cluster_test.go
derekcollison: Fixed

server/jetstream_cluster_test.go
```diff
@@ -10676,20 +10676,33 @@ func TestJetStreamClusterStreamReplicaUpdateFunctionCheck(t *testing.T) {
 	checkSubsPending(t, sub, 3*numMsgs)
 
 	// Make sure cluster replicas are current.
-	for _, r := range si.Cluster.Replicas {
-		if !r.Current {
-			t.Fatalf("Expected replica to be current: %+v", r)
+	checkFor(t, 5*time.Second, 100*time.Millisecond, func() error {
+		si, err = js.StreamInfo("TEST")
+		require_NoError(t, err)
```
kozlovic, line 10681: You can't use that here... you would need: `if err != nil { return err }`...

kozlovic (changes requested):

server/jetstream_cluster_test.go
derekcollison: We should always be able to get stream info, so that is fatal to me for the test if we get an error.

kozlovic (approved):
LGTM

PR #2960, opened 2022-03-27
Additional improvements to memory pooling and management. 

derekcollison:
During contention to the head write block, the system could perform worse memory wise compared to simple go runtime and GC. Also had some references for the subject of messages being read back in to the underlying block which was bloating memory.

Additionally a fix for firstMatching() that did unnecessary work when matching all. This would cause excessive CPU and excessive load on the garbage collector.

/cc @nats-io/core

### DIFF
--- a/server/filestore.go
+++ b/server/filestore.go
@@ -881,15 +881,16 @@ func (mb *msgBlock) rebuildStateLocked() (*LostStreamData, error) {
 			mb.bytes += uint64(rl)
 
 			// Do per subject info.
-			if mb.fss != nil {
-				if subj := string(data[:slen]); len(subj) > 0 {
-					if ss := mb.fss[subj]; ss != nil {
-						ss.Msgs++
-						ss.Last = seq
-					} else {
-						subj = mb.subjString(subj)
-						mb.fss[subj] = &SimpleState{Msgs: 1, First: seq, Last: seq}
-					}
+			if slen > 0 && mb.fss != nil {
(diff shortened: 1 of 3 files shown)


server/filestore.go
```diff
@@ -2047,6 +2069,7 @@ func (fs *fileStore) removeMsg(seq uint64, secure, needFSLock bool) (bool, error
 	// we don't lose track of the first sequence.
 	if firstSeqNeedsUpdate {
 		fs.selectNextFirst()
+		// FIXME(dlc)?
```
kozlovic, line 2072: https://github.com/nats-io/nats-server/pull/2753

server/filestore.go
```diff
@@ -3617,10 +3642,38 @@ func msgFromBuf(buf []byte, sm *StoreMsg, hh hash.Hash64) (*StoreMsg, error) {
...
+			return mb.fs.cfg.Subjects[0]
+		} else {
+			for _, subj := range mb.fs.cfg.Subjects {
+				if key == subj {
+					return subj
+				}
+			}
+		}
+	}
+	// Copy here to not reference underlying buffer.
```
kozlovic, line 3671: I don't understand this... subjString() is invoked with a string that is already a string (either from caller stack or made a string out of the underlying buffer), so I don't see how "key" is referencing the underlying buffer.

kozlovic (approved):
LGTM

PR #2967, opened 2022-03-29
Adding server limits (max ack pending/dedupe window) to js config

matthiashanel:
Also shifting consumer config check to jsConsumerCreate as in clustered
mode this was enforced in the wrong place

This is the cherry picket commit from https://github.com/nats-io/nats-server/pull/2950

### DIFF
--- a/server/jetstream_api.go
+++ b/server/jetstream_api.go
@@ -1430,7 +1430,7 @@ func (s *Server) jsStreamUpdateRequest(sub *subscription, c *client, _ *Account,
 		return
 	}
 
-	cfg, err := checkStreamCfg(&ncfg)
+	cfg, err := checkStreamCfg(&ncfg, &s.getOpts().JetStreamLimits)
 	if err != nil {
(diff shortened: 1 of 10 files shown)


server/jetstream_api.go
```diff
@@ -3324,6 +3315,17 @@ func (s *Server) jsConsumerCreate(sub *subscription, c *client, a *Account, subj
 	}
 
 	if s.JetStreamIsClustered() && !req.Config.Direct {
```
kozlovic, line 3317: Had to check again that s.JetStreamIsClustered() cannot change its evaluation (js.cluster seem to never been reset to nil)... almost wonder if we should not "snapshot" that this is clustered the first time we have the `if s.JetStreamIsClustered()` above so we have the guarantee that `js` won't be nil here..

kozlovic (approved):
LGTM

server/jetstream_api.go
matthiashanel: changed

kozlovic (approved):
LGTM!

PR #2972, opened 2022-03-29
[CHANGED] JetStream: lower default consumer's maximum ack pending

kozlovic:
The default value is lowered from 20,000 to 1,000. This does not
seem to have a performance degradation impact, but may help
with scalability at scale.

### DIFF
--- a/server/consumer.go
+++ b/server/consumer.go
@@ -295,7 +295,7 @@ const (
 	// outstanding.
 	JsFlowControlMaxPending = 1 * 1024 * 1024
 	// JsDefaultMaxAckPending is set for consumers with explicit ack that do not set the max ack pending.
-	JsDefaultMaxAckPending = 20_000
+	JsDefaultMaxAckPending = 1000
 )
 
 // Helper function to set consumer config defaults from above.


server/consumer.go
```diff
@@ -295,7 +295,7 @@ const (
 	// outstanding.
 	JsFlowControlMaxPending = 1 * 1024 * 1024
 	// JsDefaultMaxAckPending is set for consumers with explicit ack that do not set the max ack pending.
-	JsDefaultMaxAckPending = 20_000
+	JsDefaultMaxAckPending = 1000
```
kozlovic, line 298: I decided to use 1,000 instead of 1,024 as we tend to do for any kind of limit, even when not bytes limit (well, this is a bad example since previous value was 20,000, not 20*1024). Let me know if you agree with this approach.

kozlovic:
@derekcollison We already have created an issue for client maintainers in the recent past to make sure that their library is not explicitly setting a value when not provided by the user. So libraries don't set a value and server sets to whatever default it choses. This is the issue I am referring to: https://github.com/nats-io/nats-architecture-and-design/issues/100.

PR #2975, opened 2022-03-30
[IMPROVED] Performance of OrderedConsumer with longer RTT.

derekcollison:
OrderedConsumer was very conservative with slow start and small max outstanding bytes.
This is increasing perf for longer rtt for real world scenarios.

/cc @nats-io/core


kozlovic (approved):
LGTM

derekcollison:
Just added a bump to above raft layer catch up as well, just fyi..

kozlovic (approved):
LGTM

PR #2976, opened 2022-03-30
[ADDED] Logs to healthcheck handler

samuel-form3:
Kubernetes probes don't use nor log the reponse body of health
endpoints. This means that for some reason a nats node running in
Kubernetes becomes on a Not Ready state we won't have a way to know why
other than to manually access the cluster and call the /healthz endpoint
manually and see the error.

This change adds an error log so we can observe what is going wrong with
a nats node that is not ready.

Signed-off-by: Samuel Torres <samuel.torres@form3.tech>

 - [ ] Tests added

Resolves #

 - Add error logging to healthcheck endpoint 

/cc @nats-io/core

### DIFF
--- a/server/monitor.go
+++ b/server/monitor.go
@@ -2768,6 +2768,7 @@ func (s *Server) HandleHealthz(w http.ResponseWriter, r *http.Request) {
 
 	hs := s.healthz()
 	if hs.Error != _EMPTY_ {
+		s.Warnf("Healthcheck failed: %q", hs.Error)
 		w.WriteHeader(http.StatusServiceUnavailable)
 	}
 	b, err := json.Marshal(hs)


server/monitor.go
```diff
@@ -2768,6 +2768,7 @@ func (s *Server) HandleHealthz(w http.ResponseWriter, r *http.Request) {
 
 	hs := s.healthz()
 	if hs.Error != _EMPTY_ {
+		s.Errorf("Healthcheck failed: %q", hs.Error)
```
derekcollison, line 2771: Might suggest this be a Warnf since it can and should clear.

samuel-form3: :+1:

derekcollison (commented):
In general LGTM, but want Ivan to weigh in as well.

kozlovic (approved):
LGTM. No objection to this change.

PR #2982, opened 2022-03-31
added max_ack_pending setting to js account limits

matthiashanel:
because of the addition, defaults now have to be set later (depend on
these new limits now)

also re-organized the code to closer track how stream create looks

 - [ ] Tests added

Resolves #

 -
 -
 -

/cc @nats-io/core

### DIFF
--- a/server/consumer.go
+++ b/server/consumer.go
@@ -333,7 +333,7 @@ func (mset *stream) addConsumer(config *ConsumerConfig) (*consumer, error) {
 	return mset.addConsumerWithAssignment(config, _EMPTY_, nil)
 }
 
-func checkConsumerCfg(config *ConsumerConfig, lim *JSLimitOpts, cfg *StreamConfig, acc *Account, limit *JetStreamAccountLimits) *ApiError {
+func checkConsumerCfg(config *ConsumerConfig, srvLim *JSLimitOpts, cfg *StreamConfig, acc *Account, accLim *JetStreamAccountLimits) *ApiError {
 	// Check if we have a BackOff defined that MaxDeliver is within range etc.
 	if lbo := len(config.BackOff); lbo > 0 && config.MaxDeliver <= lbo {
 		return NewJSConsumerMaxDeliverBackoffError()
@@ -387,8 +387,8 @@ func checkConsumerCfg(config *ConsumerConfig, lim *JSLimitOpts, cfg *StreamConfi
 			return NewJSConsumerMaxRequestExpiresToSmallError()
 		}
 	}
-	if lim.MaxAckPending > 0 && config.MaxAckPending > lim.MaxAckPending ||
-		limit.MaxAckPending > 0 && config.MaxAckPending > limit.MaxAckPending {
+	if srvLim.MaxAckPending > 0 && config.MaxAckPending > srvLim.MaxAckPending ||
+		accLim.MaxAckPending > 0 && config.MaxAckPending > accLim.MaxAckPending {
 		return NewJSConsumerMaxPendingAckExcessError()
 	}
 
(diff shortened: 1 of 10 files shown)


server/jetstream.go
```diff
@@ -63,6 +63,7 @@ type JetStreamAccountLimits struct {
 	MaxStore             int64 `json:"max_storage"`
 	MaxStreams           int   `json:"max_streams"`
 	MaxConsumers         int   `json:"max_consumers"`
+	MaxAckPending        int   `json:"mack_ack_pending"`
```
wallyqs, line 66:
```suggestion
	MaxAckPending        int   `json:"max_ack_pending"`
```

matthiashanel: good catch, thanks

server/consumer.go
```diff
@@ -384,7 +387,8 @@ func checkConsumerCfg(config *ConsumerConfig, lim *JSLimitOpts, cfg *StreamConfi
 			return NewJSConsumerMaxRequestExpiresToSmallError()
 		}
 	}
-	if lim.MaxAckPending > 0 && config.MaxAckPending > lim.MaxAckPending {
+	if lim.MaxAckPending > 0 && config.MaxAckPending > lim.MaxAckPending ||
```
kozlovic, line 390: Is it that `lim` are server limits and `limit` are account limits? If that's the case, I wonder if `srvLim` and `accLim` (or something similar) would make it easier to understand.

server/jetstream_cluster.go
```diff
@@ -3968,25 +3968,34 @@ func (js *jetStream) createGroupForStream(ci *ClientInfo, cfg *StreamConfig) *ra
...
-		s.sendAPIErrResponse(ci, acc, subject, reply, string(rmsg), s.jsonResponse(&resp))
+		return nil, "", nil, NewJSNotEnabledForAccountError()
+	}
+
+	jsa.mu.RLock()
+	selectedLimits, tierName, ok := jsa.selectLimits(cfg)
+	jsa.mu.RUnlock()
+
+	if !ok {
+		return nil, "", nil, NewJSNoLimitsError()
```
kozlovic, line 3986: Someone would say `_EMPTY_` here ;-)

kozlovic (approved):
LGTM, but small comment about variables naming..

server/consumer.go
matthiashanel: yes, fixed

server/jetstream_cluster.go
matthiashanel: fixed

kozlovic (approved):
LGTM

server/opts.go
```diff
@@ -1659,7 +1659,7 @@ func parseGateway(v interface{}, o *Options, errors *[]error, warnings *[]error)
 	return nil
 }
 
-var dynamicJSAccountLimits = JetStreamAccountLimits{-1, -1, -1, -1, 0, 0, false}
+var dynamicJSAccountLimits = JetStreamAccountLimits{-1, -1, -1, -1, 0, 0, 0, false}
```
ripienaar, line 1662: why does `max_memory` a byte limit default to `-1` while the `memory_max_stream_bytes` also a byte limit default to 0? 

Make using these quite challenging and easy to make mistakes

matthiashanel: this will work with -1 as well. however the value 0 does not mean it's disabled, just means not limited. 
If we had picked -1, then deployed jwt would have started to fail.

matthiashanel: @ripienaar I'll make sure they return `-1` to the client and make that independent of jwt. sounds good?

PR #2983, opened 2022-03-31
skips jsz on non js machines when leader only requested

ripienaar:
This is a regression introduced in 055703f4fabc46eb9722193f9ecbed8803795228
that leads to panics in management tooling

/cc @nats-io/core


derekcollison (approved):
LGTM

PR #2985, opened 2022-03-31
In Merge validate nkey and subsequent saveIfNewer error on invalid jwt

matthiashanel:

 - [ ] Tests added

Resolves #

 -
 -
 -

/cc @nats-io/core

### DIFF
--- a/server/dirstore.go
+++ b/server/dirstore.go
@@ -28,6 +28,8 @@ import (
 	"sync"
 	"time"
 
+	"github.com/nats-io/nkeys"
+
 	"github.com/nats-io/jwt/v2" // only used to decode, not for storage
 )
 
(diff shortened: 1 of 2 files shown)


philpennock (commented):
Is dirstore not compatible with non-JWT users?  Can't people use a cluster of servers configured with static users, and JS?  I don't think this is a safe change.

matthiashanel:
> Can't people use a cluster of servers configured with static users, and JS? I don't think this is a safe change.

If they use a cluster of servers with static users and JS they won't be using this. 
this is exclusively used as backing storage by nats account resolver.
This was even done before JS went clustreed.

kozlovic (approved):
LGTM

philpennock:
Oh sorry, dirstore.go vs filestore.go ... my head went in completely the wrong direction.

server/dirstore.go
```diff
@@ -321,6 +323,9 @@ func (store *DirJWTStore) Merge(pack string) error {
 			return fmt.Errorf("line in package didn't contain 2 entries: %q", line)
 		}
 		pubKey := split[0]
+		if !nkeys.IsValidPublicAccountKey(pubKey) {
+			return fmt.Errorf("key to merge is not a valid public account key")
```
philpennock, line 327: without a %param this should probably be `errors.New()` instead, but %q auto-quoting of the key might be worthwhile.

philpennock (approved):
Double-check with Alberto please: there are I think some assumptions about propagation of the JWTs across JWT version boundaries, and this change seems to change the semantics of how a future v3 JWT could be introduced, with every nats-server in a cluster having to be recompiled with a jwt library which accepts the format, before the JWTs can be propagated.

PR #2986, opened 2022-03-31
Fixed wrong error check

matthiashanel:

Noticed the failure when I removed the `t.Skip()` in my test :-(


kozlovic (approved):
LGTM

PR #2987, opened 2022-03-31
Updated snapshot default chunk size and window size

kozlovic:
Down from 256KB to 128KB for chunk and 32MB to 8MB.


derekcollison (approved):
LGTM

PR #2988, opened 2022-03-31
start using unit test TestJWTClusteredJetStreamTiers, size was off

matthiashanel:
If both servers sent a remote update of their local use,
the limit was hit. But that limit wass to small by 200


kozlovic (approved):
LGTM

PR #2989, opened 2022-03-31
Skip SystemLimitsPlacement if we can't get the desired leader

variadico:
The system limit placement tests can fail if we can't get a specific server to be the leader. There is currently no way to force a specific server to be the leader, so we have to rely on chance for this to happen.

The main change here is that if we can't pick the right leader, then we'll skip the test instead of fail the test. I also added some more checks to `TestJetStreamSystemLimitsPlacement`.

I can't use the current JetStream cluster helpers because those assume that every server in the cluster is identical. However, the point of these tests is to verify the behavior when servers are configured differently. There may be a way to start the cluster and update individual servers after the fact, based on [this test](https://github.com/nats-io/nats-server/blob/ee98a814694820d9eeb9a2646a9aa5ee1ea346d0/server/jetstream_cluster_test.go#L10793). I could investigate this further.

For the current change, I ran `go test -v -count=100 -run=SystemLimitsPlacement &>> test.log` in 5 different shell invocations.

Here are some results from the logs.
```
$ grep -E "^--- PASS: TestJetStreamSuperClusterSystemLimitsPlacement" test.log | wc -l
500

$ grep -E "^--- PASS: TestJetStreamSystemLimitsPlacement" test.log | wc -l
475
$ grep -E "^--- SKIP: TestJetStreamSystemLimitsPlacement" test.log | wc -l
25

$ grep "FAIL" test.log | wc -l
0
```
5% of the time, `TestJetStreamSystemLimitsPlacement` wasn't able to pick the desired server to be leader. In those cases, we skip that test.

Besides those skipped tests, there were no failures in the 500 runs.

### DIFF
--- a/server/jetstream_test.go
+++ b/server/jetstream_test.go
@@ -6810,23 +6810,28 @@ func TestJetStreamSystemLimitsPlacement(t *testing.T) {
 	const mediumSystemLimit = smallSystemLimit * 2
 	const largeSystemLimit = smallSystemLimit * 3
 
+	startClusterPort := getStartClusterPort()
+
+	require_NoError(t, os.MkdirAll(tempRoot, 0755))
 	getServer := func(t *testing.T, serverName string) *Server {
 		storeDir, err := os.MkdirTemp(tempRoot, "jstests-storedir-")
 		require_NoError(t, err)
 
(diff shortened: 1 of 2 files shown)


kozlovic (approved):
LGTM

PR #2991, opened 2022-04-01
[FIXED] Sealed streams would not recover on server restart.

derekcollison:
Not allowing sealed streams to be created meant we could not recover on server restart.

/cc @nats-io/core


ripienaar (approved):
LGTM

PR #2992, opened 2022-04-01
error handling and info when moving a stream from non existing tier

matthiashanel:
adds unit test to test this scenario
improves reporting of correct error
only show info for non existing tiers where streams exist


kozlovic (approved):
LGTM

PR #2994, opened 2022-04-01
[CHANGED] Rate limit (some) similar warnings

kozlovic:
Some warnings, especially when dealing with JS limits that were
printed on a per-message basis, are now limited to ~1 per second
if the content of the warning is already found in a map.

This is also for "client" warnings, but the client porting of the
warning is not taken into account so that helps with reducing logging
for similar content, but coming from different clients.

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -274,7 +274,7 @@ type Server struct {
 
 	// To limit logging frequency
 	rateLimitLogging   sync.Map
-	rateLimitLoggingCh chan int64
+	rateLimitLoggingCh chan time.Duration
 }
 
 // For tracking JS nodes.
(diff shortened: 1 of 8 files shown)


server/server.go
```diff
@@ -3593,3 +3599,42 @@ func (s *Server) updateRemoteSubscription(acc *Account, sub *subscription, delta
...
+
+	s.startGoRoutine(func() {
+		defer s.grWG.Done()
+
+		ticker := time.NewTicker(time.Second)
+		for {
+			select {
+			case <-s.quitCh:
+				return
+			case interval = <-s.rateLimitLoggingCh:
```
derekcollison, line 3617: You lock above to set but not here to access.

server/server.go
```diff
@@ -3593,3 +3599,42 @@ func (s *Server) updateRemoteSubscription(acc *Account, sub *subscription, delta
 
 	s.updateLeafNodes(acc, sub, delta)
 }
+
+func (s *Server) startRateLimitLogExpiration() {
+	s.mu.Lock()
+	interval := int64(time.Second)
```
derekcollison, line 3605: You set here under lock, but seems not needed. And you never seem to access or use this value, you just overwrite it below when reading from the channel on line 3617.

derekcollison: I do see it could be used on line 3622, but could then this just be a variable defined in the Go routine, near where you define the ticker?

server/server.go
```diff
@@ -3593,3 +3599,42 @@ func (s *Server) updateRemoteSubscription(acc *Account, sub *subscription, delta
...
+	s.rateLimitLoggingCh = make(chan int64, 1)
```
derekcollison, line 3606: Why not just make a chan on time.Duration to avoid casts?

derekcollison (commented):
This looks pretty slick! Some comments.

server/server.go
kozlovic: The lock above is to actually create the channel to change the interval. Note that so far it will just be used for tests. So here, the method below (startRateLimitLogExpiration()) will send to the channel the new interval. No lock needed.

server/server.go
kozlovic: If I do that, then the cast will have to be done when checking if it should be expired:
```
if start := v.(int64); now-start >= interval {
```
would have to be `int64(interval)`. I originally had actual duration and then do `time.Since(start)` (by setting in the map time.Now() instead of time.Now().UnixNano(), but what I did not like is the idea that time.Since() will then evaluate the "time now" at every call?  Maybe that's not a big deal.. what do you think?

server/server.go
kozlovic: Again, the interval setting here is not the thing I want the lock for (could move out of the lock). It is the channel.. That being said, I think that the channel should be created in &Server{} when creating NewServer instead.. will revisit that.

kozlovic:
@derekcollison Made some changes that move the initialization of the channel when creating the server and remove casting by making it a channel of time.Duration. Also changed the store value to time.Time instead of int64.

derekcollison (approved):
LGTM

PR #2997, opened 2022-04-04
[FIXED] JetStream: send "bad request" response for malformed API requ…

### LINKED ISSUE #2995: When providing an invalid consumer name, consumer info returns unclear error of type Context Deadline
## Defect

- nats-server`v2.7.4`
- nats-client v1.13.1-0.20220308171302-2f2f6968e98d

- Ubuntu 20.04.4 LTS

##### MCVE test case

You can also remove nats-server startup, and simply connect it to the running nats-server by changing the `nats.Connect` call.

```golang
package main

import (
	"context"
	"testing"
	"time"

	"github.com/nats-io/nats-server/v2/server"
	"github.com/nats-io/nats.go"
	"github.com/stretchr/testify/require"
)

func TestConsumerInfoIncorrectDurableName(t *testing.T) {
	opts := &server.Options{
		JetStream: true,
		Host:      "127.0.0.1",
		Port:      -1,
		HTTPPort:  -1,
		NoLog:     true,
		NoSigs:    true,
		Debug:     true,
		Trace:     true,
		StoreDir:  t.TempDir(),
	}
	s, err := server.NewServer(opts)
	require.NoError(t, err)

	go s.Start()
	require.True(t, s.ReadyForConnections(20*time.Second))

	c, err := nats.Connect(s.ClientURL())
	require.NoError(t, err)

	jsCtx, err := c.JetStream(nats.MaxWait(30 * time.Second))
	require.NoError(t, err)

	streamName := "some-stream"
	streamCfg := &nats.StreamConfig{
		Name:     streamName,
		Subjects: []string{"subject.>"},
		Storage:  nats.MemoryStorage,
		MaxAge:   time.Hour,
	}
	_, err = jsCtx.AddStream(streamCfg)
	require.NoError(t, err)

	consumerCfg := &nats.ConsumerConfig{
		AckPolicy:     nats.AckExplicitPolicy,
		DeliverPolicy: nats.DeliverAllPolicy,
		ReplayPolicy:  nats.ReplayInstantPolicy,
		Durable:       "durable-name",
	}

	_, err = jsCtx.AddConsumer(streamName, consumerCfg)
	require.NoError(t, err)

	invalidName := "foo.bar"
	_, err = jsCtx.ConsumerInfo(streamName, invalidName)
	require.ErrorIs(t, err, context.DeadlineExceeded)
}
```

#### Expected result:

Expected to receive an error which indicates that the consumer name is invalid

#### Actual result:

Received deadline exceeded

## Investigation

I might be wrong on details, I'm not a maintainer of `nats-server`, just started using the product.

The [API-handlers](https://github.com/nats-io/nats-server/blob/main/server/jetstream_api.go#L734-L763) are by the looks of it essentially subscriptions. For the API call `JSApiConsumerInfo` the [subject](https://github.com/nats-io/nats-server/blob/main/server/jetstream_api.go#L141) is **`$JS.API.CONSUMER.INFO.*.*`**

Given the invalid consumer name of `foo.bar`, which would expand into the subject name `$JS.API.CONSUMER.INFO.some-stream.foo.bar`, it will not match the expected subject and therefore not find it's way to the appropriate handler `s.jsConsumerInfoRequest`.
Instead, it ends up in [apiDispatch](https://github.com/nats-io/nats-server/blob/main/server/jetstream_api.go#L652). Where the match function returns an empty set of SublistResult and the [request is dropped](https://github.com/nats-io/nats-server/blob/main/server/jetstream_api.go#L666), leaving the client hanging.

The found tokens are:
```
tokens = {[]string} len:6, cap:32
 0 = {string} "$JS"
 1 = {string} "API"
 2 = {string} "CONSUMER"
 3 = {string} "INFO"
 4 = {string} "some-stream"
 5 = {string} "foo"
 ```
 
Perhaps validation could be done on `subject` in the server's end of the client connectio in [processMsgResult](https://github.com/nats-io/nats-server/blob/18bdabff35cf04ea1d947f01c100474f7f0d69c5/server/client.go#L4067) to make sure `subject` conforms to the required structure?

Maybe there are somewhere better to do this.

It was quite confusing when the nats client suddenly started returning context deadline exceeded without any clear suggestion to it's the input that's wrong.

kozlovic:
…ests

An example was a "consumer info" request with a consumer name that
had tokens, which is illegal. This results in the request being
dropped in apiDispatch() because there was no interest.
The server will now return a "bad request" error in such case.

Resolves #2995

### DIFF
--- a/server/jetstream_api.go
+++ b/server/jetstream_api.go
@@ -664,6 +664,11 @@ func (js *jetStream) apiDispatch(sub *subscription, c *client, acc *Account, sub
 
 	// Shortcircuit.
 	if len(rr.psubs)+len(rr.qsubs) == 0 {
+		ci, acc, _, msg, err := s.getRequestInfo(c, rmsg)
+		if err == nil {
+			resp := &ApiResponse{Type: JSApiOverloadedType, Error: NewJSBadRequestError()}
+			s.sendAPIErrResponse(ci, acc, subject, reply, string(msg), s.jsonResponse(&resp))
+		}
 		return
 	}
 
--- a/server/jetstream_cluster_test.go
+++ b/server/jetstream_cluster_test.go
@@ -11659,6 +11659,35 @@ func TestJetStreamClusterMemoryConsumerCompactVsSnapshot(t *testing.T) {
 
 }
 
+func TestJetStreamClusterSendBadRequestResponseOnInvalidAPIRequest(t *testing.T) {
+	s := RunBasicJetStreamServer()
+	if config := s.JetStreamConfig(); config != nil {
+		defer removeDir(t, config.StoreDir)
+	}
+	defer s.Shutdown()
+
+	c := createJetStreamClusterExplicit(t, "JSC", 3)
+	defer c.shutdown()
+
+	checkBadRequest := func(t *testing.T, s *Server) {
+		nc := natsConnect(t, s.ClientURL())
+		defer nc.Close()
+
+		// Send a consumer info but with a bad consumer name.
+		// Use low-level here to by-bass validity checks done in the client library.
+		msg, err := nc.Request(fmt.Sprintf(JSApiConsumerInfoT, "BAD_REQUESTS", "bad.consumer.name"), nil, time.Second)
+		require_NoError(t, err)
+
+		var resp JSApiConsumerInfoResponse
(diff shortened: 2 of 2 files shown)


derekcollison (approved):
LGTM - Thanks!

PR #3001, opened 2022-04-04
Allow streams and consumers to migrate between clusters.

derekcollison:
The system will allow an update to a stream, and subsequently all attached consumers, to be placed in another cluster either directly or via tag placement. The meta layer will scale the underlying peerset appropriately to straddle the two clusters for both the stream and consumers, taking into account the consumer type. Control will then pass to the current leaders of the assets who will monitor the catch up status of the new peers (Note we can optimize this later to only traverse once across a GW for any given asset, but for now this is simpler). Once the original leaders have determined the assets are synched it will pass leadership to a member of the new peerset. Once the new leader has been elected, it will forward a request for the meta layer to shrink the peerset by removing the old peers.

/cc @nats-io/core


derekcollison:
FYI there may be two follow on PRs to have a time bound on the migration monitoring, and a test to check that we are not adversely affected by account limits here, meaning you can pop your usage for a short time over the account limits and not be blocked to move a stream.

server/events.go
```diff
@@ -691,6 +691,7 @@ func (s *Server) sendStatsz(subj string) {
 			ni := v.(nodeInfo)
 			ni.stats = jStat.Stats
 			ni.cfg = jStat.Config
+			ni.tags = s.getOpts().Tags
```
kozlovic, line 694: I think it would be better to make copies of the options's tags value here. Since Tags is a type but an alias to []string, you could use `copyStrings(s.getOpts().Tags)`

server/jetstream_api.go
```diff
@@ -1468,8 +1468,15 @@ func (s *Server) jsStreamUpdateRequest(sub *subscription, c *client, _ *Account,
 		return
 	}
 
+	// Handle clustered version here.
 	if s.JetStreamIsClustered() {
-		s.jsClusteredStreamUpdateRequest(ci, acc, subject, reply, rmsg, &cfg)
+		// If we are inline with client, we still may need to do a callout for stream info
+		// during this call, so place in Go routine to not block client.
+		if c.kind != ROUTER && c.kind != GATEWAY {
```
kozlovic, line 1475: The comment says "if we are inline with client", which I think you mean that the request comes from a CLIENT connection. If so, why not check on `if c.kind == CLIENT`? I mean a "lot" of types are not a CLIENT. We also have ACCOUNT, and LEAF, etc.. so would rather be explicit on what type of CLIENT should do what.

server/jetstream_cluster.go
```diff
@@ -1621,9 +1677,61 @@ func (js *jetStream) monitorStream(mset *stream, sa *streamAssignment, sendSnaps
...
+
+			// Check for migrations here. We set the state on the stream assignment update below.
+			if isLeader && migrating {
+				if peerGroup == oldPeerGroup {
+					startMigrationMonitoring()
+				} else {
+					stopMigrationMonitoring()
+					// We delay here since consumer migration, while in most cases is faster, is done after the stream itself.
+					// When we remove the stream from the old peers this will remove the consumers which might still be in flight.
+					time.AfterFunc(1000*time.Millisecond, func() { js.removeOldPeers(mset) })
```
kozlovic, line 1691: Not sure I understand the comment, but that sounds a bit scary: do we have to be sure that consumer migration is done when removeOldPeers is invoked (in 1sec)? You can't assume how long things are going to take.

Other concern: since this is happening based on an anonymous timer (that can't be stopped), is there a risk that this code here executed (scheduling the peers removal) followed within 1sec by a start migration? When the timer's function finally executes, is that not going to mess up the state of the newly started migration?

server/jetstream_cluster.go
```diff
@@ -1719,6 +1829,63 @@ func (js *jetStream) monitorStream(mset *stream, sa *streamAssignment, sendSnaps
...
+	oldPeerGroup = peerMigrateType(iota)
+	newPeerGroup
+)
+
+// Determine if we are migrating and if so if we are part of the old or new set.
+func (mset *stream) isMigrating() (bool, peerMigrateType) {
+	sa := mset.streamAssignment()
+	// During migration we will always be R>1, even when we start R1.
+	// So if we do not have a group or node we no we are not migrating.
+	if sa == nil || sa.Group == nil || sa.Group.node == nil {
```
kozlovic, line 1847: Not sure about the stream assignment itself, but I think I have seen race reports regarding those. Do we need RLock/defer RUnlock?

server/jetstream_cluster.go
```diff
@@ -1719,6 +1829,63 @@ func (js *jetStream) monitorStream(mset *stream, sa *streamAssignment, sendSnaps
...
+	}
+	return true, oldPeerGroup
+}
+
+// Determine if we are migrating and if so if we are part of the old or new set.
+func (o *consumer) isMigrating() (bool, peerMigrateType) {
+	ca := o.consumerAssignment()
+	// During migration we will always be R>1, even when we start R1.
+	// So if we do not have a group or node we no we are not migrating.
+	if ca == nil || ca.Group == nil || ca.Group.node == nil {
```
kozlovic, line 1867: Same question regarding locking here, except that you would have to be more precise because below you are getting stream's lock.

kozlovic (commented):
Some comments about locking, etc...

server/events.go
derekcollison: Will do.

server/jetstream_api.go
derekcollison: For networked accounts I want to make sure we do not block client or leaf. We already do the Go routine dispatch for Routes and Gateways, so wanted to be consistent.

server/jetstream_cluster.go
kozlovic: Actually, I think I am right since running your PR locally, got this race the very first time:
```
=== RUN   TestJetStreamClusterStreamAndConsumerScaleUpAndDown
==================
WARNING: DATA RACE
Read at 0x00c00104edd0 by goroutine 113:
  github.com/nats-io/nats-server/v2/server.(*consumer).isMigrating()
      /Users/ivan/dev/go/src/github.com/nats-io/nats-server/server/jetstream_cluster.go:1867 +0x8f
  github.com/nats-io/nats-server/v2/server.(*jetStream).monitorConsumer()
      /Users/ivan/dev/go/src/github.com/nats-io/nats-server/server/jetstream_cluster.go:3413 +0xd11
  github.com/nats-io/nats-server/v2/server.(*jetStream).processClusterCreateConsumer.func1()
      /Users/ivan/dev/go/src/github.com/nats-io/nats-server/server/jetstream_cluster.go:3140 +0x4c

Previous write at 0x00c00104edd0 by goroutine 248:
  github.com/nats-io/nats-server/v2/server.(*jetStream).processClusterCreateConsumer()
      /Users/ivan/dev/go/src/github.com/nats-io/nats-server/server/jetstream_cluster.go:3124 +0xbc4
  github.com/nats-io/nats-server/v2/server.(*jetStream).processConsumerAssignment()
      /Users/ivan/dev/go/src/github.com/nats-io/nats-server/server/jetstream_cluster.go:2931 +0xb93
  github.com/nats-io/nats-server/v2/server.(*jetStream).applyMetaEntries()
      /Users/ivan/dev/go/src/github.com/nats-io/nats-server/server/jetstream_cluster.go:1361 +0x384
  github.com/nats-io/nats-server/v2/server.(*jetStream).monitorCluster()
      /Users/ivan/dev/go/src/github.com/nats-io/nats-server/server/jetstream_cluster.go:836 +0xb2d
  github.com/nats-io/nats-server/v2/server.(*jetStream).monitorCluster-fm()
      /Users/ivan/dev/go/src/github.com/nats-io/nats-server/server/jetstream_cluster.go:776 +0x39

Goroutine 113 (running) created at:
  github.com/nats-io/nats-server/v2/server.(*Server).startGoRoutine()
      /Users/ivan/dev/go/src/github.com/nats-io/nats-server/server/server.go:2995 +0x86
  github.com/nats-io/nats-server/v2/server.(*jetStream).processClusterCreateConsumer()
      /Users/ivan/dev/go/src/github.com/nats-io/nats-server/server/jetstream_cluster.go:3140 +0xdad
  github.com/nats-io/nats-server/v2/server.(*jetStream).processConsumerAssignment()
      /Users/ivan/dev/go/src/github.com/nats-io/nats-server/server/jetstream_cluster.go:2931 +0xb93
  github.com/nats-io/nats-server/v2/server.(*jetStream).applyMetaEntries()
      /Users/ivan/dev/go/src/github.com/nats-io/nats-server/server/jetstream_cluster.go:1361 +0x384
  github.com/nats-io/nats-server/v2/server.(*jetStream).monitorCluster()
      /Users/ivan/dev/go/src/github.com/nats-io/nats-server/server/jetstream_cluster.go:836 +0xb2d
  github.com/nats-io/nats-server/v2/server.(*jetStream).monitorCluster-fm()
      /Users/ivan/dev/go/src/github.com/nats-io/nats-server/server/jetstream_cluster.go:776 +0x39

Goroutine 248 (running) created at:
  github.com/nats-io/nats-server/v2/server.(*Server).startGoRoutine()
      /Users/ivan/dev/go/src/github.com/nats-io/nats-server/server/server.go:2995 +0x86
  github.com/nats-io/nats-server/v2/server.(*jetStream).setupMetaGroup()
      /Users/ivan/dev/go/src/github.com/nats-io/nats-server/server/jetstream_cluster.go:584 +0x1092
  github.com/nats-io/nats-server/v2/server.(*Server).enableJetStreamClustering()
      /Users/ivan/dev/go/src/github.com/nats-io/nats-server/server/jetstream_cluster.go:486 +0x20a
  github.com/nats-io/nats-server/v2/server.(*Server).enableJetStream()
      /Users/ivan/dev/go/src/github.com/nats-io/nats-server/server/jetstream.go:398 +0x1168
  github.com/nats-io/nats-server/v2/server.(*Server).EnableJetStream()
      /Users/ivan/dev/go/src/github.com/nats-io/nats-server/server/jetstream.go:204 +0x651
  github.com/nats-io/nats-server/v2/server.(*Server).Start()
      /Users/ivan/dev/go/src/github.com/nats-io/nats-server/server/server.go:1728 +0x1804
  github.com/nats-io/nats-server/v2/server.RunServer·dwrap·4178()
      /Users/ivan/dev/go/src/github.com/nats-io/nats-server/server/server_test.go:90 +0x39
==================
    testing.go:1152: race detected during execution of test
```

server/jetstream_cluster.go
derekcollison: I meant to lower it to 500ms, which I will do. 

This is when we have moved control to the new stream leader in the new peerset. It is responsible to shrink the peerset, but consumers have to be expanded behind the stream itself since the consumer needs the stream state on any peer it occupies.

In some of my testing I saw this drastically slow down the consumer checks since the stream removed the old consumers and in some instances the old leader before it could transfer leadership to the new peerset etc.

server/jetstream_cluster.go
derekcollison: If the race detector complained we would need to look at locking js.

server/jetstream_cluster.go
derekcollison: Same as above, but have not seen any race complaints about this access. Have fixed (or tried to fix) the ones I have seen. The expansion of the update stream popped a few.

derekcollison: Will run a few locally myself and take a look. But its not the lock you suspect.

derekcollison: ok see it locally as well.

kozlovic: Yes, I don't know if it needs js or something else, but I can get this race quite easily.

kozlovic: Maybe you need to make those isMigrating() a function of jetStream and pass a stream and consumer (2 different funcs) and inside you need to have js lock.. but you know better..

server/jetstream_cluster.go
derekcollison: Fixed.

server/jetstream_cluster.go
derekcollison: Fixed.

server/jetstream_cluster.go
```diff
@@ -4172,8 +4434,79 @@ func (s *Server) jsClusteredStreamUpdateRequest(ci *ClientInfo, acc *Account, su
...
 		} else {
-			// We are deleting nodes here.
-			rg.Peers = rg.Peers[:newCfg.Replicas]
+			// We are deleting nodes here. We want to do our best to preserve the current leader.
+			// We have support now from above that guarantees we are in our own Go routine, so can
+			// ask for stream info from the stream leader to make sure we keep the leader in the new list.
+			var curLeader string
+			if !s.allPeersOffline(rg) {
+				// Need to release js lock.
+				js.mu.Unlock()
```
kozlovic, line 4443: Not a big fan, as you know, but I understand why you do it here. Maybe we should have had a specific lock for the s.sys.replies so that we could do it just for the registration of those handlers, without requiring server lock, but ok.

server/jetstream_cluster.go
```diff
@@ -4172,8 +4434,79 @@ func (s *Server) jsClusteredStreamUpdateRequest(ci *ClientInfo, acc *Account, su
...
+						s.Warnf("Error unmarshaling clustered stream info response:%v", err)
+						return
+					}
+					select {
+					case results <- &si:
+					default:
+						s.Warnf("Failed placing remote stream info result on internal channel")
+					}
+					// Only expect one so clean up here.
+					s.mu.Lock()
```
kozlovic, line 4460: What if there is no response received? We are going to "leak" the inbox in s.sys.replies, no? That is ok if that does not happen often, but I have no idea if that is the case. So should this code be moved out and done at the end of this function (either because the reply was received or we timeout)?

server/jetstream_cluster.go
kozlovic: Actually, you don't need to unlock. I created a file called `locksordering.txt` where I list the known ordering, and it says:
```
jetStream -> jsAccount -> Server -> client-> Account
```
So it is ok to get server lock inside js lock.

derekcollison: ok great will make that change.

derekcollison: Check that, this calls out to other servers and we do not want to hold any locks across that, so will not make that change. But good to know.

kozlovic: Not while we wait, I guess. But then you could do the cleanup of replies outside of the msgHandler, or at least after the select in case we did not receive a msg.

server/jetstream_cluster.go
derekcollison: Agree. Fixed.

server/jetstream_cluster.go
derekcollison: Yep good call.

server/jetstream_cluster.go
derekcollison: I lowered, hopefully the explanation makes sense.

server/jetstream_cluster.go
```diff
@@ -4172,8 +4434,79 @@ func (s *Server) jsClusteredStreamUpdateRequest(ci *ClientInfo, acc *Account, su
...
+					s.Warnf("Did not receive stream info results for '%s > %s'", acc, cfg.Name)
+				case si := <-results:
+					if si.Cluster != nil {
+						// The leader here is the server name, but need to convert to internal name.
+						curLeader = string(getHash(si.Cluster.Leader))
+					}
+				}
+				// Clean up here.
+				s.mu.Lock()
+				if s.sys != nil && s.sys.replies != nil {
```
kozlovic, line 4482: We use this code in other places, so likely a copy-paste, but then that means, should we check for s.sys and s.sys.replies not being nil before trying to register the handler above? The `sys.replies` (and maybe s.sys) is set to nil on shutdown. So if there is a panic it would be on shutdown, so less critical.

kozlovic (approved):
LGTM, but small comment about registering inbox without checking s.sys.replies

server/jetstream_cluster.go
derekcollison: It always exists these days except as you point out on shutdown. We could check for it, this eventually (internal messaging) needs to be reworked of course.

ripienaar:
Did some basic testing, moving something from NYC -> LON -> SFO -> LON by the time we got back to LON it got stuck with all members (sfo and lon) and never recovered and messages lost.

Will test again after the follow up PRs, but I think I'd definitely like some more logging for start/end etc, and we might want events

derekcollison:
We knew a feature like this would have some bumps in between on the road to GA, so this is to be expected.

Was the stream active in terms of inbound messages?
How many consumers were attached and of what type and activity level?

ripienaar:
No consumers, no inbound messages - so mostly idle.

It appeared that the LON->SFO completed - all peers were in SFO and messages was there, but it could have been in progress still

PR #3007, opened 2022-04-05
unit test to make sure tiered limits and stream moves work together

matthiashanel:
This needs testing because stream move adjusts the replication factor

    Because adjusting replication factor and moving is illegal, this case
    does not need to be tested

    In order to support one off configurations, added same modification
    callout to super cluster as is used with cluster


kozlovic (approved):
LGTM

PR #3008, opened 2022-04-06
[IMPROVED] Stream with multiple subjects and consumer with filter

kozlovic:
This is more of a regression introduced in v2.7.3 (with PR #2848).
When the store has a list of subjects, finding the next message
to deliver would go through the subjects map and have to match
to find out if it is a subset (if the filter had a wildcard).
In situations where there were lots of subjects (for instance 1
message per subject), but the consumer did not filter on anything
specific, then this processing was becoming slow.

We now check that if the stream has a single subject (even with
wildcard) and the consumer filters on that exact subject, then
we can do a linear scan. We also do a linear scan if the number
of messages in the block is 1/2 the number of subjects in the
subjects map.

### DIFF
--- a/server/filestore.go
+++ b/server/filestore.go
@@ -121,6 +121,7 @@ type msgBlock struct {
 	rbytes  uint64 // Total bytes (raw) including deleted. Used for rolling to new blk.
 	msgs    uint64 // User visible message count.
 	fss     map[string]*SimpleState
+	sfilter string // Single subject filter
 	sfn     string
 	kfn     string
 	lwits   int64
@@ -546,8 +547,12 @@ const (
 // This is the max room needed for index header.
 const indexHdrSize = 7*binary.MaxVarintLen64 + hdrLen + checksumSize
 
+// Lock held on entry
 func (fs *fileStore) recoverMsgBlock(fi os.FileInfo, index uint64) (*msgBlock, error) {
 	mb := &msgBlock{fs: fs, index: index, cexp: fs.fcfg.CacheExpire}
+	if len(fs.cfg.Subjects) == 1 {
+		mb.sfilter = fs.cfg.Subjects[0]
+	}
 
 	mdir := filepath.Join(fs.fcfg.StoreDir, msgDir)
 	mb.mfn = filepath.Join(mdir, fi.Name())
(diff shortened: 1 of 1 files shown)


server/filestore.go
```diff
@@ -546,8 +547,12 @@ const (
 // This is the max room needed for index header.
 const indexHdrSize = 7*binary.MaxVarintLen64 + hdrLen + checksumSize
 
+// Lock held on entry
 func (fs *fileStore) recoverMsgBlock(fi os.FileInfo, index uint64) (*msgBlock, error) {
 	mb := &msgBlock{fs: fs, index: index, cexp: fs.fcfg.CacheExpire}
+	if len(fs.cfg.Subjects) == 1 {
```
derekcollison, line 553: So on further reflection, this might need to be double checked. For instance if I start a stream with ["foo", "bar"] and later update to just ["foo"] this might not be correct.

We could use this as a hint, but for this code once we read it back in and build the fss state we know the real answer.

So after rebuilding fss for this block, we can do..

```
if len(mb.fss) == 1 {
  for sfilter := range mb.fss {
    mb.sfilter = sfilter 
  }
}

Comment (derekcollison) [reply server/filestore.go:553]: Actually since we always build fss we should only do this logic when recovering a block. That way if the stream has multiple subjects but the block only has 1 we can optimize.

File: server/filestore.go [hunk 2]
```diff
@@ -1556,6 +1573,9 @@ func (fs *fileStore) newMsgBlockForWrite() (*msgBlock, error) {
 	}
 
 	mb := &msgBlock{fs: fs, index: index, cexp: fs.fcfg.CacheExpire}
+	if len(fs.cfg.Subjects) == 1 {
```
derekcollison, line 1576: This is correct, however if we get an update we should re-evaluate this on the fs.lmb.

derekcollison (changes requested):

kozlovic:
@derekcollison Your changes look good to me. Thanks for updating the PR.

kozlovic (commented):
Confirmed that it fixes the MQTT test (TestMQTTAckWaitConfigChange) that started to fail with previous commit.

derekcollison (approved):
LGTM

PR #3010, opened 2022-04-06
Add test that demonstrated the consumer filter perf degradation

kozlovic:
This is a follow up to PR #3008.

This test fails on v2.7.4 but passes on main.


derekcollison (approved):
LGTM

PR #3011, opened 2022-04-06
Adding unique_placement_prefix to ensure matching tags are not used twice

matthiashanel:
Allows to not place a stream in the same availability zone twice

### DIFF
--- a/server/jetstream_cluster_test.go
+++ b/server/jetstream_cluster_test.go
@@ -4563,6 +4563,85 @@ func TestJetStreamClusterSuperClusterMetaPlacement(t *testing.T) {
 	}
 }
 
+func TestJetStreamUniquePlacementTag(t *testing.T) {
+	tmlp := `
+		listen: 127.0.0.1:-1
+		server_name: %s
+		jetstream: {max_mem_store: 256MB, max_file_store: 2GB, store_dir: '%s', limits: {unique_placement_prefix: az}}
+		leaf {listen: 127.0.0.1:-1}
+		cluster {
+			name: %s
+			listen: 127.0.0.1:%d
+			routes = [%s]
+		}
+		# For access to system account.
+		accounts { $SYS { users = [ { user: "admin", pass: "s3cr3t!" } ] } }
+	`
+	s := createJetStreamSuperClusterWithTemplateAndModHook(t, tmlp, 5, 2,
+		func(serverName, clustername, storeDir, conf string) string {
+			azTag := map[string]string{
+				"C1-S1": "az:same",
+				"C1-S2": "az:same",
+				"C1-S3": "az:same",
+				"C1-S4": "az:same",
+				"C1-S5": "az:same",
+				"C2-S1": "az:1",
+				"C2-S2": "az:2",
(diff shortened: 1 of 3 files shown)


kozlovic:
@matthiashanel Could you describe a bit in the PR description how the feature would be used and give an example. I am looking at the test case added (which is good) but I don't have a sense of how this works. Having more details in the PR can then be used as base for documentation, so it is not wasted. Thanks!

matthiashanel:
@kozlovic the new filed specifies a mandatory prefix of a server_tag used for placement.
When the server makes a placement decision in addition to other factors (such as placment tags provided, or disk utilization) it picks only server that have `server_tags` with the specified value as prefix and where the matching tag is unique across all selected server.  

Say a three server have the server tags below and the `unique_placement_prefix` is `az`:
```
server 1: server_tags: [az:1]
server 2: server_tags: [az:2]
server 3: server_tags: [az:3]
server 4: server_tags: [az:1]
```

An R3 stream would only be placed on servers 1-3 or 2-4. Never would server 1 and 4 be used for to host the same R3 stream.

kozlovic:
> An R3 stream would only be placed on servers 1-3 or 2-4. Never would server 1 and 4 be used for to host the same R3 stream.

Thanks, that helps a lot. Will review...

kozlovic (approved):
LGTM

server/jetstream_cluster_test.go
```diff
@@ -4563,6 +4563,85 @@ func TestJetStreamClusterSuperClusterMetaPlacement(t *testing.T) {
 	}
 }
 
+func TestJetStreamUniquePlacementTag(t *testing.T) {
+	tmlp := `
+		listen: 127.0.0.1:-1
+		server_name: %s
+		jetstream: {max_mem_store: 256MB, max_file_store: 2GB, store_dir: '%s', limits: {unique_placement_prefix: az}}
```
derekcollison, line 4570: This feels like its not a descendant of limits imo. Just inline with max_file_store etc. But could be wrong.

server/opts.go
```diff
@@ -178,8 +178,9 @@ type RemoteLeafOpts struct {
 }
 
 type JSLimitOpts struct {
-	MaxAckPending int
-	Duplicates    time.Duration
+	MaxAckPending            int
+	Duplicates               time.Duration
+	UniquePlacementTagPrefix string
```
derekcollison, line 183: These are per server yes?

How about UniqueTags?

server/opts.go
```diff
@@ -1813,6 +1814,8 @@ func parseJetStreamLimits(v interface{}, opts *Options, errors *[]error, warning
 			if err != nil {
 				*errors = append(*errors, err)
 			}
+		case "unique_placement_prefix":
```
derekcollison, line 1817: unique_tags imo is good.

derekcollison (approved):
LGTM

samuel-form3:
Awesome stuff @matthiashanel @derekcollison 

Thanks, we'll give you feedback afterwards

PR #3012, opened 2022-04-07
fix sleep in unit test to ensure updates have propagated

matthiashanel:

ran without and this and -race, spotted the issues
ran with this and -race, no issue in 40 min


kozlovic (approved):
LGTM, but is that update related to statsz update or not? If so, you may be able to speed that up as it is done with resetCh (in createJetStreamTaggedSuperCluster())

matthiashanel:
> LGTM, but is that update related to statsz update or not? If so, you may be able to speed that up as it is done with resetCh (in createJetStreamTaggedSuperCluster())

Unfortunately, a different, independent timer

kozlovic (approved):
LGTM

PR #3013, opened 2022-04-07
[ADDED] LeafNode `min_version` new option

kozlovic:
If set, a server configured to accept leafnode connections will
reject a remote server whose version is below that value. Note
that servers prior to v2.8.0 are not sending their version
in the CONNECT protocol, which means that anything below 2.8.0
would be rejected.

Configuration example:
```
leafnodes {
    port: 7422
    min_version: 2.8.0
}
```
The option is a string and can have the "v" prefix:
```
min_version: "v2.9.1"
```
Note that although suffix such as `-beta` would be accepted,
only the major, minor and update are used for the version comparison.

### DIFF
--- a/server/client.go
+++ b/server/client.go
@@ -1228,10 +1228,11 @@ func (c *client) readLoop(pre []byte) {
 		for i := 0; i < len(bufs); i++ {
 			if err := c.parse(bufs[i]); err != nil {
 				if err == ErrMinimumVersionRequired {
-					// Special case here, currently onle for leaf node connections.
-					// When processing the connect, an error was printed and sent
-					// back to the remote, but the connection is closed after a
-					// certain delay (to avoid "rapid" remote reconnection).
+					// Special case here, currently only for leaf node connections.
+					// When process the CONNECT protocol, if the minimum version
+					// required was not met, an error was printed and sent back to
+					// the remote, and connection was closed after a certain delay
+					// (to avoid "rapid" reconnection from the remote).
 					// We don't need to do any of the things below, simply return.
 					return
 				}
--- a/server/client.go
+++ b/server/client.go
@@ -202,6 +202,7 @@ const (
 	DuplicateRemoteLeafnodeConnection
 	DuplicateClientID
 	DuplicateServerName
+	MinimumVersionRequired
 )
 
 // Some flags passed to processMsgResults
@@ -1226,6 +1227,14 @@ func (c *client) readLoop(pre []byte) {
 		// to process messages, etc.
 		for i := 0; i < len(bufs); i++ {
 			if err := c.parse(bufs[i]); err != nil {
+				if err == ErrMinimumVersionRequired {
+					// Special case here, currently onle for leaf node connections.
+					// When processing the connect, an error was printed and sent
+					// back to the remote, but the connection is closed after a
+					// certain delay (to avoid "rapid" remote reconnection).
+					// We don't need to do any of the things below, simply return.
+					return
+				}
 				if dur := time.Since(start); dur >= readLoopReportThreshold {
 					c.Warnf("Readloop processing time: %v", dur)
 				}
--- a/server/leafnode.go
+++ b/server/leafnode.go
@@ -57,7 +57,7 @@ const leafNodeLoopDetectionSubjectPrefix = "$LDS."
 const leafNodeWSPath = "/leafnode"
 
 // This is the time the server will wait, when receiving a CONNECT,
-// before closing the connection is the required minimum version is not met.
+// before closing the connection if the required minimum version is not met.
 const leafNodeWaitBeforeClose = 5 * time.Second
 
 type leaf struct {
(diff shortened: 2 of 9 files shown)


derekcollison (approved):
LGTM - Thanks.

kozlovic:
@matthiashanel @philpennock @davidkemper Do you guys want to do a review or is Derek's LGTM enough and I can merge?

server/client.go
```diff
@@ -1226,6 +1227,15 @@ func (c *client) readLoop(pre []byte) {
...
 		for i := 0; i < len(bufs); i++ {
 			if err := c.parse(bufs[i]); err != nil {
+				if err == ErrMinimumVersionRequired {
+					// Special case here, currently only for leaf node connections.
+					// When process the CONNECT protocol, if the minimum version
+					// required was not met, an error was printed and sent back to
+					// the remote, and connection was closed after a certain delay
+					// (to avoid "rapid" reconnection from the remote).
+					// We don't need to do any of the things below, simply return.
+					return
```
matthiashanel, line 1237: should there be a trace here? so it's printed on either end.

server/leafnode.go
```diff
@@ -252,6 +256,18 @@ func validateLeafNode(o *Options) error {
 	if o.LeafNode.Port == 0 {
 		return nil
 	}
+
+	// If MinVersion is defined, check that it is valid.
+	if mv := o.LeafNode.MinVersion; mv != _EMPTY_ {
+		if _, _, _, err := versionComponents(mv); err != nil {
```
matthiashanel, line 262: check version components be at or above v2.8.0?
More for our users than for us

server/opts.go
```diff
@@ -1981,6 +1988,14 @@ func parseLeafNodes(v interface{}, opts *Options, errors *[]error, warnings *[]e
 		case "no_advertise":
 			opts.LeafNode.NoAdvertise = mv.(bool)
 			trackExplicitVal(opts, &opts.inConfig, "LeafNode.NoAdvertise", opts.LeafNode.NoAdvertise)
+		case "min_version", "minimum_version":
+			version := mv.(string)
+			if _, _, _, err := versionComponents(version); err != nil {
+				err := &configErr{tk, fmt.Sprintf("version parsing error: %v", err)}
+				*errors = append(*errors, err)
+				continue
+			}
```
matthiashanel, line 1997: isn't this redundant with the code in validateLeafNode

matthiashanel (commented):
Minor comments. ok with logic in general

server/leafnode.go
kozlovic: Sorry, but I don't understand this comment. Are you saying that this should be the error message logged and returned to the remove?

matthiashanel: I am wondering if we should check the configured version to be of a value that is actually usable.
Meaning 2.7.0 should not be a value one can configure.

server/opts.go
kozlovic: It is, and as you know, I often say that the check has to be done in "validate" so that it covers both config parsing and programmatic setting of the options. But when it is not too much to do, it is nice that it is done here too so that we print where in the config file the error is located, and also for the checking of the configuration file, command line `-t`, which otherwise says that the config is valid.
I am not saying that we should go back and do the checks in config parsing for all the ones that we did only in validate, but again, if "easy to do", then I think it is worth doing in both places. But I can remove if you think we should not have it in both places.

server/client.go
kozlovic: In processLeafNodeConnect we have:
```
c.sendErrAndErr(fmt.Sprintf("connection rejected since minimum version required is %q", mv))
```
which means that the accept side logs AND sends that error to the remote. The remote will see that:
```
[55435] 2022/04/07 12:11:29.131489 [INF] 127.0.0.1:7222 - lid:4 - Leafnode connection created for account: $G 
[55435] 2022/04/07 12:11:29.132505 [INF] 127.0.0.1:7222 - lid:4 - JetStream Not Extended, adding deny [$JS.API.> $KV.> $OBJ.>] for account "$G"
[55435] 2022/04/07 12:11:29.132580 [ERR] 127.0.0.1:7222 - lid:4 - Leafnode Error 'connection rejected since minimum version required is "2.9.0"'
[55435] 2022/04/07 12:11:34.133252 [INF] 127.0.0.1:7222 - lid:4 - Leafnode connection closed: Read Error account: $G
```
Again, see the read error (due to other side closing the connection) 5 seconds after being notified of the error. This is intentional so that the remote does not reconnect "right away"

matthiashanel: ok, so already logged before getting to this return

server/opts.go
matthiashanel: ok with that

server/leafnode.go
matthiashanel: I'd add that check for our users benefit.

server/client.go
kozlovic: Yes.

kozlovic:
@matthiashanel Implemented the requested changes. Have a second look when you have time.

matthiashanel (approved):
LGTM

PR #3017, opened 2022-04-09
[FIXED] Consumer deliver subject incorrect when imported and crossing a route.

derekcollison:

/cc @nats-io/core

### DIFF
--- a/server/client.go
+++ b/server/client.go
@@ -4109,6 +4109,10 @@ func (c *client) processMsgResults(acc *Account, r *SublistResult, msg, deliver,
 		// We need to rewrite the subject and the reply.
 		if li := bytes.LastIndex(creply, []byte("@")); li != -1 && li < len(creply)-1 {
 			subj, creply = creply[li+1:], creply[:li]
+			// Make sure deliver is set if inbound from a route.
+			if c.kind == ROUTER {
+				deliver = subj
+			}
 		}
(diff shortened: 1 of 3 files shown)


matthiashanel (approved):
LGTM. 
Question with @ being a special (internal) character, do we make sure client's can't publish with it? Would `pub foo@bar` be received as `bar` ?

derekcollison:
It's not on the subject, its tacked onto the reply subject and we pull that out and set local state and switch on last mile.

server/client.go
```diff
@@ -4161,6 +4163,11 @@ func (c *client) processMsgResults(acc *Account, r *SublistResult, msg, deliver,
 			} else {
 				dsubj = append(_dsubj[:0], sub.im.to...)
 			}
+
+			// Make sure deliver is set if inbound from a route.
+			if remapped && c.kind == ROUTER {
```
matthiashanel, line 4168: What about Leaf and Gateway?

matthiashanel (commented):
last minute question, leaf and gateway do not need to be considered right?

server/client.go
derekcollison: Possible, just trying to get the fix to pass which is does and not break any existing tests.

kozlovic (approved):
LGTM

PR #3020, opened 2022-04-10
[IMPROVED] Raft layer for general stability and leader election.

### LINKED ISSUE #3006: JetStream memory stream potentially duplicates/redelivers messages during restart

nats-server 2.8.0 beta 10 (92813d73707a0eaf3d3783939fa56a09ea789d3a)

Linux 5.17.1

[clustersetup.zip](https://github.com/nats-io/nats-server/files/8419504/clustersetup.zip)
[testsetup.zip](https://github.com/nats-io/nats-server/files/8419507/testsetup.zip)

* Create a jetstream cluster with three nodes
* Run the consumer `go run . -consume`
* Run the producer `go run . -publish`
* Restart node by node
* Stop the producer
* Stop the consumer

https://user-images.githubusercontent.com/2435857/161788087-a3ecfe49-6459-40a8-8848-c8694b409bcd.mp4

#### Expected result:

The output of the consumer should be a sequence without any missing numbers or duplicates. 

#### Actual result:

Some numbers show up twice.

derekcollison:

Also Resolves #3006 

/cc @nats-io/core

### DIFF
--- a/server/norace_test.go
+++ b/server/norace_test.go
@@ -3447,8 +3447,7 @@ func TestNoRaceJetStreamClusterCorruptWAL(t *testing.T) {
 	require_NoError(t, err)
 
 	// This will cause us to stepdown and truncate our WAL.
-	msgs, err := sub.Fetch(100)
-	require_NoError(t, err)
+	msgs, _ := sub.Fetch(100)
 
 	// In case we have to retry, we need to adjust the tests that follow.
 	newDelivered, ackPending, more := 200, 75, 0
@@ -3459,7 +3458,7 @@ func TestNoRaceJetStreamClusterCorruptWAL(t *testing.T) {
 		require_NoError(t, err)
 	}
 
-	checkFor(t, 20*time.Second, 500*time.Millisecond, func() error {
+	checkFor(t, 30*time.Second, 500*time.Millisecond, func() error {
 		// Make sure we changed leaders.
 		if clnew := c.consumerLeader("$G", "TEST", "dlc"); clnew == nil || cl == clnew {
 			return fmt.Errorf("Expected leader to have moved")
--- a/server/norace_test.go
+++ b/server/norace_test.go
@@ -1567,13 +1567,8 @@ func TestNoRaceJetStreamClusterSuperClusterMirrors(t *testing.T) {
 		Placement: &nats.Placement{Cluster: "C1"},
 	})
 
-	// Faster timeout since we loop below checking for condition.
-	js2, err := nc.JetStream(nats.MaxWait(50 * time.Millisecond))
-	if err != nil {
-		t.Fatalf("Unexpected error: %v", err)
-	}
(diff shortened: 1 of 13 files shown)


server/norace_test.go
```diff
@@ -3435,26 +3430,34 @@ func TestNoRaceJetStreamClusterCorruptWAL(t *testing.T) {
...
 
 	sub, err = js.PullSubscribe("foo", "dlc")
-	if err != nil {
-		t.Fatalf("Unexpected error: %v", err)
-	}
+	require_NoError(t, err)
 
 	// This will cause us to stepdown and truncate our WAL.
-	sub.Fetch(100)
+	msgs, err := sub.Fetch(100)
```
kozlovic, line 3450: I remember removing the error check in the past because it would fail there often, and based on the comment above, this call would cause the stepdown, etc.. so I figure that fetching may fail anyway. The actual check of getting messages was done below.

server/raft.go
```diff
@@ -2430,7 +2501,7 @@ func (n *raft) runAsCandidate() {
 				n.trackPeer(vresp.peer)
 				votes++
 				if n.wonElection(votes) {
-					if votes == n.numActivePeers() || lxfer {
+					if true || votes == n.numActivePeers() || lxfer {
```
kozlovic, line 2504: That `true` was probably for a debug/test you were making...

server/norace_test.go
derekcollison: The fetch would fail before with old logic that we could not fill the batch, but now we should get something all the time, which I tested. Mostly get all 100, sometimes were getting partials which is why the checks on the consumer info state would fail for those.

server/norace_test.go
```diff
@@ -3435,26 +3430,34 @@ func TestNoRaceJetStreamClusterCorruptWAL(t *testing.T) {
...
+	require_NoError(t, err)
+
+	// In case we have to retry, we need to adjust the tests that follow.
+	newDelivered, ackPending, more := 200, 75, 0
+
+	if len(msgs) != 100 {
+		more = 100 - len(msgs)
+		_, err = sub.Fetch(more)
```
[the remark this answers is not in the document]
derekcollison: Yes but I need to know if we do this to adjust below.

server/raft.go
derekcollison: Ah yes. Will fix.

kozlovic (changes requested):
Since you removed the only reference to numActivePeers() and this is not exported, staticcheck says that this is unused (and needs to be removed):
```
server/raft.go:2442:16: func (*raft).numActivePeers is unused (U1000)
```

server/norace_test.go
derekcollison: You were correct, ignore error and made fix to raft layer.

derekcollison:
ok should be good to go now.

kozlovic (approved):
LGTM

PR #3023, opened 2022-04-11
Stream Alternates

derekcollison:
This introduces stream alternates as an optional part of StreamInfo.

Alternates provide a priority list of mirrors and the source in relation to where the request originated.
So if a stream S has 2 mirrors, and we receive a user request for the stream info for S, it will include a sorted list of itself, since it could be the closest, and all mirrors, allowing clients to select read replicas if so desired.

/cc @nats-io/core

### DIFF
--- a/server/jetstream_cluster.go
+++ b/server/jetstream_cluster.go
@@ -6079,6 +6079,8 @@ func (js *jetStream) streamAlternates(ci *ClientInfo, stream string) []StreamAlt
 	defer js.mu.RUnlock()
 
 	s, cc := js.srv, js.cluster
+	// Track our domain.
+	domain := s.getOpts().JetStreamDomain
 
 	// No clustering just return nil.
 	if cc == nil {
@@ -6102,7 +6104,7 @@ func (js *jetStream) streamAlternates(ci *ClientInfo, stream string) []StreamAlt
 	for _, sa := range cc.streams[acc.Name] {
 		// Add in ourselves and any mirrors.
 		if sa.Config.Name == stream || (sa.Config.Mirror != nil && sa.Config.Mirror.Name == stream) {
-			alts = append(alts, StreamAlternate{Name: sa.Config.Name, Cluster: sa.Group.Cluster})
+			alts = append(alts, StreamAlternate{Name: sa.Config.Name, Domain: domain, Cluster: sa.Group.Cluster})
 		}
 	}
 	// If just us don't fill in.
--- a/server/jetstream_cluster.go
+++ b/server/jetstream_cluster.go
@@ -6067,6 +6067,57 @@ func (mset *stream) checkClusterInfo(ci *ClusterInfo) {
 	}
 }
 
+// Return a list of alternates, ranked by preference order to the request, of stream mirrors.
+// This allows clients to select or get more information about read replicas that could be a
+// better option to connect to versus the original source.
+func (js *jetStream) streamAlternates(ci *ClientInfo, stream string) []StreamAlternate {
+	if js == nil {
+		return nil
+	}
+
+	js.mu.RLock()
+	defer js.mu.RUnlock()
+
+	s, cc := js.srv, js.cluster
+
+	// No clustering just return nil.
+	if cc == nil {
+		return nil
+	}
+	acc, _ := s.LookupAccount(ci.serviceAccount())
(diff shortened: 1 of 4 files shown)


server/jetstream_cluster.go
```diff
@@ -6067,6 +6067,53 @@ func (mset *stream) checkClusterInfo(ci *ClusterInfo) {
 	}
 }
 
+// Return a list of alternates, ranked by preference order to the request, of stream mirrors.
+// This allows clients to select or get more information about read replicas that could be a
+// better option to connect to versus the original source.
+func (js *jetStream) streamAlternates(ci *ClientInfo, stream string) []StreamAlternate {
+	js.mu.RLock()
```
kozlovic, line 6074: Apparently, you need to check for `js != nil` here. In the code above where you invoke this, the `js.clusterInfo()` is immune to nil pointer because it checks. So either call js.streamAlternates when js is not nil, or check for nil here.

server/jetstream_cluster.go
```diff
@@ -6067,6 +6067,57 @@ func (mset *stream) checkClusterInfo(ci *ClusterInfo) {
...
+		}
+	}
+	// If just us don't fill in.
+	if len(alts) == 1 {
+		return nil
+	}
+
+	// Sort based on our weights that originate from the request itself.
+	sort.Slice(alts, func(i, j int) bool {
+		return weights[alts[i].Cluster] > weights[alts[j].Cluster]
```
kozlovic, line 6115: Normally, sort.Slice() asks to implement the "less" function. So is ">" here intended?

ripienaar:
Looks pretty good, you say optional part of stream info - but its not something i can request or not its always there, you mean optional as in it's not there post create for example right? Just to be sure we didnt miss some option somewhere

It's nice at initial start but do you anticipate for example a KV bucket would request info regularly to reconfigure itself for new mirros should they become available?

server/raft.go
```diff
@@ -2430,7 +2501,7 @@ func (n *raft) runAsCandidate() {
 				n.trackPeer(vresp.peer)
 				votes++
 				if n.wonElection(votes) {
-					if votes == n.numActivePeers() || lxfer {
+					if true || votes == n.numActivePeers() || lxfer {
```
matthiashanel, line 2504: always true?

matthiashanel (changes requested):

server/stream.go
```diff
@@ -96,13 +96,19 @@ type PubAck struct {
...
+	Domain     string              `json:"domain,omitempty"`
+	Cluster    *ClusterInfo        `json:"cluster,omitempty"`
+	Mirror     *StreamSourceInfo   `json:"mirror,omitempty"`
+	Sources    []*StreamSourceInfo `json:"sources,omitempty"`
+	Alternates []StreamAlternate   `json:"alternates,omitempty"`
+}
+
+type StreamAlternate struct {
+	Name    string `json:"name"`
+	Cluster string `json:"cluster"`
```
matthiashanel, line 111: Domain?

matthiashanel (commented):
what is the use case for this?

server/jetstream_cluster.go
derekcollison: The test makes sure the order is correct.

server/stream.go
derekcollison: This is to be able to select a read replica that is closer to you than the source.

derekcollison: Right now I don't think we can cross accounts or domains.

matthiashanel: yes It's possible to source/mirror cross domains

derekcollison: ok so you think we should include domain even if the majority of the time it will be the default or the same?

Remember we can always add things, once we do though hard if not impossible to remove them. Hence I like to start small and provide less to start.

derekcollison:
Won't merge until other PRs land and @kozlovic can do his magic to not have all the redundant commits.

kozlovic (approved):
LGTM

kozlovic:
@derekcollison Rebase done.. Let's make sure we have a green run, but the rebase itself was painless.

server/stream.go
derekcollison: I added.

PR #3026, opened 2022-04-12
[FIXED] subject renaming with hand crafted reply subject

matthiashanel:

Please read comment in client.go. probably should make this illegal, @ or not.

### DIFF
--- a/server/client.go
+++ b/server/client.go
@@ -3493,8 +3493,15 @@ func isReservedReply(reply []byte) bool {
 	if isServiceReply(reply) {
 		return true
 	}
+	rLen := len(reply)
 	// Faster to check with string([:]) than byte-by-byte
-	if len(reply) > gwReplyPrefixLen && string(reply[:gwReplyPrefixLen]) == gwReplyPrefix {
+	if rLen > jsAckPreLen+1 && string(reply[:jsAckPreLen]) == jsAckPre {
+		// Not sure what should happen if a JS message is relayed by a client (probably shouldn't support that)
+		// But it should never contain an @ (used for renaming)
+		if strings.Contains(string(reply[jsAckPreLen:]), "@") {
+			return true
+		}
+	} else if rLen > gwReplyPrefixLen && string(reply[:gwReplyPrefixLen]) == gwReplyPrefix {
 		return true
 	}
 	return false
--- a/server/client.go
+++ b/server/client.go
@@ -3495,12 +3495,8 @@ func isReservedReply(reply []byte) bool {
 	}
 	rLen := len(reply)
 	// Faster to check with string([:]) than byte-by-byte
-	if rLen > jsAckPreLen+1 && string(reply[:jsAckPreLen]) == jsAckPre {
-		// Not sure what should happen if a JS message is relayed by a client (probably shouldn't support that)
-		// But it should never contain an @ (used for renaming)
-		if strings.Contains(string(reply[jsAckPreLen:]), "@") {
-			return true
-		}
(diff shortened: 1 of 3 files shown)


server/client.go
```diff
@@ -3493,8 +3493,15 @@ func isReservedReply(reply []byte) bool {
 	if isServiceReply(reply) {
 		return true
 	}
+	rLen := len(reply)
 	// Faster to check with string([:]) than byte-by-byte
-	if len(reply) > gwReplyPrefixLen && string(reply[:gwReplyPrefixLen]) == gwReplyPrefix {
+	if rLen > jsAckPreLen+1 && string(reply[:jsAckPreLen]) == jsAckPre {
+		// Not sure what should happen if a JS message is relayed by a client (probably shouldn't support that)
+		// But it should never contain an @ (used for renaming)
+		if strings.Contains(string(reply[jsAckPreLen:]), "@") {
```
derekcollison, line 3501: Just use bytes.Contains imo.

server/client.go
```diff
@@ -3493,8 +3493,15 @@ func isReservedReply(reply []byte) bool {
 	if isServiceReply(reply) {
 		return true
 	}
+	rLen := len(reply)
 	// Faster to check with string([:]) than byte-by-byte
-	if len(reply) > gwReplyPrefixLen && string(reply[:gwReplyPrefixLen]) == gwReplyPrefix {
+	if rLen > jsAckPreLen+1 && string(reply[:jsAckPreLen]) == jsAckPre {
```
derekcollison, line 3498: We should not allow clients to use that prefix at all.

derekcollison: And need to make sure its API prefix aware, etc. Not sure if that is applicable.

derekcollison (changes requested):

server/client.go
matthiashanel: not applicable

server/client.go
matthiashanel: I'll deny the prefix, that covers this as well.

kozlovic: Do we care if this is under `$JS.ACK.`? I mean maybe we just check that there is no `@` allowed regardless of prefix?

kozlovic (approved):
LGTM, but see comment about checking of $JS.ACK prefix

kozlovic (approved):
LGTM (disregard my previous comment since was done was you were updating the PR)

matthiashanel:
tbh, I don't think anyone should be publishing with an ack prefix, `@` or not. This covers the issue as `@` renaming is a JS ack functionality. 

In addition to that I can make `@` an illegal/reserved character. 
But that should probably be done for pub subjects as well. 
Meaning, if `foo@bar` is an illegal reply subject, it should be illegal to publish on that as well.

server/client.go
matthiashanel: we only look for an `@` if the reply subject has the `jsAckPre`

```
	var remapped bool
	if len(creply) > 0 &&
		c.kind != CLIENT && c.kind != SYSTEM && c.kind != JETSTREAM && c.kind != ACCOUNT &&
		bytes.HasPrefix(creply, []byte(jsAckPre)) {
		// We need to rewrite the subject and the reply.
		if li := bytes.LastIndex(creply, []byte("@")); li != -1 && li < len(creply)-1 {
			remapped = true
			subj, creply = creply[li+1:], creply[:li]
		}
	}
```

kozlovic: Yes, I wrote that comment before you updated that we would reject JSACK regardless of presence of '@'. So I am ok with current code.

PR #3029, opened 2022-04-12
[FIXED] JetStream stream catchup issues and deadlock

kozlovic:
- A stream could become leader when it should not, causing
messages to be lost.
- A catchup could stall because the server sending data
could bail out of the runCatchup routine but still send
the EOF signal.
- Deadlock with monitoring of Jsz
    

derekcollison (approved):
LGTM

Issue #3027, opened 2022-04-12
MQTT client can't connect leafnode after cluster restart


LLLLimbo:
> This issue may be related to #3009 .

## Description
The problem is currently found to occur with the problem described in #3009.
1. Under normal circumstances, restarting the entire cluster will not cause this problem.
2. After removing the Jetstream storage from the leaf node and reboot, the MQTT client can connect to it normally.
3. When the problem occurs, I can connect to the remote node normally using the MQTT client

**Environment**
nats-version : 2.7.3
system: centos 7.6

**Cluster Information**
3 remote nodes, 1 leaf node
Configuration as follows:
https://github.com/LLLLimbo/nats-conf

The full log and Jetstream storage file as follows (After restarting the cluster, trying to connect to the leaf node with the MQTT client) :
https://github.com/LLLLimbo/nats-logs/releases/tag/before-delete
Restarted at around 2022/04/12 09:30. 
When I try to connect to the leaf node using the MQTT client, the log for the leaf node shows the following message:
```
[15891] 2022/04/12 09:44:07.556991 [TRC] 192.168.9.251:60863 - mid:311 - "mqttx_8fec4c09" - <<- [CONNECT clientID=mqttx_8fec4c09 keepAlive=1m30s username=seeiner-edge-mqtt password=****]
[15891] 2022/04/12 09:44:07.557184 [INF] Creating MQTT streams/consumers with replicas 1 for account "SEEINER"
[15891] 2022/04/12 09:44:07.557406 [TRC] 192.168.3.135:7422 - lid:15 - ->> [LS+ $MQTT.JSA.gf9flzF3.*.*]
[15891] 2022/04/12 09:44:07.557470 [TRC] 192.168.3.135:7422 - lid:15 - ->> [LS+ $MQTT.JSA.*.SP.*]
[15891] 2022/04/12 09:44:07.557504 [TRC] 192.168.3.135:7422 - lid:15 - ->> [LS+ $MQTT.sub.6mH3UM68HpFqAzAHYU9lix]
[15891] 2022/04/12 09:44:07.557521 [TRC] 192.168.3.135:7422 - lid:15 - ->> [LS+ $MQTT.JSA.*.RD]
[15891] 2022/04/12 09:44:07.558531 [DBG] JETSTREAM - JetStream connection closed: Client Closed
[15891] 2022/04/12 09:44:07.558559 [DBG] JETSTREAM - JetStream connection closed: Client Closed
[...]

Issue #3009, opened 2022-04-06
Leaf nodes do not receive messages from remote nodes properly


LLLLimbo:
# Description
I started a NATS cluster with three remote nodes and one leaf node. They are configured as follows:
https://github.com/LLLLimbo/nats-conf/tree/main
At first they work fine, the leaf nodes receive the messages correctly when I publish them to the remote nodes and vice versa, but whenever they run for a while, it happens that the leaf nodes do not receive the messages from the remote nodes correctly.
![GIF](https://user-images.githubusercontent.com/31607003/161912213-818a312c-c111-4789-84e1-e4bf8e01be13.gif)

~~The problem may be related to jetstream, when I delete the storage file of jetstream, the cluster is back to normal.~~
2022-04-12 correction：A simple restart restores communication between the remote node and the leaf node.

*Environment*
- nats-version : 2.7.3
- system: centos 7.6

derekcollison:
I suspect the authorization block for the servers receiving leafnode connections may be incorrect.

LLLLimbo:
> I suspect the authorization block for the servers receiving leafnode connections may be incorrect.

I encountered this problem again, this time I tried to delete only the Jetstream storage on the leaf node and the cluster returned to normal.

kozlovic:
@LLLLimbo Would you mind sending some reports from the NATS cli tools (`nats`) to see if you have some streams defined, how their configuration look like, etc? I believe you are using MQTT too, are you having MQTT applications subscribing to "test"? Usually, how long does it take before the failure start to occur? Any servers restart before it starts to fail?

LLLLimbo:
> @LLLLimbo Would you mind sending some reports from the NATS cli tools (`nats`) to see if you have some streams defined, how their configuration look like, etc? I believe you are using MQTT too, are you having MQTT applications subscribing to "test"? Usually, how long does it take before the failure start to occur? Any servers restart before it starts to fail?

1. I didn't define any jetstream streams.
2. I don't have any  MQTT client subscribes to `test`
3. I can't give a specific time, but usually after a few hours
4. There is no reboot

The following information is obtained（from leafnode） when the cluster is working properly.
```
PS C:\DevelopmentKit\nats\nats-cli> ./nats --server nats://seeiner-edge:seeiner123@192.168.3.131:4222 account info
Connection Information:

               Client ID: 25
               Client IP: 192.168.9.251
                     RTT: 2.2613ms
       Headers Supported: true
[...]

PR #3031, opened 2022-04-13
[FIXED] LeafNode interest propagation with imports/exports

kozlovic:
When using subscriptions through import/exports, the server with
a leafnode connection would properly send the interest over, but
if the connection is recreated, this would not happen.

In case of JetStream where that happens under the cover, message
flow would stop after the leafnode restart because the subscriptions
would be created on recovery of the JetStream assets but *before*
the LeafNode connection could be established.

Resolves https://github.com/nats-io/nats-server/issues/3024
Resolves https://github.com/nats-io/nats-server/issues/3027
Resolves https://github.com/nats-io/nats-server/issues/3009

### DIFF
--- a/server/jetstream_test.go
+++ b/server/jetstream_test.go
@@ -7361,11 +7361,13 @@ func TestJetStreamStreamStorageTrackingAndLimits(t *testing.T) {
 	nc.Flush()
 
 	state = mset.state()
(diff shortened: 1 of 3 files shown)


matthiashanel (approved):
LGTM

derekcollison (approved):
LGTM

PR #3032, opened 2022-04-13
[Adding] max_ha_assets to limit placement on server with more ha assets

matthiashanel:
server running more than max_ha_assets #raft nodes will not be used to
place new streams and fail if not enough free server can be found.
Durable Consumer creation on such server will fail as their peer size is
bound to the same set as their stream.

This also avoids updating placement where no new placement is needed.

### DIFF
--- a/server/jetstream_cluster.go
+++ b/server/jetstream_cluster.go
@@ -4126,6 +4126,14 @@ func (cc *jetStreamCluster) selectPeerGroup(r int, cluster string, cfg *StreamCo
 
 		// Otherwise check if we have enough room if maxBytes set.
 		if maxBytes > 0 && maxBytes > available {
+			s.Warnf("%s@%s (Max Bytes: %d) exceeds available %s storage of %d bytes",
+				ni.name, ni.cluster, maxBytes, cfg.Storage.String(), available)
+			continue
+		}
+		// HAAssets contain _meta_ which we want to ignore
+		if maxHaAssets > 0 && ni.stats != nil && ni.stats.HAAssets > maxHaAssets {
+			s.Warnf("%s@%s (HA Asset Count: %d) exceeds max ha asset limit of %d for stream placement",
+				ni.name, ni.cluster, ni.stats.HAAssets, maxHaAssets)
 			continue
 		}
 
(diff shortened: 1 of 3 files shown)


server/jetstream_cluster.go
```diff
@@ -4144,6 +4145,13 @@ func (cc *jetStreamCluster) selectPeerGroup(r int, cluster string, cfg *StreamCo
 				continue
 			}
 		}
+
+		// HAAssets contain _meta_ which we want to ignore
+		if maxHaAssets > 0 && ni.stats != nil && ni.stats.HAAssets > maxHaAssets {
+			s.Warnf("%s@%s (ha asset count: %d) exceeds max ha asset limit of %d for stream placement",
```
derekcollison, line 4151: (HA Asset Count: %d)

Easier to see in logs

server/jetstream_cluster.go
```diff
@@ -4144,6 +4145,13 @@ func (cc *jetStreamCluster) selectPeerGroup(r int, cluster string, cfg *StreamCo
...
+				ni.name, ni.cluster, ni.stats.HAAssets, maxHaAssets)
```
derekcollison, line 4152: We should log for storage resource overflow as well.

server/jetstream_cluster.go
```diff
@@ -5308,6 +5318,23 @@ func (s *Server) jsClusteredConsumerRequest(ci *ClientInfo, acc *Account, subjec
 				break
 			}
 		}
+		if check {
```
derekcollison, line 5321: If we placed the stream but now can't place the consumer we can't overflow a consumer, they have to overlay the same peer set. If this is just moving to new peer that is ok, the one inherited from the parent stream.

matthiashanel: are you saying, don't check the limit for consumer?

server/jetstream_cluster.go
matthiashanel: adding in next push

server/jetstream_cluster.go
derekcollison: Not saying that, just saying we can't overflow a consumer to a new cluster.

derekcollison: So we would have to fail the consumer create at that point..

matthiashanel: That's what this does.

derekcollison (approved):
LGTM!

PR #3035, opened 2022-04-14
[IMPROVED] JetStream: reduce unnecessary leader election

kozlovic:
- Wait of some sort of routing to be in place before starting
the raft run loop
- Remove use of lock in apiDispatch that was not necessary but
could have cause a route to block, causing memory growth, etc..

Unrelated rename of some tests so that they start with TestJetStream
and TestJetStreamCluster for cluster tests, fixed some flappers
and ensure that tests that change RAFT timeouts put them back
to default values on exit.

### DIFF
--- a/server/raft.go
+++ b/server/raft.go
@@ -236,13 +236,28 @@ const (
 )
 
 var (
+	minElectionTimeout time.Duration
+	maxElectionTimeout time.Duration
+	minCampaignTimeout time.Duration
+	maxCampaignTimeout time.Duration
+	hbInterval         time.Duration
+	lostQuorumInterval time.Duration
+)
+
+func init() {
+	setDefaultRaftTimeouts()
+}
+
+// This is called on init, but also by tests that want to tweak the settings
+// and then call this as a defer to restore the defaults.
+func setDefaultRaftTimeouts() {
 	minElectionTimeout = minElectionTimeoutDefault
 	maxElectionTimeout = maxElectionTimeoutDefault
 	minCampaignTimeout = minCampaignTimeoutDefault
 	maxCampaignTimeout = maxCampaignTimeoutDefault
-	hbInterval         = hbIntervalDefault
+	hbInterval = hbIntervalDefault
 	lostQuorumInterval = lostQuorumIntervalDefault
-)
+}
 
 type RaftConfig struct {
 	Name     string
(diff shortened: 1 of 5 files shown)


server/jetstream_cluster.go
```diff
@@ -1499,7 +1503,9 @@ func (js *jetStream) createRaftGroup(accName string, rg *raftGroup, storage Stor
 		s.Debugf("Error creating raft group: %v", err)
 		return err
 	}
+	js.mu.Lock()
 	rg.node = n
+	js.mu.Unlock()
 
 	// See if we are preferred and should start campaign immediately.
 	if n.ID() == rg.Preferred && n.Term() == 0 {
```
derekcollison, line 1511: We reference rg.Preferred here so I think we still need the lock or need to capture.

derekcollison (changes requested):

server/jetstream_cluster.go
kozlovic: Ok, I checked n.Campain() and not basically reseting some timer, so I could have "r.node = n" and everything else under the js.mu.Lock(), is that ok?

derekcollison: Yes I think so.

kozlovic: Ok, done

derekcollison (approved):
LGTM

kozlovic:
@derekcollison Following our discussion, I have updated the PR, please have a quick look.

kozlovic:
@derekcollison Following our discussion, I have updated the PR, please have a quick look.

server/raft.go
```diff
@@ -1528,6 +1543,12 @@ func (n *raft) resetElectionTimeout() {
 	n.resetElect(randElectionTimeout())
 }
 
+func (n *raft) resetElectionTimeoutWithLock() {
```
derekcollison, line 1546: Just have this one call resetElectWithLock(randElectionTimeout())?

server/raft.go
```diff
@@ -1543,10 +1564,41 @@ func (n *raft) resetElect(et time.Duration) {
...
+			gw.RLock()
+			ready = len(gw.out)+len(gw.in) > 0
+			gw.RUnlock()
+		}
+		s.mu.Unlock()
+		if !ready {
+			select {
+			case <-s.quitCh:
+				return
+			case <-time.After(50 * time.Millisecond):
```
derekcollison, line 1594: Make 100?

derekcollison (approved):
LGTM - Minor suggestions.

server/raft.go
kozlovic: done

kozlovic: done

server/raft.go
kozlovic: Done

kozlovic: Done

PR #3038, opened 2022-04-14
[CHANGED] JetStream: accept "Nats-Expected-Last-Sequence" with "0"

kozlovic:
We use to ignore if the seq was 0, but now would treat it as
a requirement that the stream be empty if header is present but
set to 0.

This relates to client PR: https://github.com/nats-io/nats.go/pull/958

### DIFF
--- a/server/jetstream_test.go
+++ b/server/jetstream_test.go
@@ -3478,6 +3478,16 @@ func TestJetStreamPublishExpect(t *testing.T) {
 		t.Fatalf("Expected an error, got %q", resp.Data)
 	}
 
+	// Or if we expect that there are no messages by setting "0" as the expected last seq
+	m.Header.Set(JSExpectedLastSeq, "0")
+	resp, err = nc.RequestMsg(m, 100*time.Millisecond)
+	if err != nil {
+		t.Fatalf("Unexpected error: %v", err)
+	}
+	if pa := getPubAckResponse(resp.Data); pa == nil || pa.Error == nil {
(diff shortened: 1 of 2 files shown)


ripienaar:
Looks good, but seems unlikely @derekcollison would have forgotten to add this when we did the per-subject one to support KV Create(), so maybe there was a reason for not doing it? Do you remember anything @derekcollison ?

kozlovic:
@ripienaar That's what I fear too, so a bit hesitant and would not mind if we don't add this. In the client PR you mentioned that we should have it for consistency (compared to the last-ser-per-subject). Have you changed your mind seeing the code change?

ripienaar:
I reckon it has good value both in being consistent and as a feature - but I am also aware we are very near a big release and might punt till 2.9.x, so would be fine if we waited a tad to give it proper thought.

kozlovic:
@ripienaar Could not agree more. Will switch this PR to draft and will reconsider post 2.8.0.

derekcollison (approved):
LGTM

This will mean you only want to put a message in if a stream is brand new, and you only get one shot at that since purge will drop messages but not reset the first sequence to 0.

kozlovic:
@derekcollison Ah yes. There may be a use-case, but yeah, pretty limited. Let's leave this as draft for now.

kozlovic:
@derekcollison @bruth @ripienaar I have rebased from main branch and switched from Draft to ready for review. If approved I will merge.

PR #3039, opened 2022-04-15
Make sure if we recreate something after deleting that we do not wipe valid state

derekcollison:
This window is small, less than 2 secs in most cases, but can happen.

/cc @nats-io/core


kozlovic (approved):
LGTM

PR #3040, opened 2022-04-15
Add catchup messages limit that was removed

kozlovic:
Need to also limit the number of outstanding messages during a
catchup. It was removed in PR #3029


derekcollison (approved):
LGTM

PR #3041, opened 2022-04-15
Raft improvements

derekcollison:
We would declare lost quorum too quickly on scale up.

/cc @nats-io/core

### DIFF
--- a/server/jetstream_api.go
+++ b/server/jetstream_api.go
@@ -1812,12 +1812,12 @@ func (s *Server) jsStreamInfoRequest(sub *subscription, c *client, a *Account, s
 			// while the new members work through the election and catchup process.
 			// Double check for that instead of exiting here and being silent. e.g. nats stream update test --replicas=3
 			js.mu.RLock()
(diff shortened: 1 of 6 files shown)


server/jetstream_api.go
```diff
@@ -1812,8 +1812,12 @@ func (s *Server) jsStreamInfoRequest(sub *subscription, c *client, a *Account, s
 			// while the new members work through the election and catchup process.
 			// Double check for that instead of exiting here and being silent. e.g. nats stream update test --replicas=3
 			js.mu.RLock()
-			rg, ourID := sa.Group, cc.meta.ID()
-			bail := !rg.isMember(ourID)
+			rg, bail := sa.Group, true
```
kozlovic, line 1815: You set bail to true, but it is then set to `!rg.isMember(ourID)` just below, so that does not really do anything.

derekcollison: Fair, fixed.

kozlovic (approved):
LGTM

PR #3044, opened 2022-04-16
Fixed mixed mode server without JS dropping js export on jwt update

matthiashanel:

The removed code is subsequently handled by addSystemAccountExports.
Right now I am using !standalone as signal to add in the exports. But could add an option, or check for options as well.

### DIFF
--- a/server/events.go
+++ b/server/events.go
@@ -1013,8 +1013,8 @@ func (s *Server) addSystemAccountExports(sacc *Account) {
 	if err := sacc.AddServiceExport(accSubsSubj, nil); err != nil {
 		s.Errorf("Error adding system service export for %q: %v", accSubsSubj, err)
 	}
-
-	if s.JetStreamEnabled() {
+	// in case of a mixed mode setup, enable js exports anyway
+	if s.JetStreamEnabled() || !s.standAloneMode() {
 		s.checkJetStreamExports()
 	}
 }
--- a/server/jwt_test.go
+++ b/server/jwt_test.go
@@ -6851,12 +6851,10 @@ func TestJWTSysAccUpdateMixedMode(t *testing.T) {
 
 	s := sc.clusterForName("C1").randomServer()
 
-	sysNc := natsConnect(t, s.ClientURL(), sUsr, disconnectCb)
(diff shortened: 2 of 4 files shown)


kozlovic (approved):
LGTM

server/jwt_test.go
```diff
@@ -6787,3 +6787,134 @@ func TestJWTAccountConnzAccessAfterClaimUpdate(t *testing.T) {
...
+	defer close(disconnectChan)
+	disconnectCb := nats.DisconnectErrHandler(func(conn *nats.Conn, err error) {
+		disconnectChan <- struct{}{}
+	})
+
+	s := sc.clusterForName("C1").randomServer()
+
+	sysNc := natsConnect(t, s.ClientURL(), sUsr, disconnectCb)
+	defer sysNc.Close()
+	defer sysNc.SetDisconnectErrHandler(nil)
```
kozlovic, line 6856: FYI: Instead of doing that, you could pass the option `nats.NoCallbacksAfterClientClose()` in the Connect() call. With that option, the disconnect (and close) handler(s) won't be invoked in Close()

kozlovic: And you don't need to close the disconnectChan, regardless. You are using a buffered channel, so that won't block the disconnect handler on exit, even if you were not using NoCallbacksAfterClientClose option.

matthiashanel: switched to NoCallbacksAfterClientClose instead of clearing the error handler. thanks for that tip.
I'll keep the channel close, seems off without.

server/events.go
```diff
@@ -1013,8 +1013,8 @@ func (s *Server) addSystemAccountExports(sacc *Account) {
 	if err := sacc.AddServiceExport(accSubsSubj, nil); err != nil {
 		s.Errorf("Error adding system service export for %q: %v", accSubsSubj, err)
 	}
-
-	if s.JetStreamEnabled() {
+	// in case of a mixed mode setup, enable js exports anyway
+	if s.JetStreamEnabled() || !s.standAloneMode() {
```
derekcollison, line 1017: Exports always need to be enabled regardless of standalone status.

PR #3047, opened 2022-04-18
If we moved a stream back to a cluster we were once at, the stream would be empty.

derekcollison:
The old raft node assignment would prevent proper catchup.

/cc @nats-io/core


kozlovic (approved):
LGTM

PR #3049, opened 2022-04-18
Bump ordered consumer flow control window back up.

derekcollison:
Have seen some small dips in some global systems that @ripienaar has been running and I have been testing. Originally bumped it down due to some instability but that was not associated with this in the end.

/cc @nats-io/core

### DIFF
--- a/server/consumer.go
+++ b/server/consumer.go
@@ -295,7 +295,7 @@ const (
 	JsDeleteWaitTimeDefault = 5 * time.Second
 	// JsFlowControlMaxPending specifies default pending bytes during flow control that can be
 	// outstanding.
-	JsFlowControlMaxPending = 16 * 1024 * 1024
+	JsFlowControlMaxPending = 32 * 1024 * 1024
 	// JsDefaultMaxAckPending is set for consumers with explicit ack that do not set the max ack pending.
 	JsDefaultMaxAckPending = 1000
 )


ripienaar (approved):
LGTM, though i think my drop in traffic was because majority of stream leaders were on one machine, after a bit of poking to spread leaders out traffic is as expected

server/consumer.go
```diff
@@ -295,7 +295,7 @@ const (
 	JsDeleteWaitTimeDefault = 5 * time.Second
 	// JsFlowControlMaxPending specifies default pending bytes during flow control that can be
 	// outstanding.
-	JsFlowControlMaxPending = 16 * 1024 * 1024
+	JsFlowControlMaxPending = 32 * 1024 * 1024
```
kozlovic, line 298: * Before was:
1MB and maxpb = limit / 16
* 19 days ago:
changed to 32MB *AND* changed maxpb = limit / 4
* 4 days ago:
changed to 16MB *AND* changed maxpb = limit / 16
* today:
changed back to 32MB but maxpb not changed... hard to follow..

kozlovic (changes requested):
Check if maxpb needs to be changed too.

server/consumer.go
derekcollison: Nope that is ok and will leave as is. Thanks.

kozlovic: Sorry, so are you saying that there is no need to change maxpb even if you update JSFlowConstrolMaxPending? If so, let me know and I will approve the PR.

derekcollison: Correct, I want it to start at /16 or 2mb

kozlovic (approved):
LGTM

PR #3051, opened 2022-04-18
[changed] if max_request_batch and  max_batch is 0 respond with an error

matthiashanel:
instead if setting a default value. This is to fail early during
consumer creation, not during fetch with a too large value.

### DIFF
--- a/server/errors.json
+++ b/server/errors.json
@@ -1240,10 +1240,10 @@
     "deprecates": ""
   },
   {
-    "constant": "JSConsumerMaxRequestBatchRequired",
+    "constant": "JSConsumerMaxRequestBatchRequiredF",
     "code": 400,
     "error_code": 10126,
-    "description": "consumer max request batch needs to be specified",
+    "description": "consumer max request batch needs to be specified and cannot be more than {limit}",
     "comment": "",
     "help": "",
     "url": "",
--- a/server/errors.json
+++ b/server/errors.json
@@ -1238,5 +1238,15 @@
     "help": "",
     "url": "",
     "deprecates": ""
+  },
+  {
+    "constant": "JSConsumerMaxRequestBatchRequired",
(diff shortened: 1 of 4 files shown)


server/errors.json
```diff
@@ -1238,5 +1238,15 @@
     "help": "",
     "url": "",
     "deprecates": ""
+  },
+  {
+    "constant": "JSConsumerMaxRequestBatchRequired",
+    "code": 400,
+    "error_code": 10126,
+    "description": "consumer max request batch needs to be specified",
```
kozlovic, line 1246: Should we include the current limit here? Something like: `consumer max request batch needs to be specified and cannot be more than {limit}`. If so, re-run `go generate` and then you will be able to pass the limit in the NewJSConsumerMaxRequestBatchRequiredError() function.

server/jetstream_test.go
```diff
@@ -16678,24 +16678,23 @@ func TestJetStreamLimits(t *testing.T) {
 		require_Equal(t, err.Error(), "duplicates window can not be larger then server limit of 1m0s")
 
 		ci, err := js.AddConsumer("foo", &nats.ConsumerConfig{Durable: "dur1", AckPolicy: nats.AckExplicitPolicy})
-		require_NoError(t, err)
-		require_True(t, ci.Config.MaxAckPending == 1000)
-		require_True(t, ci.Config.MaxRequestBatch == 250)
+		require_Error(t, err)
+		require_Equal(t, err.Error(), "consumer max request batch needs to be specified")
```
kozlovic, line 16682: If you accept my change, then would need to update this expected error message here.

kozlovic (changes requested):
Recommend changes, but open for discussion..

server/errors.json
matthiashanel: done

PR #3052, opened 2022-04-18
[FIXED] JetStream: some stream SOURCE issues

kozlovic:
- Possibly missing some early messages from the sourced stream
- In some cancel situations, the processing of sourced messages
would not longer work


derekcollison (approved):
LGTM! Thanks.

PR #3053, opened 2022-04-18
[IMPROVED] JetStream: add file names for invalid checksums

kozlovic:
On restart, we report when we find error in checksums, but we
did not report the name of the file.


matthiashanel (approved):
LGTM

PR #3056, opened 2022-04-19
introducing max_age/dupe_window minimum value of 100ms.

matthiashanel:

Added minimum values of 100ms (used elsewhere as well)


kozlovic (approved):
LGTM.. assuming that all tests will pass

PR #3058, opened 2022-04-20
[FIXED] LeafNode: propagation interest issue after a config reload

### LINKED ISSUE #3009: Leaf nodes do not receive messages from remote nodes properly

kozlovic:
When a configuration reload is done, the account's leaf node connections
were not transfered to the new instance of the account, causing the
interest to not be propagated until a leafnode reconnect or a server
restart.

Resolves #3009


derekcollison (approved):
LGTM

PR #3059, opened 2022-04-20
Polling too quickly could make migration fail.

derekcollison:
When we polled too quickly on migration we could check before catchup logic had even kicked in.

/cc @nats-io/core


kozlovic (approved):
LGTM

Issue #2801, opened 2022-01-20
Stream editing with external source cause message multiplying in target stream


sargikk:
## Defect

Make sure that these boxes are checked before submitting your issue -- thank you!

 - [x] Included `nats-server -DV` output
 - [x] Included a [Minimal, Complete, and Verifiable example] (https://stackoverflow.com/help/mcve)

```
[118] 2022/01/19 14:35:06.298597 [INF] Starting nats-server
[118] 2022/01/19 14:35:06.298643 [INF]   Version:  2.7.0
[118] 2022/01/19 14:35:06.298647 [INF]   Git:      [not set]
[118] 2022/01/19 14:35:06.298654 [DBG]   Go build: go1.17.6
[118] 2022/01/19 14:35:06.298685 [INF]   Name:     NDGUT5AJFRTTN5SZAO765QM3XUPFISJQV7QX4WHYVBF7IZPBIFF7PDKE
[118] 2022/01/19 14:35:06.298694 [INF]   ID:       NDGUT5AJFRTTN5SZAO765QM3XUPFISJQV7QX4WHYVBF7IZPBIFF7PDKE
[118] 2022/01/19 14:35:06.298789 [DBG] Created system account: "$SYS"
```

#### Versions of `nats-server` and affected client libraries used:
nats-server 2.7.0
synadia/nats-box:0.6.0
#### OS/Container environment:
OS: Ubuntu 20.04.2
Docker image: nats:2.7.0-alpine3.15
#### Steps or code to reproduce the issue:
All steps a very similar to the official guide:
https://docs.nats.io/running-a-nats-service/configuration/leafnodes/jetstream_leafnodes#cross-account-and-domain-import
The only difference is that we have to create stream with source, not mirror.

1. Grant permissions like in official cross account guide
2. Create stream with external source, not mirror
3. Update any stream parameter via "nats stream edit", "Duplicate Window" or subjects for example
4. Send any message to source stream with appropriate subject

#### Expected result:

Get 1 message in target stream.

#### Actual result:

Get 2 copies of original message in target stream.
Each update of stream increases number of copies by one.

#### Possible reasons:
Issues #2202 #2209 that introduces iname for using multiple sources with same name.
More precisely old config sources collected by Name.
https://github.com/nats-io/nats-server/blob/297b44394f92e990da450178cffaa7adecc88837/server/stream.go#L1024
And new config sources than compares by iname.
[...]

PR #3061, opened 2022-04-20
[FIXED] JetStream: stream sources issue in mixed mode clusters

kozlovic:
The main issue was that in mixed-mode, the interest through gateway
may still be in optimistic mode, which when creating the source
consumer would start delivery before we had a chance to setup
the subscription to receive those messages.

The approach is to create the subscription prior to sending
the consumer create request. Also refactored a bit the code in
the hope to make the retries a bit more bullet proof.

We may also look at making sure that gateways are switched to
interest-mode when detecting a mixed-mode setup.

Also fixed a defect that could cause a source to be canceled
when updating a stream.

Resolves https://github.com/nats-io/nats-server/issues/2801

### DIFF
--- a/server/ipqueue.go
+++ b/server/ipqueue.go
@@ -191,6 +191,9 @@ func (q *ipQueue) len() int {
 // notified that there is something in the queue (reading from queue's `ch`)
 // may then get nothing if `drain()` is invoked before the `pop()` or `popOne()`.
 func (q *ipQueue) drain() {
+	if q == nil {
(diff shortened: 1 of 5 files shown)


kozlovic:
@derekcollison Will likely apply the same fixes to the mirror case in a separate PR.

derekcollison (approved):
LGTM

derekcollison (approved):
LGTM!

PR #3062, opened 2022-04-20
Fix subject renaming for leaf connections and queued subs

matthiashanel:
change in `server/client.go` essentially copies the added code from the non queue part.
Both section of code add a check for leaf connections. (original unit test was wrong).


kozlovic (approved):
LGTM

PR #3060, opened 2022-04-20
Fixes issue on restart where mirror config was invalid

matthiashanel:
This issue was introduced by a bug that assigned a de dupe default value
do mirrors.
This is an invalid config that used to be checked on stream create.
Subsequently this logic was moved and then got executed on startup.
Ending up checking mirrors that got the bad default assigned.
Fix is to clear de dupe window for mirrors

Signed-off-by: Matthias Hanel <mh@synadia.com>

manual restart test.
1st start is without this change
2nd with this
```
> nats-server -sd ./store -js
[95749] 2022/04/20 13:03:42.591726 [INF] Starting nats-server
[95749] 2022/04/20 13:03:42.591833 [INF]   Version:  2.8.1-beta
[95749] 2022/04/20 13:03:42.591836 [INF]   Git:      [not set]
[95749] 2022/04/20 13:03:42.591842 [INF]   Name:     NDNENZRBTENE453IVC3G2F53VVFZU3U4TWGN7Z5KPHVKXRBFDKB5ODUB
[95749] 2022/04/20 13:03:42.591846 [INF]   Node:     f4skLVsG
[95749] 2022/04/20 13:03:42.591848 [INF]   ID:       NDNENZRBTENE453IVC3G2F53VVFZU3U4TWGN7Z5KPHVKXRBFDKB5ODUB
[95749] 2022/04/20 13:03:42.592195 [INF] Starting JetStream
[95749] 2022/04/20 13:03:42.592508 [INF]     _ ___ _____ ___ _____ ___ ___   _   __  __
[95749] 2022/04/20 13:03:42.592514 [INF]  _ | | __|_   _/ __|_   _| _ \ __| /_\ |  \/  |
[95749] 2022/04/20 13:03:42.592517 [INF] | || | _|  | | \__ \ | | |   / _| / _ \| |\/| |
[95749] 2022/04/20 13:03:42.592519 [INF]  \__/|___| |_| |___/ |_| |_|_\___/_/ \_\_|  |_|
[95749] 2022/04/20 13:03:42.592520 [INF]
[95749] 2022/04/20 13:03:42.592522 [INF]          https://docs.nats.io/jetstream
[95749] 2022/04/20 13:03:42.592523 [INF]
[95749] 2022/04/20 13:03:42.592525 [INF] ---------------- JETSTREAM ----------------
[95749] 2022/04/20 13:03:42.592528 [INF]   Max Memory:      48.00 GB
[...]

PR #3067, opened 2022-04-21
[FIXED] JetStream: Mirrors would fail to be recovered

kozlovic:
This is a continuation of PR #3060, but extends to clustering.

Verified with manual test that a mirror created with v2.7.4 has
the duplicates window set and on restart with main would still
complain about use of dedup in cluster mode. The mirror stream
was recovered but showing as R1.
With this fix, a restart of the cluster - with existing data -
will properly recover the stream as an R3 and messages that
were published while in a bad state are synchronized.

Signed-off-by: Matthias Hanel mh@synadia.com


matthiashanel (approved):
LGTM

PR #3074, opened 2022-04-25
Implement JSConsumerDeliveryNakAdvisory

### LINKED ISSUE #3072: Consider advisory subject for NAK
#### Use Case:

From a Slack user:

> Is there a native way to monitor NAKs of messages? E.g. a trigger for when x amount of NAKs in the last 5 minutes were sent. Or when a message is "stuck" because it has been NAKed for y-times. The http monitoring endpoint nor JMS support this feature as far as I can tell, right?

#### Proposed Change:

Add an advisory subject similar to the [TERM subject](https://github.com/nats-io/nats-server/blob/646b3850bffe185b3efa0b58981ea6f84b32ef49/server/jetstream_api.go#L207-L208) so that clients can monitor for NAKs and react appropriately.

Anyone who wants to monitor NAKs.

#### Alternative Approaches

Currently, there is no way to observe NAKs from the server. The alternative would be for the application to track this manually.

Coffeeri:
Resolves #3072

 - add the advisory `JSAdvisoryConsumerMsgNakPre` will be triggered when a message is naked.

/cc @nats-io/core

### DIFF
--- a/server/jetstream_events.go
+++ b/server/jetstream_events.go
@@ -110,6 +110,21 @@ type JSConsumerDeliveryExceededAdvisory struct {
 // JSConsumerDeliveryExceededAdvisoryType is the schema type for JSConsumerDeliveryExceededAdvisory
 const JSConsumerDeliveryExceededAdvisoryType = "io.nats.jetstream.advisory.v1.max_deliver"
 
+// JSConsumerDeliveryNakAdvisory is an advisory informing that a message was
+// naked by the consumer
+type JSConsumerDeliveryNakAdvisory struct {
+	TypedEvent
+	Stream      string `json:"stream"`
+	Consumer    string `json:"consumer"`
+	ConsumerSeq uint64 `json:"consumer_seq"`
+	StreamSeq   uint64 `json:"stream_seq"`
+	Deliveries  uint64 `json:"deliveries"`
+	Domain      string `json:"domain,omitempty"`
+}
+
+// JSConsumerDeliveryNakAdvisoryType is the schema type for JSConsumerDeliveryNakAdvisory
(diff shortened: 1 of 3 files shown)


server/jetstream_events.go
```diff
@@ -110,6 +110,21 @@ type JSConsumerDeliveryExceededAdvisory struct {
...
+// JSConsumerDeliveryNakAdvisory is an advisory informing that a message was
+// naked by the consumer
+type JSConsumerDeliveryNakAdvisory struct {
+	TypedEvent
+	Stream      string `json:"stream"`
+	Consumer    string `json:"consumer"`
+	ConsumerSeq uint64 `json:"consumer_seq"`
+	StreamSeq   uint64 `json:"stream_seq"`
+	Deliveries  uint64 `json:"deliveries"`
+	Domain      string `json:"domain,omitempty"`
```
ripienaar, line 122: We dont have domain on the others, probably something we should consider doing, good call

ripienaar (approved):
LGTM

server/jetstream_events.go
derekcollison: Should we do that here, adding domain to the others or in separate PR?

Coffeeri: I've removed the draft, so you @ripienaar can merge this PR. I am not quite following in which other places the domain should be added. As far as I see, all Jetstream advisories do have it.

ripienaar: You're right, sorry for the distration, I missed that they were added previously.  So I think we're good.

ripienaar:
Thank you @Coffeeri

PR #3078, opened 2022-04-26
Changes to tests that run on Travis

kozlovic:
- Remove code coverage from Travis and add it to a GitHub Action
that will be run as a nightly.
- Use tag builds to exclude some tests, such as the "norace" or
JS tests. Since "go test" does not support "negative" regexs, there
is no other way.


derekcollison (approved):
LGTM! Awesome!

PR #3079, opened 2022-04-27
[fixed] step down timing, consumer stream seqno, clear redelivery

matthiashanel:
Step down timing for consumers or streams.
Signals loss of leadership and sleeps before stepping down.
This makes it less likely that messages are being processed during step
down.

When becoming leader, consumer stream seqno got reset,
even though the consumer existed already.

Proper cleanup of redelivery data structures and timer

### DIFF
--- a/server/consumer.go
+++ b/server/consumer.go
@@ -861,8 +861,11 @@ func (o *consumer) setLeader(isLeader bool) {
 		mset.mu.RUnlock()
 
 		o.mu.Lock()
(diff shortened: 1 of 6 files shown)


matthiashanel:
I ran Marco's test without issues with 50 consumer step downs.
I'm somewhat unclear on how to best test this in a unit test.

derekcollison (approved):
LGTM!

kozlovic (approved):
LGTM

PR #3080, opened 2022-04-27
Fix for a consumer recovery issue and update to healthz.

derekcollison:
We had an issue where a recovered consumer might not recover if a source based stream which housed the consumer filter subject was removed.

We now signal that we are recovering a consumer and if so skip the filtered subject check.

Also per ops re-worked healthz to test all assigned assets, just not those that we could recover.

Also various fixes found during @matthiashanel original diagnosis.

/cc @nats-io/core

### DIFF
--- a/server/jetstream_cluster_test.go
+++ b/server/jetstream_cluster_test.go
@@ -10471,3 +10471,60 @@ func TestJetStreamClusterMirrorDeDupWindow(t *testing.T) {
 	send(100)
 	check(200)
 }
+
+func TestJetStreamClusterNewHealthz(t *testing.T) {
+	c := createJetStreamClusterExplicit(t, "JSC", 3)
+	defer c.shutdown()
+
+	nc, js := jsClientConnect(t, c.randomServer())
+	defer nc.Close()
+
+	_, err := js.AddStream(&nats.StreamConfig{
+		Name:     "R1",
+		Subjects: []string{"foo"},
+		Replicas: 1,
+	})
+	require_NoError(t, err)
+
+	_, err = js.AddStream(&nats.StreamConfig{
+		Name:     "R3",
+		Subjects: []string{"bar"},
+		Replicas: 3,
+	})
+	require_NoError(t, err)
+
+	// Create subscribers (durable and ephemeral for each)
+	fsube, err := js.SubscribeSync("foo")
+	require_NoError(t, err)
+	fsubd, err := js.SubscribeSync("foo", nats.Durable("d"))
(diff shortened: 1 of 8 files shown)


matthiashanel (commented):
As per your comment: 
> Also per ops re-worked healthz to test all assigned assets, just not those that we could recover.

does this mean healthz will or will not block for not recovered assets?

derekcollison:
It now will block on assets that could not be recovered.

matthiashanel (approved):
LGTM

server/jetstream_cluster_test.go
```diff
@@ -10471,3 +10471,124 @@ func TestJetStreamClusterMirrorDeDupWindow(t *testing.T) {
 	send(100)
 	check(200)
 }
+
+// We had a scenario where a consumer would not recover properly on restart due to
+// the cluster state not being set properly when checking source subjects.
+func TestJetStreamClusterStateOnRestartPreventsConsumerRecovery(t *testing.T) {
+	sc := createJetStreamTaggedSuperCluster(t)
```
kozlovic, line 10478: Since this looks like a super cluster test, and assuming that you are up-to-date with main branch, would you mind moving this test in jetstream_super_cluster_test.go (at the end) and call it TestJetStreamSuperClusterStateOnRestartPreventsConsumerRecovery?

server/jetstream_cluster_test.go
```diff
@@ -10471,3 +10471,124 @@ func TestJetStreamClusterMirrorDeDupWindow(t *testing.T) {
...
+
+	// Now make sure the consumer is still on this server and has restarted properly.
+	mset, err := cl.GlobalAccount().lookupStream("DS")
+	require_NoError(t, err)
+	if o := mset.lookupConsumer("dlc"); o == nil {
+		t.Fatalf("Consumer was not properly restarted")
+	}
+}
+
+func TestJetStreamClusterNewHealthz(t *testing.T) {
```
kozlovic, line 10539: This one can stay here (since simple cluster) and is properly named. Thanks!

kozlovic (changes requested):

server/jetstream_cluster_test.go
derekcollison: Yep.. Will do.

server/jetstream_cluster_test.go
derekcollison: ok sounds good.

kozlovic (approved):
LGTM

PR #3081, opened 2022-04-27
[IMPROVED] KV memory store performance

derekcollison:
For memory store KV with history of 1 we were scanning for our next first when we did not have to on rewrites.

/cc @nats-io/core


jnmoyne (approved):
Works great now, thanks!

kozlovic (commented):
LGTM

PR #3090, opened 2022-04-29
Combined canSubscribe and canQueueSubscribe

derekcollison:
Want to consistently handle canSubscribe. 

/cc @nats-io/core


server/accounts_test.go
```diff
@@ -3380,3 +3380,39 @@ func TestAccountLimitsServerConfig(t *testing.T) {
...
+    `))
+	defer removeFile(t, cf)
+
+	s, _ := RunServerWithConfig(cf)
+	defer s.Shutdown()
+
+	nc, err := nats.Connect(s.ClientURL(), nats.UserInfo("user", "pass"))
+	require_NoError(t, err)
+
+	// Now do same with qsub.
```
kozlovic, line 3408: Copy-paste from another test I presume, since we only do queue sub in this test.

kozlovic (approved):
LGTM, minor test comments.

server/accounts_test.go
derekcollison: Yes, I had both to start but the trick was it had to be solo to present the issue. Will clean up test and comments.

kozlovic (approved):
LGTM

PR #3091, opened 2022-04-29
[added] support for jwt operator option DisallowBearerToken

### LINKED ISSUE #3084: Account Policy to Restrict Bearer Token Users
## Feature Request

In operator-mode configurations (JWT-based), an account owner can create bearer token User JWTs using the `--bearer` option of `nsc` or the the JWT client libraries.

Some operators may desire to create accounts that explicitly restrict the account owner from creating users with the bearer-token option.
                
#### Use Case:

Force client to present User JWT **and** perform a challenge-and-response authentication, i.e. sign nonce using possessed private NKEY. 

#### Proposed Change:

Add an account-level claim that asserts that the account may not issue bearer User JWTs.

Organizations/operators that have a policy against bearer token usage at runtime, but otherwise enable account owners to manage their own user lifecycle.

matthiashanel:
I modified an existing data structure that held a similar attribute already.
Instead this data structure references the claim.

change 3 out of 3. Fixes #3084
corresponds to:
https://github.com/nats-io/jwt/pull/177
https://github.com/nats-io/nsc/pull/495

### DIFF
--- a/server/jwt_test.go
+++ b/server/jwt_test.go
@@ -5760,6 +5760,51 @@ func TestJWTStrictSigningKeys(t *testing.T) {
 	})
 }
 
+func TestJWTDisallowBearer(t *testing.T) {
+	opId, err := nkeys.CreateOperator()
+	require_NoError(t, err)
+	opIdPub, err := opId.PublicKey()
+	require_NoError(t, err)
+	oClaim := jwt.NewOperatorClaims(opIdPub)
+	oClaim.DisallowBearerToken = true
+	oJwt, err := oClaim.Encode(opId)
+	require_NoError(t, err)
+
+	accId, err := nkeys.CreateAccount()
+	require_NoError(t, err)
+	accIdPub, err := accId.PublicKey()
+	require_NoError(t, err)
+	aClaim := jwt.NewAccountClaims(accIdPub)
+	accJwt, err := aClaim.Encode(opId)
+	require_NoError(t, err)
+
+	uc := jwt.NewUserClaims("dummy")
+	uc.BearerToken = true
+	uOpt1 := createUserCredsEx(t, uc, accId)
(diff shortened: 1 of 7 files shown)


server/jwt_test.go
```diff
@@ -5760,6 +5760,51 @@ func TestJWTStrictSigningKeys(t *testing.T) {
...
+			%s : "%s"
+		}
+		`, oJwt, accIdPub, accJwt)))
+	defer removeFile(t, cf)
+	s, _ := RunServerWithConfig(cf)
+	defer s.Shutdown()
+
+	nc1, err := nats.Connect(s.ClientURL(), uOpt1)
+	require_Error(t, err)
+	defer nc1.Close()
```
kozlovic, line 5801: FYI: this is ok because Close() protects against the connection being nil, which should be the case here since you expected an error, that is connection did not succeed and did not return an `nc1`.

server/server.go
```diff
@@ -972,25 +972,28 @@ func (s *Server) ActivePeers() (peers []string) {
 
 // isTrustedIssuer will check that the issuer is a trusted public key.
 // This is used to make sure an account was signed by a trusted operator.
-func (s *Server) isTrustedIssuer(issuer string) bool {
```
kozlovic, line 975: You could add a little comment above that says that this function also returns "true" is bearer token are not allowed.

kozlovic (approved):
LGTM, but suggested added a comment that isTrustedIssuer returns if bearer token are allowed or not?

server/jwt_test.go
matthiashanel: fixed

server/server.go
matthiashanel: added comment

kozlovic (approved):
LGTM

PR #3092, opened 2022-05-02
[FIXED] JetStream: possible lock inversion

kozlovic:
When updating usage, there is a lock inversion in that the jetStream
lock was acquired while under the stream's (mset) lock, which is
not correct. Also, updateUsage was locking the jsAccount lock, which
again, is not really correct since jsAccount contains streams, so
it should be jsAccount->stream, not the other way around.

Removed the locking of jetStream to check for clustered state since
js.clustered is immutable.

Replaced using jsAccount lock to update usage with a dedicated lock.

Originally moved all the update/limit fields in jsAccount to new
structure to make sure that I would see all code that is updating
or reading those fields, and also all functions so that I could
make sure that I use the new lock when calling these. Once that
works was done, and to reduce code changes, I put the fields back
into jsAccount (although I grouped them under the new usageMu mutex
field).

### DIFF
--- a/locksordering.txt
+++ b/locksordering.txt
@@ -2,6 +2,9 @@ Here is the list of some established lock ordering.
 
 In this list, A -> B means that you can have A.Lock() then B.Lock(), not the opposite.
 
-jetStream -> jsAccount -> Server -> client-> Account
+jetStream -> jsAccount -> Server -> client -> Account
 
-stream -> consumer
+jetStream -> jsAccount -> stream -> consumer
+
+A lock to protect jetstream account's usage has been introduced: jsAccount.usageMu.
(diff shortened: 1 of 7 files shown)


matthiashanel (approved):
LGTM, Mind putting that lock into lockordering.txt?

kozlovic:
@matthiashanel Will update the file, but keep in mind that this lock is "independent", meaning that any other lock can get it, since no code under `updateMu` lock grabs another one (well, maybe expect ipQueue to push, but ipQueue lock again is independent, used only to add/dequeue to/from ipQueue list).

matthiashanel:
ok.

matthiashanel (approved):
LGTM

PR #3096, opened 2022-05-02
Fixed GithubAction code coverage to handle test panics

kozlovic:
We are ok with a flapper or two, because they should not affect
code coverage that much, so it is better to have those and publish
code coverage than to have to recycle the whole test suite until
we get no test failure.

However, if there is a test panic, then all other tests within this
package will NOT run, which then would have possibly a massive
impact in the code coverage percentage.

These changes will ensure that the run fails if one of the code
coverage output is "empty" (it is actually not empty, but the
initial content is "mode: atomic" and then whe code coverage is
complete, it gets filled with actual code coverage data).
On failure, the push to coverall will not happen.

### DIFF
--- a/.github/workflows/cov.yaml
+++ b/.github/workflows/cov.yaml
@@ -26,10 +26,10 @@ jobs:
 
       - name: Run code coverage
         shell: bash --noprofile --norc -x -eo pipefail {0}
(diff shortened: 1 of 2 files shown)


derekcollison (approved):
LGTM

PR #3099, opened 2022-05-03
[FIXED] JetStream: PullConsumer MaxWaiting==1 and Canceled requests

kozlovic:
There was an issue with MaxWaiting==1 that was causing a request
with expiration to actually not expire. This was because processWaiting
would not pick it up because wq.rp was actually equal to wq.wp
(that is, the read pointer was equal to write pointer for a slice
of capacity of 1).

The other issue was that when reaching the maximum of waiting pull
requests, a new request would evict an old one with a "408 Request Canceled".

There is no reason for that, instead the server will first try to
find some existing expired requests (since some of the expiration
is lazily done), but if none is expired, and the queue is full,
the server will return a "409 Exceeded MaxWaiting" to the new
request, and not a "408 Request Canceled" to an old one...

### DIFF
--- a/server/consumer.go
+++ b/server/consumer.go
@@ -2368,9 +2368,9 @@ func (wr *waitingRequest) recycle() {
 
 // waiting queue for requests that are waiting for new messages to arrive.
 type waitQueue struct {
-	rp, wp int
-	last   time.Time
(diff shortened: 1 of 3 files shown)


derekcollison (approved):
LGTM!

PR #3100, opened 2022-05-03
Raft and cluster improvements.

derekcollison:
Some small improvements to the raft layer and clustering layer trying to diagnose bad raft state during upgrades from 2.7.x to 2.8.x.

I believe some could be caused by ill-advised snapshot sends for consumers when not a leader but not only when scaling up from R1. Also fixed small bug that would not reset catchingUp state and could lead to processing logic errors in the raft layer.

During move or scale up will snapshot so followers can kick up to upper layers to sync vs raft only.

 - [ ] Tests added

Resolves #

 -
 -
 -

/cc @nats-io/core


kozlovic (approved):
LGTM

PR #3101, opened 2022-05-03
[IMPROVED] JetStream: check max-per-subject once

kozlovic:
There was a case where we may have done a check for max-per-subject
limit twice per message. That would apply to streams that have
max-per-subject and also discard_new, which is what KV configures.


derekcollison (approved):
LGTM

PR #3109, opened 2022-05-07
[FIXED] Downstream sourced retention policy streams during restart have redelivered messages

### LINKED ISSUE #3069: [JS] workqueue stream that is sourced from 2 different kvs with one of the subjects filtered, looses consumer sequence on metaleader change
I have a stream that is source from two KVs:

```
             Subjects: ngs-workers.>
     Acknowledgements: true
            Retention: File - WorkQueue
             Replicas: 3
       Discard Policy: Old
     Duplicate Window: 2m0s
    Allows Msg Delete: true
         Allows Purge: true
       Allows Rollups: false
     Maximum Messages: unlimited
  Maximum Per Subject: 1
        Maximum Bytes: unlimited
          Maximum Age: unlimited
    Maximum Consumers: unlimited
              Sources: KV_ngs-accounts, Subject: $KV.ngs-accounts.billing.*
                       KV_ngs-config
```

Several different consumers on the above stream - note consumers cannot overlap.

When a metaleader gets moved, or in cases where the nats-server may be restarted, messages (all) from the source stream will get re-sent, causing the workers to spin up even though it is not because of a change.

### LINKED ISSUE #3107: sourcing stream with retention work queue policy experiences re delivered messages
In this unit test a work queue sources another stream with messages.
It receives all messages, such that NumPending is 0.
After server restart, num pending is back to the original message count.

This happens only with work queue retention policy.

```
func TestJetStreamWorkQueueSourceRestart(t *testing.T) {
	o := s.getOpts()
	defer nc.Close()

	pubCnt := 10
	_, err := js.AddStream(&nats.StreamConfig{
		Name:     "FOO",
		Replicas: 1,
		Subjects: []string{"foo"},
	})
	for i := 0; i < pubCnt; i++ {
		_, err = js.Publish("foo", nil)
		require_NoError(t, err)
	}

		Name:     "TEST",
		Replicas: 1,
		// TODO test will pass when retention commented out
		Retention: nats.WorkQueuePolicy,
		Sources:   []*nats.StreamSource{{Name: "FOO"}}})
	require_NoError(t, err)

	time.Sleep(time.Second)

	_, err = js.AddConsumer("TEST", &nats.ConsumerConfig{Durable: "dur", AckPolicy: nats.AckExplicitPolicy})
	require_NoError(t, err)

	sub, err := js.PullSubscribe("", "dur", nats.BindStream("TEST"))
	require_NoError(t, err)

	ci, err := js.ConsumerInfo("TEST", "dur")
	require_NoError(t, err)
	require_True(t, ci.NumPending == uint64(pubCnt))

	for i := 0; i < pubCnt; i++ {
		m, err := sub.Fetch(1)
		require_NoError(t, err)
		require_NoError(t, m[0].AckSync())
	}

	ci, err = js.ConsumerInfo("TEST", "dur")
	require_NoError(t, err)
	require_True(t, ci.NumPending == 0)

	// Restart server
	s.Shutdown()
	sr := RunServer(o)
	defer sr.Shutdown()

	checkFor(t, 10*time.Second, time.Second, func() error {
		hs := sr.healthz()
		if hs.Status != "ok" {
			return fmt.Errorf("healthz %s %s", hs.Error, hs.Status)
		}
		return nil
	})

	ctest, err := js.ConsumerInfo("TEST", "dur")
	//TODO (mh) I have experienced in other tests that numwaiting has a value of 1 post restart. (
	// seems to go awary in single server setup. It's also unrelated to work queue
	// but that error seems benign.
	require_True(t, ctest.NumPending == 0)

	_, err = sub.Fetch(1, nats.MaxWait(10*time.Second))
	if err != nats.ErrTimeout {
		require_NoError(t, err)
	}
}
```

derekcollison:
We now remember the last timestamp by not removing last msgBlk when empty and during purge pull last timestamp forward until new messages arrive.

Now when a downstream stream that uses retention modes that delete messages, we will fallback to a time based start time for the new source consumers.

Resolves #3107 
Resolves #3069 

/cc @nats-io/core


server/jetstream_test.go
```diff
@@ -17160,6 +17163,108 @@ func TestJetStreamImportConsumerStreamSubjectRemapSingle(t *testing.T) {
 	})
 }
 
+func TestJetStreamWorkQueueSourceRestart(t *testing.T) {
+	s := RunBasicJetStreamServer()
+	defer s.Shutdown()
```
kozlovic, line 17168: Move down this defer after the defer of removeDir please.

kozlovic (approved):
LGTM except for minor change in one of the test.

PR #3110, opened 2022-05-07
Fix for panic due to not loaded cache during compact

derekcollison:
Reported by @ripienaar 

/cc @nats-io/core


ripienaar (approved):
LGTM, tests report a race but probably not due to this

derekcollison:
I will take a look at the race report in a bit.

derekcollison:
Fixed the race, will merge.

PR #3113, opened 2022-05-09
altered move unit test to test tiered/non tiered setup

matthiashanel:

expand unit test to cover stream move tiered/non tiered.
I also tried out deleting stream + sleep Minute + check account info. (always worked)


kozlovic (approved):
LGTM

PR #3115, opened 2022-05-10
[FIXED] JetStream: possible lockup due to a return prior to unlock

kozlovic:
This would happen in situation where a node receives an append
entry with a term higher than the node's (current leader).

This was introduced by moving a return [here](https://github.com/nats-io/nats-server/pull/3100/files#diff-0a3d4e9b87bb3880bdb735fa489ef618b571de3370dae31d8f509b26e2fb852eR2709) in PR #3100.


derekcollison (approved):
LGTM! Thanks.

PR #3126, opened 2022-05-15
Support for MaxBytes for pull requests.

derekcollison:
Getting us prepped for move to all pull consumers for clients and for the internal consumers for mirrors/sources.

/cc @nats-io/core

### DIFF
--- a/server/jetstream_test.go
+++ b/server/jetstream_test.go
@@ -650,7 +650,7 @@ func TestJetStreamConsumerMaxDeliveries(t *testing.T) {
 
 func TestJetStreamNextReqFromMsg(t *testing.T) {
 	bef := time.Now()
-	expires, _, _, _, _, err := nextReqFromMsg([]byte(`{"expires":5000000000}`)) // nanoseconds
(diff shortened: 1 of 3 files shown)


matthiashanel (approved):
LGTM

server/jetstream_test.go
```diff
@@ -17369,6 +17377,114 @@ func TestJetStreamWorkQueueSourceNamingRestart(t *testing.T) {
 	}
 }
 
+func TestJetStreamPullMaxBytes(t *testing.T) {
+	s := RunBasicJetStreamServer()
+	defer s.Shutdown()
```
kozlovic, line 17382: Move the defer shutdown past the defer removeDir please.

kozlovic (changes requested):

server/jetstream_test.go
derekcollison: ok done. I keep copying it from somewhere that is like that IIRC.

kozlovic (approved):
LGTM

PR #3128, opened 2022-05-16
Allow explicit consumer configuration of replica count and memory storage

derekcollison:
Allowing more control over consumers for large number of consumer setups as we have been seeing lately.

/cc @nats-io/core

### DIFF
--- a/server/jetstream_cluster_test.go
+++ b/server/jetstream_cluster_test.go
@@ -10528,3 +10528,89 @@ func TestJetStreamClusterNewHealthz(t *testing.T) {
 	sl = c.restartServer(sl)
 	c.waitOnServerHealthz(sl)
 }
+
+func TestJetStreamClusterConsumerOverrides(t *testing.T) {
+	c := createJetStreamClusterExplicit(t, "JSC", 3)
+	defer c.shutdown()
+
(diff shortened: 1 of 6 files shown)


server/jetstream_cluster_test.go
```diff
@@ -10528,3 +10528,89 @@ func TestJetStreamClusterNewHealthz(t *testing.T) {
...
+	require_NoError(t, err)
+
+	ci, err := nc.Request(fmt.Sprintf(JSApiDurableCreateT, "TEST", "d"), req, time.Second)
+	require_NoError(t, err)
+
+	var resp JSApiConsumerCreateResponse
+	err = json.Unmarshal(ci.Data, &resp)
+	require_NoError(t, err)
+
+	if resp.Error == nil || resp.Error.ErrCode != 10126 {
```
kozlovic, line 10566: Would do: 
```
if resp.Error == nil || !IsNatsErr(resp.Error, JSConsumerReplicasExceedsStream) {
``` 
instead of checking err code directly..

kozlovic (approved):
LGTM (but small comment about checking err code directly)

server/jetstream_cluster_test.go
derekcollison: ok done.

PR #3129, opened 2022-05-17
Enable republishing of messages once stored in a stream.

derekcollison:
This enables lightweight distribution of messages to very large number of NATS subscribers.
We add in metadata as headers that allows for gap detection which enables initial value (via JetStream, maybe KV) and real-time NATS core sub updates but all being globally ordered and gap detection allowed, even for a subset of messages.

/cc @nats-io/core

### DIFF
--- a/server/jetstream_api.go
+++ b/server/jetstream_api.go
@@ -1234,6 +1234,41 @@ func (s *Server) jsStreamCreateRequest(sub *subscription, c *client, _ *Account,
 		return
 	}
 
+	// If we have a republish directive check if we can create a transform here.
+	if cfg.RePublish != nil {
+		// Check to make sure source is a valid subset of the subjects we have.
+		// Also make sure it does not form a cycle.
+		var srcValid bool
+		for _, subj := range cfg.Subjects {
+			if SubjectsCollide(cfg.RePublish.Source, subj) {
+				srcValid = true
+				break
+			}
+		}
+		if !srcValid {
+			resp.Error = NewJSStreamInvalidConfigError(fmt.Errorf("stream configuration for republish source is not valid subset of subjects"))
+			s.sendAPIErrResponse(ci, acc, subject, reply, string(msg), s.jsonResponse(&resp))
+			return
+		}
+		var formsCycle bool
+		for _, subj := range cfg.Subjects {
+			if SubjectsCollide(cfg.RePublish.Destination, subj) {
+				formsCycle = true
+				break
+			}
+		}
+		if formsCycle {
+			resp.Error = NewJSStreamInvalidConfigError(fmt.Errorf("stream configuration for republish destination forms a cycle"))
+			s.sendAPIErrResponse(ci, acc, subject, reply, string(msg), s.jsonResponse(&resp))
+			return
+		}
+		if _, err := newTransform(cfg.RePublish.Source, cfg.RePublish.Destination); err != nil {
+			resp.Error = NewJSStreamInvalidConfigError(fmt.Errorf("stream configuration for republish not valid"))
+			s.sendAPIErrResponse(ci, acc, subject, reply, string(msg), s.jsonResponse(&resp))
+			return
+		}
+	}
+
 	// Hand off to cluster for processing.
 	if s.JetStreamIsClustered() {
 		s.jsClusteredStreamRequest(ci, acc, subject, reply, rmsg, &cfg)
--- a/server/stream.go
+++ b/server/stream.go
@@ -950,7 +950,8 @@ func (s *Server) checkStreamCfg(config *StreamConfig, acc *Account) (StreamConfi
 		return StreamConfig{}, NewJSStreamInvalidConfigError(fmt.Errorf("duplicates window can not be larger then max age"))
 	}
 	if lim.Duplicates > 0 && cfg.Duplicates > lim.Duplicates {
-		return StreamConfig{}, NewJSStreamInvalidConfigError(fmt.Errorf("duplicates window can not be larger then server limit of %v", lim.Duplicates.String()))
+		return StreamConfig{}, NewJSStreamInvalidConfigError(fmt.Errorf("duplicates window can not be larger then server limit of %v",
+			lim.Duplicates.String()))
 	}
 	if cfg.Duplicates > 0 && cfg.Duplicates < 100*time.Millisecond {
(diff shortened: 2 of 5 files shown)


server/stream.go
```diff
@@ -396,6 +415,17 @@ func (a *Account) addStreamWithAssignment(config *StreamConfig, fsConfig *FileSt
 		mset.ackq = s.newIPQueue(qpfx + "acks") // of uint64
 	}
 
+	// Check for RePublish.
+	if cfg.RePublish != nil {
+		tr, err := newTransform(cfg.RePublish.Source, cfg.RePublish.Destination)
+		if err != nil {
+			jsa.mu.Unlock()
+			return nil, fmt.Errorf("stream configuration for republish not valid")
```
kozlovic, line 423: Don't we have a place, something like checkStreamCfg where validity test should be done? If so, I wonder if here you simply do the transform but don't check for error per-se (tr will be nil I assume if there is an error).

server/util.go
```diff
@@ -316,7 +316,7 @@ func getURLsAsString(urls []*url.URL) []string {
 // copyBytes make a new slice of the same size than `src` and copy its content.
 // If `src` is nil, then this returns `nil`
 func copyBytes(src []byte) []byte {
-	if src == nil {
```
kozlovic, line 319: Would have maintained the original semantic: you pass nil, you get nil back, but if you pass empty slice, you get an empty slice. Hopefully you are not breaking code that would have depended on that...

server/stream.go
derekcollison: I check in api layer where it should be caught. So this should not trigger.

server/util.go
derekcollison: empty slice should not cause new malloc imo, so when I saw that I thought it was wrong.

derekcollison: Since if interior could be empty but non-nil.

server/stream.go
kozlovic: But what about updates, restore, etc? I think that @matthiashanel had to go through that already and tried to consolidate into a single place... but I may be wrong.

derekcollison: I could try to co-locate there for just checking. Some are there, some are at other places.

derekcollison: Can't update this atm and I am ok with that, want to see if its really useful etc.

server/util.go
kozlovic: When I wrote this helper, I added the comment that if src is nil, then result is nil, but otherwise the new slice is same size than original, which, I agree could mean allocating a slice of size 0, but that meant a matching source/destination in term of nil vs not nil. But I think that's fine... although updating the doc then would match better the new behavior.

kozlovic (commented):
Would like @matthiashanel to chime in for the configuration check..

server/stream.go
derekcollison: Actually this is where we create the transform, we already checked it in the API layer.

server/util.go
derekcollison: Updated comment.

server/jetstream_api.go
```diff
@@ -1234,6 +1234,41 @@ func (s *Server) jsStreamCreateRequest(sub *subscription, c *client, _ *Account,
 		return
 	}
 
+	// If we have a republish directive check if we can create a transform here.
```
matthiashanel, line 1237: Agree with @kozlovic , but it's this code that should move

I'd move this entire block into `func (s *Server) checkStreamCfg(config *StreamConfig, acc *Account) (StreamConfig, *ApiError)` `stream.go` line `858` or below.
This is the place for stream config checks like that.
In the current form restore would not check this.
Probably shortens the change as the error response part is dealt with by the caller already.

The previous check for `sealed` above only exists as it explicitly applies to create only.

matthiashanel (changes requested):
I'd move the check in `jsStreamCreateRequest`.

server/jetstream_api.go
derekcollison: Yep done..

matthiashanel (approved):
LGTM

server/stream.go
```diff
@@ -71,6 +74,12 @@ type StreamConfig struct {
 	AllowRollup bool `json:"allow_rollup_hdrs"`
 }
 
+// SubjectMapping allows a source subject to be mapped to a destination subject for republishing.
+type SubjectMapping struct {
+	Source      string `json:"src,omitempty"`
```
ripienaar, line 79: Is this really optional? not setting it just gives me `stream configuration for republish source is not valid subset of subjects`

ripienaar, line 79: Is this really optional? not setting it just gives me `stream configuration for republish source is not valid subset of subjects`

server/jetstream_api.go
```diff
@@ -1234,6 +1234,41 @@ func (s *Server) jsStreamCreateRequest(sub *subscription, c *client, _ *Account,
...
+				break
+			}
+		}
+		if formsCycle {
+			resp.Error = NewJSStreamInvalidConfigError(fmt.Errorf("stream configuration for republish destination forms a cycle"))
+			s.sendAPIErrResponse(ci, acc, subject, reply, string(msg), s.jsonResponse(&resp))
+			return
+		}
+		if _, err := newTransform(cfg.RePublish.Source, cfg.RePublish.Destination); err != nil {
+			resp.Error = NewJSStreamInvalidConfigError(fmt.Errorf("stream configuration for republish not valid"))
```
ripienaar, line 1266: Using this is quite hard because there's just no indication of why its wrong

server/stream.go
derekcollison: Thought was that if omitted it meant ">", but beta feature and want to see how useful this will be.

server/jetstream_api.go
derekcollison: Could attach the error if needed. Same rules as import mapping from exports/imports.

ripienaar: Yeah the errors from the is also not great - was looking to send a pr to expose err but it’s not great. 

Getting mappings right esp for partitions really hard

derekcollison:
Feel free to submit PR.

server/jetstream_api.go
derekcollison: I expected normal usage to just be match and prefix, so `foo -> RP.foo`, or even `> -> RP.>`, but you can match only a subset.

Issue #3116, opened 2022-05-11
MQTT support should be horizontal scalable


ianzhang1988:
## Feature Request
nats-server: v2.8.0 

When I benchmark MQTT protocal, I found that MQTT support is supported with 3 stream with replica set to 3:

``` 
│ $MQTT_rmsgs │             │ 2022-05-10 09:05:33 │ 0        │ 0 B     │ never        │
│ $MQTT_sess  │             │ 2022-05-10 09:05:28 │ 28,348   │ 6.1 MiB │ 6m50s        │
│ $MQTT_msgs  │             │ 2022-05-10 09:05:32 │ 561,704  │ 176 MiB │ 2h45m31s     │
```

nats stream info '$MQTT_msgs' output:
```
Configuration:

             Subjects: $MQTT.msgs.>
     Acknowledgements: true
            Retention: File - Interest
             Replicas: 3
       Discard Policy: Old
     Duplicate Window: 2m0s
    Allows Msg Delete: true
         Allows Purge: true
       Allows Rollups: false
     Maximum Messages: unlimited
        Maximum Bytes: unlimited
          Maximum Age: unlimited
 Maximum Message Size: unlimited
    Maximum Consumers: unlimited


Cluster Information:

                 Name: nats
               Leader: nats-5
              Replica: nats-3, current, seen 0.05s ago
              Replica: nats-4, current, seen 0.05s ago

State:

             Messages: 561,704
                Bytes: 176 MiB
             FirstSeq: 1 @ 2022-05-11T04:45:29 UTC
              LastSeq: 960,236 @ 2022-05-11T06:22:32 UTC
     Deleted Messages: 398532
     Active Consumers: 9419

```

As I understand, this could only utilize 3 server.

#### Use Case:
Where MQTT is used.

#### Proposed Change:
Enable horizontal scalable.

#### Who Benefits From The Change(s)?
Everyone use MQTT.

kozlovic:
@ianzhang1988 Correct, currently you cannot set the number of replicas, instead, the server determines the number based on the size of the cluster (number of routes) but cap it at 3.

We are planning at adding a `Replicas` option in the MQTT configuration block.

How many replicas are you planning on running though?  Also, I noticed that you are having quite a lot of consumers. NATS currently is a bit limited in horizontal scalability because of the number of RAFT groups (one per asset - an asset being a stream or a consumer). So the more consumers you have, the more basic RAFT traffic will increase, potentially more than actual data flow.
[...]

PR #3130, opened 2022-05-17
[ADDED] MQTT: Stream/Consumer replica count override

kozlovic:
Ability to override the stream and consumers replica count, which is by default
determined based on the cluster size.

```
mqtt {
  port: 1883
  stream_replicas: 5
  consumer_replicas: 1
}
```

The above would allow *new* MQTT streams to be created with a replicas
factor of 5 (it will be an error if the cluster does not have that
many nodes, and error will occur at runtime when the first client
on a given account connects), and new consumers would be R=1.

The MQTT existing streams/consumers for an account are not modified.

The stream_replicas can also obviously be reduced to 1 for a cluster
of 3 nodes if one desire to have those streams as R=1.

A value of 0 or negative is considered letting the server pick
the value (from 1 to 3 depending on standalone/cluster size).

There is another property that allows the consumers to be created
with memory storage instead of file:
```
mqtt {
  ..
  consumer_memory_storage: true
}
```

Those new settings are global and apply to new streams/consumers
only.

Related to https://github.com/nats-io/nats-server/issues/3116

### DIFF
--- a/server/mqtt.go
+++ b/server/mqtt.go
@@ -3601,8 +3601,9 @@ func (sess *mqttSession) processJSConsumer(c *client, subject, sid string,
 			FilterSubject:  mqttStreamSubjectPrefix + subject,
 			AckWait:        ackWait,
 			MaxAckPending:  maxAckPending,
+			MemoryStorage:  opts.MQTT.ConsumerMemoryStorage,
 		}
-		if r := c.srv.getOpts().MQTT.ConsumerReplicas; r > 0 {
+		if r := opts.MQTT.ConsumerReplicas; r > 0 {
 			cc.Replicas = r
 		}
 		if err := sess.createConsumer(cc); err != nil {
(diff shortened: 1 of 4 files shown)


server/mqtt.go
```diff
@@ -1037,49 +1039,88 @@ func (s *Server) mqttCreateAccountSessionManager(acc *Account, quitCh chan struc
...
+		}
+		if opts.MQTT.StreamReplicas == 0 {
+			return si, nil
+		}
+		sr := 1
+		if si.Cluster != nil {
+			sr += len(si.Cluster.Replicas)
+		}
+		if replicas != sr {
+			s.Warnf("account %q: %s stream replicas is %v, which does not match configured value of %v",
```
derekcollison, line 1058: Maybe use syntax we have elsewhere acc -> stream IIRC.
Also make sure log says MQTT since this is MQTT specific.

derekcollison (changes requested):

server/mqtt.go
kozlovic: Done

derekcollison (approved):
LGTM

derekcollison (approved):
LGTM

derekcollison (approved):
LGTM

PR #3131, opened 2022-05-18
Make sure consumer store is memory based when selected.

derekcollison:

/cc @nats-io/core

### DIFF
--- a/server/consumer.go
+++ b/server/consumer.go
@@ -694,7 +694,12 @@ func (mset *stream) addConsumerWithAssignment(config *ConsumerConfig, oname stri
 	o.selectStartingSeqNo()
 
 	if !config.Direct {
-		store, err := mset.store.ConsumerStore(o.name, config)
+		streamStore := mset.store
+		// We now allow overrides from a stream being a filestore type and forcing a consumer to be memory store.
+		if config.MemoryStorage && streamStore.Type() == FileStorage {
+			streamStore = emptyMemStore
+		}
+		store, err := streamStore.ConsumerStore(o.name, config)
 		if err != nil {
 			mset.mu.Unlock()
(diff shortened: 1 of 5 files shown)


server/consumer.go
```diff
@@ -694,7 +694,12 @@ func (mset *stream) addConsumerWithAssignment(config *ConsumerConfig, oname stri
 	o.selectStartingSeqNo()
 
 	if !config.Direct {
-		store, err := mset.store.ConsumerStore(o.name, config)
+		streamStore := mset.store
+		// We now allow overrides from a stream being a filestore type and forcing a consumer to be memory store.
+		if config.MemoryStorage && streamStore.Type() == FileStorage {
+			streamStore = emptyMemStore
```
kozlovic, line 700: This seem to be a "shared" mem store that all consumers from any stream will share. Is that right? Is there code that may go from consumer's store back to that one and this is going to cause a mess? Meaning emptyMemStore.ConsumerStore() will return an instance of consumerMemStore but that one will reference the "global" emptyMemStore instance.. All consumerMemStore APIs (UpdateDelivered, UpdateAcks, etc..) will work on the same stream's memory store...

derekcollison: Only thing I saw was the consumer count being added and deleted, but they would all be shared there.

kozlovic: Yes, that's what I find out after looking more into it. So I guess for now this is safe. Alternatively, streams that are file-based could hold a new field that would be an empty memory store (created only when the file-based stream is created). Then here we would assign mset.memStore or something like that.

kozlovic (approved):
LGTM

kozlovic (approved):
LGTM

PR #3132, opened 2022-05-18
With use cases bringing us more data I wanted to suggest these changes.

derekcollison:
With inlining election timeout updates we double the lock contention and most likely introduced head of line issues for routes under heavy load.
Also slowing down heartbeats with so many assets being deployed in our user ecosystem, also moved the normal follower to candidate timing further out, similar to the lost quorum.
Note that the happy path transfer will still be very quick.

/cc @nats-io/core


kozlovic:
Between 15 and 20%.

PR #3133, opened 2022-05-18
[ADDED] Monitoring: Routez's individual route has now more info

kozlovic:
Added Start, LastActivity, Uptime and Idle that we normally have
in a Connz for non route connections. This info can be useful
to determine if a route is recent, etc..


derekcollison (approved):
LGTM! This will be helpful.

PR #3135, opened 2022-05-18
Speed up raft for tests

derekcollison:

/cc @nats-io/core

### DIFF
--- a/server/norace_test.go
+++ b/server/norace_test.go
@@ -3746,10 +3746,15 @@ func TestNoRaceJetStreamClusterMaxConsumersAndDirect(t *testing.T) {
 // Make sure when we try to hard reset a stream state in a cluster that we also re-create the consumers.
 func TestNoRaceJetStreamClusterStreamReset(t *testing.T) {
 	// Speed up raft
+	omin, omax, ohb := minElectionTimeout, maxElectionTimeout, hbInterval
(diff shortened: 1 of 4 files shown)


server/norace_test.go
```diff
@@ -3746,10 +3746,15 @@ func TestNoRaceJetStreamClusterMaxConsumersAndDirect(t *testing.T) {
 // Make sure when we try to hard reset a stream state in a cluster that we also re-create the consumers.
 func TestNoRaceJetStreamClusterStreamReset(t *testing.T) {
 	// Speed up raft
+	omin, omax, ohb := minElectionTimeout, maxElectionTimeout, hbInterval
 	minElectionTimeout = 250 * time.Millisecond
 	maxElectionTimeout = time.Second
 	hbInterval = 50 * time.Millisecond
-	defer setDefaultRaftTimeouts()
+	defer func() {
```
kozlovic, line 3753: Fine like that. If we end-up having to do that more often in the future (or change more than those 3), then we will make setting the raft values a function that will be invoked in the test init() and called as a defer here. But for now, this is ok.

kozlovic (approved):
LGTM

server/norace_test.go
derekcollison: Ah gotcha.

PR #3137, opened 2022-05-19
Consumer's num pending can now rely on the stream's store

derekcollison:
We used to maintain a separate state during runtime which could be wrong under certain conditions.

/cc @nats-io/core

### DIFF
--- a/server/stream.go
+++ b/server/stream.go
@@ -3570,18 +3570,18 @@ func (mset *stream) processJetStreamMsg(subject, reply string, hdr, msg []byte, 
 	}
 
 	if err == nil && seq > 0 && numConsumers > 0 {
-		mset.mu.Lock()
+		mset.mu.RLock()
 		for _, o := range mset.consumers {
 			o.mu.Lock()
-			if o.isLeader() {
-				if seq > o.lsgap && o.isFilteredMatch(subject) {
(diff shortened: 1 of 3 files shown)


server/stream.go
```diff
@@ -3570,18 +3570,15 @@ func (mset *stream) processJetStreamMsg(subject, reply string, hdr, msg []byte,
...
 	if err == nil && seq > 0 && numConsumers > 0 {
-		mset.mu.Lock()
+		mset.mu.RLock()
 		for _, o := range mset.consumers {
 			o.mu.Lock()
-			if o.isLeader() {
-				if seq > o.lsgap && o.isFilteredMatch(subject) {
-					o.sgap++
-				}
+			if o.isLeader() && o.isFilteredMatch(subject) {
```
kozlovic, line 3576: We used to signal if o.isLeader() was true, but now will only if leader and isFilteredMatch(). Is that ok?

derekcollison: We would signal consumers that might not match, which would not affect behavior, but would wake them up and they would go back into waitForMsgs().

kozlovic: Gotcha.

kozlovic (approved):
LGTM

PR #3142, opened 2022-05-23
[FIXED] Fast routed JetStream API requests were dropped

kozlovic:
If a JS API request is received from a non client connection, it
was processed in its own go routine. To reduce the number of
such go routine, we were limiting the number of outstanding routines
to 4096. However, in some situations, it was possible to issue
many requests at the same time that would then cause those requests
to be dropped.

(an example was an MQTT benchmark tool that would create 5000
sessions, each with one QoS1 R1 consumer (with the use of consumer_replicas=1).
On abrupt exit of the tool, the consumers and their sessions needed
to be deleted. Since would cause fast incoming delete consumer requests
which would cause the original code to drop some of them)

### DIFF
--- a/server/jetstream_api.go
+++ b/server/jetstream_api.go
@@ -649,11 +649,18 @@ type JSApiStreamTemplateNamesResponse struct {
 
 const JSApiStreamTemplateNamesResponseType = "io.nats.jetstream.api.v1.stream_template_names_response"
 
-// Default max API calls outstanding.
-const defaultMaxJSApiOut = int64(4096)
-
-// Max API calls outstanding.
-var maxJSApiOut = defaultMaxJSApiOut
+// Structure that holds state for a JetStream API request that is processed
+// in a separate long-lived go routine. This is to avoid possibly blocking
+// ROUTE and GATEWAY connections.
+type jsAPIRoutedReq struct {
+	jsub    *subscription
+	sub     *subscription
+	acc     *Account
+	subject string
+	reply   string
+	msg     []byte
+	pa      pubArg
+}
 
 func (js *jetStream) apiDispatch(sub *subscription, c *client, acc *Account, subject, reply string, rmsg []byte) {
 	// No lock needed, those are immutable.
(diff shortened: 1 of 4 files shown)


server/jetstream_api.go
```diff
@@ -687,36 +694,36 @@ func (js *jetStream) apiDispatch(sub *subscription, c *client, acc *Account, sub
...
-		jsub.icb(sub, client, acc, subject, reply, rmsg)
-		atomic.AddInt64(&js.apiInflight, -1)
-	}()
+	for {
+		select {
+		case <-queue.ch:
+			reqs := queue.pop()
+			for _, req := range reqs {
+				r := req.(*jsAPIRoutedReq)
+				client.pa = r.pa
```
kozlovic, line 719: @derekcollison Want to bring to your attention that I am reusing a client that is created at the top of this function (we used to create one per request). From what I understand only the pubArg were really needed, so I think it is ok. If if there is a risk that the client's struct is used/modified by the icb() call, then let me know and we can create the client here for each request.

server/jetstream_api.go
```diff
@@ -649,11 +649,18 @@ type JSApiStreamTemplateNamesResponse struct {
...
 
-// Default max API calls outstanding.
-const defaultMaxJSApiOut = int64(4096)
-
-// Max API calls outstanding.
-var maxJSApiOut = defaultMaxJSApiOut
+// Structure that holds state for a JetStream API request that is processed
+// in a separate long-lived go routine. This is to avoid possibly blocking
+// ROUTE and GATEWAY connections.
+type jsAPIRoutedReq struct {
```
derekcollison, line 655: Should we pool these?

server/jetstream_api.go
derekcollison: Do we need to copy the pubArgs? Can they easily be overwritten?

I think its ok to re-use same client struct but we might need to deep copy pubArg.

derekcollison (changes requested):

server/jetstream_api.go
kozlovic: The code that this replace was doing a copy (not deep copy) of c.pa, however here, I agree that this is causing the copy to be done twice.

Regarding "we might need to deep copy pubArg": that was my concern when having the original code, see comment here: https://github.com/nats-io/nats-server/pull/1953#discussion_r586050261.

So now I am confused on what we should be doing here...

derekcollison: Actually I think we already copy pubargs. So should be ok.

kozlovic: Oh yes, I did not understand what you were asking here. The copy happens when pushing into the queue here: https://github.com/nats-io/nats-server/pull/3142/files#diff-9c075c47b0690a821ad35caa26308b69adeb132042f062eef75513db2a26638aR702

derekcollison: I mean what it was referring to, not the []byte structs themselves.

kozlovic: Since pubArgs is a struct, we already make a copy here:
```
s.jsAPIRoutedReqs.push(&jsAPIRoutedReq{jsub, sub, acc, subject, reply, copyBytes(rmsg), c.pa})
```
And, like I said, that bad part is that it is again making a copy (unfortunately) when doing:
```
client.pa = r.pa
```

derekcollison: ok gotcha.

derekcollison (approved):
LGTM - We can add pooling and optimize paArg copy later.

PR #3143, opened 2022-05-23
[FIXED] JetStream: consumer with deliver new may miss messages

derekcollison:
This could happen when a consumer had not sent anything to the
attached NATS subscription and there was a consumer leader
step down or server restart.

Stabilized code to set starting sequence numbers on true start versus recovered state.

/cc @nats-io/core


derekcollison:
Not sure what git fu you would want to do here on rebase. Just lmk.

kozlovic (approved):
LGTM

PR #3144, opened 2022-05-23
[FIXED] MQTT: Errors deleting consumers will now prevent deletion of session

kozlovic:
When there was a failure to delete a QoS1 consumer, the session
would still be deleted, which would cause orphaned consumers.

In case of error, the session record will not be deleted, which means
that it is still possible to restart the session and then close
it (with the clean flag).

Relates to #3116


derekcollison (approved):
LGTM

PR #3148, opened 2022-05-24
[FIXED] Consumer NumPending incorrect on certain stream setups.

derekcollison:
When a consumer was created that filtered out partial sets of messages with deliver last by subject, and this was on a stream that had max message per subject set and > 1 and the matching filtered messages did have some subjects > 1 the num pending was incorrect.

This could happen with KV streams and history > 1 with duplicates and could result in not getting the initial data set done signal for a watcher.

/cc @nats-io/core

### DIFF
--- a/server/consumer.go
+++ b/server/consumer.go
@@ -224,7 +224,7 @@ type consumer struct {
 	adflr             uint64
 	asflr             uint64
 	npc               uint64
-	npsm              uint64
+	npcm              uint64
 	dsubj             string
 	qgroup            string
 	lss               *lastSeqSkipList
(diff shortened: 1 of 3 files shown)


aricart:
I did a quick test, and it looks like tombstones are not getting reported, so DEL on a watch are missing.

aricart:
The test I have simply creates a KV, adds 4 values, then deletes some - when I ask for the keys, deleted keys are missing.

derekcollison:
Travis was not working so was a bit in flight. I just did a force push, try that one.

kozlovic (approved):
LGTM. Ran tests locally and they pass.. but check with Alberto's test that seem to be reported a change in behavior.

derekcollison:
I think he got caught in between fixes, @aricart can you re-run yours now?

aricart:
@derekcollison seems to work!

aricart:
@derekcollison all JavaScript tests pass.

PR #3149, opened 2022-05-25
[FIXED] Spurious pull consumer 408s under load 

derekcollison:
Under load we could send 408s and invalidate requests that should still be valid when multiple parallel requests are inbound.
Also move processing of Acks to its own Go routine to avoid some contention during delivery of msgs.

/cc @nats-io/core


kozlovic (approved):
LGTM

PR #3150, opened 2022-05-25
[FIXED] MQTT: Sessions with same ID in different domains were considered duplicates

kozlovic:
There is a mechanism to detect if a connection somewhere in the
cluster is using the session ID of an existing one, and if so,
close one as a duplicate.
However, when different domains are used, they should not be considered
duplicates.

### DIFF
--- a/server/mqtt.go
+++ b/server/mqtt.go
@@ -1546,8 +1546,19 @@ func (as *mqttAccountSessionManager) processSessionPersist(_ *subscription, pc *
 	if err := par.Error; err != nil {
 		return
 	}
-	// We would need to lookup the message that that is a request/reply
-	// that we can do in place here. So move that to a long-running routine
+	as.mu.RLock()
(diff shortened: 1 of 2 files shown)


server/mqtt.go
```diff
@@ -1546,8 +1546,19 @@ func (as *mqttAccountSessionManager) processSessionPersist(_ *subscription, pc *
 	if err := par.Error; err != nil {
 		return
 	}
-	// We would need to lookup the message that that is a request/reply
-	// that we can do in place here. So move that to a long-running routine
+	as.mu.RLock()
+	// Note that as.domainTk includes a terminal '.', so strip to compare to PubAck.Domain.
+	dl := len(as.domainTk)
```
derekcollison, line 1551: Feels like we can do this a bit different..

derekcollison: maybe strings.HasPrefix()?

PR #3154, opened 2022-05-27
Export the subject transformer

ripienaar:
This exports the one key function of the subject transformer
allowing external tools to be written to test mappings are
valid and see how they would interact without the hassle of
configuring a serrver

The APIs are specifically marked as being unsupported and
having kept the transform struct itself unexported one can
not cast from the interface to the real implementation

/cc @nats-io/core

### DIFF
--- a/server/accounts.go
+++ b/server/accounts.go
@@ -4225,6 +4225,20 @@ func placeHolderIndex(token string) ([]int, int32, error) {
 	return []int{-1}, -1, nil
 }
 
+// SubjectTransformer transforms subjects using mappings
+//
+// This API is not part of the public API and not subject to SemVer protections
+type SubjectTransformer interface {
+	TransformSubject(string) (string, error)
+}
+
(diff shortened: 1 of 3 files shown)


ripienaar:
As discussed with Ivan and Derek, I am proposing we open a little bit of the transform API for the `nats` CLI use, you can see below a rough implementation of what I want to achieve.

@jnmoyne and I have had a quick chat and there's a lot of improvements to be done to the code - error reporting especially makes using it really hard which is the reason for adding this tool in the CLI.  So I know you want me to copy/paste the code @derekcollison as you thought this never changes, but we definitely have some plans to change things I think and keeping things in sync would be really hard.

ripienaar:
Screenshot 2022-05-27 at 10 27 02

derekcollison (commented):
Still prefer this simply be lifted from here into NATS cli.

PR #3156, opened 2022-05-30
[FIXED] Suppress some JetStream advisories

### LINKED ISSUE #3155: Advisory updates/refinements
When a server starts up we see a lot of advisories as things recover from disk, it would be good to stop these initial advisories during recovery:

```
20:46:58 [JS API] $JS.API.CONSUMER.DURABLE.CREATE.ORDERS_54.C2 one @ one
20:46:58 [Stream Modify] KV_X
20:46:58 [JS API] $JS.API.CONSUMER.DURABLE.CREATE.ORDERS_87.C4 one @ one
```

One per every consumer on the node and the stream modify ones are streams that are R1 hosted on this server that restarted

Additionally I think the stream msg get API should not publish adivsories, in the realm of KV its essentially equivelant to a Pull request - on a busy network we see many many KV GETs each would raise this advisory, especially bad since the advisory would also include the full content of the data that was fetched.

So I propose we dont raise advisories for `JS.API.STREAM.MSG.GET.KV_X`

derekcollison:
Suppress consumer and stream advisories on server restart and any direct stream get message advisory.

Resolves #3155 

/cc @nats-io/core

### DIFF
--- a/server/jetstream_api.go
+++ b/server/jetstream_api.go
@@ -2530,7 +2530,9 @@ func (s *Server) jsMsgGetRequest(sub *subscription, c *client, _ *Account, subje
 		Data:     sm.msg,
 		Time:     time.Unix(0, sm.ts).UTC(),
 	}
-	s.sendAPIResponse(ci, acc, subject, reply, string(msg), s.jsonResponse(resp))
+
+	// Don't send response through API layer for this call.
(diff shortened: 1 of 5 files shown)


ripienaar (approved):
LGTM

PR #3158, opened 2022-05-30
[IMPROVED] Fast and Direct access to stream messages.

derekcollison:
Stream get for KV was going through the API layer, but with the popularity, needed a more performant, lighter weight and direct approach. This skips the API layer, which currently still floods and does a bunch of API layer stats and client info tracking.

We could look into using a DQ for R>1 as well, or suggest that for KV just use mirrors for horizontal scalability for reads.

/cc @nats-io/core

### DIFF
--- a/server/jetstream_api.go
+++ b/server/jetstream_api.go
@@ -118,6 +118,12 @@ const (
 	JSApiMsgGet  = "$JS.API.STREAM.MSG.GET.*"
 	JSApiMsgGetT = "$JS.API.STREAM.MSG.GET.%s"
 
+	// JSDirectMsgGet is the template for non-api layer direct requests for a message by its stream sequence number or last by subject.
+	// Will return the message similar to how a consumer receives the message, no JSON processing.
+	// If the message can not be found we will use a status header of 404. If the stream does not exist the client will get a no-responders or timeout.
+	JSDirectMsgGet  = "$JS.DS.GET.*"
+	JSDirectMsgGetT = "$JS.DS.GET.%s"
+
 	// JSApiConsumerCreate is the endpoint to create ephemeral consumers for streams.
 	// Will return JSON response.
 	JSApiConsumerCreate  = "$JS.API.CONSUMER.CREATE.*"
--- a/server/stream.go
+++ b/server/stream.go
@@ -267,10 +267,12 @@ const (
 	JSResponseType        = "Nats-Response-Type"
 )
 
-// Headers for republished messages.
+// Headers for republished messages and direct gets.
 const (
 	JSStream       = "Nats-Stream"
 	JSSequence     = "Nats-Sequence"
+	JSTimeStamp    = "Nats-Time-Stamp"
+	JSSubject      = "Nats-Subject"
 	JSLastSequence = "Nats-Last-Sequence"
 )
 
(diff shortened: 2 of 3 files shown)


server/stream.go
```diff
@@ -72,6 +72,8 @@ type StreamConfig struct {
 	// AllowRollup allows messages to be placed into the system and purge
 	// all older messages using a special msg header.
 	AllowRollup bool `json:"allow_rollup_hdrs"`
+	// Allow higher peformance, direct access to get individual messages.
+	AllowDirect bool `json:"allow_direct,omitempty"`
```
aricart, line 76: so this option makes it so that KV would have to examine not only the server version but the stream info to know if this feature is enabled. Why does this feature need stream opt-in, but the other doesn't?

ripienaar: You would only get true here on streams where its opted in AND the server is the correct version.

aricart: correct it is actually simpler, but the point is that we have to enable some option to enable the API, but we didn't for the other, so what is so special about this one?

ripienaar: From discussions with Derek - but not yet from reading the whole PR - this sets up a new subject where you can request data bypassing various expensive encodings, audits and things.  It's opt in so these things dont listen on all servers etc.

I have concerns about upgrading to the new model and breaking old clients - KV is now WIDELY used - so this is opt in behavior and behavior appropriately versions client libraries can detect and choose to use if they wish. There are also concerns about breaking existing deployments wrt we are adding new subjects and their ACLs might not cater for them meaning the client either breaks or we now have a new way to exfil data users dont know about.

Feature is there to make N million clients work rather than your garden variety day to day, and will have complexities cross accounts/leafnodes etc being on a subject outside of `$JS.API.>` (all the domain denies etc).  

So given all above, its opt-in for the moment.

derekcollison: That is correct, only new servers will return stream info with this set to true. Clients will only be able to use direct method if this is set to true, otherwise use current behavior through $JS.API route.

server/jetstream_api.go
```diff
@@ -118,6 +118,12 @@ const (
 	JSApiMsgGet  = "$JS.API.STREAM.MSG.GET.*"
 	JSApiMsgGetT = "$JS.API.STREAM.MSG.GET.%s"
 
+	// JSDirectMsgGet is the template for non-api layer direct requests for a message by its stream sequence number or last by subject.
+	// Will return the message similar to how a consumer receives the message, no JSON processing.
+	// If the message can not be found we will use a status header of 404. If the stream does not exist the client will get a no-responders or timeout.
+	JSDirectMsgGet  = "$JS.DS.GET.*"
```
aricart, line 124: How does this work with API prefixes? Currently if using prefixes, `$JS.API` is replaced with whatever prefix is provided.

aricart: @derekcollison The current implementation doesn't correctly honor mux requests, the message is mapped to the subscription, but the subject is not including the inbox - go client I believe currently has an assumption that if there's only one mux it uses that, but that doesn't work if there's any other kind of request-reply going on with the client.

< PUB $JS.DS.GET.KV_a _INBOX.4APLG7K94VJXE3ZO5ZTI72.4APLG7K94VJXE3ZO5ZTIZI 26␍␊{"last_by_subj":"$KV.a.a"}␍␊
> HMSG $KV.a.a 1  49 54␍␊NATS/1.0␍␊Nats-Stream: KV_a␍␊Nats-Sequence: 1␍␊␍␊hello␍␊

aricart: The current API for KV also provides the time when the entry was written which may or not be interesting to a client, the direct mechanism needs to provide this possibly in the headers.

derekcollison: This is at the account level, not $SYS level so no API rewrite should be needed. Just like sending to a subject to be placed into a stream.

derekcollison: Yes mux operations would need to be considered. We could make this mux safe and place original subject in header and reply with INBOX.

derekcollison: I could add in Timestamp as well.

aricart: if you add that, we can convert it and there's no difference to the current client contract.

derekcollison: ok done.

server/jetstream_api.go
```diff
@@ -118,6 +118,12 @@ const (
...
+	JSDirectMsgGetT = "$JS.DS.GET.%s"
```
ripienaar, line 125: doesnt domains allow 2 domains to host the same name stream? But the core subject space is shared by all domains in an account? So just saying stream name could lead to issues right?

Not that up to speed with domains so might just be confused

derekcollison: Fair question.

derekcollison:
Also we should decide if we want that subscriber to be a queue subscriber and make cure clients detect write them immediate read. Trivial change to enable and we could even monitor and drop peers that are too far behind from the group.

aricart:
@derekcollison This works properly with mux subscriptions, it would be awesome if regular pull subscriber requests also honored the same behaviour. As some of the operations on the request status can be greatly simplified if this is the case.

derekcollison:
We need to do a direct access for pull requests as well, agree.

server/jetstream_api.go
derekcollison: This needs to be thought through a bit more, but even now if those streams have the same subjects they will be placed into both.

I also think we should place the direct get in a DQ group by default such that replicas can share the load.

This is experimental for now as we settle in the API and subjects and mechanics. Will merge.

ripienaar: Hmm the domain scenario where 2 streams have the same name and subject space - a monumentally bad idea - isn’t about replicas it’s about like regional to hub migration/consolidation. 

So it’s not about replicas sharing the load. Data might just not be there. 

Wirh this I would say direct must not be default behaviour but sometning you opt in when you really grok your setup?

derekcollison: Direct is opt-in, agree you don't want to cross over with Domains, just saying that issues already is present with ingest subjects.

PR #3159, opened 2022-06-01
[FIXED] When stores and load for last for subject where concurrent and compet…

derekcollison:
When stores and loads for last for a subject where concurrent and competing for the same msg, we could fail to retrieve a newly placed message. We were not holding the lock so last seq calculations could be stale when we actually fetched the message since another could have placed a new value and updated.

This is from a Marco torture test.

/cc @nats-io/core


ripienaar (approved):
LGTM

PR #3160, opened 2022-06-01
[FIXED] Do not send consumer API advisories on restart.

derekcollison:
Make sure we do not resend responses on restart for certain clustered consumers after a meta snapshot.

/cc @nats-io/core


ripienaar (approved):
Lgtm

PR #3162, opened 2022-06-02
Mark meta recovering state and use to suppress api responses and api audits during restarts.

derekcollison:

/cc @nats-io/core


ripienaar (approved):
Lgtm

PR #3163, opened 2022-06-02
export the correct subject transformer subject

ripienaar:
While the TransformSubject function was doing the right
thing it did not match first and so would panic for subjects
that do not match the mapping.

The map function does the right thing so this is a more
appropriate function to export.

This undoes the exporting of unsafe TransformSubject and
exports the safer Match instead.

/cc @nats-io/core


ripienaar:
A follow up to https://github.com/nats-io/nats-server/pull/3154 that just exports a more safe function instead of one that panics

derekcollison (approved):
LGTM

PR #3168, opened 2022-06-05
[IMPROVED] Loaded server and low on resources like FDs.

### LINKED ISSUE #3167: Under heavy load a stream with lots of R>1 consumer is hard to delete
When a system is under load and behaving incorrectly, a stream delete should act like a big hammer and just work.
Currently the system sometimes does not respond, or responds slowly, and many times does not delete the stream or help stabilize the system.

2.8.4

derekcollison:
This is to help improve when a server/system is under heavy load with low resources, CPU/MEM or FDs.
If the user wants to try to stabilize by deleting a stream possibly with lots of consumers.

Resolves #3167 

/cc @nats-io/core

### DIFF
--- a/server/jetstream_cluster.go
+++ b/server/jetstream_cluster.go
@@ -2753,7 +2753,10 @@ func (js *jetStream) processClusterCreateStream(acc *Account, sa *streamAssignme
 
 	// This is an error condition.
 	if err != nil {
-		s.Warnf("Stream create failed for '%s > %s': %v", sa.Client.serviceAccount(), sa.Config.Name, err)
+		if IsNatsErr(err, JSStreamStoreFailedF) {
+			s.Warnf("Stream create failed for '%s > %s': %v", sa.Client.serviceAccount(), sa.Config.Name, err)
+			err = errStreamStoreFailed
+		}
 		js.mu.Lock()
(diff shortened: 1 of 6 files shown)


server/jetstream_cluster.go
```diff
@@ -4749,11 +4756,6 @@ func (s *Server) jsClusteredStreamDeleteRequest(ci *ClientInfo, acc *Account, st
 		s.sendAPIErrResponse(ci, acc, subject, reply, string(rmsg), s.jsonResponse(&resp))
 		return
 	}
-	// Remove any remaining consumers as well.
-	for _, ca := range osa.consumers {
-		ca.Reply, ca.State = _EMPTY_, nil
-		cc.meta.Propose(encodeDeleteConsumerAssignment(ca))
```
ripienaar, line 4755: what will remove them after this code is removed? Will they be cleaned up elsewhere?

derekcollison: Those assignments, the consumer assignments, are attached to the stream. So once the stream goes away those go with it. I also double checked the consumers themselves were stopped and deleted and they were.

ripienaar (approved):
LGTM

PR #3174, opened 2022-06-07
add global log to decoupling log from server

### LINKED ISSUE #3173: Add a global logger in server package to decoupling logger from server
## Feature Request
Currently, if nats-server wants to log something, it must receive a server struct. But actually the function shouldn't bind to type `Server`. 
So it's better to add a logger which pointers to the logger hold in server. This way makes the code clean and decreases the coupling.

#### Use Case:
The origin which wants to log an information must has a Server receiver, but actually the `Server` struct shouldn't have this function:
```go
func (s *Server) getNonLocalIPsIfHostIsIPAny(host string, all bool) (bool, []string, error) {
	ip := net.ParseIP(host)
	// If this is not an IP, we are done
	if ip == nil {
		return false, nil, nil
	}
	// If this is not 0.0.0.0 or :: we have nothing to do.
	if !ip.IsUnspecified() {
		return false, nil, nil
	}
	s.Debugf("Get non local IPs for %q", host)
	var ips []string
	ifaces, _ := net.Interfaces()
	for _, i := range ifaces {
		addrs, _ := i.Addrs()
		for _, addr := range addrs {
			switch v := addr.(type) {
			case *net.IPNet:
				ip = v.IP
			case *net.IPAddr:
				ip = v.IP
			}
			ipStr := ip.String()
			// Skip non global unicast addresses
			if !ip.IsGlobalUnicast() || ip.IsUnspecified() {
				ip = nil
				continue
			}
			s.Debugf("  ip=%s", ipStr)
			ips = append(ips, ipStr)
			if !all {
				break
			}
		}
	}
	return true, ips, nil
}
```

Instead, it is better to use the global logger to log as following and it's cleaner code.
```go
func getNonLocalIPsIfHostIsIPAny(host string, all bool) (bool, []string, error) {
	ip := net.ParseIP(host)
	// If this is not an IP, we are done
	if ip == nil {
		return false, nil, nil
	}
	// If this is not 0.0.0.0 or :: we have nothing to do.
	if !ip.IsUnspecified() {
		return false, nil, nil
	}
	logging.Debugf("Get non local IPs for %q", host)
	var ips []string
	ifaces, _ := net.Interfaces()
	for _, i := range ifaces {
		addrs, _ := i.Addrs()
		for _, addr := range addrs {
			switch v := addr.(type) {
			case *net.IPNet:
				ip = v.IP
			case *net.IPAddr:
				ip = v.IP
			}
			ipStr := ip.String()
			// Skip non global unicast addresses
			if !ip.IsGlobalUnicast() || ip.IsUnspecified() {
				ip = nil
				continue
			}
			logging.Debugf("  ip=%s", ipStr)
			ips = append(ips, ipStr)
			if !all {
				break
			}
		}
	}
	return true, ips, nil
}
```

#### Proposed Change:
Add a global logger in `log.go` file and provide an interface for `Server` struct to enable it. Provides all of log interfaces expect the `RateLimitWarnf` function.

- Developers who contributes to the nats-server could easily log.
- Make nats-server code cleaner.

xieyuschen:
- [X] Link to issue, e.g. `Resolves #3173 `
 - [X] Tests added

Resolves #

 - Add global logger 

/cc @nats-io/core

### DIFF
--- a/server/log.go
+++ b/server/log.go
@@ -17,11 +17,22 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"sync"
 	"sync/atomic"
 	"time"
 
 	srvlog "github.com/nats-io/nats-server/v2/logger"
 )
+
+var logging *Logging = &Logging{}
+
+type Logging struct {
+	sync.RWMutex
+	logger      Logger
+	trace       int32
+	debug       int32
+	traceSysAcc int32
+}
 
 // Logger interface of the NATS Server
 type Logger interface {
@@ -43,6 +54,87 @@ type Logger interface {
 
 	// Log a trace statement
 	Tracef(format string, v ...interface{})
+}
+
+// Noticef logs a notice statement
+func (l *Logging) Noticef(format string, v ...interface{}) {
+	l.executeLogCall(func(logger Logger, format string, v ...interface{}) {
+		logger.Noticef(format, v...)
+	}, format, v...)
+}
+
+// Errorf logs an error
+func (l *Logging) Errorf(format string, v ...interface{}) {
+	l.executeLogCall(func(logger Logger, format string, v ...interface{}) {
+		logger.Errorf(format, v...)
+	}, format, v...)
+}
+
+// Errors logs an error with a scope
+func (l *Logging) Errors(scope interface{}, e error) {
+	l.executeLogCall(func(logger Logger, format string, v ...interface{}) {
+		logger.Errorf(format, v...)
+	}, "%s - %s", scope, UnpackIfErrorCtx(e))
+}
+
+// Errorc logs an error with a context
+func (l *Logging) Errorc(ctx string, e error) {
+	l.executeLogCall(func(logger Logger, format string, v ...interface{}) {
+		logger.Errorf(format, v...)
+	}, "%s: %s", ctx, UnpackIfErrorCtx(e))
+}
+
+// Errorsc logs an error with a scope and context
+func (l *Logging) Errorsc(scope interface{}, ctx string, e error) {
+	l.executeLogCall(func(logger Logger, format string, v ...interface{}) {
+		logger.Errorf(format, v...)
+	}, "%s - %s: %s", scope, ctx, UnpackIfErrorCtx(e))
+}
+
+// Warnf logs a warning error
+func (l *Logging) Warnf(format string, v ...interface{}) {
+	l.executeLogCall(func(logger Logger, format string, v ...interface{}) {
+		logger.Warnf(format, v...)
+	}, format, v...)
+}
+
+// Fatalf logs a fatal error
+func (l *Logging) Fatalf(format string, v ...interface{}) {
+	l.executeLogCall(func(logger Logger, format string, v ...interface{}) {
+		logger.Fatalf(format, v...)
+	}, format, v...)
+}
+
+// Debugf logs a debug statement
+func (l *Logging) Debugf(format string, v ...interface{}) {
+	if atomic.LoadInt32(&l.debug) == 0 {
+		return
+	}
+
+	l.executeLogCall(func(logger Logger, format string, v ...interface{}) {
+		logger.Debugf(format, v...)
+	}, format, v...)
+}
+
+// Tracef logs a trace statement
+func (l *Logging) Tracef(format string, v ...interface{}) {
+	if atomic.LoadInt32(&l.trace) == 0 {
+		return
+	}
+
+	l.executeLogCall(func(logger Logger, format string, v ...interface{}) {
+		logger.Tracef(format, v...)
+	}, format, v...)
+}
+
+func (l *Logging) executeLogCall(f func(logger Logger, format string, v ...interface{}), format string, args ...interface{}) {
+	l.RLock()
+	defer l.RUnlock()
+	if l.logger == nil {
+		return
+	}
+
+	f(l.logger, format, args...)
 }
 
 // ConfigureLogger configures and sets the logger for the server.
@@ -90,7 +182,11 @@ func (s *Server) ConfigureLogger() {
 	s.SetLoggerV2(log, opts.Debug, opts.Trace, opts.TraceVerbose)
 }
 
-// Returns our current logger.
+func (s *Server) enableLogging() {
+	logging = &s.logging
+}
+
+// Logger Returns our current logger.
 func (s *Server) Logger() Logger {
 	s.logging.Lock()
 	defer s.logging.Unlock()
@@ -102,7 +198,7 @@ func (s *Server) SetLogger(logger Logger, debugFlag, traceFlag bool) {
 	s.SetLoggerV2(logger, debugFlag, traceFlag, false)
 }
 
-// SetLogger sets the logger of the server
+// SetLoggerV2 sets the logger of the server
 func (s *Server) SetLoggerV2(logger Logger, debugFlag, traceFlag, sysTrace bool) {
 	if debugFlag {
 		atomic.StoreInt32(&s.logging.debug, 1)
@@ -178,21 +274,21 @@ func (s *Server) Errorf(format string, v ...interface{}) {
 	}, format, v...)
 }
 
-// Error logs an error with a scope
+// Errors logs an error with a scope
 func (s *Server) Errors(scope interface{}, e error) {
 	s.executeLogCall(func(logger Logger, format string, v ...interface{}) {
 		logger.Errorf(format, v...)
 	}, "%s - %s", scope, UnpackIfErrorCtx(e))
 }
 
-// Error logs an error with a context
+// Errorc logs an error with a context
 func (s *Server) Errorc(ctx string, e error) {
 	s.executeLogCall(func(logger Logger, format string, v ...interface{}) {
 		logger.Errorf(format, v...)
 	}, "%s: %s", ctx, UnpackIfErrorCtx(e))
 }
 
-// Error logs an error with a scope and context
+// Errorsc logs an error with a scope and context
 func (s *Server) Errorsc(scope interface{}, ctx string, e error) {
 	s.executeLogCall(func(logger Logger, format string, v ...interface{}) {
 		logger.Errorf(format, v...)
--- a/server/log_test.go
+++ b/server/log_test.go
@@ -352,3 +352,58 @@ func TestRemovePassFromTrace(t *testing.T) {
 		})
 	}
 }
+
+func TestSetGlobalLogger(t *testing.T) {
+	server := &Server{}
+	defer server.SetLogger(nil, false, false)
+	dl := &DummyLogger{}
+	server.SetLogger(dl, true, true)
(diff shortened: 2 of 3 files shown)


xieyuschen:
Hi @derekcollison @philpennock , could you review this PR?

derekcollison:
Plan on taking a closer look next week, thanks for your patience.

philpennock:
I took a look and it's nice clean code and fixes some cruft, so I like the quality of the PR; but: what is the problem with the logger being inside the server object?  Why move it to be a global instead?

Thanks.

derekcollison:
@xieyuschen where would this be useful beyond as it is now attached to a server instance?

Are you looking to run multiple servers with logging in the same process?

xieyuschen:
> I took a look and it's nice clean code and fixes some cruft, so I like the quality of the PR; but: what is the problem with the logger being inside the server object? Why move it to be a global instead?
> 
> Thanks.



## what is the problem with the logger being inside the server object?
Hi @philpennock , in my opinion, the logger inside the `server` causes code coupling as mentioned in #3173 . 
For example, those functions which shouldn't have a `Server` receiver need to have a `Server` receiver if they want to log something. 
I paste the content here to avoid page jumping.

> The origin which wants to log an information must has a Server receiver, but actually the Server struct shouldn't have this function. 
```go
func (s *Server) getNonLocalIPsIfHostIsIPAny(host string, all bool) (bool, []string, error) {
	ip := net.ParseIP(host)
	// If this is not an IP, we are done
	if ip == nil {
		return false, nil, nil
	}
	// If this is not 0.0.0.0 or :: we have nothing to do.
	if !ip.IsUnspecified() {
		return false, nil, nil
	}
	
	var ips []string
	ifaces, _ := net.Interfaces()
	for _, i := range ifaces {
		addrs, _ := i.Addrs()
		for _, addr := range addrs {
			switch v := addr.(type) {
			case *net.IPNet:
				ip = v.IP
			case *net.IPAddr:
				ip = v.IP
			}
			ipStr := ip.String()
			// Skip non global unicast addresses
			if !ip.IsGlobalUnicast() || ip.IsUnspecified() {
				ip = nil
				continue
			}
			s.Debugf("  ip=%s", ipStr)
			ips = append(ips, ipStr)
			if !all {
				break
			}
		}
	}
	return true, ips, nil
}
````
Instead, it is better to use the global logger to log as following and it's cleaner code. It's none of the server's businees to handle the function `getNonLocalIPsIfHostIsIPAny`. Note that the `logging` **is the global logger which binds to the one in server**.
```go
func getNonLocalIPsIfHostIsIPAny(host string, all bool) (bool, []string, error) {
	ip := net.ParseIP(host)
	// If this is not an IP, we are done
	if ip == nil {
		return false, nil, nil
	}
	// If this is not 0.0.0.0 or :: we have nothing to do.
	if !ip.IsUnspecified() {
		return false, nil, nil
	}
        
	logging.Debugf("Get non local IPs for %q", host)
	var ips []string
	ifaces, _ := net.Interfaces()
	for _, i := range ifaces {
		addrs, _ := i.Addrs()
		for _, addr := range addrs {
			switch v := addr.(type) {
			case *net.IPNet:
				ip = v.IP
			case *net.IPAddr:
				ip = v.IP
			}
			ipStr := ip.String()
			// Skip non global unicast addresses
			if !ip.IsGlobalUnicast() || ip.IsUnspecified() {
				ip = nil
				continue
			}
			logging.Debugf("  ip=%s", ipStr)
			ips = append(ips, ipStr)
			if !all {
				break
			}
		}
	}
	return true, ips, nil
}
```

## Why move it to be a global instead?
For now, there are two ways to provide logging to separate log and `server` object.
1. Encapsulate our own log package and **only expose the logger interface**  
In this way, the logger object is hidden in the package itself and user cannot visit it.

2. Use a global logger which references to the server logger.  
In this way, users can use this global logger to log instead of the logger inside the server. This way frees many functions that just want to use the logger inside the server and keeps the project stable.

The reason why I use the second one is that:
- the second **doesn't take a breaking change**, we can mark those server log functions as `deprecated` and upgrade them step by step.

ripienaar:
The server is embeddable in other go apps and each instance needs to be able to have independent logging configuration.

xieyuschen:
> @xieyuschen where would this be useful beyond as it is now attached to a server instance?
> 
> Are you looking to run multiple servers with logging in the same process?

# where would this be useful beyond as it is now attached to a server instance?
As the [comment](https://github.com/nats-io/nats-server/pull/3174#issuecomment-1152870942) replied to @philpennock , it decoupling the logger from the server object.

Nats can benefit as the following shows:
- decrease the functions which shouldn't belong to the server.
- decoupling the logger and server so that the logger could also be visited by the other directly, instead of must visit the server first.

xieyuschen:
> The server is embeddable in other go apps and each instance needs to be able to have independent logging configuration.

Hi @ripienaar , could you give me the link about such app?
Also, enabling the global logger doesn't affect each nats instance as **the global one is the same logger in the server it holds**. It just holds a reference instead of creating new one.
Thanks.

ripienaar:
If it’s global in the package and server instances share a reference to the same one then each server shares the same logging setup. 

There many examples - even in our tests - of the embedded pattern.

xieyuschen:
> If it’s global in the package and server instances share a reference to the same one then each server shares the same logging setup.
> 
> There many examples - even in our tests - of the embedded pattern.

The logger is first initialized in the server, and then it's enabled and assigned to the local variable `Logging`:
```go
package server
var logging Logging
// ignore some code lines
```
So there is no shared same logging between multiple servers. Each server has its own space and has its logger.

xieyuschen:
> If it’s global in the package and server instances share a reference to the same one then each server shares the same logging setup. 
> 
> 
> 
> There many examples - even in our tests - of the embedded pattern. 

Agree with you, the global variable cause shared log in many instances.
Btw as @derekcollison mentioned above, a nats server usually runs in a process so this change won't cause a bug.
How do you think about it?

ripienaar:
The nats server does usually run in a process, but that is not the only supported usage model, we should consider all usage models.

In this PR then the package level Logg variable is unused but you propose moving to a global logger in various place here, I would not consider such a move to be a good one and it appears in this PR that isnt whats happening as the variable is unused.

I would support extracing loggers out into a package and an interface so others can be provided as dependencies etc

xieyuschen:
> The nats server does usually run in a process, but that is not the only supported usage model, we should consider all usage models.
> 
> 
> 
> In this PR then the package level Logg variable is unused but you propose moving to a global logger in various place here, I would not consider such a move to be a good one and it appears in this PR that isnt whats happening as the variable is unused.
> 
> 
> 
> I would support extracing loggers out into a package and an interface so others can be provided as dependencies etc

In my opinion, we of course should consider all of the usages. But if we want to support the separated logger for multiple server instances, it's unavoidable to couple the server with those functions which needs to log something. 
I think this a trade-off, nats-server choose to bind many functions with the server. But to decouple, using a global logger is acceptable and we just need to tell user not to run multiple nats-servers in a process. 
We pay for that user could not run many servers in a process. In return we decrease the code coupling.
I think it's not so bad.

For unused Logg, there are so many such functions which couples log inside server. I need to confirm if this feature could be merged and then work for them.

Thanks

ripienaar:
> We just need to tell user not to run multiple nats-servers in a process. We pay for that user could not run many servers in a process. In return we decrease the code coupling. I think it's not so bad.

IMO we cannot just tell users not to use a major feature of the software because its slightly more convenient for us as developers or slightly cleaner code as a result.

xieyuschen:
> > We just need to tell user not to run multiple nats-servers in a process. We pay for that user could not run many servers in a process. In return we decrease the code coupling. I think it's not so bad.
> 
> IMO we cannot just tell users not to use a major feature of the software because its slightly more convenient for us as developers or slightly cleaner code as a result.

The `net/http` server just uses `log` library in golang directly, so if users would like to run many http server in a process, all logs from different server will use the same log.
It's very similiar in this PR. As a result, I think the concern rasied by @ripienaar is not seriours. We could handle a global logger s what the server in `net/http` does.

As this feature improves our code feature and doesn't cause a seriour bug for users, I think it's ok and no need to worry about too much.
We could wait @derekcollison @philpennock to finialize this PR.
THanks.

derekcollison:
I will take a deeper look next week. The majority of our JetStream tests utilize clusters and superclusters (multi-server) in the same test process. We also have tests that use the log output to validate the test. So I think it would be wise of us to determine the level of impact on those tests before proceeding.

xieyuschen:
> I will take a deeper look next week. The majority of our JetStream tests utilize clusters and superclusters (multi-server) in the same test process. We also have tests that use the log output to validate the test. So I think it would be wise of us to determine the level of impact on those tests before proceeding.

`The majority of our JetStream tests utilize clusters and superclusters (multi-server) in the same test process. We also have tests that use the log output to validate the test. `

Thanks for briefing the background knowledge to me.

derekcollison:
I like what the PR is trying to achieve, so let's see how big of an impact it might have on the above issues.

philpennock:
I think that the concern raised by @ripienaar is serious: RI develops Choria, which embeds a NATS server.  This is a supported use-case: a nats-server object embedded inside another application.  In such a scenario, the logger provided might already have context attached to it.

By having a library server setup mutate global state, that means that the outer application needs to work around this.

Now, AFAIK Choria only embeds one nats-server per process, so this still works for that case, but this design change seems like an anti-pattern.

xieyuschen is absolutely correct about the overly broad sharing of visibility, but the options for limiting that in Go are probably uglier than what we have now.  Most of the work in this PR _could_ be used to have a NATSLogger type with methods upon it, but without making the logger anonymous inside the server object that's still going to require explicit reference to the logger for the dispatch.

So I think that the logger-in-server approach we're using is probably the "least bad" for our use-cases.  Not _good_, but the alternatives are more ugly or break supported use-cases.

xieyuschen:
> I think that the concern raised by @ripienaar is serious: RI develops Choria, which embeds a NATS server. This is a supported use-case: a nats-server object embedded inside another application. In such a scenario, the logger provided might already have context attached to it.
> 
> By having a library server setup mutate global state, that means that the outer application needs to work around this.
> 
> Now, AFAIK Choria only embeds one nats-server per process, so this still works for that case, but this design change seems like an anti-pattern.
> 
> xieyuschen is absolutely correct about the overly broad sharing of visibility, but the options for limiting that in Go are probably uglier than what we have now. Most of the work in this PR _could_ be used to have a NATSLogger type with methods upon it, but without making the logger anonymous inside the server object that's still going to require explicit reference to the logger for the dispatch.
> 
> So I think that the logger-in-server approach we're using is probably the "least bad" for our use-cases. Not _good_, but the alternatives are more ugly or break supported use-cases.

Get it, if we cannot separate the logger from the server instance, I think we could use another way to log:
The possible solution:
- do not log details in some deep function call that has a weak connection with the server.
- use context to store log information and finalize those logs in one function which has a `Server` receiver.

PR #3184, opened 2022-06-11
[IMPROVED] Performance of wildcard filtered consumer with stream with many subjects.

derekcollison:
In the presence of many subjects in a stream and a wildcard filter subject, fall back to linear scan if too many.

/cc @nats-io/core


ripienaar (approved):
LGTM

PR #3190, opened 2022-06-14
Add in support for inactivity thresholds for durable consumers.

derekcollison:
Prepping more support for JS simplification. Also will be useful in override configs to cleanup MQTT QoS-1 consumers that are orphaned.

/cc @nats-io/core


matthiashanel (approved):
LGTM

PR #3193, opened 2022-06-15
Allow for MQTT QoS-1 consumers to be auto cleaned up

derekcollison:
A server can configure the mqtt section to designate orphaned inactive consumers for auto cleanup.
Several users have reported a buildup of orphaned consumers from MQTT software they can not change or update.

/cc @nats-io/core

### DIFF
--- a/server/mqtt.go
+++ b/server/mqtt.go
@@ -3622,6 +3622,9 @@ func (sess *mqttSession) processJSConsumer(c *client, subject, sid string,
 		if r := opts.MQTT.ConsumerReplicas; r > 0 {
 			cc.Replicas = r
 		}
+		if opts.MQTT.ConsumerInactiveThreshold > 0 {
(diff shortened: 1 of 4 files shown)


ripienaar (approved):
LGTM

derekcollison:
Added in support for reload, @ripienaar could you take another quick look?

ripienaar (approved):
LGTM

derekcollison:
We have had several customers express the need for auto-cleanup from QoS1 consumers that have been abandoned. I do understand your point though.

Should we consider storing the state in the session to recreate it if it is requested again? Or send back an error?

kozlovic:
We do store the consumer config in the session, so we could do a lookup of the consumer on "reconnect" and if the consumer is not found return an error. But the problem with that is that the session would never been able to be "reopened" since the application "connect" would fail, and so we never be able to be properly removed.

So I may suggest that we have instead a timer-based check on closed sessions and removed those (including their consumers). Something to think about.

PR #3194, opened 2022-06-16
[IMPROVED] Distributed placement of streams in a cluster

derekcollison:
If no max bytes set do better randomization of placement.
If the stream is replicated take into consideration number of ha assets on the server already.

/cc @nats-io/core

### DIFF
--- a/server/jetstream_cluster.go
+++ b/server/jetstream_cluster.go
@@ -4158,6 +4158,7 @@ func (cc *jetStreamCluster) selectPeerGroup(r int, cluster string, cfg *StreamCo
 	type wn struct {
 		id    string
 		avail uint64
+		ha    int
 	}
 
 	var nodes []wn
@@ -4180,7 +4181,7 @@ func (cc *jetStreamCluster) selectPeerGroup(r int, cluster string, cfg *StreamCo
 	if uniqueTagPrefix != _EMPTY_ {
 		for _, tag := range tags {
 			if strings.HasPrefix(tag, uniqueTagPrefix) {
-				// disable uniqueness check of explicitly listed in tags
+				// disable uniqueness check if explicitly listed in tags
 				uniqueTagPrefix = _EMPTY_
 				break
 			}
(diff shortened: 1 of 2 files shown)


server/jetstream_cluster.go
```diff
@@ -4281,6 +4288,11 @@ func (cc *jetStreamCluster) selectPeerGroup(r int, cluster string, cfg *StreamCo
 	// Sort based on available from most to least.
 	sort.Slice(nodes, func(i, j int) bool { return nodes[i].avail > nodes[j].avail })
 
+	// If we are placing a replicated stream, let's sort based in haAssets, as that is more important to balance.
+	if cfg.Replicas > 1 {
+		sort.SliceStable(nodes, func(i, j int) bool { return nodes[i].ha < nodes[j].ha })
```
matthiashanel, line 4293: one sort step? 

If nodes[i].avail == nodes[i].avail {
    return nodes[i].ha < nodes[j].ha
} else {
    return nodes[i].avail > nodes[j].avail
}

server/jetstream_cluster.go
```diff
@@ -4271,7 +4278,7 @@ func (cc *jetStreamCluster) selectPeerGroup(r int, cluster string, cfg *StreamCo
 			}
 		}
 		// Add to our list of potential nodes.
-		nodes = append(nodes, wn{p.ID, available})
+		nodes = append(nodes, wn{p.ID, available, ha})
```
matthiashanel, line 4281: Ha won't cover r16 and includes consumer

server/jetstream_cluster.go
derekcollison: Right now I am only sorting and preferring HA when placing a replicated stream. These are are wittled down to a cluster, so think in terms of 3 or 5 nodes in the general case.

server/jetstream_cluster.go
derekcollison: Not following the R16 comment.. But yes, if we are placing a replicated stream I think we try to balance on HA assets per peer over available memory. Note that for the time being, all nodes can handle the stream from a storage perspective if max bytes was specified.

matthiashanel: What I am not understanding is:
If you have 5 empty R1 streams already placed.
How are they accounted for when placing anything else?

derekcollison: Currently they are not formally, but we shuffle the peers to randomize if available is all the same.

derekcollison: The test I wrote shows that the test Marco was trying we now randomize placements. If you specific MaxBytes it worked ok already.

matthiashanel (approved):
LGTM

PR #3204, opened 2022-06-20
[FIXED] Allow users directives for leafnodes to not block reloads.

### LINKED ISSUE #1921: Add support to leafnode config reload
## Feature Request

It would be useful to enable config reload for a few fields of a leafnode config, like the remotes which are not possible to prune right now:

```
[32903] 2021/02/17 19:55:43.113029 [ERR] Failed to reload server configuration: config reload not supported for LeafNode: old={ 0    [] 0 <nil> 0 false  false 1s [0xc0000d02d0] <nil> 0 0}, new={ 0    [] 0 <nil> 0 false  false 1s [0xc00028e090] <nil> 0 0}
```

#### Use Case:

Environments where it is not ideal to restart a server when modifying the leafnode edge configuration.

#### Alternative Approaches

Restart a server.

derekcollison:

Resolves #1921 

/cc @nats-io/core

### DIFF
--- a/server/leafnode_test.go
+++ b/server/leafnode_test.go
@@ -4349,3 +4349,33 @@ func TestLeafNodeStreamAndShadowSubs(t *testing.T) {
 	// Check again
 	subPubAndCheck()
 }
+
+func TestLeafNodeAuthConfigReload(t *testing.T) {
+	template := `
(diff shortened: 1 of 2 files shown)


ripienaar (approved):
LGTM

PR #3211, opened 2022-06-23
Fixes https://github.com/nats-io/nats-server/issues/3151

### LINKED ISSUE #3151: Mappings cannot have spaces in mapping functions
## Defect

a52f12613e6eb53097d6d563d9b9681b88a037cb

Linux EL8

Use configuration:

```
mappings = {
  "in.registration.*.>": "registration.{{ partition(5, 1) }}.{{wildcard(1)}}.>"
}
```

```
$ nats-server --config /tmp/server.conf
nats-server: /tmp/server.conf:2:4: Error adding mapping for "in.registration.*.>" to "registration.{{partition(5, 1)}}.{{wildcard(1)}}.>" : invalid subject
```

But if we adjust the subject mapping like this (remove all spaces):

```
mappings = {
  "in.registration.*.>": "registration.{{partition(5,1)}}.{{wildcard(1)}}.>"
}
```

It works.

This is because https://github.com/nats-io/nats-server/blob/a52f12613e6eb53097d6d563d9b9681b88a037cb/server/accounts.go#L618-L621  is checking the transform subject for validity before any parsing is done, I think probably we should remove that check.  Spaces in these mappings greatly improves readability and we specifically supported them in the code when implementing it.

jnmoyne:
 - [X] Tests added

Resolves #3151

 Allows spaces inside moustache subject mapping destinations

/cc @nats-io/core


server/sublist.go
```diff
@@ -1061,11 +1061,21 @@ func IsValidSubject(subject string) bool {
...
 		}
-		if len(t) > 1 {
-			if strings.ContainsAny(t, "\t\n\f\r ") {
+		if length > 1 {
+			if strings.ContainsAny(t, "\t\n\f\r") {
+				return false
+			}
+			// if the token is a moustache don't worry about spaces inside
+			if length > 4 {
+				if t[0] == '{' && t[1] == '{' && t[length-2] == '}' && t[length-1] == '}' {
```
ripienaar, line 1074: Not really my area of expertise but I am not sure this is the right fix, IsValidSubject is called all over the place to validate subjects not just mappings.  Spaces in subjects are always invalid and IsValidSubject should not say they are.

Where we test subjects in the process of establishing mappings we should handle the specific case where mappings are not valid subjects.

jnmoyne: Maybe I misunderstood the issue, so you would rather have a new 'isValidMapping' that accepts spaces inside mustaches to be called in https://github.com/nats-io/nats-server/blob/a52f12613e6eb53097d6d563d9b9681b88a037cb/server/accounts.go#L618-L621 ?

ripienaar: yes, or just remove the call to IsValidSubject() there, a deep validate might be good though, one that checks for like invalid mapping functions and all that stuff.

jnmoyne: For sure deeper validate and better errors to let the user know what the problem is a good thing as it's rather poor right now, will expand the scope from just #3151

ripienaar: Yeah, I think we can also tackle those in a follow up and just drop the valid subject check here tbh

jnmoyne: Yeah, that was a bit my idea (better validation/error handling in a follow up), but regardless of the breaking it down or not fair enough on a new IsValidMappingDestSubject() anyways which can be made as simple as what I have done in that initial commit or mode much more sophisticated with better errors.

jnmoyne: Now does deeper validation, and also returns new kinds of errors to help the users when creating mappings, should be good now.

server/sublist.go
```diff
@@ -1128,6 +1129,42 @@ func isValidLiteralSubject(tokens []string) bool {
...
+				return ErrBadSubjectMappingDestination
+			} else {
+				continue
+			}
+		}
+
+		switch t[0] {
+		case fwc:
+			sfwc = true
+		case ' ', '\t', '\n', '\r', '\f':
```
ripienaar, line 1161: the if and the case statement can combine to 

```
if length == 1 && t[0] == fwc {
   sfwc = true
} else if strings.ContainsAny(t, "\t\n\f\r ") {
   return ErrBadSubjectMappingDestination
}
```

Though yours might be a tad faster since sometimes it wont be calling the more complex `strings.ContainsAny`.  This is certainly more readable and maintainable though

ripienaar (commented):
Some minor comments nothing major

server/sublist.go
ripienaar:
```
BenchmarkCompare/ContainsAny
BenchmarkCompare/ContainsAny-16         	162492498	         7.037 ns/op
BenchmarkCompare/Rune_compare
BenchmarkCompare/Rune_compare-16        	1000000000	         0.0000001 ns/op
``` its faster, but is it really going to matter :)

File: server/store.go [hunk 3]
```diff
@@ -372,11 +372,11 @@ const (
...
 	switch st {
 	case MemoryStorage:
-		return strings.Title(memoryStorageString)
+		return "Memory"
 	case FileStorage:
-		return strings.Title(fileStorageString)
+		return "File"
 	case AnyStorage:
-		return strings.Title(anyStorageString)
+		return "Any"
```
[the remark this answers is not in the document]
jnmoyne: Hum... that change is not mine :) it was in one of @derekcollison commits on the main branch that fix CI tests and that got pulled in when I did a `git pull --rebase origin main` on the 25th.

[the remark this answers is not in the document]
derekcollison: For move to 1.18 to pass staticcheck for deprecated use of Title().

server/sublist.go
```diff
@@ -1128,6 +1129,42 @@ func isValidLiteralSubject(tokens []string) bool {
 	return true
 }
 
+// ValidateDestinationSubject returns nil error if the subject is a valid subject mapping destination subject
+func ValidateDestinationSubject(subject string) error {
```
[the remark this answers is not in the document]
jnmoyne: Fair enough

server/sublist.go
jnmoyne: That's how the code was and I didn't change it but I agree what you suggest is much more elegant and readable and indeed the difference in performance is not going to matter.

server/accounts_test.go
```diff
@@ -3262,7 +3262,7 @@ func TestSamplingHeader(t *testing.T) {
 func TestSubjectTransforms(t *testing.T) {
 	shouldErr := func(src, dest string) {
 		t.Helper()
-		if _, err := newTransform(src, dest); err != ErrBadSubject && err != ErrBadSubjectMappingDestination {
+		if _, err := newTransform(src, dest); err != ErrBadSubject && err != ErrBadSubjectMappingDestination && err != ErrorMappingDestinationFunctionWildcardIndexOutOfRange && err != ErrUnknownMappingDestinationFunction && err != ErrorMappingDestinationFunctionNotEnoughArguments && err != ErrorMappingDestinationFunctionInvalidArgument && err != ErrorMappingDestinationFunctionTooManyArguments {
```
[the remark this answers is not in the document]
jnmoyne: This actually led me to spend a bit of time actually improving and re-working the error handling for subject mapping destinations and implementing what I really wanted to implement in the end with IMHO much more useful error messages that should really help he user figure out their mistakes (see new commit).

jnmoyne:
Should be good for final review now

ripienaar:
May need to ask for help to get this PR be rebased on main so whats in the PR is only the bits you're changing - all this of circleci changes etc should not be here surely? I think Ivan is out git exper he might be able to help :)

kozlovic:
> Do you think the problem is doing git pull --rebase origin main 

Likely. If you do `git rebase main`, you are basically applying all your commits on top of main, which for a PR that is short-lived (not something that we have been working for months) is the right thing to do. In that case, you will see in your PR only *your* commits, which is way better for the reviewers.

jnmoyne:
Replaced by https://github.com/nats-io/nats-server/pull/3231

PR #3212, opened 2022-06-24
Add quick-n-dirty HTTP client check

kung-foo:
 - [ ] Tests added

Resolves #

A common issue we run into when deploying NATS/NGS clients at our customers' sites is verifying outbound network access. Enterprise IT often has a "block-by-default" policy, so our on-boarding process includes requesting a network exception, and then verifying the rule change. To do that, we have historically recommended the customer just open a browser to http://connect.ngs.global:4222 to check for the basic `INFO` block being shown. However, Chrome now displays a "network error" since (as expected) the NATS server doesn't speak HTTP.

This PR adds a quick-n-dirty check to see if the client is speaking HTTP rather than NATS, and if so, return an HTTP response before closing the connection.

Since the NATS protocol expects the server to first send the `INFO` message, the client should not have sent any data at the start. So we make a quick (~20ms :shrug:) read to see if `GET ` is available.

This may not be the cleanest way to do it, and it may not be worth the 20ms penalty for _all_ connections. But I thought I'd throw out some working code for feedback.

Note: this only applies to the std TCP transport. Something similar (and probably easier) would need to be added for the WebSocket transport.

/cc @nats-io/core

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -27,13 +27,12 @@ import (
 	"math/rand"
 	"net"
 	"net/http"
-	"regexp"
-
 	// Allow dynamic profiling.
 	_ "net/http/pprof"
 	"os"
 	"path"
 	"path/filepath"
+	"regexp"
 	"runtime"
 	"strconv"
 	"strings"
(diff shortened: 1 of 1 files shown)


derekcollison:
For NGS, have your customers check https://connect.ngs.global, this will return Bad Request but they will know connectivity works.

kung-foo:
True, however we currently use 4222, but in the future we'll transition to websockets over 443.

derekcollison:
You can have both running concurrently, which is what we have on NGS as well as demo server. On demo we have MQTT as well.

kung-foo:
Understand, though when a customer requests a firewall exception, it is usually for a single port. So the only test possible is for NGS on 4222 as 443 would still be blocked. This is not a critical issue, just a quality of life item. Our applications check for access on install, but we like to ensure the requisite network changes are in place _before_ deployment.

PR #3216, opened 2022-06-24
Allow consumer filter subjects to be updated

derekcollison:

/cc @nats-io/core

### DIFF
--- a/server/jetstream_test.go
+++ b/server/jetstream_test.go
@@ -17933,6 +17933,71 @@ func TestJetStreamConsumerAndStreamNamesWithPathSeparators(t *testing.T) {
 	require_Error(t, err, NewJSConsumerNameContainsPathSeparatorsError(), nats.ErrInvalidConsumerName)
 }
 
+func TestJetStreamConsumerUpdateFilterSubject(t *testing.T) {
+	s := RunBasicJetStreamServer()
+	if config := s.JetStreamConfig(); config != nil {
+		defer removeDir(t, config.StoreDir)
+	}
+	defer s.Shutdown()
+
+	nc, js := jsClientConnect(t, s)
+	defer nc.Close()
+
+	_, err := js.AddStream(&nats.StreamConfig{Name: "T", Subjects: []string{"foo", "bar", "baz"}})
+	require_NoError(t, err)
+
+	// 10 foo
+	for i := 0; i < 10; i++ {
+		js.PublishAsync("foo", []byte("OK"))
+	}
+	// 20 bar
+	for i := 0; i < 20; i++ {
+		js.PublishAsync("bar", []byte("OK"))
+	}
+	select {
+	case <-js.PublishAsyncComplete():
+	case <-time.After(5 * time.Second):
+		t.Fatalf("Did not receive completion signal")
+	}
+
+	sub, err := js.PullSubscribe("foo", "d")
+	require_NoError(t, err)
+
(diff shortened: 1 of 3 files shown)


server/jetstream_test.go
```diff
@@ -17933,6 +17933,71 @@ func TestJetStreamConsumerAndStreamNamesWithPathSeparators(t *testing.T) {
...
+	for i := 0; i < 20; i++ {
+		js.PublishAsync("bar", []byte("OK"))
+	}
+	select {
+	case <-js.PublishAsyncComplete():
+	case <-time.After(5 * time.Second):
+		t.Fatalf("Did not receive completion signal")
+	}
+
+	sub, err := js.PullSubscribe("foo", "d")
```
wallyqs, line 17963: One side effect to note is that this pull subscription that subscribed with filter subject `foo`, once the update is applied to `bar` only, when it does a Fetch it will receive the messages from the `bar` subject instead which did not match to the original FilterSubject `foo`.

derekcollison: What do we think the right behavior here should be? For bound subs in applications?

wallyqs: If the application restarts then it would now fail with a `nats: subject does not match consumer` error, so wondering whether we should be sending a status error to any inflight next msg request for the consumer on the update when the consumer config changes and there is a mismatch with the filter subject to signal that it is now invalid, then that would prevent consumers accidentally fetching messages that they were not supposed to originally.

derekcollison: @tbeets What are you thoughts? This is be driven from customer requests.

tbeets: I think we document known side-effects, including the example @wallyqs gives when we release the feature.  Frankly, as learnings show, `js.subscribe()` with a subject filter was an abstractive mistake as it misled the app developer on what was really going on and caused false assumptions.  JS Simplified abstraction will not include.

wallyqs (approved):
LGTM

server/jetstream_test.go
derekcollison: Regardless of the subject being a way to easily bind to consumers or create them, we will continue to have the notion of filtering a stream within a consumer. 

So being able to change those filtering constructs for consumers we will still have the same issue above. A subscription being bound to a consumer that changes.

tbeets: Agree.  At the end of the day JS Consumer's are server-side services owned by an Account, never by a client/user.  This enhancement is giving the service owner the ability to change its existing contract in a way that could break clients.  We can't know if this will be responsibly and sparingly used or not, but also I feel strongly this is an internal organizational choice, issue, and process, not to be solved necessarily by NATS.

I think better to educate that any subject filter specified by a client is relevant only at that moment of time, since a JS Consumer can be modified subsequently by an owner.  I think this will be rare.  IMO I would not immediately add a new signal code or error etc.  We also can't know if the changed subject filter by an owner is actually dropping a semantically identical payload to clients as the old filter.  Perhaps it is an update at consumer to switch from "blue" messages to "green" messages or something in a blue/green test cycle.

derekcollison: ok can you LGTM as well..

derekcollison: @tbeets ok to merge?

tbeets (approved):
LGTM

PR #3217, opened 2022-06-25
[Add] ability for operator to move streams (WIP)

matthiashanel:
Also added:
    ability to reload tags
    special tag (!jetstream) to remove peer from peer placement
    $JS.API.SERVER.STREAM.MOVE subject to initiate move away from a server

    This changes a detail about regular stream move as well.
    Before differing cluster names where used to start/stop a transfer.
    Now only the peer list and it's size relative to configured replica
    matter.
    Once a transfer is considered completed, excess peers will be dropped
    from the beginning of the list.
    This allows transfers within the cluster as well.

### DIFF
--- a/server/jetstream_api.go
+++ b/server/jetstream_api.go
@@ -187,6 +187,16 @@ const (
 	// Will return JSON response.
 	JSApiRemoveServer = "$JS.API.SERVER.REMOVE"
 
+	// JSApiServerStreamMove is the endpoint to move streams off a server
+	// Only works from system account.
+	// Will return JSON response.
+	JSApiServerStreamMove = "$JS.API.SERVER.STREAM.MOVE"
+
+	// JSApiServerInfo is the endpoint to abtain which stream a server contains
+	// Only works from system account.
+	// Will return JSON response.
+	//JSApiServerStreamInfo = "$JS.API.SERVER.STREAM.INFO"
+
 	// jsAckT is the template for the ack message stream coming back from a consumer
 	// when they ACK/NAK, etc a message.
 	jsAckT      = "$JS.ACK.%s.%s"
@@ -557,6 +567,36 @@ type JSApiMetaServerRemoveResponse struct {
 }
 
 const JSApiMetaServerRemoveResponseType = "io.nats.jetstream.api.v1.meta_server_remove_response"
+
+/*
+// JSApiMetaServerStreamInfoRequest will provide which streams are located on a particular server
+type JSApiMetaServerStreamInfoRequest struct {
+	// Server name of the peer to be evacuated.
+	Server string `json:"peer"`
+}
+
+// JSApiMetaServerStreamInfoResponse is the response to a peer evacuation request in the meta group.
+type JSApiMetaServerStreamInfoResponse struct {
+	ApiResponse
+	Success bool                      `json:"success,omitempty"`
+	Content map[string][]StreamConfig `json:"content,omitempty"`
+}
+
+const JSApiMetaServerStreamInfoType = "io.nats.jetstream.api.v1.meta_server_stream_info"
+*/
+
+// JSApiMetaServerStreamMoveRequest will move a stream on a server to another
+// response to this will come as JSApiStreamUpdateResponse/JSApiStreamUpdateResponseType
+type JSApiMetaServerStreamMoveRequest struct {
+	// Server name of the peer to be evacuated.
+	Server string `json:"peer"`
+	// Account name the stream to move is in
+	Account string `json:"account"`
+	// Name of stream to move
+	Stream string `json:"string"`
+	// Ephemeral placement tags for the move
+	Tags []string `json:"tags,omitempty"`
+}
 
 // JSApiMsgGetRequest get a message request.
 type JSApiMsgGetRequest struct {
@@ -1360,9 +1400,9 @@ func (s *Server) jsStreamUpdateRequest(sub *subscription, c *client, _ *Account,
 		// If we are inline with client, we still may need to do a callout for stream info
 		// during this call, so place in Go routine to not block client.
 		if c.kind != ROUTER && c.kind != GATEWAY {
-			go s.jsClusteredStreamUpdateRequest(ci, acc, subject, reply, rmsg, &cfg)
+			go s.jsClusteredStreamUpdateRequest(ci, acc, subject, reply, rmsg, &cfg, nil)
 		} else {
-			s.jsClusteredStreamUpdateRequest(ci, acc, subject, reply, rmsg, &cfg)
+			s.jsClusteredStreamUpdateRequest(ci, acc, subject, reply, rmsg, &cfg, nil)
 		}
 		return
 	}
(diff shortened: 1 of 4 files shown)


server/jetstream_cluster.go
```diff
@@ -158,6 +162,7 @@ const (
 	defaultStoreDirName  = "_js_"
 	defaultMetaGroupName = "_meta_"
 	defaultMetaFSBlkSize = 1024 * 1024
+	jsPeerExcludeTag     = "!jetstream"
```
tbeets, line 165: Suggest `jsExcludePlacement` value string, case-insensitive, as clearer meaning.

derekcollison: So we want JS to stay active while we evacuate any assets etc but just have it not applicable for placement correct?

server/jetstream_api.go
```diff
@@ -558,6 +568,36 @@ type JSApiMetaServerRemoveResponse struct {
...
+// response to this will come as JSApiStreamUpdateResponse/JSApiStreamUpdateResponseType
+type JSApiMetaServerStreamMoveRequest struct {
+	// Server name of the peer to be evacuated.
+	Server string `json:"peer"`
+	// Account name the stream to move is in
+	Account string `json:"account"`
+	// Name of stream to move
+	Stream string `json:"string"`
+	// Ephemeral placement tags for the move
+	Tags []string `json:"tags,omitempty"`
```
derekcollison, line 598: What happens if these conflict with current tags? Does the request fail?
How do you see these being used?

server/jetstream_api.go
```diff
@@ -2156,6 +2196,262 @@ func (s *Server) jsLeaderServerRemoveRequest(sub *subscription, c *client, _ *Ac
...
+		for _, s := range a {
+			contains := false
+			for _, p := range s.Group.Peers {
+				if p == found {
+					contains = true
+					break
+				}
+			}
+			if contains {
+				resp.Content[aName] = append(resp.Content[aName], *s.Config)
```
derekcollison, line 2293: Why not just inline at line 2288?

derekcollison: vs bool contains, etc.

server/jetstream_api.go
```diff
@@ -2156,6 +2196,262 @@ func (s *Server) jsLeaderServerRemoveRequest(sub *subscription, c *client, _ *Ac
...
+	}
+	js.mu.RUnlock()
+
+	if found == _EMPTY_ {
+		resp.Error = NewJSClusterServerNotMemberError()
+		s.sendAPIErrResponse(ci, acc, subject, reply, string(msg), s.jsonResponse(&resp))
+		return
+	}
+
+	resp.Content = make(map[string][]StreamConfig)
```
derekcollison, line 2281: This should only be done if we add one.

server/jetstream_api.go
```diff
@@ -2156,6 +2196,262 @@ func (s *Server) jsLeaderServerRemoveRequest(sub *subscription, c *client, _ *Ac
...
+		if peers == nil {
+			resp.Error = NewJSClusterNoPeersError()
+			s.sendAPIErrResponse(ci, acc, subject, reply, string(msg), s.jsonResponse(&resp))
+			return
+		}
+	}
+
+	cfg.Placement = origPlacement
+
+	if c.kind != ROUTER && c.kind != GATEWAY {
```
derekcollison, line 2448: Maybe add same comment to explain why since even on CLIENT or LEAF this will block waiting on other responses from other servers etc.

server/jetstream_cluster.go
```diff
@@ -4018,6 +4052,14 @@ func (js *jetStream) startUpdatesSub() {
 	if cc.peerRemove == nil {
 		cc.peerRemove, _ = s.systemSubscribe(JSApiRemoveServer, _EMPTY_, false, c, s.jsLeaderServerRemoveRequest)
 	}
+	/*
```
derekcollison, line 4055: Remove

server/jetstream_cluster.go
```diff
@@ -4039,6 +4081,16 @@ func (js *jetStream) stopUpdatesSub() {
 		cc.s.sysUnsubscribe(cc.peerRemove)
 		cc.peerRemove = nil
 	}
+	/*
```
derekcollison, line 4084: Same

server/jetstream_cluster.go
```diff
@@ -4700,42 +4761,51 @@ func (s *Server) jsClusteredStreamUpdateRequest(ci *ClientInfo, acc *Account, su
...
-				// Make sure to set if not already set.
-				if cca.Group.Preferred == _EMPTY_ {
-					cca.Group.Preferred = cca.Group.Peers[0]
+				if !found {
+					peerSet = append(peerSet, peer)
 				}
-				// We can not propose here before the stream itself so we collect them.
-				consumers = append(consumers, cca)
 			}
+			peerSet = append(peerSet, nrg.Peers...)
```
derekcollison, line 4785: This seems to ignore or overwrite any filtering above no?

server/reload.go
```diff
@@ -293,6 +293,15 @@ func (a *authTimeoutOption) Apply(server *Server) {
 	server.Noticef("Reloaded: authorization timeout = %v", a.newValue)
 }
 
+// tagsOption implements the option interface for the `tags` setting.
+type tagsOption struct {
+	noopOption // Not authOption because this is a no-op; will be reloaded with options.
+}
+
+func (u *tagsOption) Apply(server *Server) {
+	server.Noticef("Reloaded: tags")
```
derekcollison, line 302: I like this addition, but please note, AFAIK we have never written updates to config files as a result of a $SYS msg. That was by design. So we should carefully consider this and make sure we are committed here. Also need to loop in Phil imo.

There may be a better way to achieve the same goal of temporarily notifying the metaleader not to consider this server for placement, etc.

server/jetstream_api.go
matthiashanel: This is a draft still. 

This particular function I'll delete based on feedback from dave.
Basically the jsz request provides the same information. 
Dave is going to look at that and let me know if it works for him.
If yes, I'll delete this, if no I'll keep it but probably tweak the returned information.

server/jetstream_api.go
matthiashanel: same as above, will incorporate suggestions when this is keept.

server/jetstream_api.go
matthiashanel: will add

server/jetstream_cluster.go
matthiashanel: will remove once confirmed with dave

server/jetstream_cluster.go
matthiashanel: same as above

server/reload.go
matthiashanel: This is not triggered by an update on a $SYS message. 
This is driven by configuration only.

So if, for or whatever reason, we need placement to stop for a server. just place the new tag and reload the config.
Separating move and stopping of placement into two features was an intentional decision.

Enabling stopping of placement via server tags also piggy backs onto an existing mechanism (propagation already implemented) and is robust against restarts. (persistence is done via config) (just consider, meta leader changing, actual server being restarted for whatever reason...)
Furthermore, reload of tags is a worthwhile feature in its own right. 
(With this, tags can also convey the config version)

derekcollison: ok, I think black box operations are desirable, but don't think we should be updating the server config. Sounds like we should discuss before more progress is made.

matthiashanel: sent out an invite for later today

server/jetstream_api.go
matthiashanel: removed

server/jetstream_api.go
matthiashanel: the tags applied are stream placement tags + tags in request. 
This way you can only move within the constraints originally specified.

server/jetstream_cluster.go
matthiashanel: no, goal of this is to have a peer list with unique peers.  
To get there, I filter peers in peerSet that are also present in nrg.Peers.
Post concatenation, I have no duplicates.

server/reload.go
derekcollison: Thanks, was chatting with Caleb as well this am, should be a short call I think.

ripienaar: I think there is usecase for both:

 * Normal operations we need to be able to tell a machine to stop accepting stuff and evict stuff without config changes, it should be soft and in memory only driven by a $SYS style query. It should not exit once all is evicted - its not a LDM style operation
 * A config override so that, while preparing a new machine or dealing with a machine that is known to have some issue like disk performance, we can stop accepting assets but have it up for monitoring etc but it could still be fine for core traffic

Related, we need a $SYS style way to signal a config reload.

derekcollison: If we don't have a way to deliver new configs via NATS and $SYS not sure that is useful? Just have the extra layer that delivers the new config also signal after receipt and successful check?

ripienaar: No, not new configs, just signal reloads that responds back with success/err and reason.

For placement candidate it could be something that persists into the raft layer and resets either using API or force flag in the config file. But probably fine just to be in memory while the machine is up for normal operation needs.

Wdyt @davidkemper

derekcollison (approved):
LGTM

server/jetstream_cluster.go
tbeets: Correct.  A marker on a server so it is not considered for new placements.

tbeets (approved):
LGTM

server/reload.go
davidkemper: It is my understanding that putting this setting into the metaleader would open any number of cans of worms.

Where I'm going to make a big disruptive change recreating a vm, IMHO it's an acceptable compromise to set some persistent server tag as a prep step.

The prime use case is keeping r=1 highly available during a disruptive change. In this case I can spin up a replacement in the same availability zone as a target to keep all data in the same data center.

Arguably a simple tag on the doomed server would be very robust and intentional even across multiple failure modes, including having to upgrade the servers halfway through...

matthiashanel: We can always add a dedicated system call with the necessary bells and whistles later. 
The config based approach gives us exactly what we want for very little effort.

derekcollison: I wonder if we set the tag on a server that is also the meta leader, its steps down first, delays until new leader ok then sends tag metadata about placement?

server/jetstream_api.go
```diff
@@ -187,6 +187,16 @@ const (
...
 
+	// JSApiServerStreamMove is the endpoint to move streams off a server
+	// Only works from system account.
+	// Will return JSON response.
+	JSApiServerStreamMove = "$JS.API.SERVER.STREAM.MOVE"
+
+	// JSApiServerInfo is the endpoint to abtain which stream a server contains
+	// Only works from system account.
+	// Will return JSON response.
+	//JSApiServerStreamInfo = "$JS.API.SERVER.STREAM.INFO"
```
ripienaar, line 198: will this come in a follow up PR?

ripienaar: Are we also adding the ability to get a list of streams on a given server?

server/jetstream_api.go
```diff
@@ -558,6 +568,36 @@ type JSApiMetaServerRemoveResponse struct {
...
+
+// JSApiMetaServerStreamMoveRequest will move a stream on a server to another
+// response to this will come as JSApiStreamUpdateResponse/JSApiStreamUpdateResponseType
+type JSApiMetaServerStreamMoveRequest struct {
+	// Server name of the peer to be evacuated.
+	Server string `json:"peer"`
+	// Account name the stream to move is in
+	Account string `json:"account"`
+	// Name of stream to move
+	Stream string `json:"string"`
```
ripienaar, line 596: This should be `stream`, not `string`

PR #3221, opened 2022-06-28
Made direct get from a stream part of the $JS.API hierarchy vs separate.

derekcollison:
Also for direct get and for pull requests, if we are not on a client connection check how long we have been away from the read loop. If need be execute in a separate go routine.

Note this will move the direct get under the correct subject imo, so account imports/exports should be more easily handled.

Also noting that our bet going forward is that the majority of pulling messages from a system will be the pull request for a consumer and this direct get as will be used in KV, I wanted to make sure to not block any routes or gateways.

/cc @nats-io/core

### DIFF
--- a/server/consumer.go
+++ b/server/consumer.go
@@ -2602,7 +2602,20 @@ func (o *consumer) processNextMsgReq(_ *subscription, c *client, _ *Account, _, 
 		return
 	}
 	_, msg = c.msgParts(msg)
-	o.processNextMsgRequest(reply, msg)
+
+	inlineOk := c.kind != ROUTER && c.kind != GATEWAY && c.kind != LEAF
+	if !inlineOk {
+		// Check how long we have been away from the readloop for the route or gateway or leafnode.
+		// If too long move to a separate go routine.
+		if elapsed := time.Since(c.in.start); elapsed < noBlockThresh {
+			inlineOk = true
+		}
+	}
+	if inlineOk {
+		o.processNextMsgRequest(reply, msg)
+	} else {
+		go o.processNextMsgRequest(reply, copyBytes(msg))
+	}
 }
 
 func (o *consumer) processNextMsgRequest(reply string, msg []byte) {
--- a/server/jetstream_api.go
+++ b/server/jetstream_api.go
@@ -121,8 +121,11 @@ const (
 	// JSDirectMsgGet is the template for non-api layer direct requests for a message by its stream sequence number or last by subject.
 	// Will return the message similar to how a consumer receives the message, no JSON processing.
 	// If the message can not be found we will use a status header of 404. If the stream does not exist the client will get a no-responders or timeout.
-	JSDirectMsgGet  = "$JS.DS.GET.*"
-	JSDirectMsgGetT = "$JS.DS.GET.%s"
+	JSDirectMsgGet  = "$JS.API.DIRECT.GET.*"
+	JSDirectMsgGetT = "$JS.API.DIRECT.GET.%s"
+
+	// jsDirectGetPre
+	jsDirectGetPre = "$JS.API.DIRECT.GET"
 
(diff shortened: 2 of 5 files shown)


server/jetstream_api.go
```diff
@@ -121,8 +121,11 @@ const (
 	// JSDirectMsgGet is the template for non-api layer direct requests for a message by its stream sequence number or last by subject.
 	// Will return the message similar to how a consumer receives the message, no JSON processing.
 	// If the message can not be found we will use a status header of 404. If the stream does not exist the client will get a no-responders or timeout.
-	JSDirectMsgGet  = "$JS.DS.GET.*"
```
kozlovic, line 124: Were those already used in a public release or used by client libraries that have been released?

server/consumer.go
```diff
@@ -2602,7 +2602,20 @@ func (o *consumer) processNextMsgReq(_ *subscription, c *client, _ *Account, _,
...
+		// Check how long we have been away from the readloop for the route or gateway or leafnode.
+		// If too long move to a separate go routine.
+		if elapsed := time.Since(c.in.start); elapsed < noBlockThresh {
+			inlineOk = true
+		}
+	}
+	if inlineOk {
+		o.processNextMsgRequest(reply, msg)
+	} else {
+		go o.processNextMsgRequest(reply, msg)
```
kozlovic, line 2617: I think you may need a copy of the `msg` here since it will be executed in a go routine.

server/consumer.go
```diff
@@ -2602,7 +2602,20 @@ func (o *consumer) processNextMsgReq(_ *subscription, c *client, _ *Account, _,
 		return
 	}
 	_, msg = c.msgParts(msg)
-	o.processNextMsgRequest(reply, msg)
+
+	inlineOk := c.kind != ROUTER && c.kind != GATEWAY && c.kind != LEAF
+	if !inlineOk {
+		// Check how long we have been away from the readloop for the route or gateway or leafnode.
+		// If too long move to a separate go routine.
+		if elapsed := time.Since(c.in.start); elapsed < noBlockThresh {
```
kozlovic, line 2610: I am wondering if it would not be better to extend the existing `jsAPIRoutedReq` struct to contain a type of request and then have processJSAPIRoutedRequests() call o.processNextMsgRequest() or mset.getDirectRequest(). Of course, `jsAPIRoutedReq` would become a bit more bloated with fields that are not used in some API requests, etc.. but may be ok.

I am just concerned with possible "explosion" of `go o.process..` go routines and also the fact that still, when we consider that it is ok to do it inline, then the call may then block for much longer than anticipated. So with the approach I describe, changes to readCache (the start time) would not be needed and we would unconditionally process in single go routine all requests that are `!inlineOk`.

kozlovic (changes requested):

server/jetstream_api.go
derekcollison: No..

server/consumer.go
derekcollison: In my high speed tests the API version was not better and worse in some respects that just simple Go routine. Also it always does that route if not client connection, I think we can do in place as long as we know we are not stalling the route or gateway.

kozlovic: > we know we are not stalling the route or gateway.

But that's the thing: say last measurement was 250ms, so you decide to do it inline, but now a single message load blocks for 2s due to high contention, etc..

derekcollison: Also note these two in particular are handled at the user account layer not the system later.

server/consumer.go
derekcollison: Fixed.

server/consumer.go
derekcollison: Yes the first one that inlines that is bad will not allow any other behind it to be inline.. I don't think right thing to do is *always* spin it out.

kozlovic (approved):
LGTM

PR #3222, opened 2022-06-28
[fixed] reload related races

matthiashanel:
account.rm had races caused by reload copying rm from one account to
another

mset.store was used outsisde the lock

in rare cases the stasz message was not received in time.
Trigger automatically now

sometimes a statsz message received before reload cause issues.
try receiving a second time


derekcollison (approved):
LGTM - Nice.

kozlovic (approved):
LGTM

PR #3225, opened 2022-06-29
Rework startupComplete gate from PR #2360

kozlovic:
The "InProcess" change make readyForConnections() possibly return
much faster than it used to, which could cause tests to fail.

Restore the original behavior, but in case of DontListen option
wait on the startupComplete gate.

Also fixed some missing checks for leafnode connections.

### DIFF
--- a/server/jetstream_cluster_test.go
+++ b/server/jetstream_cluster_test.go
@@ -6140,6 +6140,8 @@ func TestJetStreamClusterDomains(t *testing.T) {
 		strings.ReplaceAll(jsClusterTemplWithSingleLeafNode, "store_dir:", "extension_hint: will_extend, domain: CORE, store_dir:"))
 	defer ln.Shutdown()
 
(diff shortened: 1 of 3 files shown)


kozlovic:
@derekcollison Indeed getting the test failure you mentioned on a call:
```
=== RUN   TestMQTTConsumerReplicasOverride
    mqtt_test.go:5831: Expected consumer to be R1, got: &{Name:MQTT Leader: Replicas:[0xc004252240]}
--- FAIL: TestMQTTConsumerReplicasOverride (3.25s)
```
You mentioned that you may have a fix for the test, so you can maybe update this PR? That being said, I wonder what changes made that happen (I did not see this flapping in the past: it could be also Go 1.18 vs 1.17...)

matthiashanel (approved):
LGTM

derekcollison (approved):
LGTM - Go ahead and merge and I will check and fix the other test in separate PR.

PR #3227, opened 2022-06-29
Allow filtered stream mirrors

### LINKED ISSUE #3226: Stream mirrors cannot have filters anymore
I think we used to have filters on stream mirrors, the terraform tests passed with that set, now they fail though:

```json
{
  "name": "X",
  "retention": "limits",
  "max_consumers": -1,
  "max_msgs_per_subject": -1,
  "max_msgs": -1,
  "max_bytes": -1,
  "max_age": 0,
  "max_msg_size": -1,
  "storage": "file",
  "discard": "old",
  "num_replicas": 1,
  "mirror": {
    "name": "LON",
    "filter_subject":"js.in.orders_69"
  },
  "sealed": false,
  "deny_delete": false,
  "deny_purge": true,
  "allow_rollup_hdrs": false
}
```

```
$ nats req '$JS.API.STREAM.CREATE.X' "$(cat X.json)"
14:22:23 Sending request on "$JS.API.STREAM.CREATE.X"
14:22:23 Received with rtt 945.864µs
{"type":"io.nats.jetstream.api.v1.stream_create_response","error":{"code":400,"err_code":10033,"description":"stream mirrors can not contain filtered subjects"}}

### PR DESCRIPTION
**Pull request by derekcollison**
This was allowed before but did not actually assign the filter subject so did not really work. This re-allows it and makes sure it works properly.

Tests are a tad flaky atm but not due to this.

Resolves #3226 

/cc @nats-io/core

### DIFF
--- a/server/jetstream_cluster_test.go
+++ b/server/jetstream_cluster_test.go
@@ -10689,15 +10689,6 @@ func TestJetStreamClusterMirrorDeDupWindow(t *testing.T) {
 	// Send 100 messages
 	send(100)
 
-	// First check that we can't create with a duplicates window
-	_, err = js.AddStream(&nats.StreamConfig{
-		Name:       "M",
-		Replicas:   3,
-		Mirror:     &nats.StreamSource{Name: "S"},
-		Duplicates: time.Hour,
-	})
-	require_Error(t, err)
-
 	// Now create a valid one.
 	si, err = js.AddStream(&nats.StreamConfig{
 		Name:     "M",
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: server/jetstream_cluster_test.go [hunk 1]
```diff
@@ -10689,15 +10689,6 @@ func TestJetStreamClusterMirrorDeDupWindow(t *testing.T) {
 	// Send 100 messages
 	send(100)
 
-	// First check that we can't create with a duplicates window
```
kozlovic, line 10692: Why remove this test? This is testing that we can't create a mirror with de-dup, which caused some issues in the past.

server/stream.go
```diff
@@ -1006,16 +1006,6 @@ func (s *Server) checkStreamCfg(config *StreamConfig, acc *Account) (StreamConfi
 		if len(cfg.Sources) > 0 {
 			return StreamConfig{}, NewJSMirrorWithSourcesError()
 		}
-		if cfg.Mirror.FilterSubject != _EMPTY_ {
-			return StreamConfig{}, NewJSMirrorWithSubjectFiltersError()
-		}
-		if cfg.Mirror.OptStartSeq > 0 && cfg.Mirror.OptStartTime != nil {
-			return StreamConfig{}, NewJSMirrorWithStartSeqAndTimeError()
-		}
-		if cfg.Duplicates != time.Duration(0) {
```
kozlovic, line 1015: Why removing this? (or check for startseq/starttime, although those I am not sure if it makes sense or not)

server/jetstream_cluster_test.go
derekcollison: Which issues? I think its fine if you want to create a mirror with dupe window..

server/stream.go
derekcollison: This should also be allowed. I only want to mirror last hour, or last 1k msgs.

ripienaar: Some chat on slack, but these are breaking changes that broke working code.  Now we support filters (didnt before) and I think dupe check is usable as long as we do the dupe check on the deltas of the original message timestamps

derekcollison: DeDupe imo does not need to look at original timestamp, its reference frame is when it receives a message at time T it will wait for de-dupe time window to see if another message arrives with same MsgId. Of course only stores in presence of MsgId to start with..

ripienaar: during initial sync - or just whenever there's some lag - you wll get messages arriving in mirror stream at a rate not aligned with the rate at which they arrived in source.

Presumbly the dupe window is time between mesages, and since the rate of delivery that matters there is the source rate I think source timestamp is needed?

ripienaar: @kozlovic ah - but i am sure we can handle that better than denying it.  The annoying thing here is that there's a default of 2m, clients might set that i am not sure of them all, but if they do now users must remember to specifically set it to 0 when creating a mirror which is not nice at all.

So mainly looking to a) make this not be annoying b) not be a breaking change as it used to work.

Specifically tools that create a config and later compare it - will now potentially see changes since previously they could set a dedup window that then later vanished and then later resulted in error.

kozlovic: @ripienaar I don't know the background, just that there had been issues and a change in server that once deployed would panic, etc.. I did contribute by adding the test, etc.. while working with Matthias on that. Will have to wait for @matthiashanel to comment.

derekcollison: ok, so we believe that mirrors should not do de-dupe, that is fine. I can revert, and if its also default just wipe but don't error yes?

matthiashanel: Meaning, unless there is another reason to not have dedupe window on a mirror, such as which timestamp to consider, this check can go.

ripienaar: Well I am not saying you must add that feature. I am saying if you want to do dedupe that’s all that make sense. 

I domy think it’s a priority support. I just want a nicer failure mode.

derekcollison: ok reverted last CL.

kozlovic (approved):
LGTM as long as RI or Matthias do not object.

PR #3232, opened 2022-06-30
Close leafnode connection when same cluster name detected

derekcollison:

/cc @nats-io/core

### DIFF
--- a/server/client.go
+++ b/server/client.go
@@ -206,6 +206,7 @@ const (
 	DuplicateClientID
 	DuplicateServerName
 	MinimumVersionRequired
+	ClusterNamesIdentical
 )
 
 // Some flags passed to processMsgResults
--- a/server/leafnode.go
+++ b/server/leafnode.go
@@ -1,4 +1,4 @@
-// Copyright 2019-2021 The NATS Authors
+// Copyright 2019-2022 The NATS Authors
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
 // You may obtain a copy of the License at
@@ -48,6 +48,9 @@ const leafNodeReconnectDelayAfterLoopDetected = 30 * time.Second
 // When a server receives a message causing a permission violation, the
 // connection is closed and it won't attempt to reconnect for that long.
 const leafNodeReconnectAfterPermViolation = 30 * time.Second
+
+// When we have the same cluster name as the hub.
+const leafNodeReconnectDelayAfterClusterNameSame = 30 * time.Second
 
 // Prefix for loop detection subject
 const leafNodeLoopDetectionSubjectPrefix = "$LDS."
@@ -1382,6 +1385,13 @@ func (c *client) processLeafNodeConnect(s *Server, arg []byte, lang string) erro
 		return err
 	}
 
+	// Check for cluster name collisions.
+	if cn := s.cachedClusterName(); cn != _EMPTY_ && proto.Cluster != _EMPTY_ && proto.Cluster == cn {
+		c.sendErrAndErr(ErrLeafNodeHasSameClusterName.Error())
+		c.closeConnection(ClusterNamesIdentical)
+		return ErrLeafNodeHasSameClusterName
+	}
+
 	// Reject if this has Gateway which means that it would be from a gateway
 	// connection that incorrectly connects to the leafnode port.
 	if proto.Gateway != _EMPTY_ {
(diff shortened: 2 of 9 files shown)


matthiashanel (commented):
LGTM, some tests fail due to usage of DefaultOptions

```
@@ -1262,6 +1262,7 @@ func TestLeafNodePermissions(t *testing.T) {

        u, _ := url.Parse(fmt.Sprintf("nats://%s:%d", lo1.LeafNode.Host, lo1.LeafNode.Port))
        lo2 := DefaultOptions()
+       lo2.Cluster.Name = "cde"
        lo2.LeafNode.ReconnectInterval = 5 * time.Millisecond
        lo2.LeafNode.connDelay = 100 * time.Millisecond
        lo2.LeafNode.Remotes = []*RemoteLeafOpts{
@@ -1405,6 +1406,7 @@ func TestLeafNodePermissionsConcurrentAccess(t *testing.T) {

        u, _ := url.Parse(fmt.Sprintf("nats://%s:%d", lo1.LeafNode.Host, lo1.LeafNode.Port))
        lo2 := DefaultOptions()
+       lo2.Cluster.Name = "cde"
        lo2.LeafNode.ReconnectInterval = 5 * time.Millisecond
        lo2.LeafNode.connDelay = 500 * time.Millisecond
        lo2.LeafNode.Remotes = []*RemoteLeafOpts{
@@ -1705,6 +1707,7 @@ func TestLeafNodeTLSVerifyAndMap(t *testing.T) {
        } {
                t.Run(test.name, func(t *testing.T) {
                        o := DefaultOptions()
+                       o.Cluster.Name = "cde"
                        o.Accounts = []*Account{acc}
                        o.LeafNode.Host = "127.0.0.1"
                        o.LeafNode.Port = -1
@@ -1969,7 +1972,7 @@ func TestLeafNodeOriginClusterInfo(t *testing.T) {
                        remotes [ { url: "nats://127.0.0.1:%d" } ]
                }
                cluster {
-                       name: "abc"
+                       name: "cde"
                        listen: "127.0.0.1:-1"
                }
        `, hopts.LeafNode.Port)))
@@ -1987,7 +1990,7 @@ func TestLeafNodeOriginClusterInfo(t *testing.T) {
        checkLeafNodeConnected(t, s)

        l = grabLeaf()
-       if rc := l.remoteCluster(); rc != "abc" {
+       if rc := l.remoteCluster(); rc != "cde" {
                t.Fatalf("Expected a remote cluster name of \"abc\", got %q", rc)
        }
        pcid := l.cid
@@ -4101,15 +4104,16 @@ leafnodes:{
        test(jsAA, jsLL)
 }
```

server/client.go
```diff
@@ -206,6 +206,7 @@ const (
 	DuplicateClientID
 	DuplicateServerName
 	MinimumVersionRequired
+	ClusterNamesIdentical
```
kozlovic, line 209: Add this in monitor.go (ClosedState String()):
```
	case ClusterNamesIdentical:
		return "Cluster Names Identical"
```

kozlovic (changes requested):

derekcollison:
Fixed test.

kozlovic:
This test seem to fail too, is that the one that was fixed? 
```
=== RUN   TestClientClampMaxSubsErrReport
    client_test.go:2534: Expected 1 connected leafnode(s) for server "NCZSORUN25U6ESFJWOYBYBFGSR62NZJLHN7BWUNBZJ2EKIVCIBWGREP5", got 0
--- FAIL: TestClientClampMaxSubsErrReport (5.06s)
```

server/client.go
derekcollison: ok done, thanks.

server/leafnode.go
```diff
@@ -2269,6 +2279,13 @@ func (c *client) leafPermViolation(pub bool, subj []byte) {
 
 // Invoked from generic processErr() for LEAF connections.
 func (c *client) leafProcessErr(errStr string) {
+	// Check if we got a cluster name collision.
+	if strings.Contains(errStr, "same cluster name") {
```
kozlovic, line 2283: I wonder if we should not be more strict here and use `ErrLeafNodeHasSameClusterName.Error()` so that we don't have another error some day that contains this string but is not really the error we are dealing with here?

derekcollison: I can make that change, sounds good.

kozlovic (approved):
LGTM, assuming that we are able to get a green on Travis.

derekcollison:
Will squash once green then merge.

ripienaar:
We will need to call this out in release notes warning people before upgrade it might end up splitting their networks and require config changes to fix that

kozlovic:
@ripienaar Added the the "CHANGED" section of the release notes, but noted I will probably a section at the top to point to that one.

smlx:
I'm trying to understand the implications of this change on my NATS configuration.

I see that "spoke" and "hub" cluster names for leaf clusters must now be unique. However does this mean that if there are multiple "spoke" leaf clusters they all require unique names? Or can leaf clusters have the same name as long as it is different to the "hub" cluster name?

Could anyone comment on the motivation for this change?

kozlovic:
@smlx The motivation is that cluster name is required to properly route messages between hub and spoke and with separate leaf clusters but with the same name, the hub would not be able to know where to route messages.

smlx:
Ok thanks. So that means every leaf cluster needs a unique name?

kozlovic:
> So that means every leaf cluster needs a unique name?

Yes.

PR #3234, opened 2022-07-01
fixes peer removal, simplifies move, more tests

matthiashanel:
fixes peer removal, simplifies move, more tests

Make sure when processing a peer removal that the stream assignment agrees.
When a new leader takes over it can resend a peer removal, and if the stream/consumer really was rescheduled we could remove by accident.

Also need to make sure that when we remove a stream we remove the node as part of the stream assignment.
If we didn't, if the same asset returned to this server we would not start up the monitoring loop.

Simplify migration logic in monitorStream, to be driven by leader only

Improved unit tests

Added failure when server not in peer list

Move command does not require server anymore

----

fixes FilterSubject crash

----

bumps server version

### DIFF
--- a/server/const.go
+++ b/server/const.go
@@ -41,7 +41,7 @@ var (
 
 const (
 	// VERSION is the current version for the server.
-	VERSION = "2.9.0-beta.2"
+	VERSION = "2.9.0-beta.3"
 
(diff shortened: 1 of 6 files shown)


derekcollison (approved):
LGTM

PR #3237, opened 2022-07-02
Allow maxprocs to debug log after start banner

derekcollison:

/cc @nats-io/core

### DIFF
--- a/main.go
+++ b/main.go
@@ -115,15 +115,17 @@ func main() {
 	// Configure the logger based on the flags
 	s.ConfigureLogger()
 
-	// Adjust MAXPROCS if running under linux/cgroups quotas.
-	// We ignore undo.
-	if _, err := maxprocs.Set(); err != nil {
-		server.PrintAndDie(fmt.Sprintf("failed to set GOMAXPROCS: %v", err))
-	}
-
 	// Start things up. Block here until done.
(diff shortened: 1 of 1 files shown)


main.go
```diff
@@ -115,15 +115,17 @@ func main() {
...
-
 	// Start things up. Block here until done.
 	if err := server.Run(s); err != nil {
 		server.PrintAndDie(err.Error())
 	}
+
+	// Adjust MAXPROCS if running under linux/cgroups quotas.
+	undo, err := maxprocs.Set(maxprocs.Logger(s.Debugf))
+	if err != nil {
+		server.PrintAndDie(fmt.Sprintf("failed to set GOMAXPROCS: %v", err))
```
ripienaar, line 126: is this really fatal? Seems like something we can pass over with a warning

derekcollison: From looking at the maxprocs this is really never expected to fail and if it does something is wrong, at least that is what I get the sense of looking at their code and all of their examples that consider it fatal.

I think we see if this ever fails in the wild, and with the server not starting we will have a higher probability of knowing about it then a warning, then adapt as needed.

ripienaar (approved):
LGTM

kozlovic (approved):
LGTM

PR #3238, opened 2022-07-03
Direct stream access enhancements

derekcollison:
As direct access becomes more important through materialized views like KeyValue, and having alternatives to a full blown JetStream consumers is important in future hybrid architectures, this PR introduces two enhancements.

Both of these are opt in and will be easily identified by clients.

1. DirectAccess now uses a queue group and the system allows non-leader replicas to participate.
2. Mirrors can now optionally be a part of the origin's group.

So for the second, this could allow a KV store with multiple mirrors to have a single subject to retrieve a message that operates in a single queue group. Since these are cluster aware, meaning we prefer the ones closer first and only move to a nearby cluster when no interest exists, adding a stream mirror when a large amount of traffic is direct retrieval, like geo diverse KVs, this would instantly improve latency and scale.

/cc @nats-io/core

### DIFF
--- a/server/jetstream_cluster.go
+++ b/server/jetstream_cluster.go
@@ -1707,6 +1707,7 @@ func (js *jetStream) monitorStream(mset *stream, sa *streamAssignment, sendSnaps
 	// For migration tracking.
 	var migrating bool
 	var peerGroup peerMigrateType
+
 	var mmt *time.Ticker
 	var mmtc <-chan time.Time
 
@@ -1720,10 +1721,30 @@ func (js *jetStream) monitorStream(mset *stream, sa *streamAssignment, sendSnaps
 	stopMigrationMonitoring := func() {
 		if mmt != nil {
 			mmt.Stop()
-			mmtc = nil
+			mmt, mmtc = nil, nil
 		}
 	}
 	defer stopMigrationMonitoring()
+
+	// This is to optionally track when we are ready as a non-leader for direct access participation.
+	// Either direct or if we are a direct mirror, or both.
+	var dat *time.Ticker
+	var datc <-chan time.Time
+
+	startDirectAccessMonitoring := func() {
+		if dat == nil {
+			dat = time.NewTicker(1 * time.Second)
+			datc = dat.C
+		}
+	}
+
+	stopDirectMonitoring := func() {
+		if dat != nil {
+			dat.Stop()
+			dat, datc = nil, nil
+		}
+	}
+	defer stopDirectMonitoring()
 
 	// This is triggered during a scale up from 1 to clustered mode. We need the new followers to catchup,
 	// similar to how we trigger the catchup mechanism post a backup/restore. It's ok to do here and preferred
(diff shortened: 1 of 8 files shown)


server/jetstream_cluster.go
```diff
@@ -1806,6 +1830,48 @@ func (js *jetStream) monitorStream(mset *stream, sa *streamAssignment, sendSnaps
...
+			// We are current, cancel monitoring and create the direct subs as needed.
+			if ad {
+				mset.subscribeToDirect()
+			}
+			if md {
+				mset.subscribeToMirrorDirect()
+			}
+			mset.mu.Unlock()
+			// Stop monitoring.
+			stopDirectMonitoring()
```
ripienaar, line 1873: So this monitoring is good but if I read it right its only activated on leader change, stops once caught up and then never checked again?  What if a node lags later? Likewise later where you check based on dgetCaughtUpThresh.

Really concerned about read after write consistency here for KV - so this wouold have to be a opt-in for specific KV cases where thats acceptible? When I designed KV you were pretty much saying read after write consistency is non negotiable - even the cache design being pass through not write through had to maintain that.

derekcollison: I have been thinking about that and whether or not we want to do anything there. We do not do anything like that for normal consumers, at least so far. If the server's connection is broke to a route etc the subscription will be dropped, meaning this subscriber will not participate while disconnected.

This is opt-in yes, and read after write could be an issue, and yes I was the one arguing about it for sure. I wonder if the use case is simply not common. I was using gmail as a use case, but they do not even do that now and short circuit. Meaning is you were keeping a list of keys and you updated one, either you would hand update the list, or update it once a watcher fired. Not sure you would do put then get right behind it. WDYT?

ripienaar: Agree read-after-write is not as important as initally imagined, however its out there now and cannot change by default.  

At best we can have an opt in to relax that constraint for the more kv abused as large pub-sub with cache behavior people are using it for thats causing the problems.

derekcollison: I was leaning for making this default and you could op-out. With this approach you can just add mirrors and things get better, if its an opt-in all consumers would need to be updated before the mirrors would be used for get operations.

ripienaar: Read-after-write aside, we definitely need to know its not using a !current mirror, definitely think that requires some more thought - mirrors especially can silently and forever stop working should the source be replaced (which. makes sense of course) but without that scenario being detected and that mirror declaring itself not a read candidate its trouble.

Same scenario with excessive lag or just long term badness

server/stream.go
```diff
@@ -2854,6 +2875,36 @@ func (mset *stream) subscribeToStream() error {
...
+}
+
+// Lock should be held.
+func (mset *stream) subscribeToMirrorDirect() error {
+	if mset.mirror == nil || mset.mirror.dsub != nil {
+		return nil
+	}
+	dsubj := fmt.Sprintf(JSDirectMsgGetT, mset.mirror.name)
+	// We will make this listen on a queue group by default, which can allow mirrors to participate on opt-in basis.
+	if sub, err := mset.queueSubscribeInternal(dsubj, dgetGroup, mset.processDirectGetRequest); err != nil {
```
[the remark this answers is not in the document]
derekcollison: Not by default, but you could export I think.

PR #3241, opened 2022-07-04
Make pull consumers FIFO per message, not per request.

derekcollison:
This effectively means that requests with batch > 1 will process a message and go to the end of the line.

Ivan, this was considered incorrect behavior hence no option etc.

/cc @nats-io/core


ripienaar (commented):
LGTM

tbeets (approved):
LGTM

scottf (approved):
I ran my test against this version and the threads got messages distributed perfectly instead of blocks. LGTM

PR #3243, opened 2022-07-05
Direct access to JetStream resources when a leafnode connection is down.

### LINKED ISSUE #3178: Remote PULL consumer SPOF
## Problem description

In a system where two NATS clusters are connected using leaf node connections, clients connected to the remote NATS cluster cannot consume messages from a stream when the NATS server hosting the consumer leader is unable to establish leaf node connections due to network failure. In this scenario a connection to the remote system is still available from other nodes in the cluster but messages cannot be routed to the consumer leader resulting in an unexpected point of failure for the consuming application.

## More Details

image

To consume message from the consumer C attached to the stream S, the Application publishes NATS messages on the subject `$JS.a.API.CONSUMER.MSG.NEXT.S.C`
Messages published from the application in Cluster-B cannot reach the consumer leader which resides on `Cluster-A.nats-server-1`.
Current NATS routing does not allow message forwarding within the Cluster-A even if one instance of the Cluster-A has a healthy leaf connection to the Cluster-B.

## Remediation

The current remediation for this is to manually move the consumer leader to a node in the cluster that still has leaf node connections, while this is possible to automate externally it would be a beneficial enhancement to JetStream to have this as a built in feature.

derekcollison:
This allows a soliciting leafnode config to ask that any JetStream cluster assets that are a current leader have the leader stepdown. This will allow a new leader to take over and service direct access.

Resolves #3178 

/cc @nats-io/core

### DIFF
--- a/server/jetstream_helpers_test.go
+++ b/server/jetstream_helpers_test.go
@@ -924,6 +924,31 @@ func (c *cluster) createLeafNodesWithTemplateAndStartPort(template, clusterName 
 	return lc
 }
 
+// Helper function to close and disable leafnodes.
+func (s *Server) closeAndDisableLeafnodes() {
+	var leafs []*client
+	s.mu.Lock()
+	for _, ln := range s.leafs {
+		leafs = append(leafs, ln)
+	}
+	// Disable leafnodes for now.
+	s.leafNodeEnabled = false
+	s.mu.Unlock()
+
+	for _, ln := range leafs {
+		ln.closeConnection(Revocation)
+	}
+}
+
+// Helper to set the remote migrate feature.
+func (s *Server) setJetStreamMigrateOnRemoteLeaf() {
+	s.mu.Lock()
+	for _, cfg := range s.leafRemoteCfgs {
+		cfg.JetStreamClusterMigrate = true
+	}
+	s.mu.Unlock()
+}
+
 // Will add in the mapping for the account to each server.
 func (c *cluster) addSubjectMapping(account, src, dest string) {
 	c.t.Helper()
--- a/server/leafnode.go
+++ b/server/leafnode.go
@@ -495,8 +495,13 @@ func (s *Server) connectToRemoteLeafNode(remote *leafNodeCfg, firstConnect bool)
 			if url != rURL.Host {
 				ipStr = fmt.Sprintf(" (%s)", url)
 			}
-			s.Debugf("Trying to connect as leafnode to remote server on %q%s", rURL.Host, ipStr)
-			conn, err = natsDialTimeout("tcp", url, dialTimeout)
+			if s.isLeafNodeEnabled() {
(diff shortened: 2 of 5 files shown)


server/leafnode.go
```diff
@@ -509,6 +514,8 @@ func (s *Server) connectToRemoteLeafNode(remote *leafNodeCfg, firstConnect bool)
 			case <-s.quitCh:
 				return
 			case <-time.After(reconnectDelay):
+				// Check if we should migrate any JetStream assets while this remote is down.
+				s.checkJetStreamMigrate(remote)
```
kozlovic, line 518: But if we are not connected to the rest of the network, how can migration work?

derekcollison: Check the issue referenced. We are just moving the leader such that the other side can access again.

derekcollison: Maybe should rename, just migrating the leaders.

server/leafnode.go
```diff
@@ -495,8 +495,13 @@ func (s *Server) connectToRemoteLeafNode(remote *leafNodeCfg, firstConnect bool)
 			if url != rURL.Host {
 				ipStr = fmt.Sprintf(" (%s)", url)
 			}
-			s.Debugf("Trying to connect as leafnode to remote server on %q%s", rURL.Host, ipStr)
-			conn, err = natsDialTimeout("tcp", url, dialTimeout)
+			if s.isLeafNodeEnabled() {
+				s.Debugf("Trying to connect as leafnode to remote server on %q%s", rURL.Host, ipStr)
+				conn, err = natsDialTimeout("tcp", url, dialTimeout)
+			} else {
+				s.Debugf("Will not attempt to connect to remote server on %q%s, leafnodes currently disabled", rURL.Host, ipStr)
```
kozlovic, line 502: If I read the code correctly, it means that this server will never retry to connect since you are setting "leafnodeEnabled" to false in a test helper and that value is never set back to true.
So again, if I am correct, this looks more like a "test-only" condition and wonder if it could be done differently, but I don't have other recommendation at the moment.

server/jetstream_helpers_test.go
```diff
@@ -924,6 +924,31 @@ func (c *cluster) createLeafNodesWithTemplateAndStartPort(template, clusterName
...
+	s.leafNodeEnabled = false
+	s.mu.Unlock()
+
+	for _, ln := range leafs {
+		ln.closeConnection(Revocation)
+	}
+}
+
+// Helper to set the remote migrate feature.
+func (s *Server) setJetStreamMigrateOnRemoteLeaf() {
```
kozlovic, line 944: This is so that you don't have to modify the test's template config to have the new config param, correct?

derekcollison: Correct.

kozlovic (approved):
LGTM

PR #3247, opened 2022-07-06
Allows direct get to also do get next for subject with starting sequence

derekcollison:
This continues to allow more diversity in hybrid setups combining JetStream and Core.
Also is Request() mux friendly.

/cc @nats-io/core

### DIFF
--- a/server/jetstream_api.go
+++ b/server/jetstream_api.go
@@ -600,6 +600,7 @@ type JSApiMetaServerStreamMoveRequest struct {
 type JSApiMsgGetRequest struct {
 	Seq     uint64 `json:"seq,omitempty"`
 	LastFor string `json:"last_by_subj,omitempty"`
+	NextFor string `json:"next_by_subj,omitempty"`
 }
 
 type JSApiMsgGetResponse struct {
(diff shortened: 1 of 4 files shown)


kozlovic (approved):
LGTM, but so difficult to get green these days... (even locally)

derekcollison:
I will increase time it waits. Agree its been a struggle.

kozlovic:
> I will increase time it waits

It was already waiting for 10sec.. not sure if increasing will help but let's see..

derekcollison:
Something else seems off, will dig in a bit more.

kozlovic (approved):
LGTM

derekcollison:
Correct it was not added to the old way. I think the old way should stay as is, WDYT?

ripienaar:
It’s useful and since it takes the same request structure it should either fail saying unsupported or do the right thing - I would prefer it supports this field personally

derekcollison:
Well, for older servers it will silently fail, so thought was it should not be supported at client level.
As you point out, similar processing, so not hard to add to old path as well.

ripienaar:
Long term consistency is best - else this will always be a little weird edge case no-one remembers, where it seems like adding it there would be small effort to gain the consistency

derekcollison:
Long term this method will no longer be used.

PR #3249, opened 2022-07-07
Fix for stalled catchup in endless cycle on EOF 

derekcollison:
A customer experienced and endless failure to have a stream catchup. The current leader was being asked for a message from a snapshot that was larger than what we had, resulting in EOF which silently failed.
We now detect this and signal end of catchup and redo the bad snapshot if possible.

/cc @nats-io/core


kozlovic (approved):
LGTM

PR #3250, opened 2022-07-08
[Added] account specific monitoring endpoint(s)

matthiashanel:
Added http monitoring endpoint /accstatz
It responds with a list of statz for all accounts with local connections
the argument "unused=1" can be provided to get statz for all accounts
This endpoint is also exposed as nats request under:

This monitoring endpoint is exposed via the system account.
`$SYS.REQ.ACCOUNT.*.STATZ`
Each server will respond with connection statistics for the requested
account. The format of the data section is a list (size 1) identical to the event
`$SYS.ACCOUNT.%s.SERVER.CONNS` which is sent periodically as well as on
connect/disconnect. Unless requested by options, server without the account, 
or server where the account has no local connections, will not respond. 

A PING endpoint exists as well. The response format is identical to
$SYS.REQ.ACCOUNT.*.STATZ
(however the data section will contain more than one account, if they exist)
In addition to general filter options the request takes a list of accounts and
an argument to include accounts without local connections (disabled by default)
`$SYS.REQ.ACCOUNT.PING.STATZ`

Each account has a new system account import where the local subject
`$SYS.REQ.ACCOUNT.PING.STATZ` essentially responds as if
the importing account name was used for `$SYS.REQ.ACCOUNT.*.STATZ`

The only difference between requesting `ACCOUNT.PING.STATZ` from within
the system account and an account is that the later can only retrieve
statz for the account the client requests from.

Also exposed the monitoring /healthz via the system account under
`$SYS.REQ.SERVER.*.HEALTHZ`
`$SYS.REQ.SERVER.PING.HEALTHZ`
No dedicated options are available for these.
`HEALTHZ` also accept general filter options.

### DIFF
--- a/server/events.go
+++ b/server/events.go
@@ -42,7 +42,8 @@ const (
 
 	connectEventSubj    = "$SYS.ACCOUNT.%s.CONNECT"
 	disconnectEventSubj = "$SYS.ACCOUNT.%s.DISCONNECT"
-	accReqSubj          = "$SYS.REQ.ACCOUNT.%s.%s"
+	accDirectReqSubj    = "$SYS.REQ.ACCOUNT.%s.%s"
+	accPingReqSubj      = "$SYS.REQ.ACCOUNT.PING.%s" // atm. only used for STATZ and CONNZ import from system account
 	// kept for backward compatibility when using http resolver
 	// this overlaps with the names for events but you'd have to have the operator private key in order to succeed.
 	accUpdateEventSubjOld    = "$SYS.ACCOUNT.%s.CLAIMS.UPDATE"
@@ -59,7 +60,6 @@ const (
 	leafNodeConnectEventSubj = "$SYS.ACCOUNT.%s.LEAFNODE.CONNECT" // for internal use only
 	remoteLatencyEventSubj   = "$SYS.LATENCY.M2.%s"
 	inboxRespSubj            = "$SYS._INBOX.%s.%s"
-	accConnzReqSubj          = "$SYS.REQ.ACCOUNT.PING.CONNZ"
 
 	// FIXME(dlc) - Should account scope, even with wc for now, but later on
 	// we can then shard as needed.
@@ -138,14 +138,19 @@ const DisconnectEventMsgType = "io.nats.server.advisory.v1.client_disconnect"
 // updates in the absence of any changes.
 type AccountNumConns struct {
 	TypedEvent
-	Server        ServerInfo `json:"server"`
-	Account       string     `json:"acc"`
-	Conns         int        `json:"conns"`
-	LeafNodes     int        `json:"leafnodes"`
-	TotalConns    int        `json:"total_conns"`
-	Sent          DataStats  `json:"sent"`
-	Received      DataStats  `json:"received"`
-	SlowConsumers int64      `json:"slow_consumers"`
+	Server ServerInfo `json:"server"`
+	*AccountStat
+}
+
+// AccountStat contains the data common between AccountNumConns and AccountStatz
+type AccountStat struct {
+	Account       string    `json:"acc"`
+	Conns         int       `json:"conns"`
+	LeafNodes     int       `json:"leafnodes"`
+	TotalConns    int       `json:"total_conns"`
+	Sent          DataStats `json:"sent"`
+	Received      DataStats `json:"received"`
+	SlowConsumers int64     `json:"slow_consumers"`
 }
 
 const AccountNumConnsMsgType = "io.nats.server.advisory.v1.account_connections"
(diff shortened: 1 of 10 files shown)


server/events.go
```diff
@@ -138,14 +138,19 @@ const DisconnectEventMsgType = "io.nats.server.advisory.v1.client_disconnect"
...
-	Server        ServerInfo `json:"server"`
-	Account       string     `json:"acc"`
-	Conns         int        `json:"conns"`
-	LeafNodes     int        `json:"leafnodes"`
-	TotalConns    int        `json:"total_conns"`
-	Sent          DataStats  `json:"sent"`
-	Received      DataStats  `json:"received"`
-	SlowConsumers int64      `json:"slow_consumers"`
+	Server ServerInfo `json:"server"`
+	*AccountStat
```
kozlovic, line 142: Why `*AccountStat` and not `AccountStat`? Less change of breaking existing apps relying on fields to always be present. I see that accConns() always return &AccountStat{} so it is not like there are cases where the returned object would be smaller.

server/events.go
```diff
@@ -903,17 +920,9 @@ func (s *Server) initEventTracking() {
...
+				if acc, err := extractAccount(c, subject, msg); err != nil {
 					return nil, err
 				} else {
-					if ci, _, _, _, err := c.srv.getRequestInfo(c, msg); err == nil && ci.Account != _EMPTY_ {
-						// Make sure the accounts match.
-						if ci.Account != acc {
-							// Do not leak too much here.
-							return nil, fmt.Errorf("bad request")
-						}
-						optz.ConnzOptions.isAccountReq = true
```
kozlovic, line 915: Since this is removed, the whole field isAccountReq in ConnzOptions has no relevance.

server/events.go
```diff
@@ -1005,10 +1053,14 @@ func (s *Server) addSystemAccountExports(sacc *Account) {
 	if !s.EventsEnabled() {
 		return
 	}
-	accConnzSubj := fmt.Sprintf(accReqSubj, "*", "CONNZ")
+	accConnzSubj := fmt.Sprintf(accDirectReqSubj, "*", "CONNZ")
 	if err := sacc.AddServiceExportWithResponse(accConnzSubj, Streamed, nil); err != nil {
 		s.Errorf("Error adding system service export for %q: %v", accConnzSubj, err)
 	}
+	accStatzSubj := fmt.Sprintf(accDirectReqSubj, "*", "STATZ")
```
kozlovic, line 1060: This is weird.. from reading more in the PR it looks like the account is located in index 3 (4th token), which then can be "PING" in accPingReqSubj. So do we have situations where "PING.STATZ" we don't know if it is a PING or it prevents account named "PING"?

server/events.go
```diff
@@ -1643,28 +1701,16 @@ func (s *Server) sendAccConnsUpdate(a *Account, subj ...string) {
...
-			Bytes: atomic.LoadInt64(&a.inBytes)},
-		Sent: DataStats{
-			Msgs:  atomic.LoadInt64(&a.outMsgs),
-			Bytes: atomic.LoadInt64(&a.outBytes)},
-		SlowConsumers: atomic.LoadInt64(&a.slowConsumers),
+		AccountStat: a.accConns(),
 	}
 	// Set timer to fire again unless we are at zero.
-	if localConns == 0 {
+	if m.TotalConns == 0 {
```
kozlovic, line 1713: I am planning on submitting a PR that exclude an account that has JS configured: this is because this event is used to make node "alive" and the absence of those updates cause a server to consider the remote server orphan and switch it to offline. Not sure if that causes a problem with your changes in this PR.

server/events.go
```diff
@@ -1681,6 +1727,25 @@ func (s *Server) sendAccConnsUpdate(a *Account, subj ...string) {
 	a.mu.Unlock()
 }
 
+// Lock shoulc be held on entry
+func (a *Account) accConns() *AccountStat {
```
kozlovic, line 1731: Since this is an `Account`, I would rename `accConns()` - since we know it is an account - to either conns() or connStats() or something like that.

server/events.go
matthiashanel: I probably should incorporate your change then

server/events.go
matthiashanel: Yes this happens, which is why there is a filter for ping.
This happens for server too, there it's not happening as every server subscribes to the exact name.
Not so for accounts, which is why this filter got added.
```
} else if acc == "PING" { // Filter ping subject, this happens for server too, but filter is less explicit
```

server/events.go
kozlovic: I can rebase/resolve if yours is committed first. The question I had is: is this PR fundamentally change the fact that the update is sent or not and on existing subject that the server uses to receive conns updates.

server/events.go
matthiashanel: fixed

server/events.go
matthiashanel: deleted isAccountReq and it's usage. 
I believe it's only occurrence was redundant with account being set or not.
That usage adjusted the Total, and I believe this should happen even when the request for an account comes from within the system account.

matthiashanel: deleted isAccountReq and it's usage.
I believe it's only occurrence was redundant with account being set or not.
That usage adjusted the Total, and I believe this should happen even when the request for an account comes from within the system account.

server/events.go
matthiashanel: Changed it to AccountStat. But kept it `*` in other occurances as `[]*` is used in monitoring a lot.
Example in connz
```
// Connz represents detailed information on current client connections.
type Connz struct {
	ID       string      `json:"server_id"`
	Now      time.Time   `json:"now"`
	NumConns int         `json:"num_connections"`
	Total    int         `json:"total"`
	Offset   int         `json:"offset"`
	Limit    int         `json:"limit"`
	Conns    []*ConnInfo `json:"connections"`
}

```

server/events.go
matthiashanel: This is the same for server and it's ping counter part.
There it's not observed as each server subscribes to it's ID, and is not using a wildcard.
STATZ call for an account filters when the name is PING

server/events.go
kozlovic: My comment here was because AccountNumConns{} already existed and so changing from the fields to embedding `*AccountStat` could cause existing apps to crash if one day `*AccountStat` is nil. Had it been never released, I would not even have commented.

server/events.go
matthiashanel: For me it boils down to if I should even filter accounts when the have no local connections or js.
This is what `IncludeUnused` is about. And I'm wondering if it's needed.
```
// LeafzOptions are options passed to Leafz
type AccountStatzOptions struct {
	Accounts      []string `json:"accounts"`
	IncludeUnused bool     `json:"include_unused"`
}
```

server/events.go
matthiashanel: renamed to `statz` as to not overlap with `stats` inside the account

kozlovic:
@matthiashanel You may want to do a rebase from updated main. After that, I can have a last look and LGTM it!

kozlovic (approved):
LGTM

PR #3253, opened 2022-07-08
Test move undo

matthiashanel:
 - [ ] Tests added

Resolves #

 -
 -
 -

/cc @nats-io/core

### DIFF
--- a/server/jetstream_api.go
+++ b/server/jetstream_api.go
@@ -193,12 +193,14 @@ const (
 	// JSApiServerStreamMove is the endpoint to move streams off a server
 	// Only works from system account.
 	// Will return JSON response.
-	JSApiServerStreamMove = "$JS.API.SERVER.STREAM.MOVE"
+	JSApiServerStreamMove  = "$JS.API.ACCOUNT.STREAM.MOVE.*.*"
+	JSApiServerStreamMoveT = "$JS.API.ACCOUNT.STREAM.MOVE.%s.%s"
 
 	// JSApiServerStreamCancelMove is the endpoint to cancel a stream move
 	// Only works from system account.
 	// Will return JSON response.
-	JSApiServerStreamCancelMove = "$JS.API.SERVER.STREAM.CANCEL_MOVE"
+	JSApiServerStreamCancelMove  = "$JS.API.ACCOUNT.STREAM.CANCEL_MOVE.*.*"
+	JSApiServerStreamCancelMoveT = "$JS.API.ACCOUNT.STREAM.CANCEL_MOVE.%s.%s"
 
 	// jsAckT is the template for the ack message stream coming back from a consumer
 	// when they ACK/NAK, etc a message.
@@ -580,14 +582,8 @@ type JSApiMetaServerStreamMoveRequest struct {
 	Cluster string `json:"cluster,omitempty"`
 	// Domain the sever is in
 	Domain string `json:"domain,omitempty"`
-	// Account name the stream to move is in
-	Account string `json:"account"`
-	// Name of stream to move
-	Stream string `json:"stream"`
 	// Ephemeral placement tags for the move
 	Tags []string `json:"tags,omitempty"`
-	// Cancel a previous request
-	Cancel bool `json:"cancel,omitempty"`
 }
 
 // JSApiMsgGetRequest get a message request.
@@ -2203,6 +2199,24 @@ func (s *Server) peerSetToNames(ps []string) []string {
 	return names
 }
 
+// looks up the peer id for a given server name. Cluster and domain name are optional filter criteria
+func (s *Server) nameToPeer(js *jetStream, serverName, clusterName, domainName string) string {
+	js.mu.RLock()
+	cc := js.cluster
+	defer js.mu.RUnlock()
+	for _, p := range cc.meta.Peers() {
+		si, ok := s.nodeToInfo.Load(p.ID)
+		if ok && si.(nodeInfo).name == serverName {
+			if clusterName == _EMPTY_ || clusterName == si.(nodeInfo).cluster {
+				if domainName == _EMPTY_ || domainName == si.(nodeInfo).domain {
+					return p.ID
+				}
+			}
+		}
+	}
+	return _EMPTY_
+}
+
 // Request to have the metaleader move a stream on a peer to another
 func (s *Server) jsLeaderServerStreamMoveRequest(sub *subscription, c *client, _ *Account, subject, reply string, rmsg []byte) {
 	if c == nil || !s.JetStreamEnabled() {
@@ -2229,13 +2243,10 @@ func (s *Server) jsLeaderServerStreamMoveRequest(sub *subscription, c *client, _
 		return
 	}
 
-	var resp = JSApiStreamUpdateResponse{ApiResponse: ApiResponse{Type: JSApiStreamUpdateResponseType}}
+	accName := tokenAt(subject, 6)
+	streamName := tokenAt(subject, 7)
 
-	if isEmptyRequest(msg) {
-		resp.Error = NewJSBadRequestError()
-		s.sendAPIErrResponse(ci, acc, subject, reply, string(msg), s.jsonResponse(&resp))
-		return
-	}
+	var resp = JSApiStreamUpdateResponse{ApiResponse: ApiResponse{Type: JSApiStreamUpdateResponseType}}
 
 	var req JSApiMetaServerStreamMoveRequest
 	if err := json.Unmarshal(msg, &req); err != nil {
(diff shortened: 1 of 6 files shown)


server/jetstream_cluster_test.go
```diff
@@ -3813,7 +3813,7 @@ func TestJetStreamClusterDoubleStreamReassignment(t *testing.T) {
 	moveAndCheck := func(from, to string, expectedSet ...string) {
 		move(from, to)
 		checkFor(t, 20*time.Second, 100*time.Millisecond, func() error { return moveComplete(to, expectedSet...) })
-		checkFor(t, 20*time.Second, 100*time.Millisecond, func() error { return serverEmpty(from) })
+		checkFor(t, 120*time.Second, 100*time.Millisecond, func() error { return serverEmpty(from) })
```
kozlovic, line 3816: Seem a lot... Have you seen it that it passed after getting close to (but below) 120 secs?

server/jetstream_api.go
```diff
@@ -193,7 +193,14 @@ const (
 	// JSApiServerStreamMove is the endpoint to move streams off a server
 	// Only works from system account.
 	// Will return JSON response.
-	JSApiServerStreamMove = "$JS.API.SERVER.STREAM.MOVE"
+	JSApiServerStreamMove  = "$JS.API.ACCOUNT.STREAM.MOVE.*.*"
```
kozlovic, line 196: Why the choice of removing account and stream from the request object and instead making part of the subject?

server/jetstream_api.go
```diff
@@ -2341,6 +2361,92 @@ func (s *Server) jsLeaderServerStreamMoveRequest(sub *subscription, c *client, _
 
 	cfg.Placement = origPlacement
 
+	s.Noticef("Requested move of: R=%d stream %s in account %s from old peer set %+v to new peer set %+v",
```
kozlovic, line 2364: Derek often recommends the use of `'%s > %s'` (with account and stream) to be consistent when we issue JS log statements that contain those.

server/jetstream_api.go
```diff
@@ -2341,6 +2361,92 @@ func (s *Server) jsLeaderServerStreamMoveRequest(sub *subscription, c *client, _
...
+		cfg.Replicas, streamName, accName, s.peerSetToNames(currPeers), s.peerSetToNames(peers))
+
+	// execute on a go routine for route or gateway, as this may block waiting from responses from other servers and
+	// therefore block all route/gateway traffic.
+	if c.kind != ROUTER && c.kind != GATEWAY {
```
kozlovic, line 2369: Always wonder if we need LEAF in those decisions?

server/jetstream_api.go
```diff
@@ -2341,6 +2361,92 @@ func (s *Server) jsLeaderServerStreamMoveRequest(sub *subscription, c *client, _
...
+		go s.jsClusteredStreamUpdateRequest(&ciNew, targetAcc.(*Account), subject, reply, rmsg, &cfg, peers)
```
kozlovic, line 2370: Need to make sure you have copies since you are running that in go routine. So for sure, replace `rmsg` with `copyBytes(rmsg)`. Could you do the same in existing `jsStreamUpdateRequest()`?

server/jetstream_api.go
```diff
@@ -2341,6 +2361,92 @@ func (s *Server) jsLeaderServerStreamMoveRequest(sub *subscription, c *client, _
...
+		return
+	}
+
+	// make sure client is scoped to requested account
+	ciNew := *(ci)
+	ciNew.Account = streamName
+
+	peers := currPeers[:cfg.Replicas]
+
+	s.Noticef("Requested cancel of move: R=%d stream %s in account %s from old peer set %+v to new peer set %+v",
```
kozlovic, line 2447: Same comment than above regarding acc/stream output

server/jetstream_api.go
```diff
@@ -2341,6 +2361,92 @@ func (s *Server) jsLeaderServerStreamMoveRequest(sub *subscription, c *client, _
...
+		cfg.Replicas, streamName, accName, s.peerSetToNames(currPeers), s.peerSetToNames(peers))
+
```
kozlovic, line 2449: Also, please use copyBytes(rmsg) below..

kozlovic (changes requested):
First pass...

server/jetstream_api.go
matthiashanel: This expanded from one subject to two and I know that there will be a similar one going forward.
It seemed better to have that align with stream/consumer as well and identify the resource in the subject.

server/jetstream_api.go
matthiashanel: fixed

server/jetstream_api.go
matthiashanel: I copied that from update and just did the search for this check:
update only checks for router/Gateway, direct get/next checks leaf too.
```
    Occurrences of 'c.kind != ROUTER && c.kind != GATEWAY' in Project with mask '*.go'
Found Occurrences in Project with mask '*.go'  (6 usages found)
    nats-server  (6 usages found)
        server  (6 usages found)
            consumer.go  (1 usage found)
                2611 inlineOk := c.kind != ROUTER && c.kind != GATEWAY && c.kind != LEAF
            jetstream_api.go  (4 usages found)
                732 if c.kind != ROUTER && c.kind != GATEWAY {
                1398 if c.kind != ROUTER && c.kind != GATEWAY {
                2369 if c.kind != ROUTER && c.kind != GATEWAY {
                2452 if c.kind != ROUTER && c.kind != GATEWAY {
            stream.go  (1 usage found)
                3299 inlineOk := c.kind != ROUTER && c.kind != GATEWAY && c.kind != LEAF
```

server/jetstream_api.go
matthiashanel: fixed

server/jetstream_api.go
matthiashanel: fixed

server/jetstream_api.go
matthiashanel: That copy does not need to be made


This is the code where I initially copied this from.
It does not copy either.
```
func (s *Server) jsStreamUpdateRequest(sub *subscription, c *client, _ *Account, subject, reply string, rmsg []byte) {
	...
	// Handle clustered version here.
	if s.JetStreamIsClustered() {
		// If we are inline with client, we still may need to do a callout for stream info
		// during this call, so place in Go routine to not block client.
		if c.kind != ROUTER && c.kind != GATEWAY {
			go s.jsClusteredStreamUpdateRequest(ci, acc, subject, reply, rmsg, &cfg, nil)
		} else {
			s.jsClusteredStreamUpdateRequest(ci, acc, subject, reply, rmsg, &cfg, nil)
		}
		return
	}
```

I believe the copy in apiDispatch is the reason the rmsg is not being copied in jsStreamUpdateRequest.

```
func (js *jetStream) apiDispatch(sub *subscription, c *client, acc *Account, subject, reply string, rmsg []byte) {
	// No lock needed, those are immutable.
	s, rr := js.srv, js.apiSubs.Match(subject)
...

	// If this is directly from a client connection ok to do in place.
	if c.kind != ROUTER && c.kind != GATEWAY {
		jsub.icb(sub, c, acc, subject, reply, rmsg)
		return
	}

	// If we are here we have received this request over a non client connection.
	// We need to make sure not to block. We will send the request to a long-lived
	// go routine.

	s.jsAPIRoutedReqs.push(&jsAPIRoutedReq{jsub, sub, acc, subject, reply, copyBytes(rmsg), c.pa})
}
```
As jsStreamUpdateRequest is called via apiDispatch

```
	// API handles themselves.
	pairs := []struct {
		subject string
		handler msgHandler
	}{
...
		{JSApiStreamUpdate, s.jsStreamUpdateRequest},
...
	}

	js.mu.Lock()
	defer js.mu.Unlock()

	for _, p := range pairs {
		sub := &subscription{subject: []byte(p.subject), icb: p.handler}
		if err := js.apiSubs.Insert(sub); err != nil {
			return err
		}
	}
```

server/jetstream_api.go
matthiashanel: don't think copy is needed

server/jetstream_api.go
kozlovic: Well, if this is invoked by apiDispatch and we are already making a copy and routing to a single go routine, then we should not "go" run that again here, don't you think?

But jsLeaderServerStreamMoveRequest itself seem to be attached directly to a sys sub, and then you MUST make a copy:
```
	if cc.peerStreamMove == nil {
		cc.peerStreamMove, _ = s.systemSubscribe(JSApiServerStreamMove, _EMPTY_, false, c, s.jsLeaderServerStreamMoveRequest)
	}
```

server/jetstream_api.go
```diff
@@ -2444,11 +2444,11 @@ func (s *Server) jsLeaderServerStreamCancelMoveRequest(sub *subscription, c *cli
...
 	peers := currPeers[:cfg.Replicas]
 
-	s.Noticef("Requested cancel of move: R=%d stream %s in account %s from old peer set %+v to new peer set %+v",
+	s.Noticef("Requested cancel of move: R=%d '%s > %s' from old peer set %+v to new peer set %+v",
 		cfg.Replicas, streamName, accName, s.peerSetToNames(currPeers), s.peerSetToNames(peers))
 
-	// execute on a go routine for route or gateway, as this may block waiting from responses from other servers and
-	// therefore block all route/gateway traffic.
+	// If we are inline with client, we still may need to do a callout for stream info
+	// during this call, so place in Go routine to not block client.
```
kozlovic, line 2451: Again, here I don't think jsLeaderServerStreamCancelMoveRequest is invoked from apiDispatch, but directly subscribed to:
```
if cc.peerStreamCancelMove == nil {
		cc.peerStreamCancelMove, _ = s.systemSubscribe(JSApiServerStreamCancelMove, _EMPTY_, false, c, s.jsLeaderServerStreamCancelMoveRequest)
	}
```

PR #3257, opened 2022-07-11
enhanced unit test to check for leader post move

matthiashanel:
using waitOnStreamLeader to check for leader as it has a bigger timeout value.


kozlovic (approved):
LGTM

PR #3258, opened 2022-07-11
[FIXED] JetStream: servers may be reported as orphaned

kozlovic:
In some situations, a server may report that a remote server is
detected as orphaned (and the node is marked as offline). This is
because the orphaned detection relies on conns update to be received,
however, servers would suppress the update if an account does not
have any connections attached.

This PR ensures that the update is sent regardless if the account
is JS configured (not necessarily enabled at the moment).


kozlovic:
Not sure if we want to force the update or have the orphaned detection have its own update mechanism...

derekcollison (approved):
LGTM - Thanks.

matthiashanel (approved):
LGTM

Issue #3264, opened 2022-07-14
JS consumer Maximum Deliveries update has no effect


abegaj:
#### Versions of `nats-server` and affected client libraries used:

- `nats-server: 2.8.4 `
- `nats.go: 1.16.0 `

#### OS/Container environment:

- `docker`
- `macOS Monterey`

#### Steps or code to reproduce the issue:

1. Create stream
2. Create JS consumer on stream with Maximum Deliveries set to **2**
3. Verify that messages are not delivered more that **2 times**
4. Update JS consumer and set Maximum Deliveries to **4**

#### Expected result:

- Messages are delivered max **4 times**

#### Actual result:

- Messages are still delivered max **2 times**

#### Proposed changes:
 
- On `updateConfig` function, set `o.maxdc` to the new value
- Draft PR https://github.com/nats-io/nats-server/pull/3265

PR #3265, opened 2022-07-14
[FIXED] Maximum Deliveries update has no effect

abegaj:
Resolves https://github.com/nats-io/nats-server/issues/3264

 - On update check if BackOff is defined, MaxDeliver is within range
 - Set `o.maxdc` to the new value
 - Add tests

### DIFF
--- a/server/consumer.go
+++ b/server/consumer.go
@@ -1418,6 +1418,11 @@ func (acc *Account) checkNewConsumerConfig(cfg, ncfg *ConsumerConfig) error {
 		}
 	}
 
+	// Check if BackOff is defined, MaxDeliver is within range.
+	if lbo := len(ncfg.BackOff); lbo > 0 && ncfg.MaxDeliver <= lbo {
+		return NewJSConsumerMaxDeliverBackoffError()
+	}
+
 	return nil
 }
 
@@ -1465,9 +1470,13 @@ func (o *consumer) updateConfig(cfg *ConsumerConfig) error {
 		sampleFreq, _ := strconv.Atoi(s)
 		o.sfreq = int32(sampleFreq)
 	}
+	// Set MaxDeliver if changed
+	if cfg.MaxDeliver != o.cfg.MaxDeliver {
+		o.maxdc = uint64(cfg.MaxDeliver)
+	}
 
 	// Record new config for others that do not need special handling.
-	// Allowed but considered no-op, [Description, MaxDeliver, SampleFrequency, MaxWaiting, HeadersOnly]
+	// Allowed but considered no-op, [Description, SampleFrequency, MaxWaiting, HeadersOnly]
 	o.cfg = *cfg
 
 	return nil
--- a/server/jetstream_cluster_test.go
+++ b/server/jetstream_cluster_test.go
@@ -9240,6 +9240,67 @@ func TestJetStreamClusterConsumerUpdates(t *testing.T) {
 	t.Run("Clustered", func(t *testing.T) { testConsumerUpdate(t, c.randomServer(), 2) })
 }
 
+func TestJetStreamClusterConsumerMaxDeliverUpdate(t *testing.T) {
+	c := createJetStreamClusterExplicit(t, "JSC", 3)
+	defer c.shutdown()
+
+	nc, js := jsClientConnect(t, c.randomServer())
+	defer nc.Close()
+
+	_, err := js.AddStream(&nats.StreamConfig{Name: "TEST", Subjects: []string{"foo"}, Replicas: 3})
+	require_NoError(t, err)
+
+	maxDeliver := 2
+	_, err = js.AddConsumer("TEST", &nats.ConsumerConfig{
+		Durable:       "ard",
+		AckPolicy:     nats.AckExplicitPolicy,
+		FilterSubject: "foo",
+		MaxDeliver:    maxDeliver,
+	})
+	require_NoError(t, err)
+
+	sub, err := js.PullSubscribe("foo", "ard")
+	require_NoError(t, err)
+
+	_, err = js.Publish("foo", []byte("Hello"))
+	require_NoError(t, err)
+
+	for i := 0; i <= maxDeliver; i++ {
+		msgs, err := sub.Fetch(2, nats.MaxWait(10*time.Millisecond))
+		if i < maxDeliver {
+			require_NoError(t, err)
+			require_Len(t, 1, len(msgs))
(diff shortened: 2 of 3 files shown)


server/jetstream_test.go
```diff
@@ -16276,6 +16276,73 @@ func TestJetStreamConsumerAckSamplingSpecifiedUsingUpdateConsumer(t *testing.T)
 	}
 }
 
+func TestJetStreamConsumerMaxDeliveryUpdateConsumer(t *testing.T) {
```
kozlovic, line 16279: I would like to see also a test with cluster mode, to see if your changes handle that properly.

kozlovic (changes requested):

server/consumer.go
```diff
@@ -1418,6 +1418,11 @@ func (acc *Account) checkNewConsumerConfig(cfg, ncfg *ConsumerConfig) error {
 		}
 	}
 
+	// Check if BackOff is defined, MaxDeliver is within range.
+	if lbo := len(ncfg.BackOff); lbo > 0 && ncfg.MaxDeliver <= lbo {
```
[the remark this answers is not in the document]
abegaj: Good point. If all the requests will be replayed, I believe that there can be situations where, updates made before, that are incompatible with this change (e.g. having `ncfg.MaxDeliver <= lbo`), may return an error.
Unfortunately my knowledge on the NATS internals is very limited. Is there a way to verify this occasions? How can I help?

server/jetstream_test.go
abegaj: Cluster mode tests added

server/consumer.go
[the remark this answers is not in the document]
kozlovic: After more testing and review, I think that we would have caught this configuration in earlier releases, so there should not be case where we have an existing misconfiguration.

server/jetstream_cluster_test.go
```diff
@@ -9240,6 +9240,67 @@ func TestJetStreamClusterConsumerUpdates(t *testing.T) {
...
+	})
+	require_NoError(t, err)
+
+	sub, err := js.PullSubscribe("foo", "ard")
+	require_NoError(t, err)
+
+	_, err = js.Publish("foo", []byte("Hello"))
+	require_NoError(t, err)
+
+	for i := 0; i <= maxDeliver; i++ {
```
kozlovic, line 9268: To reduce number of code lines, I would have made this an inner function:
```
checkMaxDeliver := func() {
    t.Helper()
    for i := 0; i <= maxDeliver; i++ {
        msgs, err := sub.Fetch(2, nats.MaxWait(10*time.Millisecond))
		if i < maxDeliver {
			require_NoError(t, err)
			require_Len(t, 1, len(msgs))
			_ = msgs[0].Nak()
		} else {
			require_Error(t, err, nats.ErrTimeout)
		}
	}
}
```
then call this here `checkMaxDeliver()`

server/jetstream_cluster_test.go
```diff
@@ -9240,6 +9240,67 @@ func TestJetStreamClusterConsumerUpdates(t *testing.T) {
...
+			require_NoError(t, err)
+			require_Len(t, 1, len(msgs))
+			_ = msgs[0].Nak()
+		} else {
+			require_Error(t, err, nats.ErrTimeout)
+		}
+	}
+
+	// update maxDeliver
+	maxDeliver = maxDeliver + 1
```
kozlovic, line 9280: `maxDeliver++`

server/jetstream_cluster_test.go
```diff
@@ -9240,6 +9240,67 @@ func TestJetStreamClusterConsumerUpdates(t *testing.T) {
...
+		AckPolicy:     nats.AckExplicitPolicy,
+		FilterSubject: "foo",
+		MaxDeliver:    maxDeliver,
+	})
+	require_NoError(t, err)
+
+	_, err = js.Publish("foo", []byte("Hello"))
+	require_NoError(t, err)
+
+	for i := 0; i <= maxDeliver; i++ {
```
kozlovic, line 9292: Then here invoke `checkMaxDeliver()` again instead of repeating this code.

server/jetstream_test.go
```diff
@@ -16276,6 +16276,70 @@ func TestJetStreamConsumerAckSamplingSpecifiedUsingUpdateConsumer(t *testing.T)
...
+	})
+	require_NoError(t, err)
+
+	sub, err := js.PullSubscribe("foo", "ard")
+	require_NoError(t, err)
+
+	_, err = js.Publish("foo", []byte("Hello"))
+	require_NoError(t, err)
+
+	for i := 0; i <= maxDeliver; i++ {
```
kozlovic, line 16307: Same comment than in the other test (but still would do a inner function as opposed to make a function to be used by those 2 tests, since they are quite specific).

server/jetstream_test.go
```diff
@@ -16276,6 +16276,70 @@ func TestJetStreamConsumerAckSamplingSpecifiedUsingUpdateConsumer(t *testing.T)
...
+			require_NoError(t, err)
+			require_Len(t, 1, len(msgs))
+			_ = msgs[0].Nak()
+		} else {
+			require_Error(t, err, nats.ErrTimeout)
+		}
+	}
+
+	// update maxDeliver
+	maxDeliver = maxDeliver + 1
```
kozlovic, line 16319: `maxDeliver++`

server/jetstream_cluster_test.go
```diff
@@ -9240,6 +9240,67 @@ func TestJetStreamClusterConsumerUpdates(t *testing.T) {
...
+	require_NoError(t, err)
+
+	sub, err := js.PullSubscribe("foo", "ard")
+	require_NoError(t, err)
+
+	_, err = js.Publish("foo", []byte("Hello"))
+	require_NoError(t, err)
+
+	for i := 0; i <= maxDeliver; i++ {
+		msgs, err := sub.Fetch(2, nats.MaxWait(10*time.Millisecond))
```
kozlovic, line 9269: It is good that we have low timeout for the case when we expect the failure, however, 10ms may be a bit too small and cause flappers on Travis maybe? So having 100ms would still be fast but may avoid flappers.

kozlovic (changes requested):
Apologies for the back and forth request for changes. I feel more comfortable with the backoff/maxdeliver check that you have added, I think it is safe. I would just revisit the tests.

kozlovic (approved):
LGTM

PR #3269, opened 2022-07-15
[FIXED] JetStream: stream mirror updates not rejected in standalone

kozlovic:
Updates to stream mirror config are rejected in cluster mode, but
were not in standalone. This PR adds the check in standalone mode.


matthiashanel (approved):
LGTM

PR #3270, opened 2022-07-15
Move and cancel fixes

matthiashanel:
The Move/Cancel/Downscale mechanism did not take into account that since recently, 
the consumer's replica count can be set independently. 

This alters peer selection to have the ability to skip unique tag prefix check for server that will be replaced.
Say you have 3 az, and want to add another server to az:1, in order to replace a server that is the same zone.
Without this change, uniqueTagPrefix check would filter the server to replace with and cause a failure.

The cancel move response could not be received due to the wrong account name.

### DIFF
--- a/server/const.go
+++ b/server/const.go
@@ -41,7 +41,7 @@ var (
 
 const (
 	// VERSION is the current version for the server.
-	VERSION = "2.9.0-beta-mh2.7"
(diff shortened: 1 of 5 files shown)


kozlovic (approved):
LGTM

PR #3271, opened 2022-07-15
Fix and test for RePub src ignored at runtime.

tbeets:

For a stream that is RePublish-enabled, the RePublish source configuration (RePublish message filter) was not being evaluated for each ingested message.  Every message ingested by stream would be (re)published.

This PR fixes and adds unit tests.

/cc @nats-io/core

### DIFF
--- a/server/accounts.go
+++ b/server/accounts.go
@@ -4330,7 +4330,7 @@ func (tr *transform) Match(subject string) (string, error) {
 		return _EMPTY_, ErrBadSubject
 	}
 
-	if isSubsetMatch(tts, tr.src) {
+	if (tr.src == "" || tr.src == ">") || isSubsetMatch(tts, tr.src) {
 		return tr.transform(tts)
 	}
 	return _EMPTY_, ErrNoTransforms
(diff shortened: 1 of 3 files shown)


matthiashanel (changes requested):
LGTM, pending fixing the staticcheck issue as well

derekcollison (approved):
In general LGTM, we should try to optimize that match all scenario which will most likely be the more common pattern.

tbeets:
Enhanced to skip match check if Source allows any subject.  There _may_ be room to squeak more optimization by skipping the transform and simply pre-pending non-wildcard Destination tokens to the head of the message's subject hierarchy but that is net result of the transform anyway...  @derekcollison

server/accounts.go
```diff
@@ -4330,7 +4330,7 @@ func (tr *transform) Match(subject string) (string, error) {
 		return _EMPTY_, ErrBadSubject
 	}
 
-	if isSubsetMatch(tts, tr.src) {
+	if (tr.src == "" || tr.src == ">") || isSubsetMatch(tts, tr.src) {
```
derekcollison, line 4333: for "" we use _EMPTY_ and their is a const for ">" as well.

derekcollison: If destination is empty can just use the subject as well.

derekcollison (approved):
LGTM

PR #3280, opened 2022-07-21
[IMPROVED] JetStream: stream already exists error description

### LINKED ISSUE #3273: Add detail to JSStreamNameExistErr
The text received by the client on this error
```
JSStreamNameExistErr = 10058,                       ///< Stream name already in use
```
is
```
stream name already in use
```
This issue is to request an update to the error text to something that conveys
```
stream name already exists with different settings
```
...to highlight that calling `AddStream` on a stream name already in use *can* succeed, but doesn't when the existing stream has different settings than are specified in the current call. 

This distinction may be obvious to the caller once it is known, however it was difficult to find supporting documentation, and this simple wording change would have made the problem clear.

kozlovic:
The `JSStreamNameExistErr` will now include in the description that
the stream exists with a different configuration, because that is
the error clients would get when trying to add a stream with a
different configuration (otherwise this is a no-op and client
don't get an error).

Since that error was used in case of restore, a new error is added
but uses the same description prefix "stream name already in use"
but adds ", cannot restore" to indicate that this is a restore
failure because the stream already exists.

Resolves #3273


ripienaar (approved):
LGTM

derekcollison (approved):
LGTM

PR #3281, opened 2022-07-21
fix and improve unit test

matthiashanel:
reduce number of messages sent
try move+cancel for every server

 - [ ] Tests added

Resolves #

 -
 -
 -

/cc @nats-io/core

### DIFF
--- a/server/jetstream_cluster_test.go
+++ b/server/jetstream_cluster_test.go
@@ -3718,7 +3718,7 @@ func TestJetStreamClusterMoveCancel(t *testing.T) {
 	require_NoError(t, err)
 	ephName := ci.Name
 
-	toSend := uint64(100_000)
+	toSend := uint64(1_000)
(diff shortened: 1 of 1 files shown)


kozlovic (approved):
LGTM. From 155 seconds down to 7 seconds or less, I'll take that ;-)

derekcollison (commented):
I like it too, but reason it was large was it was not showing any issues with small number of messages. So the large number was to catch those scenarios.

kozlovic:
@derekcollison I mentioned that if the big number was needed, then it should be moved to the "no race" tests, but @matthiashanel mentioned that he was able to get the desired condition with only 1000 messages.

matthiashanel:
> I like it too, but reason it was large was it was not showing any issues with small number of messages. So the large number was to catch those scenarios.

The intent was to make sure move was in progress, but since this is timer based the large number was not really needed

PR #3282, opened 2022-07-21
down scale consumer before downscale of stream

matthiashanel:
Now monitorStream waits with scaling down the stream until all
monitorConsumer have scaled down their respective consumer

Also update consumer assignment for later use in monitorConsumer
Same for stream assignment in monitorStream

 - [ ] Tests added

Resolves #

 -
 -
 -

/cc @nats-io/core

### DIFF
--- a/server/jetstream_cluster.go
+++ b/server/jetstream_cluster.go
@@ -1346,12 +1346,13 @@ func (js *jetStream) truncateOldPeers(mset *stream, newPreferred string) {
 
 	cc, csa := js.cluster, sa.copyGroup()
 	csa.Group.Peers = csa.Group.Peers[len(csa.Group.Peers)-csa.Config.Replicas:]
-	// Now do consumers actually first here, followed by the owning stream.
+	// Now do consumers still needing truncating first here, followed by the owning stream.
 	for _, ca := range csa.consumers {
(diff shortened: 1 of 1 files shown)


derekcollison (approved):
LGTM

server/jetstream_cluster.go
```diff
@@ -1634,6 +1635,34 @@ func (mset *stream) removeNode() {
...
+// as well as the first current peer and if the leader (always current) is part of the set or not
+func currentPeerCount(ci *ClusterInfo, peerSet []string, leaderId string) (currentCount int, firstPeer string, foundLeader bool) {
+	for _, peer := range peerSet {
+		// selfId is leaderId
+		foundCurrent := peer == leaderId
+		if foundCurrent {
+			foundLeader = true
+		} else {
+			for _, p := range ci.Replicas {
+				if peer == string(getHash(p.Name)) {
```
kozlovic, line 1648: Maybe we could think of storing the hash in PeerInfo - lazily, the first time we detect it is not computed, so we don't have to do that all the time. Don't think it is a problem as-is though, and there are many other places we call getHash()...

kozlovic (approved):
LGTM

PR #3283, opened 2022-07-22
[FIXED] Short index write could lead to loss of stream sequence for an empty stream

derekcollison:
When we write a msgBlock idx file that is shorter than the previous write we could fail to recover the idx properly. We do truncate when we do syncs but if the filestore is encrypted and the server is restarted before this operation the decrypt would fail since there are extra bytes. For a stream that is empty but has a non-zero start sequence, we could lose that state on a restart.

/cc @nats-io/core


matthiashanel (approved):
LGTM

kozlovic (approved):
LGTM

PR #3284, opened 2022-07-24
[FIXED] Losing stream sequence on multiple restarts and leader changes.

derekcollison:
When msgs were expired on restart recovery we could lose track on subsequent restarts of our starting sequence with no additional activity. If a leadership change happened or we were in R1 mode we would lose our starting sequence.

This could happen on low traffic streams with TTLs that frequently would empty the stream and when a server was down possibly before that event happened.

Thanks to @tbeets for the details on how to reproduce!

/cc @nats-io/core


tbeets (approved):
LGTM

PR #3287, opened 2022-07-25
[FIXED] JetStream/Cluster: Stream names/infos would return bad response

kozlovic:
If there are more stream names that the current limit of 1024, getting
the list of names would return them all instead of using pagination.

For "stream infos", the Total amount returned would be the API limit
instead of the actual number of streams.

Resolves https://github.com/nats-io/natscli/issues/541


derekcollison (approved):
LGTM - Thanks!

PR #3293, opened 2022-07-27
add implementation for consumer replica change

### LINKED ISSUE #3262: JetStream - Updating consumer from R3 to R1 doesn't work
`nats-server`: v2.8.4
`nats.go`: v1.16.0
`NATS cli`: v0.0.33

#### Step to reproduce:
1. Setup a 3-nodes JetStream cluster.
2. Create a stream with a replica count 3.
3. Create a consumer using `nats.go` client with a replica count 3.
4. Publish and consume some messages.
5. Update the consumer's replica count to 1 using `nats.go` client

#### Expected results
1. Replica count for the consumer should get updated to 1.
2. Consumer should have no replicas or metadata on replica nodes.
3. `nats con info` should not show any replicas field for a R1 consumer

Screenshot 2022-07-14 at 12 55 36 PM

4. When the leader node is killed while consuming, the client should get a "Consumer Migration" message (since, it is R1 now).

#### Actual results
1. `nats con info` does show replica count as 1 (which is correct)
2. Consumer still have its metadata on other nodes and it's being updated regularly (confirmed with `ls -l`)
3. `nats con info` shows replicas field

Screenshot 2022-07-14 at 12 54 44 PM

4. When the leader node is killed, the client gets a "Leadership Change" message (that means, it is still > R1)

matthiashanel:
fixes #3262

also check peer list on every update

 - [ ] Tests added

Resolves #

 -
 -
 -

/cc @nats-io/core

### DIFF
--- a/server/jetstream_cluster.go
+++ b/server/jetstream_cluster.go
@@ -1878,6 +1878,8 @@ func (js *jetStream) monitorStream(mset *stream, sa *streamAssignment, sendSnaps
 		case <-uch:
 			// keep stream assignment current
 			sa = mset.streamAssignment()
+			// keep peer list up to date with config
+			js.checkPeers(mset.raftGroup())
 			// We get this when we have a new stream assignment caused by an update.
 			// We want to know if we are migrating.
 			migrating := mset.isMigrating()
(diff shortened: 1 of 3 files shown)


server/jetstream_cluster.go
```diff
@@ -4764,6 +4767,52 @@ func (s *Server) jsClusteredStreamRequest(ci *ClientInfo, acc *Account, subject,
...
+
+	const timeout = 2 * time.Second
+	notActive := time.NewTimer(timeout)
+	defer notActive.Stop()
+
+	var err error
+	var data interface{}
+
+	select {
+	case <-s.quitCh:
```
derekcollison, line 4801: Would make these just errors or define them as vars once.

derekcollison (approved):
LGTM - One nit on errors but not a blocker.

server/jetstream_cluster.go
matthiashanel: incorporated

derekcollison (approved):
LGTM

PR #3301, opened 2022-07-28
Update tests to check last sequence state

derekcollison:

/cc @nats-io/core


tbeets:
LGTM.  Added LastSeq validation is good.

PR #3302, opened 2022-07-28
[FIXED] Reject update of pull consumer MaxWaiting since it is not supported

bruth:


server/jetstream_test.go
```diff
@@ -9922,6 +9922,64 @@ func TestJetStreamPullConsumerMaxWaitingOfOne(t *testing.T) {
...
+		_, err := sub.Fetch(1, nats.MaxWait(time.Second))
+		errch <- err
+	}()
+
+	var ok bool
+	err = <-errch
+	if strings.Contains(err.Error(), "Exceeded MaxWaiting") {
+		ok = true
+	}
+	err = <-errch
```
kozlovic, line 9973: Since you have MaxWaiting == 1 and have only 2 concurrent requests, only 1 should fail because of Exceeded MaxWaiting, not the 2. So I would recommend that you setup the one that you know is going to fail due to a timeout (so bigger timeout) and the second should fail due to Exceeded MaxWaiting. Let's also try to keep the MaxWait as small as possible (but without falling into flapping territory).

kozlovic (changes requested):

server/jetstream_test.go
bruth: Thanks, good point. Pushed a new commit, let me know if this is sufficient.

bruth: And for context in case @derekcollison didn't mention, this is intended to fail right now to show case the `MaxWaiting` option is not be reflected on consumer update. It still maintains its original `waitQueue` size of 512 after update.

server/jetstream_test.go
```diff
@@ -9953,30 +9953,24 @@ func TestJetStreamPullConsumerMaxWaitingUpdate(t *testing.T) {
...
-
-	var ok bool
-	err = <-errch
-	if strings.Contains(err.Error(), "Exceeded MaxWaiting") {
-		ok = true
-	}
-	err = <-errch
-	if strings.Contains(err.Error(), "Exceeded MaxWaiting") {
-		ok = true
+	// Should be rejected..
```
kozlovic, line 9964: No, we can't rely on the fact that the one in the "go func()" is the first, actually. Do as follow: use same 300ms for both. Check err from inline one: if ErrTimeout, then expect <-errch to be Exceeded. If Exceeded first, expect <-errch to be ErrTimeout. Makes sense?

kozlovic (changes requested):

server/jetstream_test.go
bruth: I am glad you said that :wink: Shouldn't be any determinism here.. hopefully should be good now. I appreciate the feedback.

kozlovic (approved):
LGTM. However, we should not merge to main or at least add a t.Skip() until the server-side code is fixed.

bruth:
Added a skip

kozlovic:
@bruth @derekcollison So now that I understand that this test was to demonstrate that the update is actually not happening, are we planning to allow that field to be updated? Unlike some other fields, this one is a bit more tricky to update. The consumer uses a slice with a capacity of MaxWaiting when created and use that capacity to know if it is full. We would have to resize this slice, but not sure what we do if the new limit is below existing waiting count? Will the server invalidate those?

derekcollison:
I thought this was to update a limit for pull requests for MaxBatch and MaxWait, which I think should be ok to update. 

MaxWaiting I would say no at this time.

kozlovic:
In light of that @bruth, should we switch it back to draft or simply close for now?

bruth:
If changing should be rejected, then I will add the check to `checkNewConsumerConfig` to prevent it from slipping through and update the test to confirm it is behaving as expected.

bruth:
@kozlovic added the check and updated the test.

kozlovic:
@bruth Wonder if we should add a cluster test because I am not sure if this is the same code path. If you check and it is, then no need.

bruth:
Just tested with the cluster test harness `createJetStreamClusterExplicit` and the same test passed. I can commit the additional if you want.

kozlovic:
@bruth Sure, let's add a test. Note that cluster tests need to have the name prefix `TestJetStreamCluster` and be located in `jetstream_cluster_test.go`. (super cluster tests names start with `TestJetStreamSuperCluster` and located in `jetstream_super_cluster_test.go`).

bruth:
Fixed the merge conflict.

kozlovic:
@bruth May want to rebase from main:
```
git checkout main
git pull
git checkout <your branch>
git rebase main
```
then you could squash all your commits.

bruth:
Squashed.

kozlovic (approved):
LGTM. Thanks!

PR #3304, opened 2022-07-29
Fixed TLS test so it passes on macOS

kozlovic:

### DIFF
--- a/server/leafnode_test.go
+++ b/server/leafnode_test.go
@@ -2624,7 +2624,11 @@ func TestLeafNodeTLSConfigReload(t *testing.T) {
 	// Wait for the error
 	select {
 	case err := <-lg.errCh:
-		if !strings.Contains(err, "unknown") && !strings.Contains(err, "broken") {
+		// Since Go 1.18, we had to regenerate certs to not have to use GODEBUG="x509sha1=1"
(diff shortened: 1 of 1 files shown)


derekcollison (approved):
LGTM

server/leafnode_test.go
```diff
@@ -2624,7 +2624,11 @@ func TestLeafNodeTLSConfigReload(t *testing.T) {
 	// Wait for the error
 	select {
 	case err := <-lg.errCh:
-		if !strings.Contains(err, "unknown") && !strings.Contains(err, "broken") {
+		// Since Go 1.18, we had to regenerate certs to not have to use GODEBUG="x509sha1=1"
+		// But on macOS, with our test CA certs, no SCTs included, it will fail
+		// for the reason "x509: “localhost” certificate is not standards compliant"
+		// instead of "unknown authority".
+		if !strings.Contains(err, "unknown") && !strings.Contains(err, "compliant") {
```
wallyqs, line 2631: Looks like this will be fixed in Go 1.20: https://github.com/golang/go/issues/51991

wallyqs (approved):
LGTM

server/leafnode_test.go
kozlovic: Thanks for looking!

PR #3307, opened 2022-07-29
Improvements to raft layer with snapshots on catchup.

derekcollison:
When we were catching up a raft follower that involved snapshots, we would send the snapshot entry followed by an entry that was one less.

Also added in more logging around exceptional conditions to help with triage and debugging production systems.

/cc @nats-io/core


kozlovic (approved):
LGTM

PR #3308, opened 2022-07-29
Encrypt meta and raft states.

derekcollison:

/cc @nats-io/core


kozlovic (approved):
LGTM. However, this makes me nervous. You have tested manual upgrade, but once files are encrypted, there is no way to downgrade server version (not that we want users to do that), correct?

derekcollison:
The code to allow upgrade to encrypted has existed with original work. Downgrades have never been possible directly in the server by removing the key from the config. Upgrades do work with just adding key to config and restarting. You can backup and restore an encrypted asset if downgrading is needed today.

PR #3314, opened 2022-07-31
[FIXED] Instability with encrypted systems

derekcollison:
Fixed and improved several system instability issues with encrypted state that were a result of block counters being incorrect for message blocks either due to compaction or when messages expired and we would keep markers for sequence state.

/cc @nats-io/core


tbeets (approved):
LGTM

kozlovic (approved):
LGTM

PR #3315, opened 2022-07-31
[FIXED] MQTT: Possible panic when clients misbehave

### LINKED ISSUE #3313: mqtt nil pointer (without reproducer)
Hi
At some time we get an error
изображение
We do not have a reproducer

Have any ideas why `mqttAccountSessionManager` have is nil?
https://github.com/nats-io/nats-server/blob/66524ed715a72db192ec93465bf24391f7696f3f/server/mqtt.go#L1853

https://github.com/nats-io/nats-server/blob/66524ed715a72db192ec93465bf24391f7696f3f/server/mqtt.go#L3522-L3527

But `mqttAccountSessionManager` is sayed `quick reference to account session manager, immutable after processConnect()`
https://github.com/nats-io/nats-server/blob/66524ed715a72db192ec93465bf24391f7696f3f/server/mqtt.go#L313

In another goroutine trying to work with a client who is not yet ready?
https://github.com/nats-io/nats-server/blob/66524ed715a72db192ec93465bf24391f7696f3f/server/mqtt.go#L2815

Please, any ideas. Thx

PS nats-server v2.8.4

kozlovic:
If a client with a given client ID is connected and while connected
another client tries to reuse the same client ID, the spec says that
the old client be closed and the new one accepted.
However, the server protects from this flapping happening all the time
by rejecting new clients that try to connect at a very fast pace.

However, the server was closing a misbehaving client after a second
delay (to prevent immediate reconnect if the client library does that)
but was not blocking the read loop and the compounding issue was that
if that misbehaving client is sending more protocols (for instance SUB) 
the server would panic because the client was not fully configured.

To prevent that, the server will now "block" this misbehaving client
in its readLoop before closing the connection, preventing processing
of possible protocols that follow the CONNECT.

EDIT: It is allowed for a client to send other protocols before receiving
the CONNACK from the server, so I edited the text above stating that
the client would be really misbehaving if doing that.

Resolves #3313

### DIFF
--- a/server/mqtt.go
+++ b/server/mqtt.go
@@ -2708,11 +2708,17 @@ CHECK:
 			// Will hold this client for a second and then close it. We
 			// do this so that if the client has a reconnect feature we
 			// don't end-up with very rapid flapping between apps.
-			time.AfterFunc(mqttSessJailDur, func() {
-				c.closeConnection(DuplicateClientID)
-			})
+			// We need to wait in place and not schedule the connection
+			// close because if this is a misbehaved client that does
+			// not wait for the CONNACK and sends other protocols, the
+			// server would not have a fully setup client and may panic.
 			asm.mu.Unlock()
-			return nil
+			select {
+			case <-s.quitCh:
+			case <-time.After(mqttSessJailDur):
+			}
+			c.closeConnection(DuplicateClientID)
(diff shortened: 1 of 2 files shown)


derekcollison (approved):
LGTM

PR #3318, opened 2022-08-02
Chaos tests for JS cluster

mprimi:
 - [x] Tests added
 - [x] Changes squashed to a single commit ** => intentionally split in 2 commits for easier review**

- Add chaos controller
- Add "cluster bouncer" chaos monkey
- Add 2 tests that use it
- Add some utility methods

/cc @nats-io/core

### DIFF
--- a/scripts/runTestsOnTravis.sh
+++ b/scripts/runTestsOnTravis.sh
@@ -35,7 +35,7 @@ elif [ "$1" = "js_tests" ]; then
     # tests by using the `skip_js_cluster_tests` and `skip_js_super_cluster_tests`
     # build tags.
 
-    go test -race -v -run=TestJetStream ./server -tags=skip_js_cluster_tests,skip_js_super_cluster_tests -count=1 -vet=off -timeout=30m -failfast
+    go test -race -v -run=TestJetStream ./server -tags=skip_js_cluster_tests,skip_js_super_cluster_tests,skip_js_cluster_chaos_tests -count=1 -vet=off -timeout=30m -failfast
 
 elif [ "$1" = "js_cluster_tests" ]; then
 
@@ -50,6 +50,13 @@ elif [ "$1" = "js_super_cluster_tests" ]; then
     # tests with `TestJetStreamSuperCluster`.
 
     go test -race -v -run=TestJetStreamSuperCluster ./server -count=1 -vet=off -timeout=30m -failfast
+
+elif [ "$1" = "js_chaos_tests" ]; then
+
+    # Run JetStream chaos tests. By convention, all JS cluster chaos
+    # tests with `TestJetStreamChaos`.
+
+    go test -race -v -p=1 -run=TestJetStreamChaos ./server -count=1 -vet=off -timeout=30m -failfast
 
 elif [ "$1" = "srv_pkg_non_js_tests" ]; then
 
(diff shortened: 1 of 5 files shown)


scripts/runTestsOnTravis.sh
```diff
@@ -51,6 +51,13 @@ elif [ "$1" = "js_super_cluster_tests" ]; then
 
     go test -race -v -run=TestJetStreamSuperCluster ./server -count=1 -vet=off -timeout=30m -failfast
 
+elif [ "$1" = "js_cluster_chaos_tests" ]; then
+
+    # Run JetStream clustered chaos tests. By convention, all JS cluster chaos
+    # tests with `TestJetStreamClusterChaos`.
+
+    go test -race -v -run=TestJetStreamClusterChaos ./server -count=1 -vet=off -timeout=30m -failfast
```
kozlovic, line 59: Either name this `TestJetStreamChaosCluster` or then you need to add `skip_js_cluster_chaos_tests` to the "js_cluster_tests" section above...

server/jetstream_chaos_helpers_test.go
```diff
@@ -0,0 +1,148 @@
+// Copyright 2020-2022 The NATS Authors
```
kozlovic, line 1: Since this is a new file, `2022` is enough

server/jetstream_chaos_helpers_test.go
```diff
@@ -0,0 +1,148 @@
...
+import (
+	"math/rand"
+	"sync"
+	"testing"
+	"time"
+)
+
+// Support functions
+
+func init() {
```
kozlovic, line 25: Would remove that since it is empty?

server/jetstream_cluster_chaos_test.go
```diff
@@ -0,0 +1,79 @@
+// Copyright 2020-2022 The NATS Authors
```
kozlovic, line 1: Same

kozlovic (changes requested):
Check for my comments and notice some issues on Travis regarding fmt or unused imports. Also, may be worth (at least to experiment), add an entry in ".travis.yml" to actually trigger the tests, just to make sure that all is setup properly for that.

mprimi (commented):
Thank you for the review @kozlovic 

All comments addressed.

kozlovic (changes requested):
Still have some issues, such as:
```
# github.com/nats-io/nats-server/v2/server [github.com/nats-io/nats-server/v2/server.test]
332server/jetstream_chaos_helpers_test.go:17:2: imported and not used: "fmt"
333server/jetstream_chaos_helpers_test.go:18:2: imported and not used: "log"
334server/jetstream_chaos_helpers_test.go:20:2: imported and not used: "strings"
```

kozlovic (changes requested):
What about adding this to .travis.yml, at least to make sure that they run correctly (even if we were to remove/comment out once it is time to merge if we want to not run them all the time):
```
    - name: "Run JetStream chaos tests"
      env: TEST_SUITE=js_chaos_tests
```
You could put that say below the line of the cluster tests `env: TEST_SUITE=js_cluster_tests`

mprimi:
@kozlovic 
> What about adding this to .travis.yml, 

I wasn't sure about adding this to the PR build critical path. Some of the tests are failing, and I need to set a large timeout once I add them.

Anyway, added now, we can always remove it later.

kozlovic:
@mprimi I just wanted your to add a new commit with that to make sure that it would trigger and run at least to make sure everything was ok. Yeah, seeing that it is 14 minutes long (just for 2 tests), maybe we don't include them by default in Travis, but devs will have options to run them locally (with naming convention). So please remove from travis.yml now that we know everything is setup correctly.

mprimi:
@kozlovic, gotch. Removed from travis

kozlovic (approved):
LGTM

Issue #3306, opened 2022-07-29
Stream Lost Quorum triggered for streams in deleted account as well


sourabhaggrawal:
## Defect
If account is deleted without deleting the streams then on restarting the peer nodes other node starts complaining about 'No Quorum' for streams which belongs to deleted account and never recovers.

 - [ ] Included `nats-server -DV` output
 - [x] Included a [Minimal, Complete, and Verifiable example] (https://stackoverflow.com/help/mcve)

#### Versions of `nats-server` and affected client libraries used:
2.8.4
#### OS/Container environment:
ubuntu-20.04
#### Steps or code to reproduce the issue:
In a 3 node cluster
1. Create one jetstream enabled account
2. Create one stream under the new account and publish couple of messages to it.
3. Deleted the account without deleting streams
4. Restart any node.

#### Expected result:
It should restart smoothly and should not log any error related to deleted account streams or account itself.
#### Actual result:
Other node starts logging "No quorum, stalled" for new stream we just created. The logs does not stop until we restart the node which is complaining. 

All accounts mentioned in the logs are deleted account.

`[24314] 2022/07/29 10:01:37.935503 [WRN] Account fetch failed: open /var/lib/nats/superadmin/jwt/ACGREIUVT2PT4ERQAFNQFY6SJHMCWH77E3LQKCCCUZF2F7B47ZXBKTH6.jwt: no such file or directory
[24314] 2022/07/29 10:01:38.903683 [WRN] Account fetch failed: open /var/lib/nats/superadmin/jwt/ACPU7YPBAVITL5RAZNURPROSASX4N26PZYVOR6ODSQMNFR6MBUI2PPXJ.jwt: no such file or directory
[24314] 2022/07/29 10:01:38.907527 [WRN] Account fetch failed: open /var/lib/nats/superadmin/jwt/AC6P4P3CINXZ4MFTQSWXGCJI53BHGZLYMNQFU4SHY2ANOBY2JUXZFRIO.jwt: no such file or directory
[24314] 2022/07/29 10:01:40.818752 [WRN] JetStream cluster stream 'ACN5OEYPSG7VDT2HWUUHVEZIKPDPMKKO4VQQBKIIC2BPKUEYKY6TEXQ5 > 5e83a915-c06a-4dd4-9b2a-07e4da85b7ad_stream' has NO quorum, stalled
[...]

Issue #3186, opened 2022-06-13
nats-server tries to recover data for deleted JetStream account


goku321:
`nat-server version`: 2.8.4

Steps to reproduce:
  1. Create an account with JetStream enabled
  2. Create a stream and publish some data
  3. Delete the account by publishing to `$SYS.REQ.CLAIMS.DELETE`
  4. Restart the nats-server
  
Even after the account is deleted nats-server tries to recover the JetStream data and then times out as it is not able to fetch the account. This can lead to multiple failed account lookups if the number of deleted accounts (with JetStream enabled) is high.

I've tried deleting all the streams and consumers by publishing to `$JS.API.STREAM.DELETE.*` and `$JS.API.CONSUMER.DELETE.*.*` respectively but that doesn't delete the entire store directory for an account. And, as long as the directory is present, nats-server will try to recover the JetStream data.

**Proposal**:
1. There should be a way to clean up the entire JetStream data directory for an account.
2. (OR) Account JWT deletion should also take care of deleting the JetStream data.
3. (OR) nats-server shouldn't try to recover JetStream data for a deleted account.

derekcollison:
For now we suggest you reinstate the account, delete the streams, and the disable JetStream from the account, then re-remove.

We will improve this experience for sure but that should get you unstuck.

matthiashanel:
@goku321 we will be adding your proposal 1, triggered by a system account message

goku321:
Thanks @matthiashanel that would be great! For now, I'll try to include Derek's suggestion in our workflow

goku321:
> For now we suggest you reinstate the account, delete the streams, and the disable JetStream from the account, then re-remove.
> 
> We will improve this experience for sure but that should get you unstuck.

I tried this but no luck. I can still see the account fetch message in the logs. The only solution that works for me is to delete the account directory under the JetStream store directory. Maybe, I'm missing something?

derekcollison:
After you re-instated the account you could see the fetch account being successful?

goku321:
My bad. I actually deleted the account JWT along with deleting the streams and disabling JetStream. I thought I had to remove the account as well. I'll list the steps for clarity:
1. Delete the streams
2. Disable JetStream
3. Restart nats-server

Account fetch is successful and the server doesn't try to recover JetStream data.

derekcollison:
Thanks for the feedback.

caleblloyd:
[...]

PR #3319, opened 2022-08-02
Adding account purge operation

matthiashanel:
The new request is available for the system account.
The subject to send the request to is `$JS.API.ACCOUNT.PURGE.*`
With the name of the account to purge instead of the wildcard.

Also added directory cleanup code such that server do not
end up with empty streams directories and account dirs that
only contain streams

Also adding ACCOUNT to leaf node domain rewrite table

Addresses #3186 and #3306 by providing a way to
get rid of the streams for existing and non existing accounts

### DIFF
--- a/server/jetstream.go
+++ b/server/jetstream.go
@@ -94,20 +94,22 @@ type JetStreamAPIStats struct {
 // This is for internal accounting for JetStream for this server.
 type jetStream struct {
 	// These are here first because of atomics on 32bit systems.
-	apiInflight    int64
-	apiTotal       int64
-	apiErrors      int64
-	memReserved    int64
-	storeReserved  int64
-	memUsed        int64
-	storeUsed      int64
-	clustered      int32
-	mu             sync.RWMutex
-	srv            *Server
-	config         JetStreamConfig
-	cluster        *jetStreamCluster
-	accounts       map[string]*jsAccount
-	apiSubs        *Sublist
+	apiInflight   int64
+	apiTotal      int64
+	apiErrors     int64
+	memReserved   int64
+	storeReserved int64
+	memUsed       int64
+	storeUsed     int64
+	clustered     int32
+	mu            sync.RWMutex
+	srv           *Server
+	config        JetStreamConfig
+	cluster       *jetStreamCluster
+	accounts      map[string]*jsAccount
+	apiSubs       *Sublist
+	// System level request to purge a stream move
+	accountPurge   *subscription
 	metaRecovering bool
 	standAlone     bool
 	disabled       bool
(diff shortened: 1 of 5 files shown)


server/jetstream.go
```diff
@@ -879,8 +902,13 @@ func (s *Server) shutdownJetStream() {
 			accounts = append(accounts, a)
 		}
 	}
+	accPurgeSub := js.accountPurge
 	js.mu.RUnlock()
 
+	if accPurgeSub != nil {
+		s.sysUnsubscribe(accPurgeSub)
```
derekcollison, line 909: Should we set js.accountPurge to nil?

server/jetstream_cluster.go
```diff
@@ -3074,6 +3094,24 @@ func (js *jetStream) processClusterDeleteStream(sa *streamAssignment, isMember,
 		sa.Group.node.Delete()
 	}
 
+	// Cleanup in case account does not exist or node was nil
+	if sacc := s.SystemAccount(); sacc != nil {
```
derekcollison, line 3098: In the happy path this is always handled, this looks a bit strange here.
If the sa.Group.node.Delete() was called from above this is already handled etc.

server/jetstream_cluster.go
```diff
@@ -3499,12 +3542,22 @@ func (js *jetStream) processClusterDeleteConsumer(ca *consumerAssignment, isMemb
 		ca.Group.node.Delete()
 	}
 
+	// Cleanup in case account does not exist or node was nil
+	if sacc := s.SystemAccount(); sacc != nil {
```
derekcollison, line 3546: Again this looks out of place here, maybe we should have this to be pedantic to an extreme, but..

derekcollison (commented):
I see we added in a purge subscription to capture these commands. Would we not want this to be a replicated meta state such that servers offline, once restarted would also perform the purge?

derekcollison (commented):
Ignore comment about subscription vs meta state, got it now..

server/jetstream_cluster.go
```diff
@@ -3094,8 +3094,10 @@ func (js *jetStream) processClusterDeleteStream(sa *streamAssignment, isMember,
 		sa.Group.node.Delete()
 	}
 
-	// Cleanup in case account does not exist or node was nil
-	if sacc := s.SystemAccount(); sacc != nil {
+	// This is a stop gap cleanup in case
+	// 1) the account does not exist (and mset couldn't be stopped) and/or
+	// 2) node was nil (and couldn't be deleted)
+	if sacc := s.SystemAccount(); sacc != nil && (!stopped || sa.Group.node == nil) {
```
derekcollison, line 3100: Would reverse so in happy path do not need server lock to lookup system account. Also deref on sa.Group.node should be under js lock, so maybe capture above.

server/jetstream_cluster.go
```diff
@@ -3542,8 +3546,10 @@ func (js *jetStream) processClusterDeleteConsumer(ca *consumerAssignment, isMemb
 		ca.Group.node.Delete()
 	}
 
-	// Cleanup in case account does not exist or node was nil
-	if sacc := s.SystemAccount(); sacc != nil {
+	// This is a stop gap cleanup in case
+	// 1) the account does not exist (and mset consumer couldn't be stopped) and/or
+	// 2) node was nil (and couldn't be deleted)
+	if sacc := s.SystemAccount(); sacc != nil && (!stopped || ca.Group.node == nil) {
```
derekcollison, line 3552: Same here.

server/jetstream_cluster.go
matthiashanel: fixed (also modified earlier sa.Group.node access)

derekcollison (approved):
LGTM

PR #3327, opened 2022-08-03
Auto-set and upgrade AllowDirect when MaxMsgsPerSubject is set.

derekcollison:
Also allow mirrors to inherit properly.

This is the preferred path moving forward for KV setups. This provides the smoothest experience during upgrade to 2.9 without external tooling and maintenance needed from customers/users.

/cc @nats-io/core


ripienaar:
Be good to also upgrade other past things we have in `nats kv upgrade` so that can die.

```
AllowRollup = true
Discard = DiscardNew
```

then I can remove that silly command.

derekcollison:
This change does auto-upgrade assets that are recovered on disk during startup.

derekcollison:
> Be good to also upgrade other past things we have in `nats kv upgrade` so that can die.
> 
> ```
> AllowRollup = true
> Discard = DiscardNew
> ```
> 
> then I can remove that silly command.

It does that correctly now. But not the Discard upgrade.. But we do auto promote AllowDirect now.

kozlovic (approved):
I was thinking for a whole now, that it would have been better to have a new boolean in the stream config that indicates that this is a "KV" stream. Assuming that this is going to be used as KV based on some params (like MaxMsgsPer) is in my opinion removing the flexibility and choice for the users to decide what is best for them.
In other words, if someone wants to create a stream with MaxMsgsPer but not want direct gets, they can't. If there was a KV boolean in the config, then server could have done all those settings freely.

derekcollison:
Yes my intent is to not make that settable.

However, old msg get semantics will continue to work.

ripienaar:
Am with Ivan. For example would rather have the reject new work differently rather than what it does now only for KV - it makes other things impossible.

derekcollison:
The AllowDirect was just for transitioning. My intent is that it is not settable, but if you use the old method directly that sill works fine.

ripienaar:
There is a whole set of funky things - not direct related - to KV streams that would be better done just to KV streams. 

The example I had recently is how reject new reject on number of keys not messages per subject as would be more intuitive. It makes sense for KV but for other things. So the KV flag could flip just those behaviours without shutting the doors on others

kozlovic:
Also, I was thinking that: what if we realize that for KV a different type of file storage would be better (like a kv type store, bbolt - not that one, but you get the idea), then it would definitively make sense to know for sure that the intent of the stream is to be KV, not simply to have some of the fields that looks like it could be used for KV...

derekcollison:
These are all great points and agree going forward, but not at this time, imo.

PR #3329, opened 2022-08-04
[FIXED] DirectGet working across leafnodes with JetStream Domains.

derekcollison:
DirectGet needed to be added to the import map to allow cross domain traffic.
However, DirectGets were part of a queue group, and that exposed a bug that would prevent the origin ingest server from a leafnode from processing it for local queue subs since the interest was for the mapping which had no queue context.
 
/cc @nats-io/core


matthiashanel (approved):
LGTM

PR #3333, opened 2022-08-04
Use specific boolean for a leaf test instead of using leafNodeEnabled

kozlovic:
A test TestJetStreamClusterLeafNodeSPOFMigrateLeaders was added at
some point that needed the remotes to stop (re)connecting. It made
use of existing leafNodeEnabled that was used for GW/Leaf interest
propagation races to disable the reconnect, but that may not be
the best approach since it could affect users embedding servers
and adding leafnodes "dynamically".

So this PR introduced a specific boolean specific for that test.

### DIFF
--- a/server/jetstream_helpers_test.go
+++ b/server/jetstream_helpers_test.go
@@ -951,7 +951,7 @@ func (s *Server) closeAndDisableLeafnodes() {
 		leafs = append(leafs, ln)
 	}
 	// Disable leafnodes for now.
-	s.leafNodeEnabled = false
+	s.leafDisableConnect = true
 	s.mu.Unlock()
 
 	for _, ln := range leafs {
--- a/server/leafnode.go
+++ b/server/leafnode.go
@@ -495,12 +495,13 @@ func (s *Server) connectToRemoteLeafNode(remote *leafNodeCfg, firstConnect bool)
 			if url != rURL.Host {
 				ipStr = fmt.Sprintf(" (%s)", url)
 			}
-			if s.isLeafNodeEnabled() {
-				s.Debugf("Trying to connect as leafnode to remote server on %q%s", rURL.Host, ipStr)
-				conn, err = natsDialTimeout("tcp", url, dialTimeout)
-			} else {
+			// Some test may want to disable remotes from connecting
+			if s.isLeafConnectDisabled() {
 				s.Debugf("Will not attempt to connect to remote server on %q%s, leafnodes currently disabled", rURL.Host, ipStr)
 				err = ErrLeafNodeDisabled
+			} else {
+				s.Debugf("Trying to connect as leafnode to remote server on %q%s", rURL.Host, ipStr)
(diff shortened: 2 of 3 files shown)


derekcollison (approved):
LGTM

ripienaar:
I think I can do some end to end testing now, let me see

ripienaar:
So this works for me as in it avoids the core problem of leafnode startup, but I wonder if we should also consider this:

https://github.com/nats-io/nats-server/blob/5df91797d6d7b2c88424177bd95c267f4859bf2c/server/consumer.go#L2608

and 

https://github.com/nats-io/nats-server/blob/5df91797d6d7b2c88424177bd95c267f4859bf2c/server/consumer.go#L2926

These 2 places are still essentially gated around the initial options only and doesnt consider current leafnode reality if someone adds them after NewServer

kozlovic:
@ripienaar But keep in mind that that code is to "override" the notion of interest in case there is a race, that is, it does not mean that if the boolean is disabled that leaf and consumer won't work. Also leafnodes are currently not configuration-reloadable, so if you start a server without accepting leafnode port or remote leafnodes, you can't change that with a config reload.
I am not sure - even programmatically - how you would be able to suddenly have a server be able to accept leaf or connect remotes if they were not configured prior to start.

ripienaar:
@kozlovic basiccally like this:

```
opts := &Options{}
// set initial options
s, _ := NewServer(opts)

// now I want to add accounts programatically, this is done using like
s.LookupOrRegisterAccount("foo")

// now I want to add a leafnode for 'foo', I have to do this here because account has to exist
opts.LeafNode.Remotes=append(.....)

// now leaf nodes boolean is false
s.Start()
```

So I think @derekcollison suggestion to also check the len(remotes) and port in Start() along with your fix would close the loop entirely

kozlovic:
@ripienaar Oh wow.. did not realize that server was holding direct reference to the user provided options, which you may find useful in this context, but to me is prone to bad/race behavior (if you change values of server options that server then uses internally without any ability to protect with lock). But ok, I will make the change if that helps close the loop.

ripienaar:
Indeed its messy, I definitely do not change these post Start() thats would be really bad as you say, Start() might want to copy opts but probably another discussion.

We'd need a bunch of APIs to modify options as today VAST amounts of things can only be done this way when embedded

kozlovic:
>  I definitely do not change these post Start()

Good. And also note that past a Reload(), then the original user provided options will have no effect in the server since a new object is created from the config reload and then reassigned to s.opts, breaking the bound with the user-provided one.

kozlovic:
@ripienaar Ok, I made the change. I kept in NewServer() and updated also in Start() (maybe here would be the only place needed, but not sure, so not taking any risk close to v2.9.0).

ripienaar (approved):
LGTM, post 2.9 we can work together on a better way to manage this, agree its grim

PR #3335, opened 2022-08-04
[ADDED] LeafNode: Support for a SignatureHandler in remote config

### LINKED ISSUE #3331: Add nkey support to RemoteLeafOpts
## Feature Request

Add ability configure an nkey programmatically in the leafnode options.

#### Use Case:

We use a NATS server embedded in our client application. It operates as a leaf node, connecting to NGS. The nkey is _not_ stored on disk, and comes from the Windows credential store and is decrypted on demand. When using the nats.go client, we use the `Nkey` option (see https://github.com/nats-io/nats.go/blob/v1.16.0/nats.go#L1066-L1075). As we migrate to an embedded leafnode, we need the same ability to configure the user.

#### Proposed Change:

Add an option similar to the nats.go package with `type SignatureHandler func([]byte) ([]byte, error)`.

me :partying_face: 

#### Alternative Approaches

Write the credentials out to a temporary file and hope it gets deleted on panic. And also write them out on every server reconnect...

kozlovic:
This would allow in embedded use-cases where the user does not
have the ability to use a credentials file. Instead, a signature
callback is specified and invoked by the server sends the CONNECT
protocol. The user is responsible to provide the JWT and sign the
nonce.

Resolves #3331

### DIFF
--- a/server/leafnode.go
+++ b/server/leafnode.go
@@ -707,8 +707,24 @@ func (c *client) sendLeafConnect(clusterName string, tlsRequired, headers bool) 
 		DenyPub:   c.leaf.remote.DenyImports,
 	}
 
-	// Check for credentials first, that will take precedence..
-	if creds := c.leaf.remote.Credentials; creds != _EMPTY_ {
+	// If a signature callback is specified, this takes precedence over anything else.
+	if cb := c.leaf.remote.SignatureCB; cb != nil {
+		nonce := c.nonce
+		c.mu.Unlock()
+		jwt, sigraw, err := cb(nonce)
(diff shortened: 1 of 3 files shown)


derekcollison (approved):
LGTM

kozlovic:
@ripienaar Do you want to comment on this PR? Do you see value? The user that asked for the ER has tested it and works well for their use case. Otherwise I can merge, let me know.

ripienaar:
Yeah, I would probably one day use this as my auth plugin requires custom signers for nonce and stuff, so +1 from me.

PR #3338, opened 2022-08-05
[FIXED] Memory leak when unsubscribing the last queue subscription

kozlovic:
A server maintains a map for the subject+queue to know the number
of members on the same group. However, on unsubscribe when we get
to the last one being unsubscribed, we were removing from the map
but then unfortunately adding back with a value of 0, which caused
a leak. If the same subscription was coming back, then this map
entry would be reused, but if it is a never coming back queue sub,
then memory could increase continously.


derekcollison (approved):
LGTM!

PR #3340, opened 2022-08-05
Fix chaos tests build tags for Travis

mprimi:
 - [ ] Tests added

Resolves #

[[FIXED] Wrong flag in Travis to exclude chaos tests](https://github.com/nats-io/nats-server/commit/60b0180102c78a83afb3f5d10f466fccdfb730d5)  …
The `js_tests` build target was using the wrong tag to exclude chaos 
tests.
As a result, chaos tests would run as part of the default testing.

[Exclude chaos tests helpers from default build](https://github.com/nats-io/nats-server/commit/60ef3e228283673e85610bda5a223a2c70e91437)

/cc @nats-io/core

### DIFF
--- a/scripts/runTestsOnTravis.sh
+++ b/scripts/runTestsOnTravis.sh
@@ -13,7 +13,7 @@ if [ "$1" = "compile" ]; then
     $(exit $(go fmt $GO_LIST | wc -l));
     go vet $GO_LIST;
     find . -type f -name "*.go" | xargs misspell -error -locale US;
-    staticcheck $GO_LIST
+    staticcheck -tags=js_chaos_tests $GO_LIST
     if [ "$TRAVIS_TAG" != "" ]; then
         go test -race -v -run=TestVersionMatchesTag ./server -count=1 -vet=off
     fi
@@ -35,7 +35,7 @@ elif [ "$1" = "js_tests" ]; then
     # tests by using the `skip_js_cluster_tests` and `skip_js_super_cluster_tests`
     # build tags.
 
-    go test -race -v -run=TestJetStream ./server -tags=skip_js_cluster_tests,skip_js_super_cluster_tests,skip_js_chaos_tests -count=1 -vet=off -timeout=30m -failfast
+    go test -race -v -run=TestJetStream ./server -tags=skip_js_cluster_tests,skip_js_super_cluster_tests -count=1 -vet=off -timeout=30m -failfast
 
 elif [ "$1" = "js_cluster_tests" ]; then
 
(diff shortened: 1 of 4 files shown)


scripts/runTestsOnTravis.sh
```diff
@@ -35,7 +35,7 @@ elif [ "$1" = "js_tests" ]; then
     # tests by using the `skip_js_cluster_tests` and `skip_js_super_cluster_tests`
     # build tags.
 
-    go test -race -v -run=TestJetStream ./server -tags=skip_js_cluster_tests,skip_js_super_cluster_tests,skip_js_cluster_chaos_tests -count=1 -vet=off -timeout=30m -failfast
+    go test -race -v -run=TestJetStream ./server -tags=skip_js_cluster_tests,skip_js_super_cluster_tests,skip_js_chaos_tests -count=1 -vet=off -timeout=30m -failfast
```
kozlovic, line 38: I really would want to use the reverse, that is, not have flags here but instead have a "positive" tag when you want to run the chaos tests (in a section down below).

server/jetstream_chaos_helpers_test.go
```diff
@@ -11,6 +11,9 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+//go:build !skip_js_tests && !skip_js_cluster_tests && !skip_js_chaos_tests
```
kozlovic, line 14: Would replace all that with `js_chaos_tests` and have the same tag in the non helper file. Then, you would need to probably add to the staticcheck command that tag so that it can check that too: `staticcheck -tags js_chaos_tests $GO_LIST`. However, if we don't include that one, then it would complain that some of the helpers you added in other files are not used (since chaos tests would not be compiled), such as `func (*cluster).waitOnClusterHealthz is unused (U1000)`, etc...

kozlovic (changes requested):
We can discuss if you prefer.

scripts/runTestsOnTravis.sh
mprimi: I'm working on the reversed tag. 
I just wanted to get this fix out in case someone is unable to pass PR build because chaos tests are failing

kozlovic: They were not running for some reason... I don't see them in the run on main for instance..

mprimi: Probably the other build tags were excluding them

```
//go:build !skip_js_tests && !skip_js_cluster_tests && !skip_js_chaos_tests
```

kozlovic: I think this is because you already had `!skip_js_cluster_tests` in the chaos tests?

mprimi:
@kozlovic Added one more commit that flips to default exclude

kozlovic:
@mprimi Would you mind doing a rebase from main? I have merged some PRs that move us to Go 1.19 but also removed references to io/ioutil.

mprimi:
@kozlovic just rebased, and checks are green

kozlovic (approved):
LGTM

PR #3341, opened 2022-08-05
Move to Go 1.19, remote io/util, fix data race and a flapper

kozlovic:
Package io/util is deprecated since Go 1.16. Also fixed a race that showed up when I first tried Go 1.19 on my machine and finally tweaked a test that I saw flap several times in the past.


derekcollison (approved):
LGTM

PR #3342, opened 2022-08-05
better error when peer selection fails

matthiashanel:
It is pretty hard to diagnose what went wrong when not enough peers for
an operation where found. This change now returns counts of reasons why
peers where discarded.

**This is a first shot at this and we don't want to reveal too much information.
So please let me know if you think I'm showing too much and/or should
change the format. The intent is that whoever sees this error, has enough 
information to why it failed, without having to rely on logging.**

Changed the error to `JSClusterNoPeers` as it seems more appropriate
of an error for that operation. Not having enough resources is one of
the conditions for a peer not being considered. But so is having a non
matching tag. Which is why `JSClusterNoPeers` seems more appropriate
In addition, `JSClusterNoPeers` was already used as error after one call
to `selectPeerGroup` already.

example:
```
no suitable peers for placement: peer selection cluster 'C' with 3 peers
offline: 0
excludeTag: 1
noTagMatch: 2
noSpace: 0
uniqueTag: 0
misc: 0
```

Examle for mqtt:
```
mid:12 - "mqtt" - unable to connect: create sessions stream for account "$G":
no suitable peers for placement: peer selection cluster 'MQTT' with 3 peers
        offline: 0
        excludeTag: 0
        noTagMatch: 0
        noSpace: 0
        uniqueTag: 0
        misc: 0
         (10005)
```

### DIFF
--- a/server/errors.json
+++ b/server/errors.json
@@ -80,10 +80,10 @@
     "deprecates": ""
   },
   {
-    "constant": "JSClusterNoPeersErr",
+    "constant": "JSClusterNoPeersErrF",
     "code": 400,
     "error_code": 10005,
-    "description": "no suitable peers for placement",
+    "description": "no suitable peers for placement: {err}",
     "comment": "",
     "help": "",
     "url": "",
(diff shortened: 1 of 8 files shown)


derekcollison (approved):
LGTM - Maybe say NoStorage vs NoSpace..

derekcollison (approved):
LGTM - Make sure to rebase once main settles with pop to Go 1.19

kozlovic (approved):
LGTM

PR #3347, opened 2022-08-08
[IMPROVED] Replicas ordering and info regarding unknown in stream info

kozlovic:
If a cluster is brought down and then partially restarted, the
replica information about the non restarted node would be completely
missing. The CLI could report replicas 3 but then only the leader
and the running replicas, but nothing about the other node.
Since this node's server name is not know, this PR adds an entry
with something similar to this:
```
<unknown (peerID: jZ6RvVRH)>, outdated, OFFLINE, not seen
```

Also, replicas array is now ordered, which will help when using
a watcher or repeating stream info commands in that the replicas
output will be stable in regards to the list of replicas.

### DIFF
--- a/server/jetstream_cluster.go
+++ b/server/jetstream_cluster.go
@@ -6852,21 +6852,34 @@ func (js *jetStream) clusterInfo(rg *raftGroup) *ClusterInfo {
 			if current && lastSeen > lostQuorumInterval {
 				current = false
 			}
+			// Create a peer info with common settings if the peer has not been seen
+			// yet (which can happen after the whole cluster is stopped and only some
+			// of the nodes are restarted).
+			pi := &PeerInfo{
+				Current: current,
+				Offline: true,
+				Active:  lastSeen,
+				Lag:     rp.Lag,
+				peer:    rp.ID,
+			}
+			// If node is found, complete/update the settings.
 			if sir, ok := s.nodeToInfo.Load(rp.ID); ok && sir != nil {
 				si := sir.(nodeInfo)
-				pi := &PeerInfo{
-					Name:    si.name,
-					Current: current,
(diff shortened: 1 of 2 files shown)


derekcollison (approved):
LGTM

ripienaar (approved):
Weird, I thought we knew the peer name as long as he was once known?

Anyway, I have seen the phantom missing peers before and yeah super confusing you think you need to fix something but dont know what.  Not sure this `unknown` though will not leed to more user questions though, but I also dont have a better suggestion so I guess LGTM

derekcollison:
It will be in the cluster section of the streamAssignment (sa).

kozlovic:
@derekcollison First, not sure how to get the sa in clusterInfo() since we don't have the account information, but even then, where would be the server name in this:
```
type streamAssignment struct {
	Client  *ClientInfo   `json:"client,omitempty"`
	Created time.Time     `json:"created"`
	Config  *StreamConfig `json:"stream"`
	Group   *raftGroup    `json:"group"`
	Sync    string        `json:"sync"`
	Subject string        `json:"subject"`
	Reply   string        `json:"reply"`
	Restore *StreamState  `json:"restore_state,omitempty"`
	// Internal
	consumers map[string]*consumerAssignment
	responded bool
	err       error
}
```
The `Client` (ClientInfo) server/cluster is about the client that issued the request, when I try with that, I do not get the server name that is actually missing.

kozlovic:
@ripienaar @derekcollison I don't think we have the server name (on in nodesToInfo that is updated only when the nodes connect to each other). However, I have updated the text to better describe that this is a transient situation.

ripienaar:
Looks good, concur we dont know the name yet

PR #3348, opened 2022-08-08
[IMPROVED] Catchup improvements

derekcollison:
Some general stability improvements to catchup logic.

/cc @nats-io/core


server/jetstream_cluster_test.go
```diff
@@ -12041,3 +12041,206 @@ func TestJetStreamClusterDirectGetFromLeafnode(t *testing.T) {
...
+
+	// Now scale up to R3.
+	cfg.Replicas = 3
+	_, err = js.UpdateStream(cfg)
+	require_NoError(t, err)
+	nl := c.randomNonStreamLeader("$G", "TEST")
+	c.waitOnStreamCurrent(nl, "$G", "TEST")
+
+	// Make sure we only sent the number of catchup msgs we expected.
+	if nmsgs, _, _ := sub.Pending(); nmsgs != (cfg.Replicas-1)*(num+1) {
```
kozlovic, line 12194: I was getting failures when running this locally, so would recommend making that a checkFor:
```
checkFor(t, 5*time.Second, 50*time.Millisecond, func() error {
		if nmsgs, _, _ := sub.Pending(); nmsgs != (cfg.Replicas-1)*(num+1) {
			return fmt.Errorf("expected %d catchup msgs, but got %d", (cfg.Replicas-1)*(num+1), nmsgs)
		}
		return nil
	})
```

server/jetstream_cluster.go
```diff
@@ -6610,13 +6641,23 @@ RETRY:
 			notActive.Reset(getActivityInterval())
 
 			mrecs := msgsQ.pop()
+
+			// Send acks first for longer RTT situations.
+			for _, mreci := range mrecs {
```
kozlovic, line 6646: Drawback with that is that remote my send the next batch although we may be failing in middle of a batch (in code below), but arguably, this would clearly not be the happy path!

server/jetstream_cluster.go
```diff
@@ -6491,7 +6490,7 @@ func (mset *stream) processSnapshot(snap *streamSnapshot) (e error) {
 		if gotMsgs || activityInterval == maxActivityInterval {
 			return maxActivityInterval
 		}
-		activityInterval *= 2
+		activityInterval *= 5
```
kozlovic, line 6493: That will limit this whole thing to likely just 1 retry (1sec -> 5sec and then cap of 10sec - as opposed to 1s -> 2s -> 4s -> 8s-> 10s). As discussed before, I would be ok that we remove all that and simply adjust one of the test that was sensitive of the fact that the first catchup request was missed by the leader (before itself was just becoming leader and the sync request subscription was not fully setup).

server/jetstream_cluster.go
kozlovic: I was planning on proposing that the other side does "wait" (with a max wait) for say 1/2 of the previous batch size before sending new batch: basically the sub that updates the outm would have a chance to execute before the next batch is sent allowing to send more at a time. But maybe this approach here is as good (and much simpler).

server/jetstream_cluster.go
```diff
@@ -6637,11 +6678,21 @@ RETRY:
 				} else {
 					s.Warnf("Catchup for stream '%s > %s' errored, will retry: %v", mset.account(), mset.name(), err)
 					msgsQ.recycle(&mrecs)
+
```
kozlovic, line 6681: I was thinking that in any condition where we return an error or we `goto RETRY`, we should use the current mrec's reply to send the error back to the leader. The leader does not care of the content at the moment, so it would be backward compatible, but in runCatchup() we could have the sub check that if the FC reply's body is not empty, we know that the remote has stopped the catchup and the body content is the error message.
We would need to coordinate with that sub and the rest of the go routine loop, but that is perfectly doable. This would help by "releasing" resources on the runCatchup because otherwise that routine will wait up to 5 seconds without activity from the remote before returning (which then releases the outb that counts toward the server-wide limit if I am not mistaken).

kozlovic (changes requested):
Need to add a checkFor() in a test to avoid flapping and made some suggestions.

server/jetstream_cluster.go
derekcollison: I plan on testing on NGS when this lands, so we will see. Might need more tweaking.

server/jetstream_cluster.go
derekcollison: Which test? Do you recall?

server/jetstream_cluster.go
derekcollison: I like that idea, once I free up will take a look.

server/jetstream_cluster.go
kozlovic: Yes, it was TestNoRaceJetStreamClusterDirectAccessAllPeersSubs. Maybe just change the timeout here to more than 10 seconds (say 20secs): https://github.com/nats-io/nats-server/blob/d90854a45fe9405198093ab1489f8b3e5e11dcf8/server/norace_test.go#L5318. 
Also this check has been already extended to 20secs (used to be 10 originally I think): https://github.com/nats-io/nats-server/blob/d90854a45fe9405198093ab1489f8b3e5e11dcf8/server/norace_test.go#L5343

derekcollison: ok will do that and remove dynamic activityInterval logic.

kozlovic (approved):
LGTM for current changes, not sure if you want to add the "return error to leader" as a new commit in this PR or not.

derekcollison:
I say we do, I am tied up for a few hours, want me to merge and you have a go at it? If not will circle back in a few hours.

kozlovic:
@derekcollison I rebased, resolved the conflicts and push forced. Will check that we have green and then I can merge.

PR #3349, opened 2022-08-08
[IMPROVED] JetStream catchup can be aborted and better flow control

kozlovic:
If the leader sends messages but the follower for any reason aborts
or retry the snapshot process, it will now send the error that
caused this and the leader can then abort the catchup instead of
waiting for its inactivity threshold of 5 seconds.

Also make the send of a batch be delayed for a bit until the number
of "acks" is 1/2 of the batch size or after reaching 100ms. This
helps avoid trickling of messages. Tested with the new test
TestJetStreamSuperClusterStreamCathupLongRTT() and see better
results both in size of batches and overall time is smaller or
similar but not longer.

### DIFF
--- a/server/jetstream_cluster.go
+++ b/server/jetstream_cluster.go
@@ -6497,6 +6497,14 @@ func (mset *stream) isCurrent() bool {
 // Maximum requests for the whole server that can be in flight.
 const maxConcurrentSyncRequests = 8
 
+var (
+	errCatchupCorruptSnapshot = errors.New("corrupt stream snapshot detected")
+	errCatchupStalled         = errors.New("catchup stalled")
+	errCatchupStreamStopped   = errors.New("stream has been stopped") // when a catchup is terminated due to the stream going away.
+	errCatchupBadMsg          = errors.New("bad catchup msg")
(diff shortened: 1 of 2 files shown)


server/jetstream_cluster.go
```diff
@@ -6752,6 +6758,8 @@ RETRY:
 			msgsQ.recycle(&mrecs)
 		case <-notActive.C:
 			if mrecs := msgsQ.pop(); len(mrecs) > 0 {
+				mrec := mrecs[0].(*im)
+				notifyLeaderStopCatchup(mrec, fmt.Errorf("catchup stalled"))
```
derekcollison, line 6762: Make a var error with errors.New?

derekcollison (approved):
LGTM!

server/jetstream_cluster.go
kozlovic: I see we do that even inside functions, which kind of does not make much sense because there will still be a new alloc per invocation. So I think it should be done outside of function (like a global) no?

kozlovic: Check last commit where I did extract some of the catchup errors and declared them outside of the funcs.

derekcollison (approved):
LGTM!

PR #3350, opened 2022-08-09
Fix race between stream stop and monitorStream

matthiashanel:
monitorCluster stops the stream, when doing so, monitorStream
needs to be stopped to avoid miscounting of store size.
In a test stop and reset of store size happened first and then
was followed by storing more messages via monitorStream

 - [ ] Tests added

Resolves #

 -
 -
 -

/cc @nats-io/core

### DIFF
--- a/server/jetstream_cluster.go
+++ b/server/jetstream_cluster.go
@@ -2011,6 +2011,8 @@ func (js *jetStream) monitorStream(mset *stream, sa *streamAssignment, sendSnaps
 			// If we were successful lookup up our stream now.
 			if err == nil {
 				if mset, err = acc.lookupStream(sa.Config.Name); mset != nil {
+					mset.monitorWg.Add(1)
+					defer mset.monitorWg.Done()
 					mset.setStreamAssignment(sa)
 					// Make sure to update our updateC which would have been nil.
 					uch = mset.updateC()
@@ -2736,10 +2738,8 @@ func (js *jetStream) processUpdateStreamAssignment(sa *streamAssignment) {
 				node.StepDown(sa.Group.Preferred)
 			}
 			node.ProposeRemovePeer(ourID)
-		}
-		// shut down monitor by shutting down raft
-		if n := mset.node; n != nil {
-			n.Delete()
+			// shut down monitor by shutting down raft
+			node.Delete()
 		}
 		// wait for monitor to be shut down
 		mset.monitorWg.Wait()
@@ -3099,7 +3099,7 @@ func (js *jetStream) processClusterDeleteStream(sa *streamAssignment, isMember, 
 	if acc, _ = s.LookupAccount(sa.Client.serviceAccount()); acc != nil {
 		if mset, _ := acc.lookupStream(sa.Config.Name); mset != nil {
 			// shut down monitor by shutting down raft
-			if n := mset.node; n != nil {
+			if n := mset.raftNode(); n != nil {
 				n.Delete()
 			}
 			// wait for monitor to be shut down
--- a/server/jetstream_cluster.go
+++ b/server/jetstream_cluster.go
@@ -1478,6 +1478,7 @@ func (js *jetStream) applyMetaEntries(entries []*Entry, rm *recoveryRemovals) (b
 				if isRecovering {
 					js.setStreamAssignmentRecovering(sa)
 				}
+
 				js.processUpdateStreamAssignment(sa)
 			default:
 				panic("JetStream Cluster Unknown meta entry op type")
@@ -1647,7 +1648,9 @@ func currentPeerCount(ci *ClusterInfo, peerSet []string, leaderId string) (curre
 func (js *jetStream) monitorStream(mset *stream, sa *streamAssignment, sendSnapshot bool) {
 	s, cc := js.server(), js.cluster
 	defer s.grWG.Done()
-
+	if mset != nil {
+		defer mset.monitorWg.Done()
+	}
 	js.mu.RLock()
 	n := sa.Group.node
 	meta := cc.meta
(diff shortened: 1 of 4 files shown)


server/jetstream_cluster.go
```diff
@@ -1647,7 +1648,9 @@ func currentPeerCount(ci *ClusterInfo, peerSet []string, leaderId string) (curre
 func (js *jetStream) monitorStream(mset *stream, sa *streamAssignment, sendSnapshot bool) {
 	s, cc := js.server(), js.cluster
 	defer s.grWG.Done()
-
+	if mset != nil {
+		defer mset.monitorWg.Done()
```
derekcollison, line 1652: Should we check below when we create one of these from restore and set it there as well

server/jetstream_cluster.go
```diff
@@ -2763,6 +2772,7 @@ func (js *jetStream) processClusterUpdateStream(acc *Account, osa, sa *streamAss
 				mset.setLeader(false)
 				js.createRaftGroup(acc.GetName(), rg, storage)
 			}
+			mset.monitorWg.Add(1)
```
derekcollison, line 2775: Should we just put this in startup of monitorStream? If mset != nil?

server/jetstream_cluster.go
```diff
@@ -2938,6 +2948,9 @@ func (js *jetStream) processClusterCreateStream(acc *Account, sa *streamAssignme
 	// Start our monitoring routine.
 	if rg.node != nil {
 		if !alreadyRunning {
+			if mset != nil {
+				mset.monitorWg.Add(1)
```
derekcollison, line 2952: Same here.

server/jetstream_cluster.go
```diff
@@ -3085,6 +3098,12 @@ func (js *jetStream) processClusterDeleteStream(sa *streamAssignment, isMember,
 	// Go ahead and delete the stream if we have it and the account here.
 	if acc, _ = s.LookupAccount(sa.Client.serviceAccount()); acc != nil {
 		if mset, _ := acc.lookupStream(sa.Config.Name); mset != nil {
+			// shut down monitor by shutting down raft
```
derekcollison, line 3101: I wonder if we should either make function called kickMonitor() and possibly make its own channel which we create at the beginning on monitorStream() as long a mset != nil and we simple close this channel in kickMonitor?

kozlovic (changes requested):

server/jetstream_cluster.go
```diff
@@ -1647,7 +1648,9 @@ func currentPeerCount(ci *ClusterInfo, peerSet []string, leaderId string) (curre
 func (js *jetStream) monitorStream(mset *stream, sa *streamAssignment, sendSnapshot bool) {
 	s, cc := js.server(), js.cluster
 	defer s.grWG.Done()
-
+	if mset != nil {
```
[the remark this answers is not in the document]
matthiashanel: not sure I follow, we only defer when started with `mset != nil`. Setting it later will not cause a defer.

server/jetstream_cluster.go
```diff
@@ -2734,6 +2737,12 @@ func (js *jetStream) processUpdateStreamAssignment(sa *streamAssignment) {
 			}
 			node.ProposeRemovePeer(ourID)
 		}
+		// shut down monitor by shutting down raft
+		if n := mset.node; n != nil {
```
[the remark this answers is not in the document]
matthiashanel: done

server/jetstream_cluster.go
```diff
@@ -3085,6 +3098,12 @@ func (js *jetStream) processClusterDeleteStream(sa *streamAssignment, isMember,
 	// Go ahead and delete the stream if we have it and the account here.
 	if acc, _ = s.LookupAccount(sa.Client.serviceAccount()); acc != nil {
 		if mset, _ := acc.lookupStream(sa.Config.Name); mset != nil {
+			// shut down monitor by shutting down raft
+			if n := mset.node; n != nil {
```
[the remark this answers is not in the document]
matthiashanel: done

server/jetstream_cluster.go
matthiashanel: is it possible to have monitoStream starte, exit, be started again?

server/jetstream_cluster.go
matthiashanel: added

server/jetstream_cluster.go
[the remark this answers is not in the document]
matthiashanel: ok, I wouldn't call a done because of this.
But agree that we would be blind to a restored stream. adding add/defer done where it's set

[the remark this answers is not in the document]
kozlovic: Sorry... I missed that the check for mset != nil is done outside of the defer.

server/jetstream_cluster.go
```diff
@@ -2008,6 +2011,8 @@ func (js *jetStream) monitorStream(mset *stream, sa *streamAssignment, sendSnaps
 			// If we were successful lookup up our stream now.
 			if err == nil {
 				if mset, err = acc.lookupStream(sa.Config.Name); mset != nil {
+					mset.monitorWg.Add(1)
+					defer mset.monitorWg.Done()
```
kozlovic, line 2015: So even if - not sure if that is possible - we enter with mset != nil and end-up in this code, the number of Add() and Done() would match, so that is ok!

kozlovic (approved):
LGTM

server/jetstream_cluster.go
matthiashanel: basically

PR #3351, opened 2022-08-09
[IMPROVED] DirectGet performance and memory usage for large streams.

derekcollison:
This PR is to improve the worst case lookup times for direct get by subject, e.g. KV.

When the number of blocks became very large, the linear scan walking backwards for the correct block, and the linear scan walking forward to find the previous entry to delete was becoming slow. This would become evident in > 100M messages or small block sizes, e.g. 4k.

We now use better indexing for lookups by soft tracking the first and last block per subject.
We also improve memory usage by some degree on very large stores by swapping out the per block subject info and swap out the indexing data structures more aggressively.

This approximately doubles the number of messages we can store and manage with direct gets.

More work to be done but this is a start.

 - [ ] Tests added

Resolves #

 -
 -
 -

/cc @nats-io/core

### DIFF
--- a/server/filestore.go
+++ b/server/filestore.go
@@ -76,6 +76,12 @@ const (
 	defaultFilePerms = os.FileMode(0640)
 )
 
+type psi struct {
+	total uint64
+	fblk  uint32
+	lblk  uint32
+}
+
 type fileStore struct {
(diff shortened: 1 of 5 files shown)


kozlovic (approved):
LGTM

PR #3354, opened 2022-08-10
fix move cancel issue where tags and peers diverge

matthiashanel:
This can happen if the move was initiated by the user.
A subsequent cancel resets the initial peer list.
The original peer list was picked on the old set of tags.
A cancel would then keep the new list of tags but reset
to the old peers. Thus tags and peers diverge.

The problem is that at the time of cancel, the old
placement tags can't be found anymore.

This fix causes cancel to remove the placement tags, if
the old peers do not satisfy the new placement tags.


derekcollison (approved):
LGTM

kozlovic (approved):
LGTM

PR #3355, opened 2022-08-10
fixed consumer restart on source filter update

matthiashanel:
When a stream source filter subject was updated, the internal consumer
was not re created

If the upstream stream contains a tail of previously filtered messages,
these will now be delivered

PLEASE check my method to restart and if my use of mset.setSourceConsumer(s.iname, si.sseq+1) is the right thing to do here. Test etc... look al right. just want to point this one out.

### DIFF
--- a/server/stream.go
+++ b/server/stream.go
@@ -1397,13 +1397,13 @@ func (mset *stream) updateWithAdvisory(config *StreamConfig, sendAdvisory bool) 
 
 		// Check for Sources.
 		if len(cfg.Sources) > 0 || len(ocfg.Sources) > 0 {
-			current := make(map[string]struct{})
+			current := make(map[string]string)
 			for _, s := range ocfg.Sources {
-				current[s.iname] = struct{}{}
+				current[s.iname] = s.FilterSubject
 			}
 			for _, s := range cfg.Sources {
 				s.setIndexName()
-				if _, ok := current[s.iname]; !ok {
+				if oFilter, ok := current[s.iname]; !ok {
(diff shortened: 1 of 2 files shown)


tbeets:
In a server restart scenario, was the internal consumer hydrated with the updated config from store at that time?  Is this fix making sure that the config update (source change) happens immediately?

derekcollison (approved):
LGTM - I think that looks correct.

server/stream.go
```diff
@@ -1412,6 +1412,13 @@ func (mset *stream) updateWithAdvisory(config *StreamConfig, sendAdvisory bool)
 					mset.sources[s.iname] = si
 					mset.setStartingSequenceForSource(s.iname)
 					mset.setSourceConsumer(s.iname, si.sseq+1)
+				} else if oFilter != s.FilterSubject {
+					if si, ok := mset.sources[s.iname]; ok {
+						mset.cancelSourceInfo(si)
```
kozlovic, line 1417: I believe this is not necessary since setSourceConsumer will cancel if 'iname' is found in mset.sources, but that is ok I guess.

kozlovic (approved):
LGTM

server/stream.go
matthiashanel: yes, agreed.

matthiashanel:
> In a server restart scenario, was the internal consumer hydrated with the updated config from store at that time? Is this fix making sure that the config update (source change) happens immediately?

@tbeets yes, happens immediately now

PR #3356, opened 2022-08-10
Added check for source/mirror filter subjects

matthiashanel:
When the origin stream exists, the sourec/mirror filter subject
will be checked against the stream subjects.
If there is no overlap, an error will be returned

### DIFF
--- a/server/stream.go
+++ b/server/stream.go
@@ -1007,6 +1007,20 @@ func (s *Server) checkStreamCfg(config *StreamConfig, acc *Account) (StreamConfi
 		return exists, cfg.MaxMsgSize, cfg.Subjects
 	}
 
+	hasFilterSubjectOverlap := func(filter string, streamSubs []string) bool {
+		if filter == _EMPTY_ || len(streamSubs) == 0 {
+			return true
+		}
+		subjectOverlap := false
+		for _, sub := range streamSubs {
+			if SubjectsCollide(sub, filter) {
+				subjectOverlap = true
+				break
+			}
+		}
+		return subjectOverlap
+	}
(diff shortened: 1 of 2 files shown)


server/stream.go
```diff
@@ -1007,6 +1007,20 @@ func (s *Server) checkStreamCfg(config *StreamConfig, acc *Account) (StreamConfi
 		return exists, cfg.MaxMsgSize, cfg.Subjects
 	}
 
+	hasFilterSubjectOverlap := func(filter string, streamSubs []string) bool {
+		if filter == _EMPTY_ || len(streamSubs) == 0 {
+			return true
+		}
+		subjectOverlap := false
```
derekcollison, line 1014: No need for this, when you see condition just return true.

server/stream.go
```diff
@@ -1007,6 +1007,20 @@ func (s *Server) checkStreamCfg(config *StreamConfig, acc *Account) (StreamConfi
...
+		for _, sub := range streamSubs {
+			if SubjectsCollide(sub, filter) {
+				subjectOverlap = true
```
derekcollison, line 1017: Just return true here I think.

derekcollison (approved):
LGTM - Minor nit comment

kozlovic (approved):
LGTM but agree with Derek that the function could simply return true/false without need for the boolean declaration.

server/stream.go
matthiashanel: fixed

server/stream.go
matthiashanel: fixed

PR #3358, opened 2022-08-10
Fixed message timestamp formatting for direct message get feature

kozlovic:
In normal message get, the returned format is RFC3339Nano, which
is what is being used when using JSON marshaling. However, for
the direct get we had to pass a string to construct the header
and we were using time.Time.String() which was using a different
layout. So use ts.Format(time.RFC3339Nano) to be consistent with
the non-direct get message.

Libraries that already parsed the non RFC3339Nano time format
can be updated since none should have been released yet (since
the feature in the server is not released yet)

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -6,7 +6,7 @@ require (
 	github.com/klauspost/compress v1.15.9
 	github.com/minio/highwayhash v1.0.2
 	github.com/nats-io/jwt/v2 v2.3.0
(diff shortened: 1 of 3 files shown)


ripienaar (approved):
LGTM

scottf (approved):
Tested against this branch and get direct message returned proper RFC3339 format

PR #3359, opened 2022-08-10
fix unit test

matthiashanel:

### DIFF
--- a/server/jetstream_super_cluster_test.go
+++ b/server/jetstream_super_cluster_test.go
@@ -3082,7 +3082,7 @@ func TestJetStreamSuperClusterPeerEvacuationAndStreamReassignment(t *testing.T) 
 	nc, js := jsClientConnect(t, srv)
 	defer nc.Close()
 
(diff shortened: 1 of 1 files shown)


server/jetstream_super_cluster_test.go
```diff
@@ -3208,7 +3208,9 @@ func TestJetStreamSuperClusterPeerEvacuationAndStreamReassignment(t *testing.T)
 			if !listFrom {
 				// when needed determine which server move moved away from
 				si, err := js.StreamInfo("TEST", nats.MaxWait(2*time.Second))
-				require_NoError(t, err)
+				if err != nil {
```
kozlovic, line 3211: Yeah, in checkFor(), any "expected" error need to be a return fmt.Errorf(), since t.Fatal() will stop the test. Now Derek sometimes uses require_NoError() inside checkFor() in cases where arguably the call should not fail, regardless of time.

kozlovic (approved):
LGTM

PR #3363, opened 2022-08-11
[FIXED] JetStream: subject overlap error should be returned

### LINKED ISSUE #3362: JetStream: non specific error returned with overlapping subjects in standalone
This was reported as an issue in the C client (https://github.com/nats-io/nats.c/issues/572), but it turns out to be an issue that the server does not return the proper NewJSStreamSubjectOverlapError() error in standalone mode (but does it correctly in cluster mode).

Will have a PR to fix this.

kozlovic:
In standalone mode, when attempting to create a stream which has
subjects that overlap with an existing stream, the generic
stream create error "10049" was returned instead of the more
accurate "10065" error code corresponding to subject overlap,
as it was the case in clustered mode.

Resolves #3362

### DIFF
--- a/server/jetstream_test.go
+++ b/server/jetstream_test.go
@@ -18908,3 +18908,45 @@ func TestJetStreamDirectGetAutoSet(t *testing.T) {
 		t.Fatalf("Expected to see the mirror respond at least once")
 	}
 }
+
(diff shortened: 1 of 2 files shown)


derekcollison (approved):
LGTM

PR #3364, opened 2022-08-11
fixed consumer source update receiving upstream msgs

matthiashanel:
if an origin stream contains:
1M msgs with subject foo and 1M msgs with subject bar

IF the source consumer changes their filter from foo to bar
Then it would have received messages for subject bar.
This happens because this tail was filtered and their
respective seqno was not communicated to the consumer.

This is somewhat unexpected. It is also coincidental.
Had the last message in the stream had subject foo then
this wouldn't happen.

Therefore, when completely changing the subject say,
from foo to bar, we only receive messages received
after the time the change was made.

However, if the old and new subject overlap in any way,
we go by sequence number. Meaning in these cases the
outlined behavior remains in order to not induce artificial
message loss for the part of the subject space that is
covered by old and new filter.

 - [ ] Tests added

Resolves #

 -
 -
 -

/cc @nats-io/core

### DIFF
--- a/server/jetstream_cluster_test.go
+++ b/server/jetstream_cluster_test.go
@@ -5431,9 +5431,22 @@ func TestJetStreamClusterSourcesFilterSubjectUpdate(t *testing.T) {
 		Replicas: 2,
 	})
 	require_NoError(t, err)
-	// no send was necessary as we received previously filtered messages
-	checkSync(500, 300)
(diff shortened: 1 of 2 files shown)


kozlovic (approved):
LGTM

derekcollison (approved):
LGTM

PR #3365, opened 2022-08-11
[FIXED] JetStream: issue with max deliver and server/cluster restart

### LINKED ISSUE #3361: Zombie re-deliveries
## Defect

If you have a durable consumer with explicit ack and some message reached the maximum number of delivery attempts, if you restart the nats-server then the server tries again to re-deliver those messages.

2.8.4

Create a storage=file stream and a durable (push or pull), start a subscriber on that consumer that doesn't acknowledge messages (e.g. `nats sub --no-akc` or `nats consumer next --no-ack`) publish a message into the stream and see that the message is delivered only n times (n = max deliveries), after that is over, stop and restart the nats-server(s) and after that restart the server attempts to deliver that message again.

kozlovic:
This is a regression introduced in v2.8.3. If a message reaches
the max redeliver count, it stops being delivered to the consumer.
However, after a server or cluster restart, those messages would
be redelivered again.

Resolves #3361

### DIFF
--- a/server/norace_test.go
+++ b/server/norace_test.go
@@ -3562,6 +3562,8 @@ func TestNoRaceJetStreamClusterCorruptWAL(t *testing.T) {
 	}
 	// Grab underlying raft node and the WAL (filestore) and we will attempt to "corrupt" it.
 	node := o.raftNode().(*raft)
+	// We are doing a stop here to prevent the internal consumer snapshot from happening on exit
+	node.Stop()
 	fs := node.wal.(*fileStore)
 	fcfg, cfg := fs.fcfg, fs.cfg.StreamConfig
 	// Stop all the servers.
--- a/server/norace_test.go
+++ b/server/norace_test.go
@@ -2183,7 +2183,7 @@ func TestNoRaceJetStreamClusterMirrorExpirationAndMissingSequences(t *testing.T)
 	checkMirror(20)
 }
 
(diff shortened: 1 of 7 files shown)


derekcollison (approved):
LGTM

kozlovic:
@derekcollison I am seeing that the cluster test is flapping, so will investigate more. Not merging at this time until I figure out if it is test timing or something more.

server/norace_test.go
```diff
@@ -3561,6 +3561,9 @@ func TestNoRaceJetStreamClusterCorruptWAL(t *testing.T) {
 		t.Fatalf("Error looking up consumer %q", "dlc")
 	}
 	// Grab underlying raft node and the WAL (filestore) and we will attempt to "corrupt" it.
+	o.mu.Lock()
+	o.skipSnapshotOnExit = true
```
derekcollison, line 3565: I think I would rather pull the consumer in this test and stop the raft node directly versus having a boolean for every consumer in normal operations.

So just call o.raftNode().Stop()

derekcollison (changes requested):

server/norace_test.go
kozlovic: Removed the boolean and did just that. Force pushed this commit.

derekcollison (approved):
LGTM

PR #3366, opened 2022-08-12
[FIXED] Make sure when SubjectState is called we have loaded fss state.

derekcollison:

/cc @nats-io/core

### DIFF
--- a/server/filestore.go
+++ b/server/filestore.go
@@ -1536,9 +1536,29 @@ func (fs *fileStore) SubjectsState(subject string) map[string]SimpleState {
 		return nil
 	}
 
+	start, stop := fs.blks[0], fs.lmb
+	// We can short circuit if not a wildcard using psim for start and stop.
+	if !subjectHasWildcard(subject) {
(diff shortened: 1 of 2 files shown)


kozlovic (commented):
I have checked "all" references of mb.fss to see if there were other cases where we did not make sure that the per-subject info was loaded. The only place I found is writeMsgRecord() where we say that if `mb.fss != nil` then we track per-subject info. Does that mean that if the state was flushed then we would no longer keep track of new messages?

kozlovic (approved):
I was altering the new test to see if I could make it fail but I couldn't. I did increase the block size to 32K and yes, I can verify that in that case after expiration the writeMsgRecord does not track the subject, however since we ensure that we rebuild when checking for mb.fss then it should be ok. I would like confirmation that we are ok in that situation (expiration and then write that are not tracked) before LGTM it.

kozlovic (approved):
LGTM

derekcollison:
we don't need fss load to write a new msg, we will update psim but fss not needed.

PR #3371, opened 2022-08-16
[ADDED] Support for AES-GCM as a cipher along with ChaChaPoly. 

derekcollison:
Also supports upgrade to different cipher for existing store assets.

/cc @nats-io/core

### DIFF
--- a/server/filestore.go
+++ b/server/filestore.go
@@ -16,6 +16,7 @@ package server
 import (
 	"archive/tar"
 	"bytes"
+	"crypto/aes"
 	"crypto/cipher"
 	"crypto/rand"
 	"crypto/sha256"
@@ -55,12 +56,32 @@ type FileStoreConfig struct {
 	SyncInterval time.Duration
 	// AsyncFlush allows async flush to batch write operations.
 	AsyncFlush bool
+	// Cipher is the cipher to use when encrypting.
+	Cipher StoreCipher
 }
 
 // FileStreamInfo allows us to remember created time.
 type FileStreamInfo struct {
 	Created time.Time
 	StreamConfig
+}
+
+type StoreCipher int
+
+const (
+	ChaCha StoreCipher = iota
+	AES
+)
+
+func (cipher StoreCipher) String() string {
+	switch cipher {
+	case ChaCha:
+		return "ChaCha"
+	case AES:
+		return "AES"
+	default:
+		return "Unknown StoreCipher"
+	}
 }
 
 // File ConsumerInfo is used for creating consumer stores.
@@ -113,7 +134,7 @@ type msgBlock struct {
 	mu      sync.RWMutex
 	fs      *fileStore
 	aek     cipher.AEAD
-	bek     *chacha20.Cipher
+	bek     cipher.Stream
 	seed    []byte
 	nonce   []byte
 	mfn     string
@@ -215,8 +236,8 @@ const (
 	JetStreamMetaFileKey = "meta.key"
 
 	// AEK key sizes
-	metaKeySize = 72
-	blkKeySize  = 72
+	minMetaKeySize = 64
+	minBlkKeySize  = 64
 
 	// Default stream block size.
 	defaultLargeBlockSize = 8 * 1024 * 1024 // 8MB
@@ -444,7 +465,7 @@ func dynBlkSize(retention RetentionPolicy, maxBytes int64) uint64 {
 }
 
 // Generate an asset encryption key from the context and server PRF.
-func (fs *fileStore) genEncryptionKeys(context string) (aek cipher.AEAD, bek *chacha20.Cipher, seed, encrypted []byte, err error) {
+func (fs *fileStore) genEncryptionKeys(context string) (aek cipher.AEAD, bek cipher.Stream, seed, encrypted []byte, err error) {
 	if fs.prf == nil {
 		return nil, nil, nil, nil, errNoEncryption
 	}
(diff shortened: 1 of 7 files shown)


server/filestore.go
```diff
@@ -453,31 +474,71 @@ func (fs *fileStore) genEncryptionKeys(context string) (aek cipher.AEAD, bek *ch
 	if err != nil {
 		return nil, nil, nil, nil, err
 	}
-	kek, err := chacha20poly1305.NewX(rb)
+
+	sc := fs.fcfg.Cipher
+
+	var kek cipher.AEAD
+	if sc == ChaCha {
```
kozlovic, line 481: Looks like this block repeats 11 times in this PR, I think it justifies being made a function? Just pass the sc and rb.

kozlovic (changes requested):

server/filestore.go
derekcollison: Will take a look.

server/filestore.go
```diff
@@ -453,31 +493,60 @@ func (fs *fileStore) genEncryptionKeys(context string) (aek cipher.AEAD, bek *ch
...
-	if n, err := rand.Read(seed); err != nil || n != 32 {
+
+	const seedSize = 32
+	seed = make([]byte, seedSize)
+	if n, err := rand.Read(seed); err != nil || n != seedSize {
 		return nil, nil, nil, nil, err
 	}
-	aek, err = chacha20poly1305.NewX(seed)
+
+	if sc == ChaCha {
```
kozlovic, line 511: Use the new function that you created?

server/filestore.go
```diff
@@ -333,6 +354,10 @@ func newFileStoreWithCreated(fcfg FileStoreConfig, cfg StreamConfig, created tim
 			if err := fs.writeStreamMeta(); err != nil {
 				return nil, err
 			}
+		} else if err != nil {
+			return nil, err
+		} else {
+			// Check to make sure we can read our key file.
```
kozlovic, line 360: Something missing here or what?

server/filestore.go
```diff
@@ -611,28 +681,44 @@ func (fs *fileStore) recoverMsgBlock(fi os.FileInfo, index uint32) (*msgBlock, e
...
-				return nil, err
+				// We may be here on a cipher conversion, so attempt to convert.
+				if err = mb.convertCipher(); err != nil {
+					return nil, err
+				}
+			} else {
+				mb.seed, mb.nonce = seed, ekey[:ns]
+			}
+
+			if sc == ChaCha {
```
kozlovic, line 709: Again, we can call the new function here I think.

server/filestore.go
```diff
@@ -5571,21 +5727,40 @@ func (fs *fileStore) ConsumerStore(name string, cfg *ConsumerConfig) (ConsumerSt
...
+			nonce := ekey[:ns]
+			seed, err := kek.Open(nil, nonce, ekey[ns:], nil)
 			if err != nil {
-				return nil, err
+				// We may be here on a cipher conversion, so attempt to convert.
+				if err = o.convertCipher(); err != nil {
+					return nil, err
+				}
+			} else {
+				if sc == ChaCha {
```
kozlovic, line 5753: I don't see the use of the new function...

server/jetstream.go
```diff
@@ -252,11 +250,20 @@ func (s *Server) decryptMeta(ekey, buf []byte, acc, context string) ([]byte, err
 	if err != nil {
 		return nil, err
 	}
-	aek, err := chacha20poly1305.NewX(seed[:])
+	var aek cipher.AEAD
+	if sc == ChaCha {
```
kozlovic, line 254: Use the new function

server/filestore.go
```diff
@@ -5640,6 +5815,74 @@ func (fs *fileStore) ConsumerStore(name string, cfg *ConsumerConfig) (ConsumerSt
...
+	}
+	ns := kek.NonceSize()
+	nonce := ekey[:ns]
+	seed, err := kek.Open(nil, nonce, ekey[ns:], nil)
+	if err != nil {
+		return err
+	}
+
+	var aek cipher.AEAD
+	if sc == AES {
```
kozlovic, line 5853: Can use the new function here too...

kozlovic (changes requested):

server/filestore.go
kozlovic: Still missing...

server/filestore.go
kozlovic: That one too.

server/filestore.go
kozlovic: That one too.

server/filestore.go
kozlovic: Sorry, this is conversion so you are reversing things...

server/jetstream.go
kozlovic: Use new function.

kozlovic (changes requested):
Looks like you still missed some places where you can use the function, however, the one with the conversion can't use it, but maybe adding a comment here to make it clear!

derekcollison:
Can't use it for aek which requires a seed. kek requires rb and sc. I could add a convenience function for aeks if you feel strongly about it.

kozlovic:
But you pass a byte slice:`rb` and `seed` are both `[]byte` so what's the difference? Meaning it looks like the same chacha or aes functions are invoked in both cases, just that the param passed to them is different byte slice content, but...

Here is a line where I commented that it is still missing: https://github.com/nats-io/nats-server/pull/3371/files#diff-384c189826934c9a6fc3554dafc63dab2076245010e3d6fce5c71a93e15e9877R507
```
if sc == ChaCha {
	aek, err = chacha20poly1305.NewX(seed)
} else if sc == AES {
	block, e := aes.NewCipher(seed)
	if e != nil {
		return nil, nil, nil, nil, err
	}
	aek, err = cipher.NewGCMWithNonceSize(block, block.BlockSize())
}
```
how is that code different from the function:
```
func genKEK(sc StoreCipher, rb []byte) (kek cipher.AEAD, err error) {
	if sc == ChaCha {
		kek, err = chacha20poly1305.NewX(rb)
	} else if sc == AES {
		block, e := aes.NewCipher(rb)
		if e != nil {
			return nil, err
		}
		kek, err = cipher.NewGCMWithNonceSize(block, block.BlockSize())
	} else {
		err = errUnknownCipher
	}
	return kek, err
}
```

derekcollison:
Will take another look.

derekcollison:
You are correct, updated. Thanks!

kozlovic (approved):
LGTM

PR #3373, opened 2022-08-17
fix jwt template ordering issue and error message

matthiashanel:
ordering of templates got messed up by a map (now removed)
Also improved error message when template generation fails


kozlovic (approved):
LGTM

PR #3374, opened 2022-08-17
[FIXED] Gateway: possible panic if monitor endpoint inspected too soon

kozlovic:
The monitoring http server is started early and the gateway setup
(when configured) may not be fully ready when the `/gatewayz`
endpoint is inspected and could cause a panic.


derekcollison (approved):
LGTM

PR #3377, opened 2022-08-17
When filestore is used for raft layer do not attempt to track subjects

derekcollison:

/cc @nats-io/core

### DIFF
--- a/server/filestore.go
+++ b/server/filestore.go
@@ -5098,12 +5098,10 @@ func (mb *msgBlock) generatePerSubjectInfo(hasLock bool) error {
 		return nil
 	}
 
-	var shouldExpire bool
 	if mb.cacheNotLoaded() {
 		if err := mb.loadMsgsWithLock(); err != nil {
 			return err
 		}
-		shouldExpire = true
 	}
 
 	// Create new one regardless.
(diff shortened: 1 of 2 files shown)


server/filestore.go
```diff
@@ -5213,6 +5221,10 @@ func (mb *msgBlock) readPerSubjectInfo(hasLock bool) error {
 		}
 	}()
 
+	// Make sure we run the cache expire timer.
+	mb.llts = time.Now().UnixNano()
```
kozlovic, line 5225: Btw, should this be protected by the lock, which is acquired only later on in this function when invoked with !hasLock?

server/filestore.go
```diff
@@ -646,6 +647,11 @@ const indexHdrSize = 7*binary.MaxVarintLen64 + hdrLen + checksumSize
 func (fs *fileStore) recoverMsgBlock(fi os.FileInfo, index uint32) (*msgBlock, error) {
 	mb := &msgBlock{fs: fs, index: index, cexp: fs.fcfg.CacheExpire}
 
+	// Check if we should be monitoring subjects.
+	if len(fs.cfg.Subjects) == 0 {
```
kozlovic, line 651: Humm... I know that fs.cfg.Subjects can change (in fileStore.UpdateConfig()), so does that mean that we would then not track subject at all (because of setting noSubj=true)? Or maybe we need to update this value in all blocks in UpdateConfig(), but then it means that noSubj is no longer immutable and should be checked under lock in readPerSubjectInfo() (since that function can be invoked without the mb lock)

server/filestore.go
derekcollison: True wil fix..

server/filestore.go
derekcollison: I think this ok, upper layers have 1 or more subjects vs raft layer always none.

kozlovic (approved):
LGTM

PR #3378, opened 2022-08-17
We can't purge directories here since not 100% sure all state is in snapshot.

derekcollison:

/cc @nats-io/core


matthiashanel (approved):
LGTM

PR #3381, opened 2022-08-18
induce delay prior to scale down

matthiashanel:
This is to avoid a narrow race between adding server and them catching
up where they also register as current.

Also wait for all peers to be caught up. 

This also avoids clearing catchup marker once catchup stalled.
A stalled catchup would remove the marker causing the peer to
register as current.

### DIFF
--- a/server/jetstream_cluster.go
+++ b/server/jetstream_cluster.go
@@ -7329,6 +7329,7 @@ func (mset *stream) runCatchup(sendSubject string, sreq *streamSyncRequest) {
 		case <-qch:
 			return
 		case <-remoteQuitCh:
+			mset.clearCatchupPeer(sreq.Peer)
 			return
 		case <-notActive.C:
 			s.Warnf("Catchup for stream '%s > %s' stalled", mset.account(), mset.name())
--- a/server/jetstream_cluster.go
+++ b/server/jetstream_cluster.go
@@ -1630,6 +1630,9 @@ func currentPeerCount(ci *ClusterInfo, peerSet []string, leaderId string) (curre
 	return
 }
 
+// how many migration tracker ticks of delay to induce
+const scaleDownDelayTicks = 2
+
 // Monitor our stream node for this stream.
 func (js *jetStream) monitorStream(mset *stream, sa *streamAssignment, sendSnapshot bool) {
 	s, cc := js.server(), js.cluster
@@ -1709,18 +1712,20 @@ func (js *jetStream) monitorStream(mset *stream, sa *streamAssignment, sendSnaps
 	// For migration tracking.
 	var mmt *time.Ticker
 	var mmtc <-chan time.Time
+	var mDelayTc int
 
 	startMigrationMonitoring := func() {
 		if mmt == nil {
 			mmt = time.NewTicker(1 * time.Second)
 			mmtc = mmt.C
+			mDelayTc = 0
 		}
 	}
 
 	stopMigrationMonitoring := func() {
 		if mmt != nil {
 			mmt.Stop()
-			mmt, mmtc = nil, nil
+			mmt, mmtc, mDelayTc = nil, nil, 0
 		}
 	}
 	defer stopMigrationMonitoring()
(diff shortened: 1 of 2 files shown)


server/jetstream_cluster.go
```diff
@@ -7193,7 +7214,6 @@ func (mset *stream) runCatchup(sendSubject string, sreq *streamSyncRequest) {
 	// Setup sequences to walk through.
 	seq, last := sreq.FirstSeq, sreq.LastSeq
 	mset.setCatchupPeer(sreq.Peer, last-seq)
-	defer mset.clearCatchupPeer(sreq.Peer)
```
derekcollison, line 7196: What other ways can we exit without clearing? We ok with those?

server/jetstream_cluster.go
```diff
@@ -3861,15 +3875,21 @@ func (js *jetStream) monitorConsumer(o *consumer, ca *consumerAssignment) {
...
+				}
 				// Remove the old peers or transfer leadership (after which new leader resumes with peer removal).
 				// stopMigrationMonitoring is invoked on actual leadership change or
 				// on the next tick when migration completed.
 				// In case these operations fail, the next tick will retry
 				if !foundLeader {
+					s.Debugf("Scale down of '%s > %s' step down ('%s' preferred)", ca.Stream, ca.Name, firstPeer)
 					n.StepDown(firstPeer)
 				} else {
+					s.Noticef("Scale down of '%s > %s' to %+v", ca.Stream, ca.Name, newPeerSet)
```
derekcollison, line 3892: Should be notice? Every time we log?

derekcollison (approved):
LGTM - Few comments.

server/jetstream_cluster.go
matthiashanel: These are the only return statements after the removed line.
We do not clear, when there is no quit channel, quit channel or server quit channel gets closed.

We do not clear it when the other side aborts catchup and closes remoteQuitCh.
(log msg: Catchup for stream '%s > %s' was aborted on the remote due to: %q")
@derekcollison For this situation I'm leaning towards not clearing, unless you say this actually signals success. 

Basically we only clear when this side determines catchup is finished.
As a fallback to not clearing catchup, when a server looses leadership it clears all of them. 

```
	if qch == nil {
		return
	}
  ...
	for {
		select {
		case <-s.quitCh:
			return
		case <-qch:
			return
		case <-remoteQuitCh:
			return
		case <-notActive.C:
			s.Warnf("Catchup for stream '%s > %s' stalled", mset.account(), mset.name())
			return
		case <-nextBatchC:
			if !sendNextBatchAndContinue(qch) {
				mset.clearCatchupPeer(sreq.Peer)
				return
			}
		case <-cbKick:
			if !sendNextBatchAndContinue(qch) {
				mset.clearCatchupPeer(sreq.Peer)
				return
			}
		}
	}
}
```

server/jetstream_cluster.go
matthiashanel: I find this to be a rather significant event, worth of a log entry. But not set on this. @derekcollison WDYT?

The whole process starts with this notice:
	s.Noticef("Requested move of: R=%d stream '%s > %s' from old peer set %+v to new peer set %+v",
		cfg.Replicas, streamName, accName, s.peerSetToNames(currPeers), s.peerSetToNames(peers))

server/jetstream_cluster.go
derekcollison: Would need to look closer, but I would think maybe clear on that..

server/jetstream_cluster.go
derekcollison: ok good to bookend and keep consistent.

derekcollison (approved):
LGTM

server/jetstream_cluster.go
matthiashanel: fixed

derekcollison (approved):
LGTM

PR #3383, opened 2022-08-19
[CHANGED] Gateway: Switch all accounts to interest-only mode

kozlovic:
We are phasing out the optimistic-only mode. Servers accepting
inbound gateway connections will switch the accounts to interest-only
mode.

The servers with outbound gateway connection will check interest
and ignore the "optimistic" mode if it is known that the corresponding
inbound is going to switch the account to interest-only. This is
done using a boolean in the gateway INFO protocol.

### DIFF
--- a/server/gateway.go
+++ b/server/gateway.go
@@ -118,6 +118,14 @@ func (im GatewayInterestMode) String() string {
 	}
 }
 
+var gwDoNotForceInterestOnlyMode bool
+
+// GatewayDoNotForceInterestOnlyMode is used ONLY in tests.
+// DO NOT USE in normal code or if you embed the NATS Server.
+func GatewayDoNotForceInterestOnlyMode(doNotForce bool) {
+	gwDoNotForceInterestOnlyMode = doNotForce
+}
+
 type srvGateway struct {
 	totalQSubs int64 //total number of queue subs in all remote gateways (used with atomic operations)
 	sync.RWMutex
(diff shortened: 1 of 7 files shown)


server/gateway.go
```diff
@@ -118,6 +118,14 @@ func (im GatewayInterestMode) String() string {
 	}
 }
 
+var gwDoNotForceInterestOnlyMode bool
+
+// GatewayDoNotForceInterestOnlyMode is used ONLY in tests.
+// DO NOT USE in normal code or if you embed the NATS Server.
+func GatewayDoNotForceInterestOnlyMode(doNotForce bool) {
```
derekcollison, line 125: Should we make lower case to enforce this a bit more? Will work for tests under server but not test then..

server/gateway.go
```diff
@@ -1122,8 +1144,9 @@ func (c *client) processGatewayInfo(info *Info) {
 		js := s.js
 		s.mu.Unlock()
 
-		// Switch JetStream accounts to interest-only mode.
-		if js != nil {
+		// If running in some tests, maintain the original behavior.
+		if gwDoNotForceInterestOnlyMode && js != nil {
```
derekcollison, line 1148: Should we still do this if the server not running JS but has accounts with JS enabled with limits?

derekcollison (commented):
Lots of tests, sorry ;)

I wonder since 2.9 if we should consider sending accounts we would want to see interest in?

What I mean is for the server that is creating an outbound, it could send an A+ with an array of accounts that it has active interest on via a client connection or leafnode to the other side. Then the other side would only need to send actual subs for matching accounts.

I am ok with this being for 3.0 consideration as well.

server/gateway.go
kozlovic: That's for that reason I have it exported, I need it in the test/ package.

server/gateway.go
kozlovic: This is only for tests that want to maintain the old behavior. If not for that test variable (say the executable code compiled outside of test framework), the "else" (without even the check for the variable since it will be always false so compiled out) clause is executed, which is to switch ALL accounts.

derekcollison (approved):
LGTM

alexbozhenko:
@kozlovic Is it true that optimistic mode is not a thing anymore, and all code related to it can be reaped out?

kozlovic:
> Is it true that optimistic mode is not a thing anymore, and all code related to it can be reaped out?

@alexbozhenko It is true that since v2.9.0, servers at that same level will switch to interest-only mode. If you "reap out" the code related to optimistic mode, you will still have to handle connections from older servers (reject the connection for instance).

PR #3386, opened 2022-08-22
Some more checks for cc.meta == nil

kozlovic:
Missed those when re-running the previous test for longer period
of time.


derekcollison (approved):
LGTM

PR #3387, opened 2022-08-22
[ADDED] Monitoring: TLS Peer Certificates in Connz when auth is on

### LINKED ISSUE #3317: Include TLS client information in connz output
## Feature Request
When clients connect via TLS with client side certificate validation, no information about the client is available in the HTTP monitoring endpoints. This is what the data looks like right now:

image

#### Use Case:
Better insight into NATS server activity and its client connections.

#### Proposed Change:
Add public key, fingerprints, serial, subject to connz endpoint when query parameter `auth` is truthy.

kozlovic:
Add basic peer certificates information in /connz endpoint when
the "auth" option is provided.

```
{
  "server_id": "NDA6JPO2S27UL7QCA5OO7LP5JT324OMEYXTNEWDCSH2JRJYORSMW4HWD",
  "now": "2022-08-24T20:15:58.287276Z",
  "num_connections": 1,
  "total": 1,
  "offset": 0,
  "limit": 1024,
  "connections": [
    {
      "cid": 4,
      "kind": "Client",
      "type": "nats",
      "ip": "127.0.0.1",
      "port": 55361,
      "start": "2022-08-24T20:15:58.275853Z",
      "last_activity": "2022-08-24T20:15:58.284893Z",
      "rtt": "9.041ms",
      "uptime": "0s",
      "idle": "0s",
      "pending_bytes": 0,
      "in_msgs": 0,
      "out_msgs": 0,
      "in_bytes": 0,
      "out_bytes": 0,
      "subscriptions": 0,
      "lang": "go",
      "version": "1.16.0",
      "tls_version": "1.3",
      "tls_cipher_suite": "TLS_AES_128_GCM_SHA256",
      "tls_peer_certs": [
        {
          "subject": "CN=localhost,OU=nats.io,O=Synadia,ST=California,C=US",
          "spki_sha256": "bf6f821f09fde09451411ba3b42c0f74727d61a974c69fd3cf5257f39c75f0e9",
          "cert_sha256": "1316eeeb26856757afe1f4fc8b0621b24ca382c7bc73b285b0e282c714cf6935"
        }
      ]
    }
  ]
}
```

Resolves #3317

### DIFF
--- a/server/monitor.go
+++ b/server/monitor.go
@@ -14,7 +14,10 @@
 package server
 
 import (
+	"bytes"
+	"crypto/sha256"
 	"crypto/tls"
+	"crypto/x509"
 	"encoding/json"
 	"fmt"
 	"net"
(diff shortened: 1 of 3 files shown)


derekcollison (approved):
LGTM - Let's see what @philpennock has to say on format etc.

kozlovic:
@derekcollison @philpennock Sorry, forgot to add that the connz endpoint/curl would look like this:
```
{
  "server_id": "NA4VK47TEAD7D5MMI4H2EP6URWM6YSJYJF44P3PBRC5SLW5P3B7GDUKS",
  "now": "2022-08-22T17:46:01.041592Z",
  "num_connections": 1,
  "total": 1,
  "offset": 0,
  "limit": 1024,
  "connections": [
    {
      "cid": 4,
      "kind": "Client",
      "type": "nats",
      "ip": "127.0.0.1",
      "port": 59452,
      "start": "2022-08-22T17:46:01.030711Z",
      "last_activity": "2022-08-22T17:46:01.03961Z",
      "rtt": "8.9ms",
      "uptime": "0s",
      "idle": "0s",
      "pending_bytes": 0,
      "in_msgs": 0,
      "out_msgs": 0,
      "in_bytes": 0,
      "out_bytes": 0,
      "subscriptions": 0,
      "lang": "go",
      "version": "1.16.0",
      "tls_version": "1.3",
      "tls_cipher_suite": "TLS_AES_128_GCM_SHA256",
      "tls_peer_certs": [
        {
          "subject": "CN=localhost,OU=nats.io,O=Synadia,ST=California,C=US",
          "fprints": "13:16:EE:EB:26:85:67:57:AF:E1:F4:FC:8B:06:21:B2:4C:A3:82:C7:BC:73:B2:85:B0:E2:82:C7:14:CF:69:35"
        }
      ]
    }
  ]
}
```
See the last `tls_peer_certs` entry above.

derekcollison:
Would make fprints -> finger_prints

kozlovic:
@philpennock If you don't mind reviewing since we are getting close to v2.9.0 release. Thanks!

philpennock:
1. Do we always verify the peer cert by default?
2. Should we have something to warn when the peer is unverified?
3. An `issuer` field, much the same as `subject`, is helpful when confirming that transitions are happening as expected.
4. What is the purpose of the fingerprint here?

Note that the `pinned_certs` directive takes a fingerprint of the _public key_, not the _certificate_ (slightly misnamed directive), because this is more stable in the face of certificate reissuance.  This is a fairly common pattern and is the same thing seen in HPKP and TLSA records: the term "SPKI" means "Subject Public Key Information".

See the docs in <https://docs.nats.io/running-a-nats-service/configuration/securing_nats/tls>.
Fingerprint _changes_ over time as a term: the algorithm gets updated.

So I would rename `finger_prints` to `cert_sha256` (and singular, as it's one string) and add `spki_sha256` as a sibling field.

kozlovic:
> Do we always verify the peer cert by default?

No, if the server is not configured with tls_verify, then there won't be a client certificate.

> Should we have something to warn when the peer is unverified?

We could change - but that would be quite breaking -, but as of now it is up to the admin to configure the tls block to require client certificates.

> An issuer field, much the same as subject, is helpful when confirming that transitions are happening as expected.

I am not clear on that, could you please elaborate?

> What is the purpose of the fingerprint here?

From the OP: https://github.com/nats-io/nats-server/issues/3317#issuecomment-1202165750
```
Certificate subject and fingerprints would be a good start I think. Subject is typically human readable to quickly identify a certificate, the fingerprint is the truely unique identifier that can be used for, for example, database lookups.
```
And (https://github.com/nats-io/nats-server/issues/3317#issuecomment-1215391214)
```
Yes, the fingerprint (sometimes also called thumbprint) is the hash of the certificate. MD5 used to be common but nowadays sha256 is used, for example: ...
```

> So I would rename finger_prints to cert_sha256 (and singular, as it's one string) and add spki_sha256 as a sibling field.

So keep existing but rename as `cert_sha256` and add a new `spki_sha256` that would be the fingerprint of the public key? But would that be `RawSubjectPublicKeyInfo` in this: https://pkg.go.dev/crypto/x509#Certificate?

philpennock:
* Issuer: one of the obnoxious parts of preventing breaking with PKIX is when you can't tell if the certificate is issued by the new issuer, before the old one expires.  More generally, I have repeatedly found that having _both_ the subject and the issuer fields has been useful in various tasks, although not directly relevant here.  So, for me, `issuer:` is "I think it would be a very good idea, but it's not blocking"; but see the next item
* If the peer's certificate is not necessarily verified, then that's something which should be exposed.  But it could be at a higher level instead of per-cert, of just "we don't verify these".  If we are not verifying, then the `issuer` field should _not_ be included, because it becomes an unverified claim and potentially misleading.
* "So keep existing but rename as `cert_sha256` and add a new `spki_sha256` that would be the fingerprint of the public key?" -- correct
* "But would that be `RawSubjectPublicKeyInfo` [...]" -- yes, see Wally's code in `server/auth.go` function `matchesPinnedCert()`

philpennock (approved):

PR #3388, opened 2022-08-22
Reduce length of some clustering tests

kozlovic:
Since PR #3381, the 2 tests modified here would take twice as
long (around 245 seconds) to complete.
Talking with Matthias, he suggested using a variable instead of
a const and set it to 0 for those 2 tests since they don't really
need that to be set.


matthiashanel (approved):
LGTM

PR #3389, opened 2022-08-22
fix consumer subject validation on recovery

matthiashanel:
This fixes an issue introduced in #3080
The consumer filter subject check was skipped on recovery.

The intent was to bypass the upstream stream subjects.
But it also filtered the downstream stream subject.
This became a problem when the downstream was itself an upstream.

Then during recover, the stream subject was not checked, which
lead to delivery of filtered messages that should never have been
delivered.

-- Detail explanation --

he issue was was introduced by
```
commit f702e279ab4de56f907db10a563a64407fd64831
Author: Derek Collison <derek@nats.io>
Date:   Tue Apr 26 17:06:43 2022 -0700

    Fix for a consumer recovery issue.
    Also update healthz to check all assets that are assigned, not just running.

```

In `checkConsumerCfg` we basically skip filter subject validation by adding `&& !isRecovering` to the existing code
```
	// As best we can make sure the filtered subject is valid.
	if config.FilterSubject != _EMPTY_ && !isRecovering {
		subjects, hasExt := allSubjects(cfg, acc)
		if !validFilteredSubject(config.FilterSubject, subjects) && !hasExt {
			return NewJSConsumerFilterNotSubsetError()
		}
	}
```

I believe this was added to compensate for
```
// We had a scenario where a consumer would not recover properly on restart due to
// the cluster state not being set properly when checking source subjects.
func TestJetStreamSuperClusterStateOnRestartPreventsConsumerRecovery(t *testing.T) {
```

From the test I'd say the original issue was caused by deleting upstream
```
	// Pull source out from underneath the downstream stream.
	err = js.DeleteStream("SOURCE")
	require_NoError(t, err)
```

The original fix may be too broad here.
It intended to cover upstream filter subject validation.
But it also caused to skip the subject validation of the downstream.
This change always checks the downstream subject validation and skips upstream when in recovery mode

### DIFF
--- a/server/consumer.go
+++ b/server/consumer.go
@@ -446,9 +446,14 @@ func checkConsumerCfg(
 	}
 
 	// As best we can make sure the filtered subject is valid.
-	if config.FilterSubject != _EMPTY_ && !isRecovering {
-		subjects, hasExt := allSubjects(cfg, acc)
-		if !validFilteredSubject(config.FilterSubject, subjects) && !hasExt {
+	if config.FilterSubject != _EMPTY_ {
+		subjects := copyStrings(cfg.Subjects)
+		// explicitly skip validFilteredSubject when recovering
+		hasExt := isRecovering
+		if !isRecovering {
+			subjects, hasExt = gatherSourceMirrorSubjects(subjects, cfg, acc)
+		}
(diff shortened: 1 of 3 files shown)


derekcollison (approved):
LGTM! - Thanks!

kozlovic (approved):
LGTM

PR #3390, opened 2022-08-23
fix signing key template processing dropping allow

matthiashanel:
Scoped signing keys allow for optional values in allow rules
If an allow rule therefore gets removed because a tag is not present,
the removal needs to be compensated by adding in a deny >

### DIFF
--- a/server/auth.go
+++ b/server/auth.go
@@ -514,8 +514,8 @@ func processUserPermissionsTemplate(lim jwt.UserPermissionLimits, ujwt *jwt.User
 		return emittedList, nil
 	}
 
-	subAllowWasEmpty := len(lim.Permissions.Sub.Allow) > 0
-	pubAllowWasEmpty := len(lim.Permissions.Pub.Allow) > 0
+	subAllowWasNotEmpty := len(lim.Permissions.Sub.Allow) > 0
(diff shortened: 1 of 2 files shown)


server/auth.go
```diff
@@ -523,6 +527,14 @@ func processUserPermissionsTemplate(lim jwt.UserPermissionLimits, ujwt *jwt.User
 	} else if lim.Permissions.Pub.Deny, err = applyTemplate(lim.Permissions.Pub.Deny, true); err != nil {
 		return jwt.UserPermissionLimits{}, err
 	}
+
+	// if pub/sub allow where not empty, but are empty post template processing, add in a deny to compensate
```
kozlovic, line 531: `were` instead of `where`. But more importantly, looks like the code is opposite of the comment. That is, you wrote if they were not empty, but are empty after processing add deny. Then should it not be `if !subAllowWasEmpty && ..` (notice the "!"). Same for the pub.

kozlovic (changes requested):

server/auth.go
matthiashanel: wrong variable name. fixed

matthiashanel: thanks for pointing that out

kozlovic (approved):
LGTM

PR #3391, opened 2022-08-23
Add test that demonstrate cons pending < stream first seq is OK

kozlovic:


derekcollison (approved):
LGTM

PR #3392, opened 2022-08-23
General improvements to clustered streams during server restart and KV/CAS.

derekcollison:

/cc @nats-io/core

### DIFF
--- a/server/filestore.go
+++ b/server/filestore.go
@@ -4085,11 +4085,11 @@ func (fs *fileStore) loadLast(subj string, sm *StoreMsg) (lsm *StoreMsg, err err
 // The subject can be a wildcard.
 func (fs *fileStore) LoadLastMsg(subject string, smv *StoreMsg) (sm *StoreMsg, err error) {
 	if subject == _EMPTY_ || subject == fwcs {
-		sm, err = fs.msgForSeq(fs.lastSeq(), sm)
+		sm, err = fs.msgForSeq(fs.lastSeq(), smv)
 	} else {
-		sm, err = fs.loadLast(subject, sm)
+		sm, err = fs.loadLast(subject, smv)
 	}
-	if err != nil && err != ErrStoreClosed {
+	if sm == nil || (err != nil && err != ErrStoreClosed) {
 		err = ErrStoreMsgNotFound
 	}
 	return sm, err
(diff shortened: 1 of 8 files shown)


server/filestore.go
```diff
@@ -4080,16 +4083,16 @@ func (fs *fileStore) loadLast(subj string, sm *StoreMsg) (lsm *StoreMsg, err err
 
 // LoadLastMsg will return the last message we have that matches a given subject.
 // The subject can be a wildcard.
-func (fs *fileStore) LoadLastMsg(subject string, sm *StoreMsg) (*StoreMsg, error) {
+func (fs *fileStore) LoadLastMsg(subject string, smv *StoreMsg) (sm *StoreMsg, err error) {
```
kozlovic, line 4086: But changing `sm` to `smv` in function parameters, we are not using `smv` at all, so that seem wrong.

server/filestore.go
```diff
@@ -4080,16 +4083,16 @@ func (fs *fileStore) loadLast(subj string, sm *StoreMsg) (lsm *StoreMsg, err err
...
 	if subject == _EMPTY_ || subject == fwcs {
-		sm, _ = fs.msgForSeq(fs.lastSeq(), sm)
+		sm, err = fs.msgForSeq(fs.lastSeq(), sm)
 	} else {
-		sm, _ = fs.loadLast(subject, sm)
+		sm, err = fs.loadLast(subject, sm)
 	}
-	if sm == nil {
-		return nil, ErrStoreMsgNotFound
+	if err != nil && err != ErrStoreClosed {
```
kozlovic, line 4092: This is causing a panic on some client tests and that is because the low level fs.loadLast() may not find the message but not have an error, so `sm` will be nil. The original code as you can see above was checking for `sm` nil and in that case return `ErrStoreMsgNotFound`. We still need that or then make that test in fs.loadLast().

kozlovic (changes requested):

server/filestore.go
derekcollison: Good catch, will fix.

server/filestore.go
derekcollison: Wil lfix.

kozlovic (approved):
LGTM

kozlovic (approved):
LGTM

PR #3394, opened 2022-08-23
Fixed a bug that would not track per subject info for streams that were mirrors/sources.

derekcollison:
This was seen in internal testing by @ripienaar manifested as a consumer that would get stuck for a bit and skip records due to incorrect EOF signaling.

/cc @nats-io/core

### DIFF
--- a/server/filestore.go
+++ b/server/filestore.go
@@ -167,7 +167,7 @@ type msgBlock struct {
 	lchk    [8]byte
 	loading bool
 	flusher bool
-	noSubj  bool
+	noTrack bool
 	closed  bool
 }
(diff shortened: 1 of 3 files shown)


kozlovic (approved):
LGTM... But I am thinking: the change to not track fss is quite recent and for instance we will not do that for RAFT logs now, but what happens if a server is updated to v2.9.0 and recovers fss files from previous shutdown (when at say v2.8.4) but now the code does not track fss. Will that cause any sort of problem?

derekcollison:
They will be empty, so no effect. Thanks.

PR #3395, opened 2022-08-24
Scale down

matthiashanel:
scale down happened too soon

    when currentCount != replicas


davidkemper (approved):
Matthias has high confidence that this will resolve early exit from scaling for move.

PR #3400, opened 2022-08-25
CAS operations improved,

derekcollison:
We now hold stream locks through storage and use a separate lock for the consumer list and any storage updates from the store level.

/cc @nats-io/core

### DIFF
--- a/server/consumer.go
+++ b/server/consumer.go
@@ -4150,7 +4150,8 @@ func (o *consumer) decStreamPending(sseq uint64, subj string) {
 	// If it was pending process it like an ack.
 	// TODO(dlc) - we could do a term here instead with a reason to generate the advisory.
 	if wasPending {
-		o.processAckMsg(sseq, p.Sequence, rdc, false)
+		// We could have lock for stream so do this in a go routine.
+		go o.processTerm(sseq, p.Sequence, rdc)
 	}
 }
 
--- a/server/stream.go
+++ b/server/stream.go
@@ -219,6 +219,10 @@ type stream struct {
 	// For republishing.
 	tr *transform
 
+	// For processing consumers as a list without main stream lock.
+	clsMu sync.RWMutex
+	cList []*consumer
+
 	// TODO(dlc) - Hide everything below behind two pointers.
 	// Clustered mode.
 	sa       *streamAssignment
@@ -1534,23 +1538,17 @@ func (mset *stream) updateWithAdvisory(config *StreamConfig, sendAdvisory bool) 
 
 // Purge will remove all messages from the stream and underlying store based on the request.
 func (mset *stream) purge(preq *JSApiStreamPurgeRequest) (purged uint64, err error) {
-	mset.mu.Lock()
-	if mset.client == nil {
-		mset.mu.Unlock()
+	mset.mu.RLock()
+	if mset.client == nil || mset.store == nil {
+		mset.mu.RUnlock()
 		return 0, errors.New("invalid stream")
 	}
 	if mset.cfg.Sealed {
+		mset.mu.RUnlock()
 		return 0, errors.New("sealed stream")
 	}
-	var _obs [4]*consumer
-	obs := _obs[:0]
-	for _, o := range mset.consumers {
-		if preq != nil && !o.isFilteredMatch(preq.Subject) {
-			continue
-		}
-		obs = append(obs, o)
-	}
-	mset.mu.Unlock()
+	store := mset.store
+	mset.mu.RUnlock()
 
 	if preq != nil {
 		purged, err = mset.store.PurgeEx(preq.Subject, preq.Sequence, preq.Keep)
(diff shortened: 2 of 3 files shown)


server/stream.go
```diff
@@ -1534,23 +1538,17 @@ func (mset *stream) updateWithAdvisory(config *StreamConfig, sendAdvisory bool)
 
 // Purge will remove all messages from the stream and underlying store based on the request.
 func (mset *stream) purge(preq *JSApiStreamPurgeRequest) (purged uint64, err error) {
-	mset.mu.Lock()
-	if mset.client == nil {
+	mset.mu.RLock()
+	if mset.client == nil || mset.store == nil {
 		mset.mu.Unlock()
```
kozlovic, line 1543: Need to change to RUnlock()

server/stream.go
```diff
@@ -1534,23 +1538,17 @@ func (mset *stream) updateWithAdvisory(config *StreamConfig, sendAdvisory bool)
...
 		return 0, errors.New("invalid stream")
 	}
 	if mset.cfg.Sealed {
+		mset.mu.Unlock()
```
kozlovic, line 1547: Same

server/stream.go
```diff
@@ -3818,7 +3811,7 @@ func (mset *stream) processJetStreamMsg(subject, reply string, hdr, msg []byte,
 		} else if mset.numFilter > 0 {
 			// Assume no interest and check to disqualify.
 			noInterest = true
-			for _, o := range mset.consumers {
+			for _, o := range mset.cList {
```
kozlovic, line 3814: But even if you are under stream lock, you would need to protect with the new mutex to avoid races with other code that does not use stream lock but uses only the new lock

server/stream.go
```diff
@@ -3836,17 +3829,17 @@ func (mset *stream) processJetStreamMsg(subject, reply string, hdr, msg []byte,
...
 		mset.lseq = store.SkipMsg()
 		mset.lmsgId = msgId
+		// If we have a msgId make sure to save.
+		if msgId != _EMPTY_ {
+			mset.storeMsgIdLocked(&ddentry{msgId, seq, ts})
+		}
 		mset.mu.Unlock()
 
 		if canRespond {
 			response = append(pubAck, strconv.FormatUint(mset.lseq, 10)...)
```
kozlovic, line 3839: Should have captured mset.lseq above actually!

server/stream.go
```diff
@@ -4403,16 +4399,11 @@ func (mset *stream) getMsg(seq uint64) (*StoredMsg, error) {
...
 func (mset *stream) getConsumers() []*consumer {
-	mset.mu.RLock()
-	defer mset.mu.RUnlock()
-
-	var obs []*consumer
-	for _, o := range mset.consumers {
-		obs = append(obs, o)
-	}
-	return obs
+	mset.clsMu.Lock()
```
kozlovic, line 4404: Since this is getting the list, could be RLock. Do we need to check for cList being nil?

server/stream.go
```diff
@@ -4422,11 +4413,11 @@ func (mset *stream) numPublicConsumers() int {
 
 // This returns all consumers that are not DIRECT.
 func (mset *stream) getPublicConsumers() []*consumer {
-	mset.mu.RLock()
-	defer mset.mu.RUnlock()
+	mset.clsMu.Lock()
```
kozlovic, line 4416: RLock?

server/stream.go
```diff
@@ -4469,11 +4477,11 @@ func (mset *stream) lookupConsumer(name string) *consumer {
 }
 
 func (mset *stream) numDirectConsumers() (num int) {
-	mset.mu.RLock()
-	defer mset.mu.RUnlock()
+	mset.clsMu.Lock()
```
kozlovic, line 4480: RLock?

server/stream.go
```diff
@@ -4258,6 +4253,7 @@ func (mset *stream) stop(deleteFlag, advisory bool) error {
 	for _, o := range mset.consumers {
 		obs = append(obs, o)
 	}
+	mset.consumers, mset.cList = nil, nil
```
kozlovic, line 4256: Again, cList should be protected by new lock...

kozlovic (changes requested):

server/stream.go
derekcollison: Not following this one..

server/stream.go
derekcollison: range will work ok, will move to RLock.

server/stream.go
kozlovic: See the mset.lseq in FormatUint()? We should capture mset.lseq, say as lseq, under the mset lock above.

kozlovic (changes requested):

server/consumer.go
```diff
@@ -4150,7 +4150,8 @@ func (o *consumer) decStreamPending(sseq uint64, subj string) {
 	// If it was pending process it like an ack.
 	// TODO(dlc) - we could do a term here instead with a reason to generate the advisory.
 	if wasPending {
-		o.processAckMsg(sseq, p.Sequence, rdc, false)
+		// We could have lock for sytream so do this in a go routine.
```
kozlovic, line 4153: little typo: `systream`. But more importantly: is it going to be an issue if those "acks" are processed possibly out of order? Imagine a fast pace ingestion with low stream limits (even possibly KV use-case) that cause the "first" message to be removed, so those go routines may first: start to accumulate, second: processed "out of order".

Since you are now changing to a processTerm, then maybe the comment above "wasPending" can be revisited: also, it means that now an advisory is going to be generated for every message that is dropped. Is that ok?

Also, when invoked from storeUpdates(), we have something like:
```
...
    for _, o := range cl {
        o.decStreamPending(seq, subj)
    }
}
if mset.jsa != nil {
	mset.jsa.updateUsage(mset.tier, mset.stype, bd)
}
```
Previously, the ack processing was done in place, now it is not and so is that not possible that updateUsage() will have wrong information?

derekcollison: I think term is more appropriate here, that is what we are really doing.

The jsa update could be wrong, but quickly will be corrected.. IMO

kozlovic: Ok then

kozlovic (approved):
LGTM

PR #3401, opened 2022-08-25
Try to dump any cached state including fss on recovery to avoid memory bloat on restart.

derekcollison:

/cc @nats-io/core


kozlovic (approved):
LGTM

PR #3406, opened 2022-08-25
Don't fail if auto setting of maxprocs fails.

derekcollison:
Can fail in some embedded systems.

/cc @nats-io/core


kozlovic (approved):
LGTM

PR #3408, opened 2022-08-27
Fix testdata CA expiration

philpennock:
- Set cert lifetime on CA cert too
- Regenerate test cert data

/cc @nats-io/core


derekcollison (approved):
LGTM - Thanks.

PR #3409, opened 2022-08-27
[IMPROVED] Secure consumer create

derekcollison:
New consumer create that allows elevation of the stream and consumer names, and optional filter subject to the request subject.

Similar to changes in direct get allows proper security if needed for filter subject selection.

/cc @nats-io/core

### DIFF
--- a/server/consumer.go
+++ b/server/consumer.go
@@ -48,7 +48,9 @@ type ConsumerInfo struct {
 }
 
 type ConsumerConfig struct {
+	// Durable is deprecated. All consumers will have names. picked by clients.
 	Durable         string          `json:"durable_name,omitempty"`
+	Name            string          `json:"name,omitempty"`
 	Description     string          `json:"description,omitempty"`
 	DeliverPolicy   DeliverPolicy   `json:"deliver_policy"`
 	OptStartSeq     uint64          `json:"opt_start_seq,omitempty"`
@@ -353,10 +355,6 @@ func setConsumerConfigDefaults(config *ConsumerConfig, lim *JSLimitOpts, accLim 
 	}
 }
 
-func (mset *stream) addConsumer(config *ConsumerConfig) (*consumer, error) {
-	return mset.addConsumerWithAssignment(config, _EMPTY_, nil, false)
-}
-
 // Check the consumer config. If we are recovering don't check filter subjects.
 func checkConsumerCfg(
 	config *ConsumerConfig,
@@ -530,6 +528,10 @@ func checkConsumerCfg(
 	return nil
 }
 
+func (mset *stream) addConsumer(config *ConsumerConfig) (*consumer, error) {
+	return mset.addConsumerWithAssignment(config, _EMPTY_, nil, false)
+}
+
 func (mset *stream) addConsumerWithAssignment(config *ConsumerConfig, oname string, ca *consumerAssignment, isRecovering bool) (*consumer, error) {
 	mset.mu.RLock()
 	s, jsa, tierName, cfg, acc := mset.srv, mset.jsa, mset.tier, mset.cfg, mset.acc
@@ -589,8 +591,14 @@ func (mset *stream) addConsumerWithAssignment(config *ConsumerConfig, oname stri
 	}
 
 	// If this one is durable and already exists, we let that be ok as long as only updating what should be allowed.
+	var cName string
 	if isDurableConsumer(config) {
-		if eo, ok := mset.consumers[config.Durable]; ok {
+		cName = config.Durable
+	} else if config.Name != _EMPTY_ {
+		cName = config.Name
+	}
+	if cName != _EMPTY_ {
+		if eo, ok := mset.consumers[cName]; ok {
 			mset.mu.Unlock()
 			err := eo.updateConfig(config)
 			if err == nil {
@@ -674,11 +682,17 @@ func (mset *stream) addConsumerWithAssignment(config *ConsumerConfig, oname stri
 	} else if oname != _EMPTY_ {
 		o.name = oname
 	} else {
-		for {
-			o.name = createConsumerName()
-			if _, ok := mset.consumers[o.name]; !ok {
-				break
+		if config.Name != _EMPTY_ {
+			o.name = config.Name
+		} else {
+			// Legacy ephemeral auto-generated.
+			for {
+				o.name = createConsumerName()
+				if _, ok := mset.consumers[o.name]; !ok {
+					break
+				}
 			}
+			config.Name = o.name
 		}
 	}
 	// Create ackMsgs queue now that we have a consumer name
@@ -768,6 +782,8 @@ func (mset *stream) addConsumerWithAssignment(config *ConsumerConfig, oname stri
 		// Ephemerals will always have inactive thresholds.
 		// Add in 1 sec of jitter above and beyond the default of 5s.
 		o.dthresh = JsDeleteWaitTimeDefault + time.Duration(rand.Int63n(1000))*time.Millisecond
+		// Only stamp config with default sans jitter.
+		o.cfg.InactiveThreshold = JsDeleteWaitTimeDefault
 	}
 
 	if o.isPushMode() {
--- a/server/jetstream_api.go
+++ b/server/jetstream_api.go
@@ -131,10 +131,14 @@ const (
 	// jsDirectGetPre
 	jsDirectGetPre = "$JS.API.DIRECT.GET"
 
-	// JSApiConsumerCreate is the endpoint to create ephemeral consumers for streams.
+	// JSApiConsumerCreate is the endpoint to create consumers for streams.
+	// This was also the legacy endpoint for ephemeral consumers.
+	// It now can take consumer name and optional filter subject, which when part of the subject controls access.
 	// Will return JSON response.
-	JSApiConsumerCreate  = "$JS.API.CONSUMER.CREATE.*"
-	JSApiConsumerCreateT = "$JS.API.CONSUMER.CREATE.%s"
+	JSApiConsumerCreate    = "$JS.API.CONSUMER.CREATE.*"
+	JSApiConsumerCreateT   = "$JS.API.CONSUMER.CREATE.%s"
+	JSApiConsumerCreateEx  = "$JS.API.CONSUMER.CREATE.*.>"
+	JSApiConsumerCreateExT = "$JS.API.CONSUMER.CREATE.%s.%s.%s"
 
 	// JSApiDurableCreate is the endpoint to create durable consumers for streams.
 	// You need to include the stream and consumer name in the subject.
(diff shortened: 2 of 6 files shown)


server/consumer.go
```diff
@@ -589,8 +591,14 @@ func (mset *stream) addConsumerWithAssignment(config *ConsumerConfig, oname stri
 	}
 
 	// If this one is durable and already exists, we let that be ok as long as only updating what should be allowed.
+	var cName string
 	if isDurableConsumer(config) {
-		if eo, ok := mset.consumers[config.Durable]; ok {
+		cName = config.Durable
+	} else if config.Name != _EMPTY_ {
+		cName = config.Name
+	}
```
ripienaar, line 599: Maybe fail if both are set

server/consumer.go
```diff
@@ -48,7 +48,9 @@ type ConsumerInfo struct {
 }
 
 type ConsumerConfig struct {
+	// Durable is deprecated. All consumers will have names. picked by clients.
 	Durable         string          `json:"durable_name,omitempty"`
+	Name            string          `json:"name,omitempty"`
```
ripienaar, line 53: These kinds of depreations add ambiguity without much value, in effect in future all consuers will be durable.  So the existing key name is fine and accurate.  More so because its essentially an internal only detail hidden from users.

Not changing has the benefit of not adding weird changes or impacting systems that are based on remediation loops like terraform/k8s which would then see this as a change which will need remediation and would require quite careful specific handling in each of those systemsetc.  Or older client versions that does strict config compares etc. 

The cost/benefit of this change seems off when the existing name is correct already, even in future.

Etherway, this close to 2.9 this seems risky when lookin at the related changes overall vs the stated goal of this PR to me.

server/jetstream_api.go
```diff
@@ -131,10 +131,14 @@ const (
...
-	// JSApiConsumerCreate is the endpoint to create ephemeral consumers for streams.
+	// JSApiConsumerCreate is the endpoint to create consumers for streams.
+	// This was also the legacy endpoint for ephemeral consumers.
+	// It now can take consumer name and optional filter subject, which when part of the subject controls access.
 	// Will return JSON response.
-	JSApiConsumerCreate  = "$JS.API.CONSUMER.CREATE.*"
-	JSApiConsumerCreateT = "$JS.API.CONSUMER.CREATE.%s"
+	JSApiConsumerCreate    = "$JS.API.CONSUMER.CREATE.*"
+	JSApiConsumerCreateT   = "$JS.API.CONSUMER.CREATE.%s"
+	JSApiConsumerCreateEx  = "$JS.API.CONSUMER.CREATE.*.>"
```
ripienaar, line 140: This seems a bit limited/weird, if I wanted to filter on `x.>` then I would need to publish to `$JS.API....CREATE.NAME.x.>` as a subject, something that we do accept, though iirc not in all clients, but it seems very confusing to do this, also not sure how you would write an ACL to match the example destination subject?

I know there's an immediate need to solve this somehow but this doesnt seem like the best way.  

In the short term we really need prioritise the framework to allow arbitrary external policies infront of services so users can write something like OPA policies to achieve this rather than try to weirdly wedge every edge case into this format.  Though sure thats probably longer term than we can wait on this one. I am just weary of putting weird things in here that will not stand the test of time.

server/consumer.go
derekcollison: Both can be set imo. As we transition from Durable to Name.

server/consumer.go
derekcollison: Not all consumers will need to be durable, they will have a Name and optional InactiveThreshold, and Durable will be phased out.

This was needed for JS simplification as well.

server/jetstream_api.go
derekcollison: This should all be hidden by clients, and callouts will be the long term customization effort for sure, but for base level stuff elevating to a subject is our go to.

I am not sure I would want to require external components to be managed just to provide this security.

server/consumer.go
ripienaar: A durable wirh a timeout would survive restarts before the timeout? So it’s a durable. 

Chafing this key will complicate things like terraform and CRDs and doesn’t bring any functionality. It’s cosmetic at best no effect here but big knock on elsewhere.

derekcollison: An ephemeral that is backed by disk with an Inactive timeout would as well.

ripienaar: So then changing the key is just not needed. The behaviour is always durable with an optional idle timeout. 

Durable is a fine name and doesn’t break other things.

derekcollison: Durable meant no cleanup, and we had the notion of Name when you asked for consumer info which put together durable and ephemerals.

This is the right direction.

server/jetstream_api.go
aricart: Just to be clear, we are going to allow publishing to subjects containing wildcards?

aricart: Also durability is now a factor of configuration on the server for the consumer on `inactive_threshold`
An old client on a new server, will not supply the name - so it is durable with the default `inactive_threashold`. Test for durability on clients is effectively checking `inactive_threshold` to be set.

derekcollison: No, you can only publish to literals, but in this case we treat wildcards as literal tokens.

derekcollison: Old client will supply Durable name which is still supported. If they do not and only use legacy ephemeral API that is also still supported.

Meaning, unless you use the new style nothing changes.

aricart: so `$JS.API.CONSUMER.CREATE.streamName.consumerName.>` is not how I would create consumerName on streamName with access to all subjects?

derekcollison: That is the new style, not the old which again is still supported.

If you want new you can use above, but do not need the trailing ">". Just `$JS.API.CONSUMER.CREATE.streamName.consumerName`

The inactive threshold will determine what we used to define as "durable" vs "ephemeral"

server/consumer.go
aricart, line 53: Instead of adding a new property to check, why not simply just say all consumers are durable. And that durables can be reaped by the server based on their `inactive_threshold`. Server can still autoname "ephemerals" when name is empty.

This would simplify the logic on clients, at the expense that `durable_name` would always be set, and `name` in ConsumerInfo would duplicate it.

aricart: The only other concern then becomes how to reflect the new use of the API. If the consumer specifies a filter in the config, and the server matches the version, the client could simply use the new API. Otherwise, it can use the old API, allowing security compliance of the `$JS.API.CREATE.CONSUMER...` API without having to deal with model changes.

derekcollison: Durables could not be reaped, so I prefer to not change that behavior and leave as is for existing clients.

Going forward all consumers will have a name and optional inactive threshold.
Note also that consumer info already had Name, so it now simply matches that for new style.

aricart:
So this means that if you want to create a durable, the client specifies `durable_name`, otherwise `name`, or empty `name` and `durable_name` for an ephemeral

derekcollison:
> So this means that if you want to create a durable, the client specifies `durable_name`, otherwise `name`, or empty `name` and `durable_name` for an ephemeral

You can also create a named ephemeral by specifying name and inactive threshold.

server/jetstream_test.go
```diff
@@ -19284,3 +19284,132 @@ func TestJetStreamConsumerEOFBugNewFileStore(t *testing.T) {
...
+	require_True(t, si.State.Msgs == 300)
+
+	// Trying to create a consumer with non filtered API should fail.
+	js, err = nc.JetStream(nats.MaxWait(200 * time.Millisecond))
+	require_NoError(t, err)
+
+	_, err = js.SubscribeSync("ID.foo.*")
+	require_Error(t, err, nats.ErrTimeout)
+
+	js, err = nc.JetStream(nats.MaxWait(200 * time.Millisecond))
```
kozlovic, line 19347: Curious as to why you need to recreate the JS context here? (since it was created few lines above).

kozlovic (approved):
LGTM (a comment in the test). I have verified that with this branch the nats.go and nats.c client libraries test suite pass.

aricart (approved):
LGTM

ripienaar:
So feature wise we have still Durable and Ephemeral. 

Both will survive reboots
Both have a idle timeout
Both have a name
Both supports the user giving a name

Is it then that the only difference is essentially just positioning - a durable without a timeout is ephemeral? 

Then I guess implementation wise everything is a durable, the server only knows about durable consumers.  If a client doesnt set `durable_name` or `inactive_timeout` we supply them in line with ephemeral behaviors of today. The result is a durable with a timeout and a generated name.

You can additionally have R1 durables, memory only durables and so forth.  But all are durables.

Keeping the structures as is seem acceptible to me, no client changes needed to create an ephemeral or durable - the API simplification can use the features as it wish any old code continue to work?

So if above is not correct, what is the technical need for adding/deprecating around names, what can not be implemented without deprecating durable name? 

These kind of API key changes, not on a major version boundary is really disruptive, so much so that I am thinking of just abandoning the terraform provider if the server simply cant keep stable interfaces since those systems require that.

server/consumer.go
```diff
@@ -48,7 +48,9 @@ type ConsumerInfo struct {
 }
 
 type ConsumerConfig struct {
+	// Durable is deprecated. All consumers will have names. picked by clients.
```
aricart, line 51: this is still deprecated...

server/consumer.go
```diff
@@ -768,6 +782,8 @@ func (mset *stream) addConsumerWithAssignment(config *ConsumerConfig, oname stri
 		// Ephemerals will always have inactive thresholds.
 		// Add in 1 sec of jitter above and beyond the default of 5s.
 		o.dthresh = JsDeleteWaitTimeDefault + time.Duration(rand.Int63n(1000))*time.Millisecond
+		// Only stamp config with default sans jitter.
+		o.cfg.InactiveThreshold = JsDeleteWaitTimeDefault
```
kozlovic, line 786: Discussing with Alberto, he has a point that if Durable is now deprecated, if the app were to specify only Name and no inactive threshold, I believe that the code will still fall into that "if" statement and make it an ephemeral, no?

kozlovic (changes requested):

server/jetstream_test.go
derekcollison: Probably don't since it is done above, will remove.

derekcollison: The subscribe is supposed to fail, and would by default take 10s, hence why I recreate but could also just do for that one call I think..

server/consumer.go
derekcollison: Correct, you would have to fill in Durable or provide a filter subject.

Durable is not deprecated per se, but previously it was very specific in semantics that it would not go away until removed.

Moving forward we will support both legacy models, but new way is all have a Name and an optional Inactive Threshold and R factor.

kozlovic (approved):
LGTM

server/consumer.go
derekcollison: We could soften if you like..

server/consumer.go
ripienaar, line 53: Why `omitempty` if `All consumers will have names`?

derekcollison: It's good hygiene for older consumers and clients since it does not introduce this on the wire in those situations.

PR #3411, opened 2022-08-29
[FIXED] Stream info numbers may be 0 after cluster restart

kozlovic:
This would happen after multiple replicas changes.


derekcollison:
Fixed the issue with account purge test, also rebased to main.

derekcollison (approved):
LGTM

PR #3412, opened 2022-08-29
[FIXED] Gateway: system request/replies may not work properly

kozlovic:
When a subscription is recently made, gateway code ensures that if
there is a reply subject, the reply is "mapped" or rewritten to allow
the reply to come back to the origin cluster, regardless of subscription
interest propagation.

The issue was that this uses a map with a `*client` as the key
but the pointer for SYSTEM clients would not always be the same,
which meant that the rewrite would not happen, causing possible
"loss" of replies.


derekcollison (approved):
LGTM

PR #3413, opened 2022-08-30
FSS state could skew on expire on recover with no msgs left.

derekcollison:
Also added in sanity check on server start.

/cc @nats-io/core


davidkemper (approved):
LGTM on green unit tests.

kozlovic (commented):
LGTM

PR #3416, opened 2022-08-30
When writing a msg after the fss state was expired we would count the msg twice.

derekcollison:
This was due to the fact that we updated the accounting and then checked if we needed to load state which would then fallback to generating the state on the fly.

/cc @nats-io/core


ripienaar (approved):
LGTM, cant repro locally, will test upgrades of existing streams once merged

kozlovic (commented):
LGTM

PR #3418, opened 2022-08-30
[ADDED] JetStream: Ability to configure the per server max catchup bytes

kozlovic:
The original value was hardcoded to 128MB and 32MB per stream. The
per-server limit is lowered to 32MB but is configurable with
a new configuration parameter:
```
jetstream {
   max_catchup: 8MB
}
```

The per-stream limit was also lowered from 32MB/128,000msgs to
8MB/32,000 messages.

Tests have shown no difference in performance for fast links.

### DIFF
--- a/server/opts.go
+++ b/server/opts.go
@@ -255,6 +255,7 @@ type Options struct {
 	JetStreamCipher       StoreCipher   `json:"-"`
 	JetStreamUniqueTag    string
 	JetStreamLimits       JSLimitOpts
+	JetStreamMaxCatchup   int64
 	StoreDir              string            `json:"-"`
 	JsAccDefaultDomain    map[string]string `json:"-"` // account to domain name mapping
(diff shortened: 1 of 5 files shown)


server/opts.go
```diff
@@ -1948,6 +1949,12 @@ func parseJetStream(v interface{}, opts *Options, errors *[]error, warnings *[]e
 				}
 			case "unique_tag":
 				opts.JetStreamUniqueTag = strings.ToLower(strings.TrimSpace(mv.(string)))
+			case "max_catchup", "max_catchup_bytes", "max_total_catchup_bytes":
```
derekcollison, line 1952: Maybe max_oustanding_catchup?

derekcollison (approved):
LGTM - minor naming suggestion.

derekcollison (approved):
LGTM

PR #3419, opened 2022-08-30
Allow optional compression in raft and stream catchup, move improvements for streams and consumers

derekcollison:
We will not optionally compress raft traffic and stream catchup messages iff all peers support it. Current threshold (fixed atm) is 4k for review.

Improved stream and consumer moves.

/cc @nats-io/core


kozlovic (approved):
LGTM

PR #3423, opened 2022-08-31
[FIXED] Subjects collide would not properly return false.

### LINKED ISSUE #3422: SubjectsCollide() returns true for non matching subjects
## Defect

`SubjectsCollide()` will match subjects which do not match same literal subject (e.g. `foo.*` and `foo.*.bar.>`). That issue impacts  creating consumers with `FilterSubject`, as well as listing streams with subject filter (and probably more).

I added a simple test to verify the issue:
```go
func TestSubectsCollide(t *testing.T) {
	s1 := "foo.*.bar.>"
	s2 := "foo.*"

	res := SubjectsCollide(s1, s2)

        // prints true
	fmt.Println(res)
}
```

derekcollison:
When a subject with only partial wildcards was compared against a subject with a fwc but more tokens would incorrectly return true. e.g. `foo.*` vs `foo.bar.>`

Resolves #3422 	

/cc @nats-io/core

### DIFF
--- a/server/sublist.go
+++ b/server/sublist.go
@@ -1215,10 +1215,18 @@ func SubjectsCollide(subj1, subj2 string) bool {
 	if !fwc1 && !fwc2 && len(toks1) != len(toks2) {
 		return false
 	}
+	if lt1, lt2 := len(toks1), len(toks2); lt1 != lt2 {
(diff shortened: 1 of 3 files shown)


kozlovic (approved):
LGTM

PR #3426, opened 2022-08-31
Test for PublishAsyncComplete

mprimi:
 - [x] Tests added

Related to https://github.com/nats-io/nats.go/issues/1064

 - Add a test for `<-js.PublishAsyncComplete()` verifying all PubAckFutures are in fact completed

At the moment, this test occasionally fails (~3-5% of runs)

/cc @nats-io/core

### DIFF
--- a/server/jetstream_test.go
+++ b/server/jetstream_test.go
@@ -4576,6 +4576,41 @@ func TestJetStreamSnapshotsAPI(t *testing.T) {
 	}
 }
 
+func TestJetStreamPubAckFutureComplete(t *testing.T) {
+	s := RunBasicJetStreamServer()
+	if config := s.JetStreamConfig(); config != nil {
+		defer removeDir(t, config.StoreDir)
+	}
+	defer s.Shutdown()
+
+	nc, js := jsClientConnect(t, s)
+	defer nc.Close()
+
+	if _, err := js.AddStream(&nats.StreamConfig{Name: "foo", Storage: nats.MemoryStorage}); err != nil {
+		t.Fatalf("Unexpected error: %v", err)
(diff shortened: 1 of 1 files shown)


server/jetstream_test.go
```diff
@@ -4576,6 +4576,41 @@ func TestJetStreamSnapshotsAPI(t *testing.T) {
...
+	defer s.Shutdown()
+
+	nc, js := jsClientConnect(t, s)
+	defer nc.Close()
+
+	if _, err := js.AddStream(&nats.StreamConfig{Name: "foo", Storage: nats.MemoryStorage}); err != nil {
+		t.Fatalf("Unexpected error: %v", err)
+	}
+
+	toSend := 1_000_000
```
kozlovic, line 4593: Sending that many messages when running under `-race` is not good. So you should move it to the `norace_test.go`, but more importantly, is this a test to show that there is a server issue or a client issue? If client (since I saw that you opened an issue in the nats.go repo), then I would not add it this in server test.

kozlovic (changes requested):

server/jetstream_test.go
mprimi: I put this here because I clone-updated an existing test in this file.
This test fails now, so don't want to merge this. We can close it. It was just a way to share test code with you.

I don't know if this is a client or server issue. I'm just reporting a problem I noticed.

kozlovic:
@mprimi Sorry for the confusion, I thought that you determined that it was an issue in the client since you opened the issue there. Will check with above test when I have some time and will make the determination if this is a server or client issue. Thanks!

PR #3427, opened 2022-08-31
[FIXED] JetStream: possible routing issues through gateways

kozlovic:
Internally jetstream may subscribe to some subject and then send
a request with a reply subject matching that subscription.
Due to interest propagation through a super cluster, it is possible
that the reply comes back to a node that is not yet aware of
the subscription interest which would cause the reply to be dropped.

Some code detects that the subscription is recent and "map" the
reply subject so that it can be routed back to the origin server.
However, this was done with the use of the connection object that
created the subscription, but at the time of the send, a different
internal "*client" object may be used which would then cause
the code to not be aware of the recent subscription and not do
the mapping.

This code was changed to scope at the account level instead of
connection.

A recent change in PR #3412 is no longer needed and was reverted
in favor of changes in this PR.

### DIFF
--- a/server/gateway.go
+++ b/server/gateway.go
@@ -162,15 +162,8 @@ type srvGateway struct {
 		m map[string]map[string]*sitally
 	}
 
-	// This is to track recent subscriptions for a given connection
+	// This is to track recent subscriptions for a given account
 	rsubs sync.Map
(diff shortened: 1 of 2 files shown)


derekcollison (approved):
LGTM

PR #3429, opened 2022-08-31
[FIXED] Do not allow overlap between multiple subjects for a stream.

derekcollison:
We detected overlap between different streams for subjects but not same stream with multiple subjects configured.

/cc @nats-io/core

### DIFF
--- a/server/stream.go
+++ b/server/stream.go
@@ -411,19 +411,6 @@ func (a *Account) addStreamWithAssignment(config *StreamConfig, fsConfig *FileSt
 		return nil, NewJSStreamSubjectOverlapError()
 	}
 
-	// Now check if we have multiple subjects they we do not overlap ourselves
-	// which would cause duplicate entries (assuming no MsgID).
-	if len(cfg.Subjects) > 1 {
-		for _, subj := range cfg.Subjects {
-			for _, tsubj := range cfg.Subjects {
-				if tsubj != subj && SubjectsCollide(tsubj, subj) {
-					jsa.mu.Unlock()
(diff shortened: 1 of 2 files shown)


server/stream.go
```diff
@@ -404,12 +404,26 @@ func (a *Account) addStreamWithAssignment(config *StreamConfig, fsConfig *FileSt
...
 	}
 
+	// Now check if we have multiple subjects they we do not overlap ourselves
+	// which would cause duplicate entries (assuming no MsgID).
+	if len(cfg.Subjects) > 1 {
+		for _, subj := range cfg.Subjects {
+			for _, tsubj := range cfg.Subjects {
+				if tsubj != subj && SubjectsCollide(tsubj, subj) {
+					jsa.mu.Unlock()
+					return nil, fmt.Errorf("subject %q overlaps with %q", subj, tsubj)
```
kozlovic, line 421: 2 things:
- is there a formal NewJSxx error already for that or not?
- more importantly, if a stream already exists like that, does it mean that it will not be recovered? does that leave the system in a precarious condition?

I would recommend make a manual test and see how it behaves on an upgrade to make sure it is not too bad.

derekcollison: Good point, yes if it was created it would not be able to restore.

No error for that one, we have one for colliding with another stream's subjects.

Will also add test for update stream, etc.

derekcollison: ok added test for update, will do manual test on recover now.

derekcollison: I think it is ok. Can always use old server to restore and snapshot and edit subjects.

image

kozlovic (approved):
LGTM

PR #3432, opened 2022-09-01
Replace verbose API error detail with concise message

tbeets:
Re-placement (move) errors due to no match generated extreme-verbosity API error detail.  This PR opts for brief API error detail instead.  The mechanisms that collect verbose detail left in place if/when needed in a custom build.  

Simplification made in two places as there are two implementations of move -- one as a side effect of a stream update and the other a direct call to new move API.

/cc @nats-io/core

### DIFF
--- a/server/jetstream_api.go
+++ b/server/jetstream_api.go
@@ -2380,7 +2380,8 @@ func (s *Server) jsLeaderServerStreamMoveRequest(sub *subscription, c *client, _
 			errs = append(errs, e)
 		}
 		if peers == nil {
-			resp.Error = NewJSClusterNoPeersError(&errs)
+			// per-cluster errs helpful for development debug, but concise message sufficient for API return
+			resp.Error = NewJSClusterNoPeersError(fmt.Errorf("no clusters match placement criteria for move"))
 			s.sendAPIErrResponse(ci, acc, subject, reply, string(msg), s.jsonResponse(&resp))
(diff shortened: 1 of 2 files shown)


derekcollison:
What does the output from a NATS cli command look like now? Possible to paste example/screenshot here?

derekcollison (commented):
We want more detail than this IMO. Why could it not be placed, but a summary, not the level of detail currently being sent.

kozlovic (changes requested):
A lot of information was added to help understand why the placement failed. To be honest, I may not understand the original output myself, but completely getting rid of it may not be the right approach. We should consult with Matthias that added it, likely for a reason.

Regardless, the current test suite fails because there is a test checking for the error report.

At the very least, I wonder if we should add the special error in the server log?

tbeets:
Agree @kozlovic .  It is not obvious how to get to a "Goldilocks" message verbosity here since the factors for non-placement are several, could be multiple, and could be different for every single cluster.  That is a lot of information.  Also, as Travis caught, @matthiashanel assumes the full verbosity in validating functional tests.

I recommend we ship 2.9.0 GA as is then dialogue about approach.  From typical usage perspective (at CLI) could be better to ship detailed information from server and then allow the CLI to have verbosity option to shape the terminal output.  Need to factor in how much cluster detail to share with account owners.

PR #3433, opened 2022-09-01
[FIXED] JetStream: Some nodes may never be reported as offline

kozlovic:
In some rare situations, it is possible that nodes are added
to the cluster but are not properly tracked and not shown as
offline when they exit the cluster.

Relates to #3258


derekcollison (approved):
LGTM

PR #3434, opened 2022-09-03
[FIXED] Parsing of strings starting with numbers and K/G/etc.. suffix

### LINKED ISSUE #3431: String envar incorrectly inferred as number
## Defect

A String envar that starts with a digit (0-9) followed by the second character being anyone of these...
```
return r == 'k' || r == 'K' || r == 'm' || r == 'M' || r == 'g' || r == 'G' || r == 't' || r == 'T' || r == 'p' || r == 'P' || r == 'e' || r == 'E'
```
and has a length > 2, is incorrectly inferred as a number and not a string.

After stepping through the code [isNumberSuffix()](https://github.com/nats-io/nats-server/blob/c5c8e385abffcc4f0b6911a9ec25f55d2ffcca0d/conf/lex.go#L1141) gets called but I feel like it should only infer it's a number if the suffix is at the end of the string.

 - [x] Included a [Minimal, Complete, and Verifiable example] 

Here is a unit test that shows the issue. It can go in `conf/parse_test.go`

```go
func TestEnvVariableString(t *testing.T) {
	ex := map[string]interface{}{
		"foo": "2GP3JZCZ4G7JNWkzBhVYx9ZXtUiu",
	}
	evar := "__UNIQ22__"
	os.Setenv(evar, "2GP3JZCZ4G7JNWkzBhVYx9ZXtUiu")
	defer os.Unsetenv(evar)
	test(t, fmt.Sprintf("foo = $%s", evar), ex)
}
```

#### Expected result:
The test passes

#### Actual result:
Errors with 
```
Received err: variable reference for '__UNIQ22__' on line 1 could not be parsed: Parse error on line 1: 'Expected a top-level value to end with a new line, comment or EOF, but got 'P' instead.'
```

kozlovic:
If a configuration variable starts with numbers and has a character
that such as K/k/G/g/etc.. it would assume that it was a number
(expressed in Kb, Gb, etc..).

This PR checks that if the special characters are not the suffix,
that is, the variable does not end after those characters, then
the parsing will treat the whole thing as a string.

Resolves #3431

### DIFF
--- a/conf/lex.go
+++ b/conf/lex.go
@@ -1009,9 +1009,19 @@ func lexConvenientNumber(lx *lexer) stateFn {
 	case r == 'b' || r == 'B' || r == 'i' || r == 'I':
 		return lexConvenientNumber
 	}
-	lx.backup()
-	lx.emit(itemInteger)
-	return lx.pop()
+	if isNL(r) || r == eof || r == mapEnd || r == optValTerm || r == mapValTerm || isWhitespace(r) || unicode.IsDigit(r) {
+		lx.backup()
+		lx.emit(itemInteger)
+		return lx.pop()
+	}
+	// This is not a number, so we have to backup to the start and consider
+	// this to be a string
+	pos, start := lx.pos, lx.start
+	for i := pos; i > start; i-- {
+		lx.backup()
+	}
+	lx.stringStateFn = lexString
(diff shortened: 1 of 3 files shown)


derekcollison (approved):
LGTM

PR #3435, opened 2022-09-03
js: add per account reserved mem/store bytes

wallyqs:
This makes the `reserved_memory` and `reserved_storage` quotas for an account to be filled. For example given the following configuration:

```hcl
accounts {
  sys {
    users = [{ user: 'sys', pass: 'sys' }]
  }
  a {
    users = [{ user: 'foo', pass: 'foo' }]
    jetstream {
      max_mem: 500mi
      max_store: 1gi
    }
  }
```

Now making a request to `http://127.0.0.1:8222/jsz?accounts=true` it would make the reserved_memory and reserved_storage show up instead of 0s:

```js
"account_details": [
    {
      "name": "a",
      "id": "a",
      "memory": 0,
      "storage": 28956,
      "reserved_memory": 524288000,
      "reserved_storage": 1073741824,
      "accounts": 0,
      "ha_assets": 0,
      "api": {
        "total": 0,
        "errors": 0
      }
    }
  ]
```

Also fixes `$SYS.REQ.ACCOUNT.*.JSZ`:

```
nats req -s 'nats://sys:sys@127.0.0.1:4222' '$SYS.REQ.ACCOUNT.a.JSZ' '{"streams": true, "consumers":true}'
09:21:33 Sending request on "$SYS.REQ.ACCOUNT.a.JSZ"
09:21:33 Received with rtt 228µs
{"server":{"name":"NBAUUNWTIZHQZ6K5TGZJWO6PJ2LQSMKT3FR2BQWBMDYDUXC37SWPVAJ4","host":"0.0.0.0","id":"NBAUUNWTIZHQZ6K5TGZJWO6PJ2LQSMKT3FR2BQWBMDYDUXC37SWPVAJ4","ver":"2.9.0-RC.16","seq":128,"jetstream":true,"time":"2022-09-03T16:21:33.1023Z"},"data":{"name":"a","id":"a","memory":0,"storage":28956,"reserved_memory":524288000,"reserved_storage":1073741824,"accounts":0,"ha_assets":0,"api":{"total":0,"errors":0},"stream_detail":[{"name":"foo","cluster":{"leader":"NBAUUNWTIZHQZ6K5TGZJWO6PJ2LQSMKT3FR2BQWBMDYDUXC37SWPVAJ4"},"state":{"messages":762,"bytes":28956,"first_seq":1,"first_ts":"2022-09-02T22:00:17.334501Z","last_seq":762,"last_ts":"2022-09-03T16:12:31.677082Z","num_subjects":1,"consumer_count":0}}]}}
```

/cc @nats-io/core

### DIFF
--- a/server/monitor.go
+++ b/server/monitor.go
@@ -2700,6 +2700,11 @@ func (s *Server) accountDetail(jsa *jsAccount, optStreams, optConsumers, optCfg 
 		},
 		Streams: make([]StreamDetail, 0, len(jsa.streams)),
 	}
+	if reserved, ok := jsa.limits[_EMPTY_]; ok {
+		detail.JetStreamStats.ReservedMemory = uint64(reserved.MaxMemory)
+		detail.JetStreamStats.ReservedStore = uint64(reserved.MaxStore)
+	}
+
 	jsa.usageMu.RUnlock()
 	var streams []*stream
 	if optStreams {
--- a/server/monitor_test.go
+++ b/server/monitor_test.go
@@ -4205,6 +4205,41 @@ func TestMonitorJsz(t *testing.T) {
 			}
 		}
 	})
+	t.Run("accounts reserved metrics", func(t *testing.T) {
+		for _, url := range []string{monUrl1, monUrl2} {
+			info := readJsInfo(url + "?accounts=true&acc=ACC")
+			if len(info.AccountDetails) != 1 {
+				t.Fatalf("expected single account")
+			}
+			acc := info.AccountDetails[0]
+			got := int(acc.ReservedMemory)
+			expected := 5242880
+			if got != expected {
+				t.Errorf("Expected: %v, got: %v", expected, got)
+			}
+			got = int(acc.ReservedStore)
+			expected = 4194304
+			if got != expected {
+				t.Errorf("Expected: %v, got: %v", expected, got)
+			}
+
+			info = readJsInfo(url + "?accounts=true&acc=BCC_TO_HAVE_ONE_EXTRA")
+			if len(info.AccountDetails) != 1 {
+				t.Fatalf("expected single account")
+			}
+			acc = info.AccountDetails[0]
+			got = int(acc.ReservedMemory)
+			expected = -1
+			if got != expected {
+				t.Errorf("Expected: %v, got: %v", expected, got)
+			}
+			got = int(acc.ReservedStore)
+			expected = -1
+			if got != expected {
+				t.Errorf("Expected: %v, got: %v", expected, got)
+			}
+		}
+	})
 	t.Run("offset-too-big", func(t *testing.T) {
 		for _, url := range []string{monUrl1, monUrl2} {
 			info := readJsInfo(url + "?accounts=true&offset=10")


server/monitor.go
```diff
@@ -2709,6 +2709,15 @@ func (s *Server) accountDetail(jsa *jsAccount, optStreams, optConsumers, optCfg
 	}
 	jsa.mu.RUnlock()
 
+	accStats := s.js.lookupAccount(acc)
+	if accStats != nil && accStats.limits != nil {
+		limits, ok := accStats.limits[_EMPTY_]
```
derekcollison, line 2714: Can make this one line.. if limits,ok ..

server/monitor.go
```diff
@@ -2709,6 +2709,15 @@ func (s *Server) accountDetail(jsa *jsAccount, optStreams, optConsumers, optCfg
 	}
 	jsa.mu.RUnlock()
 
+	accStats := s.js.lookupAccount(acc)
```
kozlovic, line 2712: I am not familiar enough with "reserved" values to meaningfully comment, but we have a `jsa` object already and you are now looking up the `jsAccount` for the account of the `jsa` that you were given? They should be the same! (I know that there is a bug in the server that could lead to a race where a given registered account's jsAccount is not the one stored in `js.accounts[acc.Name]`, but not sure you were facing this issue).

So the limits that you check here should be done above for the given `jsa` and under the jsa.usageMu's lock.

kozlovic (changes requested):

server/monitor.go
wallyqs: Thanks, will take a look now and add tests.

wallyqs: Thanks, ok updated now.

server/monitor.go
```diff
@@ -2709,6 +2710,13 @@ func (s *Server) accountDetail(jsa *jsAccount, optStreams, optConsumers, optCfg
 	}
 	jsa.mu.RUnlock()
 
+	if limits != nil {
```
kozlovic, line 2713: But why doing all that outside of the lock, that was my point? Sure you capture `limits` under the lock but then the lookup is done outside of the lock which can still produce races. I would do all that under the jsa.usageMu's lock at line 2703.

kozlovic (changes requested):

server/monitor.go
wallyqs: yeah that makes more sense. thanks, just fixed in the next commit.

server/monitor.go
```diff
@@ -2701,6 +2701,13 @@ func (s *Server) accountDetail(jsa *jsAccount, optStreams, optConsumers, optCfg
 		Streams: make([]StreamDetail, 0, len(jsa.streams)),
 	}
 	limits := jsa.limits
```
kozlovic, line 2703: As Derek mentioned previous, this could be done in one line, such as: `if limits := jsa.limits; limits != nil {`, but knowing that lookup of a nil map is valid, you could even replace this with:
```
if reserved, ok := jsa.limits[_EMPTY_]; ok {
 ...
```
However, I think that there may be case (if not always) that by default we store `defaultJSAccountTiers` for the `_EMPTY_` key, which means that MaxMemory/MaxStore are set to -1, and you take uint64() which will then translate to big number, which is likely ok, just wanted to bring this to your attention.

kozlovic (changes requested):
Still a little code simplification I think.

server/monitor.go
wallyqs: updated to add test that confirms that the other account without reserved quota is actually getting the `-1` limit values.

server/monitor_test.go
```diff
@@ -4212,13 +4212,29 @@ func TestMonitorJsz(t *testing.T) {
 				t.Fatalf("expected single account")
 			}
 			acc := info.AccountDetails[0]
-			got := acc.ReservedMemory
-			expected := uint64(5242880)
```
kozlovic: I would have kept expected as uint64() since after all acc.ReservedMemory is a uint64.

server/monitor_test.go
```diff
@@ -4212,13 +4212,29 @@ func TestMonitorJsz(t *testing.T) {
...
+			if got != expected {
+				t.Errorf("Expected: %v, got: %v", expected, got)
+			}
+
+			info = readJsInfo(url + "?accounts=true&acc=BCC_TO_HAVE_ONE_EXTRA")
+			if len(info.AccountDetails) != 1 {
+				t.Fatalf("expected single account")
+			}
+			acc = info.AccountDetails[0]
+			got = int(acc.ReservedMemory)
```
kozlovic, line 4231: Again, would keep got as acc.ReservedMemory, but instead have `expected = uint64(-1)`.

kozlovic (changes requested):

server/monitor_test.go
kozlovic: Compiler does not let you do uint64(-1), so disregard.

kozlovic (approved):
LGTM

PR #3438, opened 2022-09-06
Use testing.TempDir() where possible

mprimi:
Rather than relying on manual deferred cleanup of files, switch most tests to use testing.TempDir() which is automatically cleaned up.

 - [x] Tests added

Resolves #

 - Change test utility methods to use `testing.TempDir()`
 - Remove (now superfluous) deferred directory and file cleanup

/cc @nats-io/core

### DIFF
--- a/logger/log_test.go
+++ b/logger/log_test.go
@@ -25,8 +25,6 @@ import (
 	"testing"
 )
 
-var tempRoot = filepath.Join(os.TempDir(), "nats-server")
-
 func TestStdLogger(t *testing.T) {
 	logger := NewStdLogger(false, false, false, false, false)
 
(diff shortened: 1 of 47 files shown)


kozlovic (changes requested):
Although I like the idea that we would not have to do all those defer remove, it is going to be painful to verify that this change is not "breaking" something. It seems that the test `TestLeafNodeOperatorBadCfg` is failing, twice in Travis (did not have chance to run locally to check). So there may be something there. Regardless, we may want to keep this one for after 2.9.0 is released...

kozlovic (changes requested):
I see why: t.TempDir() is creating a directory based on the test name. And this test has a VERY long name (that should be fixed regardless). But you see that this may then become a problem...

mprimi:
@kozlovic holding until after 2.9 sounds good (PR is still marked `draft`, I only opened to see if Travis would surface any error I wasn't seeing locally (it did, fixed)

PR #3440, opened 2022-09-06
[FIXED] JetStream: Restarted queue subscriptions may not receive msgs

### LINKED ISSUE #1066: Subscriptions are not propagated correctly upon new leaf node joining
- [x] Defect

#### Versions of `gnatsd` and affected client libraries used:
nats-server 2.0.0 official released

### Consider the following setup

topsetup

publish foo.bar will failed (not received on L1) if with the following order:

Assuming operator, account, user have configured correctly

1. Setup `main cluster`
2. `L1` join to main cluster
3. Subscribe `foo.bar` from `L1`
4. Publish `foo.bar`

Result: the subscriber will receive nothing

I think this related to #982

kozlovic:
The server was not properly handling queue subscriptions internal
notifications which could lead to messages not being delivered
to applications using a queue group to consume from a JetStream
consumer after they were restarted.

Resolves #1066

### DIFF
--- a/server/sublist.go
+++ b/server/sublist.go
@@ -427,7 +427,6 @@ func (s *Sublist) Insert(sub *subscription) error {
 	} else {
 		if n.qsubs == nil {
 			n.qsubs = make(map[string]map[*subscription]*subscription)
-			isnew = true
 		}
 		qname := string(sub.queue)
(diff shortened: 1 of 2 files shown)


derekcollison (approved):
LGTM

PR #3441, opened 2022-09-06
Do not auto-promote direct get functionality. 

derekcollison:

/cc @nats-io/core


ripienaar (approved):
LGTM

kozlovic (changes requested):
Upgrade of client library causes lots of failure, do not merge!

ripienaar:
I dont thnk, for this, we need the client update - `omitempty` is a noop in the context of this PR. So can safely revert, however this is perhaps due to the consumer name/durable_name change?

kozlovic:
@ripienaar And some other changes to error handling where the library returns now a tiny different error text (but also returns full fledge JS API error where one can inspect the code). So that will require quite a bit of careful update of tests. So if that PR can be done without client update, maybe that's the way to go and have a pass after 2.9.0?

ripienaar:
Thanks @kozlovic  been going through and saw that as well, surprised the client library landed that in main thought it was part of the new simplification erfforts only.

I dont think the update is needed - however this is a worrying indication of compatibility issues this close to 2.9.

ripienaar:
I tested just removing these 3 lines that force the change without any omitempty updates, client updates etc - and using old clients that has omitempty.  

The issue remains resolved without client updates

server/test_test.go
```diff
@@ -98,8 +98,15 @@ func require_Error(t *testing.T, err error, expected ...error) {
 	if len(expected) == 0 {
 		return
 	}
+	// Try to strip nats prefix from Go library if present.
+	const natsErrPre = "nats: "
+	eStr := err.Error()
+	if strings.HasPrefix(eStr, natsErrPre) {
```
kozlovic, line 104: Maybe strip only if the expected errors do not themselves have the `nats: ` prefix (or strip also from the expected errors). Still seeing errors such as:
```
=== RUN   TestJWTJetStreamTiers
    jwt_test.go:5517: require equal, but got: nats: maximum number of streams reached != maximum number of streams reached
--- FAIL: TestJWTJetStreamTiers (0.06s)
```

kozlovic (changes requested):

go.sum
```diff
@@ -15,8 +15,18 @@ github.com/minio/highwayhash v1.0.2 h1:Aak5U0nElisjDCfPSG79Tgzkn2gl66NxOMspRrKnA
 github.com/minio/highwayhash v1.0.2/go.mod h1:BQskDq+xkJ12lmlUUi7U0M5Swg3EWR+dLTk+kldvVxY=
 github.com/nats-io/jwt/v2 v2.3.0 h1:z2mA1a7tIf5ShggOFlR1oBPgd6hGqcDYsISxZByUzdI=
 github.com/nats-io/jwt/v2 v2.3.0/go.mod h1:0tqz9Hlu6bCBFLWAASKhE5vUA4c24L9KPUUgvwumE/k=
+github.com/nats-io/nats.go v1.16.0 h1:zvLE7fGBQYW6MWaFaRdsgm9qT39PJDQoju+DS8KsO1g=
```
kozlovic, line 18: Question: did you do a `go mod tidy`?

kozlovic (changes requested):

go.sum
derekcollison: Sorry I always forget that, will try to do better.

kozlovic (approved):
LGTM

PR #3442, opened 2022-09-06
NoAck allowed now on pull consumers

derekcollison:
Still required for WorkQueue retention policy streams.

/cc @nats-io/core

### DIFF
--- a/server/consumer.go
+++ b/server/consumer.go
@@ -400,8 +400,8 @@ func checkConsumerCfg(
 			return NewJSConsumerSmallHeartbeatError()
 		}
 	} else {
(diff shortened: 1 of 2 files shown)


server/consumer.go
```diff
@@ -401,7 +401,7 @@ func checkConsumerCfg(
 		}
 	} else {
 		// Pull mode / work queue mode require explicit ack.
```
kozlovic, line 403: Could update the comment here...

kozlovic (approved):
LGTM, but comment could be updated

server/consumer.go
derekcollison: ok done.

PR #3445, opened 2022-09-07
Suppress leaderless claims if JetStream just started.

derekcollison:
Only complain about leaderless group with previous leader if we know JetStream has been running for some threshold.

/cc @nats-io/core


kozlovic (approved):
LGTM

PR #3451, opened 2022-09-08
JetStream: R1 durables were incorrectly migrated on shutdown

kozlovic:
This could happen for stream with R>1 but with a durable that has an override of R=1.

Fixed a test to make sure assets have an elected leader.

Also fixed a gateway test that would cause a data race.


server/jetstream.go
```diff
@@ -832,7 +832,9 @@ func (s *Server) migrateEphemerals() {
 		for _, sa := range asa {
 			if rg := sa.Group; rg != nil && len(rg.Peers) > 1 && rg.isMember(ourID) && len(sa.consumers) > 0 {
 				for _, ca := range sa.consumers {
-					if ca.Group != nil && len(ca.Group.Peers) == 1 && ca.Group.isMember(ourID) {
+					// Make sure this is not a durable that has an override
+					// of the replicas count.
```
derekcollison, line 836: Can be one line since line below long.

derekcollison (commented):
LGTM

PR #3454, opened 2022-09-08
Adds pagination to stream info

jnmoyne:
 - [X] Tests added

Resolves #

 - Adds pagination to the JS Stream Info API (backwards compatible) that was previously hard-coded to a max number of 100K subject states

/cc @nats-io/core

### DIFF
--- a/server/jetstream_api.go
+++ b/server/jetstream_api.go
@@ -396,16 +396,18 @@ type JSApiStreamDeleteResponse struct {
 
 const JSApiStreamDeleteResponseType = "io.nats.jetstream.api.v1.stream_delete_response"
 
-// Maximum number of subject details we will send in the stream info.
+// JSMaxSubjectDetails The limit of the number of subject details we will send in a stream info response.
 const JSMaxSubjectDetails = 100_000
 
 type JSApiStreamInfoRequest struct {
+	ApiPagedRequest
 	DeletedDetails bool   `json:"deleted_details,omitempty"`
 	SubjectsFilter string `json:"subjects_filter,omitempty"`
 }
 
 type JSApiStreamInfoResponse struct {
 	ApiResponse
+	ApiPaged
 	*StreamInfo
 }
 
@@ -1788,6 +1790,7 @@ func (s *Server) jsStreamInfoRequest(sub *subscription, c *client, a *Account, s
 
 	var details bool
 	var subjects string
+	var offset int
 	if !isEmptyRequest(msg) {
 		var req JSApiStreamInfoRequest
 		if err := json.Unmarshal(msg, &req); err != nil {
@@ -1796,6 +1799,7 @@ func (s *Server) jsStreamInfoRequest(sub *subscription, c *client, a *Account, s
 			return
 		}
 		details, subjects = req.DeletedDetails, req.SubjectsFilter
+		offset = req.Offset
 	}
 
 	mset, err := acc.lookupStream(streamName)
@@ -1825,17 +1829,39 @@ func (s *Server) jsStreamInfoRequest(sub *subscription, c *client, a *Account, s
 	// Check if they have asked for subject details.
 	if subjects != _EMPTY_ {
 		if mss := mset.store.SubjectsState(subjects); len(mss) > 0 {
-			if len(mss) > JSMaxSubjectDetails {
-				resp.StreamInfo = nil
-				resp.Error = NewJSStreamInfoMaxSubjectsError()
-				s.sendAPIErrResponse(ci, acc, subject, reply, string(msg), s.jsonResponse(&resp))
-				return
+			// As go iterates over map in a non-consistent order, no choice but to buffer it a slice
+
+			buffer := make([]string, 0, len(mss))
+			for subj := range mss {
+				buffer = append(buffer, subj)
 			}
-			sd := make(map[string]uint64, len(mss))
-			for subj, ss := range mss {
-				sd[subj] = ss.Msgs
+
+			// Sort it
+			sort.Strings(buffer)
+
+			if offset > len(buffer) {
+				offset = len(buffer)
 			}
+
+			end := offset + JSMaxSubjectDetails
+			if end > len(buffer) {
+				end = len(buffer)
+			}
+
+			actualSize := end - offset
+			var sd map[string]uint64
+
+			if actualSize > 0 {
+				sd = make(map[string]uint64, actualSize)
+				for _, ss := range buffer[offset:end] {
+					sd[ss] = mss[ss].Msgs
+				}
+			}
+
 			resp.StreamInfo.State.Subjects = sd
+			resp.Offset = offset
+			resp.Limit = JSMaxSubjectDetails
+			resp.Total = len(mss)
 		}
 
 	}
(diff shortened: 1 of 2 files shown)


server/jetstream_api.go
```diff
@@ -396,16 +396,18 @@ type JSApiStreamDeleteResponse struct {
 
 const JSApiStreamDeleteResponseType = "io.nats.jetstream.api.v1.stream_delete_response"
 
-// Maximum number of subject details we will send in the stream info.
-const JSMaxSubjectDetails = 100_000
```
derekcollison, line 400: This was technically public so can't remove.

server/jetstream_api.go
```diff
@@ -1825,17 +1829,45 @@ func (s *Server) jsStreamInfoRequest(sub *subscription, c *client, a *Account, s
...
+				subject  string
+				messages uint64
 			}
-			sd := make(map[string]uint64, len(mss))
+
+			buffer := make([]SubjectsState, len(mss))
+			var i int
 			for subj, ss := range mss {
-				sd[subj] = ss.Msgs
+				buffer[i] = SubjectsState{subj, ss.Msgs}
```
derekcollison, line 1841: Just use append and add 0 to make above.

derekcollison: Then don't have to do the `i` thing.

server/jetstream_api.go
```diff
@@ -1825,17 +1829,45 @@ func (s *Server) jsStreamInfoRequest(sub *subscription, c *client, a *Account, s
...
+			})
+
+			limitSize := JSSubjectDetailsRespLimit
+			sd := make(map[string]uint64, limitSize)
+			end := offset + limitSize
+			if end > len(buffer) {
+				end = len(buffer)
+			}
+
+			if offset+limitSize > len(buffer) {
```
derekcollison, line 1857: Can this not just be one iter loop with [offset:end]?

server/jetstream_api.go
jnmoyne: ok

server/jetstream_api.go
jnmoyne: Ok so renamed back to `JSMaxSubjectDetails`

derekcollison (approved):
LGTM - but let's have Ivan eyeball too. Also feel free to squash.

server/jetstream_api.go
```diff
@@ -1825,17 +1829,37 @@ func (s *Server) jsStreamInfoRequest(sub *subscription, c *client, a *Account, s
...
-				s.sendAPIErrResponse(ci, acc, subject, reply, string(msg), s.jsonResponse(&resp))
-				return
+			// As go iterates over map in a non-consistent order, no choice but to buffer it a slice
+			type SubjectsState struct {
+				subject  string
+				messages uint64
 			}
-			sd := make(map[string]uint64, len(mss))
+
+			buffer := make([]SubjectsState, 0, len(mss))
```
kozlovic, line 1838: Why not simply a slice of strings since we care only about sorting the subjects, not the combination of subject and messages. At the end, when it is time to "feed" the map `sd` you would do something like `sd[subject] = mss[subject]`.

(because right now it is worse that you not make it a []*SubjectsState because sort is then done moving structures in that slice, as opposed to move pointers).

server/jetstream_api.go
```diff
@@ -1825,17 +1829,37 @@ func (s *Server) jsStreamInfoRequest(sub *subscription, c *client, a *Account, s
...
+			})
+
+			limitSize := JSMaxSubjectDetails
+			sd := make(map[string]uint64, limitSize)
+			end := offset + limitSize
+			if end > len(buffer) {
+				end = len(buffer)
+			}
+
+			for _, ss := range buffer[offset:end] {
```
kozlovic, line 1855: You need to check for `offset`, not simply cap `end` because then you could have offset > end and get a panic "slice bounds out of range"

kozlovic (changes requested):
Some changes needed, especially to prevent panic.
Also, it means that all clients would need to be updated to support this pagination options in the StreamInfo() API, right?

server/jetstream_api.go
jnmoyne: Ah yes, good catch added protection against bad requests with offset out of range

jnmoyne:
Yes, the clients need to be updated to be able to leverage pagination. As soon as this is merged I can create PRs (I already have the coding done) for natscli, jsm.go and nats.go (I can probably have a go at Java as well).
Existing clients still work (these changes are backwards compatible) though obviously then can only get the first 100k subject details

server/jetstream_api.go
```diff
@@ -1824,18 +1828,32 @@ func (s *Server) jsStreamInfoRequest(sub *subscription, c *client, a *Account, s
 
 	// Check if they have asked for subject details.
 	if subjects != _EMPTY_ {
-		if mss := mset.store.SubjectsState(subjects); len(mss) > 0 {
-			if len(mss) > JSMaxSubjectDetails {
-				resp.StreamInfo = nil
-				resp.Error = NewJSStreamInfoMaxSubjectsError()
-				s.sendAPIErrResponse(ci, acc, subject, reply, string(msg), s.jsonResponse(&resp))
-				return
+		if mss := mset.store.SubjectsState(subjects); len(mss) > 0 && offset < len(mss) {
```
kozlovic, line 1831: I would not add the offset check here. See below.

server/jetstream_api.go
```diff
@@ -1824,18 +1828,32 @@ func (s *Server) jsStreamInfoRequest(sub *subscription, c *client, a *Account, s
...
+			for subj := range mss {
+				buffer = append(buffer, subj)
+			}
+
+			// Sort it
+			sort.Strings(buffer)
+
+			limitSize := JSMaxSubjectDetails
+			sd := make(map[string]uint64, limitSize)
+			end := offset + limitSize
```
kozlovic, line 1844: Would replace this and lines below with:
```
if offset > len(buffer) {
   offset = len(buffer)
}
end := offset + limitSize
if end > len(buffer) {
   end = len(buffer)
}
```
The advantage of that vs checking above is that you will send a response with proper offset, but no subjects, signaling the end of the pagination.

server/jetstream_api.go
```diff
@@ -1824,18 +1828,32 @@ func (s *Server) jsStreamInfoRequest(sub *subscription, c *client, a *Account, s
...
+
+			buffer := make([]string, 0, len(mss))
+			for subj := range mss {
+				buffer = append(buffer, subj)
+			}
+
+			// Sort it
+			sort.Strings(buffer)
+
+			limitSize := JSMaxSubjectDetails
```
kozlovic, line 1842: Move these 2 lines below computation of offset and end, and you could have:
```
var sd map[string]uint64
limitSize := end - offset
if limitSize > 0 {
   sd = make(map[string]uint64, limitSize)
   for _, ss := range buffer[offset:end] {
   ...
}
resp.StreamInfo.State.Subjects = sd
...
```

kozlovic (changes requested):

server/jetstream_api.go
jnmoyne: I see what you mean, implemented the optimization thanks for the suggestion

kozlovic (approved):
LGTM

PR #3456, opened 2022-09-08
[IMPROVED] Pull request logic

### LINKED ISSUE #3455: Possible regression in pull consumer fetch to maximize batch size
In nightly, given a stream that is pre-populated and a pull consumer with two (or more) subscriptions using separate connections, a combined fetch size (between all subscriptions) that is smaller than max ack pending, I am observing varying returned batch sizes rather than the max size.

I tested the same code against the nightly tagged for July 5th and the issue was not observed. I tested the nightly tagged July 6th and the issue is present. Here is the [commit range from July 4-6](https://github.com/nats-io/nats-server/commits/main?since=2022-07-04&until=2022-07-06). Possible suspect is #3241

derekcollison:
Prepping for pull consumer increased usage in clients.

1. Do better accounting on per message basis of bytes used for MaxBytes in pull requests.
2. Always used the MaxBytes allowed unless first delivery.
3. Fixed bug that would cause us to expire one shots on hitting max ack pending.

Resolves #3455 

/cc @nats-io/core

### DIFF
--- a/server/consumer.go
+++ b/server/consumer.go
@@ -3070,10 +3070,12 @@ func (o *consumer) loopAndGatherMsgs(qch chan struct{}) {
 	// Deliver all the msgs we have now, once done or on a condition, we wait for new ones.
 	for {
 		var (
-			pmsg  *jsPubMsg
-			dc    uint64
-			dsubj string
-			delay time.Duration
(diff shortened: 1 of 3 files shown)


kozlovic (approved):
LGTM

PR #3457, opened 2022-09-08
Fixed some tests

kozlovic:
Code change:
- Do not start the processMirrorMsgs and processSourceMsgs go routine if the server has been detected to be shutdown. This would otherwise leave some go routine running at the end of some tests.
- Pass the fch and qch to the consumerFileStore's flushLoop otherwise in some tests this routine could be left running.

Tests changes:
- Added missing defer NATS connection close
- Added missing defer server shutdown


derekcollison (approved):
LGTM - Awesome!

PR #3459, opened 2022-09-08
changed format of JSClusterNoPeers error

matthiashanel:
This error was introduced in #3342 and reveals too much information 
This change gets rid of cluster names and peer counts.

All other counts where changed to booleans,
which are only included in the output when the filter was hit.

In addition, the set of not matching tags is included.

### DIFF
--- a/server/jetstream_cluster.go
+++ b/server/jetstream_cluster.go
@@ -4555,7 +4555,6 @@ type selectPeerError struct {
 
 func (e *selectPeerError) Error() string {
 	b := strings.Builder{}
-	b.WriteString("peer selection")
 	var firstWritten bool
 	writeBoolErrReason := func(hasErr bool, errMsg string) {
 		if !hasErr {
@@ -4567,7 +4566,6 @@ func (e *selectPeerError) Error() string {
 		firstWritten = true
 		b.WriteString(errMsg)
 	}
-	b.WriteString(" failures: ")
 	writeBoolErrReason(e.offline, "peer offline")
 	writeBoolErrReason(e.excludeTag, "exclude tag set")
 	writeBoolErrReason(e.noStorage, "insufficient storage")
--- a/server/jetstream_cluster.go
+++ b/server/jetstream_cluster.go
@@ -4555,17 +4555,14 @@ type selectPeerError struct {
 
 func (e *selectPeerError) Error() string {
 	b := strings.Builder{}
-	var firstWritten bool
 	writeBoolErrReason := func(hasErr bool, errMsg string) {
 		if !hasErr {
 			return
 		}
-		if firstWritten {
-			b.WriteString(", ")
-		}
-		firstWritten = true
+		b.WriteString(", ")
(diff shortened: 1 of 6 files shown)


kozlovic:
With a local cluster of 3, when trying to update with a tag (that does not exist), I am getting this with the CLI:
```
nats: error: could not edit Stream TEST: no suitable peers for placement: peer selection failures: tags not matched {'geo:europess'} (10005)
```
I think that is good. @derekcollison I think that matches what you wanted, no?

derekcollison:
Lots of colons, anyway to cut that down to plain sentence?

```
no suitable peers for placement: no tags matching 'geo:europess'
```

server/jetstream_cluster.go
```diff
@@ -4544,43 +4544,89 @@ func (cc *jetStreamCluster) remapStreamAssignment(sa *streamAssignment, removePe
...
-offline: %d
-excludeTag: %d
-noTagMatch: %d
-noStorage: %d
-uniqueTag: %d
-misc: %d
-`,
-		e.cluster, e.clusterPeers, e.offline, e.excludeTag, e.noTagMatch, e.noStorage, e.uniqueTag, e.misc)
+	b := strings.Builder{}
+	b.WriteString("peer selection")
```
kozlovic, line 4558: Wonder if `peer select failures` is needed though, because we already return through NewJSClusterNoPeersError() which already has "no suitable peers for placement: "

derekcollison (commented):
Still prefer a simpler error that looks more like a sentence. Too many colons.

kozlovic:
Removing the "peer selection failures: " would result in:
```
nats: error: could not edit Stream TEST: no suitable peers for placement: tags not matched {'geo:europess'} (10005)
```
Now, the beginning "nats: error: could not edit Stream TEST:" is from the NATS CLI, so not much we can do there.

derekcollison:
Are we sure we are not returning `error:` prefix from server? Go client puts on `nats:` prefix.

Would prefer..

```
nats: could not edit stream 'TEST': no suitable peers for placement, tags not matched {'geo:europess'} (10005)
```

Note no error:, stream not capitalized, stream name in quotes, and comma vs another colon between placement and tags.

kozlovic:
@derekcollison That is totally NATS CLI. The response is a stream update response, so that's CLI that format this way. Here is the -trace (if we were to remove the "peer selection failures"
```
18:32:11 <<< $JS.API.STREAM.UPDATE.TEST
{"type":"io.nats.jetstream.api.v1.stream_update_response","error":{"code":400,"err_code":10005,"description":"no suitable peers for placement: tags not matched {'geo:europess'}"}}
```

derekcollison (approved):
LGTM

derekcollison:
The tags not matched should be an array, not a map.

Please change to..

```
no suitable peers for placement: tags not matched ['geo:europess']
```

kozlovic (approved):
LGTM

derekcollison (approved):
LGTM

PR #3462, opened 2022-09-09
bump: Update golang.org/x/crypto in response on CVE-2021-44716 

### LINKED ISSUE #3461: nats-servers has CVE-2021-44716
See details in https://nvd.nist.gov/vuln/detail/CVE-2021-44716

The problem becomes from https://github.com/nats-io/nats-server/blob/v2.9.0/go.mod#L13

See at https://cs.opensource.google/go/x/crypto/+/c86fa9a7:go.mod;l=6

denis-tingaikin:
Signed-off-by: Denis Tingaikin <denis.tingajkin@xored.com>

 - [ ] Tests added

Resolves #3461

Closes #3461

/cc @nats-io/core

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -10,7 +10,7 @@ require (
 	github.com/nats-io/nkeys v0.3.0
 	github.com/nats-io/nuid v1.0.1
 	go.uber.org/automaxprocs v1.5.1
-	golang.org/x/crypto v0.0.0-20220829220503-c86fa9a7ed90
+	golang.org/x/crypto v0.0.0-20210314154223-e6e6c4f2bb5b
 	golang.org/x/sys v0.0.0-20220906135438-9e1f76180b77
 	golang.org/x/time v0.0.0-20220722155302-e5dcc9cfc0b9
 )
--- a/go.sum
+++ b/go.sum
@@ -26,9 +26,8 @@ github.com/prashantv/gostub v1.1.0 h1:BTyx3RfQjRHnUWaGF9oQos79AlQ5k8WNktv7VGvVH4
 github.com/stretchr/testify v1.7.1 h1:5TQK59W5E3v0r2duFAb7P95B6hEeOyEnHRa8MjYSMTY=
 go.uber.org/automaxprocs v1.5.1 h1:e1YG66Lrk73dn4qhg8WFSvhF0JuFQF0ERIp4rpuV8Qk=
 go.uber.org/automaxprocs v1.5.1/go.mod h1:BF4eumQw0P9GtnuxxovUd06vwm1o18oMzFtK66vU6XU=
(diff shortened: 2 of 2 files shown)


go.mod
```diff
@@ -10,7 +10,7 @@ require (
 	github.com/nats-io/nkeys v0.3.0
 	github.com/nats-io/nuid v1.0.1
 	go.uber.org/automaxprocs v1.5.1
-	golang.org/x/crypto v0.0.0-20220829220503-c86fa9a7ed90
+	golang.org/x/crypto v0.0.0-20210314154223-e6e6c4f2bb5b
```
kozlovic, line 13: What? This is going back to 2021? I don't think this is right... If anything, we should update to the latest.

kozlovic (changes requested):

go.mod
derekcollison: I did bump it only a day or two back, just FYI..

kozlovic: Yes. The change looks weird since we would go back to an old revision for a CVE that affected Go 1.16 and 1.17.5-, so I asked the original poster some clarifications in the issue report. May close both if there is not a good explanation on why it would be necessary to go back to an old revision!

denis-tingaikin: Agreed. Something is not correct. Checking.

denis-tingaikin: We see a vulnerability 
```
	CVE-2021-44716 - (trivy):
		golang: net/http: limit growth of header canonicalization cache
		net/http in Go before 1.16.12 and 1.17.x before 1.17.5 allows uncontrolled memory consumption in the header canonicalization cache via HTTP/2 requests.
		Package name: golang.org/x/net
		Installed version: v0.0.0-20211112202133-69e39bad7dc2
		Fixed version:     v0.0.0-20211209124913-491a49abca63
```
That is going from the [ttps://github.com/nats-io/nats-server/blob/v2.9.0/go.mod#L13](https://github.com/nats-io/nats-server/blob/v2.9.0/go.mod#L13) => https://cs.opensource.google/go/x/crypto/+/c86fa9a7:go.mod;l=6

Do you have any ideas on how to fix it?

ripienaar: I am guessing you are using a vulnerability scanner that doesn’t really understand how go works. 

But also just reading the CVE and looking what version of go we use tells you we are not vulnerable 

I suggest you run “go version -m nats-server” to see what module versions are actually used and what version go is used. 

Your tool is reporting vulnerabilities found in go.sun which isn’t relevant.

kozlovic:
@denis-tingaikin I am leaning toward closing the issues/PR regarding this. Unless there are objections, I will close "shortly"

kozlovic:
I am now closing this issue since there has not been any replies and I do believe that our dependencies were updated well past the revision that contained the fix, and also that we were not even affected in the first place.

PR #3468, opened 2022-09-13
Optimise `needAck`

neilalexander:
This PR optimises the `needAck` function by pooling the `StoreMsg`s and reusing them. It turns out this is a hot path if there are lots of consumers on a stream, so using a `sync.Pool` here and reusing the existing slices pretty much eliminates the bulk of allocations on this path and almost completely eliminates GC pressure.

(Why it's pegging the CPU in the first place doing this is a mystery I'm still working on, but this improves things at least.)

CPU profile before:
Screenshot 2022-09-13 at 13 07 53

CPU profile after:
Screenshot 2022-09-13 at 13 33 21

Allocation profile before:
Screenshot 2022-09-13 at 13 34 40

Allocation profile after:
Screenshot 2022-09-13 at 13 34 53

### DIFF
--- a/server/consumer.go
+++ b/server/consumer.go
@@ -2320,6 +2320,15 @@ func (o *consumer) isFiltered() bool {
 	return true
 }
 
+// needAck can be called lots of times if there are multiple consumers,
+// so having a pool here drastically reduces the number of allocations
+// and GC pressure.
+var storeMsgPool = &sync.Pool{
+	New: func() interface{} {
+		return &StoreMsg{}
+	},
+}
+
 // Check if we need an ack for this store seq.
 // This is called for interest based retention streams to remove messages.
 func (o *consumer) needAck(sseq uint64) bool {
@@ -2329,10 +2338,11 @@ func (o *consumer) needAck(sseq uint64) bool {
 	o.mu.RLock()
 
 	// Check first if we are filtered, and if so check if this is even applicable to us.
-	var svp StoreMsg
 	if o.isFiltered() && o.mset != nil {
+		svp := storeMsgPool.Get().(*StoreMsg)
 		svp.clear()
-		sm, err := o.mset.store.LoadMsg(sseq, &svp)
+		defer storeMsgPool.Put(svp)
+		sm, err := o.mset.store.LoadMsg(sseq, svp)
 		if err != nil || !o.isFilteredMatch(sm.subj) {
 			o.mu.RUnlock()
 			return false
--- a/server/consumer.go
+++ b/server/consumer.go
@@ -2329,8 +2329,9 @@ func (o *consumer) needAck(sseq uint64) bool {
 	o.mu.RLock()
 
 	// Check first if we are filtered, and if so check if this is even applicable to us.
+	var svp StoreMsg
 	if o.isFiltered() && o.mset != nil {
-		var svp StoreMsg
+		svp.clear()
 		sm, err := o.mset.store.LoadMsg(sseq, &svp)
 		if err != nil || !o.isFilteredMatch(sm.subj) {
 			o.mu.RUnlock()
--- a/server/store.go
+++ b/server/store.go
@@ -547,7 +547,5 @@ func (sm *StoreMsg) clear() {
 	if sm == nil {
 		return
 	}
-	sm.subj = _EMPTY_
-	sm.hdr, sm.msg, sm.buf = sm.hdr[:0], sm.msg[:0], sm.buf[:0]
-	sm.seq, sm.ts = 0, 0
+	*sm = StoreMsg{_EMPTY_, sm.hdr[:0], sm.msg[:0], sm.buf[:0], 0, 0}
 }
--- a/server/store.go
+++ b/server/store.go
@@ -547,8 +547,7 @@ func (sm *StoreMsg) clear() {
 	if sm == nil {
 		return
 	}
-	*sm = StoreMsg{_EMPTY_, nil, nil, sm.buf, 0, 0}
-	if len(sm.buf) > 0 {
-		sm.buf = sm.buf[:0]
-	}
+	sm.subj = _EMPTY_
+	sm.hdr, sm.msg, sm.buf = sm.hdr[:0], sm.msg[:0], sm.buf[:0]
+	sm.seq, sm.ts = 0, 0
 }


derekcollison:
Will take a look a bit later. Thanks. We also discussed this on slack.

server/consumer.go
```diff
@@ -2330,8 +2339,10 @@ func (o *consumer) needAck(sseq uint64) bool {
 
 	// Check first if we are filtered, and if so check if this is even applicable to us.
 	if o.isFiltered() && o.mset != nil {
-		var svp StoreMsg
-		sm, err := o.mset.store.LoadMsg(sseq, &svp)
+		svp := storeMsgPool.Get().(*StoreMsg)
+		svp.clear()
```
kozlovic, line 2343: Should this be part of the pool Get() so one does not have to call it every time (and likely forget ;-))?

server/store.go
```diff
@@ -547,8 +547,7 @@ func (sm *StoreMsg) clear() {
 	if sm == nil {
 		return
 	}
-	*sm = StoreMsg{_EMPTY_, nil, nil, sm.buf, 0, 0}
-	if len(sm.buf) > 0 {
-		sm.buf = sm.buf[:0]
-	}
+	sm.subj = _EMPTY_
```
kozlovic, line 550: The advantage of the method above is that if StoreMsg{} has a new field, the code above would fail to compile until we update it to take care of the new field.
With your current approach, in that same situation, it requires someone to not forget to add that.

server/store.go
```diff
@@ -547,8 +547,7 @@ func (sm *StoreMsg) clear() {
...
+	sm.hdr, sm.msg, sm.buf = sm.hdr[:0], sm.msg[:0], sm.buf[:0]
```
kozlovic, line 551: The comment at top of function was saying that all fields are nil/empty, except for buf if present but reset to [:0]. I honestly don't know the implications of sm.hdr[:0] as opposed to set it to nil (same for sm.msg), in that if caller checks (incorrectly?) nil vs len(hdr)==0 to decide what to do, then this is a change. Again, correct code should not care and so that should not make a difference, but again, not sure. If that does not change your profiling, maybe keep those at nil?

server/consumer.go
neilalexander: The `Get` method is only called when the pool needs to create a new instance. It doesn’t get called when recycling something that was returned to the pool, so we need to do it this way.

server/store.go
neilalexander: This is true, although it will be more expensive to copy the full struct. I don’t know how much that would really matter in the real world.

server/store.go
neilalexander: I can’t be sure there aren’t checks for `nil` precisely, but that would be quite surprising if so. I’ve been running the branch today and it seems to be OK admittedly.

server/consumer.go
kozlovic: You are right!

server/store.go
kozlovic: Will see if @derekcollison thinks that this could be an issue or not.

server/consumer.go
```diff
@@ -2329,8 +2329,9 @@ func (o *consumer) needAck(sseq uint64) bool {
 	o.mu.RLock()
 
 	// Check first if we are filtered, and if so check if this is even applicable to us.
+	var svp StoreMsg
```
derekcollison: What is the reasoning for moving outside the block here? Seems scoping is better where it was?
Also why call clear on it here since its on the stack already and initialized?

server/store.go
derekcollison: We could combine both approaches possibly.

```go
*sm = StoreMsg{_EMPTY_, sm.hdr[:0], ...}
```

But for this to be effective we would need to reuse any capacity there in the lower layers of filestore and memstore.

Also, I know `sm.hdr = sm.hdr[:0]` still works if original is nil but I did a double take at first.

server/consumer.go
```diff
@@ -2329,10 +2338,11 @@ func (o *consumer) needAck(sseq uint64) bool {
 	o.mu.RLock()
 
 	// Check first if we are filtered, and if so check if this is even applicable to us.
-	var svp StoreMsg
 	if o.isFiltered() && o.mset != nil {
+		svp := storeMsgPool.Get().(*StoreMsg)
```
derekcollison, line 2342: If the pooled messages hold onto interior items this could be beneficial but at first is counter intuitive, since we are adding work compared to accessing a stack variable and struct.

We could determine where we are stressed the most with the interior slices and preallocate that as well on the stack and place it into svp.

derekcollison: After reviewing I can see we reuse the buf filed of a StoredMsg if present, so the benefit is coming from there for sure.

derekcollison (commented):
So left some comments and now just did a deep dive on how that is filled in inside filestore and memstore.

We spin completely off of the buf pointer and we intelligently re-use it if not nil.  So I think we can get the same benefit here with a bit of a simpler approach.

And we still need (IMO) to understand why needAck() was being called in the scenario in the first place. We can see it is rooted in processInboundAcks but I am interested in what queued up all the acks in the scenario originally described.

derekcollison:
I also see a better optimization. So may code that up and ask @neilalexander to take a look. With lots of consumers we should load the message once before looping though asking about interest etc.

server/consumer.go
neilalexander: The struct itself probably would stay on the stack in the happy case but the slices are bleeding out into the heap either way, so admittedly there's a bit of a trade-off to make here. Moving it to the heap (via the `sync.Pool`) and keeping the slice references is probably the best way to keep the underlying capacity alive, since that's where you're going to save on GC pressure. Otherwise every time the struct goes out of scope, the inner slices are marked for GC.

Am I right in thinking this code path is hot in just about any situation where `AckAll` is used? In my case I think I hit it by accident due to trying to ack messages that no longer exist (as we discussed) but it's probably also a genuine case when there *are* messages to ack too.

server/store.go
neilalexander: Agreed, I've made that change now.

server/consumer.go
neilalexander: I think this is a comment on an out-of-date commit.

neilalexander:
@derekcollison Sounds good — more than happy to test other branches!

server/consumer.go
derekcollison: I will dig in today. And thanks again for this PR and the insight, much appreciated. Will ping/tag you on PR when ready so you can review and test.

neilalexander:
Superseded by #3472.

PR #3472, opened 2022-09-14
Optimize needAck.

derekcollison:
Based off excellent insight and PR #3468 by @neilalexander.

This reuses pools we already had and also only loads the msg once for all consumers.

/cc @nats-io/core

### DIFF
--- a/server/consumer.go
+++ b/server/consumer.go
@@ -2328,17 +2328,22 @@ func (o *consumer) isFiltered() bool {
 
 // Check if we need an ack for this store seq.
 // This is called for interest based retention streams to remove messages.
-func (o *consumer) needAck(sseq uint64) bool {
+func (o *consumer) needAck(sseq uint64, subj string) bool {
 	var needAck bool
 	var asflr, osseq uint64
 	var pending map[uint64]*Pending
 	o.mu.RLock()
 
-	// Check first if we are filtered, and if so check if this is even applicable to us.
+	// Check if we are filtered, and if so check if this is even applicable to us.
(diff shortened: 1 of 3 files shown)


kozlovic (approved):
LGTM. Let's see if @neilalexander can run that and see if that helps as it shown to help on the added test.

derekcollison:
I added another optimization in case we are chasing lots of sequences below our first as well.

kozlovic (approved):
LGTM

neilalexander (approved):
This looks good — I've just tested it and it is no longer burning resources on `needAck` when creating the new ephemeral consumer after application startup.

server/consumer.go
```diff
@@ -2328,17 +2328,22 @@ func (o *consumer) isFiltered() bool {
...
-	// Check first if we are filtered, and if so check if this is even applicable to us.
+	// Check if we are filtered, and if so check if this is even applicable to us.
 	if o.isFiltered() && o.mset != nil {
-		var svp StoreMsg
-		sm, err := o.mset.store.LoadMsg(sseq, &svp)
-		if err != nil || !o.isFilteredMatch(sm.subj) {
+		if subj == _EMPTY_ {
+			var svp StoreMsg
+			if _, err := o.mset.store.LoadMsg(sseq, &svp); err != nil {
+				return false
```
ashumkin, line 2342: I guess, `o.mu.RUnlock()` is missed here (before `return`)?

neilalexander: Looks like it — https://github.com/nats-io/nats-server/pull/3555

PR #3473, opened 2022-09-15
[ADDED] JetStream: ability to remove a server by peer ID instead of name

kozlovic:
This can be helpful after a partial cluster restart since in that case the server name may not be known. However "server report jetstream" would report the peer ID that then can be used.

For instance here is the output after a cluster restart where server "C" is not restarted.

```
nats -s nats://sys:pwd@localhost:4222 server report jetstream
...
╭────────────────────────────────────────────────────────────────────────────────────────────────╮
│                                  RAFT Meta Group Information                                   │
├─────────────────────────────────────────────────────┬────────┬─────────┬────────┬────────┬─────┤
│ Name                                                │ Leader │ Current │ Online │ Active │ Lag │
├─────────────────────────────────────────────────────┼────────┼─────────┼────────┼────────┼─────┤
│ A                                                   │ yes    │ true    │ true   │ 0.00s  │ 0   │
│ B                                                   │        │ true    │ true   │ 0.53s  │ 0   │
│ Server name unknown at this time (peerID: jZ6RvVRH) │        │ false   │ false  │ 0.00s  │ 0   │
╰─────────────────────────────────────────────────────┴────────┴─────────┴────────┴────────┴─────╯
```

With a change to the NATS CLI we could have something like:
```
nats -s nats://sys:pwd@localhost:4222 server raft peer-remove jZ6RvVRH --by_id
```


kozlovic:
@derekcollison @ripienaar Not sure if the "ByID" makes sense or it should be a different struct/subject.

@ripienaar I wanted to have a go at submitting a PR for the NATS CLI, but it is way more involved than I thought (have to modify jsm.go, and was not sure about the schemas, how to regenerate them, etc..). So I hack'ed my way just to see if I could make it work somehow. Here is a diff that shows a new option for the CLI and manual request to the subject with a local struct. Of course, this is not proper way... again, more of a POC for me to check experience with CLI against this server PR:
```
diff --git a/cli/server_raft_command.go b/cli/server_raft_command.go
index 8716a7e..41ab5aa 100644
--- a/cli/server_raft_command.go
+++ b/cli/server_raft_command.go
@@ -22,11 +22,13 @@ import (
        "github.com/choria-io/fisk"
        "github.com/nats-io/jsm.go/api"
        "github.com/nats-io/nats-server/v2/server"
+       "github.com/nats-io/nats.go"
 )
 
 type SrvRaftCmd struct {
        json             bool
        force            bool
+       byID             bool
        peer             string
        placementCluster string
 }
@@ -43,6 +45,7 @@ func configureServerRaftCommand(srv *fisk.CmdClause) {
        rm := raft.Command("peer-remove", "Removes a server from a JetStream cluster").Alias("rm").Alias("pr").Action(c.metaPeerRemove)
        rm.Arg("name", "The Server Name to remove from the JetStream cluster").StringVar(&c.peer)
        rm.Flag("force", "Force removal without prompting").Short('f').UnNegatableBoolVar(&c.force)
+       rm.Flag("by_id", "Use the name as the peer ID, not the server name").UnNegatableBoolVar(&c.byID)
 }
 
 func (c *SrvRaftCmd) metaPeerRemove(_ *fisk.ParseContext) error {
@@ -60,29 +63,31 @@ func (c *SrvRaftCmd) metaPeerRemove(_ *fisk.ParseContext) error {
                return fmt.Errorf("did not receive a response from the meta leader, ensure the account used has system privileges and appropriate permissions")
        }
 
-       type jszr struct {
-               Data   server.JSInfo     `json:"data"`
-               Server server.ServerInfo `json:"server"`
-       }
+       if !c.byID {
+               type jszr struct {
+                       Data   server.JSInfo     `json:"data"`
+                       Server server.ServerInfo `json:"server"`
+               }
 
-       found := false
-       srv := &jszr{}
-       err = json.Unmarshal(res[0], srv)
-       if err != nil {
-               return err
-       }
+               found := false
+               srv := &jszr{}
+               err = json.Unmarshal(res[0], srv)
+               if err != nil {
+                       return err
+               }
 
-       for _, r := range srv.Data.Meta.Replicas {
-               if r.Name == c.peer {
-                       if !r.Offline {
-                               return fmt.Errorf("can only remove offline nodes")
+               for _, r := range srv.Data.Meta.Replicas {
+                       if r.Name == c.peer {
+                               if !r.Offline {
+                                       return fmt.Errorf("can only remove offline nodes")
+                               found = true
                        }
-                       found = true
                }
-       }
 
-       if !found {
-               return fmt.Errorf("did not find a replica named %s", c.peer)
+               if !found {
+                       return fmt.Errorf("did not find a replica named %s", c.peer)
+               }
        }
 
        if !c.force {
@@ -96,12 +101,48 @@ func (c *SrvRaftCmd) metaPeerRemove(_ *fisk.ParseContext) error {
                }
        }
 
-       err = mgr.MetaPeerRemove(c.peer)
+       if c.byID {
+               err = c.removeByID(nc)
+       } else {
+               err = mgr.MetaPeerRemove(c.peer)
+       }
        fisk.FatalIfError(err, "Could not remove %s", c.peer)
 
        return nil
 }
 
+func (c *SrvRaftCmd) removeByID(nc *nats.Conn) error {
+       type removeByID struct {
+               api.JSApiMetaServerRemoveRequest
+               ByID bool `json:"by_id,omitempty"`
+       }
+       req := removeByID{}
+       req.Server, req.ByID = c.peer, true
+
+       var jreq []byte
+       jreq, err := json.Marshal(&req)
+       if err != nil {
+               return err
+       }
+
+       var mresp *nats.Msg
+       mresp, err = nc.Request(api.JSApiRemoveServer, jreq, opts.Timeout)
+       if err != nil {
+               return err
+       }
+
+       var resp api.JSApiMetaServerRemoveResponse
+       err = json.Unmarshal(mresp.Data, &resp)
+       if err != nil {
+               return err
+       }
+
+       if !resp.Success {
+               return fmt.Errorf("unknown error while requesting leader step down")
+       }
+       return nil
+}
+
 func (c *SrvRaftCmd) metaLeaderStandDown(_ *fisk.ParseContext) error {
        nc, mgr, err := prepareHelper("", natsOpts()...)
        if err != nil {
```

server/jetstream_api.go
```diff
@@ -581,6 +581,7 @@ const JSApiLeaderStepDownResponseType = "io.nats.jetstream.api.v1.meta_leader_st
 type JSApiMetaServerRemoveRequest struct {
 	// Server name of the peer to be removed.
 	Server string `json:"peer"`
+	ByID   bool   `json:"by_id,omitempty"`
```
derekcollison, line 584: Just `Peer`

derekcollison (approved):
LGTM - Would just say peer.

ripienaar:
In order for me to build friendly UI i need to let people pick names and such, or at least show them in nice columns and so forth.

So it would be good if `MetaClusterInfo` included this ID in the data, I dont want to have to parse `Server name unknown at this time (peerID: ae1WMeTj)` which wouldnt be reliable

ripienaar:
@kozlovic thanks for the poc :) indeed once schema changes are involved it gets a bit complicated

kozlovic:
> In order for me to build friendly UI i need to let people pick names and such, or at least show them in nice columns and so forth. So it would be good if MetaClusterInfo included this ID in the data

I am a bit confused because right now, even if I want to remove by server name, if I don't specify the server name in the command "nats -s nats://sys:pwd@localhost:4222 server raft peer-remove", I am not proposed a list. So why would it be different for the peer ID?

And as I have shown in the PR description, since v2.9.0, the server will report the peer ID for a server name that is currently unknown. So if the user wants to remove by peer ID, the ID is present.

I understand what you were saying about having the peer ID separate to show in column or for a selection, but that is not currently happening. Maybe it is the case for the "stream peer-remove", but not the raft.

ripienaar:
I am not prsenting a list *today*, but its something I want to improve, so ensuring its possible to do that in fuure while touching this code now means the server is ready.

kozlovic:
Ok, I will revisit that then. Thanks!

server/jetstream_api.go
ripienaar: Why not take whats given in `peer` and try to use it as a name, if it fails try to use it as a ID and if fails errors

kozlovic: I thought about that, and we would not need to update the request field, which is nice, but then I got concerned that what if a server name is equal to the peer ID (the hash) of another server. Say server name "A" has id "B" (I know it is 8 chars, but you get the idea) and server "B"'s id is "A". It would have to be VERY clear that the name provided to the peer-remove command is by "server name" first, and only if not found would we remove looking by ID. Because if the user uses the "raft peer-remove B" and assumes that B is the peer ID, then instead of remove server "A" (whose ID is "B"), the server "B" (id "A") is being removed. Sounds crazy because very unlikely that peer ID and server names ever cross-match, but since the command is "peer-remove" (not server-remove), for a new user it could be possible to make the mistake, especially if we do provide the "peer ID" as a new column.

Don't get me wrong, I like your approach and that simplifies a lot (and does not need a CLI change), but I wanted to raise the far-fetched concern above. Especially if we start returning peer ID to the CLI that would display it as a column. One could say that if we do what you propose maybe we don't need to return explicitly the peer ID in PeerInfo objects, since the Name has the peerID info for the cases where it is not known. What do you think?

ripienaar: Yeah that's a valid concern - and why I like the restriction to only remove offline nodes.  But yes, of course you could have the situation also with offline peers.

I don't want to parse that sentence to extract the peer id (again thinking about the case of presenting a list of nodes)

kozlovic: So you would still be ok to have a command that explicitly ask to remove by ID? (so update of the request struct, etc..)

ripienaar: Perhaps, but are we sure a clash in name/id wont be a problem later? Is uniqueness not something in these things not something we should care about?

kozlovic: But I am saying that if the CLI has a new EXPLICIT command to remove by id (instead of name), then the struct will have "Peer string" and the server will only look for raft peer ID (not server names). So then it is no longer a problem.

kozlovic: Meaning that you have 2 ways to remove: by name or by id, then since it is explicit there is no risk of clash. You can still have typos/messing up, but the same way right now if you remove by name and pick the wrong one :-)

kozlovic:
@derekcollison @ripienaar I have updated the PR to use a Peer (a string) and if that is specified, server will use that to remove by Peer ID. And for @ripienaar I did change PeerInfo to export the Peer id so that it can be visible in the MetaGroupInfo's replicas and could be listed as a separate column, etc..

I have also added a new file for the cluster tests so we keep the running time on Travis at around 5min for the whole test suite.

derekcollison (approved):
LGTM - Thanks.

ripienaar (approved):
LGTM

ripienaar:
Seems good thanks @kozlovic the only problem is we dont know the peer id of the leader as leader isnt in replicas :(

PR #3474, opened 2022-09-15
[FIXED] JetStream: Pull requests closed due to max_bytes were silent

kozlovic:
If the client pull requests has a max_bytes value and the server cannot deliver a single message (because size is too big), it is sending a 409 to signal that to the client library. However, if it sends at least a message then it would close the request without notifying the client with a 409, which would cause the client library to have to wait for its expiration/timeout.

### DIFF
--- a/server/consumer.go
+++ b/server/consumer.go
@@ -2619,11 +2619,10 @@ func (o *consumer) nextWaiting(sz int) *waitingRequest {
 					wr.n = 1
 				}
 			} else {
-				// If we have not delivered anything to the requestor let them know.
-				if wr.d == 0 {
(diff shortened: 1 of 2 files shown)


derekcollison (approved):
LGTM

kozlovic:
@aricart @Jarema @scottf Giving you a chance to comment/raise concerns before I merge this PR (to avoid code in main that may be "broken" from client library perspective).

aricart:
LGTM - currently, we have to parse the messages, so the actual error code is not as important on the error status.

aricart (approved):

PR #3476, opened 2022-09-16
[ADDED] JetStream: peer (the hash of server name) in statsz/jsz

kozlovic:
A request to `$SYS.REQ.SERVER.PING.JSZ` would now return something
like this:
```
...
    "meta_cluster": {
      "name": "local",
      "leader": "A",
      "peer": "NUmM6cRx",
      "replicas": [
        {
          "name": "B",
          "current": true,
          "active": 690369000,
          "peer": "b2oh2L6w"
        },
        {
          "name": "Server name unknown at this time (peerID: jZ6RvVRH)",
          "current": false,
          "offline": true,
          "active": 0,
          "peer": "jZ6RvVRH"
        }
      ],
      "cluster_size": 3
    }
```
Note the "peer" field following the "leader" field that contains
the server name. The new field is the node ID, which is a hash of
the server name.

Related to #3473

### DIFF
--- a/server/consumer.go
+++ b/server/consumer.go
@@ -3857,7 +3857,7 @@ func (o *consumer) String() string {
 }
 
 func createConsumerName() string {
-	return string(getHash(nuid.Next()))
+	return getHash(nuid.Next())
 }
 
 // deleteConsumer will delete the consumer from this stream.
--- a/server/events.go
+++ b/server/events.go
@@ -16,6 +16,7 @@ package server
 import (
 	"bytes"
 	"compress/gzip"
+	"crypto/sha256"
 	"encoding/json"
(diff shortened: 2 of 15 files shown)


derekcollison (approved):
LGTM but would just call it peer, so leader/peer for leader and name/peer for replicas.

kozlovic:
@derekcollison I have updated as a separate commit (https://github.com/nats-io/nats-server/pull/3476/commits/d36321bc4265d2b5d2bdf847b2014202fd19caa5), but already updated the PR's title and content. Will do squash and 1st commit's message update when squashing prior to merge.

derekcollison:
ok sounds good.

PR #3477, opened 2022-09-16
[FIXED] JetStream: possible panic on peer remove on server shutdown

kozlovic:
This was discovered by new test TestJetStreamClusterRemovePeerByID. I saw this on Travis and repeating the test locally with -count=10 I was able to reproduce. The issue is cc.meta being nil but accessing cc.meta.ID() directly.


derekcollison (approved):
LGTM

PR #3478, opened 2022-09-16
When filtering a source stream use new consumer create API subject.

derekcollison:

/cc @nats-io/core

### DIFF
--- a/server/stream.go
+++ b/server/stream.go
@@ -2481,12 +2481,20 @@ func (mset *stream) setSourceConsumer(iname string, seq uint64, startTime time.T
 		return
 	}
 
-	b, _ := json.Marshal(req)
-	subject := fmt.Sprintf(JSApiConsumerCreateT, si.name)
+	var subject string
+	if req.Config.FilterSubject != _EMPTY_ {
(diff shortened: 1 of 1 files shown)


server/stream.go
```diff
@@ -2481,13 +2481,21 @@ func (mset *stream) setSourceConsumer(iname string, seq uint64, startTime time.T
 		return
 	}
 
-	b, _ := json.Marshal(req)
-	subject := fmt.Sprintf(JSApiConsumerCreateT, si.name)
+	var subject string
+	if req.Config.FilterSubject != _EMPTY_ {
+		req.Config.Name = fmt.Sprintf("src-%s", createConsumerName())
+		subject = fmt.Sprintf(JSApiConsumerCreateExT, si.name, req.Config.Name, req.Config.FilterSubject)
```
kozlovic, line 2487: What if the server connects to a pre v2.9.0? This request will timeout and likely be retried for ever...

derekcollison: Yep that is correct.. I am ok with that.

kozlovic (approved):
LGTM

PR #3487, opened 2022-09-22
Allow discard new per subject for certain KV type scenarios.

derekcollison:
Require general DiscardNewPolicy to bet set as well.

/cc @nats-io/core

### DIFF
--- a/server/jetstream_cluster_3_test.go
+++ b/server/jetstream_cluster_3_test.go
@@ -134,7 +134,6 @@ func TestJetStreamClusterDiscardNewAndMaxMsgsPerSubject(t *testing.T) {
 				Name:          "KV",
 				Subjects:      []string{"KV.>"},
 				Storage:       test.storage,
-				MaxMsgsPer:    1,
 				AllowDirect:   true,
 				DiscardNewPer: true,
 				MaxMsgs:       10,
@@ -145,8 +144,17 @@ func TestJetStreamClusterDiscardNewAndMaxMsgsPerSubject(t *testing.T) {
 			} else if apiErr.ErrCode != 10052 || !strings.Contains(apiErr.Description, "discard new per subject requires discard new policy") {
 				t.Fatalf("Got wrong error: %+v", apiErr)
 			}
+
 			// Set broad discard new policy to engage DiscardNewPer
 			cfg.Discard = DiscardNew
+			// We should also error here since we have not setup max msgs per subject.
+			if _, apiErr := addStreamWithError(t, nc, cfg); apiErr == nil {
+				t.Fatalf("Expected API error but got none")
+			} else if apiErr.ErrCode != 10052 || !strings.Contains(apiErr.Description, "discard new per subject requires max msgs per subject > 0") {
+				t.Fatalf("Got wrong error: %+v", apiErr)
+			}
+
+			cfg.MaxMsgsPer = 1
 			addStream(t, nc, cfg)
 
 			// We want to test that we reject new messages on a per subject basis if the
--- a/server/jetstream_cluster_3_test.go
+++ b/server/jetstream_cluster_3_test.go
@@ -18,6 +18,7 @@ package server
 
 import (
 	"encoding/json"
+	"strings"
 	"testing"
 	"time"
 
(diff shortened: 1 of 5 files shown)


server/jetstream_cluster_3_test.go
```diff
@@ -106,3 +107,55 @@ func TestJetStreamClusterRemovePeerByID(t *testing.T) {
...
+			cfg.Discard = DiscardNew
+			addStream(t, nc, cfg)
+
+			// We want to test that we reject new messages on a per subject basis if the
+			// max msgs per subject limit has been hit, even if other limits have not.
+			_, err := js.Publish("KV.foo", nil)
+			require_NoError(t, err)
+
+			_, err = js.Publish("KV.foo", nil)
+			require_Error(t, err)
```
ripienaar, line 158: would probably check, to be safe, that it would reject on MaxMsgsPer > 1, and that the error is the one we exxpect to get.

aricart:
Is the intent of this for max per subject of 1, and discard new per always replaces?

server/jetstream_cluster_3_test.go
derekcollison: ok makes sense.

derekcollison:
@aricart the default right now is with DiscardNew and MaxMsgPer we treat per subject as discard old and the discard new only applies to upper level stream limits like total max msgs or max bytes. This allows the discard new to be effective on a per subject basis, allowing distributed locks, semaphore gates etc.

aricart:
> @aricart the default right now is with DiscardNew and MaxMsgPer we treat per subject as discard old and the discard new only applies to upper level stream limits like total max msgs or max bytes. This allows the discard new to be effective on a per subject basis, allowing distributed locks, semaphore gates etc.

So this is clamping on the discard to take effect within the scope of the key?

ripienaar:
@aricart this gets back past behavior before DiscardNew was turned into a "how many keys a KV bucket can have" so we're back to what it says on the name - ensures a subject can have just n messages.

aricart:
this is awesome!

ripienaar:
Yeah, we cant change default behavior, but I do many non KV things with max msgs per subject - some which became impossible when that change landed.

server/jetstream_cluster_3_test.go
ripienaar: Good, but I meant in addition to test on MaxMsgsPer=10 for eg so that it only rejects messages once we hit the right number but allowing the rest, and then that the error you get on publishing the 11th message is the actual error you expect.

derekcollison: Oh you mean the lowest level add stream, ok.

derekcollison: ok added.

ripienaar (approved):
LGTM but lets hear from @kozlovic of course

kozlovic (commented):
LGTM. I have verified that it does not break some of the existing clients (C and Go), which is expected since the code addition is about a new field that the clients don't know about yet :-)

server/jetstream_cluster_3_test.go
```diff
@@ -106,3 +108,64 @@ func TestJetStreamClusterRemovePeerByID(t *testing.T) {
...
+			addStream(t, nc, cfg)
+
+			// We want to test that we reject new messages on a per subject basis if the
+			// max msgs per subject limit has been hit, even if other limits have not.
+			_, err := js.Publish("KV.foo", nil)
+			require_NoError(t, err)
+
+			_, err = js.Publish("KV.foo", nil)
+			// Go client does not have const for this one.
+			require_Error(t, err, errors.New("nats: maximum messages per subject exceeded"))
```
aricart, line 168: The top level error code returned here by the server is a `503` which is a no responders/timeout.  It would be much better if it was something else.
```
> MSG _INBOX.TAK1HLDINPNI7SATVNPK66.TAK1HLDINPNI7SATVNPKHU 1  117␍␊{"error":{"code":503,"err_code":10077,"description":"maximum messages per subject exceeded"},"stream":"KV_A","seq":0}␍␊
```

kozlovic: Not really. 503 is considered no responders when it is a message header (Status 503), but this is part of the JSAPiError, which is treated differently than a no-payload message with header status 503.

PR #3489, opened 2022-09-22
[FIXED] LeafNode: connecting using websocket and no_auth_user

kozlovic:
If the `no_auth_user` is set in the `websocket{}` block and a server creates a leafnode connection using the websocket port, and does not provide credentials, that no_auth_user should be used, but was not.


derekcollison (approved):
LGTM

PR #3498, opened 2022-09-26
[FIXED] JetStream: prevent panic on consumer assignment

kozlovic:
It could be that while the routine processing the consumer assignment runs the stream is being stopped, which would lead to a panic.


derekcollison (approved):
LGTM

PR #3501, opened 2022-09-27
[FIXED] JetStream: File store memory usage

kozlovic:
The write cache may be pinned for longer than needed when creating
a new write block. This could be seen in some benchmark tests.

The old block cache would be kept for 5 more seconds, which, with
a fast rate of inserts could start to show in some memory profiling.

This was a change introduced in https://github.com/nats-io/nats-server/pull/3351
which was different than code in v2.8.4

### DIFF
--- a/server/filestore.go
+++ b/server/filestore.go
@@ -4782,7 +4782,11 @@ func (fs *fileStore) Compact(seq uint64) (uint64, error) {
 			if err = os.WriteFile(smb.mfn, nbuf, defaultFilePerms); err != nil {
 				goto SKIP
 			}
+			// Make sure to remove fss state.
 			smb.fss = nil
+			if smb.sfn != _EMPTY_ {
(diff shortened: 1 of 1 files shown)


server/filestore.go
```diff
@@ -4782,7 +4782,11 @@ func (fs *fileStore) Compact(seq uint64) (uint64, error) {
 			if err = os.WriteFile(smb.mfn, nbuf, defaultFilePerms); err != nil {
 				goto SKIP
 			}
+			// Make sure to remove fss state.
 			smb.fss = nil
+			if smb.sfn != _EMPTY_ {
+				os.Remove(smb.sfn)
```
kozlovic, line 4788: LGTM! (I can't "formally" approve with "Review changes" since I created the PR)

derekcollison (approved):
LGTM

PR #3504, opened 2022-09-28
[IMPROVED] Fan-out performance

kozlovic:
There was an observed degradation (around 5%) for large fan out in v2.9.0 compared to earlier release. This is because we added accounting of the in/out messages for the account, which result in 4 atomic operations, 2 for in and 2 for out, however, it means that for a fan-out of say 100 matching subscriptions, it is now 2 + 2 * 100 = 202.

This PR rework how the stats accounting is done which removes the regression and even boost a bit the numbers since we are doing the server stats update as an aggregate too.

There are still degradation for queues and no-sub at all that need to be looked at.

### DIFF
--- a/server/client.go
+++ b/server/client.go
@@ -3084,7 +3084,7 @@ var needFlush = struct{}{}
 
 // deliverMsg will deliver a message to a matching subscription and its underlying client.
 // We process all connection/client types. mh is the part that will be protocol/client specific.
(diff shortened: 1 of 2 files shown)


kozlovic:
@mprimi This PR should help with the fan-out cases, but I still see some differences with queue specifically, so needs more digging.

derekcollison (approved):
LGTM

PR #3508, opened 2022-09-29
Allow meta layer to snapshot on a clean shutdown.

derekcollison:
Added protection for meta-layer.

/cc @nats-io/core


server/jetstream.go
```diff
@@ -946,6 +952,21 @@ func (s *Server) shutdownJetStream() {
 		cc.meta = nil
 	}
 	js.mu.Unlock()
+
+	// If we were clustered signal the monitor consumer go routine.
```
kozlovic, line 956: You mean the monitor cluster, right?

server/jetstream.go
```diff
@@ -946,6 +952,21 @@ func (s *Server) shutdownJetStream() {
...
+	// Do this without the lock.
+	if qch != nil {
+		select {
+		case qch <- struct{}{}:
+			select {
+			case <-qch:
+			case <-time.After(2 * time.Second):
+				s.Warnf("Did not receive signal for successful shutdown of cluster routine")
+			}
+		default:
```
kozlovic, line 967: Not sure that works then, because if the monitorCluster is not actively consuming from that channel at the time of the send line 961, then this is where you end-up (since you created an unbuffered channel). If it was a buffered, then you would have the issue at line 963 where you would possibly consume your own notification and not let the monitorCluster go routine consume it.

server/jetstream.go
derekcollison: Yes will correct comment.

server/jetstream.go
derekcollison: It works, tested it and verified with prints that it does right thing, one test has the routine exit before and that is why its selective.

kozlovic: If you don't want to use a waitGroup to wait on monitorCluster to finish and still want to be time bound, then you need 2 channels, one that you sent to (buffered size 1) and still use default if you think that there may be 2 "shutdown" invocations possible, and the monitorCluster will close the other channel that this routine is going to wait on.

derekcollison: Also new test there if snapshot is not performed will fail at some point with count=20.

PR #3510, opened 2022-09-29
[FIXED] bug in consumer names paging, did not honor limits and returned duplicate results.

derekcollison:

/cc @nats-io/core


ripienaar (approved):
LGTM if tests green

kozlovic (commented):
LGTM

PR #3526, opened 2022-10-06
[ADDED] Ability to build on NetBSD

### LINKED ISSUE #3525: Enable nats-server to be built on NetBSD
#### Use Case:

I want to build Dendrite (Matrix Server) on NetBSD. Nats server is a dependency. Currently, Nats server cannot be built on NetBSD because of the following loopholes:

 - pse not supported
 - disk availability not supported

#### Proposed Change:

The necessary adjustments in the source code for a minimal version for NetBSD are minor. Process Usage (pse) can be adopted from OpenBSD. Disk availability could be taken over from Windows (dummy code).

NetBSD users

#### Alternative Approaches

unknown

MatthiasPetermann:
Signed-off-by: Matthias Petermann <mp@petermann-it.de>

 - [-] Tests added
 - [-] Changes squashed to a single commit (described [here](http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html))
 - [?] Build is green in Travis CI

Resolves #3525

 - Add pse support on NetBSD (based on OpenBSD work)
 - Add (dummy) support for disk availability on NetBSD (similiar to Windows work) - I will work on real support on NetBSD later

/cc @nats-io/core

### DIFF
--- a/server/disk_avail_netbsd.go
+++ b/server/disk_avail_netbsd.go
@@ -1,4 +1,4 @@
-// Copyright 2020-2022 The NATS Authors
+// Copyright 2022 The NATS Authors
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
 // You may obtain a copy of the License at
@@ -11,28 +11,12 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-//go:build !windows && !openbsd && !netbsd && !wasm
-// +build !windows,!openbsd,!netbsd,!wasm
+//go:build netbsd
+// +build netbsd
 
 package server
 
-import (
-	"os"
-	"syscall"
-)
-
+// TODO - See if there is a version of this for NetBSD.
 func diskAvailable(storeDir string) int64 {
-	var ba int64
-	if _, err := os.Stat(storeDir); os.IsNotExist(err) {
-		os.MkdirAll(storeDir, defaultDirPerms)
-	}
-	var fs syscall.Statfs_t
-	if err := syscall.Statfs(storeDir, &fs); err == nil {
-		// Estimate 75% of available storage.
-		ba = int64(uint64(fs.Bavail) * uint64(fs.Bsize) / 4 * 3)
-	} else {
-		// Used 1TB default as a guess if all else fails.
-		ba = JetStreamMaxStoreDefault
-	}
-	return ba
+	return JetStreamMaxStoreDefault
 }
--- a/server/pse/pse_netbsd.go
+++ b/server/pse/pse_netbsd.go
@@ -1,0 +1,36 @@
+// Copyright 2022 The NATS Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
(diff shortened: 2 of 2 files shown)


MatthiasPetermann:
Hello, thank you very much for your reply. Please also excuse my late response. I have looked again at the facts of the statfs syscall. In fact it is unfortunately the case that this is already not supported in the Golang standard library (for NetBSD). Since this would be the right place for an appropriate addition, the best choice for the nats-server at the moment is to continue with the dummy implementation, as is the case in the Windows variant. Would this be possible?

kozlovic:
> the best choice for the nats-server at the moment is to continue with the dummy implementation, as is the case in the Windows variant. Would this be possible?

So you mean as you have it in this PR, where `diskAvailable` returns `JetStreamMaxStoreDefault`, correct?

I mean, we do this for Windows (and wasm stub), so I guess this would be ok if there is no other alternative. @derekcollison what do you think?

kozlovic:
@MatthiasPetermann Ok, could you modify your PR to be made against the branch called "dev" (instead of main)? We are trying to keep changes/additions for the next release separate from the main branch. Let me know if there is any issue with that.

MatthiasPetermann:
Hello, thank you for your assessment. I have now changed the pull request to the dev branch. Please let me know if there is anything else to do. Kind regards, Matthias

server/disk_avail.go
```diff
@@ -11,8 +11,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-//go:build !windows && !openbsd && !wasm
-// +build !windows,!openbsd,!wasm
+//go:build !windows && !openbsd && !netbsd && !wasm
```
kozlovic, line 14: Since you modified this file, could you please add `-2022` to the very first line of this file (the copyright banner)?

server/disk_avail_netbsd.go
```diff
@@ -0,0 +1,22 @@
+// Copyright 2021 The NATS Authors
```
kozlovic, line 1: New file, so please change this year to `2022`.

server/pse/pse_netbsd.go
```diff
@@ -0,0 +1,36 @@
+// Copyright 2015-2018 The NATS Authors
```
kozlovic, line 1: Same, new file, so simply use `2022` for the year (instead of this range).

kozlovic (changes requested):
@MatthiasPetermann Sorry for not catching this early. Just a nit about copyright and then we will be able to merge. Also, if you don't mind, could you squash and have the commit's title be something like:
```
[ADDED] Ability to build on NetBSD
```
This will help us later when writing the release notes to easily see the commits that need to be included there. Thanks!

MatthiasPetermann:
Hello, thanks for the tips. I made the small changes and in the meantime also learned how to combine several previous commits ;-) I would be happy if you could look over the updated pull request again when you get a chance. If more changes are needed, please don't hesitate to write me. Kind regards, Matthias

kozlovic (approved):
LGTM. Thanks for the contribution!

Issue #3524, opened 2022-10-05
 nats-server missing flags in the usage message 


ariel-zilber:
### Defect
The printed usage message of server options for the nats server should describe all possilbe cli options.
Howerver by  looking in the  ConfigureOptions method in the opts.go file 
it can be seen that some options were are "hidden"

#### 1
```
fs.IntVar(&opts.MaxTracedMsgLen, "max_traced_msg_len", 0, "Maximum printable length for traced messages. 0 for unlimited.")

```
#### 2
```
fs.StringVar(&opts.Cluster.ListenStr, "cluster_listen", "", "Cluster url from which members can solicit routes.")

```
#### 3
```
fs.StringVar(&opts.PortsFileDir, "ports_file_dir", "", "Creates a ports file in the specified directory (<executable_name>_<pid>.ports).")

```
#### 4
```
fs.IntVar(&opts.ProfPort, "profile", 0, "Profiling HTTP port.")
```

#### 5
```
fs.Int64Var(&opts.LogSizeLimit, "log_size_limit", 0, "Logfile size limit being auto-rotated")
```

### action required
Please update the usage guide to support this option

### Versions of `nats-server` and affected client libraries used:
Latest commited version "37ca972251d517d3ca5a4046b647b1c3824cfeb8"

Issue #3521, opened 2022-10-05
nats-server missing third flag option for server name in the usage message


ariel-zilber:
## Defect
The printed usage message of server options for the nats server describes two methods of setting up the server Name:

```
    -n, --name <server_name>         Server name (default: auto)
```

However a third "hidden" flag possabilty exists in the opts.go file in the ConfigureOptions method:
```
	fs.StringVar(&opts.ServerName, "n", "", "Server name.")
	fs.StringVar(&opts.ServerName, "name", "", "Server name.")
	fs.StringVar(&opts.ServerName, "server_name", "", "Server name.")
```

Please either remove the third option "server_name" in the ConfigureOptions method in the opts.go file or  update the usage guide

#### Versions of `nats-server` and affected client libraries used:
Latest commited version "37ca972251d517d3ca5a4046b647b1c3824cfeb8"

kozlovic:
Closed by #3527

(Note, in fure PRs, if you add this to your commit message, the issue will auto-close when the PR is merged:
```
Resolves #<number>
```

PR #3527, opened 2022-10-06
Fix to issues  #3524 and #3521

ariel-zilber:
### description
* The usage message for nats core does not contain all the exisitng options for setting up the server
* Also some flags we missing

### Adding resolve to issues:
https://github.com/nats-io/nats-server/issues/3521
https://github.com/nats-io/nats-server/issues/3524
Updates the usage guide for nats server

### DIFF
--- a/main.go
+++ b/main.go
@@ -28,57 +28,61 @@ var usageStr = `
 Usage: nats-server [options]
 
 Server Options:
-    -a, --addr, --net <host>         Bind to host address (default: 0.0.0.0)
-    -p, --port <port>                Use port for clients (default: 4222)
-    -n, --name <server_name>         Server name (default: auto)
-    -P, --pid <file>                 File to store PID
-    -m, --http_port <port>           Use port for http monitoring
-    -ms,--https_port <port>          Use port for https monitoring
-    -c, --config <file>              Configuration file
-    -t                               Test configuration and exit
-    -sl,--signal <signal>[=<pid>]    Send signal to nats-server process (stop, quit, reopen, reload)
-                                     <pid> can be either a PID (e.g. 1) or the path to a PID file (e.g. /var/run/nats-server.pid)
-        --client_advertise <string>  Client URL to advertise to other servers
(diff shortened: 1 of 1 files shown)


main.go
```diff
@@ -28,57 +28,61 @@ var usageStr = `
...
-    -m, --http_port <port>           Use port for http monitoring
-    -ms,--https_port <port>          Use port for https monitoring
-    -c, --config <file>              Configuration file
-    -t                               Test configuration and exit
-    -sl,--signal <signal>[=<pid>]    Send signal to nats-server process (stop, quit, reopen, reload)
-                                     <pid> can be either a PID (e.g. 1) or the path to a PID file (e.g. /var/run/nats-server.pid)
-        --client_advertise <string>  Client URL to advertise to other servers
+    -a, --addr, --net <host>                        Bind to host address (default: 0.0.0.0)
+    -p, --port <port>                               Use port for clients (default: 4222)
+    -n, --name, --server_name <server_name>         Server name (default: auto)
```
kozlovic, line 33: You could have put `--server_name` below maybe so to not have to shift the description of other fields to the right?
Originally we were trying to keep the columns to 80 characters, but to be frank even in main it is already not the case for few options. Maybe we take the opportunity to fix that in that PR?

main.go
```diff
@@ -28,57 +28,61 @@ var usageStr = `
...
+        --connect_retries <number>                  For implicit routes, number of connect retries
+        --cluster_listen <url>                      Cluster url from which members can solicit routes.
 Common Options:
-    -h, --help                       Show this message
-    -v, --version                    Show version
-        --help_tls                   TLS help
+    -h, --help                                      Show this message
+    -v, --version                                   Show version
+        --help_tls                                  TLS help
+Profiling Options:
```
kozlovic, line 84: Maybe move this section after Cluster Options but before Common Options? (and use new line between sections)

kozlovic (changes requested):

kozlovic:
@ariel-zilber Do you think you are going to be able to make the changes?

PR #3530, opened 2022-10-07
Test to make sure a consumer that is deleted while a server is down recovers correctly.

derekcollison:

/cc @nats-io/core


kozlovic (approved):
LGTM

PR #3532, opened 2022-10-08
Add error and its check for collision with JS API

### LINKED ISSUE #3531: Setting up a stream over JS API subject breaks the API
## Defect

If a stream is created that covers the JS API subject (or any of its child subjects), then all the API messages covered by the stream definition are grabbed by JS and replied to with JS ACKs, effectively breaking the API and possibly making the deletion of the stream impossible.

any library client 
any NATS-server version since JetStream (tested today on 2.9.2)

any, tested today on windows 10, reproduced on linux. 

The actual bug is the break of jetstream own protocol by itself, is OS independent

try to create a stream that covers `$JS`/ the JS api prefix 

#### Expected result:

stream creation would be rejected

#### Actual result:

creation of the stream :

        nats stream create --config stream.config.json
        Stream anything was created
        
        Information for Stream anything created 2022-10-08T15:49:23+02:00
        
        Configuration:
        
                     Subjects: >
             Acknowledgements: true
                    Retention: Memory - Limits
                     Replicas: 1
               Discard Policy: Old
             Duplicate Window: 2m0s
            Allows Msg Delete: true
                 Allows Purge: true
               Allows Rollups: false
             Maximum Messages: unlimited
                Maximum Bytes: unlimited
                  Maximum Age: unlimited
            Maximum Consumers: unlimited
        
        
        State:
        
                     Messages: 0
                        Bytes: 0 B
                     FirstSeq: 0
                      LastSeq: 0
             Active Consumers: 0

then all the API messages are grabbed by JS and replied to with JS ACKs, effectively breaking the API and making the deletion of the stream impossible.

        nats stream ls
        nats: error: could not list streams: server response is not a valid "io.nats.jetstream.api.v1.stream_list_response" message: /type: value must be "io.nats.jetstream.api.v1.stream_list_response"
        missing properties: 'error'
        /streams: expected array, but got null, try --help

nats -DV output not really useful but attached [output.log](https://github.com/nats-io/nats-server/files/9739367/output.log)
stream config file (zipped to pass github's rules ) [stream.config.json.zip](https://github.com/nats-io/nats-server/files/9739385/stream.config.json.zip)

### LINKED ISSUE #3551: Unable to concurrently update a streams config.
## Defect

When trying to concurrently update a streams config not all updates are being honored. 

2.9.3

Mac

Using the nats npm package

```
async function updateStream(stream, new_subject, jsm) {
	const si = await jsm.streams.info(stream);
	si.config.subjects.push(new_subject);
	await jsm.streams.update(stream, si.config);
}

async function update() {
	const jsm = await nc.jetstreamManager();

	const stream = 'mystream';
	const subj = 'mystream.*';
	await jsm.streams.add({ name: stream, subjects: [subj] });

	let update_promises = [];
	for (let x = 0; x < 3; x++) {
		update_promises.push(updateStream(stream, x.toString(), jsm));
	}
        
        await Promise.all(update_promises);
	const si = await jsm.streams.info(stream);
	console.log(si.config.subjects);
}

update.then(() =>{}).catch((err) => {console.error(err)})
```

#### Expected result:

I expect all the subjects to be in the streams config
si.config.subjects to equal `[ "mystream.*", "0", "1", "2" ]`

#### Actual result:

si.config.subjects equals `[ "mystream.*", "2" ]`

AlbericC:
 - [x] Tests added

Resolves #3531

 - check for collision with JS API when account creates a stream
 - New error for this case

/cc @nats-io/core

### DIFF
--- a/server/stream.go
+++ b/server/stream.go
@@ -1147,6 +1147,12 @@ func (s *Server) checkStreamCfg(config *StreamConfig, acc *Account) (StreamConfi
 		}
 	}
 
+	for _, subj := range cfg.Subjects {
+		if SubjectsCollide(subj, JSApiPrefix) {
+			return StreamConfig{}, NewJSStreamApiPrefixOverlapError(JSApiPrefix, subj)
+		}
+	}
+
 	// cycle check for source cycle
 	toVisit := []*StreamConfig{&cfg}
 	visited := make(map[string]struct{})
--- a/server/stream.go
+++ b/server/stream.go
@@ -1205,8 +1205,10 @@ func (s *Server) checkStreamCfg(config *StreamConfig, acc *Account) (StreamConfi
 				return StreamConfig{}, NewJSStreamInvalidConfigError(fmt.Errorf("duplicate subjects detected"))
 			}
 			// Also check to make sure we do not overlap with our $JS API subjects.
-			if subjectIsSubsetMatch(subj, "$JS.API.>") {
-				return StreamConfig{}, NewJSStreamInvalidConfigError(fmt.Errorf("subjects overlap with jetstream api"))
+			if !cfg.NoAck { // unless Ack is disabled
+				if subjectIsSubsetMatch(subj, "$JS.API.>") {
+					return StreamConfig{}, NewJSStreamInvalidConfigError(fmt.Errorf("subjects overlap with jetstream api"))
+				}
 			}
 			// Make sure the subject is valid.
 			if !IsValidSubject(subj) {
--- a/server/stream.go
+++ b/server/stream.go
@@ -1147,9 +1147,12 @@ func (s *Server) checkStreamCfg(config *StreamConfig, acc *Account) (StreamConfi
 		}
 	}
 
-	for _, subj := range cfg.Subjects {
-		if SubjectsCollide(subj, JSApiPrefix) {
-			return StreamConfig{}, NewJSStreamApiPrefixOverlapError(JSApiPrefix, subj)
+	// if Ack is on, forbid a creating stream that would respond Acks in lieu of the API.
+	if !cfg.NoAck {
+		for _, subj := range cfg.Subjects {
+			if SubjectsCollide(subj, JSApiPrefix) {
+				return StreamConfig{}, NewJSStreamApiPrefixOverlapError(JSApiPrefix, subj)
+			}
 		}
 	}
 
--- a/server/errors.json
+++ b/server/errors.json
@@ -1313,7 +1313,7 @@
     "constant": "JSStreamApiPrefixOverlapError",
     "code": 400,
     "error_code": 10133,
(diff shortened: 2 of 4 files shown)


server/stream.go
```diff
@@ -1147,6 +1147,12 @@ func (s *Server) checkStreamCfg(config *StreamConfig, acc *Account) (StreamConfi
 		}
 	}
 
+	for _, subj := range cfg.Subjects {
+		if SubjectsCollide(subj, JSApiPrefix) {
```
ripienaar, line 1151: This will help of course - we planned to add this before but with cross accounts, domains etc it’s a mine field. And of course many other subjects are problematic also. 

We should for sure also stop just “>” for example also

ripienaar: I guess this sound stop > also though as it would over lap

AlbericC: that was my first thought also, but with (possibly) exported JetStream services, I wanted to target the JS API  more specifically. 

Although '>' sure poses a problem every time, I can imagine a (twisted) use-case where one would want to use a stream to keep an history of `$JS.EVENT.ADVISORY.>` for instance.

ripienaar: Yeah so we definitely shouldn’t stop advisories from being infested that’s a good point. 

And to be pedantic you CAN ingest > you just need to turn off ack on the stream.

AlbericC:
I made this pull request kind of quick, do not hesitate to reject it if you want me to do it again in a cleaner way (branched, rebased, squashed).

server/stream.go
ripienaar: I do seem to recall some users having a stream on the API subject without acks for audit purposes.

AlbericC: Pretty good point I'll take that into account and correct my code accordingly

AlbericC: for the record a test exists to ensure events, etc are OK https://github.com/nats-io/nats-server/blob/cb086bcef9e5d9f051c74319ef63c2412de093e8/server/jetstream_test.go#L1085

AlbericC: > I guess this should stop > also though as it would over lap

yes, it does. Included a test case for it to make that explicit

kozlovic (changes requested):
Since I would consider this a breaking change (since possibly existing streams would fail to be recovered, etc..) should this go only on the next 2.10.0 release? I have created the `dev` branch today that will hold the changes in `main` (the patches) and all PRs that are targeted for the next release. If we all agree that this is possibly breaking and should go in 2.10.0, could you alter your PR to be against the `dev` branch instead of `main`?

bruth:
An undertone of this PR is the need to extend the authorization model transparently for JetStream primitives. That said, there may be an incremental improvement here as an opt-in (non-breaking) account limit that could guard against creating streams that bind to the JS API.

wallyqs:
Addressed via https://github.com/nats-io/nats-server/pull/5548

PR #3537, opened 2022-10-10
[FIXED] JetStream: User-defined ephemeral Name not used in cluster mode

kozlovic:
If the user sends a CONSUMER.CREATE request with a configuration that specifies the name that the user wants for the ephemeral consumer, this would not work on cluster mode, that is, the server would still pick a name instead of using the provided one.


derekcollison (approved):
LGTM

PR #3539, opened 2022-10-10
[FIXED] JetStream: return error on negative replicas count

kozlovic:
If a stream is created or updated with a negative replicas count, and error is now returned. Same for consumers.


derekcollison (approved):
LGTM

caleblloyd (approved):
LGTM

PR #3552, opened 2022-10-13
[FIXED] MQTT: Subjects mapping were not handled

### LINKED ISSUE #3547: Subject mapping doesn't work with MQTT

```
[53] 2022/10/12 12:53:59.709111 [INF] Starting nats-server
[53] 2022/10/12 12:53:59.709136 [INF]   Version:  2.9.1
[53] 2022/10/12 12:53:59.709145 [INF]   Git:      [2363a2c]
[53] 2022/10/12 12:53:59.709151 [DBG]   Go build: go1.19.1
```

Ubuntu 22.04.1 LTS

I'm trying to use subject mapping with MQTT and the mapping doesn't work.

My nats server is running with this config:
```
listen: 0.0.0.0:4222
http: 0.0.0.0:8222

server_name: "mqtt_server"

jetstream {
    store_dir=/var/lib/nats/data
    max_file=10Gi
}

mappings = {
  foo: bar
}

mqtt {
   port: 1883
}
```

I subscribe on the topic `bar`:
```sh
docker run --rm --net host hivemq/mqtt-cli sub --topic bar -V 3
```

In another terminal I publish a message to `foo`:
```sh
docker run --rm --net host hivemq/mqtt-cli pub --topic foo -m "test" -V 3
```

#### Expected result:

I expect the message "test" to be received on the topic `bar`.

#### Actual result:

The message is not received on `bar`. I can see the message if I subscribe on `foo` instead but it means there is no mapping.

kozlovic:
A simple configuration like this:
```
...
mappings = {
  foo: bar
}

mqtt {
   port: 1883
}
```
would cause an MQTT subscription on "bar" to not receive messages published on "foo".

In otherwords, the subject transformation was not done when parsing a PUBLISH packet.

This PR also handles the case of service imports where transformation occurs after the initial publish parsing.

Resolves #3547

### DIFF
--- a/server/client.go
+++ b/server/client.go
@@ -3927,7 +3927,8 @@ func (c *client) processServiceImport(si *serviceImport, acc *Account, msg []byt
 	// Now check to see if this account has mappings that could affect the service import.
 	// Can't use non-locked trick like in processInboundClientMsg, so just call into selectMappedSubject
 	// so we only lock once.
-	if nsubj, changed := si.acc.selectMappedSubject(to); changed {
+	nsubj, changed := si.acc.selectMappedSubject(to)
+	if changed {
 		c.pa.mapped = []byte(to)
 		to = nsubj
 	}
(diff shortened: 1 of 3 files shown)


derekcollison (approved):
LGTM

PR #3558, opened 2022-10-14
Fix setting `InactivityThreshold` on existing durable consumer

neilalexander:
Beforehand, the threshold and timer weren't resetting properly when an existing durable consumer was updated with a new `InactivityThreshold` so consumers wouldn't get cleaned up until the NATS Server was restarted. This PR refactors that to use a common function both when the consumer is set up and when it is updated, which effectively fixes the bug. It also updates the unit tests to verify that it works both ways.

 - [X] Tests added

/cc @nats-io/core

### DIFF
--- a/server/consumer.go
+++ b/server/consumer.go
@@ -786,15 +786,7 @@ func (mset *stream) addConsumerWithAssignment(config *ConsumerConfig, oname stri
 	o.nextMsgSubj = fmt.Sprintf(JSApiRequestNextT, mn, o.name)
 
 	// If the user has set the inactive threshold, set that up here.
-	if o.cfg.InactiveThreshold > 0 {
-		o.dthresh = o.cfg.InactiveThreshold
-	} else if !o.isDurable() {
-		// Ephemerals will always have inactive thresholds.
-		// Add in 1 sec of jitter above and beyond the default of 5s.
-		o.dthresh = JsDeleteWaitTimeDefault + time.Duration(rand.Int63n(1000))*time.Millisecond
-		// Only stamp config with default sans jitter.
-		o.cfg.InactiveThreshold = JsDeleteWaitTimeDefault
-	}
+	o.cfg.InactiveThreshold = o.updateInactiveThresholdLocked(o.cfg.InactiveThreshold, false)
 
 	if o.isPushMode() {
 		if !o.isDurable() {
@@ -1523,11 +1515,41 @@ func (o *consumer) updateConfig(cfg *ConsumerConfig) error {
 		o.maxdc = uint64(cfg.MaxDeliver)
 	}
 
+	// Set InactiveThreshold if changed.
+	if cfg.InactiveThreshold != o.cfg.InactiveThreshold {
+		cfg.InactiveThreshold = o.updateInactiveThresholdLocked(cfg.InactiveThreshold, true)
+	}
+
 	// Record new config for others that do not need special handling.
 	// Allowed but considered no-op, [Description, SampleFrequency, MaxWaiting, HeadersOnly]
 	o.cfg = *cfg
 
 	return nil
+}
+
+// updateInactiveThresholdLocked sets the dthres and either resets the
+// dtmr (if there is one) or forcibly starts one, depending on the value
+// of the "forceStart" parameter. It then returns what the final value
+// for InactiveThreshold will be (so that the jittery default for ephemeral
+// consumers can be captured in config).
+func (o *consumer) updateInactiveThresholdLocked(threshold time.Duration, forceStart bool) time.Duration {
+	started := o.dtmr != nil
+	defer func() {
+		if forceStart || started {
+			stopAndClearTimer(&o.dtmr)
+			o.dtmr = time.AfterFunc(o.dthresh, func() { o.deleteNotActive() })
+		}
+	}()
+	if threshold > 0 {
+		o.dthresh = threshold
+	} else if !o.isDurable() {
+		// Ephemerals will always have inactive thresholds.
+		// Add in 1 sec of jitter above and beyond the default of 5s.
+		o.dthresh = JsDeleteWaitTimeDefault + time.Duration(rand.Int63n(1000))*time.Millisecond
+		// Only stamp config with default sans jitter.
+		return JsDeleteWaitTimeDefault
+	}
+	return threshold
 }
 
 // This is a config change for the delivery subject for a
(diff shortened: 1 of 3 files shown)


server/consumer.go
```diff
@@ -1523,13 +1515,43 @@ func (o *consumer) updateConfig(cfg *ConsumerConfig) error {
...
+	}()
+	if threshold > 0 {
+		o.dthresh = threshold
+	} else if !o.isDurable() {
+		// Ephemerals will always have inactive thresholds.
+		// Add in 1 sec of jitter above and beyond the default of 5s.
+		o.dthresh = JsDeleteWaitTimeDefault + time.Duration(rand.Int63n(1000))*time.Millisecond
+		// Only stamp config with default sans jitter.
+		return JsDeleteWaitTimeDefault
+	}
```
kozlovic, line 1551: So what if the user uses a durable consumer and had a threshold and now wants to set it to 0. That would never been reset to 0. If we support that case, then we would need to set it here and also change the defer func() to not reset the o.dtmr if threshold is 0.

kozlovic: Also, as it stands, any update to the consumer reset its inactivity timer, which may be ok, but that will delay the cleanup (even if there is no interest or no pull made for pull consumers). I think it is "ok" because if the consumer is updated it is likely that one wants the consumer to stick around, but let's make sure we understand the consequences.

kozlovic (commented):
I have some comments (mainly regarding if we want to support resetting the InactiveThreshold to 0 for a durable) and would like @derekcollison to chime in too before we approve.

server/consumer.go
neilalexander: So I think setting it to `0` is an easy enough fix and it would seem like it's reasonable to reset the timer and not restart it if that's the case.

Modifying the `defer` is more complicated. I did have different behaviour there originally but most permutations seem to break the `TestJetStreamSuperClusterEphemeralCleanup` test. I'm not 100% sure I understand yet what that test does or why we would expect a timer to not be running there.

server/consumer.go
```diff
@@ -1532,24 +1532,34 @@ func (o *consumer) updateConfig(cfg *ConsumerConfig) error {
...
+		if o.dthresh > 0 {
 			o.dtmr = time.AfterFunc(o.dthresh, func() { o.deleteNotActive() })
 		}
 	}()
-	if threshold > 0 {
+	if o.isDurable() || threshold > 0 {
+		// Allow setting a durable consumer threshold to either a specific
+		// value or to 0 (effectively disabling it). Also allow setting an
+		// ephemeral consumer to a non-zero value.
 		o.dthresh = threshold
```
kozlovic, line 1553: I was not asking to make the changes yet, since I wanted @derekcollison input to see if this is something we want to support (reset durable to 0).

If we do and you want to keep this code, you now have for the isDurable() case, make sure that threshold has not been given as a negative value. So anything < 0 should become 0 I guess.

kozlovic (changes requested):

server/consumer.go
neilalexander: I'll revert for now but leave the commits on the branch rather than rebasing.

server/consumer.go
kozlovic: It broke because we used to change the threshold value but with existing code the timer was known to not be set yet. With your changes (and calling that new function), it would be started as right there (in previous commit you were calling with forceStart==false, so were not starting it).

It could be that the test needs updating if the timer *should* have been started. Sometimes - unfortunately - tests are written with the assumption that the current code behaves as it should :-)

server/consumer.go
```diff
@@ -1532,24 +1532,34 @@ func (o *consumer) updateConfig(cfg *ConsumerConfig) error {
...
-func (o *consumer) updateInactiveThresholdLocked(threshold time.Duration, forceStart bool) time.Duration {
-	started := o.dtmr != nil
+func (o *consumer) updateInactiveThresholdLocked(threshold time.Duration) time.Duration {
+	origThresh := o.dthresh
 	defer func() {
-		if forceStart || started {
-			stopAndClearTimer(&o.dtmr)
+		if origThresh == o.dthresh {
+			// Nothing has changed, so don't do anything.
+			return
```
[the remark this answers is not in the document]
kozlovic: Agreed.

server/consumer.go
neilalexander: I think actually I might have figured out a solution. I've left the setting durable to `0` in but it appears to work and pass either way with/without that, so easy enough to revert if there's no appetite to support that.

neilalexander: Bah. Spoke too soon.

server/consumer.go
derekcollison: It sounds like might be reasonable to set back to 0, although I want in general for systems to avoid that since we have issues with abandoned durables taking up resources in systems.

server/consumer.go
```diff
@@ -786,15 +786,7 @@ func (mset *stream) addConsumerWithAssignment(config *ConsumerConfig, oname stri
...
-	if o.cfg.InactiveThreshold > 0 {
-		o.dthresh = o.cfg.InactiveThreshold
-	} else if !o.isDurable() {
-		// Ephemerals will always have inactive thresholds.
-		// Add in 1 sec of jitter above and beyond the default of 5s.
-		o.dthresh = JsDeleteWaitTimeDefault + time.Duration(rand.Int63n(1000))*time.Millisecond
-		// Only stamp config with default sans jitter.
-		o.cfg.InactiveThreshold = JsDeleteWaitTimeDefault
-	}
+	o.cfg.InactiveThreshold = o.updateInactiveThresholdLocked(o.cfg.InactiveThreshold, false)
```
kozlovic, line 789: I have my doubts creating the timer from here. It looks to me that this function (or addConsumer) may be invoked from non consumer leader servers. Could it then lead a server from deleting the consumer because it thinks that there is no interest/pull requests? I need to run some tests...

kozlovic:
@neilalexander FYI: I have found an issue with update consumer (where server does not send a response). Will send a PR soon. I will also then send another PR that build on yours but a bit different (my suspicion was correct that your approach would likely start the timer in any node).

server/consumer.go
neilalexander: I did wonder about that, I assumed there was probably a reason for the unit test specifically testing that the timer wasn't already set but I don't really know enough about the cluster behaviour at this point.

neilalexander:
> @neilalexander FYI: I have found an issue with update consumer (where server does not send a response). Will send a PR soon. I will also then send another PR that build on yours but a bit different (my suspicion was correct that your approach would likely start the timer in any node).

No problem, I did wonder whether that might be undesirable.

kozlovic:
@neilalexander Closing this in favor of #3575. Thank you for the contribution (you will be mentioned in the release notes for the PR I made).

PR #3560, opened 2022-10-14
fix: Prevent nil map assignment panic if disk is full

fantashley:
If the server is unable to write to disk, there is a failure mode that results in the fss map being nil when writing message block data to it, resulting in a panic. This fix checks whether an error is returned by the function that loads the fss map, and if so, returns an error to prevent future assignment.

This issue was brought up in matrix-org/dendrite#2796. I'm not sure if it's appropriate to return early if an error is encountered, or to simply allocate the map and continue with the function. Or perhaps I could add an explicit check for whether the error returned is an "out of space" error and only return in that instance. Please let me know if there's a better alternative, thanks!

 - [ ] Tests added

 - Fix panic that occurs when disk is full

/cc @nats-io/core

### DIFF
--- a/server/filestore.go
+++ b/server/filestore.go
@@ -3160,7 +3160,9 @@ func (mb *msgBlock) writeMsgRecord(rl, seq uint64, subj string, mhdr, msg []byte
 
 	// Check if we are tracking per subject for our simple state.
 	if len(subj) > 0 && !mb.noTrack {
-		mb.ensurePerSubjectInfoLoaded()
+		if err := mb.ensurePerSubjectInfoLoaded(); err != nil {
+			return fmt.Errorf("error loading fss: %w", err)
+		}
 		if ss := mb.fss[subj]; ss != nil {
 			ss.Msgs++
 			ss.Last = seq


server/filestore.go
```diff
@@ -3160,7 +3160,9 @@ func (mb *msgBlock) writeMsgRecord(rl, seq uint64, subj string, mhdr, msg []byte
 
 	// Check if we are tracking per subject for our simple state.
 	if len(subj) > 0 && !mb.noTrack {
-		mb.ensurePerSubjectInfoLoaded()
+		if err := mb.ensurePerSubjectInfoLoaded(); err != nil {
```
derekcollison, line 3163: Looking at the code, but might not be looking hard enough, if there is an error when calling into `loadPerSubjectInfo()` that is called from `readSubjectInfo()` if we have an error we regenerate the subject state on the fly.

Were you encountering a write error from inside of `generateSubjectInfo()`? Or could it have been that the msg couht was zero and we did not create the fss on line 5103?

derekcollison (commented):
Want to dig in a bit more to really understand the failure scenario here.

Thanks though for the PR! Some questions inline.

server/filestore.go
fantashley: Thanks for taking a look! The only way I found that this could happen is if it follows the progression of function calls below, with a failure encountered on the last one while trying to write. I don't believe it's because the message count is 0, as that's accounted for [here](https://github.com/nats-io/nats-server/blob/19ff7f5f7addf7ce59463f899e4ee0a8d1d946bf/server/filestore.go#L5185).
`ensurePerSubjectInfoLoaded()` -> `readPerSubjectInfo()` -> `generatePerSubjectInfo()` -> `loadMsgsWithLock()` -> `mb.mfd.WriteAt(buf, woff)`

derekcollison: So this would be the last block and be actively being written to, and this call would fail yes? That is the theory?

https://github.com/nats-io/nats-server/blob/19ff7f5f7addf7ce59463f899e4ee0a8d1d946bf/server/filestore.go#L3488

fantashley: Correct! The other paths I traced through the code resulted in `mb.fss` being allocated at some point.

fantashley: Here's the snippet of logs from the linked issue that I was working off of
```
Oct 14 13:43:41 dendrite-monolith-server[6356]: panic: assignment to entry in nil map
Oct 14 13:43:41 dendrite-monolith-server[6356]: goroutine 252 [running]:
Oct 14 13:43:41 dendrite-monolith-server[6356]: github.com/nats-io/nats-server/v2/server.(*msgBlock).writeMsgRecord(0xc00e03aea0, 0xdc4, 0xb6fe, {0xc00a35e1c0, 0x35}, {0xc00542506b, 0x35, 0xf95}, {0xc0054250a0, 0xd38, ...}, ...)
Oct 14 13:43:41 dendrite-monolith-server[6356]:         github.com/nats-io/nats-server/v2@v2.9.2/server/filestore.go:3168 +0x8cf
Oct 14 13:43:41 dendrite-monolith-server[6356]: github.com/nats-io/nats-server/v2/server.(*fileStore).writeMsgRecord(0xc0000cea00, 0x0?, 0x0?, {0xc00a35e1c0, 0x35}, {0xc00542506b, 0x35, 0xf95}, {0xc0054250a0, 0xd38, ...})
Oct 14 13:43:41 dendrite-monolith-server[6356]:         github.com/nats-io/nats-server/v2@v2.9.2/server/filestore.go:3308 +0x1f2
Oct 14 13:43:41 dendrite-monolith-server[6356]: github.com/nats-io/nats-server/v2/server.(*fileStore).storeRawMsg(0xc0000cea00, {0xc00a35e1c0, 0x35}, {0xc00542506b, 0x35, 0xf95}, {0xc0054250a0, 0xd38, 0xf60}, 0xb6fe, ...)
Oct 14 13:43:41 dendrite-monolith-server[6356]:         github.com/nats-io/nats-server/v2@v2.9.2/server/filestore.go:1911 +0x385
```

derekcollison: Nice work!

Your suggested fix seems reasonable, however what seems to be happening in this case is we are re-entering the flush pending blocks because we are about to flush the current block. The original filestore was async, but we moved to suppress that behavior by default (its still possible and will revisit one day). So flush should really never be called but here we can see that we are doing that even thought we should not (if I am reading the code correctly).

So I am going to take a deeper look here before signing off, not that this is not correct but I think this might highlight something that we could optimize as well.

fantashley: Ah, that makes sense, thanks for the explanation!

derekcollison: Your change is good, just want to take a peek at something with fresh eyes, just returned back from Korea this morning ;)

derekcollison (approved):
LGTM - Thanks Again!

Will do other fix in separate PR.

PR #3561, opened 2022-10-14
[FIXED] JetStream: User given named ephemeral lost after migration

### LINKED ISSUE #3550: Named consumer recreated with generated name after Consumer Leader down.
## Defect

When the server with the Consumer Leader is taken down, the server recreates the consumer on another replica, but does so with a generated name instead of the given name. All other settings are remembered. 

This causes subsequent pull requests to fail because the original consumer name is used in the request.

`v2.9.3`

Windows

- Start a 5 node cluster
- Create a stream with file storage and replication 3 
- Publish 100 messages
- Consuming a message using pull with batch size of 1
  - a named ephemeral pull consumer
  - inactive threshold of 1 hour
  - default [consumer] replicas (so 3)
- do `nats c info` or look at the server output to determine the Consumer Leader
- kill the consumer leader server

#### Expected result:
The consumer is recreated on one of the replicas with the user given name.

#### Actual result:
The consumer is recreated on one of the replicas with a generated name.

#### Consumer Info after creation 
```
? Select a Stream ConStream
? Select a Consumer consumer1665667190256
Information for Consumer ConStream > consumer1665667190256 created 2022-10-13T09:19:50-04:00

Configuration:

           Pull Mode: true
      Filter Subject: ConSubject
      Deliver Policy: All
          Ack Policy: Explicit
            Ack Wait: 30s
       Replay Policy: Instant
     Max Ack Pending: 1,000
   Max Waiting Pulls: 512
  Inactive Threshold: 1h0m0s

Cluster Information:

                Name: cluster
              Leader: server6

State:

   Last Delivered Message: Consumer sequence: 1 Stream sequence: 1 Last delivery: 6.29s ago
     Acknowledgment floor: Consumer sequence: 1 Stream sequence: 1 Last Ack: 6.28s ago
         Outstanding Acks: 0 out of maximum 1,000
     Redelivered Messages: 0
     Unprocessed Messages: 99
            Waiting Pulls: 0 of maximum 512
```

#### Consumer after server6 is killed:
```
C:\nats>nats c info
? Select a Stream ConStream
? Select a Consumer wlpY49ui
Information for Consumer ConStream > wlpY49ui created 2022-10-13T09:19:50-04:00

Configuration:

           Pull Mode: true
      Filter Subject: ConSubject
      Deliver Policy: All
          Ack Policy: Explicit
            Ack Wait: 30s
       Replay Policy: Instant
     Max Ack Pending: 1,000
   Max Waiting Pulls: 512
  Inactive Threshold: 1h0m0s

Cluster Information:

                Name: cluster
              Leader: server8

State:

   Last Delivered Message: Consumer sequence: 1 Stream sequence: 1
     Acknowledgment floor: Consumer sequence: 1 Stream sequence: 1
         Outstanding Acks: 0 out of maximum 1,000
     Redelivered Messages: 0
     Unprocessed Messages: 99
            Waiting Pulls: 0 of maximum 512
```

kozlovic:
If an ephemeral was given a name by the user, if the consumer leader was then shutdown, the ephemeral would be migrated using a server generated new name instead of keeping the user given name.

Also, in some cases the migration would not even occur. This was likely due to the fact that RAFT node(s) were shutdown prior to the ephemeral migration code was invoked.

Resolves #3550

### DIFF
--- a/server/jetstream.go
+++ b/server/jetstream.go
@@ -873,7 +873,9 @@ func (s *Server) migrateEphemerals() {
 					cc.meta.ForwardProposal(encodeDeleteConsumerAssignment(ca))
 					// Encode state and new name.
 					ca.State = state
-					ca.Name = createConsumerName()
+					if ca.Config.Name == _EMPTY_ {
+						ca.Name = createConsumerName()
+					}
 					addEntry := encodeAddConsumerAssignmentCompressed(ca)
 					cc.meta.ForwardProposal(addEntry)
 					js.mu.Unlock()
--- a/server/jetstream_cluster_3_test.go
+++ b/server/jetstream_cluster_3_test.go
@@ -520,7 +520,7 @@ func TestJetStreamClusterNegativeReplicas(t *testing.T) {
 	t.Run("Clustered", func(t *testing.T) { testBadReplicas(t, c.randomServer(), "TEST2") })
 }
 
-func TestJetStreamClusterUserSelectedConsName(t *testing.T) {
+func TestJetStreamClusterUserGivenConsName(t *testing.T) {
 	s := RunBasicJetStreamServer()
 	if config := s.JetStreamConfig(); config != nil {
 		defer removeDir(t, config.StoreDir)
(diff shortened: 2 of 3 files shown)


derekcollison (approved):
LGTM

PR #3562, opened 2022-10-16
Reduced default logging for JetStream setup over a leafnode.

derekcollison:
Also added in specific tests for using non system extended setup similar to NGS.

/cc @nats-io/core


kozlovic:
@derekcollison Fine with the change, but this may be one that maybe we should keep for the "dev" branch? This is not really a fix, but possibly a change since we go from Notice to Debug. I am fairly certain that there would be no one affected by that if it was going to main (in that I don't think users would scan their log for those specific entries), but that could be practice for the new release process.

derekcollison:
ok so clear this one and submit against dev? I guess I need to change my whole setup to always use dev branch now ;)

kozlovic:
If you can't change from the existing PR, you could indeed close this one and in the GUI, when opening the PR select "dev" branch instead of "main".

derekcollison:
ok

PR #3564, opened 2022-10-17
Reduced default logging for JetStream setup over a leafnode.

derekcollison:
Also added in specific tests for using non system extended setup similar to NGS.

/cc @nats-io/core

### DIFF
--- a/server/jetstream_helpers_test.go
+++ b/server/jetstream_helpers_test.go
@@ -890,6 +890,13 @@ var jsLeafFrag = `
 		]
 	}
 `
+var jsLeafNoSysFrag = `
+	leaf {
+		remotes [
+			{ urls: [ %s ] }
+		]
+	}
+`
 
 func (c *cluster) createLeafNodes(clusterName string, numServers int, domain string) *cluster {
 	c.t.Helper()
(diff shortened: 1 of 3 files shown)


server/jetstream_helpers_test.go
```diff
@@ -890,6 +890,13 @@ var jsLeafFrag = `
 		]
 	}
 `
+var jsLeafNoSysFrag = `
```
kozlovic, line 893: Sorry about that. I just realized that if you think we may need those facilities in main before we release 2.10.0, then we should have the test part of this PR made against main (since it is just tests) and leave the PR with just the logging change. I would be happy to do all that for you (since you already had to change the PR from main to dev). Just let me know.

derekcollison: I think my game plan is to just work always in `dev` branch from now on. So its ok.

kozlovic: But that's not the plan though. If it is a fix that would go into a patch, it would need to be done directly toward main. I don't want to have to cherry pick commits from dev into main. If I do what I said above (make the test changes in PR toward main), I will then merge main into dev (meaning dev is up-to-date with main), so even dev would benefit from the tests changes. Again, happy to do all that for you.

derekcollison: ok, but do you consider tests with no fix a patch here? I am happy to follow your lead.

derekcollison (commented):
LGTM

kozlovic (approved):
LGTM

PR #3565, opened 2022-10-17
Added specific tests for using non system extended setup similar to NGS

kozlovic:
Signed-off-by: Derek Collison derek@nats.io


derekcollison (approved):
LGTM

PR #3568, opened 2022-10-18
Make sure to rebuild per subject info on write error

derekcollison:
We could potentially fail a write and not rebuild the state since we tried to ensurePerSubjectInfo was loaded after we set the flush pending state. So if it had to be rebuilt it would try to flushPending first.

This was referenced in #3560 

So we moved this above setting the pending state, and also made sure the rebuildState also took into account the per subject info as well.

/cc @nats-io/core

### DIFF
--- a/server/filestore.go
+++ b/server/filestore.go
@@ -168,6 +168,9 @@ type msgBlock struct {
 	flusher bool
 	noTrack bool
 	closed  bool
+
+	// Used to mock write failures.
+	mockWriteErr bool
 }
 
 // Write through caching layer that is also used on loading messages.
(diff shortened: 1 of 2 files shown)


server/filestore.go
```diff
@@ -3095,6 +3098,21 @@ func (mb *msgBlock) writeMsgRecord(rl, seq uint64, subj string, mhdr, msg []byte
...
+	// Check if we are tracking per subject for our simple state.
+	// Do this before changing the cache that would trigger a flush pending msgs call
+	// if we needed to regenerate the per subject info.
+	if len(subj) > 0 && !mb.noTrack {
+		if err := mb.ensurePerSubjectInfoLoaded(); err != nil {
+			return err
+		}
+		if ss := mb.fss[subj]; ss != nil {
+			ss.Msgs++
+			ss.Last = seq
```
kozlovic, line 3110: Do we need to check that `seq >= ss.Last` before doing the assignment or are we sure that it always be that writeMsgRecord is called with the "last" sequence?

server/filestore.go
```diff
@@ -3551,9 +3567,6 @@ func (mb *msgBlock) flushPendingMsgsLocked() (*LostStreamData, error) {
 		}
 	}
 
-	// set write err to any error.
-	mb.werr = err
```
kozlovic, line 3555: This original code would have cleared `mb.werr` if this flush was successful (err would be nil). Is that something we wanted? If so, this change would break that, no?

kozlovic (commented):
Asking for some clarifications before approving.

server/filestore.go
derekcollison: It is always called with the last sequence, or next to be more exact, and is locked.

server/filestore.go
derekcollison: Good catch, will fix.

kozlovic (approved):
LGTM pending travis run.

Issue #3571, opened 2022-10-21
Router connection closed: Slow Consumer And no longer reconnect #48


wubumihuo:
We use nats -v2.1.7，and have three nodes in cluster： 168.63.70.231/233/235
the 231‘s config is :
```
# Monitoring and Tracing
port: 4222
http_port: 8222
# limit
max_pending: 64Mb
write_deadline: 5s
# cluster options
cluster {
  listen: 168.63.70.231:4248
  routes = [
  ]
}
```
and 235's config is ：
```
# Monitoring and Tracing
port: 4222
http_port: 8222

# limit
max_pending: 64Mb
write_deadline: 5s

# cluster options
cluster {
  listen: 168.63.70.235:4248
  routes = [
    nats://168.63.70.231:4248
  ]
}
```
On 2022/10/18 begin at 15:03 ,we found on 168.63.70.231,there are frequent slow consumer on 168.63.70.235 ,the log in below:
```
[99971] 2022/10/18 14:59:53.531838 [DBG] 168.63.70.235:4248 - rid:303 - Router Ping Timer
[99971] 2022/10/18 14:59:53.531868 [DBG] 168.63.70.235:4248 - rid:303 - Delaying PING due to client activity 0s ago
[99971] 2022/10/18 15:01:53.532037 [DBG] 168.63.70.235:4248 - rid:303 - Router Ping Timer
[99971] 2022/10/18 15:01:53.532065 [DBG] 168.63.70.235:4248 - rid:303 - Delaying PING due to client activity 0s ago
[99971] 2022/10/18 15:02:35.853998 [INF] 168.63.70.235:4248 - rid:303 - Slow Consumer Detected: WriteDeadline of 5s exceeded with 197 chunks of 12863290 total bytes.
[99971] 2022/10/18 15:02:40.854214 [INF] 168.63.70.235:4248 - rid:303 - Slow Consumer Detected: WriteDeadline of 5s exceeded with 344 chunks of 22340957 total bytes.
[99971] 2022/10/18 15:02:45.854474 [INF] 168.63.70.235:4248 - rid:303 - Slow Consumer Detected: WriteDeadline of 5s exceeded with 448 chunks of 29152757 total bytes.
[99971] 2022/10/18 15:02:50.854702 [INF] 168.63.70.235:4248 - rid:303 - Slow Consumer Detected: WriteDeadline of 5s exceeded with 518 chunks of 33696862 total bytes.
[...]

PR #3573, opened 2022-10-24
[FIXED] Solicited route may not retry to reconnect

kozlovic:
Originally, only solicited routes were retried in case of a disconnect, but that was before gossip protocol was introduced. Since then, two servers that connect to each other due to gossip should retry to reconnect if the connection breaks, even if the route is not explicit. However, server will retry only once or more accurately, ConnectRetries+1.

This PR solves the issue that the reconnect attempt was not initiated for a "solicited route" that was not explicit.

Maybe related to #3571

### DIFF
--- a/server/client.go
+++ b/server/client.go
@@ -4937,7 +4937,10 @@ func (c *client) reconnect() {
 		return
 	}
 	if c.route != nil {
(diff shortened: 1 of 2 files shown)


derekcollison (approved):
LGTM

PR #3575, opened 2022-10-25
[FIXED] JetStream: InactivityThreshold updates not always working

kozlovic:
This is based of @neilalexander PR #3558.

It ensures that the timer is reset/canceled on configuration update (by the leader only).

Fixed also the issue with a super-cluster where the delete timer would always be reset at every gateway interval check.


derekcollison (approved):
LGTM

PR #3576, opened 2022-10-25
Removed ephemeral consumer migration.

derekcollison:
We still will signal R1 pull consumers with pending requests.
Also fixed a bug that would not honor replicas for new style named consumers.

/cc @nats-io/core

### DIFF
--- a/server/jetstream.go
+++ b/server/jetstream.go
@@ -811,81 +811,41 @@ func (s *Server) JetStreamEnabledForDomain() bool {
 	return jsFound
 }
 
-// Will migrate off ephemerals if possible.
-// This means parent stream needs to be replicated.
-func (s *Server) migrateEphemerals() {
-	js, cc := s.getJetStreamCluster()
-	// Make sure JetStream is enabled and we are clustered.
-	if js == nil || cc == nil {
+// Will signal that all pull requests for consumers on this server are now invalid.
+func (s *Server) signalPullConsumers() {
+	js := s.getJetStream()
+	if js == nil {
 		return
 	}
 
-	var consumers []*consumerAssignment
-
-	js.mu.Lock()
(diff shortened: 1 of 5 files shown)


server/jetstream.go
```diff
@@ -811,81 +811,41 @@ func (s *Server) JetStreamEnabledForDomain() bool {
...
+	var didSend bool
+
+	for _, jsa := range js.accounts {
+		jsa.mu.RLock()
+		for _, stream := range jsa.streams {
+			stream.mu.RLock()
+			for _, o := range stream.consumers {
+				o.mu.RLock()
+				// Only signal on R1.
+				if o.cfg.Replicas <= 1 {
```
kozlovic, line 835: Do we care if this is not a real ephemeral? Meaning what if this is a durable with R1, do we still want that? Should we have the original test (`if ca.Config.Durable == _EMPTY_ && ca.Group != nil && len(ca.Group.Peers) == 1 && ca.Group.isMember(ourID)`)?

At very least, I think we should do this only for pull consumers, no? (isPullMode()). Otherwise push consumers would receive this 409 on the "data" plane (the lib intercept those "protocols" for pull consumers, not push).
I understand that o.pendingRequestReplies() would return nil, so there would be nothing to range over, so code right now is correct, but it could be more telling to check on isPullMode().

kozlovic (commented):
Maybe be explicit that we are dealing with pull consumers?

server/jetstream.go
derekcollison: Yes since the requests that are pending for us will be invalidated no matter what.

We could be more explicit, but the call to gather pending only works on pull consumers, so its a no-op otherwise.

kozlovic (approved):
LGTM

PR #3577, opened 2022-10-25
Make sure to stop raft nodes on all monitor exits.

derekcollison:

/cc @nats-io/core

### DIFF
--- a/server/jetstream_cluster.go
+++ b/server/jetstream_cluster.go
@@ -888,6 +888,9 @@ func (js *jetStream) monitorCluster() {
 	s.Debugf("Starting metadata monitor")
 	defer s.Debugf("Exiting metadata monitor")
 
+	// Make sure to stop the raft group on exit to prevent accidental memory bloat.
+	defer n.Stop()
+
 	const compactInterval = 2 * time.Minute
 	t := time.NewTicker(compactInterval)
 	defer t.Stop()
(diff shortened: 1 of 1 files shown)


server/jetstream_cluster.go
```diff
@@ -888,6 +888,9 @@ func (js *jetStream) monitorCluster() {
 	s.Debugf("Starting metadata monitor")
 	defer s.Debugf("Exiting metadata monitor")
 
+	// Make sure to stop the raft group on exit to prevent accidental memory bloat.
+	defer n.Stop()
```
kozlovic, line 892: I am wondering: I believe that there are legitimate cases where the "monitor" go routine (true for all monitor routines - cluster, stream, consumer) can exit, say because of a leader change, etc... and may be restarted after. Since I am not sure we have a solid (or any at all) "wait for routine to be done before restarting", when restarting the monitor routine, are we not taking the risk of having this defer of the previous instance actually execute *after* the new instance has been started, effectively killing the restart?

derekcollison: The monitorRoutines run for all members, not just the leader. They will not exit because of a leader change etc.

We have tests that test for a restart, meaning come out and go back in and all tests passed.

kozlovic (approved):
LGTM

PR #3578, opened 2022-10-25
Added warning for high message lag into a clustered stream.

derekcollison:

/cc @nats-io/core

### DIFF
--- a/server/jetstream_cluster.go
+++ b/server/jetstream_cluster.go
@@ -6636,8 +6636,8 @@ func (mset *stream) processClusteredInboundMsg(subject, reply string, hdr, msg [
 	// TODO(dlc) - Make this a limit where we drop messages to protect ourselves, but allow to be configured.
 	const warnThreshold = 10_000
 	if mset.clseq-lseq > warnThreshold {
-		err := fmt.Errorf("JetStream stream '%s > %s' has high message lag", jsa.acc().Name, mset.cfg.Name)
-		s.RateLimitWarnf(err.Error())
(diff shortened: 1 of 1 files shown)


server/jetstream_cluster.go
```diff
@@ -6631,6 +6631,14 @@ func (mset *stream) processClusteredInboundMsg(subject, reply string, hdr, msg [
 	if err != nil && mset.clseq > 0 {
 		mset.clseq--
 	}
+
+	// Check to see if we are being overrun.
+	// TODO(dlc) - Make this a limit where we drop messages to protect ourselves, but allow to be configured.
+	const warnThreshold = 10_000
+	if mset.clseq-lseq > warnThreshold {
+		err := fmt.Errorf("JetStream stream '%s > %s' has high message lag", jsa.acc().Name, mset.cfg.Name)
```
kozlovic, line 6639: Let's make sure that someone does not inadvertently remove the ":" from ":=" otherwise we would fail the publish (although this is your goal in the future, not at the moment). We could have used a local error variable with a different name here - to avoid confusion - or simply to a one liner with s.RateLimitWarnf(fmt.Errorf(...).Error()) although that would be a long line :-)

kozlovic (approved):
LGTM (but see comment)

server/jetstream_cluster.go
derekcollison: Agree, fixed.

kozlovic (approved):
LGTM

PR #3579, opened 2022-10-26
[FIXED] MaxMsgPerSubject recovered on corrupt or missing index file.

derekcollison:
On abnormal server exit, for streams or KV with max msgs per subject set we could recover more than N msgs per subject.
This fix allows for recovery of correct state on restart when index files are missing or not current.

/cc @nats-io/core

### DIFF
--- a/server/filestore.go
+++ b/server/filestore.go
@@ -1178,6 +1178,11 @@ func (fs *fileStore) recoverMsgs() error {
 		fs.startAgeChk()
 	}
 
+	// If we have max msgs per subject make sure the is also enforced.
+	if fs.cfg.MaxMsgsPer > 0 {
+		fs.enforceMsgPerSubjectLimit()
+	}
+
 	return nil
 }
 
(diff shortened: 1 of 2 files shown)


server/filestore.go
```diff
@@ -2136,6 +2141,76 @@ func (fs *fileStore) enforceBytesLimit() {
...
+
+	// We want to suppress callbacks from remove during this process
+	// since these should have already been deleted and accounted for.
+	cb := fs.scb
+	fs.scb = nil
+	defer func() { fs.scb = cb }()
+
+	// Also suppress writeIndex info since we will do that manually at the end.
+	fip := fs.fip
+	fs.fip = false
```
kozlovic, line 2159: As seen in a test failure, this causes removeMsg() to spin a go routine to do the flush asynchronously, which was detected by a data race. Maybe you need a new boolean in fs struct to handle this case?

kozlovic (changes requested):

server/filestore.go
derekcollison: Ah thanks, its just optimization and we have a flush timer to not go too fast, will remove.

derekcollison: fixed

server/filestore_test.go
```diff
@@ -4724,3 +4724,77 @@ func TestFileStoreOutOfSpaceRebuildState(t *testing.T) {
...
+	fs.mu.RLock()
+	for _, mb := range fs.blks {
+		mb.mu.Lock()
+		if err := mb.readIndexInfo(); err != nil {
+			mb.mu.Unlock()
+			fs.mu.RUnlock()
+			t.Fatalf("Unexpected error reading index info: %v", err)
+		}
+		if mb.msgs == 0 {
+			mb.mu.Lock()
```
kozlovic, line 4793: Change to `mb.mu.Unlock()`

kozlovic (changes requested):
Fix needed in the test

server/filestore_test.go
derekcollison: Fixed and squashed, sorry about that.

kozlovic (approved):
LGTM

PR #3580, opened 2022-10-27
adds the notion of a connection deadline to User

ripienaar:
This will be used mainly by CustomClientAuthentication implementations to indicate that the user connection should be disconnected at some point in future - like when a certificate or token expires

/cc @nats-io/core

### DIFF
--- a/server/auth.go
+++ b/server/auth.go
@@ -71,6 +71,7 @@ type User struct {
 	Password               string              `json:"password"`
 	Permissions            *Permissions        `json:"permissions,omitempty"`
 	Account                *Account            `json:"account,omitempty"`
+	ConnectionDeadline     time.Time           `json:"connection_deadline,omitempty"`
 	AllowedConnectionTypes map[string]struct{} `json:"connection_types,omitempty"`
 }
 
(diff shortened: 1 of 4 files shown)


ripienaar:
@kozlovic draft PR to discuss what we spoke about, seemed easier than anticipated so I think I must be missing something, let me know your thoughts please?

server/auth.go
```diff
@@ -83,6 +84,14 @@ func (u *User) clone() *User {
 	clone := &User{}
 	*clone = *u
 	clone.Permissions = u.Permissions.clone()
+
+	if len(u.AllowedConnectionTypes) > 0 {
```
ripienaar, line 88: dont know if this is needed, just noticed it might not be cloning this struct

kozlovic: I guess this is good to do.

kozlovic (approved):
I think this is good. Make sure it covers all the cases you want and then you could make it a formal PR against dev branch.

ripienaar:
OK, will finish it tomorrow thanks

ripienaar:
Actually it does what I want right now, so I am happy to merge unless you have anythign else @kozlovic

PR #3585, opened 2022-10-27
[FIXED] Existing subs would be sent to leafnodes even though they violated permissions.

derekcollison:
The LS+ would be sent for existing subscriptions upon leafnode connect but then would be silently denied due to permissions violations. This would make it tough for anyone with only access to the leafnode to understand why a request might be getting NoResponders.

We now do not allow existing subs to pass, so a trace log on the leafnode would not show an LS+ for the service subject.
If the LS+ does get through we debug log that it was denied.

/cc @nats-io/core

### DIFF
--- a/server/leafnode.go
+++ b/server/leafnode.go
@@ -1670,7 +1670,8 @@ func (s *Server) initLeafNodeSmapAndSendSubs(c *client) {
 	c.leaf.smap = make(map[string]int32)
 	for _, sub := range subs {
 		subj := string(sub.subject)
-		if c.isSpokeLeafNode() && !c.canSubscribe(subj) {
+		// Check perms regardless of role.
+		if !c.canSubscribe(subj) {
 			c.Debugf("Not permitted to subscribe to %q on behalf of %s%s", subj, accName, accNTag)
 			continue
 		}
@@ -1970,6 +1971,7 @@ func (c *client) processLeafSub(argo []byte) (err error) {
 	if checkPerms && subjectIsLiteral(string(sub.subject)) && !c.pubAllowedFullCheck(string(sub.subject), true, true) {
 		c.mu.Unlock()
 		c.leafSubPermViolation(sub.subject)
+		c.Debugf(fmt.Sprintf("Permissions Violation for Subscription to %q", sub.subject))
 		return nil
 	}
 
--- a/server/leafnode_test.go
+++ b/server/leafnode_test.go
@@ -4564,3 +4564,101 @@ func TestLeafNodeSignatureCB(t *testing.T) {
 	defer sl.Shutdown()
 	checkLeafNodeConnected(t, sl)
 }
+
+type testLeafTraceLogger struct {
+	DummyLogger
+	ch chan string
+}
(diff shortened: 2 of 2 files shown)


server/leafnode_test.go
```diff
@@ -4564,3 +4564,101 @@ func TestLeafNodeSignatureCB(t *testing.T) {
...
+
+	// Need to have ot reconnect to trigger since logger attaches too late.
+	ln.mu.Lock()
+	for _, c := range ln.leafs {
+		c.mu.Lock()
+		c.nc.Close()
+		c.mu.Unlock()
+	}
+	ln.mu.Unlock()
+	checkLeafNodeConnectedCount(t, ln, 0)
```
kozlovic, line 4643: There is a danger that this flap if the reconnect where to happen too quickly, no? Although by default the reconnect interval is 1 second, I believe, so should be ok. (I sometimes in test reduce the reconnect interval to something smaller)

server/leafnode.go
```diff
@@ -1670,7 +1670,8 @@ func (s *Server) initLeafNodeSmapAndSendSubs(c *client) {
 	c.leaf.smap = make(map[string]int32)
 	for _, sub := range subs {
 		subj := string(sub.subject)
-		if c.isSpokeLeafNode() && !c.canSubscribe(subj) {
+		// Check perms regardless of role.
+		if !c.canSubscribe(subj) {
```
kozlovic, line 1674: I am concerned about this one since "blame" seem to indicate that there was a reason "isSpokeLeafNode()" was added: https://github.com/nats-io/nats-server/commit/6294c0b0c77279543d2fa7ac276613226fc20782 (meaning that prior to that commit there was no c.isSpokeLeafNode() check, only !c.canSubscribe(), so concerned that we could introduce a regression)

server/leafnode_test.go
derekcollison: Without it test would always pass when it should have failed, so I need to wait for it to go to 0 before waiting for it to go back to 1.

server/leafnode.go
derekcollison: I thought same thing around wider scoped subs being blocked but verified that they are not..

derekcollison: On the git blame did we add a test?

server/leafnode_test.go
kozlovic: Understood, but I was just concerned that if it reconnects too quickly, then this actual test would fail (since count would be 1). But like I said, I think the default reconnect interval is 1s, so that should be fine. Marking as resolved.

server/leafnode.go
kozlovic: no, unfortunately.

kozlovic: But it was linked to a user error report, so maybe worth going back and see if that causes the issue to reappear.

derekcollison: I verified it works properly against that report.

kozlovic (approved):
LGTM

PR #3589, opened 2022-10-28
[FIXED] Move code that references sreq past the check for nil

derekcollison:
Signed-off-by: Derek Collison <derek@nats.io>

/cc @nats-io/core

PR #3596, opened 2022-11-01
[FIXED] JetStream: processing of snapshot with expired messages

kozlovic:
The issue that a "first sequence mismatch" during processing of a snapshot was causing the state to be reset and caused a lot of catchup from the follower. An attempt to fix that in PR #3567 caused an issue that was addressed in PR #3589. However, this was then causing the follower to sometime never able to catchup or took a very long time.
This PR - we believe - addresses the original and subsequent issues.


derekcollison (approved):
LGTM

PR #3599, opened 2022-11-02
[FIXED] JetStream: possible panic on some rare cases

kozlovic:
Very difficult to reproduce. Had to run TestJetStreamSuperClusterMoveCancel in covermode=atomic on a slow machine to hit the condition where the monitorConsumer go routine is started by RAFT node is nil, which caused the warning message to produce the panic (since n is nil)


derekcollison (approved):
LGTM

PR #3601, opened 2022-11-02
[FIXED] On stream proposal failures we could accidentally warn about high stream lag.

derekcollison:
We were not taking the failures (mset.clfs) into account.

/cc @nats-io/core

### DIFF
--- a/server/jetstream_cluster.go
+++ b/server/jetstream_cluster.go
@@ -6468,6 +6468,9 @@ func (mset *stream) checkAllowMsgCompress(peers []string) {
 	mset.mu.Unlock()
 }
 
+// To warn when we are getting too far behind from what has been proposed vs what has been committed.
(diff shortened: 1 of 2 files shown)


server/jetstream_cluster.go
```diff
@@ -6468,6 +6468,9 @@ func (mset *stream) checkAllowMsgCompress(peers []string) {
 	mset.mu.Unlock()
 }
 
+// To warn when we are getting tooo far behind from what has been proposed vs what has been committed.
```
kozlovic, line 6471: typo `too`

kozlovic (approved):
LGTM!

server/jetstream_cluster.go
derekcollison: Will fix and then merge once green. Thanks.

PR #3604, opened 2022-11-03
[FIXED] LeafNode: possible duplicate messages in complex setup

### LINKED ISSUE #3191: Duplicates message received while consuming with leafnode connection
## Defect

Messages are delivered twice to the same client.

The NATS client receives each message 2 times.
Both messages are delivered almost at the same time (<10ms delta) and are identical (Same headers, sequence id, body, delivery count)

- We experience this issue on all our environments
- We are able to reproduce the issue in a local kind cluster
- Unfortunately we are not able to make up a deterministic procedure to create the issue
- We need to restart servers multiple times in a sequence we could not understand for the problem to appear

[nats-server v2.8.4](https://github.com/nats-io/nats-server/releases/tag/v2.8.4)  
[nats.go v1.16.0](https://pkg.go.dev/github.com/nats-io/nats.go@v1.16.0)

Running nats in Kubernetes cluster with official [helm chart](https://docs.nats.io/running-a-nats-service/nats-kubernetes/helm-charts).

Setup diagram:
jetstream-issue

Given:
 - 2 nats clusters (A and B) composed of 2 nats-server instances each with jetstream running on all instances
 - Cluster A establishes leaf node connections to the cluster B
 - A go service (Application 1, Application 2) using the client lib version `nats.go@1.16.0` is connected to the cluster B and consumes messages from a stream that lives in the cluster A (thus via a leafnode connection).
 
When:
 - an unexpected sequence of events happen, nats server restarts, network partitions...
 
Then:
 - the cluster enters incoherent state and doesn't recover (all nodes running, no network partition)
 - The application instance receives each messages 2 times 
 - The duplicated messages are received in the same millisecond on the same instance
 - Restarting the application does not resolve the issue
 - The issue does not self resolve with time
 - Messages are systematically and constantly received 2 times.
 - Restarting the nats-server that sends the duplicates (the one which the client is connected to) is the ONLY way to stop that problem

Logs from a NATS server that sends duplicates (duplicates start at sequence 104):
[nats-0-trace.log](https://github.com/nats-io/nats-server/files/8901072/nats-0-trace.log)

#### Expected result:

Each message should be received once from the same consumer or subscription.

#### Actual result:

The go client subscribing to the remote consumer receives duplicated message.

kozlovic:
This is specific to setup described [here](https://github.com/nats-io/nats-server/issues/3191#issuecomment-1296974382) and does not require JetStream to be reproduced. The added test reproduces the above setup but without JetStream enabled in the accounts.

Each cluster has a leafnode for a given account to the other cluster. The accounts import/export a subject. When a consumer is connected to cluster "B" and the producer is on cluster "A" there was a duplicate message. Due to shadow subscription caused by the import/export rules, an additional subscription was sent across the leafnode.

Resolves #3191

### DIFF
--- a/server/leafnode.go
+++ b/server/leafnode.go
@@ -816,6 +816,7 @@ func (s *Server) removeLeafNodeURL(urlStr string) bool {
 
 // Server lock is held on entry
 func (s *Server) generateLeafNodeInfoJSON() {
+	s.leafNodeInfo.Cluster = s.cachedClusterName()
 	s.leafNodeInfo.LeafNodeURLs = s.leafURLsMap.getAsStringSlice()
 	s.leafNodeInfo.WSConnectURLs = s.websocket.connectURLsMap.getAsStringSlice()
 	b, _ := json.Marshal(s.leafNodeInfo)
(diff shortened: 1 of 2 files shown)


derekcollison (approved):
LGTM

chenchunping:
I've tested this issue after upgrading nats 2.9.6 and it doesn't occur anymore, thank you very much for fixing this issue so quickly.
We rely heavily on nats in our business and I'd like to know why this is causing duplicate messages, can you email or github me if you can?

PR #3605, opened 2022-11-03
[FIXED] Notifying waiting pull consumers on consumer or parent stream deletion.

### LINKED ISSUE #3603: Pull consumer - notify the client about stream/consumer no being available
## Feature Request

Currently, a `$JS.API.CONSUMER.MSG.NEXT.<stream>.<consumer>` API call does not notify the client if a given stream or consumer does not exist. Because of this, clients have to rely on heartbeats to detect any issues, but that comes with a delay and does not give any context on the error.

#### Proposed Change:

Server should send a status message on the inbox, with details on why a pull has failed (stream/consumer does not exist) and terminate the pull request immediately.

derekcollison:

Resolves #3603 

/cc @nats-io/core

### DIFF
--- a/server/consumer.go
+++ b/server/consumer.go
@@ -1882,6 +1882,25 @@ func (o *consumer) checkPendingRequests() {
 	o.prm = nil
 }
 
+// This will release any pending pull requests if applicable.
+// Should be called only by the leader being deleted or stopped.
+// Lock should be held.
+func (o *consumer) releaseAnyPendingRequests() {
+	if o.mset == nil || o.outq == nil || o.waiting.len() == 0 {
+		return
+	}
+	hdr := []byte("NATS/1.0 409 Consumer Deleted\r\n\r\n")
+	wq := o.waiting
+	o.waiting = nil
+	for i, rp := 0, wq.rp; i < wq.n; i++ {
+		if wr := wq.reqs[rp]; wr != nil {
+			o.outq.send(newJSPubMsg(wr.reply, _EMPTY_, _EMPTY_, hdr, nil, nil, 0))
+			wr.recycle()
+		}
+		rp = (rp + 1) % cap(wq.reqs)
+	}
+}
+
 // Process a NAK.
 func (o *consumer) processNak(sseq, dseq, dc uint64, nak []byte) {
 	o.mu.Lock()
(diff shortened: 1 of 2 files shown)


server/consumer.go
```diff
@@ -1882,6 +1882,25 @@ func (o *consumer) checkPendingRequests() {
...
 }
 
+// This will release any pending pull requests if applicable.
+// Should be called only by the leader being deleted or stopped.
+// Lock should be held.
+func (o *consumer) releaseAnyPendingRequests() {
+	if o.mset == nil || o.outq == nil || o.waiting.len() == 0 {
+		return
+	}
+	hdr := []byte("NATS/1.0 409 Consumer Deleted\r\n\r\n")
```
kozlovic, line 1892: I guess we don't need to send the pending msgs/bytes in this case since the consumer is deleted.

server/jetstream_cluster_3_test.go
```diff
@@ -1003,3 +1003,50 @@ func TestJetStreamClusterStreamLagWarning(t *testing.T) {
...
+	require_Error(t, err, expectedErr)
+
+	// Check that we bailed early.
+	if time.Since(start) > time.Second {
+		t.Fatalf("Took to long to bail out on consumer delete")
+	}
+
+	time.AfterFunc(250*time.Millisecond, func() { js.DeleteStream("TEST") })
+	start = time.Now()
+	_, err = sub2.Fetch(1, nats.MaxWait(10*time.Second))
```
kozlovic, line 1047: I guess that could be one that should be considered "terminal" in the client (longer discussion) and would invalidate the subscription (close subscription): so that an application check Fetch() error and bail on connection closed or invalid subscription (as opposed to a leader change, etc... where Fetch() could be tried again).

kozlovic (approved):
LGTM

server/consumer.go
derekcollison: I think that is correct, but should we add it?

server/jetstream_cluster_3_test.go
derekcollison: The deletion of parent stream deletes the consumer. Just wanted to test both.

kozlovic (approved):
LGTM. I don't think we need to add, but we could, up to you.

server/jetstream_cluster_3_test.go
derekcollison: Going to leave as is for now.

PR #3610, opened 2022-11-05
[FIXED] #3499

### LINKED ISSUE #3499: Memory skyrockets after `nats stream view`
nats-server: 2.9.1
nats: 0.0.34
3 node cluster, AWS EC2 Ubuntu

Load before:
before

Load after `nats stream view`
after

After a few seconds CPU returned to before level. After a few minutes memory too returned to before level.

If curious about the max memory limit, its because previously (and with previous nats-server version) these spikes exhausted entire node memory causing node outage. Setting this limit doesn't seem to have affected the normal workings of nats-server.

derekcollison:
When we deleted a consumer from an interest policy stream we would make sure to clean up any unacked messages. However we only based start from the ack floor for the consumer and did not take into account the first sequence of the stream.

Resolves #3499 

/cc @nats-io/core


ripienaar (approved):
LGTM

PR #3611, opened 2022-11-08
[FIXED] Routing: TLS connections to discovered server may fail

kozlovic:
The server was not setting "server name" in the TLS configuration for route connections, which may lead to failed (re)connect if the certificate does not allow for the IP and the URL did not have the hostname, which would happen with gossip protocol.


derekcollison (approved):
LGTM

PR #3613, opened 2022-11-08
Updates to PR #3611

kozlovic:
- Save the TLS name only if not already set
- Use the passed URLs slice instead of using s.getOpts().Routes
- Enhanced the test
- Fixed an unrelated DATA RACE report


derekcollison (approved):
LGTM

PR #3615, opened 2022-11-10
[FIXED] JetStream: Account max streams/consumers not always honoured

kozlovic:
This could happen during concurrent requests where the assignments are not yet fully processed.


derekcollison (approved):
LGTM

PR #3616, opened 2022-11-10
[WIP] Integrate oss-fuzz into CI on pull requets

piotrpio:


kozlovic (approved):
I see that there was a run on GA already and seem to work fine. To be honest, I would not know what we need to know if that fails, but for this PR itself, LGTM.

piotrpio:
@kozlovic I am not sure we want to merge it, as it adds extra 15 minutes to the build time. I created it to verify the errors reported by these tests:
```
==33== ERROR: libFuzzer: out-of-memory (used: 2801Mb; limit: 2560Mb)
```

PR #3624, opened 2022-11-14
[IMPROVED] Don't copy pending on needAck check.

derekcollison:
When determining whether we need an ack, no need to copy since under consumer lock.

/cc @nats-io/core

### DIFF
--- a/server/filestore.go
+++ b/server/filestore.go
@@ -6445,6 +6445,16 @@ func (o *consumerFileStore) Type() StorageType { return FileStorage }
 // State retrieves the state from the state file.
 // This is not expected to be called in high performance code, only on startup.
 func (o *consumerFileStore) State() (*ConsumerState, error) {
+	return o.stateWithCopy(true)
+}
+
+// This will not copy pending or redelivered, so should only be done under the
+// consumer owner's lock.
+func (o *consumerFileStore) BorrowState() (*ConsumerState, error) {
+	return o.stateWithCopy(false)
+}
+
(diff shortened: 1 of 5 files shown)


server/memstore.go
```diff
@@ -1171,6 +1171,16 @@ func (o *consumerMemStore) StreamDelete() error {
 }
 
 func (o *consumerMemStore) State() (*ConsumerState, error) {
+	return o.stateWithCopy(true)
+}
+
+// This will not copy pending or redelivered, so should only be down under the
```
kozlovic, line 1177: typo: down->done

server/filestore.go
```diff
@@ -6445,6 +6445,14 @@ func (o *consumerFileStore) Type() StorageType { return FileStorage }
 // State retrieves the state from the state file.
 // This is not expected to be called in high performance code, only on startup.
 func (o *consumerFileStore) State() (*ConsumerState, error) {
+	return o.stateWithCopy(true)
+}
+
+func (o *consumerFileStore) BorrowState() (*ConsumerState, error) {
```
kozlovic, line 6451: What am I missing? I don't see this function being called?

kozlovic (changes requested):

server/memstore.go
derekcollison: Will fix.

server/filestore.go
derekcollison: lol forgot to plug in. fixed.

kozlovic (approved):
LGTM

PR #3627, opened 2022-11-14
[FIXED] JetStream: Account removal leaks subscriptions

### LINKED ISSUE #3623: Nats deleted account leak subscriptions
## Defect
When adding a jetstream enabled account, it creates 6 (internal?) subscriptions, when deleting the account, those subscriptions remains until the server is restarted.
The step bellow reproduce it on a single server and a single account. But we found it on a cluster with several (a lot of) accounts where the number of subscriptions, the messages rate (in and out) and the memory used by the 3 server process where getting higher after each account creation and never got down when the accounts where deleted.

[nats-server.log](https://github.com/nats-io/nats-server/files/10002937/nats-server.log)

[commands.txt](https://github.com/nats-io/nats-server/files/10002917/commands.txt)

nats-server: 2.9.6

debian 11

see attached commands.txt

#### Expected result:
The number of subscriptions should drop back to the initial values (from 59 to 53 in the example case) after deleting the account.

#### Actual result:
The number of subscriptions only drop after restarting the server.

kozlovic:
Resolve #3623

### DIFF
--- a/server/accounts.go
+++ b/server/accounts.go
@@ -3813,9 +3813,16 @@ func removeCb(s *Server, pubKey string) {
 	a.mconns = 0
 	a.mleafs = 0
 	a.updated = time.Now().UTC()
+	jsa := a.js
 	a.mu.Unlock()
 	// set the account to be expired and disconnect clients
 	a.expiredTimeout()
+	// For JS, we need also to disable JS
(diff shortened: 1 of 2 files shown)


derekcollison (approved):
LGTM

PR #3629, opened 2022-11-15
When reading state, even on consumer init, should lock.

derekcollison:

/cc @nats-io/core


kozlovic (approved):
LGTM

Jarema (approved):
LGTM

PR #3633, opened 2022-11-15
Offload signaling to consumers when number is large.

derekcollison:
This will offload consumer processing and signaling from the stream ingest routine when the number of consumers passes a threshold.

/cc @nats-io/core


kozlovic (approved):
LGTM

PR #3638, opened 2022-11-16
[FIXED] Make mirror consumers use filtered version of consumer create

derekcollison:
Creating source consumers used this but mirrors were overlooked, this fixes that.

/cc @nats-io/core


kozlovic (approved):
LGTM

PR #3640, opened 2022-11-17
[FIXED] JetStream: WorkQueue not preventing overlapping consumers

### LINKED ISSUE #3639: WorkQueuePolicy incorrectly allows consumers with overlapping subjects
Reported by @Kilio22

### Discussed in https://github.com/nats-io/nats-server/discussions/3637

<div type='discussions-op-text'>

<sup>Originally posted by **Kilio22** November 16, 2022</sup>
Hi,

I am new to NATS, and I am playing with Jetstream and the different retention policies, especially the WorkQueuePolicy.  
I noticed in the following example https://natsbyexample.com/examples/jetstream/workqueue-stream/go that you cannot create two consumers with overlaping filtrer subjects.

However, I manage to do the following using nats-cli and a local nats-server started in Jetstream mode:
* create a stream with `toto.>` as subject
* create a first pull consumer with `toto.*.tata` as subject
* create a second pull consumer with `toto.>` as subject

Here, to me the subjects are overlaping, as both can match a message published on a subject such as `toto.tutu.tata`.  
I was thinking that maybe it would result in receiving a message published on subject `toto.tutu.tata` only on the first consumer, but no, it is received by both consumers.

Could you explain why please?

Thanks!</div>

kozlovic:
A stream with a WorkQueue retention policy is supposed to allow more than one consumer if they user filtered subjects, but those subjects should not overlap.

There was an issue that if a new consumer had a filter subject "wider" than an existing one, the error was not detected and the new consumer was incorrectly accepted.

Resolves #3639

### DIFF
--- a/server/consumer.go
+++ b/server/consumer.go
@@ -650,9 +650,25 @@ func (mset *stream) addConsumerWithAssignment(config *ConsumerConfig, oname stri
 				mset.mu.Unlock()
 				return nil, NewJSConsumerWQMultipleUnfilteredError()
 			} else if !mset.partitionUnique(config.FilterSubject) {
-				// We have a partition but it is not unique amongst the others.
-				mset.mu.Unlock()
-				return nil, NewJSConsumerWQConsumerNotUniqueError()
+				// Prior to v2.9.7, on a stream with WorkQueue policy, the servers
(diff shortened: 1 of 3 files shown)


derekcollison (approved):
LGTM

PR #3659, opened 2022-11-22
If no_auth_user is set, clear auth required from server info to client.

derekcollison:

/cc @nats-io/core


wallyqs (approved):
LGTM

PR #3663, opened 2022-11-23
Add a system account responder for IDZ along with STATSZ etc.

derekcollison:
This will avoid write locks and any non-static data.

/cc @nats-io/core


ripienaar (approved):
LGTM

PR #3667, opened 2022-11-25
[FIXED] Operator Mode could return no_auth_required set to false on client connect.

derekcollison:
In operator mode the system could incorrectly return no_auth_required when a client tries to connect.
This is due to the mechanism used to allow system users but not official users on the $G (global account) and a change in 2.9.8 that if no_auth_user was set that we would not return auth_required set to true.

Any client (java) that suppressed sending credentials because of the flag could be affected.

This fix only executes that code when in non-operator mode now and when we are looking to suppress that flag for clients we make sure it is not the anonymous user if it was created.

A simple fix for those affected is to have the server reload its config which will clear this behavior.

This only is present in 2.9.8 server version and when the server is in operator mode with an account resolver.

/cc @nats-io/core

### DIFF
--- a/server/jwt_test.go
+++ b/server/jwt_test.go
@@ -3512,6 +3512,7 @@ func TestJWTAccountNATSResolverFetch(t *testing.T) {
 		system_account: %s
 		resolver: {
 			type: full
+
 			dir: '%s'
 			interval: "200ms"
 			limit: 4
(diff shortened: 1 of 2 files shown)


ColinSullivan1 (approved):
LGTM.

ripienaar (approved):
LGTM

PR #3671, opened 2022-11-28
[ADDED] User Information requests at $SYS.REQ.USER.INFO

derekcollison:
Allow connected users to query information about bound account name and user permissions. This even works for users in the system account itself.

We do not return detailed account information here to allow for separate security/subject permissions for that and jetstream account information.

This should be added to NATS cli as a parallel call to account info when merged.

Note many of the changes are fixes to low level historical tests. When you add a base level system subscription that is imported by other accounts lots of test updates need to happen.

/cc @nats-io/core


server/events.go
```diff
@@ -1064,6 +1075,38 @@ func (s *Server) initEventTracking() {
 	}
 }
 
+// UserInfo returns basic information to a user about bound account and user permissions.
+// For account information they will need to ping that separately, and this allows security
+// controls on each subsystem if desired, e.g. account info, jetstream account info, etc.
+type UserInfo struct {
+	UserID      string       `json:"user"`
+	Account     string       `json:"account"`
+	Permissions *Permissions `json:"permissions,omitempty"`
```
ColinSullivan1, line 1084: Also include Time of Day restrictions and CIDR block restrictions?  Expiration might be useful here too. wdyt?

derekcollison: Let me add expiration, I think we hold on the others since that could open up a bunch of options on what is reported that blurs between user and accounts.

Will add expiration, that is a good add for sure.

ColinSullivan1 (approved):
LGTM.

ripienaar (approved):
LGTM will put into nats account info and see what else might be useful

tbeets:
Love this.  No more peeking at connz to validate during testing!

PR #3674, opened 2022-12-02
allow expiration timer logic to be called unlocked

ripienaar:
This ensures the logic that triggers user disconnections share the same code paths and sets the same variables

/cc @nats-io/core


derekcollison (approved):
LGTM

PR #3678, opened 2022-12-04
tag policies not honored in reassignment after peer remove

tbeets:
Resolves issue in which tag policies (tag affinity and tag uniqueness constraint) were not honored when selecting a new peer for a stream RG to replace one just removed.  

/cc @nats-io/core


server/jetstream_cluster.go
```diff
@@ -4797,6 +4790,13 @@ func (cc *jetStreamCluster) selectPeerGroup(r int, cluster string, cfg *StreamCo
 			continue
 		}
 
+		// If ignore skip
+		if ip != nil {
```
derekcollison, line 4794: nil maps ok to dereference. So you can just do line 4795 direct.

https://go.dev/play/p/XLBsPw7NUBM

derekcollison (approved):
LGTM - One small suggestion.

server/jetstream_cluster.go
```diff
@@ -4769,6 +4753,13 @@ func (cc *jetStreamCluster) selectPeerGroup(r int, cluster string, cfg *StreamCo
 		}
 	}
 
+	// Map ignore
+	var ip map[string]struct{}
+	ip = make(map[string]struct{})
```
derekcollison, line 4758: Should still conditionally create here.

server/jetstream_cluster.go
derekcollison: This is where you can just dereference, and if nil will be ok.

derekcollison (changes requested):

server/jetstream_cluster.go
tbeets: Yep, just noticed.  Will fix.  Thx.

derekcollison (approved):
LGTM

PR #3680, opened 2022-12-05
Do not gather full state for snapshots

ripienaar:
This avoids full state outputs for streams with many deleted items

/cc @nats-io/core

### DIFF
--- a/server/filestore.go
+++ b/server/filestore.go
@@ -5748,7 +5748,8 @@ func (fs *fileStore) Snapshot(deadline time.Duration, checkMsgs, includeConsumer
 	}
 
 	// We can add to our stream while snapshotting but not delete anything.
-	state := fs.State()
+	var state StreamState
+	fs.FastState(&state)
 
 	// Stream in separate Go routine.
 	go fs.streamSnapshot(pw, &state, includeConsumers)


server/filestore.go
```diff
@@ -5749,6 +5749,8 @@ func (fs *fileStore) Snapshot(deadline time.Duration, checkMsgs, includeConsumer
 
 	// We can add to our stream while snapshotting but not delete anything.
 	state := fs.State()
+	var state StreamState
```
derekcollison, line 5752: Don't need above command?

ripienaar: Ah bad commit will fix. This file gets messed up by my IDE every time I save it

ripienaar:
fixed

derekcollison:
This for dev or could be for main?

ripienaar:
If you are certain you don’t use the extra state details to rebuild anything on restore we can do main else dev.

derekcollison:
We do not use any of the state tbh.. So edit this PR to merge to main.

ripienaar:
Allright, its pointing at main now.

derekcollison (approved):
LGTM

PR #3683, opened 2022-12-06
Update --signal usage help

bruth:
This includes the `ldm` and `term` signals that are valid options, but not documented.

/cc @nats-io/core


derekcollison (approved):
LGTM - Thanks!

PR #3690, opened 2022-12-06
Fix to reenable JetStream for account when reenabled (dev → main)

wallyqs:
PR to upstream change from #3687 to main

/cc @nats-io/core


derekcollison (approved):
LGTM

PR #3691, opened 2022-12-07
Make stream removal from a server consistent.

derekcollison:

/cc @nats-io/core


tbeets (approved):
LGTM

Issue #3682, opened 2022-12-06
Leaf node fails to reconnect, due to ping messages being held off indefinitely


sandykellagher:
## Defect

We are using a LeafNode NATS server to connect to a cluster, and see a strange effect which prevents the LeafNode reconnecting properly in the event that its link to the cluster goes down.

The LeafNode has two local clients which connect to it, and which generate traffic on a continuous basis. And in short, this continuous client traffic results in the outbound Pings from LeafNode to cluster being held off/delayed indefinitely, with the message "Leaf Node Ping Timer", "Delaying Ping due to client activity". And because the Pings are held off, the LeafNode doesn't detect a stale connection, and hence doesn't close the connection and attempt to reconnect.

I believe I understand the issue.

In NATS server client.go:: processPingTimer() there is a check to test whether to delay sending an outgoing ping, in two cases:

- we recently (within specified pingInterval) received a data message or sub/unsub message from the remote end (Client activity)
- we recently received a ping from the remote end (Remote ping)

This makes perfect sense: incoming receive messages mean we still have a link and don't need to send a ping.

However, the first test above is derived from the client.last field ("last packet" time) which is set in two cases:

- when the readLoop parsing determines we have received a new message or sub/unsub
- in flushClients() routine, which is called when we have received some messages that need to be forwarded to other client connections

But I believe that this second case is incorrect: we should only hold off pings when there has been receive traffic on the specified connection, and this isn't so in the second case. We received traffic on one connection, but are resetting the c.last field on another connection where we are forwarding/sending the message.

If I remove the line of code in flushClients() (about line 1113) that updates cp.last, then the Stale Client Connection fires fine in my testing.

#### Versions of `nats-server` and affected client libraries used:
Latest master NATS server as of 2/12/2022, eg git commit c4c87612

#### OS/Container environment:
Linux ARM64 - but not a factor

#### Steps or code to reproduce the issue:
Configure leaf node server with local clients sending traffic that an external client is subscribed for, and then break the connection to the external NATS cluster

#### Expected result:
Leaf node server should automatically trigger a reconnect when the connection to an external NATS cluster is lost, with the detection time as configured by ping_interval and ping_max parameters.

#### Actual result:
Leaf node server does not automatically reconnect. Instead, it remains in a zombie state indefinitely. 
To be more precise, it might possibly recover when the network stack TCP keepalive timeout expires (which by default is 2 hours), but that is much too long to be useful

sandykellagher:
[...]

PR #3692, opened 2022-12-07
Fix for #3682 (Take 2): do not delay PINGs for GATEWAY or spoke LEAF connections

sandykellagher:
This change is required to ensure failover of leaf node connections

This contribution is my original work and I license the work to the project under the [Apache 2 license](https://github.com/nats-io/nats-server/blob/main/LICENSE)

/cc @nats-io/core

### DIFF
--- a/server/client.go
+++ b/server/client.go
@@ -4547,19 +4547,20 @@ func (c *client) processPingTimer() {
 
 	var sendPing bool
 
-	// If we have had activity within the PingInterval then
-	// there is no need to send a ping. This can be client data
-	// or if we received a ping from the other side.
 	pingInterval := c.srv.getOpts().PingInterval
 	if c.kind == GATEWAY {
 		pingInterval = adjustPingIntervalForGateway(pingInterval)
-		sendPing = true
 	}
 	now := time.Now()
 	needRTT := c.rtt == 0 || now.Sub(c.rttStart) > DEFAULT_RTT_MEASUREMENT_INTERVAL
 
-	// Do not delay PINGs for GATEWAY connections.
-	if c.kind != GATEWAY {
+	// Do not delay PINGs for GATEWAY or spoke LEAF connections.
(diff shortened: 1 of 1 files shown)


derekcollison:
ok will take a look in a bit. Thanks.

sandykellagher:
First simple change was not correct, because sendPing flag would never get set for spokeLeafNode connections.
So I have tidied up the logic and inverted the test for gateway or spokeLeafNode to be a positive test (reads easier!)

sandykellagher:
Hi,
We have run into a similar issue with an intra-cluster connection, when testing system robustness by disconnecting or powering down one of the servers in our standard 3-cluster server.

The issue is that it takes a long time (several minutes) before the other servers in the cluster detect the loss of server A. And the logs show that this happens as a result of a WriteDeadline failure rather than through ping-pong detection mechanism.

In the time between physical network disconnection and the loss being detected, any request messages published to a queue group subscription may get forwarded to the dead server - which of course cannot respond.

So I think in client line 4558, the test for always sending a ping should include c.Kind == ROUTER as well as c.Kind == GATEWAY. 

TBH my personal preference would be to update c.last only for received traffic, but IIRC making that change triggered a Travis test failure. Or perhaps the condition for holding off a ping is limited to receiving a ping from the other side within the specified interval (rather than general traffic)

Does that make sense?  As it stands, it seems that intra-cluster ROUTER connections remain in a zombie state for quite a long time when the network connection is pulled, where I would expect them to be closed by the ping-pong monitoring mechanism.

sandykellagher:
Thinking some more, I'd suggest that we don't need to check the type of connection (c.Kind).

I think the rule should be "send a ping if we've not received any message traffic or a ping from the other end within the specified ping interval, or an RTT measurement is required". And this applies for all connection types.

If we update cp.ping.last for receive traffic as well as pings, then a single test on cp.ping.last is all we need.

PR #3695, opened 2022-12-08
Fix JWT tests JS store dir

wallyqs:
This was causing longer than necessary startup times since by default they would have been using the same temporary directory and the extra account lookup failures due to a previous JS run using different accounts and leaving those artifacts in the shared folder. Also renamed some of these tests to be possible to run them via `-run TestJetStreamJWT`.


derekcollison (approved):
LGTM

PR #3700, opened 2022-12-09
Temporary test files cleanup

mprimi:
 - [x] Tests added

 - Switch most tests to [testing.T::TempDir()](https://pkg.go.dev/testing#T.TempDir), which gets cleaned up automatically after test execution
 - Remove some no-longer-necessary deferred cleanup
 - Minor refactoring
 - Negative 1600 LOC!

Notice the PR is split into 5 atomic commits, to make review somewhat more manageable.

/cc @nats-io/core

### DIFF
--- a/logger/log_test.go
+++ b/logger/log_test.go
@@ -25,8 +25,6 @@ import (
 	"testing"
 )
 
-var tempRoot = filepath.Join(os.TempDir(), "nats-server")
-
 func TestStdLogger(t *testing.T) {
 	logger := NewStdLogger(false, false, false, false, false)
 
@@ -104,9 +102,7 @@ func TestStdLoggerTraceWithOutDebug(t *testing.T) {
 }
 
 func TestFileLogger(t *testing.T) {
-	tmpDir := createDir(t, "_nats-server")
-	defer removeDir(t, tmpDir)
-
+	tmpDir := t.TempDir()
 	file := createFileAtDir(t, tmpDir, "nats-server:log_")
 	file.Close()
 
@@ -176,8 +172,7 @@ func TestFileLoggerSizeLimit(t *testing.T) {
 	}
 	logger.Close()
 
-	tmpDir := createDir(t, "nats-server")
-	defer removeDir(t, tmpDir)
+	tmpDir := t.TempDir()
 
 	file := createFileAtDir(t, tmpDir, "log_")
 	file.Close()
@@ -210,10 +205,7 @@ func TestFileLoggerSizeLimit(t *testing.T) {
 		t.Fatalf("Should be statement about rotated log and backup name, got %s", content)
 	}
 
-	// Remove all files
-	removeDir(t, tmpDir)
-	tmpDir = createDir(t, "nats-server")
-	defer removeDir(t, tmpDir)
+	tmpDir = t.TempDir()
 
 	// Recreate logger and don't set a limit
 	file = createFileAtDir(t, tmpDir, "log_")
@@ -325,18 +317,6 @@ func expectOutput(t *testing.T, f func(), expected string) {
 	}
 }
 
-func createDir(t *testing.T, prefix string) string {
-	t.Helper()
-	if err := os.MkdirAll(tempRoot, 0700); err != nil {
-		t.Fatal(err)
-	}
-	dir, err := os.MkdirTemp(tempRoot, prefix)
-	if err != nil {
-		t.Fatal(err)
-	}
-	return dir
-}
-
 func createFileAtDir(t *testing.T, dir, prefix string) *os.File {
 	t.Helper()
 	f, err := os.CreateTemp(dir, prefix)
@@ -344,11 +324,4 @@ func createFileAtDir(t *testing.T, dir, prefix string) *os.File {
 		t.Fatal(err)
 	}
 	return f
-}
-
-func removeDir(t *testing.T, dir string) {
-	t.Helper()
-	if err := os.RemoveAll(dir); err != nil {
-		t.Fatal(err)
-	}
 }
--- a/server/accounts_test.go
+++ b/server/accounts_test.go
@@ -337,7 +337,6 @@ func TestAccountIsolationExportImport(t *testing.T) {
 			`,
 				c.exp, c.imp,
 			)))
-			defer removeFile(t, cf)
 			s, _ := RunServerWithConfig(cf)
 			defer s.Shutdown()
 
@@ -381,7 +380,6 @@ func TestMultiAccountsIsolation(t *testing.T) {
 			]
 		}
 	}`))
-	defer removeFile(t, conf)
 
 	s, _ := RunServerWithConfig(conf)
 	if config := s.JetStreamConfig(); config != nil {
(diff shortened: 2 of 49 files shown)


logger/log_test.go
```diff
@@ -25,8 +25,6 @@ import (
 	"testing"
 )
 
-var tempRoot = filepath.Join(os.TempDir(), "nats-server")
```
wallyqs, line 28: this part got introduced to make it easier cleanup the tests at some point, but `testing.TempDir()` (available since Go 1.15) already creates a folder with an easy to recognize name using the test like `TestFileLogger.../001` plus it auto cleans up, so this is a good change 👍

mprimi:
This is passing check and ready for review.

server/jetstream_test.go
```diff
@@ -9605,7 +9251,6 @@ func TestJetStreamStoreDirectoryFix(t *testing.T) {
 	defer removeDir(t, sd)
 
 	conf := createConfFile(t, []byte(fmt.Sprintf("listen: 127.0.0.1:-1\njetstream: {store_dir: %q}\n", sd)))
-	defer removeFile(t, conf)
```
wallyqs, line 9608: maybe we should make a note that for this test not using t.TempDir unlike other tests and have to rely on os.TempDir

server/jwt_test.go
```diff
@@ -6347,11 +6198,7 @@ func TestJWTOperatorPinnedAccounts(t *testing.T) {
 		kps[i], pubs[i] = createKey(t)
 		jwts[i] = encodeClaim(t, jwt.NewAccountClaims(pubs[i]), pubs[i])
 	}
-	sysCreds := newUser(t, kps[0]) // index 0 is handled as system account
-	defer removeFile(t, sysCreds)
-
-	dirSrv := createDir(t, "srv")
-	defer removeDir(t, dirSrv)
+	newUser(t, kps[0]) // index 0 is handled as system account
```
wallyqs, line 6201:
```suggestion
	// create system account user credentials, index 0 is handled as system account
	newUser(t, kps[0])
```

server/jwt_test.go
```diff
@@ -6347,11 +6198,7 @@ func TestJWTOperatorPinnedAccounts(t *testing.T) {
 		kps[i], pubs[i] = createKey(t)
 		jwts[i] = encodeClaim(t, jwt.NewAccountClaims(pubs[i]), pubs[i])
 	}
-	sysCreds := newUser(t, kps[0]) // index 0 is handled as system account
-	defer removeFile(t, sysCreds)
-
-	dirSrv := createDir(t, "srv")
```
wallyqs, line 6353: this directory was not even used in the test looks like

server/jetstream_test.go
```diff
@@ -88,12 +88,11 @@ func TestJetStreamBasicNilConfig(t *testing.T) {
 	}
 }
 
-func RunBasicJetStreamServer() *Server {
+func RunBasicJetStreamServer(t testing.TB) *Server {
```
wallyqs, line 91: I think we need to leave this one like this and a have a different one that takes the TB interface, maybe:
```suggestion
func RunTestJetStreamServer(t testing.TB) *Server {
```
or a private one would work `runBasicJetStreamServer`

server/opts_test.go
```diff
@@ -2466,6 +2437,7 @@ func TestParsingLeafNodeRemotes(t *testing.T) {
 		os.Setenv("HOME", "/home/foo")
 
 		content := `
+		port: -1
```
wallyqs, line 2440: server not being run on this tests so can omit adding this

server/opts_test.go
```diff
@@ -2426,6 +2397,7 @@ func TestParsingLeafNodesListener(t *testing.T) {
 func TestParsingLeafNodeRemotes(t *testing.T) {
 	t.Run("parse config file with relative path", func(t *testing.T) {
 		content := `
+		port: -1
```
wallyqs, line 2400: same here, server not being started so can skip port -1 config

wallyqs (changes requested):
Looks great, left some comments

derekcollison:
Once @wallyqs comments are addressed let's squash down and I will take a look. Thanks!

server/jetstream_test.go
mprimi: The method I modified is defined in `jetstream_test.go`, I don't think anyone outside this module would be able to see such symbol, no?

wallyqs: you are right, yes it would not have been exposed so would be be ok change

mprimi:
@derekcollison comments addressed. Just squashed & rebased.

derekcollison:
ok will take a look, but might not be til tomorrow.

We need to tidy some things up on all of these test under server directory, getting a but unwieldy..

logger/log_test.go
```diff
@@ -210,10 +205,7 @@ func TestFileLoggerSizeLimit(t *testing.T) {
 		t.Fatalf("Should be statement about rotated log and backup name, got %s", content)
 	}
 
-	// Remove all files
-	removeDir(t, tmpDir)
-	tmpDir = createDir(t, "nats-server")
-	defer removeDir(t, tmpDir)
+	tmpDir = t.TempDir()
```
derekcollison, line 208: So we do not need to clean these up? The directory?

server/accounts_test.go
```diff
@@ -536,8 +533,11 @@ func accountNameExists(name string, accounts []*Account) bool {
 }
 
 func TestAccountSimpleConfig(t *testing.T) {
-	confFileName := createConfFile(t, []byte(`accounts = [foo, bar]`))
-	defer removeFile(t, confFileName)
+	cfg1 := `
```
derekcollison, line 536: Would inline with call below.

server/accounts_test.go
```diff
@@ -552,9 +552,12 @@ func TestAccountSimpleConfig(t *testing.T) {
 		t.Fatal("Expected a 'bar' account")
 	}
 
+	cfg2 := `
```
derekcollison, line 555: Same.

server/accounts_test.go
```diff
@@ -2834,13 +2826,13 @@ func TestAccountMultiWeightedRouteMappings(t *testing.T) {
...
 func TestGlobalAccountRouteMappingsConfiguration(t *testing.T) {
 	cf := createConfFile(t, []byte(`
+	port: -1
 	mappings = {
 		foo: bar
 		foo.*: [ { dest: bar.v1.$1, weight: 40% }, { destination: baz.v2.$1, weight: 20 } ]
 		bar.*.*: RAB.$2.$1
     }
     `))
-	defer removeFile(t, cf)
```
derekcollison, line 2843: Like here, does this file just remain?

logger/log_test.go
wallyqs: yes not needed, the testing framework auto cleans these up when the test is done

derekcollison: ok nice! And we have actually test that is the case yes?

derekcollison (approved):
LGTM!

wallyqs (approved):
LGTM

PR #3702, opened 2022-12-09
Rate limit frequent debug log lines

ripienaar:
With the increase use of subject based limits not being able to store due to limits exceeded happens frequently and makes running the server in debug quite noisy, so we rate limit this log line even in debug

/cc @nats-io/core


ripienaar:
@derekcollison as discussed on slack, though I am not really convinced this is a good idea

derekcollison (approved):
LGTM

PR #3704, opened 2022-12-09
Fix /healthz?js-enabled=true behavior

### LINKED ISSUE #3703: Fix `/healthz` check for `js-enabled=true`
`js-enabled=true` is not a well defined health check request parameter currently.  It currently has the same effect that looking at the NATS Config to see if JS Should be enabled:

https://github.com/nats-io/nats-server/blob/d6f243f39bac2e0eaa577cd5e2bfa2d848a86389/server/monitor.go#L3012-L3021

We should instead just check the NATS Config to determine if JetStream should be enabled, and compare the current state to the config for the health check.

The `js-enabled=true` parameter should be changed to `js-enabled-only=true` and should terminate the Healthz endpoint after checking the Desired state (from the config) of JetStream against the Current state of JetStream

- `js-enabled-only=true` -> exit after checking if JetStream is enabled
- `js-server-only=true` -> exit after checking if Server is Current with Meta Leader
- no paramaters -> check everything

bruth:
When js-enabled is set to true, the condition was only checked if
the `getJetStream()` call returned `nil`. However, if it non-nil,
all remaining checks were executed, including assessing the health
of the assets (streams and consumers).

This change addresses two issues:

- Switch to use `js.isEnabled()` which will check whether the value
  is nil OR `js.disabled = true` which can occur if the subsystem
  is temporarily disabled (insufficient resources).
- Correctly exit the check after the assertion and before meta and
  asset checks are performed.

In addition, the option has been renamed to `js-enabled-only` to align
with the `js-server-only` naming. The previous `js-enabled` name still
works, but is mapped to this new option. A warning is emitted noting
the previous option is deprecated.

Fix https://github.com/nats-io/nats-server/issues/3703

Signed-off-by: Byron Ruth <b@devel.io>

/cc @nats-io/core

### DIFF
--- a/server/monitor.go
+++ b/server/monitor.go
@@ -2631,8 +2631,8 @@ type JSzOptions struct {
 
 // HealthzOptions are options passed to Healthz
 type HealthzOptions struct {
-	JSEnabled    bool `json:"js-enabled,omitempty"`
-	JSServerOnly bool `json:"js-server-only,omitempty"`
+	JSEnabledOnly bool `json:"js-enabled-only,omitempty"`
+	JSServerOnly  bool `json:"js-server-only,omitempty"`
 }
 
 type StreamDetail struct {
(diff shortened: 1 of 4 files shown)


server/monitor.go
```diff
@@ -3008,16 +3015,26 @@ func (s *Server) healthz(opts *HealthzOptions) *HealthStatus {
...
 			return health
 		}
 		return health
+	} else if opts.JSEnabledOnly {
+		return health
+	}
+
+	// Access the Jetstream state to perform additional checks.
+	js := s.getJetStream()
+	if js == nil {
```
wallyqs, line 3034: With this change then `js-server-only` then would be the recommended check to use whether JetStream has been disabled/enabled during runtime

bruth: Fixing some logic and pushing soon.. on a call with @caleblloyd :smile:

bruth:
Just noting that the one remaining Travis test that is failing seems to be unrelated to this fix.. there is a health check in the test, but the error happens after that.

caleblloyd:
@bruth remember, the behavior of `JSEnabledOnly` has now changed to exit the health check earlier, wheras before `JSEnabled` would run the entire Health Check.  So tests that relied on `JSEnabled=true` actually now need `JSEnabledOnly=false` to have the same effect

bruth:
@caleblloyd Yes, I will update, but I did test that locally and the test still failed.

bruth:
Of course.. I tried again locally.. and it works :man_facepalming: .

caleblloyd (approved):

wallyqs (approved):
LGTM

bruth:
^I squashed the commits and updated the commit message to be accurate with the changes.

derekcollison (approved):
LGTM

PR #3707, opened 2022-12-12
[FIXED] Make sure we remove node from stream assignment.

derekcollison:

/cc @nats-io/core


tbeets (approved):
Fix LGTM.  I will merge and then optimize and formalize the test.

PR #3708, opened 2022-12-13
zero-state condition on stream peer re-assignment

tbeets:
Fix and test that resolves a zero-state condition on stream peer re-assignment regression (peer rejoining it's former group). 

/cc @nats-io/core


derekcollison:
Yea seems you pulled in some other stuff here.

PR #3709, opened 2022-12-13
Fix for regression in which peer re-assign to a former RG would zero state

tbeets:
Fix and test for regression in which peer re-assign to an RG in which it was formerly a member, would silently zero state on that peer.

/cc @nats-io/core


derekcollison:
Looks like you have a data race. Check Travis.

Also you can re-order commits and squash all the ones for the test imo

tbeets:
@derekcollison fixed up the test and removed data race.  Squashed the last iterations of test into one after your server fix was merged.

With your approval, can merge with GitHub squash and merge to see if we can get all into one main commit.

derekcollison (approved):
LGTM!

PR #3710, opened 2022-12-13
Improve performance and latency with large number of sparse consumers.

derekcollison:
When a stream had a large number of consumers on a server that were sparse, the signaling mechanism would do a linear scan to signal matching consumers. As usage patterns have continued to have more consumers that are filtered and sparse, meaning a message is destined for a single or small number of consumers.

This change moves selection to a sublist that tracks only active consumer leaders for selection, which optimizes selection of consumers to signal when the number of consumers is large.

/cc @nats-io/core

### DIFF
--- a/server/consumer.go
+++ b/server/consumer.go
@@ -306,6 +306,9 @@ type consumer struct {
 
 	// Ack queue
 	ackMsgs *ipQueue
+
+	// For stream signaling.
+	sigSub *subscription
 }
 
 type proposal struct {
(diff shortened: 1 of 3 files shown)


bruth (approved):

tbeets:
@derekcollison Are we skipping the TestNoRaceJetStreamEndToEndLatency test?   Not needed?

derekcollison:
It takes 15+ seconds when its broken, but I could let it run and add in ability to make sure it received the last message in an expected period of time.

derekcollison:
ok fixed. Will squash once we think its good.

tbeets (approved):
LGTM

Issue #3712, opened 2022-12-14
R3 Memory Stream cannot recover when 1 server restarts in 2.9.9


caleblloyd:
R3 memory stream has RAFT errors in `2.9.9`.  This does not happen in `2.9.8`

1. Create a cluster of 3
2. Create an R3 memory stream `nats bench js.bench --js --pub 2 --msgs 1000000 --storage=memory --replicas=3 --size=1024 --purge`
    ```
    Cluster Information:

                 Name: nats
               Leader: nats-0
              Replica: nats-1, current, seen 0.63s ago
              Replica: nats-2, current, seen 0.63s ago
    ```
3. Restart one of the servers, in this case I did the replica `nats-2`
    ```
    Cluster Information:

                 Name: nats
               Leader: 
              Replica: nats-0, outdated, seen 2m57s ago, 160,069 operations behind
              Replica: nats-1, current, seen 4.61s ago
              Replica: nats-2, outdated, seen 4.61s ago, 1 operation behind

    ```
    
Logs from all 3 servers:

`nats-0`:
```
[7] 2022/12/14 16:01:56.024985 [INF] Starting nats-server
[7] 2022/12/14 16:01:56.025015 [INF]   Version:  2.9.9
[7] 2022/12/14 16:01:56.025017 [INF]   Git:      [825949b]
[7] 2022/12/14 16:01:56.025018 [INF]   Cluster:  nats
[7] 2022/12/14 16:01:56.025019 [INF]   Name:     nats-0
[7] 2022/12/14 16:01:56.025020 [INF]   Node:     S1Nunr6R
[7] 2022/12/14 16:01:56.025021 [INF]   ID:       NCQWINNGATEV3LXMT2FT2WBY6GQYVGA2W6FMTFBE5KWBCO2ACTOSQ7UG
[7] 2022/12/14 16:01:56.025026 [INF] Using configuration file: /etc/nats-config/nats.conf
[7] 2022/12/14 16:01:56.025471 [INF] Starting http monitor on 0.0.0.0:8222
[7] 2022/12/14 16:01:56.025495 [INF] Starting JetStream
[7] 2022/12/14 16:01:56.025566 [INF]     _ ___ _____ ___ _____ ___ ___   _   __  __
[7] 2022/12/14 16:01:56.025569 [INF]  _ | | __|_   _/ __|_   _| _ \ __| /_\ |  \/  |
[...]

PR #3713, opened 2022-12-15
fix and test for clustered mem store asset no-quorum if leader restarted

tbeets:
Resolves [Issue #3712](https://github.com/nats-io/nats-server/issues/3712)

If the leader peer of a clustered memory-store stream was restarted, stream would become no quorum after the peer returned.

/cc @nats-io/core


derekcollison (approved):
LGTM

derekcollison:
Once tests are green will merge. Thanks.

derekcollison:
You have some staticcheck failures..

PR #3714, opened 2022-12-15
Simplify locking for consumer info requests.

derekcollison:

/cc @nats-io/core

### DIFF
--- a/server/consumer.go
+++ b/server/consumer.go
@@ -1122,10 +1122,7 @@ func (o *consumer) setLeader(isLeader bool) {
 }
 
 func (o *consumer) handleClusterConsumerInfoRequest(sub *subscription, c *client, _ *Account, subject, reply string, msg []byte) {
-	o.mu.RLock()
-	sysc := o.sysc
-	o.mu.RUnlock()
-	sysc.sendInternalMsg(reply, _EMPTY_, nil, o.info())
+	o.infoWithSnapAndReply(false, reply)
 }
 
 // Lock should be held.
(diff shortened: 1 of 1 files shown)


ripienaar (approved):
LGTM

Jarema (approved):
LGTM

ripienaar:
Probably should have gone to dev branch as it’s not a big fix

bruth:
The release document, as defined today, allows for any "change" that does not change the API nor has a change in behavior that users/applications otherwise expected/relied on. Of course a bug fix that corrects the intended behavior is fine. In practice this includes:

- Bug fixes
- Refactoring
- Optimization
- Code documentation

Assuming this PR is purely a refactor and _safe_, then it can target main for a patch release.

derekcollison:
We are still settling on the exact nature, @bruth is leading the effort. Some are in a gray area but if potential performance or stability issues..

ripienaar:
Need to consider the potential impact of those also. Changing locking - given how complex it is in here - can cause severe bugs.

derekcollison:
If it was a severe bug I would imagine it would be caught by our over 600+ jetstream tests, at least one.

bruth:
Aside from the unit tests, @ripienaar what would _you_ do to test this change so you are confident in it working as expected? Load testing with concurrent clients of some kind? The more we can feed @mprimi and @wallyqs with the higher-level testing, the better..

ripienaar:
We have had many bugs introduced by locking changes that was only found during manual tests later. 

We fixed a memory raft bug yesterday that got introduced during what seems like work done during quite minor bug fixes. 

These things happen. We are 10 releases deep into 2.9.x there is a clear signal here that we need to be extra vigilant.

bruth:
> These things happen. We are 10 releases deep into 2.9.x there is a clear signal here that we need to be extra vigilant.

Yes agreed. Although delaying a change in a patch vs minor release won't itself solve the "test sufficiency" problem. It will provide more time for us to possibly stumble upon it during the release cycle, but that is obviously not a dependable way to build confidence.

The items above that are in scope for a "fix release" is the desired state.. what we should be able to do (with confidence). So the more we can translate the "manual testing" approaches into automated tests the better. We should allocate sufficient time/effort to perform some manual QA (and subsequently codifying the steps) before a release, so if we happen to find an issue with the "locking refactor" for instance, then we can fix it before being published.

derekcollison:
That is a fair comment, but locking order was preserved, just eliminated 2 additional lock and unlocks for the consumer. I need to probably take it off the route and gw processing but that is a 2.10 thing. Hear you about being vigilant.

ripienaar:
Agree on your points. For me I have like a few red flag code areas. Locks. Anything with interest streams and a few other bits. Those changes are always high risk. 

Very often we have had lock bugs found during code review on completely green tests by reviewers (I am terrible at finding them myself) so when I see PRs like this my immediate thought is to find ways to manually test them. Now this isn’t my role and I miss some and my manual testing isn’t exhaustive but imo in those few red areas we should, as you say, maximise our chances by not putting them in main.

bruth:
That is great perspective.. I have a [draft PR](https://github.com/nats-io/nats-server/pull/3686/files) which is intended to have a checklist for both the author and reviewers.. so we could indicate "red flag areas" to call out for more diligent review, some quick QA, or it requires two people to approve or something. Incremental steps, but awareness and getting into the habit will help.

PR #3715, opened 2022-12-15
fix and test for overlapping stream subscription on update

tbeets:
In a non-clustered JetStream system, subject overlap with other streams in the account was not checked on stream update.

/cc @nats-io/core

### DIFF
--- a/server/stream.go
+++ b/server/stream.go
@@ -413,7 +413,7 @@ func (a *Account) addStreamWithAssignment(config *StreamConfig, fsConfig *FileSt
 
 	// Check for overlapping subjects with other streams.
 	// These are not allowed for now.
-	if jsa.subjectsOverlap(cfg.Subjects) {
+	if jsa.subjectsOverlap(cfg.Subjects, nil) {
 		jsa.mu.Unlock()
 		return nil, NewJSStreamSubjectOverlapError()
 	}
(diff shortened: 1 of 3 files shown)


server/stream.go
```diff
@@ -1434,6 +1452,11 @@ func (mset *stream) update(config *StreamConfig) error {
 
 // Update will allow certain configuration properties of an existing stream to be updated.
 func (mset *stream) updateWithAdvisory(config *StreamConfig, sendAdvisory bool) error {
+	_, jsa, err := mset.acc.checkForJetStream()
```
derekcollison, line 1455: Can we assume mset.acc non nil here? Can't remember off top of my head.

server/stream.go
```diff
@@ -1444,6 +1467,13 @@ func (mset *stream) updateWithAdvisory(config *StreamConfig, sendAdvisory bool)
 		return NewJSStreamInvalidConfigError(err, Unless(err))
 	}
 
+	jsa.mu.Lock()
```
derekcollison, line 1470: This an RW? Can we use RLock?

server/stream.go
tbeets: There is an mset.account() accessor function.  Better to use that and check for nil return?

derekcollison: I think so.

server/stream.go
tbeets: Yes, we can use RLock here I will change.   subjectsOverlap() is called under Lock but that is a side-effect of single caller's RW critical section.

tbeets:
@derekcollison fixing update validation will reject existing streams on server start that already overlap.  Update is called on restore if any assignments or config have possibly changed (not deep equal).

```
[80468] 2022/12/15 10:56:42.550434 [INF]   Starting restore for stream '$G > 8576ea68-70b6-4155-a8fc-a21e86c18878'
[80468] 2022/12/15 10:56:42.550521 [WRN]   Error recreating stream "8576ea68-70b6-4155-a8fc-a21e86c18878": subjects overlap with an existing stream (10065)
```

tbeets:
Note that the issue with offending streams (overlapping sub with other stream) failing to recreate on server startup is a current **v2.9.9** issue for non-JS clustered environments only.  This PR plugs a validation hole that can create that scenario in the first place.

derekcollison (approved):
LGTM

PR #3718, opened 2022-12-15
Update nsc go install to use go modules

wallyqs:
Use v2 form to fetch the nsc tool which is now supported in that repo.

/cc @nats-io/core


philpennock (approved):

derekcollison (approved):
LGTM

PR #3719, opened 2022-12-15
Authorization Callouts

derekcollison:
An implementation for a server-signed authorization request mechanism. The mechanism will be used in both server configuration and operator mode servers. This allows external NATS applications to securely authorize users with external AuthN.

/cc @nats-io/core


server/auth.go
```diff
@@ -274,6 +274,27 @@ func (s *Server) configureAuthorization() {
...
+		_, err := s.lookupAccount(opts.AuthCallout.Account)
+		if err != nil {
+			s.Errorf("Authorization callout account %q not valid", opts.AuthCallout.Account)
+		}
+		for _, u := range opts.AuthCallout.AuthUsers {
+			// Check for user in users and nkeys since this is server config.
+			var found bool
+			if len(s.users) > 0 {
+				_, found = s.users[u]
+			} else if len(s.nkeys) > 0 && !found {
```
philpennock, line 290: Double check, this `&& ! found` has to always be true, right?

server/auth.go
```diff
@@ -557,6 +578,54 @@ func (s *Server) processClientOrLeafAuthentication(c *client, opts *Options) boo
 		err  error
 		ao   bool // auth override
 	)
+
+	// Check if we have auth callouts enabled at the server level or in the bound account.
+	defer func() {
```
philpennock, line 583: This is a lot of security-critical work to be happening inside a deferred function.  I'm worried about this being triggered after a panic.  Is this something which could be put into a new intermediary function instead?

server/auth_callout.go
```diff
@@ -0,0 +1,329 @@
...
+	c.mu.Lock()
+	c.fillClientInfo(&claim.ClientInformation)
+	c.fillConnectOpts(&claim.ConnectOptions)
+	// If we have a sig in the client opts, fill in nonce.
+	if claim.ConnectOptions.SignedNonce != _EMPTY_ {
+		claim.ClientInformation.Nonce = string(c.nonce)
+	}
+
+	// TLS
+	if c.flags.isSet(handshakeComplete) && c.nc != nil {
```
ColinSullivan1, line 224: Not necessarily for the first pass, but wondering if this should be driven by account configuration.  At scale this could result in a fair amount of extra data an auth callout wouldn't necessarily use.

server/auth.go
```diff
@@ -557,6 +578,54 @@ func (s *Server) processClientOrLeafAuthentication(c *client, opts *Options) boo
...
+			if juc == nil && opts.AuthCallout != nil {
+				authAccountName = opts.AuthCallout.Account
+			} else if juc != nil {
+				authAccountName = acc.Name
+			}
+			c.mu.Lock()
+			if c.acc != nil && c.acc.Name == authAccountName {
+				c.mergeDenyPermissions(pub, []string{AuthCalloutSubject})
+			}
+			c.mu.Unlock()
```
philpennock, line 625: If this weren't inside a huge defer doing critical work, then you could make this a more normal/safer deferred call to unlock.

server/auth_callout.go
```diff
@@ -0,0 +1,329 @@
...
+	processReply := func(_ *subscription, rc *client, racc *Account, subject, reply string, rmsg []byte) {
+		_, msg := rc.msgParts(rmsg)
+		// This signals not authorized.
+		// Since this is an account subscription will always have "\r\n".
+		if len(msg) == LEN_CR_LF {
+			respCh <- false
+			return
+		}
+		// Strip trailing CRLF.
+		msg = msg[:len(msg)-LEN_CR_LF]
```
philpennock, line 81: bug: this will panic if given a message length 1 from a misbehaving responder, won't it?

I see the comment just above and I don't follow why that might be guaranteed.  Can we just change the prior guard to `if len(msg) <= LEN_CR_LF` instead?

server/auth_callout.go
```diff
@@ -0,0 +1,329 @@
...
+
+		// If we sent an encrypted request the response could be encrypted as well.
+		if xkp != nil && len(msg) > len(jwtPrefix) && !bytes.HasPrefix(msg, []byte(jwtPrefix)) {
+			var err error
+			msg, err = xkp.Open(msg, pubAccXKey)
+			if err != nil {
+				s.Warnf("Error decrypting auth callout response: %v", err)
```
philpennock, line 88: Can this message please get `acc.Name` as a logged field too, so that we can see which account might be doing this?

If we allow authorization callouts in NGS, then a customer using them and having mistakes in their logic will be spamming our logs.  In operator mode, should this be a NATS advisory event instead of a server log?

server/auth_callout.go
```diff
@@ -0,0 +1,329 @@
...
+				respCh <- false
+				return
+			}
+		}
+
+		juc, err := jwt.DecodeUserClaims(string(msg))
+		if err != nil {
+			s.Warnf("Error decoding auth callout response: %v")
```
philpennock, line 96: Similarly here and below, we should either have `acc.Name` in the log message, to give us a chance of chasing down which account has issues, or these warnings should be events.

server/auth_callout.go
```diff
@@ -0,0 +1,329 @@
...
+		if opts.AuthCallout != nil {
+			issuer = opts.AuthCallout.Issuer
+		} else {
+			// Operator mode is who we send the request on unless switching accounts.
+			issuer = acc.Name
+		}
+		// By default issuer needs to match server config or the requesting account in operator mode.
+		if juc.Issuer != issuer {
+			if !isOperatorMode {
+				s.Warnf("Wrong issuer for auth callout response: %v")
```
philpennock, line 119: bug: you've got `%v` in the format string but aren't passing a parameter to be decoded.

server/auth_callout.go
```diff
@@ -0,0 +1,329 @@
...
+		}
+		allowNow, validFor := validateTimes(juc)
+		if !allowNow {
+			c.Errorf("Outside connect times")
+			respCh <- false
+			return
+		}
+		allowedConnTypes, err := convertAllowedConnectionTypes(juc.AllowedConnectionTypes)
+		if err != nil && len(allowedConnTypes) == 0 {
+			respCh <- false
```
philpennock, line 136: Should something be done with the error?  Perhaps an account advisory?

server/auth_callout.go
```diff
@@ -0,0 +1,329 @@
...
+			return
+		}
+		// Apply to this client.
+		targetAcc := acc
+		// Check if we are being asked to switch accounts.
+		if aname := juc.Audience; aname != _EMPTY_ {
+			targetAcc, err = s.LookupAccount(aname)
+			if err != nil {
+				s.Warnf("No valid account %q for auth callout response: %v", aname, err)
```
philpennock, line 145: Should this record `acc.Name` to identify the origin account too?

server/auth_callout.go
```diff
@@ -0,0 +1,329 @@
...
+
+		// Build internal user and bind to the targeted account.
+		nkuser := buildInternalNkeyUser(juc, allowedConnTypes, targetAcc)
+		if err := c.RegisterNkeyUser(nkuser); err != nil {
+			s.Warnf("Could not register auth callout user: %v", err)
+			respCh <- false
+			return
+		}
+
+		// See if the response wants to override the userID.
```
philpennock, line 165: This comment doesn't appear to describe the actions to be taken

server/auth_callout.go
```diff
@@ -0,0 +1,329 @@
...
+	claim.Server = jwt.ServerID{
+		Name: s.info.Name,
+		Host: s.info.Host,
+		ID:   s.info.ID,
+	}
+	// Check if we have been requested to encrypt.
+	if xkp != nil {
+		claim.Server.XKey = xkey
+	}
+	s.mu.RUnlock()
```
philpennock, line 204: This server-wide unlock can be above the XKey setting, right?

server/auth_callout.go
```diff
@@ -0,0 +1,329 @@
...
+		return false
+	}
+	req := []byte(b)
+	var hdr map[string]string
+
+	// Check if we have been asked to encrypt.
+	if xkp != nil {
+		req, err = xkp.Seal([]byte(req), pubAccXKey)
+		if err != nil {
+			s.Warnf("Error encrypting auth request claim: %v", err)
```
philpennock, line 268: Since this might vary based on the account information, should this log the account for which we tried this?  And should this be an account advisory in operator mode?

server/auth_callout.go
```diff
@@ -0,0 +1,329 @@
...
+}
+
+// Fill in client information for the request.
+// Lock should be held.
+func (c *client) fillClientInfo(ci *jwt.ClientInformation) {
+	if c == nil || (c.kind != CLIENT && c.kind != LEAF && c.kind != JETSTREAM && c.kind != ACCOUNT) {
+		return
+	}
+
+	// Do it this way to fail to compile if fields are added to jwt.ClientInformation.
```
philpennock, line 293: I don't see what would cause a compilation failure here: the other fields would just be left on their default values.

server/auth_callout.go
```diff
@@ -0,0 +1,329 @@
...
+// Fill in client options.
+// Lock should be held.
+func (c *client) fillConnectOpts(opts *jwt.ConnectOptions) {
+	if c == nil || (c.kind != CLIENT && c.kind != LEAF && c.kind != JETSTREAM && c.kind != ACCOUNT) {
+		return
+	}
+
+	o := c.opts
+
+	// Do it this way to fail to compile if fields are added to jwt.ClientInformation.
```
philpennock, line 316: Similarly

philpennock (dismissed):
Reviewed everything except `server/auth_callout_test.go`

ripienaar:
I have a few comments / wishes to add here:

 * Can we extend the client connect event to include information that this was a user authorized by a callout and how long that call out took
 * Can we have events about callout fails which might include reasons from the auth callout side
 * Server should log that this is setup in static config mode, just a warning at start saying a callout is in use
 * In this pr the nil reply is all thats accepted, I think we should consider also accepting a json document with optional error details that could be logged or included in an event (first point) to asist with debugging

server/auth_callout.go
derekcollison: Let's wait and see, not that much really compared to production systems that I have seen send 100k msgs/sec that are 4-8k.

server/accounts.go
```diff
@@ -2984,6 +2992,65 @@ func (a *Account) traceLabel() string {
 	return a.Name
 }
 
+// Check is an account has external auth set.
```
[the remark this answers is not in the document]
derekcollison: Thanks will fix.

server/auth.go
```diff
@@ -274,6 +274,27 @@ func (s *Server) configureAuthorization() {
 	s.wsConfigAuth(&opts.Websocket)
 	// And for mqtt config
 	s.mqttConfigAuth(&opts.MQTT)
+
+	// Check for server configured auth callouts.
+	if opts.AuthCallout != nil {
+		// Make sure we have a valid account and auth_users.
+		_, err := s.lookupAccount(opts.AuthCallout.Account)
```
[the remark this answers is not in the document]
derekcollison: lookupAccount can reach out. Like SYSTEM account, if you are running in operator mode, this account should also be part of the preloads. But if it isn't we will try to fetch it. If that fails we will error.

server/auth.go
```diff
@@ -274,6 +274,27 @@ func (s *Server) configureAuthorization() {
...
+		for _, u := range opts.AuthCallout.AuthUsers {
+			// Check for user in users and nkeys since this is server config.
+			var found bool
+			if len(s.users) > 0 {
+				_, found = s.users[u]
+			} else if len(s.nkeys) > 0 && !found {
+				_, found = s.nkeys[u]
+			}
+			if !found {
+				s.Errorf("Authorization callout user %q not valid", u, err)
```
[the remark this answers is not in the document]
derekcollison: Yes thx, will fix.

server/auth.go
derekcollison: That clause will only enter if nkeys > 0 and found == false.

server/auth.go
derekcollison: This needs to be a defer to be able to reference the result of authorized which is a return variable.

server/auth.go
derekcollison: 625 matches the lock on 621. Normal locking unlocking. For small blocks without branching I enclose, for more complicated stuff I try to leverge defer unlock. Defer could be used in hear but is function scoped, not block.

server/auth_callout.go
derekcollison: Account internal callbacks always have the `\r\n` which is required at the protocol layer, so would not make it this far if the message was not properly formed from a proto perspective. 

I can add more pedantic check on line 76.

server/auth_callout.go
derekcollison: Will update to add in the account.

We have the ability to rate limit these if they are exactly the same repeating very often, but would want to add that in later.

server/auth_callout.go
derekcollison: Will fix.

server/auth_callout.go
derekcollison: Changed to match behavior of this error in normal auth flow.

Not against advisories, but would want that in separate PR after some discussion and thought.

server/auth_callout.go
derekcollison: Will fix.

server/auth_callout.go
derekcollison: yes.

server/auth_callout.go
derekcollison: Will change, see above on advisories and rate limiting for server warnings etc.

server/auth_callout.go
derekcollison: The assignment part. Left side, not right side.

server/auth_callout.go
derekcollison: Same explanation.

derekcollison:
> I have a few comments / wishes to add here:
> 
> * Can we extend the client connect event to include information that this was a user authorized by a callout. Maybe include the issuer pub key.

This is visible in the account of the user and not the service, so might be considered leaking info.

> * Can we have events about callout fails which might include reasons from the auth callout side.

Phil asked about this as well. I think starting with server logs is right approach, and we can add in events. Not opposed.

> * Server should log that this is setup in static config mode, just a warning at start saying a callout is in use.

Can you give me an example of what would be helpful?

> * In this pr the nil reply is all thats accepted, I think we should consider also accepting a json document with optional error details that could be logged or included in an event (first point) to assist with debugging. JSON because we might want to extend later.

Possibly, we would need to go back into the jwt lib and add a response claim that would include the user claim such that we could trust the error response etc. So non-trivial.

> * Please include server tags in `server_id` so I can for example say some user is only allowed to connect in the eu

Good idea. I can add that, will also require update to jwt lib.

> * More logging

Examples?

ripienaar:
> > I have a few comments / wishes to add here:
> > 
> > * Can we extend the client connect event to include information that this was a user authorized by a callout. Maybe include the issuer pub key.
> 
> This is visible in the account of the user and not the service, so might be considered leaking info.

I mean the ones that get sent to `$SYS.ACCOUNT.*.CONNECT` those are system only

> 
> > * Can we have events about callout fails which might include reasons from the auth callout side.
> 
> Phil asked about this as well. I think starting with server logs is right approach, and we can add in events. Not opposed.

fair

> 
> > * Server should log that this is setup in static config mode, just a warning at start saying a callout is in use.
> 
> Can you give me an example of what would be helpful?

If a server config significantly change the authentication parameters of behavior that's definitely something you want to communicate and warn about.  It stands out so there's no inadvertant mistakes or no surprises.

> > * In this pr the nil reply is all thats accepted, I think we should consider also accepting a json document with optional error details that could be logged or included in an event (first point) to assist with debugging. JSON because we might want to extend later.
> 
> Possibly, we would need to go back into the jwt lib and add a response claim that would include the user claim such that we could trust the error response etc. So non-trivial.
> 
> > * Please include server tags in `server_id` so I can for example say some user is only allowed to connect in the eu
> 
> Good idea. I can add that, will also require update to jwt lib.

cool

> 
> > * More logging
> 
> Examples?

At debug it would for sure be good to know you're calling out and when you got a response/or not/or timeout etc. It can be hard to debug this as its a bit of effect at a distance so debug info would help developers working on auth plugins.

derekcollison:
Should also add in cluster to server portion if we are adding tags.

@ripienaar for the AuthroizationResponse, should it be code + description similar to how we evolved in JetStream?

ripienaar:
Don’t think it needs code - since this is not for internal use it’s user supplied code so would be hard to standardise on coded errors. 

Yeah cluster would be good for sure

derekcollison:
So just error as a string an empty user_claim sounds good to you?

ripienaar:
> So just error as a string an empty user_claim sounds good to you?

I thought maybe `{"error":"X"}` as a claim for now, we can also add additional fields later that way, where if we just do error string now it becomes a future pain to support both methods

server/auth.go
```diff
@@ -557,6 +578,54 @@ func (s *Server) processClientOrLeafAuthentication(c *client, opts *Options) boo
...
+		}
+		// Check if we are and in the auth callout account, and if so add in deny publish permissions for the auth subject.
+		if authorized {
+			var authAccountName string
+			if juc == nil && opts.AuthCallout != nil {
+				authAccountName = opts.AuthCallout.Account
+			} else if juc != nil {
+				authAccountName = acc.Name
+			}
+			c.mu.Lock()
```
neilalexander, line 621: It looks like the `if`/`else if` above isn't exhaustive, so if `authAccountName` isn't found for some reason, does it make sense to take out the lock and continue?

server/auth_callout.go
```diff
@@ -0,0 +1,333 @@
...
+			return
+		}
+
+		// Make sure correct issuer.
+		var issuer string
+		if opts.AuthCallout != nil {
+			issuer = opts.AuthCallout.Issuer
+		} else {
+			// Operator mode is who we send the request on unless switching accounts.
+			issuer = acc.Name
```
neilalexander, line 114: This can panic if `acc == nil`, doesn't look as though L36-47 guarantees that it can't be.

server/opts.go
```diff
@@ -3581,6 +3618,66 @@ func parseAllowedConnectionTypes(tk token, lt *token, mv interface{}, errors *[]
...
+	pm, ok := mv.(map[string]interface{})
+	if !ok {
+		return nil, &configErr{tk, fmt.Sprintf("Expected authorization callout to be a map/struct, got %+v", mv)}
+	}
+	for k, v := range pm {
+		tk, mv = unwrapValue(v, &lt)
+
+		switch strings.ToLower(k) {
+		case "issuer":
+			ac.Issuer = mv.(string)
```
neilalexander, line 3640: Might be safer for these various assertions to have `, ok =` otherwise a malformed response could result in a panic.

server/auth.go
derekcollison: The account connected to that for the static config test is checked on startup. This is just checking if we are in that account and making sure this client can not send requests on the server behalf.

server/auth_callout.go
derekcollison: Subscription callbacks are guaranteed to have an account parameter, and at top of this function we make sure acc is resolved. A client's acc is always non-nil.

server/opts.go
derekcollison: See top of function, we capture panics and resolve. I think this was @wallyqs approach..

derekcollison:
I force pushed update that has the AuthorizationResponse now. 

Still working on the correct approach to auth error events offline with the team.

ripienaar (approved):
LGTM, didnt have time to end to end test the new event but looks good

PR #3733, opened 2022-12-21
Refactor client outbound queues

neilalexander:
Previously, dynamically-sized buffers were used and discarded often in the client outbound queues. This PR removes the dynamic sizing and instead replaces them with fixed-size coalesced buffers, as well as a `sync.Pool` to enable efficient buffer reuse. The result is that the server now allocates dramatically less memory when sending data to clients.

The following results are generated by running `-bench=BenchmarkJetStreamConsume -benchtime=100000x -memprofile=`.

Memory allocations on `main` (2.9.10) as of this morning:

Screenshot 2022-12-21 at 11 17 36
Screenshot 2022-12-21 at 11 17 27

Memory allocations on this PR:

Screenshot 2022-12-21 at 11 17 53
Screenshot 2022-12-21 at 11 17 43

I don't know if this will make a huge difference to throughput in the usual case, but for servers that are pushing a lot of traffic or handling a large number of concurrent clients, this should reduce the GC burden noticeably and, with any luck, bring load averages down a little bit.

There are a couple TODOs around WebSockets but figured they would be better dealt with separately. 

/cc @nats-io/core

### DIFF
--- a/server/client.go
+++ b/server/client.go
@@ -294,19 +294,46 @@ type pinfo struct {
 
 // outbound holds pending data for a socket.
 type outbound struct {
-	p   []byte        // Primary write buffer
-	s   []byte        // Secondary for use post flush
-	nb  net.Buffers   // net.Buffers for writev IO
-	sz  int32         // limit size per []byte, uses variable BufSize constants, start, min, max.
-	sws int32         // Number of short writes, used for dynamic resizing.
+	nb  net.Buffers   // Pending buffers for send, each has fixed capacity as per nbPool below.
+	wnb net.Buffers   // Working copy of "nb", reused on each flushOutbound call, partial writes may leave entries here for next iteration.
 	pb  int64         // Total pending/queued bytes.
-	pm  int32         // Total pending/queued messages.
 	fsp int32         // Flush signals that are pending per producer from readLoop's pcd.
 	sg  *sync.Cond    // To signal writeLoop that there is data to flush.
 	wdl time.Duration // Snapshot of write deadline.
 	mp  int64         // Snapshot of max pending for client.
 	lft time.Duration // Last flush time for Write.
 	stc chan struct{} // Stall chan we create to slow down producers on overrun, e.g. fan-in.
+}
+
+const nbPoolSizeSmall = 4096  // Underlying array size of small buffer
+const nbPoolSizeLarge = 65536 // Underlying array size of large buffer
+
+var nbPoolSmall = &sync.Pool{
+	New: func() any {
+		b := [nbPoolSizeSmall]byte{}
+		return &b
+	},
+}
+
+var nbPoolLarge = &sync.Pool{
+	New: func() any {
+		b := [nbPoolSizeLarge]byte{}
+		return &b
+	},
+}
+
+func nbPoolPut(b []byte) {
+	switch cap(b) {
+	case nbPoolSizeSmall:
+		b := (*[nbPoolSizeSmall]byte)(b[0:nbPoolSizeSmall])
+		nbPoolSmall.Put(b)
+	case nbPoolSizeLarge:
+		b := (*[nbPoolSizeLarge]byte)(b[0:nbPoolSizeLarge])
+		nbPoolLarge.Put(b)
+	default:
+		// Ignore frames that are the wrong size, this might happen
+		// with WebSocket/MQTT messages as they are framed
+	}
 }
 
 type perm struct {
@@ -585,7 +612,6 @@ func (c *client) initClient() {
 	c.cid = atomic.AddUint64(&s.gcid, 1)
 
 	// Outbound data structure setup
-	c.out.sz = startBufSize
 	c.out.sg = sync.NewCond(&(c.mu))
 	opts := s.getOpts()
 	// Snapshots to avoid mutex access in fast paths.
@@ -1405,11 +1431,6 @@ func (c *client) collapsePtoNB() (net.Buffers, int64) {
 	if c.isWebsocket() {
 		return c.wsCollapsePtoNB()
 	}
-	if c.out.p != nil {
-		p := c.out.p
-		c.out.p = nil
-		return append(c.out.nb, p), c.out.pb
-	}
 	return c.out.nb, c.out.pb
 }
 
@@ -1420,9 +1441,6 @@ func (c *client) handlePartialWrite(pnb net.Buffers) {
 		c.ws.frames = append(pnb, c.ws.frames...)
 		return
 	}
-	nb, _ := c.collapsePtoNB()
-	// The partial needs to be first, so append nb to pnb
-	c.out.nb = append(pnb, nb...)
 }
 
 // flushOutbound will flush outbound buffer to a client.
@@ -1446,26 +1464,37 @@ func (c *client) flushOutbound() bool {
 		return true // true because no need to queue a signal.
 	}
 
-	// Place primary on nb, assign primary to secondary, nil out nb and secondary.
-	nb, attempted := c.collapsePtoNB()
-	c.out.p, c.out.nb, c.out.s = c.out.s, nil, nil
-	if nb == nil {
-		return true
-	}
+	// In the case of a normal socket connection, "collapsed" is just a ref
+	// to "nb". In the case of WebSockets, additional framing is added to
+	// anything that is waiting in "nb". Also keep a note of how many bytes
+	// were queued before we release the mutex.
+	collapsed, attempted := c.collapsePtoNB()
 
-	// For selecting primary replacement.
-	cnb := nb
-	var lfs int
-	if len(cnb) > 0 {
-		lfs = len(cnb[0])
-	}
+	// Frustratingly, (net.Buffers).WriteTo() modifies the receiver so we
+	// can't work on "nb" directly — while the mutex is unlocked during IO,
+	// something else might call queueOutbound and modify it. So instead we
+	// need a working copy — we'll operate on "wnb" instead. Note that in
+	// the case of a partial write, "wnb" may have remaining data from the
+	// previous write, and in the case of WebSockets, that data may already
+	// be framed, so we are careful not to re-frame "wnb" here. Instead we
+	// will just frame up "nb" and append it onto whatever is left on "wnb".
+	// "nb" will be reset back to its starting position so it can be modified
+	// safely by queueOutbound calls.
+	c.out.wnb = append(c.out.wnb, collapsed...)
+	orig := append(net.Buffers{}, c.out.wnb...)
+	c.out.nb = c.out.nb[:0]
+
+	// Since WriteTo is lopping things off the beginning, we need to remember
+	// the start position of the underlying array so that we can get back to it.
+	// Otherwise we'll always "slide forward" and that will result in reallocs.
+	startOfWnb := c.out.wnb[0:]
 
 	// In case it goes away after releasing the lock.
 	nc := c.nc
-	apm := c.out.pm
 
 	// Capture this (we change the value in some tests)
 	wdl := c.out.wdl
+
 	// Do NOT hold lock during actual IO.
 	c.mu.Unlock()
 
@@ -1477,7 +1506,7 @@ func (c *client) flushOutbound() bool {
 	nc.SetWriteDeadline(start.Add(wdl))
 
 	// Actual write to the socket.
-	n, err := nb.WriteTo(nc)
+	n, err := c.out.wnb.WriteTo(nc)
 	nc.SetWriteDeadline(time.Time{})
 
 	lft := time.Since(start)
@@ -1485,11 +1514,35 @@ func (c *client) flushOutbound() bool {
 	// Re-acquire client lock.
 	c.mu.Lock()
 
+	// At this point, "wnb" has been mutated by WriteTo and any consumed
+	// buffers have been lopped off the beginning, so in order to return
+	// them to the pool, we need to look at the difference between "orig"
+	// and "wnb".
+	for i := 0; i < len(orig)-len(c.out.wnb); i++ {
+		nbPoolPut(orig[i])
+	}
+
+	// At this point it's possible that "nb" has been modified by another
+	// call to queueOutbound while the lock was released, so we'll leave
+	// those for the next iteration. Meanwhile it's possible that we only
+	// managed a partial write of "wnb", so we'll shift anything that
+	// remains up to the beginning of the array to prevent reallocating.
+	// Anything left in "wnb" has already been framed for WebSocket conns
+	// so leave them alone for the next call to flushOutbound.
+	c.out.wnb = append(startOfWnb[:0], c.out.wnb...)
+
+	// If we've written everything but the underlying array of our working
+	// buffer has grown excessively then free it — the GC will tidy it up
+	// and we can allocate a new one next time.
+	if len(c.out.wnb) == 0 && cap(c.out.wnb) > nbPoolSizeLarge*8 {
+		c.out.wnb = nil
+	}
+
 	// Ignore ErrShortWrite errors, they will be handled as partials.
 	if err != nil && err != io.ErrShortWrite {
 		// Handle timeout error (slow consumer) differently
 		if ne, ok := err.(net.Error); ok && ne.Timeout() {
-			if closed := c.handleWriteTimeout(n, attempted, len(cnb)); closed {
+			if closed := c.handleWriteTimeout(n, attempted, len(c.out.nb)); closed {
 				return true
 			}
 		} else {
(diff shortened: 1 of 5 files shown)


neilalexander:
Investigating `TestWSCompressionWithPartialWrite` failure.

server/client.go
```diff
@@ -294,19 +294,25 @@ type pinfo struct {
...
+	sg   *sync.Cond    // To signal writeLoop that there is data to flush.
+	wdl  time.Duration // Snapshot of write deadline.
+	mp   int64         // Snapshot of max pending for client.
+	lft  time.Duration // Last flush time for Write.
+	stc  chan struct{} // Stall chan we create to slow down producers on overrun, e.g. fan-in.
+}
+
+const nbPoolSize = 512
+
+var nbPool = &sync.Pool{
```
derekcollison, line 311: We most likely will need multiple pools. The dynamic sizing on blind reads is directly responsible for raw publish performance. This will greatly reduce.

server/client.go
```diff
@@ -585,8 +591,8 @@ func (c *client) initClient() {
 	c.cid = atomic.AddUint64(&s.gcid, 1)
 
 	// Outbound data structure setup
-	c.out.sz = startBufSize
```
derekcollison, line 588: Having very small buffers helps with idle connections, but the pool may do just as well. But this could cause a run on the bank so to speak in certain situations, like sending a ping to a large number of clients who send a response. They only do this once in a while, but each response would load the buffer. It is 512 now but that is also limiting performance quite a bit.

neilalexander:
4K+64K buffers vs `main`:
```
benchmark                           old MB/s     new MB/s     speedup
Benchmark______Pub0b_Payload-16     185.50       178.70       0.96x
Benchmark______Pub8b_Payload-16     313.33       303.27       0.97x
Benchmark_____Pub32b_Payload-16     673.18       652.82       0.97x
Benchmark____Pub128B_Payload-16     1807.57      1720.96      0.95x
Benchmark____Pub256B_Payload-16     2937.77      2905.47      0.99x
Benchmark______Pub1K_Payload-16     7438.41      7205.29      0.97x
Benchmark______Pub4K_Payload-16     5349.21      5728.48      1.07x
Benchmark______Pub8K_Payload-16     4083.02      2379.72      0.58x
Benchmark_____Pub32K_Payload-16     2168.19      1626.91      0.75x
Benchmark__AuthPub0b_Payload-16     85.36        84.67        0.99x
Benchmark______FanOut_8x1x10-16     232.15       325.98       1.40x
Benchmark_____FanOut_8x1x100-16     245.85       403.99       1.64x
Benchmark____FanOut_8x10x100-16     138.72       236.03       1.70x
Benchmark___FanOut_8x10x1000-16     361.59       440.56       1.22x
Benchmark___FanOut_8x100x100-16     425.92       385.28       0.90x
Benchmark__FanOut_8x100x1000-16     428.71       422.80       0.99x
Benchmark__FanOut_8x10x10000-16     451.37       419.97       0.93x
Benchmark___FanOut_8x500x100-16     429.13       371.97       0.87x
Benchmark___FanOut_128x1x100-16     464.37       2116.05      4.56x
Benchmark__FanOut_128x10x100-16     509.55       846.38       1.66x
Benchmark_FanOut_128x10x1000-16     954.77       2317.24      2.43x
Benchmark_FanOut_128x100x100-16     1627.18      1897.50      1.17x
BenchmarkFanOut_128x100x1000-16     1816.02      2136.97      1.18x
BenchmarkFanOut_128x10x10000-16     1790.16      2157.75      1.21x
BenchmarkFanOut__128x500x100-16     1720.79      1826.24      1.06x
Benchmark_FanOut_512x100x100-16     3498.85      4873.54      1.39x
Benchmark__FanOut_512x100x1k-16     4091.18      5626.64      1.38x
Benchmark____FanOut_1kx10x1k-16     3822.01      8174.05      2.14x
Benchmark__FanOut_1kx100x100-16     3795.23      7438.82      1.96x
Benchmark_____RFanOut_8x1x10-16     107.98       303.84       2.81x
Benchmark____RFanOut_8x1x100-16     179.57       405.74       2.26x
Benchmark___RFanOut_8x10x100-16     179.14       252.19       1.41x
Benchmark__RFanOut_8x10x1000-16     443.64       453.00       1.02x
Benchmark__RFanOut_8x100x100-16     414.94       388.85       0.94x
Benchmark_RFanOut_8x100x1000-16     432.00       417.43       0.97x
Benchmark_RFanOut_8x10x10000-16     448.99       419.25       0.93x
Benchmark_RFanOut_1kx10x1000-16     3665.21      8164.32      2.23x
Benchmark_____FanIn_1kx100x1-16     3691.06      4420.21      1.20x
Benchmark_____FanIn_4kx100x1-16     5184.43      6001.12      1.16x
Benchmark_____FanIn_8kx100x1-16     5545.94      7457.00      1.34x
Benchmark____FanIn_16kx100x1-16     5594.28      7070.22      1.26x
Benchmark____FanIn_64kx100x1-16     4329.60      5892.91      1.36x
Benchmark___FanIn_128kx100x1-16     4764.75      5648.84      1.19x
Benchmark___BumpPubCount_1x3-16     112.43       294.79       2.62x
Benchmark___BumpPubCount_2x3-16     267.78       434.67       1.62x
Benchmark___BumpPubCount_5x3-16     398.54       121.10       0.30x
Benchmark__BumpPubCount_10x3-16     85.19        355.62       4.17x
Benchmark____GWs_Opt_1kx01x0-16     3645.74      3653.55      1.00x
Benchmark____GWs_Opt_2kx01x0-16     5565.68      5460.11      0.98x
Benchmark____GWs_Opt_4kx01x0-16     5558.51      5508.58      0.99x
Benchmark____GWs_Opt_1kx10x0-16     9385.95      9781.65      1.04x
Benchmark____GWs_Opt_2kx10x0-16     12544.53     12614.63     1.01x
Benchmark____GWs_Opt_4kx10x0-16     11868.34     11857.18     1.00x
Benchmark____GWs_Opt_1kx01x1-16     701.42       5266.21      7.51x
Benchmark____GWs_Opt_2kx01x1-16     918.64       7144.90      7.78x
Benchmark____GWs_Opt_4kx01x1-16     1112.30      8082.00      7.27x
Benchmark____GWs_Opt_1kx10x1-16     868.60       4851.29      5.59x
Benchmark____GWs_Opt_2kx10x1-16     1160.86      6486.34      5.59x
Benchmark____GWs_Opt_4kx10x1-16     1344.30      7582.30      5.64x
Benchmark____GWs_Int_1kx01x0-16     3637.42      3592.07      0.99x
Benchmark____GWs_Int_2kx01x0-16     5359.40      5335.23      1.00x
Benchmark____GWs_Int_4kx01x0-16     5613.72      5842.08      1.04x
Benchmark____GWs_Int_1kx10x0-16     9298.72      9134.23      0.98x
Benchmark____GWs_Int_2kx10x0-16     13618.82     12871.17     0.95x
Benchmark____GWs_Int_4kx10x0-16     11922.15     12748.66     1.07x
Benchmark____GWs_Int_1kx01x1-16     687.65       5290.96      7.69x
Benchmark____GWs_Int_2kx01x1-16     867.25       7448.40      8.59x
Benchmark____GWs_Int_4kx01x1-16     1006.29      8412.82      8.36x
Benchmark____GWs_Int_1kx10x1-16     975.18       5038.44      5.17x
Benchmark____GWs_Int_2kx10x1-16     3365.42      6686.50      1.99x
Benchmark____GWs_Int_4kx10x1-16     4201.79      7518.25      1.79x
Benchmark__GWs_Reqs_1_SubAll-16     224.87       234.15       1.04x
Benchmark__GWs_Reqs_1SubEach-16     38.13        34.25        0.90x
```

8 byte message:
```
=== RUN   TestClientOutboundQueueMemory
Message size:       0.0KB
Subscribed clients: 50000
Heap allocs before: 2733.0MB
Heap allocs after:  2925.2MB
Heap allocs delta:  7.0%
```

48000 byte message:
```
=== RUN   TestClientOutboundQueueMemory
Message size:       46.9KB
Subscribed clients: 50000
Heap allocs before: 2734.7MB
Heap allocs after:  2871.0MB
Heap allocs delta:  5.0%
```

server/client.go
```diff
@@ -294,19 +294,58 @@ type pinfo struct {
...
+	stc  chan struct{} // Stall chan we create to slow down producers on overrun, e.g. fan-in.
+}
+
+const nbPoolSizeSmall = 4096  // Underlying array size of small buffer
+const nbPoolSizeLarge = 65536 // Underlying array size of large buffer
+const nbStartCount = 128      // TODO(nat): This seems OK but maybe there is a more sane value
+
+var nbPoolSmall = &sync.Pool{
+	New: func() any {
+		b := [nbPoolSizeSmall]byte{}
```
derekcollison, line 315: Maybe just `return make([]byte, size)`

server/client.go
```diff
@@ -294,19 +294,58 @@ type pinfo struct {
...
+		return &b
+	},
+}
+
+var nbPoolLarge = &sync.Pool{
+	New: func() any {
+		b := [nbPoolSizeLarge]byte{}
```
derekcollison, line 322: Same here unless there is an advantage this way.

server/client.go
```diff
@@ -585,8 +624,8 @@ func (c *client) initClient() {
 	c.cid = atomic.AddUint64(&s.gcid, 1)
 
 	// Outbound data structure setup
-	c.out.sz = startBufSize
 	c.out.sg = sync.NewCond(&(c.mu))
+	c.out.nb = make(net.Buffers, 0, nbStartCount)
```
derekcollison, line 628: We want to optimize to no more than 8 yes?

server/client.go
```diff
@@ -1446,26 +1477,37 @@ func (c *client) flushOutbound() bool {
...
+	// In the case of a normal socket connection, "collapsed" is just a ref
+	// to "nb". In the case of WebSockets, additional framing is added to
+	// anything that is waiting in "nb". Also keep a note of how many bytes
+	// were queued before we release the mutex.
+	collapsed, attempted := c.collapsePtoNB()
+
+	// Frustratingly, (net.Buffers).WriteTo() modifies the receiver so we
+	// can't work on "nb" directly — while the mutex is unlocked during IO,
+	// something else might call queueOutbound and modify it. So instead we
+	// need a working copy — we'll operate on "wnb" instead. Note that in
```
derekcollison, line 1489: Can this just be on the stack vs part of client?

server/client.go
```diff
@@ -1477,19 +1519,38 @@ func (c *client) flushOutbound() bool {
 	nc.SetWriteDeadline(start.Add(wdl))
 
 	// Actual write to the socket.
-	n, err := nb.WriteTo(nc)
+	n, err := c.out.wnb.WriteTo(nc)
```
derekcollison, line 1522: We want to limit wnb to 8 yes for the actual calls into OS?

server/client.go
```diff
@@ -2050,6 +2079,40 @@ func (c *client) queueOutbound(data []byte) {
...
+	// Take a copy of the slice ref so that we can chop bits off the beginning
+	// without affecting the original "data" slice.
+	toBuffer := data
+
+	// All of the queued []byte have a fixed capacity, so if there's a []byte
+	// at the tail of the buffer list that isn't full yet, we should top that
+	// up first. This helps to ensure we aren't pulling more []bytes from the
+	// pool than we need to.
+	if len(c.out.nb) > 0 {
+		last := &c.out.nb[len(c.out.nb)-1]
```
derekcollison, line 2091: Why does this need to be addr? Seems its re-assigned below but seemingly with no effect?

server/client.go
```diff
@@ -2050,6 +2079,40 @@ func (c *client) queueOutbound(data []byte) {
...
+			toBuffer = toBuffer[free:]
+		}
+	}
+
+	// Now we can push the rest of the data into new []bytes from the pool
+	// in fixed size chunks. This ensures we don't go over the capacity of any
+	// of the buffers and end up reallocating.
+	for len(toBuffer) > 0 {
+		new := nbPoolGet(len(toBuffer))
+		//new := nbPool.Get().(*[nbPoolSize]byte)[:0]
```
derekcollison, line 2106: Remove

server/client.go
```diff
@@ -2050,6 +2079,40 @@ func (c *client) queueOutbound(data []byte) {
...
+			*last = append(*last, toBuffer[:free]...)
+			toBuffer = toBuffer[free:]
+		}
+	}
+
+	// Now we can push the rest of the data into new []bytes from the pool
+	// in fixed size chunks. This ensures we don't go over the capacity of any
+	// of the buffers and end up reallocating.
+	for len(toBuffer) > 0 {
+		new := nbPoolGet(len(toBuffer))
```
derekcollison, line 2105: If we have existing nb, we should just do another 64k and not fragment. Should only be small iff no other pending imo.

server/client.go
```diff
@@ -3310,8 +3321,6 @@ func (c *client) deliverMsg(prodIsMQTT bool, sub *subscription, acc *Account, su
 		client.queueOutbound([]byte(CR_LF))
 	}
 
-	client.out.pm++
```
derekcollison, line 3313: Might be good to track for reporting?

server/client.go
```diff
@@ -3326,7 +3335,7 @@ func (c *client) deliverMsg(prodIsMQTT bool, sub *subscription, acc *Account, su
 	// to intervene before this producer goes back to top of readloop. We are in the producer's
 	// readloop go routine at this point.
 	// FIXME(dlc) - We may call this alot, maybe suppress after first call?
-	if client.out.pm > 1 && client.out.pb > maxBufSize*2 {
+	if len(client.out.nb) > 1 {
```
derekcollison, line 3338: != 0 maybe preferred?

derekcollison (changes requested):

server/client.go
neilalexander: There's two reasons for doing it this way:
1. It allows us to return a pointer, which is essential for `sync.Pool` to make a difference — if you send a slice into a `sync.Pool`, the slice header itself ends up being copied onto the heap and this results in more allocations
2. It ensures we always have a reference to the beginning of the underlying array, which is important when coalescing

server/client.go
neilalexander: 8 comes from the quick path for `writev`, but we can still save on some potential re-allocations by ensuring we have a reasonable amount of space after we've exceeded those 8 values. This is just an array of `[][]byte` references so they are compact, even with 128 of them.

server/client.go
neilalexander: I think by limiting to 8 buffers before having to call `flushOutbound` again, we're limiting the potential for `writev` to send potentially more than that if it can. The path in `writev` is slightly slower for more than a certain number of buffers (as it requires heap allocations in the kernel) but it still completes in a single context switch regardless, so it's less expensive on syscalls.

server/client.go
neilalexander: This is the coalescing in action — at this point, `c.out.nb` might still have things queued up from a previous write, so we take the reference to the last buffer, see whether there's any remaining space that we can "top up" with this write and append that many bytes onto the existing buffer before we allocate a new one.

i.e. if we have an 8K buffer, the last write was only 7K, we have 1K left, so we can write another 1K onto the end of that buffer before we need to pull another one out of the pool.

server/client.go
neilalexander: By this point we've already "filled up" the existing last buffer in `nb` and we will only allocate new buffers if there's still something we've yet to queue up. That way we can reduce the number of vectors to `writev` for lots of small successive messages.

server/client.go
derekcollison: Yes bit what I am saying is that for a call into the kernel let's limit to 8, so that would most likely be 8*64k or 256k.. WDYT?

server/client.go
neilalexander: The reason for having it on the client is that we can reuse the underlying array by just trimming it down with `wnb[:0]` and store it for next time around which saves on some more allocations.

server/client.go
derekcollison: ok reason I bring it up is seeing us get stuck in certain situations calling in with 1000s of entries and like 100s of MBs of data. I think we want to avoid that maybe if we can.

server/client.go
derekcollison: The small buffers should only be used if no pending at all imo, for the ping to all issue. If there is any pending they should all be 64k. Since that means the publisher is publishing at a fast rate.

server/client.go
derekcollison: So we buffer in two spots, pool and wnb?

derekcollison: Can these be cleaned up if client goes back to idle?

server/client.go
neilalexander: I can modify it to ensure that later buffers are always large ones, sure.

server/client.go
neilalexander: `wnb` is reused each time `flushOutbound` calls. Actually it's a good point that right now we don't check if the capacity of `wnb` has grown uncontrollably so I'll add something to check for that.

server/client.go
neilalexander: Yep, good point. I'll have a think on how best to handle that.

server/client.go
derekcollison: Yes, so if the message to be queued is small, and if there is no pending get a small buffer if the message will fit. Otherwise always do 64k.

server/client.go
neilalexander: I've done some researching on this and it appears there's no benefit, when heavily loaded we end up just making more syscalls so the time lost to context switching is worse than the gain from hitting the `writev` fast path. :-(

server/client.go
```diff
@@ -585,8 +613,8 @@ func (c *client) initClient() {
 	c.cid = atomic.AddUint64(&s.gcid, 1)
 
 	// Outbound data structure setup
-	c.out.sz = startBufSize
 	c.out.sg = sync.NewCond(&(c.mu))
+	c.out.nb = make(net.Buffers, 0, nbStartCount)
```
derekcollison, line 617: Would leave this nil for now and will be allocated when we add to this imo.

test/cluster_test.go
```diff
@@ -248,6 +252,10 @@ func TestClusterQueueSubs(t *testing.T) {
 	// Send to A
 	sendA("PUB foo 2\r\nok\r\n")
 
+	// Give plenty of time for the messages to flush, so that we don't
+	// accidentally only read some of them.
+	time.Sleep(time.Second / 4)
```
derekcollison, line 257: Just 250 * time.Millisecond?

derekcollison (changes requested):

neilalexander:
@derekcollison Review comments addressed!

derekcollison (approved):
LGTM!

derekcollison (approved):
LGTM

PR #3744, opened 2022-12-26
Add back existing HealthzOptions.JSEnabled field

bruth:
This fixes a backwards incompat change introduced in #3704 for library usage as well as using the healthz NATS API which depends on the JSON payload.

/cc @nats-io/core


ripienaar (approved):
LGTM

derekcollison (approved):
LGTM

PR #3752, opened 2022-12-30
Healthz works on single server systems and allow recovery if stream meta not flushed.

derekcollison:
If a stream's meta was not properly written but the file existed, we could re-add the stream but a subsequent restart would lose the stream again.

Also added in healthz for single server systems to make sure all stream directories resulted in recovered streams.

/cc @nats-io/core


wallyqs (approved):
LGTM

neilalexander (approved):

PR #3754, opened 2022-12-30
Switch to `golangci-lint`

neilalexander:
This replaces the separate linters in `runTestsOnTravis.sh` with the more modern `golangci-lint`. This is the recommended approach now and also means that the `gopls` language server stands a much better chance of picking up the same issues that CI will.

It also adds a `.golangci-lint.yml` file to control the linter behaviours, as well as fixing a few things in the code that it yelled about.

There are more linters we can enable in the future if we want to — I commented out a few but they can be addressed in separate PRs.

/cc @nats-io/core

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -22,7 +22,7 @@ linters:
   disable-all: true
   enable:
     # - errcheck
-    # - gofmt
+    - gofmt
     # - goimports
     - gosimple
     - govet
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,0 +1,60 @@
+# Config file for golangci-lint
+run:
+  concurrency: 4
+  deadline: 1m
+  issues-exit-code: 1
+  tests: true
+  skip-dirs:
+    - .github
+    - doc
+    - docker
+    - logos
+    - scripts
+    - util
+  modules-download-mode: readonly
+
+output:
+  format: colored-line-number
+  print-issued-lines: true
+  print-linter-name: true
+
+linters:
+  disable-all: true
+  enable:
+    # - errcheck
+    # - gofmt
+    # - goimports
+    - gosimple
+    - govet
+    - ineffassign
+    # - maligned
+    - megacheck
+    - misspell
+    # - prealloc
+    - staticcheck
+    # - unparam
+    - unused
+
+linters-settings:
+  errcheck:
+    check-type-assertions: false
+    check-blank: false
+  govet:
+    check-shadowing: false
+    settings:
+      printf:
+        funcs:
+          - (github.com/golangci/golangci-lint/pkg/logutils.Log).Infof
+          - (github.com/golangci/golangci-lint/pkg/logutils.Log).Warnf
+          - (github.com/golangci/golangci-lint/pkg/logutils.Log).Errorf
+          - (github.com/golangci/golangci-lint/pkg/logutils.Log).Fatalf
+  gofmt:
+    simple: true
+  misspell:
+    locale: US
+  unparam:
+    check-exported: false
+  prealloc:
+    simple: true
+    range-loops: true
+    for-loops: true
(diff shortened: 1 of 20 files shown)


scripts/runTestsOnTravis.sh
```diff
@@ -3,17 +3,12 @@
...
-    GO_LIST=$(go list ./...);
+    # First check that NATS builds.
     go build;
-    $(exit $(go fmt $GO_LIST | wc -l));
-    go vet $GO_LIST;
-    find . -type f -name "*.go" | xargs misspell -error -locale US;
-    staticcheck -tags=js_chaos_tests $GO_LIST
+
+    # Now run the linters.
+    curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(go env GOPATH)/bin v1.50.1
```
wallyqs, line 10: we could do `go install github.com/golangci/golangci-lint/cmd/golangci-lint@latest` but this binary is about 30mb so prefer installing via the script too and just getting the release

wallyqs (approved):
LGTM

scripts/runTestsOnTravis.sh
neilalexander: Yeah, I did consider using `go install` but the `golangci-lint` documentation advises against it, mostly because it's considerably slower than just fetching a binary.

.golangci.yml
```diff
@@ -0,0 +1,60 @@
...
+output:
+  format: colored-line-number
+  print-issued-lines: true
+  print-linter-name: true
+
+linters:
+  disable-all: true
+  enable:
+    # - errcheck
+    # - gofmt
```
derekcollison, line 25: Why disabled?

scripts/runTestsOnTravis.sh
derekcollison: How much time does it add to the travis runs?

.golangci.yml
derekcollison: We should run gofmt.

server/jetstream_test.go
```diff
@@ -7691,6 +7691,9 @@ func TestJetStreamRequestAPI(t *testing.T) {
 
 	// Make sure list names works.
 	resp, err = nc.Request(JSApiStreams, nil, time.Second)
+	if err != nil {
```
derekcollison, line 7694: We do `require_NoError(t, err)` now.

server/jetstream_test.go
```diff
@@ -7714,6 +7717,9 @@ func TestJetStreamRequestAPI(t *testing.T) {
 
 	// Now do detailed version.
 	resp, err = nc.Request(JSApiStreamList, nil, time.Second)
+	if err != nil {
```
derekcollison, line 7720: Same

scripts/runTestsOnTravis.sh
neilalexander: About 20-30 seconds.

.golangci.yml
neilalexander: I should have added a comment here — it's mostly because of a bug in `golangci-lint` on macOS Ventura where it fails unless you `brew install diffutils`.

The issue is fixed upstream but there's no tagged release that includes it yet. I guess we could use the `go install` install method to pin a more recent commit of `golangci-lint` and then enable `gofmt`? (I just didn't want to break anyone!)

scripts/runTestsOnTravis.sh
derekcollison: ok will check the timing differences on your current run. Thanks.

.golangci.yml
derekcollison: Yes that might be good, ci should have go fmt.

Or, use old method for now until it lands proper here?

derekcollison (approved):
LGTM!

PR #3760, opened 2023-01-05
Ensure we add in new consumers from a meta snapshot from the leader.

derekcollison:
This could lead to missing consumers for R1s that were assigned to a server while it was down and on restart the metaleader had snapshotted it's meta state.

/cc @nats-io/core


Jarema (approved):
LGTM

PR #3763, opened 2023-01-05
Set and clear observer state for servers with disconnected leafnodes.

derekcollison:
When migrating leaders off a server when the leafnode is not connected, also ensure leaders can not return until reconnected.

/cc @nats-io/core


neilalexander (approved):
Looks reasonable to me!

Issue #3730, opened 2022-12-20
Improve memory footprint for subscriptions


bruth:

bruth:
Closing this since some improvements were made in scope in 2.10. More detailed proposals will be written up for future optimizations.

derekcollison:
We actually have not too much here and is still on my list..

PR #3768, opened 2023-01-06
Deduplicate `*subscription` in `Sublist`

neilalexander:
Currently we are storing the same `*subscription` pointer both in the map keys and values, which seems like an unnecessary duplication. This replaces the map value with `struct{}` as it has an effective width of zero in memory.

cc #3730

/cc @nats-io/core


derekcollison (approved):
LGTM!

PR #3774, opened 2023-01-10
Add `profilez` server endpoint

neilalexander:
This PR adds a `profilez` endpoint to the system account, which accepts two parameters: `profile` for the name of the profile (i.e. `allocs`, `block`, `goroutine`, `heap`, `mutex`, `threadcreate`) and `debug` for the return format (defaults to `0`, which is usually gzipped for `go tool pprof`).

It will only be possible to query this endpoint if `PROF_PORT` is already enabled, otherwise it will return a `Profiling is not enabled` error.

/cc @nats-io/core

### DIFF
--- a/server/monitor.go
+++ b/server/monitor.go
@@ -14,6 +14,7 @@
 package server
 
 import (
+	"bytes"
 	"crypto/sha256"
 	"crypto/tls"
 	"crypto/x509"
@@ -26,6 +27,7 @@ import (
 	"os"
 	"path/filepath"
 	"runtime"
+	"runtime/pprof"
 	"sort"
 	"strconv"
 	"strings"
@@ -2639,6 +2641,12 @@ type HealthzOptions struct {
 	JSServerOnly  bool `json:"js-server-only,omitempty"`
 }
 
+// ProfilezOptions are options passed to Profilez
+type ProfilezOptions struct {
+	Name  string `json:"name"`
+	Debug int    `json:"debug"`
+}
+
 type StreamDetail struct {
 	Name     string              `json:"name"`
 	Created  time.Time           `json:"created"`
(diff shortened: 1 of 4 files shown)


ripienaar:
We also have configuration option in the server to enable it at runtime and when embedded. This kind of thing I think should be an opt-in as it can have runtime performance impact on the servers

neilalexander:
> This kind of thing I think should be an opt-in as it can have runtime performance impact on the servers

So far the only thing that `StartProfiler` does that has any overhead is to call `runtime.SetBlockProfileRate`, which `profilez` does not do admittedly, so maybe we can just guard off the `block` profile unless `StartProfiler` has been called.

There's no runtime overhead on things like `allocs`, `goroutine`, `heap`, `threadcreate` as they just are internally extracting statistics out of the `runtime` package.

ripienaar:
There's certainly some cost associated with gathering a profile though - if its same as when you hit the normal pprof end points the go team aims to keep the impact aroun 5%, its still an impact though and a potential vector for abuse that someone might want to turn off.

neilalexander:
I should have probably noted in the original PR that it isn't actually possible to collect a CPU profile via `profilez` right now (like you would with `/debug/pprof/profile`) because the `pprof` package registers that straight onto `DefaultServeMux` and not into the `profiles` struct, so the only accessible profiles [are these](https://cs.opensource.google/go/go/+/refs/tags/go1.19.4:src/runtime/pprof/pprof.go;l=182-195;drc=0202ad0b3a2bfddf9f3eafb94e19d5a0fa3d1f31). 

I can make it opt-in, although it doesn't look as though the monitoring port actually has a configuration section of its own, so I guess I could do one of the following:
1. Add a `http_profiling` option, to control whether or not it should be exposed on both the monitoring port and the system account;
2. Only expose it on the system account only (and not the HTTP port) when `PROF_PORT` is already enabled?

ripienaar:
2nd options sounds good - the current `/debug/pprof` stuff is nice cos its just exactly what `go tool pprof` wants for real time debugging etc. so for HTTP I think its great as is, then just additionally expose on system account once opted in is great

server/monitor.go
```diff
@@ -3123,3 +3131,36 @@ func (s *Server) healthz(opts *HealthzOptions) *HealthStatus {
...
+	Error   string `json:"error"`
+}
+
+func (s *Server) profilez(opts *ProfilezOptions) *ProfilezStatus {
+	if s.profiler == nil {
+		return &ProfilezStatus{
+			Error: "Profiling is not enabled",
+		}
+	}
+	if opts.Name == "" {
```
derekcollison, line 3146: Use _EMPTY_, old pattern but want to stay consistent.

server/monitor.go
```diff
@@ -3123,3 +3131,36 @@ func (s *Server) healthz(opts *HealthzOptions) *HealthStatus {
...
+		}
+	}
+	var buffer bytes.Buffer
+	if err := profile.WriteTo(&buffer, opts.Debug); err != nil {
+		return &ProfilezStatus{
+			Error: fmt.Sprintf("Profile %q error: %s", opts.Name, err),
+		}
+	}
+	return &ProfilezStatus{
+		Profile: buffer.Bytes(),
```
derekcollison, line 3164: We support optional compression, we should make sure that is supported here, WDYT?

server/monitor_test.go
```diff
@@ -4548,3 +4548,38 @@ func TestMonitorWebsocket(t *testing.T) {
...
+		{Name: "block", Debug: 0},
+		{Name: "goroutine", Debug: 0},
+		{Name: "goroutine", Debug: 1},
+		{Name: "goroutine", Debug: 2},
+		{Name: "heap", Debug: 0},
+		{Name: "heap", Debug: 1},
+		{Name: "mutex", Debug: 0},
+		{Name: "threadcreate", Debug: 0},
+	} {
+		if ps := s.profilez(try); ps.Error != "" {
```
derekcollison, line 4581: _EMPTY_

derekcollison (changes requested):

derekcollison:
Please do a rebase onto dev vs merge.

PR #3780, opened 2023-01-12
Use highwayhash for last stream, consumer and cluster snapshots

neilalexander:
These snapshots can grow quite large, especially in the case of lots of interior deletes on streams, lots of pending ACKs on consumers etc. By using hashes instead, we can determine whether or not the snapshot has changed without having to hold the entire snapshot on the heap.

/cc @nats-io/core

### DIFF
--- a/server/jetstream_cluster.go
+++ b/server/jetstream_cluster.go
@@ -30,6 +30,7 @@ import (
 	"time"
 
 	"github.com/klauspost/compress/s2"
+	"github.com/minio/highwayhash"
 	"github.com/nats-io/nuid"
 )
 
@@ -949,6 +950,10 @@ func (js *jetStream) monitorCluster() {
 		beenLeader   bool
 	)
 
+	// Highwayhash key for generating hashes.
+	key := make([]byte, 32)
+	rand.Read(key)
+
 	// Set to true to start.
 	js.setMetaRecovering()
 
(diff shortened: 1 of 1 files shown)


server/jetstream_cluster.go
```diff
@@ -949,6 +951,9 @@ func (js *jetStream) monitorCluster() {
 		beenLeader   bool
 	)
 
+	// Highwayhash key for generating hashes.
+	key := sha256.Sum256([]byte("clusters"))
```
derekcollison, line 955: Can just create random bytes here, or if you prefer just use sha256 vs HH? HH is quick.

server/jetstream_cluster.go
```diff
@@ -1812,16 +1818,21 @@ func (js *jetStream) monitorStream(mset *stream, sa *streamAssignment, sendSnaps
 	}
 	accName := acc.GetName()
 
+	// Hash of the last snapshot (fixed size in memory).
 	var lastSnap []byte
 
+	// Highwayhash key for generating hashes.
+	key := sha256.Sum256([]byte("streams"))
```
derekcollison, line 1825: Same as above.

server/jetstream_cluster.go
```diff
@@ -3936,11 +3951,13 @@ func (js *jetStream) monitorConsumer(o *consumer, ca *consumerAssignment) {
...
 
-		if snap, err := o.store.EncodedState(); err == nil && (!bytes.Equal(lastSnap, snap) || needSnap) {
-			if err := n.InstallSnapshot(snap); err == nil {
-				lastSnap, lastSnapTime = snap, time.Now()
-			} else {
-				s.Warnf("Failed to install snapshot for '%s > %s > %s' [%s]: %v", o.acc.Name, ca.Stream, ca.Name, n.Group(), err)
+		if snap, err := o.store.EncodedState(); err == nil {
+			if hash := highwayhash.Sum(snap, key[:]); !bytes.Equal(hash[:], lastSnap) || needSnap {
+				if err := n.InstallSnapshot(snap); err == nil {
+					lastSnap, lastSnapTime = hash[:], time.Now()
```
derekcollison, line 3957: Do same one liner for stream which is 2 lines IIRC?

derekcollison (approved):
LGTM!

PR #3781, opened 2023-01-12
Tweak logic for sending leaf node subscription updates

neilalexander:
Currently the logic here doesn't feel too tight — currently we are only checking movement upward if `n` is exactly 0, and movement downward can be triggered again even if we're already below 0.

This new line should be clearer and more explicit that we're trying to capture *transitions* to/from +ve and -ve only.

/cc @nats-io/core


derekcollison (approved):
LGTM

PR #3783, opened 2023-01-13
ADDED hard_delete option on resolver configuration

### LINKED ISSUE #3782: Add option to the nats resolver to hard delete the jwt instead of renaming them to `.deleted`.
## Feature Request
Add option to the nats resolver to hard delete the jwt instead of renaming them to `.deleted`.

I'll propose and MR to implement the option listed bellow.

#### Use Case:
We manage the accounts in our database so if we want to recover them we can use the database so we don't need the copy on disk.
We can do some manual (or cron) cleanup but as the nats code already has an HardDelete option internally, why not expose it to the configuration.

#### Proposed Change:
Add an option in the resolver config, for instance: `hard_delete: bool`. 

Any user that do not need the `.deleted` files.

#### Alternative Approaches
Overloading `allow_delete: hard_delete` but that may be problematic for the parser and backward compatibility

JulienVdG:
 - [ ] Tests added

Resolves #3782

 - Add hard_delete option in the resolver config and use it to set the deleteType in NewDirAccResolver

/cc @nats-io/core

### DIFF
--- a/server/accounts.go
+++ b/server/accounts.go
@@ -4049,18 +4049,14 @@ func (dr *DirAccResolver) apply(opts ...DirResOption) error {
 	return nil
 }
 
-func NewDirAccResolver(path string, limit int64, syncInterval time.Duration, delete bool, opts ...DirResOption) (*DirAccResolver, error) {
+func NewDirAccResolver(path string, limit int64, syncInterval time.Duration, delete deleteType, opts ...DirResOption) (*DirAccResolver, error) {
 	if limit == 0 {
 		limit = math.MaxInt64
 	}
 	if syncInterval <= 0 {
 		syncInterval = time.Minute
 	}
-	deleteType := NoDelete
-	if delete {
-		deleteType = RenameDeleted
-	}
-	store, err := NewExpiringDirJWTStore(path, false, true, deleteType, 0, limit, false, 0, nil)
+	store, err := NewExpiringDirJWTStore(path, false, true, delete, 0, limit, false, 0, nil)
 	if err != nil {
 		return nil, err
 	}
--- a/server/opts.go
+++ b/server/opts.go
@@ -1130,6 +1130,8 @@ func (o *Options) processConfigFileLine(k string, v interface{}, errors *[]error
 			}
 		case map[string]interface{}:
 			del := false
+			hdel := false
+			hdel_set := false
 			dir := ""
 			dirType := ""
 			limit := int64(0)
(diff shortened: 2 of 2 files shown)


derekcollison:
Need to fix test failure..

JulienVdG:
on my computer it fails event on main before my changes :(

derekcollison:
The tests pass on main proper, and on your branch when I kept restarting yours it kept failing on the same test.

We do have some flappers but I think there is something legit here.

JulienVdG:
OK, Thanks for checking I'm going to retry and fix that ;)

derekcollison:
Any updates? We are starting prep for a 2.9.12 release.

JulienVdG:
No sorry I don't get why it fails: I though it would be just a config change but there might be strange side effects.
I tried several changes but even if the changes should not affect the default configuration the tests failed on Travis (On my computer they fail at different stages depending on the options I set to go test like -parallel ... rerunning the failing test alone always work)
Don't wait for me for your release, I'll revisit this MR later.
Thank you !

derekcollison:
ok, and FYI we will be moving away from the NATS resolver in the near future for a JetStream KV based resolver.

JulienVdG:
> FYI we will be moving away from the NATS resolver in the near future for a JetStream KV based resolver.
oh nice, thanks !

goku321:
> ok, and FYI we will be moving away from the NATS resolver in the near future for a JetStream KV based resolver.

Is there a doc/ADR for this? Would like to know more about this.

derekcollison:
There is not as of yet, but in the simple form you would store account claims in a KV that the server would be configured to use to resolve accounts through KV gets and watchers.

KVs today can already be set up with as many mirrors as you want and KV gets will use the closest replica to respond to the request.

We will convert our own global network NGS to it once complete.

goku321:
Thanks for the info. Good to see the JetStream KV store being used in exciting ways!

derekcollison:
We are prepping for 2.9.16 release, do you want this reviewed again for possible inclusion?

derekcollison (commented):
Where is the behavior enforced?

JulienVdG:
> Where is the behavior enforced?

The behavior **was already present** in the DirJWTStore implementation (it's used for the caching resolver: NewCacheDirAccResolver hardcode the delete parameter to `HardDelete`).

So this PR only allows to use this parameter when creating the DirJWTStore via NewDirAccResolver and does the necessary checks on the configuration options (with hopefully clear enough error messages).

PR #3784, opened 2023-01-13
[CHANGE] [AC] flip aud and sub for consistency

aricart:
The auth request, the `sub` for the JWT is `nats-authorization-request`, and the `aud` is the nkey, typically our JWTs have `sub` be an identity and the `aud` is just a string describing where this jwt is applicable.

/cc @nats-io/core

### DIFF
--- a/server/auth_callout.go
+++ b/server/auth_callout.go
@@ -204,8 +204,13 @@ func (s *Server) processClientOrLeafCallout(c *client, opts *Options) (authorize
 	}
 	defer acc.unsubscribeInternal(sub)
 
-	// Build our request claims.
-	claim := jwt.NewAuthorizationRequestClaims(AuthRequestSubject)
+	// Build our request claims - jwt subject should be nkey
+	jwtSub := acc.Name
+	if opts.AuthCallout != nil {
+		jwtSub = opts.AuthCallout.Issuer
+	}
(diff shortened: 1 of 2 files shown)


neilalexander:
Looks like the target branch should be `dev` and not `main`?

aricart:
> Looks like the target branch should be `dev` and not `main`?

correct and vendor stuff too - we should default to dev.

server/auth_callout.go
```diff
@@ -204,8 +204,13 @@ func (s *Server) processClientOrLeafCallout(c *client, opts *Options) (authorize
 	}
 	defer acc.unsubscribeInternal(sub)
 
-	// Build our request claims.
-	claim := jwt.NewAuthorizationRequestClaims(AuthRequestSubject)
+	// Build our request claims - jwt subject should be nkey
+	jwtSub := acc.Name
```
aricart, line 208: Also wondering if this really should `acc.NKey` and name is a last resort.

derekcollison:
Let's redo this to target dev..

aricart:
> Let's redo this to target dev..

it is

derekcollison (approved):
LGTM - Flip out of draft mode and we can merge.

derekcollison (approved):
LGTM

PR #3794, opened 2023-01-20
Add small tolerance in stream health check

neilalexander:
If messages are being async published faster than the cluster can apply Raft updates, then the `HEALTHZ` check will report unhealthy, even if we are still making forward progress. This might cause problems in instances where hosting infrastructure will restart the NATS Server if it reports an unhealthy state for a certain period of time, for example.

Adding a slight tolerance to the health check should improve this situation and report only if we are provably not making forward progress. There's also a unit test that demonstrates the problem.

/cc @nats-io/core

### DIFF
--- a/server/raft.go
+++ b/server/raft.go
@@ -47,6 +47,7 @@ type RaftNode interface {
 	Leader() bool
 	Quorum() bool
 	Current() bool
+	Healthy() bool
 	Term() uint64
 	GroupLeader() string
 	HadPreviousLeader() bool
@@ -1160,13 +1161,15 @@ func (n *raft) isCatchingUp() bool {
 	return n.catchup != nil
 }
 
-// Lock should be held.
-func (n *raft) isCurrent() bool {
-	// First check if have had activity and we match commit and applied.
-	if n.commit == 0 || n.commit != n.applied {
-		n.debug("Not current, commit %d != applied %d", n.commit, n.applied)
+// Lock should be held. This function may block for up to ~5ms to check
+// forward progress in some cases.
+func (n *raft) isCurrent(includeForwardProgress bool) bool {
+	// Check whether we've made progress on any state, 0 is invalid so not healthy.
+	if n.commit == 0 {
+		n.debug("Not current, no commits")
 		return false
 	}
+
 	// Make sure we are the leader or we know we have heard from the leader recently.
 	if n.state == Leader {
 		return true
@@ -1181,14 +1184,40 @@ func (n *raft) isCurrent() bool {
 	if n.leader != noLeader && n.leader != n.id && n.catchup == nil {
 		okInterval := int64(hbInterval) * 2
 		ts := time.Now().UnixNano()
-		if ps := n.peers[n.leader]; ps != nil && ps.ts > 0 && (ts-ps.ts) <= okInterval {
-			return true
+		if ps := n.peers[n.leader]; ps == nil || ps.ts == 0 && (ts-ps.ts) > okInterval {
+			n.debug("Not current, no recent leader contact")
+			return false
 		}
-		n.debug("Not current, no recent leader contact")
 	}
 	if cs := n.catchup; cs != nil {
 		n.debug("Not current, still catching up pindex=%d, cindex=%d", n.pindex, cs.cindex)
 	}
+
+	if n.commit == n.applied {
+		// At this point if we are current, we can return saying so.
+		return true
+	} else if !includeForwardProgress {
+		// Otherwise, if we aren't allowed to include forward progress
+		// (i.e. we are checking "current" instead of "healthy") then
+		// give up now.
+		return false
+	}
+
+	// Otherwise, wait for a short period of time and see if we are making any
+	// forward progress.
+	if startDelta := n.commit - n.applied; startDelta > 0 {
+		for i := 0; i < 10; i++ { // 5ms, in 0.5ms increments
+			n.RUnlock()
+			time.Sleep(time.Millisecond / 2)
+			n.RLock()
+			if n.commit-n.applied < startDelta {
+				// The gap is getting smaller, so we're making forward progress.
+				return true
+			}
+		}
+	}
+
+	n.warn("Falling behind in health check, commit %d != applied %d", n.commit, n.applied)
 	return false
 }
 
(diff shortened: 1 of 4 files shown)


jnmoyne:
This is a good improvement 👍. In my experiments during benchmarks I have noticed this behavior. Empirical observation showed that (in my case) it could fall behind for up to a few tens of messages and then quickly catch up, so 5 may be a little bit too low. That is falls behind a bit temporarily is quite normal if the publisher is bursty, what's bad is if the number of messages behind keeps increasing over time.

derekcollison:
Neil and I discussed and he is updating to just verify we are making progress on applied since this is all local to the server. But you raise a good point, which is we should make sure that applied is increasing but should also check that the delta between committed and applied is decreasing before return ok to healthy.

server/raft.go
```diff
@@ -1160,13 +1161,40 @@ func (n *raft) isCatchingUp() bool {
...
+	// First check if have had activity and that commit and applied are equal.
+	// If not, wait for a short period of time and see if we are making forward
+	// progress.
+	startCommit, startDelta := n.commit, n.commit-n.applied
+	if startCommit == 0 || startDelta > 0 {
+		var progress bool
+		n.RUnlock()
+		for i := 0; i < 10; i++ { // 5ms, in 0.5ms increments
+			time.Sleep(time.Millisecond / 2)
+			n.RLock()
```
derekcollison, line 1176: Let's just simplify locking and put unlock just around the sleep.

n.RUnlock()
time.Sleep()
n.RLock()

We also need to make sure we are holding the RLock and not Lock..

derekcollison (changes requested):

server/raft.go
neilalexander: You're right, that's simpler — done.

derekcollison: Are we guaranteed to only be holding the RLock?

neilalexander: Both the call sites for `isCurrent` are taking out RLocks only.

server/raft.go
```diff
@@ -1160,13 +1161,38 @@ func (n *raft) isCatchingUp() bool {
...
+				break
+			}
+			if startCommit < curCommit && curDelta <= startDelta {
+				// New work has come in and the gap is either getting smaller
+				// or has stayed the same, so we're still making progress.
+				break
+			}
+		}
+		if !progress {
+			n.debug("Falling behind, commit %d != applied %d", n.commit, n.applied)
```
derekcollison, line 1191: Should this be a warn maybe?

neilalexander: It was debug level before, if we want to change it to warn, is there more information we should include?

derekcollison: Say it is inside health check logic, etc?

neilalexander: Done :-)

server/raft.go
```diff
@@ -1160,13 +1161,38 @@ func (n *raft) isCatchingUp() bool {
...
+			if startCommit == curCommit && curDelta < startDelta {
+				// Nothing new has come in but the gap is getting smaller, so
+				// we're making progress on the work we already had.
+				progress = true
+				break
+			}
+			if startCommit < curCommit && curDelta <= startDelta {
+				// New work has come in and the gap is either getting smaller
+				// or has stayed the same, so we're still making progress.
+				break
```
derekcollison, line 1187: Should progress be set to true here?

neilalexander: Ah, yes, good spot. I've fixed it.

server/raft.go
```diff
@@ -1160,13 +1161,39 @@ func (n *raft) isCatchingUp() bool {
...
+	// If not, wait for a short period of time and see if we are making forward
+	// progress.
+	startCommit, startDelta := n.commit, n.commit-n.applied
+	if startCommit == 0 || startDelta > 0 {
+		var progress bool
+		for i := 0; i < 10; i++ { // 5ms, in 0.5ms increments
+			n.RUnlock()
+			time.Sleep(time.Millisecond / 2)
+			n.RLock()
+			curCommit, curDelta := n.commit, n.commit-n.applied
```
derekcollison, line 1177: I think this can be simplified to a simple, is the delta smaller, wdyt?

Just track delta and make sure it descreases.

neilalexander: Originally I checked that the delta was getting smaller only, but then if you have the same number of items coming in as being processed, it doesn't look as though anything is being done. By being able to check whether the commit has increased as well, we can at least check if forward progress is being made, even if the gap isn't getting smaller. I'm not sure if that's something we want to consider "healthy" though?

derekcollison: If the delta remains same or increases that is not healthy imo. We are not catching up and should return false. So I think healthy is we are caught up, applied == committed, or we know delta is decreasing.

server/raft.go
```diff
@@ -1160,13 +1161,31 @@ func (n *raft) isCatchingUp() bool {
...
+	if n.commit == 0 || startDelta > 0 {
+		var progress bool
+		for i := 0; i < 10; i++ { // 5ms, in 0.5ms increments
+			n.RUnlock()
+			time.Sleep(time.Millisecond / 2)
+			n.RLock()
+			if curDelta := n.commit - n.applied; curDelta < startDelta {
+				// The gap is getting smaller, so we're making forward progress.
+				progress = true
+				break
```
derekcollison, line 1180: Could just return true here, and if we fail out just warn after loop. No need for progress bool now.

neilalexander: Is it OK to skip the later checks by doing that, i.e. that either we're the leader and/or have heard from the leader recently?

neilalexander: Right, so the way the loop is set up at the moment is if there's no `progress` then we return false, otherwise we just fall through to the leader checks.

derekcollison: We should do leader checks first and only check deltas if we are a follower and we know we have been in recent contact with the leader.

My bad there, did not catch that.

neilalexander: OK, I've refactored it to do leader/follower checks first.

server/raft.go
```diff
@@ -1181,14 +1178,36 @@ func (n *raft) isCurrent() bool {
...
+			n.RLock()
+			if n.commit-n.applied < startDelta {
+				// The gap is getting smaller, so we're making forward progress.
+				return true
+			}
+		}
+	}
+
+	// At this point if we are current, we can return saying so.
+	if n.commit-n.applied == 0 {
```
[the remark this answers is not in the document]
neilalexander: Done!

derekcollison:
Make sure tests are green. Then ping me to take a look. Thanks.

derekcollison (approved):
LGTM!

PR #3796, opened 2023-01-20
Update snapshots to numCores and maxProcs after maxrocs.Set()

### LINKED ISSUE #3793: Reported number of gomaxprocs in varz not updated after automaxprocs runs
## Defect

When creating the varz struct, the server picks up the initial gomaxprocs before automaxprocs runs
and it is not updated later.  As a result, the reported value in `/varz` will remain the initial one without
considering the limits:
https://github.com/nats-io/nats-server/blob/a9a9b92f6d98cae92c299744ad6b602d4981c13e/server/monitor.go#L45

derekcollison:

Resolves #3793 

/cc @nats-io/core

### DIFF
--- a/main.go
+++ b/main.go
@@ -134,6 +134,8 @@ func main() {
 		s.Warnf("Failed to set GOMAXPROCS: %v", err)
 	} else {
 		defer undo()
+		// Reset these from the snapshots from init for monitor.go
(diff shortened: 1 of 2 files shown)


wallyqs (approved):
LGTM

PR #3799, opened 2023-01-21
Auth-Callout connect events

derekcollison:
Make sure connection events during auth callouts are correct.

Fixed one extraneous account update for $G. We sent for the addition before switching but suppressed the change back to 0. We now suppress all for $G as was designed.

/cc @nats-io/core


ripienaar (approved):
LGTM

PR #3801, opened 2023-01-22
[FIXED] Filtered state for all subjects when the first seqs are deleted.

derekcollison:
Discovered doing the optimizations work for interior deletes branch for dev.
Also brought over a simple optimization for large interior deletes if we decide to do a 2.9.12 release.

/cc @nats-io/core

### DIFF
--- a/server/filestore_test.go
+++ b/server/filestore_test.go
@@ -4994,3 +4994,49 @@ func TestFileStoreMsgBlkFailOnKernelFaultLostDataReporting(t *testing.T) {
 	require_True(t, state.Lost != nil)
 	require_True(t, len(state.Lost.Msgs) == 93)
 }
+
+func TestFileStoreAllFilteredStateWithDeleted(t *testing.T) {
+	storeDir := t.TempDir()
+	fs, err := newFileStore(
+		FileStoreConfig{StoreDir: storeDir, BlockSize: 1024},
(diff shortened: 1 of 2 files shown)


server/filestore_test.go
```diff
@@ -4994,3 +4994,49 @@ func TestFileStoreMsgBlkFailOnKernelFaultLostDataReporting(t *testing.T) {
...
+			t.Fatalf("Expected %d to be first, got %d", first, fss.First)
+		}
+		if fss.Last != uint64(last) {
+			t.Fatalf("Expected %d to be last, got %d", last, fss.Last)
+		}
+	}
+
+	checkFilteredState(1, 100, 1, 100)
+	remove(2)
+	checkFilteredState(2, 98, 3, 100)
```
bruth, line 5036: Just confirming the correct way to read this... we attempt to start at sequence 2, however the actual first sequence is 3 (since 2 was deleted) and there are 98 messages (including 3) available to be read in the block. Where as if we started at 1, there would be 99 since the first message was not deleted.

bruth (approved):
Assuming my comment is accurate.. LGTM

server/filestore_test.go
derekcollison: Correct..

PR #3805, opened 2023-01-24
Move test to "norace_test.go"

kozlovic:
The test TestJetStreamClusterConsumerListPaging was in the jetstream_cluster_3_test.go and because of `-race` flag would take more than 440 seconds (7+ minutes) as seen here:

https://app.travis-ci.com/github/nats-io/nats-server/jobs/593984385#L335

Without the `-race` flag, this test takes ~17 seconds.


derekcollison (approved):
LGTM

PR #3812, opened 2023-01-24
Lint warnings on `fmt.Print`, `fmt.Printf`, `fmt.Println`

neilalexander:
This will create linter warnings on `fmt.Print`, `fmt.Printf` and `fmt.Println`, except for in unit tests and a couple sanctioned places.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -22,6 +22,7 @@ linters:
   disable-all: true
   enable:
     # - errcheck
+    - forbidigo
     - gofmt
     # - goimports
     - gosimple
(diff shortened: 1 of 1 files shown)


wallyqs (approved):

derekcollison (approved):
LGTM

My check though wants to capture any added fmt.Printf and warn, even in tests.

neilalexander:
We can include tests, but there are a couple instances of `fmt.Printf` being used in tests already. I can `t.Logf` or something those instead?

derekcollison:
Yea I don't check if they are there period, only if the branch is adding them..

PR #3813, opened 2023-01-25
Allow to Reload Nats Config through monitor http

ramonberrutti-f3:

Allow to reload the config with a http call to the monitor server.
The route is: /reloadz following the same pattern of the other routes.

TODO:
- [ ] Only allow post/put method. 
- [ ] Update the reloader service to support http call: https://github.com/nats-io/nack
- [ ] ~~Allow to reload the config with an internal API call... $SYS.REQ.SERVER.RELOADZ~~

/cc @nats-io/core

### DIFF
--- a/server/monitor.go
+++ b/server/monitor.go
@@ -14,6 +14,7 @@
 package server
 
 import (
+	"bytes"
 	"crypto/sha256"
 	"crypto/tls"
 	"crypto/x509"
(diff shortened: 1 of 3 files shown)


derekcollison:
Thank you for the contribution, however the monitoring endpoints are read only by design. We do have an opportunity to allow this via the system account and the NATS cli. 

Also if in the K8S environment, the helm chart IIRC handles this automatically when it detects the config has changed.

ripienaar:
Big +1 for a system account API to do this. I would also love a API to retrieve the running config.

PR #3814, opened 2023-01-25
[ADDED] Stream subject mapping transforms

jnmoyne:
- [X] Tests added
- [X] Branch rebased on top of current dev

Adds support for subject mapping transforms in streams

Introduces the following changes:

- Extracts subject transform code and tests from accounts.go to it's own separate set of files.
- Ability to add a subject mapping transform destination to sourced streams, allowing to map message subjects as they are received from the stream being sourced
- Ability to source more than once from the same stream, each time using a specific subject filter and destination transform
- Ability to add a stream input subject mapping transform to a stream which will not act as a filter but will transform the subject of messages matching the transform's source according to the transform's destination. This mapping happens regardless of how the messages are received into the stream (mirror, or subject(s) and/or source(s))

Stream transform-2023-01-17-1401-3

This enables many interesting use cases, such as KV bucket sourcing, or inserting a partition number token in the subject in order to then create a consumer per partition, directly at the stream definition level rather than having to resort to using the Core NATS subject mapping functionality (which requires administrative access to the server config file or account key to be configured).

/cc @nats-io/core

### DIFF
--- a/server/jetstream_test.go
+++ b/server/jetstream_test.go
@@ -11343,20 +11343,10 @@ func TestJetStreamMirrorAndSourcesFilteredConsumers(t *testing.T) {
 			t.Fatalf("Unexpected error: %v", err)
 		}
 	}
-	expectFail := func(sn, fs string) {
-		t.Helper()
-		_, err = js.AddConsumer(sn, &nats.ConsumerConfig{DeliverSubject: dsubj, FilterSubject: fs})
-		if err == nil {
-			t.Fatalf("Expected error but got none")
-		}
-	}
 
 	createConsumer("M", "foo")
 	createConsumer("M", "bar")
 	createConsumer("M", "baz.foo")
-	expectFail("M", "baz")
-	expectFail("M", "baz.1.2")
-	expectFail("M", "apple")
 
 	// Now do some sources.
 	if _, err := js.AddStream(&nats.StreamConfig{Name: "O1", Subjects: []string{"foo.*"}}); err != nil {
@@ -11377,9 +11367,6 @@ func TestJetStreamMirrorAndSourcesFilteredConsumers(t *testing.T) {
 
 	createConsumer("S", "foo.1")
 	createConsumer("S", "bar.1")
-	expectFail("S", "baz")
-	expectFail("S", "baz.1")
-	expectFail("S", "apple")
 
 	// Chaining
 	// Create Mirror now.
@@ -11394,9 +11381,6 @@ func TestJetStreamMirrorAndSourcesFilteredConsumers(t *testing.T) {
 	createConsumer("M2", "foo")
 	createConsumer("M2", "bar")
 	createConsumer("M2", "baz.foo")
-	expectFail("M2", "baz")
-	expectFail("M2", "baz.1.2")
-	expectFail("M2", "apple")
 }
 
 func TestJetStreamMirrorBasics(t *testing.T) {
@@ -11623,7 +11607,7 @@ func TestJetStreamSourceBasics(t *testing.T) {
 		Name:    "MS",
 		Storage: FileStorage,
 		Sources: []*StreamSource{
-			{Name: "foo"},
+			{Name: "foo", SubjectTransform: "foo2.>"},
 			{Name: "bar"},
 			{Name: "baz"},
 		},
@@ -11646,6 +11630,14 @@ func TestJetStreamSourceBasics(t *testing.T) {
 		}
 		return nil
 	})
+
+	m, err := js.GetMsg("MS", 1)
+	if err != nil {
+		t.Fatalf("Unexpected error: %v", err)
+	}
+	if m.Subject != "foo2.foo" {
+		t.Fatalf("Expected message subject foo2.foo, got %s", m.Subject)
+	}
 
 	// Test Source Updates
 	ncfg := &nats.StreamConfig{
@@ -11688,7 +11680,7 @@ func TestJetStreamSourceBasics(t *testing.T) {
 		return nil
 	})
 	// Double check first starting.
-	m, err := js.GetMsg("FMS", 1)
+	m, err = js.GetMsg("FMS", 1)
 	if err != nil {
 		t.Fatalf("Unexpected error: %v", err)
 	}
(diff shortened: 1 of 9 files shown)


server/jetstream_test.go
```diff
@@ -11647,6 +11631,14 @@ func TestJetStreamSourceBasics(t *testing.T) {
 		return nil
 	})
 
+	m, err := js.GetMsg("MS", 1)
+	if err != nil {
```
derekcollison, line 11635: requie_NoError(t, err)

server/jetstream_test.go
```diff
@@ -11729,6 +11721,49 @@ func TestJetStreamSourceBasics(t *testing.T) {
...
+
+	createStream(&StreamConfig{Name: "T1", Subjects: []string{"foo"}, InputSubjectTransform: &InputSubjectTransform{Source: ">", Destination: "transformed.>"}, Storage: MemoryStorage})
+
+	// publish a message
+	if _, err := js.Publish("foo", []byte("OK")); err != nil {
+		t.Fatalf("Unexpected publish error: %v", err)
+	}
+
+	m, err := js.GetMsg("T1", 1)
+	if err != nil {
```
derekcollison, line 11759: require_NoError()

server/stream.go
```diff
@@ -58,6 +58,9 @@ type StreamConfig struct {
 	Mirror       *StreamSource   `json:"mirror,omitempty"`
 	Sources      []*StreamSource `json:"sources,omitempty"`
 
+	// Allow applying a subject transform to incoming messages before doing anything else
+	InputSubjectTransform *InputSubjectTransform `json:"input_subject_transform,omitempty"`
```
derekcollison, line 62: Shorter json name possible?

server/stream.go
```diff
@@ -1707,138 +1718,18 @@ func (mset *stream) sourcesInfo() (sis []*StreamSourceInfo) {
...
-}
-
 // Lock should be held
 func (mset *stream) sourceInfo(si *sourceInfo) *StreamSourceInfo {
 	if si == nil {
 		return nil
 	}
 
-	ssi := &StreamSourceInfo{Name: si.name, Lag: si.lag, Error: si.err}
+	var ssi *StreamSourceInfo
```
derekcollison, line 1727: Maybe keep same original assignment and just add the field if si.tr != nil?

server/subject_transform.go
```diff
@@ -0,0 +1,518 @@
...
+	if err != nil {
+		return BadTransform, []int{}, -1, _EMPTY_, &mappingDestinationErr{token, ErrorMappingDestinationFunctionInvalidArgument}
+	}
+
+	return transformType, []int{i}, int32(mappingFunctionIntArg), _EMPTY_, nil
+}
+
+// Helper to ingest and index the subjectTransform destination token (e.g. $x or {{}}) in the token
+// returns a transformation type, and three function arguments: an array of source subject token indexes, and a single number (e.g. number of partitions, or a slice size), and a string (e.g.a split delimiter)
+
```
derekcollison, line 176: Remove empty line, make line above 2.

derekcollison (commented):
Just a few minor nits.

server/stream.go
jnmoyne: I thought of just input_transform but then thought maybe we want to have transform on other things than subjects later. And also thought since it's only in stream config / stream info it's not critical to optimize the size and as humans will be editing those JSONs a lot of the time I erred on the side of maybe a bit too verbose figuring it couldn't hurt 🤷‍♂️ but yeah it's like 'input' it's the best I could come up with but not especially attached to it...

ripienaar: Given it’s “subjects” when setting the interest subjects for the stream naming this input sometning is confusing so maybe subject_transform or even subjects_transform to be very clear

server/jetstream_test.go
jnmoyne: Done, since this pattern was present in many places in the file I made it consistent and replaced with require_NoError in all these other places in the file as well

server/jetstream_test.go
jnmoyne: Done (see above)

server/stream.go

server/subject_transform.go
jnmoyne: Done

server/stream.go
jnmoyne: Yeah I put input in there to try and denote that it's not just messages from the 'subjects' that get transformed but also those from sources or the mirror, but really for lack of a better term

jnmoyne: Trying to convey it's a 'stream pre-processing' transform that happens after the subjects/mirror/sources step (which itself can already have another transform, at least for sources)

ripienaar: We cant possibly convey all this by choice of a single word :) So seems we should at least just pick a term and stick with it, even better if its one we already use.  It's subject(s).

jnmoyne: Maybe trying to cram too much meaning in a field name (picking good names can be hard 😅) and "subject_transform" is enough... @derekcollison let me know what name you like best and I'll adjust everywhere (companion branches) for it.

jnmoyne: @ripienaar fair enough, I will go with "subject_transform" then (unless someone else comes with a better one by tomorrow morning)

derekcollison:
LMK when you want me to take another look for review.

jnmoyne:
Still need to do the changes for input_subject_transform naming but I’m in meetings most of the time until sometime in the afternoon today, I’ll let you know. I believe that name change should be the only thing left.

derekcollison (approved):
LGTM

bruth:
@derekcollison @jnmoyne Playing around, I was able to construct a cycle with the two stream setup with this PR and #3825 applied by setting the `subject_transform_dest` to the local subject subject, e.g.

```
{
  "name": "n1",
  "subjects": ["n1.>"],
  "sources": [
    {
      "name": "n2",
      "filter_subject": "n2.>",
      "subject_transform_dest": "n1.>"
    }
  ]
}
```

And the equivalent for `n2` (swapping `n1` and `n2`). This is to be expected since the transform occurs before sourcing occurs, but just wanted to call it out.

jnmoyne:
Will take a look

derekcollison:
I added in ability to have two sources reference each other as long as their subjects did not overlap. 

Just need to add in the transform logic in dev branch.

bruth:
I had created a branch that rebased this one on top of main, incorporating both sets of changes called `bjr-suse`.

jnmoyne:
Can't reproduce using the current top of dev, which also has #3825 merged in

PR #3817, opened 2023-01-26
[FIXED] Force consumer replicas to match for interest policy streams

derekcollison:
Consumers must match replica of parent stream if interest based policy retention.

For MQTT, this was ill-advised by me, not understanding that the messages stream for MQTT was interest policy based. Interest policy based streams require consumers to match the replica count.

/cc @nats-io/core

### DIFF
--- a/server/consumer.go
+++ b/server/consumer.go
@@ -382,6 +382,17 @@ func checkConsumerCfg(
 	if config.Replicas < 0 {
 		return NewJSReplicasCountCannotBeNegativeError()
 	}
+	// If the stream is interest or workqueue retention make sure the replicas
+	// match that of the stream. This is REQUIRED for now.
+	if cfg.Retention == InterestPolicy || cfg.Retention == WorkQueuePolicy {
+		// Only error here if not recovering.
+		// We handle recovering in a different spot to allow consumer to come up
+		// if previous version allowed it to be created. We do not want it to not come up.
+		if !isRecovering && config.Replicas != 0 && config.Replicas != cfg.Replicas {
+			fmt.Printf("config is %+v\n", config)
+			return NewJSConsumerReplicasShouldMatchStreamError()
+		}
+	}
 
 	// Check if we have a BackOff defined that MaxDeliver is within range etc.
 	if lbo := len(config.BackOff); lbo > 0 && config.MaxDeliver <= lbo {
--- a/server/consumer.go
+++ b/server/consumer.go
@@ -389,7 +389,6 @@ func checkConsumerCfg(
 		// We handle recovering in a different spot to allow consumer to come up
 		// if previous version allowed it to be created. We do not want it to not come up.
 		if !isRecovering && config.Replicas != 0 && config.Replicas != cfg.Replicas {
-			fmt.Printf("config is %+v\n", config)
 			return NewJSConsumerReplicasShouldMatchStreamError()
 		}
 	}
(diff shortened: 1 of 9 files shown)


server/opts.go
```diff
@@ -439,6 +439,9 @@ type MQTTOpts struct {
 	// replicas count (lower than StreamReplicas if specified, but also lower
 	// than the automatic value determined by cluster size).
 	// Note that existing consumers are not modified.
+	//
+	// UPDATE: This is no longer used while messages stream has interest policy retention
```
wallyqs, line 443: Should we mark as `Deprecated`?
```suggestion
	// Deprecated: This is no longer used while messages stream has interest policy retention
```

wallyqs (approved):
LGTM

server/consumer.go
```diff
@@ -382,6 +382,17 @@ func checkConsumerCfg(
...
 		return NewJSReplicasCountCannotBeNegativeError()
 	}
+	// If the stream is interest or workqueue retention make sure the replicas
+	// match that of the stream. This is REQUIRED for now.
+	if cfg.Retention == InterestPolicy || cfg.Retention == WorkQueuePolicy {
+		// Only error here if not recovering.
+		// We handle recovering in a different spot to allow consumer to come up
+		// if previous version allowed it to be created. We do not want it to not come up.
+		if !isRecovering && config.Replicas != 0 && config.Replicas != cfg.Replicas {
+			fmt.Printf("config is %+v\n", config)
```
wallyqs, line 392: leftover Printf
```suggestion
```

server/opts.go
derekcollison: I decided against in just in case I change over the stream to not be interest policy based in 2.10.

server/consumer.go
derekcollison: uug and I even have a check utility for that! Thanks!

derekcollison: Fixed

wallyqs (approved):

server/consumer.go
```diff
@@ -382,6 +382,16 @@ func checkConsumerCfg(
 	if config.Replicas < 0 {
 		return NewJSReplicasCountCannotBeNegativeError()
 	}
+	// If the stream is interest or workqueue retention make sure the replicas
+	// match that of the stream. This is REQUIRED for now.
+	if cfg.Retention == InterestPolicy || cfg.Retention == WorkQueuePolicy {
```
sourabhaggrawal, line 387: Hi @derekcollison PR description mentioned changes only for InterestPolicy, any specific reason to consider WorkQueue here ? 
Does it mean going forward with new version we wont be able to create consumer for InterestBased as well as WorkQueue based if replica is < stream replica ?

derekcollison: For now that is correct. The way the system is architected currently this is a requirement.

However, we are looking at changes for 2.11 to the complete way consumers work, so stay tuned.

sourabhaggrawal: Hi @derekcollison we have a prod use case where we want to have million of consumer in total out of few hundreds workqueue stream. Our stream is R3 file and consumer is R1 memory. We cant have R3 consumer  as with millions of consumer there will be way too much of network traffic.

Also could you please share the timeline for 2.11 ?

derekcollison: You can not have an R1 consumer with and R2 stream. 

We have plans to reach those goals and are working with some customers now. 

Feel free to send me an email directly and we can setup a meeting.

derek at synadia dot com

Kaarel:
Hi, I think we are impacted by this change. See issue here #3858

PR #3818, opened 2023-01-26
[FIXED] Properly process updates on a stream on restart.

### LINKED ISSUE #3636: Consumer Replica R3 -> R2 while Server is offline results in hung health probe
Issue reported by a customer

## Version

```
Version:  2.10.0-beta.1
Git:      [deadc05c]
Go build: go1.19.2
```

## Summary
3-pod cluster.  `Pod 0`, `Pod 1`, and `Pod 2`.  `Pod 1` was taken offline due to issues, and `Stream S` with `Consumer C` were changed from R3 to R2 to remove `Pod 1` as a peer

## Issue 1

`Pod 1` came back online.  It was stuck starting up because the `/healthz` endpoint reported that it was still trying to catch-up `Consumer C` even though that was an R2 consumer that was no longer assigned to `Pod 1`

## Fix Attempt 1

`Stream S` was changed from R2 back to R3

## Issue 2

Even though `Stream S` changed from R2 to R3 and `Consumer C` had num_replicas=0 (I think this means inherit replica count from stream)?, `Consumer C` still reflected that it was R2.  `Pod 1` was still stuck trying to catch up.

## Fix Attempt 2

`Consumer C` was deleted, and `Pod 1` finally started up

derekcollison:
During restart if the stream existed in the server's state but was also in a meta-snapshot delivered by the leader we would not process the update properly.

Resolves #3636 

/cc @nats-io/core

### DIFF
--- a/server/jetstream_cluster.go
+++ b/server/jetstream_cluster.go
@@ -1264,6 +1264,16 @@ func (js *jetStream) applyMetaSnapshot(buf []byte) error {
 			js.processConsumerAssignment(ca)
 		}
 	}
+
+	// Perform updates on those in saChk. These were existing so make
+	// sure to process any changes.
+	for _, sa := range saChk {
+		if isRecovering {
+			js.setStreamAssignmentRecovering(sa)
+		}
+		js.processUpdateStreamAssignment(sa)
+	}
+
 	// Now do the deltas for existing stream's consumers.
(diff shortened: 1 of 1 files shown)


wallyqs (approved):
LGTM

neilalexander (approved):

PR #3824, opened 2023-01-27
Check if connection name was already set when storing it

wallyqs:

/cc @nats-io/core

### DIFF
--- a/server/client.go
+++ b/server/client.go
@@ -1883,7 +1883,7 @@ func (c *client) processConnect(arg []byte) error {
 			}
 		}
 		if ncs != _EMPTY_ {
-			c.ncs.Store(fmt.Sprintf("%s - %q", c, ncs))
+			c.ncs.CompareAndSwap(nil, fmt.Sprintf("%s - %q", c, ncs))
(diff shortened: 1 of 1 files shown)


server/client.go
```diff
@@ -1883,7 +1883,7 @@ func (c *client) processConnect(arg []byte) error {
 			}
 		}
 		if ncs != _EMPTY_ {
-			c.ncs.Store(fmt.Sprintf("%s - %q", c, ncs))
+			c.ncs.CompareAndSwap(nil, fmt.Sprintf("%s - %q", c, ncs))
```
derekcollison, line 1886: This Go 1.19 only?

wallyqs: Looks like built-in since Go 1.17: https://pkg.go.dev/sync/atomic#Value.CompareAndSwap

derekcollison: Does this change pass the fuzz test?

wallyqs: yes this would fix it

derekcollison (approved):
LGTM!

PR #3825, opened 2023-01-27
[FIXED] Sourced streams that do not overlap subjects reported as a cycle.

derekcollison:
Allow cycles between source streams if subjects do not overlap.

This is a common pattern we can use for 2 node setups.

/cc @nats-io/core


wallyqs (approved):
LGTM

PR #3828, opened 2023-01-28
A few small optimizations

derekcollison:
1. Only snapshot with minSnap time window like consumers and meta. Make it consistent for all to 5s.
2. Only snapshot at the end of processing all entries pending vs inside the loop.
3. Use fast state when calculating sync request, do not need deleted details there.
4. Only hold on to so many raft pending in memory, will fetch from WAL

/cc @nats-io/core


Jarema (approved):
LGTM

PR #3831, opened 2023-01-30
Minor fixes and optimizations for snapshots.

derekcollison:
We were snapshotting more then needed, so double check that we should be doing this at the stream and consumer level. At the raft level, we should have always been compacting the WAL to last+1, so made that consistent. Also fixed bug that would not skip last if more items behind the snapshot.


wallyqs (approved):
LGTM

PR #3833, opened 2023-01-31
Add `logtime_utc` option

### LINKED ISSUE #3826: Configure log timestamp as UTC
## Feature Request

Log timestamp can be enabled or disabled (default enabled), but timestamp is always in the host's currently configured timezone (e.g. PST).  The timezone is not recorded in the timestamp format.

Users would benefit from an option to log timestamp as UTC regardless of host's currently configured timezone.

#### Use Case:

Logs from multiple servers are scraped into a central log aggregator service where timezone normalization is needed or required.

#### Proposed Change:

Add a server configuration option (not default to preserve back compatibility) to always log timestamps (if enabled) as normalized UTC.

Users who aggregate server logs from multiple hosts that may not always be uniformly configured as UTC timezone at host level.

#### Alternative Approaches

Output timestamp in format that preserves timezone information which could allow log aggregators to normalize downstream as required.

neilalexander:
Fixes #3826 by adding the `logtime_utc` configuration option for both the stdout and the file logger.

### DIFF
--- a/logger/log.go
+++ b/logger/log.go
@@ -38,12 +38,36 @@ type Logger struct {
 	fl         *fileLogger
 }
 
-// NewStdLogger creates a logger with output directed to Stderr
-func NewStdLogger(time, debug, trace, colors, pid bool) *Logger {
+type LogOption interface {
+	isLoggerOption()
+}
+
+// LogUTC controls whether timestamps in the log output should be UTC or local time.
+type LogUTC bool
+
+func (l LogUTC) isLoggerOption() {}
+
+func logFlags(time bool, opts ...LogOption) int {
 	flags := 0
 	if time {
 		flags = log.LstdFlags | log.Lmicroseconds
 	}
+
+	for _, opt := range opts {
+		switch v := opt.(type) {
+		case LogUTC:
+			if time && bool(v) {
+				flags |= log.LUTC
+			}
+		}
+	}
+
+	return flags
+}
+
+// NewStdLogger creates a logger with output directed to Stderr
+func NewStdLogger(time, debug, trace, colors, pid bool, opts ...LogOption) *Logger {
+	flags := logFlags(time, opts...)
 
 	pre := ""
 	if pid {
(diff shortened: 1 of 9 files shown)


logger/log.go
```diff
@@ -66,10 +69,13 @@ func NewStdLogger(time, debug, trace, colors, pid bool) *Logger {
 }
 
 // NewFileLogger creates a logger with output directed to a file
-func NewFileLogger(filename string, time, debug, trace, pid bool) *Logger {
+func NewFileLogger(filename string, time, utc, debug, trace, pid bool) *Logger {
```
derekcollison, line 72: Public function, so probably can not change.

Could add optional to it or separate new function?

gedw99:
Very happy to see this..

logger/log.go
neilalexander: Good idea, I've modified the functions to take a variadic set of options at the end, so the original function signature is still valid.

logger/log.go
```diff
@@ -38,13 +38,37 @@ type Logger struct {
 	fl         *fileLogger
 }
 
-// NewStdLogger creates a logger with output directed to Stderr
-func NewStdLogger(time, debug, trace, colors, pid bool) *Logger {
+type LogOption interface {
+	isLoggerOption()
```
derekcollison, line 42: Should this be upper case so it is also public?

derekcollison: Another approach could be add new constructor?

NewStdUTCLogger() and NewFileUTCLogger() 

Just a thought, not suggesting either way is better or not.

neilalexander: The interface is exported but the function to match isn't, that way it prevents other packages from defining new `LogOptions` that won't make sense to us (as they can't implement the `isLoggerOption` to do so).

Also I did think about adding new functions for UTC specifically, but figured an options-based approach would be more future-proof in case there are more options we want to add in the future.

derekcollison: ok thanks.. Appreciate the feedback.

derekcollison (approved):
LGTM

server/opts.go
```diff
@@ -4805,6 +4809,7 @@ func ConfigureOptions(fs *flag.FlagSet, args []string, printVersion, printHelp,
 	fs.BoolVar(&dbgAndTrcAndVerboseTrc, "DVV", false, "Enable Debug and Verbose Trace logging. (Traces system account as well)")
 	fs.BoolVar(&opts.Logtime, "T", true, "Timestamp log entries.")
 	fs.BoolVar(&opts.Logtime, "logtime", true, "Timestamp log entries.")
+	fs.BoolVar(&opts.LogtimeUTC, "logtime_utc", true, "Timestamps in UTC instead of local timezone.")
```
bruth, line 4812: @neilalexander To not change the existing behavior, the default should be `false` here, correct?

PR #3837, opened 2023-02-02
[FIXED] Do not let !NeedSnapshot() avoid snapshots and compaction.

derekcollison:
Had a logic bug that would incorrectly suppress snapshots and compaction of the raft logs.

/cc @nats-io/core


wallyqs (approved):
LGTM

PR #3844, opened 2023-02-04
[FIXED] Snapshots Issue

derekcollison:
Due to a bug, in rare circumstances we could write an empty snapshot for applied == 0. This would cause a spinning behavior at the raft layer.

To fix this issue, given that these could be out in deployments potentially.

    1. Allow Truncate() to also properly do a reset of the store when terms were the only mismatch.
    2. During testing fixed memstore truncate and also made sure per subject info was also cleaned up.
    3. Added fix to detect a bad snapshot on initialization and remove.
    4. Do not allow snapshots for applied == 0.

Also small fixes for new staticcheck.

/cc @nats-io/core

### DIFF
--- a/server/raft.go
+++ b/server/raft.go
@@ -939,7 +939,7 @@ func (n *raft) InstallSnapshot(data []byte) error {
 	var state StreamState
 	n.wal.FastState(&state)
 
-	if n.snapfile != _EMPTY_ && state.FirstSeq >= n.applied {
+	if state.FirstSeq >= n.applied {
 		n.Unlock()
 		return nil
 	}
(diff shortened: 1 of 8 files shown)


wallyqs (approved):
LGTM

server/raft.go
```diff
@@ -1123,6 +1124,15 @@ func (n *raft) loadLastSnapshot() (*snapshot, error) {
 		data:      buf[20+lps : hoff],
 	}
 
+	// We had a bug in 2.9.12 that would allow snapshots on last index of 0.
+	// Detect that here and return err.
+	if snap.lastIndex == 0 {
+		n.warn("Snapshot with last index 0 is invalid, cleaning up")
```
ripienaar, line 1130: On starting my servers I now see pages of this, is it essentially a one off? and I wonder if it really should be a warning since no-one can do anything about it really

wallyqs: sounds good to me to change to debug

derekcollison:
Only happens once on startup for consumer that are HA and never had a single state transition, which is contrived in R.I.'s setup. We discussed this am on Slack, will leave as is.

PR #3845, opened 2023-02-06
[FIXED] Updating consumer filter subject possible panic and deadlock.

derekcollison:
For updating a consumer filter subject make sure locking order is correct and that our sublist is present.

/cc @nats-io/core

### DIFF
--- a/server/stream.go
+++ b/server/stream.go
@@ -4688,25 +4688,41 @@ func (mset *stream) removeConsumerAsLeader(o *consumer) {
 }
 
 // swapSigSubs will update signal Subs for a new subject filter.
-// Lock should be held.
+// consumer lock should not be held.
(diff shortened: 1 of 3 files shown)


server/stream.go
```diff
@@ -4688,25 +4688,41 @@ func (mset *stream) removeConsumerAsLeader(o *consumer) {
 }
 
 // swapSigSubs will update signal Subs for a new subject filter.
-// Lock should be held.
+// conumer lock should not be held.
```
wallyqs, line 4691:
```suggestion
// consumer lock should not be held.
```

derekcollison: fixed.

Jarema (approved):
LGTM (assuming green CI)

derekcollison:
Will wait on green tests.

PR #3853, opened 2023-02-07
[WIP] enable signing keys

aricart:
 - [ ] Tests added

Resolves #

 -
 -
 -

/cc @nats-io/core

### DIFF
--- a/server/auth_callout.go
+++ b/server/auth_callout.go
@@ -121,7 +121,7 @@ func (s *Server) processClientOrLeafCallout(c *client, opts *Options) (authorize
 			}
 			// this will validate the scope - so if it doesn't have the right issuer it gets rejected
 			if scope, ok := userAccount.hasIssuer(u.Issuer); !ok {
-				respCh <- fmt.Sprint("User JWT issuer is not known")
+				respCh <- "User JWT issuer is not known"
 				return
 			} else if scope != nil {
 				// this possibly has to be different because it could just be a plain issued by a non-scoped signing key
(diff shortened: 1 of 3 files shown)


server/auth_callout.go
```diff
@@ -118,6 +118,9 @@ func (s *Server) processClientOrLeafCallout(c *client, opts *Options) (authorize
 		}
 		// By default issuer needs to match server config or the requesting account in operator mode.
 		if arc.Issuer != issuer {
+			if arc.IssuerAccount != "" {
```
derekcollison, line 121: _EMPTY_ vs "" in the server.

server/auth_callout.go
```diff
@@ -118,6 +118,9 @@ func (s *Server) processClientOrLeafCallout(c *client, opts *Options) (authorize
...
+				arc.Issuer = arc.IssuerAccount
```
sethjback, line 122: `arc.Issuer` is used in 177 below to see if it is a signing key. This change would seem however to always set the `arc.issuer` to the account root key (`IssuerAccount`).

Since the server is relying on the `aud` of the user claims to indicate the target account, would it be simpler to change 127 to:
```golang
} else if !acc.isAllowedAcount(arc.User.Audience) {
```

aricart:
would be good to have a chat about this - note that I think this is an intermediate step, I had to figure out how to get the scoped limits to be ingested.

server/auth_callout.go
aricart: fixed

PR #3855, opened 2023-02-07
YAML engineering: quote go version string

philpennock:
We're currently using Go 1.19; we'll switch to 1.20 when the NATS Maintainers
make the call to switch.  Prepare by making sure that 1.20 won't turn into 1.2
instead, by quoting the string to be updated and adding a warning.

/cc @nats-io/core

### DIFF
--- a/.github/workflows/go-test.yaml
+++ b/.github/workflows/go-test.yaml
@@ -5,7 +5,9 @@ jobs:
   test:
     strategy:
       matrix:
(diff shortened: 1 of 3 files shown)


wallyqs (approved):
LGTM

philpennock:
just waiting for CI tests to pass ... flaky test on Travis.  I'm "confident", but want to see it green first.

PR #3859, opened 2023-02-10
Small fix to nats-server-hardened.service

wallyqs:
Importing as is would have reported:
```
/etc/systemd/system/nats.service:15: Failed to parse resource value, ignoring: 800000 # JetStream requires 2 FDs open per stream.
```

/cc @nats-io/core


derekcollison (approved):
LGTM - Thanks Wally!

PR #3860, opened 2023-02-10
Make sure we adjust per subject info when doing a Compact().

derekcollison:

/cc @nats-io/core


wallyqs (approved):
LGTM

neilalexander (approved):

PR #3862, opened 2023-02-10
Fix stack overflow panic in reverse entry check when inbox ends with wildcard

neilalexander:
In the case that the reverse entries also contain wildcards then this leads to infinite recursion. By breaking out the function to accept a parameter to tell us whether or not we have recursed, we can detect this condition and give up.

/cc @nats-io/core


aricart (approved):
LGTM

neilalexander:
@derekcollison Would you like this on `dev` or `main`? It might go nicely in a patch release.

derekcollison (approved):
LGTM - I think main branch here.

derekcollison (approved):
LGTM

PR #3864, opened 2023-02-10
Simplify authresponse

aricart:
This PR reworks the auth callout to simply return an user JWT.

 - [X] Tests added

The original auth callout response, is simpler:
- the response is just JSON `{error: string, user_token: string}`. If encrypted the entire payload is encrypted.
- The server id which was set as an audience, is now set as a tag, this allows the server to verify the server (albeit by turning the server ID lower case)
- For operator mode, the callout should include it's public key, and sign the payload, and return both of these values in a header. This way the server can verify that the payload was sent from the expected service. On encrypted payloads this is not required, since the payload is encrypted and the server can verify. 

The callout service can now use account signing keys and scoped signing keys. The server will validate the signing keys used to match the target account, and in the case of the scoped signing key, will assign limits as recorded in the destination account. The one requirement is that the limits be default (zero) struct (otherwise the resulting JWT will be invalid once validation runs).

The audience field is now reserved for account placement on non-operator mode. In operator mode, the user JWT has all the proper attributions to determine the target account.

/cc @nats-io/core


server/auth_callout.go
```diff
@@ -159,30 +174,60 @@ func (s *Server) processClientOrLeafCallout(c *client, opts *Options) (authorize
 		// Apply to this client.
 		targetAcc := acc
 		// Check if we are being asked to switch accounts.
-		if aname := juc.Audience; aname != _EMPTY_ {
+		// FIXME: this is a directive, shouldn't we be checking here if the
+		//   audience account matches the jwtIssuer, because that is the account...
+		if aname := arc.Audience; aname != _EMPTY_ {
```
caleblloyd, line 179: I think this conditional can go away and the line above can become:

```
targetAcc := s.LookupAccount(jwtIssuer)
```

server/auth_callout.go
```diff
@@ -116,33 +132,32 @@ func (s *Server) processClientOrLeafCallout(c *client, opts *Options) (authorize
 			// Operator mode is who we send the request on unless switching accounts.
 			issuer = acc.Name
 		}
-		// By default issuer needs to match server config or the requesting account in operator mode.
-		if arc.Issuer != issuer {
+
+		// the jwt issuer can be a signing key
+		jwtIssuer := arc.Issuer
+		if arc.IssuerAccount != "" {
```
caleblloyd, line 138: This conditional should be `if isOperatorMode && arc.IssuerAccount != ""`

In non-operator mode there is no validation for signing keys

server/auth_callout.go
aricart: Almost - while on the surface the idea of the authcallout is to create a credential that maps to an account, the code is currently treating audience as a signal to switch accounts.

server/auth_callout.go
aricart: actually - if it is NOT operator mode, and they set an issuer account, it should fail...

server/auth_callout.go
```diff
@@ -18,22 +18,42 @@ import (
...
+func getTagValue(uc *jwt.UserClaims, key string) string {
+	// we are expecting `tag: serverID` or `tag:serverID`
+	// note that tags lower case
+	k := fmt.Sprintf("%s:", strings.ToLower(key))
+	for _, t := range uc.Tags {
+		if strings.HasPrefix(t, k) {
+			prefix := len(key) + 1
+			if len(t) > prefix {
+				return strings.TrimSpace(t[prefix:])
+			}
```
caleblloyd, line 45:
```suggestion
			return strings.TrimSpace(t[len(k):])
```

nit - this length check is redundant, [CutPrefix from go 1.20](https://cs.opensource.google/go/go/+/refs/tags/go1.20:src/strings/strings.go;l=1273-1278) doesn't use one

server/auth_callout.go
```diff
@@ -116,73 +161,141 @@ func (s *Server) processClientOrLeafCallout(c *client, opts *Options) (authorize
...
-				return
-			} else if !acc.isAllowedAcount(arc.Issuer) {
-				respCh <- fmt.Sprintf("Account %q not permitted as valid account option for auth callout for account %q",
-					arc.Issuer, racc.Name)
-				return
-			}
+
+		// the jwt issuer can be a signing key
+		jwtIssuer := arc.Issuer
+		if !isOperatorMode && arc.IssuerAccount != "" {
```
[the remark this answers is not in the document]
aricart: actually glad you like that - because I do as well - was just trying to match how DLC does it - but then there's always the ambiguity of `if false && false` which is not here...

derekcollison:
Let me know when we think this one is ready for review.

PR #3867, opened 2023-02-14
Improvements to Filestore

derekcollison:
Only update on disk per subject information if we know we have an update.

Increase limit for internal blocking IO vs num cores. Will help in slow IO situations and lots of expire cache calls.

Improve expireMsgs minAge calculation for when lots of messages to expire in each callback.

/cc @nats-io/core


neilalexander (approved):
LGTM, I've done some testing for the last couple hours and it looks good so far.

derekcollison:
Will merge, I have a followup coming that is an issue for monitoring probes and certain JetStream setups inside Kubernetes environments.

PR #3876, opened 2023-02-16
[FIXED] Extended Purge accounting was not updating account usage.

derekcollison:
Make sure we update storage accounting on extended version purge for filestore.

/cc @nats-io/core


wallyqs (approved):
LGTM

PR #3877, opened 2023-02-17
[IMPROVED] ConsumerInfo request processing.

derekcollison:
Do not re-calculate NumPending on all consumer info calls.

We noticed that consumer info calls were being called alot in some environments. When the consumer had a filtered subject with a wildcard and the stream had a high cardinality of subjects and was falling behind this could take a substantial amount of time.

/cc @nats-io/core


neilalexander (approved):
LGTM

PR #3880, opened 2023-02-18
[IMPROVED] Parallel consumer 

derekcollison:
Parallel consumer creation could drop responses (create and info) and could also possibly run monitorConsumer twice.

/cc @nats-io/core


wallyqs (approved):
LGTM, no longer run into consumer info timeout with these changes.

PR #3882, opened 2023-02-18
[FIXED] Do not warn if consumer replicas configured to 0

derekcollison:
If a consumer had an older configuration with replicas of 0, we do not need to warn for an interest policy stream since the replicas will be inherited from the parent stream.

/cc @nats-io/core


Jarema (approved):
LGTM

PR #3884, opened 2023-02-19
[FIXED] Named push consumers with inactive thresholds deleted when still active.

derekcollison:
Fixed a bug where a named consumer under an interest policy stream was incorrectly spinning up inactive threshold timers in all replicas not just the leader.

/cc @nats-io/core


Jarema (approved):
LGTM

PR #3886, opened 2023-02-19
Check if connection name was already set when storing it

wallyqs:
Same change from https://github.com/nats-io/nats-server/pull/3824 but to main branch instead.
/cc @nats-io/core


derekcollison (approved):
LGTM!

PR #3893, opened 2023-02-20
[FIXED] Spurious "Error storing entry to WAL" log messages

derekcollison:
Needed to compact wal on snapshot to pindex+1, not pindex to make sure next append entry would not fail.
Also we could spin on a certain condition due to not processing if equal as well as less than.

/cc @nats-io/core


wallyqs (approved):
LGTM

Jarema (approved):
LGTM

neilalexander (approved):
LGTM!

PR #3894, opened 2023-02-21
Refactor `ipQueue` to use generics

neilalexander:
This refactors `ipQueue` to use Go generics. Since there are now considerably less interface boundaries, there should be far less small runtime allocations and less potential for `nil`s to creep out by accident and cause panics during type assertions.

### DIFF
--- a/server/consumer.go
+++ b/server/consumer.go
@@ -307,7 +307,7 @@ type consumer struct {
 	ptail *proposal
 
 	// Ack queue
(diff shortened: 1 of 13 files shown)


derekcollison (approved):
LGTM

ripienaar:
Are there any comparative benchmarks? Might be a non issue but generics do have some perf concerns

gedw99:
Nice work @neilalexander

PR #3898, opened 2023-02-22
Install snapshot when consumer WAL grows, even when no state changes occur

neilalexander:
This prevents the case that the WAL of an R3 pull consumer can grow endlessly as a result of fetches when there's no activity on the stream.

/cc @nats-io/core


server/jetstream_cluster.go
```diff
@@ -4022,15 +4022,26 @@ func (js *jetStream) monitorConsumer(o *consumer, ca *consumerAssignment) {
...
-			if ne, nb := n.Size(); ne < compactNumMin && nb < compactSizeMin {
+			if ne < compactNumMin && nb < compactSizeMin {
 				return
 			}
 		}
 
 		if snap, err := o.store.EncodedState(); err == nil {
-			if hash := highwayhash.Sum(snap, key); !bytes.Equal(hash[:], lastSnap) {
+			hash := highwayhash.Sum(snap, key)
+			switch {
```
derekcollison, line 4035: Would a simple if be better? 

```go
if !bytes.Equal(hash[:], lastSnap) || ne >= compactNumMin || nb >= compactSizeMin {
}
```

derekcollison: Also add in force as a qualifier.

server/jetstream_cluster_3_test.go
```diff
@@ -2779,3 +2779,47 @@ func TestJetStreamClusterInterestPolicyEphemeral(t *testing.T) {
...
+		nats.ConsumerReplicas(3),
+	)
+	require_NoError(t, err)
+
+	for i := 0; i < 100000; i++ {
+		_, _ = sub.Fetch(1, nats.MaxWait(time.Microsecond))
+	}
+
+	s := c.randomNonConsumerLeader(globalAccountName, "TEST", "durable")
+	dc := s.js.accounts[globalAccountName].streams["TEST"].consumers["durable"]
```
derekcollison, line 2813: Needs to be locked.

server/jetstream_cluster_3_test.go
```diff
@@ -2779,3 +2779,47 @@ func TestJetStreamClusterInterestPolicyEphemeral(t *testing.T) {
...
+
```
derekcollison, line 2814: Need to wait 5s and then do more fetches then check. Only do a few fetches, like 10.

derekcollison (changes requested):

derekcollison (approved):
LGTM!

PR #3900, opened 2023-02-22
[FIXED] KV accounting errors under heavy concurrent usage

derekcollison:
When having a max msgs per subject (e.g. KV) under heavy concurrent usage we could skew the accounting for the underlying filestore.

This fixes the incorrect accounting.

/cc @nats-io/core


wallyqs (approved):
LGTM

PR #3901, opened 2023-02-23
[IMPROVED] Consumer create performance in some circumstances.

derekcollison:
In cases where we had a large subject space, a filestore with many msg blocks, and a filtered consumer with a wildcard filtered subject, creating a consumer could take more memory and time then we wanted. This improvement works when the consumer is DeliverAll and we used the upper layer in memory psim structure to scan but only in memory and avoid a file read for each msg block.

/cc @nats-io/core

### DIFF
--- a/server/filestore.go
+++ b/server/filestore.go
@@ -27,6 +27,7 @@ import (
 	"fmt"
 	"hash"
 	"io"
+	"math"
 	"net"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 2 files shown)


server/filestore.go
```diff
@@ -1709,7 +1779,7 @@ func (fs *fileStore) SubjectsState(subject string) map[string]SimpleState {
 	}
 
 	start, stop := fs.blks[0], fs.lmb
-	// We can short circuit if not a wildcard using psim for start and stop.
+	// We can short circuit if not a wildcard using psim1 for start and stop.
```
wallyqs, line 1782: (Here `psim1` referring to `psim` I think)

wallyqs (approved):
LGTM

server/filestore.go
derekcollison: Fixed thx.

PR #3902, opened 2023-02-23
Log asset name for RAFT debug/warn/error messages

neilalexander:
This should make it considerably easier to identify which clustered streams/consumers a given RAFT log line applies to without having to dissect the file store.

/cc @nats-io/core

### DIFF
--- a/server/jetstream_cluster.go
+++ b/server/jetstream_cluster.go
@@ -700,7 +700,7 @@ func (js *jetStream) setupMetaGroup() error {
 	}
 
 	// Start up our meta node.
-	n, err := s.startRaftNode(sysAcc.GetName(), cfg)
+	n, err := s.startRaftNode(sysAcc.GetName(), "_meta_", cfg)
 	if err != nil {
 		s.Warnf("Could not start metadata controller: %v", err)
 		return err
@@ -1784,7 +1784,7 @@ func (rg *raftGroup) setPreferred() {
 }
 
 // createRaftGroup is called to spin up this raft group if needed.
-func (js *jetStream) createRaftGroup(accName string, rg *raftGroup, storage StorageType) error {
+func (js *jetStream) createRaftGroup(accName, assetName string, rg *raftGroup, storage StorageType) error {
 	js.mu.Lock()
 	s, cc := js.srv, js.cluster
 	if cc == nil || cc.meta == nil {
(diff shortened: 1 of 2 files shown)


server/jetstream_cluster.go
```diff
@@ -1696,7 +1696,7 @@ func (js *jetStream) createRaftGroup(accName string, rg *raftGroup, storage Stor
 	}
 	js.mu.Unlock()
 
-	s.Debugf("JetStream cluster creating raft group:%+v", rg)
+	s.Debugf("JetStream cluster creating raft group:%+v for asset %q", rg, assetName)
```
ripienaar, line 1699: would be sweet if in the `StreamConfig` that is used to make the `S-` and `C-` streams we use the new metadata to also record the asset and account the raft asset relates to.

derekcollison: 2.10?

server/jetstream_cluster.go
```diff
@@ -1672,7 +1672,7 @@ func (rg *raftGroup) setPreferred() {
 }
 
 // createRaftGroup is called to spin up this raft group if needed.
-func (js *jetStream) createRaftGroup(accName string, rg *raftGroup, storage StorageType) error {
+func (js *jetStream) createRaftGroup(accName, assetName string, rg *raftGroup, storage StorageType) error {
```
derekcollison, line 1675: Why not just use cfg.Name?

derekcollison (changes requested):
We should just have the debug line figure out on the fly which name to use instead of passing things around IMO.

server/jetstream_cluster.go
ripienaar: right didnt see this targetted main, fair.

derekcollison:
Any updates here?

server/jetstream_cluster.go
neilalexander: `cfg.Name` is set from `rg.Name`, which seems to be the raft ID, not the actual asset name.

derekcollison: Yes its the store name. But if you are trying to tie these together at this level we would need to be more expressive then assetName I believe.

server/jetstream_cluster.go
derekcollison: Yes this will need to be for dev.

server/jetstream_cluster.go
derekcollison: E.g. a consumer would also need which stream it belongs to.

Let's chat about this a bit on a Zoom at some point.

PR #3903, opened 2023-02-23
Set code owners, tweak PR template

neilalexander:
This will remove the "cc" from the PR template (which typically doesn't work for outside contributors) and instead replaces it with a code owners file, which will mean that PRs against this repository will automatically request review from the new @nats-io/server team.


derekcollison (approved):
LGTM

PR #3904, opened 2023-02-23
[IMPROVED] When becoming meta-leader optionally send current snapshot to followers.

derekcollison:
This can help sync on restarts and improve ghost ephemerals. Also added more code to suppress respnses and API audits when we know we are recovering.

/cc @nats-io/core

### DIFF
--- a/server/jetstream_cluster.go
+++ b/server/jetstream_cluster.go
@@ -127,9 +127,10 @@ type streamAssignment struct {
 	Reply   string        `json:"reply"`
 	Restore *StreamState  `json:"restore_state,omitempty"`
 	// Internal
-	consumers map[string]*consumerAssignment
-	responded bool
-	err       error
+	consumers  map[string]*consumerAssignment
+	responded  bool
+	recovering bool
+	err        error
 }
 
 // consumerAssignment is what the meta controller uses to assign consumers to streams.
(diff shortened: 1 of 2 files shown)


server/jetstream_cluster.go
```diff
@@ -1070,16 +1084,16 @@ func (js *jetStream) monitorCluster() {
...
-			}
-			js.processLeaderChange(isLeader)
-			if isLeader && !beenLeader {
-				beenLeader = true
-				if n.NeedSnapshot() {
-					if err := n.InstallSnapshot(js.metaSnapshot()); err != nil {
-						s.Warnf("Error snapshotting JetStream cluster state: %v", err)
+				// Install a snapshot as we become leader. We will also send to the cluster.
+				if snap := doSnapshot(); snap != nil {
+					// If we are caught up distibute our current state to followers.
```
wallyqs, line 1093:
```suggestion
					// If we are caught up distribute our current state to followers.
```

wallyqs (approved):
LGTM

server/jetstream_cluster.go
derekcollison: Fixed. Not sure why mispell does not catch those..

neilalexander (approved):

PR #3905, opened 2023-02-23
[IMPROVED] Make sure preferred peer for stepdown is healthy.

derekcollison:

### DIFF
--- a/server/raft.go
+++ b/server/raft.go
@@ -1281,7 +1281,6 @@ func (n *raft) StepDown(preferred ...string) error {
 	n.debug("Being asked to stepdown")
 
 	// See if we have up to date followers.
-	nowts := time.Now().UnixNano()
 	maybeLeader := noLeader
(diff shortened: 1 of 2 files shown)


server/raft.go
```diff
@@ -1290,21 +1289,42 @@ func (n *raft) StepDown(preferred ...string) error {
 			preferred = nil
 		}
 	}
+	// Can't pick ourselves.
+	if maybeLeader != noLeader && maybeLeader == n.id {
```
neilalexander, line 1293: Not a big deal but I guess this could just be `if maybeLeader == n.id`

neilalexander (approved):
LGTM!

Jarema (approved):
LGTM

server/raft.go
derekcollison: Will update. Thanks.

wallyqs (approved):

PR #3907, opened 2023-02-24
[FIXED] Snapshots would not compact through applied. 

derekcollison:
This means a subsequent request for exactly applied index would return that entry only and not the full state snapshot.
Also fixed bug where we would not snapshot when we should.


wallyqs (approved):
LGTM

Jarema (approved):
Lgtm

PR #3909, opened 2023-02-24
Don't compact past `applied` in `InstallSnapshot`

neilalexander:
This seems to be an off-by-one again, in this case `snap.lastIndex` is already set to `n.applied` so we'll compact up to but not including the applied entry.

With this, `TestJetStreamClusterStreamCatchupNoState` no longer churns on `Wrong index, ae is...` in RAFT indefinitely.


derekcollison:
Found the issue.

PR #3911, opened 2023-02-26
[FIXED] If a truncate for a raft WAL failed we could spin.

derekcollison:
When the truncate failed we would continue to process append entries without changing state.
This could result in raft stores filling up disks and systems becoming unresponsive.

The server log could also fill up with warnings similar to this one.

```[WRN] RAFT [z3WIzPtj - S-R2F-76KY41ow] Wrong index, ae is &{leader:z3WIzPtj term:1 commit:100 pterm:1 pindex:100 entries: 1}, index stored was 115, n.pindex is 100```


wallyqs (approved):
LGTM 👍

PR #3914, opened 2023-02-27
Add raft query parameter to /jsz to include raft group info

wallyqs:
This adds a new `raft` query parameter to be able to map a stream/consumer name with its active raft group:

```
/jsz?consumers=1&raft=1
```

```js
 "stream_raft_group": "S-R3F-8Qml10Ei",
"consumer_raft_groups": [
  {
    "name": "1",
    "raft_group": "C-R3F-5c87s2aq"
  },
  {
    "name": "10",
    "raft_group": "C-R3F-PF8WWIue"
  },
  {
    "name": "100",
    "raft_group": "C-R3F-r58xhrRe"
  },
  {
    "name": "12",
    "raft_group": "C-R3F-faSs9EY1"
  },
```

PR #3915, opened 2023-02-27
[main] Add raft query parameter to /jsz to include raft group info

wallyqs:
backport of https://github.com/nats-io/nats-server/pull/3914


neilalexander (approved):
LGTM!

PR #3920, opened 2023-02-28
Change logtime_utc default to false

bruth:
This ensures that the existing, default behavior of the server emitting the local timezone is not changed.


neilalexander (approved):
Good catch, thanks!

Jarema (approved):
LGTM

PR #3923, opened 2023-02-28
Update LEAFZ to include leafnode server/connection name

### LINKED ISSUE #3921: Add Server Name to Leafz Monitoring Endpoint
## Feature Request
In the monitoring endpoint, for /Leafz, add serverName attribute of the Leafnode that connected. 

#### Use Case:
The IP doesn't give us all of the information that we need to uniquely identify which of the many Leafnodes that are connecting to the core nats server. For example, we have connections coming outside of our network that get DNAT to a different IP that the original leafnode server was using. Also, deploying within a K8S environment uses the internal K8S IP and tracking it down can be difficult.

#### Proposed Change:
Change the LeafInfo struct to include Server Name of the Leaf node and then within the loop that is building the LeafNodes call ln.leaf.remoteServer for the cluster name.

Anyone that uses K8S or bridges networks together where DNAT is involved. 

#### Alternative Approaches
Build out a better way of showing the entire nats / leafnodes topology.

codegangsta:
resolves #3921 

 - [x] Tests added

Adds `Name` field to `LEAFZ` info for incoming and outgoing leafnode connections. This should allow observability tools to more easily identify leaf nodes without having to require SYS accounts set up for each leaf node connection.

### DIFF
--- a/server/leafnode.go
+++ b/server/leafnode.go
@@ -888,7 +888,13 @@ func (s *Server) createLeafNode(conn net.Conn, rURL *url.URL, remote *leafNodeCf
 	}
 	now := time.Now().UTC()
 
-	c := &client{srv: s, nc: conn, kind: LEAF, opts: defaultOpts, mpay: maxPay, msubs: maxSubs, start: now, last: now}
+	var clientOpts = ClientOpts{
+		Name:     opts.ServerName,
+		Verbose:  defaultOpts.Verbose,
+		Pedantic: defaultOpts.Pedantic,
+		Echo:     defaultOpts.Echo,
+	}
+	c := &client{srv: s, nc: conn, kind: LEAF, opts: clientOpts, mpay: maxPay, msubs: maxSubs, start: now, last: now}
 	// Do not update the smap here, we need to do it in initLeafNodeSmapAndSendSubs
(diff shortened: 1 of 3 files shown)


derekcollison (approved):
LGTM

server/leafnode.go
```diff
@@ -888,7 +888,13 @@ func (s *Server) createLeafNode(conn net.Conn, rURL *url.URL, remote *leafNodeCf
 	}
 	now := time.Now().UTC()
 
-	c := &client{srv: s, nc: conn, kind: LEAF, opts: defaultOpts, mpay: maxPay, msubs: maxSubs, start: now, last: now}
+	var clientOpts = ClientOpts{
+		Name:     opts.ServerName,
```
codegangsta, line 892: This is the only bit I'm not feeling 100% about. For outgoing connections I set the Name as the current server name. I believe this is probably what we want, since folks looking at LEAFZ should be able to tell the difference between incoming and outgoing connections here, but would like some feedback

derekcollison: LMK if you want me to merge it.

codegangsta: Go for it!

PR #3925, opened 2023-03-01
Updated LEAFZ names to use remoteServer name/id and added is_spoke

codegangsta:
This is a followup on #3923 that improves LEAFZ names to use `ln.leaf.remoteServer` instead of the connection name.

I've also added `is_spoke` to the leafz LeafInfo struct so that we can easily tell if a leaf connection is a hub or spoke.


derekcollison (approved):
LGTM

PR #3928, opened 2023-03-01
Revert behavior for JetStream messages published directly from client.

derekcollison:
This matches inline behavior for messages from a client.


neilalexander (approved):

Jarema (approved):
LGTM

PR #3932, opened 2023-03-01
Fixed flapping test since we have two streams and order in slice not guaranteed.

derekcollison:


wallyqs (approved):
Thanks for the fix!

PR #3952, opened 2023-03-10
Only mark fss dirty if a change is made

neilalexander:
Otherwise we might unnecessarily write out the fss file for no reason, which is possibly happening inline.


derekcollison (approved):
LGTM!

PR #3956, opened 2023-03-13
Don't use string builder in `subjString` (it is slow)

neilalexander:
We were already converting `key` to a `string` in the loop, so the use of a `strings.Builder` is additionally wasteful and burns CPU cycles for no reason. Let's just convert `key` once instead — when it becomes a string, it already stops referring to the underlying array anyway.


Jarema (approved):
LGTM!

PR #3961, opened 2023-03-14
adds a nightly build action and build nightly rc from main

ripienaar:


ripienaar:
This creates a reuable action for building images, then use that in a new `rc` image, leaving the `nightly` one as is so it wont fail.

If this new build work I will upate the `nightly` tomorrow to use the same reusable action.

wallyqs (approved):
LGTM

bruth:
Given your last suggestion on Slack, I think `nightly-main` is more intuitive than `rc`.

.github/workflows/rc_nightly.yaml
```diff
@@ -0,0 +1,17 @@
...
+
+jobs:
+  rc_release:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: ./.github/actions/nightly-release
+        with:
+          go: "1.19"
+          branch: main
+          label: nightly-main
```
ripienaar, line 17: updated as discussed, will merge on green.

PR #3964, opened 2023-03-14
Fix leaf client connection failing in OCSP setups

### LINKED ISSUE #3959: When OCSP is enabled, if leaf mTLS configured, mTLS handshake succeeds but leaf connection fails
## When OCSP is enabled, if leaf mTLS configured, mTLS handshake succeeds but leaf connection fails

In server up to 2.9.15, where OCSP is enabled, operator is able to configure a Leaf listener with mTLS enablement (client auth as well as server auth), but ultimate leaf connection fails after successful mTLS handshake.

1. Configure a NATS server with the OCSP staple feature enabled
2. Add a leaf configuration with a TLS block with `verify: true`
3. Attempt NATS leaf remote connection to the NATS Server with mTLS (client sending valid cert)

#### Expected result:

mTLS handshake succeeds and Leaf connection established

#### Actual result:

mTLS handshake succeeds, but Leaf connection is rejected by the OCSP-enabled "hub".  Debug entry observed in hub logs:

`[1] 2023/03/13 20:40:24.564561 [ERR] 10.xx.xx.xx:42480 - lid:11392 - TLS leafnode handshake error: Leafnode client missing OCSP Staple`

If OCSP is disabled at hub, mTLS handshake and Leaf connection succeed.

wallyqs:
This changes ocsp handling so that leafnode remote client connections behave more similarly to how regular nats clients work, rather than mimic the bidirectional verification that happens for routes/gateways.

Resolves #3959

### DIFF
--- a/server/ocsp.go
+++ b/server/ocsp.go
@@ -433,18 +433,18 @@ func (srv *Server) NewOCSPMonitor(config *tlsConfigKind) (*tls.Config, *OCSPMoni
 			}, nil
 		}
 
-		// Check whether need to verify staples from a client connection depending on the type.
+		// Check whether need to verify staples from a peer router or gateway connection.
 		switch kind {
 		case kindStringMap[ROUTER], kindStringMap[GATEWAY]:
 			tc.VerifyConnection = func(s tls.ConnectionState) error {
 				oresp := s.OCSPResponse
 				if oresp == nil {
-					return fmt.Errorf("%s client missing OCSP Staple", kind)
+					return fmt.Errorf("%s peer missing OCSP Staple", kind)
 				}
 
-				// Client route connections will verify the response of the staple.
+				// Peer connections will verify the response of the staple.
 				if len(s.VerifiedChains) == 0 {
-					return fmt.Errorf("%s client missing TLS verified chains", kind)
+					return fmt.Errorf("%s peer missing TLS verified chains", kind)
 				}
 
 				chain := s.VerifiedChains[0]
@@ -453,7 +453,7 @@ func (srv *Server) NewOCSPMonitor(config *tlsConfigKind) (*tls.Config, *OCSPMoni
 
 				resp, err := ocsp.ParseResponseForCert(oresp, leaf, parent)
 				if err != nil {
-					return fmt.Errorf("failed to parse OCSP response from %s client: %w", kind, err)
+					return fmt.Errorf("failed to parse OCSP response from %s peer: %w", kind, err)
 				}
 				if resp.Certificate == nil {
 					if err := resp.CheckSignatureFrom(parent); err != nil {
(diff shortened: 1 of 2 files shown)


server/ocsp.go
```diff
@@ -558,10 +558,11 @@ func (s *Server) configureOCSP() []*tlsConfigKind {
 			tlsConfig: config,
 			tlsOpts:   opts,
 			apply: func(tc *tls.Config) {
-
 				// RequireAndVerifyClientCert is used to tell a client that it
 				// should send the client cert to the server.
-				tc.ClientAuth = tls.RequireAndVerifyClientCert
+				if opts.Verify {
```
tbeets, line 563: Will `opts.Verify` reflect `verify:` and `verify_and_map:` TLS config block entries?

server/ocsp.go
```diff
@@ -558,10 +558,11 @@ func (s *Server) configureOCSP() []*tlsConfigKind {
...
+					tc.ClientAuth = tls.RequireAndVerifyClientCert
+				}
 				// GetClientCertificate is used by a client to send the client cert
 				// to a server. We're a server, so we must not set this.
 				tc.GetClientCertificate = nil
```
tbeets, line 568: Comment is confusing.  We are the server (leaf hub) in this case.  Does it mean that we would only use `tc.GetClientCertificate` to obtain the client's certificate just used in the successful TLS handshake?

If we as Leaf Hub are configured for `verify_and_map: true` then post-handshake we need to evaluate cert attributes for user mapping purposes.   Setting `nil` here defeat that function?

server/ocsp.go
```diff
@@ -435,7 +435,7 @@ func (srv *Server) NewOCSPMonitor(config *tlsConfigKind) (*tls.Config, *OCSPMoni
 
 		// Check whether need to verify staples from a client connection depending on the type.
```
tbeets, line 436: This comment is confusing.   Feels like throughout we are introducing confusion (at least I get confused!) by mixing the c-word "client" as a mix of TLS directional role and NATS client.   For OCSP and TLS code recommend trying to disambiguate and lean towards TLS directional role for clarity.

Should it read: 

// Check whether need to verify **server's** staple from **outbound** connection depending on the type.

philpennock (approved):
So in discussion, the core problem here is that I made a mistake in the original specification for OCSP support, in that we conflate (1) TLS server fetching and managing its own OCSP staple for presenting to clients; with (2) the TLS client demanding a staple of the peer.

Solving this fully will require a new OCSP block keyword of `peer`, and probably letting the `ocsp{}` block be embedded in a `tls{}` block, as originally spec'd.

In the meantime, given the split in administrative domains between "cluster/supercluster" and "leafnodes", it makes sense to just have the peer policy match for cluster/supercluster without an explicit `peer` block, and turn off OCSP for leafnodes.

The trade-off which we are accepting with this PR is that the NATS server connecting as a leafnode to a hub is also losing the ability to verify the staple as presented by the hub.  Fixing this will require the ability to set `tls{ocsp{peer: ...}}` in config.  In the meantime this is a withdrawal of functionality to get to a "more likely to work" deployment.  We will fix more correctly in the near-ish future.

server/ocsp.go
wallyqs: I simplified the comment and also added a `verify_and_map` test to confirm.

server/ocsp.go
wallyqs: Updated the terminology to use `peers` instead of server/client since cannot make the distinction at this stage yet.

server/ocsp.go
wallyqs: Yes verify will be enabled if `verify_and_map` is enabled.

philpennock (approved):

tbeets (approved):
LGTM

PR #3970, opened 2023-03-16
Remove msgs from interest based stream on consumer snapshot.

derekcollison:
Make sure to cleanup stream messages on a follower consumer for an interest based stream when the consumer leader sends a state snapshot.

When we would receive a state snapshot vs individual replicated acks we would not remove the stream's messages.


wallyqs (approved):
LGTM

PR #3972, opened 2023-03-16
move dev nightly to same shared action

ripienaar:


ripienaar:
Both nightlies worked last night on schedule, so just updating the dev nightly to use the same shared action.

derekcollison (approved):
LGTM

PR #3975, opened 2023-03-16
Use pooled buffer for flushing encrypted message blocks

neilalexander:
We already have a pool of buffers that are going to be more than big enough most of the time, so this should save a lot of allocations if the file store is encrypted.


derekcollison (approved):
LGTM

PR #3976, opened 2023-03-16
Bump JWT lib

wallyqs:


philpennock (approved):

derekcollison:
Any thoughts?

wallyqs:
We should wait for v2.10 just in case, I will close for now.

PR #3978, opened 2023-03-17
Don't recycle buffer more than once

neilalexander:
This should fix a panic with overlapping slices ending up in the message block buffer pool.


Jarema (approved):
LGTM!

PR #3979, opened 2023-03-17
Remove snapshotting of cores and maxprocs.

derekcollison:
We used to snapshot to avoid runtime costs, but with containers etc. probably easiest to call at runtime.
Don't think it is as expensive as it once was either.


neilalexander (approved):
LGTM

wallyqs (approved):

PR #3980, opened 2023-03-19
[FIXED] Fixed an issue with consumer states growing and causing instability.

derekcollison:
Under certain circumstances, a replicated consumer configured for explicit ack could have the consumer state for the replicas grow very large with ack pending that should have been cleared.

This bug could trip from an off by one bug during out of order acks, but would only manifest if the ack pending did not reach 0 which would clear and reset the ack floor properly and mask the issue.

Also in this PR is some work on the general test framework to introduce network proxy for emulating RTT and throughput into the base cluster vs just the gateway connections.


wallyqs (approved):
LGTM

PR #4001, opened 2023-03-29
[ADDED] Multiple routes and ability to have per-account routes

kozlovic:
New configuration fields:
```
cluster {
   ...
   pool_size: 5
   accounts: ["A", "B"]
}
```

The configuration `pool_size` in the example above means that this
server will create 5 routes to a remote server, assuming that that
server has the same `pool_size` setting.

Accounts (which are not part of the `accounts[]` configuration)
are assigned a specific route in this pool, and this will be the
same route on all servers in the cluster.

Accounts that are defined in the `accounts` field will each have
a dedicated route connection. This will allow suppression of the
account name in some of the route protocols, reducing bytes transmitted
which may increase performance.


server/route.go
```diff
@@ -1396,98 +1616,311 @@ func (s *Server) createRoute(conn net.Conn, rURL *url.URL) *client {
...
-		// If the INFO contains a Gateway URL, add it to the list for our cluster.
-		if info.GatewayURL != "" && s.addGatewayURL(info.GatewayURL) {
-			s.sendAsyncGatewayInfo()
+		// In pool mode, if we did not yet reach the cap, try to connect a new connection
+		if pool && didSolicit && sz != s.routesPoolSize {
+			s.startGoRoutine(func() {
+				select {
+				case <-time.After(time.Duration(rand.Intn(100)) * time.Millisecond):
+				case <-s.quitCh:
+					s.grWG.Done()
```
derekcollison, line 1871: Just defer at top of func for optics? Behavior is correct already I think.

server/server.go
```diff
@@ -828,16 +912,32 @@ func (s *Server) configureAccounts() error {
 		for _, si := range acc.imports.services {
 			if v, ok := s.accounts.Load(si.acc.Name); ok {
 				si.acc = v.(*Account)
+				// TODO: Not sure if it is possible for an account to have a
+				// service import from itself, but if that is the case,
```
derekcollison, line 916: Yes it is possible now to allow for latency tracking inside your own account.

server/server.go
```diff
@@ -828,16 +912,32 @@ func (s *Server) configureAccounts() error {
...
+				// we are already lock, otherwise use locking to protect
```
derekcollison, line 917: lock -> locked

server/server.go
```diff
@@ -1736,8 +1874,9 @@ func (s *Server) Start() {
 		// In operator mode, when the account resolver depends on an external system and
 		// the system account is the bootstrapping account, start fetching it.
 		if len(opts.TrustedOperators) == 1 && opts.SystemAccount != _EMPTY_ && opts.SystemAccount != DEFAULT_SYSTEM_ACCOUNT {
+			opts := s.getOpts()
 			_, isMemResolver := ar.(*MemAccResolver)
-			if v, ok := s.accounts.Load(s.opts.SystemAccount); !isMemResolver && ok && v.(*Account).claimJWT == "" {
+			if v, ok := s.accounts.Load(opts.SystemAccount); !isMemResolver && ok && v.(*Account).claimJWT == "" {
```
derekcollison, line 1879: _EMPTY_

derekcollison: vs ""

derekcollison (commented):
Looks pretty good! Few minor comments. Thanks again!

server/jetstream.go
```diff
@@ -1029,9 +1029,12 @@ func (a *Account) EnableJetStream(limits map[string]JetStreamAccountLimits) erro
 	}
 
 	js.mu.Lock()
-	if _, ok := js.accounts[a.Name]; ok && a.JetStreamEnabled() {
+	if jsa, ok := js.accounts[a.Name]; ok {
+		a.mu.Lock()
```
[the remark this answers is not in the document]
kozlovic: No, since I am assigning `a.js = jsa` just below.

server/route.go
kozlovic: Actually no (and I have added a comment in the upcoming changes). The function connectToRoute() is itself calling s.grWG.Done(), so we need to do it only if that function is not invoked. Otherwise it would lead to two calls to Done() for the same Add(), which is bad.

server/route.go
```diff
@@ -1185,10 +1380,34 @@ func (s *Server) sendSubsToRoute(route *client) {
 				eSize += 5
 			}
 		}
-		a.mu.RUnlock()
-		return true
-	})
-	s.mu.Unlock()
+	}
+	// Send over our account subscriptions.
+	accs := make([]*Account, 0, 32)
```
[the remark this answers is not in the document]
kozlovic: I made it 1024, can always change it later.

kozlovic:
@derekcollison I have updated the PR with 2 commits, one addressing your comments, the second fixes some tests that missed some cleanup code that caused possibly other tests to fail.

However, I find that the following test is often failing (even on my machine, macOS bare metal):
```
=== RUN   TestNoRaceJetStreamSuperClusterStreamMoveLongRTT
    norace_test.go:5588: require no error, but got: context deadline exceeded
--- FAIL: TestNoRaceJetStreamSuperClusterStreamMoveLongRTT (23.24s)
```
This is true even in `dev` and `main` branches so I am confident that this is unrelated to changes in this PR.

More worrisome, is this test:
```
=== RUN   TestNoRaceJetStreamInterestStreamCheckInterestRaceBug
    norace_test.go:7043: Ackfloor not correct yet
--- FAIL: TestNoRaceJetStreamInterestStreamCheckInterestRaceBug (24.89s)
```
It passes 10/10 on `main` but fails 10/10 on `dev` (so again, unrelated to this PR). I can't get the build green because of these 2.

derekcollison (approved):
LGTM!

For the test fixes, possibly to do those on main and we can inherit them or do separate PR for them on main?

kozlovic:
For the tests, done in: https://github.com/nats-io/nats-server/pull/4005

kozlovic:
@derekcollison I cherry-picked the tests fixes an submitted the PR to main, which was approved by @wallyqs and merged into main, then into dev and I have rebased this PR out of dev, which dropped 3 commits that were now upstream.

kozlovic:
@derekcollison I have rebased and fixed conflicts and run on Travis is green.

PR #4004, opened 2023-03-30
File store compression

neilalexander:
This PR adds internal support for compression of message blocks in the file store. Message blocks will be compressed when they reach their maximum configured block size and we have moved onto the next block, so as to avoid slowing down on-going writes as far as possible.

Compressed blocks are preambled with a small amount of metadata describing the compression algorithm (although at this point only S2 compression is supported) and compression takes place before encryption (so as to maximise the efficiency of the compression where a stream contains redundant or repeated information in the messages).

This PR **does not** provide any way to configure compression from the stream level yet (or indeed any other level exposed to clients or operators). It also does not go out of its way to compress old uncompressed blocks after turning on compression unless they are compacted or truncated. These omissions will be addressed in follow-up PRs.

New permutations have been added to the file store tests to also ensure test coverage when compression is enabled.

We should note that, when file store encryption is not in use, storage-level or filesystem-level deduplication or compression (such as those provided by some disk arrays or filesystems like ZFS) may end up being considerably more space-efficient than file store compression would. Conversely, when file store encryption is enabled, enabling compression at the file store level may achieve better ratios.

### DIFF
--- a/server/filestore.go
+++ b/server/filestore.go
@@ -57,6 +57,8 @@ type FileStoreConfig struct {
 	AsyncFlush bool
 	// Cipher is the cipher to use when encrypting.
 	Cipher StoreCipher
+	// Compression is the algorithm to use when compressing.
+	Compression StoreCompression
 }
 
 // FileStreamInfo allows us to remember created time.
@@ -83,6 +85,24 @@ func (cipher StoreCipher) String() string {
 		return "None"
 	default:
 		return "Unknown StoreCipher"
+	}
+}
+
+type StoreCompression uint8
+
+const (
+	NoCompression StoreCompression = iota
+	S2Compression
+)
+
+func (alg StoreCompression) String() string {
+	switch alg {
+	case NoCompression:
+		return "None"
+	case S2Compression:
+		return "S2"
+	default:
+		return "Unknown StoreCompression"
 	}
 }
 
@@ -145,6 +165,7 @@ type msgBlock struct {
 	mfd     *os.File
 	ifn     string
 	ifd     *os.File
+	cmp     StoreCompression // Effective compression at the time of loading the block
 	liwsz   int64
 	index   uint32
 	bytes   uint64 // User visible bytes count.
@@ -226,6 +247,8 @@ const (
 	consumerDir = "obs"
 	// Index file for a consumer.
 	consumerState = "o.dat"
+	// The suffix that will be given to a new temporary block during compression.
+	compressTmpSuffix = ".tmp"
 	// This is where we keep state on templates.
 	tmplsDir = "templates"
 	// Maximum size of a write buffer we may consider for re-use.
@@ -984,6 +1007,11 @@ func (mb *msgBlock) rebuildStateLocked() (*LostStreamData, error) {
 			return nil, err
 		}
 		mb.bek.XORKeyStream(buf, buf)
+	}
+
+	// Check for compression.
+	if buf, err = mb.decompressIfNeeded(buf); err != nil {
+		return nil, err
 	}
 
 	mb.rbytes = uint64(len(buf))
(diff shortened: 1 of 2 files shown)


derekcollison:
Will take a look in am, thanks!

If you want you can rebase on dev to get a clean travis run now.

server/filestore.go
```diff
@@ -3791,6 +3869,166 @@ func (fs *fileStore) writeMsgRecord(seq uint64, ts int64, subj string, hdr, msg
...
+	}
+
+	meta := &CompressionInfo{}
+	if _, err := meta.UnmarshalMetadata(origBuf); err != nil {
+		// An error is only returned here if there's a problem with parsing
+		// the metadata. If the file has no metadata at all, no error is
+		// returned and the algorithm defaults to no compression.
+		return fmt.Errorf("failed to read existing metadata header: %w", err)
+	}
+	if meta.Algorithm == alg {
```
derekcollison, line 3927: If we think this is common should we just peek at this part before loading in the whole block?

server/filestore.go
```diff
@@ -3791,6 +3869,166 @@ func (fs *fileStore) writeMsgRecord(seq uint64, ts int64, subj string, hdr, msg
...
+		mb.bek = bek
+		mb.bek.XORKeyStream(cmpBuf, cmpBuf)
+	}
+
+	// Write the new block data (which might be compressed or encrypted) to the
+	// temporary file.
+	if n, err := tmpFD.Write(cmpBuf); err != nil {
+		return fmt.Errorf("failed to write to temporary file: %w", err)
+	} else if n != len(cmpBuf) {
+		return fmt.Errorf("short write to temporary file (%d != %d)", n, len(cmpBuf))
```
derekcollison, line 3983: Clean up the file since we did write something?

server/filestore.go
```diff
@@ -3791,6 +3869,166 @@ func (fs *fileStore) writeMsgRecord(seq uint64, ts int64, subj string, hdr, msg
...
+	}
+	if err := tmpFD.Sync(); err != nil {
+		return fmt.Errorf("failed to sync temporary file: %w", err)
```
derekcollison, line 3986: Same

server/filestore.go
```diff
@@ -3791,6 +3869,166 @@ func (fs *fileStore) writeMsgRecord(seq uint64, ts int64, subj string, hdr, msg
...
+	}
+	if err := tmpFD.Close(); err != nil {
```
derekcollison, line 3988: This should be a defer up above so that any exit will always close.

server/filestore.go
```diff
@@ -3791,6 +3869,166 @@ func (fs *fileStore) writeMsgRecord(seq uint64, ts int64, subj string, hdr, msg
...
+		return fmt.Errorf("failed to close temporary file: %w", err)
+	}
+
+	// Now swap the files around. First we'll move the original file aside,
+	// then we'll move the temporary file into place, and then finally we will
+	// delete the original file.
+	if err := os.Rename(origFN, oldFN); err != nil {
+		return fmt.Errorf("failed to move original file aside: %w", err)
```
derekcollison, line 3996: Remove tmp one?

server/filestore.go
```diff
@@ -3791,6 +3869,166 @@ func (fs *fileStore) writeMsgRecord(seq uint64, ts int64, subj string, hdr, msg
...
+	}
+	if err := os.Rename(tmpFN, origFN); err != nil {
```
derekcollison, line 3998: I think we only want this one? Do we need the one above to move old then remove it?

server/filestore.go
```diff
@@ -7338,3 +7593,107 @@ func (ts *templateFileStore) Store(t *streamTemplate) error {
...
+	c.Algorithm = StoreCompression(b[3])
+	c.OriginalSize, n = binary.Uvarint(b[4:])
+	if n <= 0 {
+		return 0, fmt.Errorf("metadata incomplete")
+	}
+	return 4 + n, nil
+}
+
+func (alg StoreCompression) Compress(buf []byte) ([]byte, error) {
+	if len(buf) < checksumSize {
```
derekcollison, line 7633: Probably should be a separate const threshold.

server/filestore.go
neilalexander: I've made sure that errors result in clean-up, but I wanted to make sure the descriptor was closed before the rename.

server/filestore.go
neilalexander: Since we need to preserve the checksum, the `checksumSize` seems like it makes sense, as we need to make sure the buffer is big enough to do that.

derekcollison: What I mean is probably does not make sense if block is below 1-2MB or so.

neilalexander: I wonder if that might be surprising behaviour, what threshold do you think makes most sense?

derekcollison: 1M? WDYT?

neilalexander: Looking at the constants, maybe `defaultSmallBlockSize / 2` makes most sense, which would be 0.5MB today, but I’m also aware of the fact that some of the unit tests set the block size to be deliberately small and by adding in a threshold like this, we may lose out on test coverage of compression by accident.

derekcollison: Make your best call and we can get this merged in tomorrow.

neilalexander: I think the best thing to do for now is to leave it as-is so that the test coverage is maximal. If we think that compressing small blocks is hurting performance we can always follow it up later with another PR.

derekcollison (approved):
LGTM

PR #4007, opened 2023-03-31
Updates to preacks when multiple consumers are mutually exclusive

derekcollison:
When consumers where mutually exclusive and acks came in before a message we did not properly cleanup the messages.


wallyqs (approved):
LGTM

PR #4010, opened 2023-03-31
Allow republish for mirroring/sourcing streams

### LINKED ISSUE #4000: JetStream RePublish should work with SOURCES and MIRROR message ingest paths
A stream may ingest messages from subscription on one or more defined SUBJECT filters, but also may ingest messages replicated from an upstream via SOURCES or MIRROR configuration.

In the current release (2.9.x), if RePublish is configured on the stream, it acts only on ingested messages from SUBJECT filter(s).  It would be desirable and useful for RePublish to work on any ingested message into the stream.

jnmoyne:
 - [ ] Tests added

Resolves #4000

Removes the check on the republish filter needing to have an overlap with the listened subjects, as with the new stream subject transformation changes we do not assume anymore that the subjects in a stream must match the subjects being listened to.

This allows the use of republish for mirroring and/or sourcing streams

### DIFF
--- a/server/stream.go
+++ b/server/stream.go
@@ -1291,19 +1291,9 @@ func (s *Server) checkStreamCfg(config *StreamConfig, acc *Account) (StreamConfi
 	if cfg.RePublish != nil {
 		// Check to make sure source is a valid subset of the subjects we have.
 		// Also make sure it does not form a cycle.
-		var srcValid bool
 		// Empty same as all.
 		if cfg.RePublish.Source == _EMPTY_ {
 			cfg.RePublish.Source = fwcs
-		}
-		for _, subj := range cfg.Subjects {
-			if SubjectsCollide(cfg.RePublish.Source, subj) {
-				srcValid = true
-				break
-			}
-		}
-		if !srcValid {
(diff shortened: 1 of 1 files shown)


server/stream.go
```diff
@@ -1291,20 +1291,10 @@ func (s *Server) checkStreamCfg(config *StreamConfig, acc *Account) (StreamConfi
 	if cfg.RePublish != nil {
 		// Check to make sure source is a valid subset of the subjects we have.
 		// Also make sure it does not form a cycle.
-		var srcValid bool
 		// Empty same as all.
 		if cfg.RePublish.Source == _EMPTY_ {
 			cfg.RePublish.Source = fwcs
 		}
-		for _, subj := range cfg.Subjects {
```
derekcollison, line 1299: Anyway to tell if their config will never publish any messages?

derekcollison: Fair, I wonder though if we do best effort to some degree? Or you think leave without checks/

jnmoyne: I would leave without checks, in case someone sets republish with a filter that doesn't match any messages in the stream it's pretty easy now to check that using `nats stream subjects` which will show you exactly what is in the stream not just trying to check with sourcing, sourcing chains, subject transforms, etc...

derekcollison (approved):
LGTM

PR #4013, opened 2023-04-02
[IMPROVED] Protect against concurrent creation of streams and consumers.

derekcollison:
Also make sure we have exited monitoring routines when doing resets for both streams and consumers.

### DIFF
--- a/server/stream.go
+++ b/server/stream.go
@@ -374,7 +374,10 @@ func (a *Account) addStreamWithAssignment(config *StreamConfig, fsConfig *FileSt
 		ifwg.Wait()
 	} else {
 		ifwg.Add(1)
-		defer ifwg.Done()
+		defer func() {
+			jsa.inflight.Delete(cfg.Name)
+			ifwg.Done()
+		}()
 	}
 
 	js, isClustered := jsa.jetStreamAndClustered()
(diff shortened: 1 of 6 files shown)


server/stream.go
```diff
@@ -366,10 +366,21 @@ func (a *Account) addStreamWithAssignment(config *StreamConfig, fsConfig *FileSt
 		return nil, ApiErrors[JSStreamReplicasNotSupportedErr]
 	}
 
+	// Make sure we are ok when these are done in parallel.
+	var wg sync.WaitGroup
+	v, loaded := jsa.inflight.LoadOrStore(cfg.Name, &wg)
```
neilalexander, line 371: Minor nit but `LoadOrStore(cfg.Name, &sync.WaitGroup{})` is clearer and doesn't pollute the scope with extra names.

neilalexander (approved):
LGTM!

Jarema (approved):
LGTM!

server/stream.go
derekcollison: Good call! Will update.

derekcollison: Done.

wallyqs (approved):
LGTM

Issue #4014, opened 2023-04-03
Disconnection of a NATS server in a cluster is not detected by ping-pong mechanism


sandykellagher:
**Defect**
This is more or less a follow on to #3682 

We are using three NATS servers in a cluster, and see a bug which prevents the loss of a particular server being detected by the other two servers. This matters because we use queue group subscriptions: when a server goes offline, we need the queue group subscription for services on the dead server to be removed. Otherwise, NATS can end up selecting entities on the dead server as the nominated responder for pub messages to the queue group.

In our system, there is continual traffic exchanged between all three servers during normal operation. And in short, this continuous traffic results in all outbound Pings from one server to another being held off/delayed indefinitely. And because the Pings are held off, the servers don't detect a stale connection, and hence doesn't close the connection and attempt to reconnect.

I believe I understand the issue.

In NATS server client.go:: processPingTimer() there is a check to test whether to delay sending an outgoing ping, in two cases:

- we recently (within specified pingInterval) received traffic (a message or sub/unsub) from the remote end
- we recently received a ping from the remote end (Remote ping)

This makes perfect sense: incoming receive messages mean we still have a link and don't need to send a ping.

However, the first test above is derived from the client.last field ("last packet" time) which is set for both incoming and outgoing traffic on this link. 

I think the solution is to store a separate client.ping.lastRxMsg field which is updated for receive traffic only, and then use that field in the test of whether to hold off sending a ping.

Also, with this change, I think the same logic can be applied to all connection types (CLIENT, ROUTER, GATEWAY, LEAF). So we can remove the tests to always send ping for a GATEWAY.

**OS/Container environment**
Linux ARM64 - but not a factor

**Steps or code to reproduce the issue**
Configure 3 server cluster, and then break the connection to one of the servers

**Expected result**
The remaining servers should automatically detect loss of connection, with the detection time as configured by ping_interval and ping_max parameters.

**Actual result**
The remaining servers do eventually detect loss of connection via a WriteBuffer deadline being exceeded, but it takes a very long time (presumably) depending on the volume of traffic and the size of the buffer.

sandykellagher:
I've closed PR #4015 in favour of #4016

PR #4016, opened 2023-04-03
Alternative fix for issue 4014, where we always send pings on ROUTER connections

sandykellagher:
Alternative fix for issue https://github.com/nats-io/nats-server/issues/4014: where we always send pings for ROUTER, GATEWAY and LEAF spoke connections.

This is my own original work that I license to the project

[ x] Link to issue, e.g. Resolves #NNN
 Documentation added (if applicable)
 Tests added
[ x] Branch rebased on top of current main (git pull --rebase origin main)
[ x] Changes squashed to a single commit (described [here](http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html))
[x] Build is green in Travis CI
[ x] You have certified that the contribution is your original work and that you license the work to the project under the [Apache 2 license](https://github.com/nats-io/nats-server/blob/main/LICENSE)
Resolves #

Changes proposed in this pull request:
- in processPingTimer, add explicit test for ROUTER connection, and only hold off sending ping based on receive ping

### DIFF
--- a/server/client.go
+++ b/server/client.go
@@ -253,7 +253,9 @@ type client struct {
 	ping       pinfo
 	msgb       [msgScratchSize]byte
 	last       time.Time
-	headers    bool
+	lastIn     time.Time
+
+	headers bool
 
 	rtt      time.Duration
 	rttStart time.Time
@@ -286,9 +288,8 @@ type rrTracking struct {
 
 // Struct for PING initiation from the server.
 type pinfo struct {
-	tmr  *time.Timer
-	last time.Time
-	out  int
+	tmr *time.Timer
+	out int
 }
 
 // outbound holds pending data for a socket.
@@ -1307,8 +1308,10 @@ func (c *client) readLoop(pre []byte) {
 		c.mu.Lock()
 
 		// Activity based on interest changes or data/msgs.
+		// Also update last receive activity for ping sender
 		if c.in.msgs > 0 || c.in.subs > 0 {
 			c.last = last
+			c.lastIn = last
 		}
 
 		if n >= cap(b) {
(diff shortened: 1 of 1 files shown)


server/client.go
```diff
@@ -4564,16 +4564,13 @@ func (c *client) processPingTimer() {
...
 	} else {
-		// If we have had activity within the PingInterval then
-		// there is no need to send a ping. This can be client data
-		// or if we received a ping from the other side.
-		if delta := now.Sub(c.last); delta < pingInterval && !needRTT {
-			c.Debugf("Delaying PING due to client activity %v ago", delta.Round(time.Second))
-		} else if delta := now.Sub(c.ping.last); delta < pingInterval && !needRTT {
+		// If we received a ping from the other side within the PingInterval then
+		// there is no need to send a ping.
+		if delta := now.Sub(c.ping.last); delta < pingInterval && !needRTT {
```
derekcollison, line 4573: This feels like it should be any inbound activity not just a ping.

sandykellagher:
I've made a change so that the c.ping.last timestamp is updated for receive client data as well as pings, and squashed this commit with the previous one. I've also marked the other PR as closed.

server/client.go
derekcollison: So other inbound data from a client connection will not suppress the pings here, that is the intended behavior correct?

sandykellagher: Well no. As per your suggestion on the other PR, in this update the c.ping.last timestamp is used to record both received pings and client data. There is another change to the end of the readLoop routine where c.ping.last is updated.

derekcollison: ok maybe rename then to `in`, so c.in.last or c.inbound.last?

derekcollison: I think we are close, and we are prepping to release 2.9.16. 

If we can get this merged later today or no later then tomorrow it will make it in.

sandykellagher: OK.  Not 100% sure this is what you meant, but I have changed the ping check to work with a `c.lastIn` timestamp, which is updated on receiving client data or a ping, and deleted the `c.ping.last` field.

derekcollison (approved):
LGTM - Thanks!

PR #4017, opened 2023-04-03
Possibly de-race `sysRequest`

neilalexander:
This may fix a race condition in `sysRequest` where multiple inbox responses could try to mutate the same input object, so instead we'll let it instantiate its own, although it isn't clear to me yet why that would happen in the first place.

### DIFF
--- a/server/jetstream_cluster.go
+++ b/server/jetstream_cluster.go
@@ -5573,19 +5573,20 @@ var (
 
 // blocking utility call to perform requests on the system account
 // returns (synchronized) v or error
-func (s *Server) sysRequest(v interface{}, subjFormat string, args ...interface{}) (interface{}, error) {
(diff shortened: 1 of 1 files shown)


server/jetstream_cluster.go
```diff
@@ -5573,48 +5573,40 @@ var (
...
+		if err := json.Unmarshal(msg, &v); err != nil {
 			s.Warnf("Error unmarshalling response for request '%s':%v", isubj, err)
 			return
 		}
-		select {
-		case results <- v:
-		default:
-			s.Warnf("Failed placing request response on internal channel")
-		}
+		results <- &v
```
derekcollison, line 5588: If this was running and got blocked for whatever reason this go routine would never exit.

derekcollison (approved):
LGTM

PR #4020, opened 2023-04-03
[IMPROVED] Raft layer improvements

derekcollison:
Two changes here.

1. Revert of a change regarding handling of leadership transfer entries that caused instability in some tests and a bad upgrade experience.

2. Make sure if we detect that a stream or a consumer is already running the monitoring routines that we do not stop the underlying raft group since it will be being used in the monitor routine and would stall that asset.


wallyqs (approved):
LGTM 💯

neilalexander (approved):
LGTM!

neilalexander (approved):
LGTM!

Jarema (approved):
LGTM!

PR #4022, opened 2023-04-04
[IMPROVED] Make sure to never process next message requests inline.

derekcollison:

### DIFF
--- a/server/consumer.go
+++ b/server/consumer.go
@@ -2860,7 +2860,7 @@ func newNextMsgReq(reply string, msg []byte) *nextMsgReq {
 	} else {
 		nmr = &nextMsgReq{}
 	}
-	// When getting something from a pool it is criticical that all fields are
+	// When getting something from a pool it is critical that all fields are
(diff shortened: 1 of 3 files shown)


server/consumer.go
```diff
@@ -2833,6 +2844,37 @@ func (o *consumer) nextWaiting(sz int) *waitingRequest {
...
+
+func newNextMsgReq(reply string, msg []byte) *nextMsgReq {
+	var nmr *nextMsgReq
+	m := nextMsgReqPool.Get()
+	if m != nil {
+		nmr = m.(*nextMsgReq)
+	} else {
+		nmr = &nextMsgReq{}
+	}
+	// When getting something from a pool it is criticical that all fields are
```
wallyqs, line 2863:
```suggestion
	// When getting something from a pool it is critical that all fields are
```

derekcollison: Fixed, and fixed where I copied it from ;)

wallyqs (approved):
LGTM

PR #4026, opened 2023-04-04
[IMPROVED] Do not block routes, leafnodes or gateways on received internal system msgs.

derekcollison:
To increase system stability in large NATS systems we do not want to process inbound system messages that are destined to be processed by the server itself to be processed inline on the inbound connection.


wallyqs (approved):
LGTM!

PR #4028, opened 2023-04-06
[IMPROVED] Optimize non-inline direct gets to not use simple go routines.

derekcollison:
This was causing regression in certain KV performance tests.


neilalexander (approved):
LGTM!

PR #4029, opened 2023-04-06
Move stepdowns of streams and consumers to not be inline.

derekcollison:


neilalexander (approved):
LGTM

PR #4031, opened 2023-04-06
[IMPROVED] Finer grained locking for /healthz with large number of JetStream assets.

derekcollison:
For checking the health of jetstream, do not hold the lock as we traverse the streams and consumers.


neilalexander (approved):
LGTM

PR #4032, opened 2023-04-06
[FIXED] Make sure on reverse match to compensate for wider target subjects.

derekcollison:
If we have a wider subject, meaning more tokens, but our subs end with a wildcard token make sure that matches properly.


wallyqs (approved):
LGTM

PR #4036, opened 2023-04-09
[IMPROVED] Memory allocations and pressure in the raft layer.

derekcollison:
Improved some hot spots on the memory subsystem with sync.Pools.

### DIFF
--- a/server/filestore.go
+++ b/server/filestore.go
@@ -4490,24 +4490,21 @@ func (mb *msgBlock) subjString(skey []byte) string {
 	if len(skey) == 0 {
 		return _EMPTY_
 	}
-	key := string(skey)
-
 	if lsubjs := len(mb.fs.cfg.Subjects); lsubjs > 0 {
 		if lsubjs == 1 {
 			// The cast for the comparison does not make a copy
-			if key == mb.fs.cfg.Subjects[0] {
+			if string(skey) == mb.fs.cfg.Subjects[0] {
 				return mb.fs.cfg.Subjects[0]
 			}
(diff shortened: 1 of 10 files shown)


server/filestore.go
```diff
@@ -4483,31 +4483,50 @@ func (mb *msgBlock) msgFromBuf(buf []byte, sm *StoreMsg, hh hash.Hash64) (*Store
 	return sm, nil
 }
 
+// Used to intern strings for subjects.
+// Based on idea from https://github.com/josharian/intern/blob/master/intern.go
```
neilalexander, line 4487: Well this is a pretty clever approach!

server/raft.go
```diff
@@ -1989,16 +2055,25 @@ func (ar *appendEntryResponse) encode(b []byte) []byte {
 	return buf[:appendEntryResponseLen]
 }
 
+// Track all peers we may have ever seen to use an string interns for appendEntryResponse decoding.
+var peers sync.Map
```
neilalexander, line 2059: It looks like nothing ever gets removed from this `sync.Map`?

neilalexander (approved):
LGTM overall

server/raft.go
derekcollison: For peers, yes, even on a large system like NGS with peer replacements would be like 40 entries.

derekcollison: We could add a remove in there for peer remove entry..

derekcollison: Added.

PR #4045, opened 2023-04-13
[FIXES/IMPROVED] Raft and filestore logic 

derekcollison:
When failing to send a snapshot to a follower, in addition to stepping down reset our state.
For a filestore when expiring messages on recovery make sure that the first message is not a deleted message.
Fix a bad unlock of wrong type which could cause a crash during shutdown.
Add in raft specific test framework.


neilalexander (approved):
LGTM!

PR #4046, opened 2023-04-13
Limit filestore block size to 2MB when encryption is enabled

neilalexander:
Some write operations in the filestore require re-encrypting the entire block, so having large blocks can make these operations slow and hurt performance.


derekcollison (approved):
LGTM - Small tweak to comment if possible before merge.

PR #4048, opened 2023-04-13
[FIXED] #4043

### LINKED ISSUE #4043: Extreme JetStream write performance degredation when subject counts are limited
## Defect

Writes to JetStream seem to slow down drastically when there is a constrained number of subjects _and_ there is a size/count/ttl limit applied.

For example, with a stream limited to 100k messages and writing 1m messages

| test | MsgsPerSec |
|------|------------|
| single subject | 52414 |
| all unique subjects | 30191 |
| 16 subjects | 37574 |
| 256 subjects | 8984 |
| 1024 subjects: |2417 |

Tested with 2.9.6 and 2.10.0-beta.30 (via nats cli)

* OS: Linux Mint 21.1
* Kernel: 6.2.7-x64v3-xanmod1
* CPU: i7-8565U

```
$ export NATS_CONTEXT=nats_development
$ rm -rf ~/.local/share/nats/nats_development/jetstream
$ nats server run --jetstream --verbose
$ nats stream add --defaults --max-msgs=100000 --subjects 'foo.>' --max-bytes=1GB TEST
```

```
$ nats bench --pub=1 --js foo.bar --stream TEST --msgs=1000000 --size=200 --pubbatch=1024 --csv /dev/stdout --multisubject
```

#### Expected result:

Running with different `--multisubjectmax` values should result in similar performace

#### Actual result:

```csv
RunID,ClientID,MsgCount,MsgBytes,MsgsPerSec,BytesPerSec,DurationSecs
single subject,P0,2000000,200000000,52414,10482974.154080,19.078555
--multisubjectmax=0,P0,2000000,200000000,30191,6038270.572961,33.122067
--multisubjectmax=16,JoCx5tctLCR9yhX1yC7zzt,P0,2000000,200000000,37574,7514918.051133,26.613730
--multisubjectmax=256,P0,2000000,200000000,8984,1796898.424830,111.302897
--multisubjectmax=1024,P0,2000000,200000000,2417,483465.501359,413.679982
```

derekcollison:
Improve performance on storing msgs when multiple subjects exists with multiple messages and we have store limits that are being hit.

This will not fix any performance degradation per se when per subject limits are being hit constantly across a vast array of keys. That will be addressed more so in 2.10.

Resolves #4043 

### DIFF
--- a/server/filestore.go
+++ b/server/filestore.go
@@ -1358,7 +1358,7 @@ func (fs *fileStore) expireMsgsOnRecover() {
 			}
 			// Update fss
 			// Make sure we have fss loaded.
-			mb.removeSeqPerSubject(sm.subj, seq, nil)
+			mb.removeSeqPerSubject(sm.subj, seq)
 			fs.removePerSubject(sm.subj)
 		}
 		// Make sure we have a proper next first sequence.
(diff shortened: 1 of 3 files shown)


wallyqs (approved):
LGTM

PR #4056, opened 2023-04-14
[FIXED] Do not spin on multi-leader with same term.

derekcollison:
If we see another leader with same term we should step down, otherwise we could spin with catchups and WAL truncates.


wallyqs (approved):
LGTM! 🚀

PR #4057, opened 2023-04-15
[FIXED] Fix for a regression in behavior.

derekcollison:
When tracking information per subject went from >1 back to only 1 we needed to make sure we cleared firstNeedsUpdate.

Thanks to @scottf for finding it.


Jarema (approved):
LGTM!

PR #4058, opened 2023-04-15
[IMPROVED] Recover on boundary conditions that could cause RAFT layer to spin.

derekcollison:
Reset our WAL on edge conditions instead of trying to recover.
Also if we are timing out and trying to become a candidate but are doing a catchup check if we are stalled.


wallyqs (approved):
LGTM

neilalexander (approved):
LGTM

PR #4072, opened 2023-04-18
Apply stream compression from file store config

neilalexander:
This PR fills the missing link between the stream config and the actual compression state.

Related: nats-io/jsm.go#445, nats-io/natscli#762

### DIFF
--- a/server/filestore.go
+++ b/server/filestore.go
@@ -107,6 +107,35 @@ func (alg StoreCompression) String() string {
 	}
 }
 
+func (alg StoreCompression) MarshalJSON() ([]byte, error) {
+	var str string
+	switch alg {
+	case S2Compression:
+		str = "s2"
+	case NoCompression:
+		str = "none"
+	default:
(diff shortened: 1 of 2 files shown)


server/filestore.go
```diff
@@ -107,6 +107,35 @@ func (alg StoreCompression) String() string {
 	}
 }
 
+func (alg StoreCompression) MarshalJSON() ([]byte, error) {
+	var str string
+	switch alg {
+	case S2Compression:
+		str = "s2"
```
derekcollison, line 114: I suggest S2, I think that is how they refer to it.

derekcollison (approved):
LGTM

server/filestore.go
neilalexander: I stuck with the existing pattern from the JS schema here, which is that the values are all lower-case, although there is a separate `String()` function in the `jsm.go` PR that is for pretty-printing in the CLI (`"S2 compression"`) so a CLI user never sees it.

PR #4083, opened 2023-04-21
Stream mirror sources filter

### LINKED ISSUE #4082: Support filter subject in mirror and sources for streams
## Feature Request

It is possible to setup consumers with `filter_subject` in order to have granular authorizations over `$JS.*.API.CONSUMER.CREATE`. However, the same is not possible when using stream sources and mirrors.

#### Use Case:

With this change it is possible to create streams and KV stores that only mirror/source a subset of the original one. Very useful for authorization purposes, very useful for the bandwidth.

#### Proposed Change:

The `FilterSubject` is already parsed and passed around, it is just a matter of putting it in the creation of the consumer.

Everyone that has a "central" stream/kv store that needs to be replicated to leaf nodes with high granularity.

#### Alternative Approaches

N/A

CC @paolobarbolini

dodomorandi:
 - [ ] Tests added

Resolves #4082 

 - Forward `Mirror.FilterSubject` to `CreateConsumerRequest` inside `setupMirrorConsumer`
 - Forward `ssi.FilterSubject` to `CreateConsumerRequest` inside `setSourceConsumer`

I splitted the commits in two just because they can be seen as two differente `feat`s. If you would rather keep into a single commit, I can squash them without problems.

CC @paolobarbolini

### DIFF
--- a/server/stream.go
+++ b/server/stream.go
@@ -2535,6 +2535,7 @@ func (mset *stream) setSourceConsumer(iname string, seq uint64, startTime time.T
 	// Determine subjects etc.
 	var deliverSubject string
 	ext := ssi.External
+	filter := ssi.FilterSubject
 
 	if ext != nil && ext.DeliverPrefix != _EMPTY_ {
 		deliverSubject = strings.ReplaceAll(ext.DeliverPrefix+syncSubject(".S"), "..", ".")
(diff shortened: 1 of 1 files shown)


jnmoyne:
Hum.... unless I'm missing something here this is already been done a few lines later

e.g. https://github.com/M4SS-Code/nats-server/blob/e96ae0bf79a2a3737b00376636cd1d04194c244c/server/stream.go#L2277-L2280

derekcollison:
Looks like it, this line.

https://github.com/M4SS-Code/nats-server/blob/e96ae0bf79a2a3737b00376636cd1d04194c244c/server/stream.go#L2307

dodomorandi:
As I wrote in the related issue, I was definitely wrong -- sorry guys :pray:

PR #4084, opened 2023-04-21
Outbound queue improvements

neilalexander:
This extends the previous work in #3733 with the following:

1. Remove buffer coalescing, as this could result in a race condition during the `writev` syscall in rare circumstances
2. Add a third buffer size, to ensure that we aren't allocating more than we need to without coalescing
3. Refactor buffer handling in the WebSocket code to reduce allocations and ensure owned buffers aren't incorrectly being pooled resulting in further race conditions

Fixes nats-io/nats.ws#194.

### DIFF
--- a/server/client.go
+++ b/server/client.go
@@ -302,12 +302,20 @@ type outbound struct {
 	stc chan struct{} // Stall chan we create to slow down producers on overrun, e.g. fan-in.
 }
 
-const nbPoolSizeSmall = 4096  // Underlying array size of small buffer
+const nbPoolSizeSmall = 512   // Underlying array size of small buffer
+const nbPoolSizeMedium = 4096 // Underlying array size of medium buffer
 const nbPoolSizeLarge = 65536 // Underlying array size of large buffer
 
 var nbPoolSmall = &sync.Pool{
 	New: func() any {
 		b := [nbPoolSizeSmall]byte{}
+		return &b
+	},
+}
+
+var nbPoolMedium = &sync.Pool{
+	New: func() any {
+		b := [nbPoolSizeMedium]byte{}
 		return &b
 	},
 }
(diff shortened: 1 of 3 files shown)


server/client.go
```diff
@@ -312,18 +313,44 @@ var nbPoolSmall = &sync.Pool{
...
+	return new
+}
+
 func nbPoolPut(b []byte) {
 	switch cap(b) {
 	case nbPoolSizeSmall:
 		b := (*[nbPoolSizeSmall]byte)(b[0:nbPoolSizeSmall])
 		nbPoolSmall.Put(b)
+	case nbPoolSizeMedium:
+		b := (*[nbPoolSizeMedium]byte)(b[0:nbPoolSizeMedium])
```
derekcollison, line 352: Why do we have these?

derekcollison (commented):
Have we tested against the setup from yesterday?

server/client.go
neilalexander: Without the coalescing, we have to allocate more buffers when traffic is in flight, so I added an extra pool size so that there's middle-ground between small and large messages.

delaneyj:
Testing now

delaneyj:
Previous race issue in go and web socket parse/connection issues not found in testing live app! :shipit:   

PS)  Browser app is now MUCH faster in full 2.9.16

server/client.go
derekcollison: I meant the casting before placing back into the pool. We cast when we pull it out, seems we might not need?

neilalexander: Ah, the code was doing this before too, the reason is that `sync.Pool` is most efficient when the contents are pointerised, otherwise it creates lots of copies of the slice header on the heap and they get re-boxed passing through the `interface{}` boundary.

derekcollison (approved):
LGTM!

PR #4086, opened 2023-04-21
Run a check for ack floor drift.

derekcollison:
Also periodically check. If all normal will be very cheap.

### DIFF
--- a/server/consumer.go
+++ b/server/consumer.go
@@ -3237,12 +3237,81 @@ func (o *consumer) hbTimer() (time.Duration, *time.Timer) {
 	return o.cfg.Heartbeat, time.NewTimer(o.cfg.Heartbeat)
 }
 
+// Check here for conditions when our ack floor may have drifted below the streams first sequence.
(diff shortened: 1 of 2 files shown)


server/consumer.go
```diff
@@ -3237,13 +3237,82 @@ func (o *consumer) hbTimer() (time.Duration, *time.Timer) {
...
+			o.processTerm(seq, p.Sequence, rdc)
+		}
+	}
+
+	// Do one final check here.
+	o.mu.Lock()
+	defer o.mu.Unlock()
+
+	// If we are here, and this should be rare, we still are off with our ack floor.
+	// We will set it explcitly to 1 behind our current lowest in pending, or if
```
neilalexander, line 3282: Explicitly

neilalexander (approved):
LGTM

wallyqs (approved):
LGTM

server/consumer.go
derekcollison: Thanks will fix.

derekcollison: Thanks will fix.

PR #4090, opened 2023-04-22
fix formatting of raft debug log

wallyqs:
Fixes this entry showing up:

```
[36571] 2023/04/22 06:53:37.278744 [DBG] RAFT [2DfzK1X6 - C-R3F-bzb7ShZ0] Stepping down from %!s(uint64=66), detected higher term: 65 vs %!d(string=leader)
```


derekcollison (approved):
LGTM - Thanks!

PR #4093, opened 2023-04-24
Restore outbound queue coalescing

neilalexander:
This PR effectively reverts part of #4084 which removed the coalescing from the outbound queues as I initially thought it was the source of a race condition.

Further investigation has proven that not only was that untrue (the race actually came from the WebSocket code, all coalescing operations happen under the client lock) but removing the coalescing also worsens performance.

### DIFF
--- a/server/client.go
+++ b/server/client.go
@@ -328,19 +328,14 @@ var nbPoolLarge = &sync.Pool{
 }
 
 func nbPoolGet(sz int) []byte {
-	var new []byte
 	switch {
 	case sz <= nbPoolSizeSmall:
-		ptr := nbPoolSmall.Get().(*[nbPoolSizeSmall]byte)
-		new = ptr[:0]
+		return nbPoolSmall.Get().(*[nbPoolSizeSmall]byte)[:0]
 	case sz <= nbPoolSizeMedium:
-		ptr := nbPoolMedium.Get().(*[nbPoolSizeMedium]byte)
-		new = ptr[:0]
+		return nbPoolMedium.Get().(*[nbPoolSizeMedium]byte)[:0]
 	default:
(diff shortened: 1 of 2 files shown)


server/client.go
```diff
@@ -302,8 +302,7 @@ type outbound struct {
 	stc chan struct{} // Stall chan we create to slow down producers on overrun, e.g. fan-in.
 }
 
-const nbPoolSizeSmall = 512   // Underlying array size of small buffer
```
derekcollison, line 305: Still think this might have been a good idea.

derekcollison: Could you test with small payloads, 32, 64 or 128 bytes and watch RSS?

neilalexander: Interestingly, on NATS Core publishes, there seems to be virtually no difference in how two vs three buffers manifest, but in JetStream publishes, some of the benchmarks show better results. I've added the small buffer back as a result.

server/client.go
```diff
@@ -1447,7 +1442,8 @@ func (c *client) flushOutbound() bool {
 	// "nb" will be reset back to its starting position so it can be modified
 	// safely by queueOutbound calls.
 	c.out.wnb = append(c.out.wnb, collapsed...)
-	orig := append(net.Buffers{}, c.out.wnb...)
+	var _orig [1024][]byte
+	orig := append(_orig[:0], c.out.wnb...)
```
derekcollison, line 1446: Where is this used?

neilalexander: L1479-1485 to return consumed buffers to the pool. The `net.Buffers.WriteTo()` function modifies the receiver slice by trimming sent buffers off the beginning, so we need to keep a copy of those references to recycle them.

derekcollison (approved):
LGTM!

Issue #4094, opened 2023-04-24
Allow custom stream sources and mirrors heartbeat


dodomorandi:
## Feature Request

It would be interesting to allow user-defined heartbeats for JetStream sources and mirrors.

#### Use Case:

Heartbeats are pretty cheap, but they introduce a small overhead. If in normal cases this overhead is totally meaningless, in scenarios in which bandwidth is precious, the default health checks performed once per second is not negligible anymore.

If my estimation is correct, the bandwidth related to health checks at the moment is ~20MiB/day, which means ~600MiB per month. If you just imagine using NATS on embedded devices using SIM mobile network connection, this amount of data is _somehow expensive_.

Giving to the user the possibility of customizing the heartbeat could make them in the position to sacrifice some reliability in order to gain in terms of bandwidth. Given that JetStream are extremely robust with data re-transmission, this seems a valuable approach.

#### Proposed Change:

Add a `HeartBeat` field to `StreamSource`, deserialized from the JSON `idle_heartbeat` (proposed for consistency with `ConsumerConfig` and `JSApiConsumerGetNextRequest`, otherwise it could be just `heartbeat`) and use this value instead of the default `sourceHealthCheckInterval`.

Two utility functions can be used in order to easily get the value from the mirror or the source, respectively, when specified, using `sourceHealthCheckInterval` as fallback.

The non trivial downside for this is that all the bindings need to be changed in order to make things work flawlessly from all the supported languages.

#### Who Benefits From The Change(s)?

All people that have strict bandwidth requirements. In theory this approach could also be used to _increment_ the frequency of the heartbeat, but in this case I cannot think of a real use-case.

#### Alternative Approaches

A different approach could be simply to use just one generic value that can be configured globally. This is a questionable approach, because at that point you need to use the same configuration for every node in the network. Or, at least, you need that the difference between all the heartbeats is less than `3*heartbeat`, otherwise one node consider another _slow_ node as stalled (correct me if I misunderstood how things work).

---

I am going to open a PR for this, with the changes related to `nats-server`. If you think that this change is interesting, I can add the checks for the other repositories (feel free to suggest me some high-priority repos) and start helping with the changes.

CC @paolobarbolini

PR #4095, opened 2023-04-24
Allow custom StreamSource heartbeat

dodomorandi:
 - [ ] Tests added

Main issue: #4094  

- Added `Heartbeat` field to `StreamSource`
- Added `getMirrorHeartbeat` helper function
- Added `getStreamSourceHeartbeat` and `getSourceHeartbeat` helper functions
- Use `getMirrorHeartbeat` to define mirror heartbeat and related health checks
- Use `getStreamSourceHeartbeat` and `getSourceHeartbeat` to define stream sources heartbeat and related health checks

CC @paolobarbolini

### DIFF
--- a/server/stream.go
+++ b/server/stream.go
@@ -189,6 +189,7 @@ type StreamSource struct {
 	SubjectTransformDest string                   `json:"subject_transform_dest,omitempty"`
 	SubjectTransforms    []SubjectTransformConfig `json:"subject_transforms,omitempty"`
 	External             *ExternalStream          `json:"external,omitempty"`
+	Heartbeat            time.Duration            `json:"idle_heartbeat,omitempty"`
 
 	// Internal
 	iname string // For indexing when stream names are the same for multiple sources.
@@ -2081,7 +2082,7 @@ func (mset *stream) processMirrorMsgs(mirror *sourceInfo, ready *sync.WaitGroup)
 	// Signal the caller that we have captured the above fields.
 	ready.Done()
 
-	t := time.NewTicker(sourceHealthCheckInterval)
+	t := time.NewTicker(mset.getMirrorHeartbeat())
 	defer t.Stop()
 
 	for {
@@ -2103,7 +2104,7 @@ func (mset *stream) processMirrorMsgs(mirror *sourceInfo, ready *sync.WaitGroup)
 		case <-t.C:
 			mset.mu.RLock()
 			isLeader := mset.isLeader()
-			stalled := mset.mirror != nil && time.Since(mset.mirror.last) > 3*sourceHealthCheckInterval
+			stalled := mset.mirror != nil && time.Since(mset.mirror.last) > 3*mset.getMirrorHeartbeat()
 			mset.mu.RUnlock()
 			// No longer leader.
 			if !isLeader {
@@ -2117,6 +2118,14 @@ func (mset *stream) processMirrorMsgs(mirror *sourceInfo, ready *sync.WaitGroup)
 				mset.retryMirrorConsumer()
 			}
 		}
+	}
+}
+
+func (mset *stream) getMirrorHeartbeat() time.Duration {
+	if mset.cfg.Mirror.Heartbeat == 0 {
+		return sourceHealthCheckInterval
+	} else {
+		return mset.cfg.Mirror.Heartbeat
 	}
 }
 
@@ -2426,7 +2435,7 @@ func (mset *stream) setupMirrorConsumer() error {
 			AckPolicy:      AckNone,
 			AckWait:        22 * time.Hour,
 			MaxDeliver:     1,
-			Heartbeat:      sourceHealthCheckInterval,
+			Heartbeat:      mset.getMirrorHeartbeat(),
 			FlowControl:    true,
 			Direct:         true,
 		},
@@ -2761,7 +2770,7 @@ func (mset *stream) setSourceConsumer(iname string, seq uint64, startTime time.T
 			AckPolicy:      AckNone,
 			AckWait:        22 * time.Hour,
 			MaxDeliver:     1,
-			Heartbeat:      sourceHealthCheckInterval,
+			Heartbeat:      ssi.getStreamSourceHeartbeat(),
 			FlowControl:    true,
 			Direct:         true,
 		},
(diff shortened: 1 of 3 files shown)


dodomorandi:
Except for the failing CI (for which TBH I am a bit perplexed), feel free to give me some feedback about the value of the PR (and the relative issue). If you think it is basically fine, I will add some tests for the new behavior.

dodomorandi:
@derekcollison can you give me a hint about a possible cause? The reason I don't understand is that, without specifying any `idle_heartbeat` parameter, the behavior should have not been changed in any possible way. What I am not seeing?

derekcollison:
The stack for the datarace will give you the info needed, in terms of when it violates and what was previous operation, etc..

dodomorandi:
@derekcollison Thanks for the patient! There was just a single point of data race it slipped through. Tests are now green.

Let me know if the change is reasonable for you, in that case I can add a few tests.

derekcollison (commented):
Looks reasonable, may need to be targeted at dev branch for 2.10 vs main which is 2.9.x atm.

Let's add some tests etc.

Thanks!

dodomorandi:
@derekcollison Sorry to bother, but I am banging my head against the tests and I probably need a hint from someone who knows the codebase.

The ideal test would be to use both (relatively) small and large custom heartbeat and monitor the behavior of the consumers in order to check everything is working as expected. For instance, I should expect a failing test if I change back the heartbeat value to the previously default one in the checks for stalled streams.

The issue I am facing is I am unable to verify this situation. For instance, the test `TestJetStreamPushConsumerIdleHeartbeats` exactly performs this task, but it has direct access to the testing consumer (because it just creates it) in order to subscribe to the replies and check for the heartbeat. On the other hand, when a stream is configured to use a mirror or some sources, the created consumer is not exposed in any way, and I am unable to get the heartbeat messages _just subscribing_ to `>`.

I honestly do not know if I am missing something obvious or the situation is really a bit tricky, in any case I think people with the knowledge of the codebase could give me very useful information. Even if my changes have a small impact, I think that there are a couple of sharp edges that really need some additional tests.

Thank you in advance!

derekcollison:
Yes can be tricky in there, the consumer is gettable, but a bit nuanced.

Can you point me to a line in your test where you would need direct access and I can probably help.

dodomorandi:
Sure, and thank you for helping me! :blush:

I just pushed a [_wip_ commit](https://github.com/nats-io/nats-server/pull/4095/commits/776fd9939dcfc40452ca4820ace0d3e91a7257da), which is pretty much stolen from [TestJetStreamPushConsumerIdleHeartbeats](https://github.com/nats-io/nats-server/blob/eb1eb3c49e835280eea52c61c3bad38e4717a343/server/jetstream_test.go#L7283).

As you can see [I am trying](https://github.com/nats-io/nats-server/blob/776fd9939dcfc40452ca4820ace0d3e91a7257da/server/jetstream_test.go#L20864) to [replicate](https://github.com/nats-io/nats-server/blob/eb1eb3c49e835280eea52c61c3bad38e4717a343/server/jetstream_test.go#L7303) the subscription to the consumer in order to [get](https://github.com/nats-io/nats-server/blob/776fd9939dcfc40452ca4820ace0d3e91a7257da/server/jetstream_test.go#L20870-L20877) the heartbeat messages.

However (and pretty obviously), in the original test the subject is used as a delivery subject for the consumer. However, in my case the subject of the consumer [is changed a bit](https://github.com/nats-io/nats-server/blob/eb1eb3c49e835280eea52c61c3bad38e4717a343/server/stream.go#L2195) and [other parameters are set](https://github.com/nats-io/nats-server/blob/eb1eb3c49e835280eea52c61c3bad38e4717a343/server/stream.go#L2209-L2217) -- and my _noobness_ with the codebase does not let me grasp what is the specific reason I am unable to see the the heartbeat messages from the _inbox subscription_ :sweat_smile:.

I would be really happy to solve this without bothering you, but I think I need some help :disappointed:. Thank you in advance for all the support!

derekcollison:
We have a customer that is looking at this functionality so we will introduce for 2.10 in some for or fashion.

server/stream.go
```diff
@@ -1909,6 +1910,14 @@ func (mset *stream) processMirrorMsgs(mirror *sourceInfo, ready *sync.WaitGroup)
 	}
 }
 
+func (mset *stream) getMirrorHeartbeat() time.Duration {
+	if mset.cfg.Mirror.Heartbeat == 0 {
```
bruth, line 1914: @dodomorandi My guess is that the data race is in this method since no read lock is being used (unlike all of the other methods on `mset` when fields are accessed).

bruth: Actually.. most places the lock is already held except for here: https://github.com/nats-io/nats-server/pull/4095/files#diff-2f4991438bb868a8587303cde9107f83127e88ad70bd19d5c6a31c238a20c299R1874

dodomorandi: The data race now is fixed, the only thing is that I keep using the initial value of the heartbeat. AFAIK this should not be an issue because I don't expect the value to change across locking blocks. Is my assumption correct or it would be better to use the _current_ value of the heartbeat [here](https://github.com/nats-io/nats-server/pull/4095/files#diff-2f4991438bb868a8587303cde9107f83127e88ad70bd19d5c6a31c238a20c299L2742)?

bruth:
Looks like the final bit to remove is the local `replace` in go.mod?

dodomorandi:
@bruth Sorry for the delay. I just resynced the repo. Regarding to your last comment, I think that this PR needs a change in `nats.go` to correctly support the new `Heartbeat` field before being merged, and the change is pretty simple [as you can see from here](https://github.com/nats-io/nats.go/compare/main...M4SS-Code:nats.go:stream-source-heartbeat).

If you want I can open a PR in `nats.go` in order to add that field and link to this PR, just let me know.

paolobarbolini:
The approach of #4352 looks interesting. Since this didn't make it in 2.10 we're going to patch nats-server in our Yocto build and experiment with whether this option makes sense to be configured globally or at a per stream-by-stream level.

maurobalbi:
> We have a customer that is looking at this functionality so we will introduce for 2.10 in some for or fashion.

Has this been introduced?

PR #4097, opened 2023-04-24
[ADDED] Checking HEALTHZ for specific accounts/stream

piotrpio:
This PR adds `account` and `stream` options to `HEALTHZ` system request and `/healthz` monitoring endpoint.
It allows for checking health of a specific stream, without having to rely on other streams, which (when under stress) may not have reached consensus yet and would return an error.

/cc @wallyqs

### DIFF
--- a/server/monitor.go
+++ b/server/monitor.go
@@ -2647,9 +2647,11 @@ type JSzOptions struct {
 // HealthzOptions are options passed to Healthz
 type HealthzOptions struct {
 	// Deprecated: Use JSEnabledOnly instead
-	JSEnabled     bool `json:"js-enabled,omitempty"`
-	JSEnabledOnly bool `json:"js-enabled-only,omitempty"`
-	JSServerOnly  bool `json:"js-server-only,omitempty"`
+	JSEnabled     bool   `json:"js-enabled,omitempty"`
+	JSEnabledOnly bool   `json:"js-enabled-only,omitempty"`
+	JSServerOnly  bool   `json:"js-server-only,omitempty"`
+	Account       string `json:"account,omitempty"`
+	Stream        string `json:"stream,omitempty"`
 }
 
 // ProfilezOptions are options passed to Profilez
@@ -3038,6 +3040,8 @@ func (s *Server) HandleHealthz(w http.ResponseWriter, r *http.Request) {
 		JSEnabled:     jsEnabled,
 		JSEnabledOnly: jsEnabledOnly,
 		JSServerOnly:  jsServerOnly,
+		Account:       r.URL.Query().Get("account"),
+		Stream:        r.URL.Query().Get("stream"),
 	})
 	if hs.Error != _EMPTY_ {
 		s.Warnf("Healthcheck failed: %q", hs.Error)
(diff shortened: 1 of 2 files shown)


server/monitor.go
```diff
@@ -2647,9 +2647,11 @@ type JSzOptions struct {
...
 type HealthzOptions struct {
 	// Deprecated: Use JSEnabledOnly instead
-	JSEnabled     bool `json:"js-enabled,omitempty"`
-	JSEnabledOnly bool `json:"js-enabled-only,omitempty"`
-	JSServerOnly  bool `json:"js-server-only,omitempty"`
+	JSEnabled     bool   `json:"js-enabled,omitempty"`
+	JSEnabledOnly bool   `json:"js-enabled-only,omitempty"`
+	JSServerOnly  bool   `json:"js-server-only,omitempty"`
+	Account       string `json:"account,omitempty"`
+	Stream        string `json:"stream,omitempty"`
```
wallyqs, line 2654: could it be possible to add Consumer as well as an option?

piotrpio: Done

server/monitor.go
```diff
@@ -3114,10 +3190,51 @@ func (s *Server) healthz(opts *HealthzOptions) *HealthStatus {
...
+		if !details {
+			health.Status = "error"
+			health.Error = fmt.Sprintf("Bad request: %q must not be empty when checking stream health", "account")
+		} else {
+			health.Errors = append(health.Errors, HealthzError{
+				Type:  HealthzErrorBadRequest,
+				Error: fmt.Sprintf("Bad request: %q must not be empty when checking stream health", "account"),
+			})
+		}
+		return health
```
wallyqs, line 3212: maybe add a Code field to the health result here? It would be nice to get an http 400 for these and a 404 for when the streams/consumers are not found for example

server/monitor.go
```diff
@@ -3089,10 +3156,19 @@ func (s *Server) HandleHealthz(w http.ResponseWriter, r *http.Request) {
...
+	}
+
 	hs := s.healthz(&HealthzOptions{
 		JSEnabled:     jsEnabled,
 		JSEnabledOnly: jsEnabledOnly,
 		JSServerOnly:  jsServerOnly,
+		Account:       r.URL.Query().Get("account"),
+		Stream:        r.URL.Query().Get("stream"),
+		Consumer:      r.URL.Query().Get("consumer"),
+		Details:       includeDetails,
```
wallyqs, line 3171: thinking whether we should narrow down results to streams and consumers details as an option, but this option to gather all the results in one pass is useful.

server/monitor.go
wallyqs: or could base on the healthzError.Type to return a custom http code

piotrpio: I'm not sure about the custom error code as this will work fine for http-based `HealthZ` only, and not for `$SYS.REQ.SERVER.%s.HEALTHZ`, unless we add a custom header. I will add a `StatusCode` field in response.

neilalexander (approved):
LGTM but please do a final rebase as quite a few things have merged since yesterday.

Nice job on the thorough tests!

PR #4101, opened 2023-04-25
Test that makes sure that assets can change be scaled after a cluster change.

derekcollison:
This is specifically when a cluster is reconfigured and the servers are restarted with a new cluster name.


neilalexander (approved):
LGTM

PR #4107, opened 2023-04-26
[IMPROVED] Clustered consumer improvements

derekcollison:
Consumer state from Jsz() would not be consistent for a leader vs follower.

ConsumerFileStore could encode an empty state or update an empty state on startup.
We needed to make sure at the lowest level that the state was read from disk and not depend on the upper layer consumer.


wallyqs (approved):
LGTM

PR #4115, opened 2023-04-28
[ADDED] Support for route S2 compression

kozlovic:
The new field `compression` in the `cluster{}` block allows to
specify which compression mode to use between servers.

It can be simply specified as a boolean or a string for the
simple modes, or as an object for the "s2_auto" mode where
a list of RTT thresholds can be specified.

By default, if no compression field is specified, the server
will default to "accept", which means that a server will accept
compression from a remote and switch to that same compression
mode, but will otherwise not initiate compression. That is,
if 2 servers are configured with "accept", then compression
will actually be "off". If one of the server had say s2_fast
then they would both use this mode.

Here is the way to specify compression with a simple string:
```
cluster {
..
  # Possible values are "disabled", "off", "enabled", "on",
  # "accept", "s2_fast", "s2_better", "s2_best" or "s2_auto"
  compression: s2_fast
}
```
If the compression field is simply set to "s2_auto", then
the server will use default RTT thresholds of 10ms, 50ms 
and 100ms for the "uncompressed", "fast", "better"
and "best" modes.

To specify a different list of thresholds for the s2_auto,
here is how it would look like:
```
cluster {
..
  compression: {
    mode: s2_auto
    # This means that for RTT up to 5ms (included), then
    # the compression level will be "uncompressed", then
    # from 5ms+ to 15ms, the mode will switch to "s2_fast",
    # then from 15ms+ to 50ms, the level will switch to
    # "s2_better", and anything above 50ms will result
    # in the "s2_best" compression mode.
    rtt_thresholds: [5ms, 15ms, 50ms]
  }
}
```

If a server has compression mode set (other than "off") but
connects to an older server, there will be no compression between
those 2 routes.


server/accounts.go
```diff
@@ -1338,19 +1338,19 @@ func (a *Account) sendTrackingLatency(si *serviceImport, responder *client) bool
 	// FIXME(dlc) - We need to clean these up but this should happen
 	// already with the auto-expire logic.
 	if responder != nil && responder.kind != CLIENT {
-		a.mu.Lock()
+		si.acc.mu.Lock()
```
derekcollison, line 1341: Should this be cherry picked to main?

server/client.go
```diff
@@ -408,8 +411,12 @@ const (
 type readCacheFlag uint16
 
 const (
-	hasMappings readCacheFlag = 1 << iota // For account subject mappings.
-	sysGroup                  = "_sys_"
+	hasMappings         readCacheFlag = 1 << iota // For account subject mappings.
+	switchToCompression readCacheFlag = 1 << 1
+)
+
+const (
```
derekcollison, line 418: Just one line for now since singelton?

server/events.go
```diff
@@ -686,11 +686,13 @@ func routeStat(r *client) *RouteStat {
 		return nil
 	}
 	r.mu.Lock()
+	// Note: *client.out[Msgs|Bytes] are not set using atomics,
+	// unlike in[Msgs|Bytes].
 	rs := &RouteStat{
 		ID: r.cid,
 		Sent: DataStats{
-			Msgs:  atomic.LoadInt64(&r.outMsgs),
```
derekcollison, line 692: Don
t need atomic access anymore?

server/monitor.go
```diff
@@ -123,6 +123,7 @@ type ConnInfo struct {
 	OutMsgs        int64          `json:"out_msgs"`
 	InBytes        int64          `json:"in_bytes"`
 	OutBytes       int64          `json:"out_bytes"`
+	SentBytes      int64          `json:"sent_bytes"`
```
derekcollison, line 126: Should we not have ReceivedBytes as well if other side sending us compressed?
I do not like long winded but should it be SentWireBytes?

server/monitor.go
```diff
@@ -2095,18 +2101,19 @@ type LeafzOptions struct {
...
+	IsSpoke   bool     `json:"is_spoke"`
+	Account   string   `json:"account"`
+	IP        string   `json:"ip"`
+	Port      int      `json:"port"`
+	RTT       string   `json:"rtt,omitempty"`
+	InMsgs    int64    `json:"in_msgs"`
+	OutMsgs   int64    `json:"out_msgs"`
+	InBytes   int64    `json:"in_bytes"`
+	OutBytes  int64    `json:"out_bytes"`
+	SentBytes int64    `json:"sent_bytes"`
```
derekcollison, line 2114: We doing leaf compression in this PR too?

server/reload.go
```diff
@@ -379,10 +381,38 @@ func (c *clusterOption) Apply(s *Server) {
 		s.routeInfo.WSConnectURLs = s.websocket.connectURLs
 	}
 	s.setRouteInfoHostPortAndIP()
+	var routes []*client
+	if c.compChanged {
```
derekcollison, line 385: Maybe compressChanged

server/server.go
```diff
@@ -325,9 +327,223 @@ type stats struct {
...
+	CompressionS2Auto         = "s2_auto"
+	CompressionS2Uncompressed = "s2_uncompressed"
+	CompressionS2Fast         = "s2_fast"
+	CompressionS2Better       = "s2_better"
+	CompressionS2Best         = "s2_best"
+)
+
+// defaultCompressionS2AutoRTTThresholds is the default of RTT thresholds for
+// the CompressionS2Auto mode.
+var defaultCompressionS2AutoRTTThresholds = []time.Duration{
```
derekcollison, line 353: If I set best on my side, what does other side do?

Reason I am asking is for asymmetrical routes, like home internet where downstream and upstream is different. So I would like to set mine to best but have other side sending to me with say fast.

server/accounts.go
kozlovic: It is already like that in main. I had it changed to `a.mu.Lock` in the "route pool and per account" PR because I got a race. But since then, you have fixed what was actually causing the race. See comment: https://github.com/nats-io/nats-server/pull/4115/commits/8d2683a062c56343b4de91d04027c1f168cb1197

server/client.go
kozlovic: Will do

server/events.go
kozlovic: Never did for `client.stats.outMsgs`. See comment I added few lines above:
```
// Note: *client.out[Msgs|Bytes] are not set using atomics,
// unlike in[Msgs|Bytes].
```

server/monitor.go
kozlovic: The problem with the receive side is that we would only get the bytes read *after* decompression (since we have a s2 reader that wraps the net.Conn).
As for `SentBytes`, I originally used `WriteBytes`, but decided on `SentBytes`. I can certainly change to `SentWireBytes`. Please let me know and I will make the change.

server/monitor.go
kozlovic: No, but since the count is on a `*client` it would be available to all type of connections, that's what I added it to all. And again, this value is simply the number of bytes wrote to the TCP connection, regardless if compression is used or not. It is different from OutBytes in that it contains ANY data written, so for messages that includes the MSG protocol and any overhead, etc..

server/reload.go
kozlovic: Will do

server/server.go
kozlovic: Compression levels are asymmetric, meaning that you can have S1 chose "best" and the other side "fast". If one side has "accept", then it will use the same than given by the other side, but otherwise, each side can select the one they want. So in short, yes, it works as you would want/described above.

server/events.go
derekcollison: Ah ok.

server/monitor.go
kozlovic: (edited, I wrote `WriteBytes` when I meant that I originally used `WireBytes`)

derekcollison: I would remove for now, we need better and more symmetric reporting long term but not here and I want us to think a bit more about it.

server/monitor.go
kozlovic: That being said, happy to remove from all objects except RouteInfo for the moment...

derekcollison: ok so without compression you still think this could be valuable?

derekcollison: I think remove for now from this PR and give some more thought.

server/monitor.go
kozlovic: will do

server/monitor.go
kozlovic: ok, will do.

derekcollison (approved):
LGTM

derekcollison (approved):
LGTM

PR #4116, opened 2023-04-29
[IMPROVED] Health repair

derekcollison:
Under certain scenarios we have witnessed healthz() that will never return healthy due to a stream or consumer being missing or stopped. This will now allow the healthz() call to attempt to restart those assets.

We will also periodically call this in clustered mode from the monitorCluster routine.


wallyqs (approved):
LGTM

PR #4117, opened 2023-05-01
[IMPROVED] Updating of a large fleet of leafnodes.

derekcollison:
When a fleet of leafnodes are isolated (not routed but using same cluster) we could do better at optimizing how we update the other leafnodes since if they are all in the same cluster and we know we are isolated we can skip.

We can improve further in 2.10.


wallyqs (approved):
LGTM

PR #4118, opened 2023-05-01
[FIXED] A stream raft node could stay running after a stop().

derekcollison:
This can happen when we reset a stream internally and the stream had a prior snapshot.
Also make sure to always release resources back to the account regardless if the store is no longer present.


Jarema (approved):
LGTM

wallyqs (approved):
LGTM

neilalexander (approved):

wallyqs (approved):
LGTM

PR #4119, opened 2023-05-01
[ADDED] LeafNode: TLSHandshakeFirst option

kozlovic:
A new field in `tls{}` blocks force the server to do TLS handshake before sending the INFO protocol.
```
leafnodes {
   port: 7422
   tls {
      cert_file: ...
      ...
      handshake_first: true
   }
   remotes [
       {
         url: tls://host:7423
         tls {
            ...
            handshake_first: true
         }
       }
   ]
}
```
Note that if `handshake_first` is set in the "accept" side, the first `tls{}` block in the example above, a server trying to create a LeafNode connection to this server would need to have `handshake_first` set to true inside the `tls{}` block of the corresponding remote.

Configuration reload of leafnodes is generally not supported, but TLS certificates can be reloaded and the support for this new field was also added.

### DIFF
--- a/server/client.go
+++ b/server/client.go
@@ -1752,7 +1752,11 @@ func (c *client) markConnAsClosed(reason ClosedState) {
 	// we use Noticef on create, so use that too for delete.
 	if c.srv != nil {
 		if c.kind == LEAF {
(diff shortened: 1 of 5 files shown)


derekcollison (approved):
LGTM

PR #4120, opened 2023-05-02
Tweak outbound queue memory test to only log if growth exceeds 10%

neilalexander:
This makes the `TestNoRaceClientOutboundQueueMemory` test useful and stops unnecessary logging.


derekcollison (approved):
LGTM

PR #4123, opened 2023-05-02
[IMPROVED] Do not hold filestore lock during remove that needs to do IO.

derekcollison:
When removing a msg and we need to load the msg block and incur IO, unlock fs lock to avoid stalling other activity on other blocks. E.g removing and adding msgs at the same time.


neilalexander (approved):
LGTM

PR #4124, opened 2023-05-02
Don't allow compression on memory storage

neilalexander:
Compression only works on file storage, so let's be explicit about that rather than having an ineffective configuration option when memory storage is selected.


derekcollison (commented):
I think we should allow compression on memory streams, just will be per message vs per block as we have it now in filestore.

neilalexander:
No longer required as of #4186.

PR #4130, opened 2023-05-03
[FIXED] Subscription interest issue due to configuration reload

kozlovic:
This would impact only cases with accounts defined in configuration file (as opposed to operator mode). During the configuration reload, new accounts and sublists were created to later be replaced with existing ones. That left a window of time where a subscription could have been added (or attempted to be removed) from the "wrong" sublist. This could lead to route subscriptions seemingly not being forwarded.


server/route.go
```diff
@@ -1772,17 +1772,18 @@ func (s *Server) startRouteAcceptLoop() {
 
 // Similar to setInfoHostPortAndGenerateJSON, but for routeInfo.
 func (s *Server) setRouteInfoHostPortAndIP() error {
-	if s.opts.Cluster.Advertise != "" {
-		advHost, advPort, err := parseHostPort(s.opts.Cluster.Advertise, s.opts.Cluster.Port)
+	opts := s.getOpts()
+	if opts.Cluster.Advertise != "" {
```
derekcollison, line 1776: _EMPTY_

server/server.go
```diff
@@ -739,40 +738,85 @@ func (s *Server) globalAccount() *Account {
...
+	awcsti := make(map[string]struct{})
+
 	// Create the global account.
 	if s.gacc == nil {
 		s.gacc = NewAccount(globalAccountName)
 		s.registerAccountNoLock(s.gacc)
 	}
 
-	opts := s.opts
+	opts := s.getOpts()
```
derekcollison, line 754: We use lock to get, but don't copy. So when we do s.getOpts().Port lets say and its a copy of a value that seems good, but the above does not offer any protection really does it or am I missing something?

server/server.go
```diff
@@ -817,16 +862,32 @@ func (s *Server) configureAccounts() error {
 		for _, si := range acc.imports.services {
 			if v, ok := s.accounts.Load(si.acc.Name); ok {
 				si.acc = v.(*Account)
+				// TODO: Not sure if it is possible for an account to have a
```
derekcollison, line 865: Its is possible now for latency tracking.

derekcollison (approved):
In general LGTM, a few small comments.

server/server.go
kozlovic: It is to avoid possible data races, because during a config reload, s.opts gets replaced with the new version. That is, doing s.opts.Port could be reported as a data race during the reload because it would be a read of s.opts where the reload does a write of s.opts. By using getOpts(), we get under lock the pointer to the "old" options, so that prevent the data race.

server/route.go
kozlovic: will replace

server/server.go
derekcollison: That is what I figured, but won't prevent any issues with actual values if they are being written/changed.

server/server.go
kozlovic: Yeah.. that because I cherry-picked the commit where this fix was done, and the change to this comment was done in the PR's last commit (base on code review). Will update.

server/server.go
kozlovic: It won't cause issues with actual values if they are being written/changed, but the check may be made on the "old" value. Again doing s.getOpts().Port is data-race safe, while s.opts.Port is not. For instance:
- thread 1 does `if s.opts.Port == 1234`
- thread 2 does a config reload set the new options: https://github.com/nats-io/nats-server/blob/main/server/reload.go#L907 (which result in https://github.com/nats-io/nats-server/blob/b61e411b44af43a125f89e549d0ac7337c88db89/server/server.go#L731)
This would result in the data race because of reading s.opts and writing s.opts, not because of accessing Port.

Imagine s.opts is currently a pointer to Options with address 0x123456789, during the config reload, a new separate Options object is created, say address 0x987654321. Once the reload is "complete", that's when s.opts is swapped from 0x123456789 to 0x987654321, under the s.optsMu mutex.

kozlovic: So if a thread N does s.getOpts() and got pointer 0x123456789, any field in that Options object are not going to change, the updates were made in the object 0x987654321. Hence my comment of check made possibly against "old" values.

derekcollison (approved):
LGTM! Thanks.

PR #4132, opened 2023-05-04
Stop using UTC for time in flushClients

wallyqs:
In #1943 it was adopted to use `UTC()` in some timestamps, but an unintended side effect from this is that it strips  the monotonic time (https://github.com/golang/go/commit/e5646b23dee5eb6f896a3eef45959aa130857988), so it can be prone to clock skews when subtracting time in other areas of the code.


derekcollison (approved):
LGTM

PR #4133, opened 2023-05-05
Record the stream and consumer info timestamps

ripienaar:
This records the server time when info for streams and consumers are created so that tools such as the nats cli can calculate time deltas for last ack, last delivered and so forth in the context of the server clock.

This will help aleviate problems with client devices experiencing clock jitter that can show up in user interfaces as negative seconds since last ack etc

### DIFF
--- a/server/consumer.go
+++ b/server/consumer.go
@@ -55,6 +55,8 @@ type ConsumerInfo struct {
 	NumPending     uint64          `json:"num_pending"`
 	Cluster        *ClusterInfo    `json:"cluster,omitempty"`
 	PushBound      bool            `json:"push_bound,omitempty"`
+	// TimeStamp indicates when the info was gathered
+	TimeStamp time.Time `json:"ts"`
 }
 
 type ConsumerConfig struct {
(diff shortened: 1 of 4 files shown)


server/consumer.go
```diff
@@ -55,6 +55,8 @@ type ConsumerInfo struct {
 	NumPending     uint64          `json:"num_pending"`
 	Cluster        *ClusterInfo    `json:"cluster,omitempty"`
 	PushBound      bool            `json:"push_bound,omitempty"`
+	// TimeStamp indicates when the info was gathered
+	TimeStamp time.Time `json:"time_stamp"`
```
wallyqs, line 59: We have `"first_ts"` and `"last_ts"` being passed around so would prefer shorter version too here:
```suggestion
	TimeStamp time.Time `json:"ts"`
```

wallyqs (approved):
LGTM

derekcollison (approved):
LGTM - Should we consider that when this is used elsewhere it is called Now() in the struct?

wallyqs:
`JSInfo` uses `Now` too, maybe standardize on using that for the struct name instead? https://github.com/nats-io/nats-server/blob/91607d845960c0df45e4942feb5ea3d859a730c1/server/monitor.go#L2684

ripienaar:
To me, `now` just seems wrong :) I think since we already use `ts` in the jetstream API as wally points out we should stick to that

ripienaar:
All monitor things using `now` is good, all api things using the same is alao good - better if all used the same of course, but we cant change the js api ones so better be consistent in the api?

derekcollison:
I am not hard set on either one, just brought it up to gather thoughts and opinions.

I think since we are putting it in Stream and Consumer info should match the precedent set in JetStream to date, which sounds like what we already have in this PR.

PR #4136, opened 2023-05-08
Bump GH Actions versions (Go; node12 deprecation)

philpennock:
Bump various GitHub Actions versions to move away from those using the deprecated node.js 12 runtime, and so remove the warnings.

For the coveralls action, I switched to just `@v2`, per their example docs and as we do for most actions, instead of major.minor.patch.

For Golang, update one 1.16 to 1.19, and quote the remaining unquoted instances, even though they had `.x` in there so had to parse as strings, because we're just switching to always-quoting for consistency.

### DIFF
--- a/.github/workflows/cov.yaml
+++ b/.github/workflows/cov.yaml
@@ -15,7 +15,7 @@ jobs:
 
     steps:
       - name: Checkout code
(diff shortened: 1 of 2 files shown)


derekcollison (approved):
LGTM

PR #4145, opened 2023-05-10
[IMPROVED] Leadership transfer 

derekcollison:
When doing leadership transfer stepdown as soon as we know we have sent the EntryLeaderTransfer entry.

Delaying could allow something to be sent from the old leader which would cause the new leader to bail on being a candidate even though it would have gotten all the votes.


wallyqs (approved):
LGTM

PR #4146, opened 2023-05-11
[FIXED] Can't scale up some older streams

derekcollison:
For some older R1 streams created by previous servers we could have no cluster for the stream assignment group which would prevent scale up with newer servers.

This will inherit cluster if detected as absent from the placement tags or client cluster designation.


wallyqs (approved):
LGTM

PR #4151, opened 2023-05-11
resolver: improve signaling for missing account lookups

wallyqs:
When using the account resolver and a JWT is not found, the client could often get an i/o timeout error due to not receiving a timely response before the account resolver fetch request times out.  In this PR, instead of waiting for the fetch request to timeout, a resolver without JWTs will notify as well that it could not find a matching JWT, waiting for a response from all servers ~~but with a delay to let resolver that do have the JWT to respond first~~.

Also included in this PR is some cleanup to the logs emitted by the resolver.

 - [X] Tests added

### DIFF
--- a/server/accounts.go
+++ b/server/accounts.go
@@ -4233,30 +4233,24 @@ func (s *Server) fetch(res AccountResolver, name string, timeout time.Duration) 
 	respC := make(chan []byte, 1)
 	accountLookupRequest := fmt.Sprintf(accLookupReqSubj, name)
 	s.mu.Lock()
-	// Resolver will wait for detected active servers to reply
-	// before serving an error in case there weren't any found.
-	var expectedServers int
 	if s.sys == nil || s.sys.replies == nil {
 		s.mu.Unlock()
 		return _EMPTY_, fmt.Errorf("eventing shut down")
 	}
-	if s.sys != nil && s.sys.servers != nil {
-		expectedServers = len(s.sys.servers)
-	}
+	// Resolver will wait for detected active servers to reply
+	// before serving an error in case there weren't any found.
+	expectedServers := len(s.sys.servers)
 	replySubj := s.newRespInbox()
 	replies := s.sys.replies
 
 	// Store our handler.
 	replies[replySubj] = func(sub *subscription, _ *client, _ *Account, subject, _ string, msg []byte) {
 		var clone []byte
-		var isEmpty bool
-		if len(msg) > 0 {
+		isEmpty := len(msg) == 0
+		if !isEmpty {
 			clone = make([]byte, len(msg))
 			copy(clone, msg)
-		} else {
-			isEmpty = true
 		}
-
 		s.mu.Lock()
 		defer s.mu.Unlock()
 		expectedServers--
--- a/server/accounts.go
+++ b/server/accounts.go
@@ -3991,7 +3991,6 @@ func (dr *DirAccResolver) Start(s *Server) error {
 	defer dr.Unlock()
 	dr.Server = s
 	dr.operator = opKeys
-	fetchTimeout := dr.fetchTimeout
 	dr.DirJWTStore.changed = func(pubKey string) {
 		if v, ok := s.accounts.Load(pubKey); ok {
 			if theJwt, err := dr.LoadAcc(pubKey); err != nil {
@@ -4082,9 +4081,7 @@ func (dr *DirAccResolver) Start(s *Server) error {
 		if theJWT, err := dr.DirJWTStore.LoadAcc(accName); err != nil {
 			if errors.Is(err, fs.ErrNotExist) {
 				s.Debugf("DirResolver - Could not find account %q", accName)
-				// Reply with empty response to signal absence of JWT to others,
-				// but not too fast to mitigate beating actual responses with content.
-				time.Sleep(fetchTimeout / 10)
+				// Reply with empty response to signal absence of JWT to others.
 				s.sendInternalMsgLocked(reply, _EMPTY_, nil, nil)
 			} else {
 				s.Errorf("DirResolver - Error looking up account %q: %v", accName, err)
@@ -4250,18 +4247,16 @@ func (s *Server) fetch(res AccountResolver, name string, timeout time.Duration) 
 	replies := s.sys.replies
 
 	// Store our handler.
-	replies[replySubj] = func(sub *subscription, _ *client, _ *Account, subject, reply string, msg []byte) {
+	replies[replySubj] = func(sub *subscription, _ *client, _ *Account, subject, _ string, msg []byte) {
 		clone := make([]byte, len(msg))
 		copy(clone, msg)
 
 		s.mu.Lock()
+		defer s.mu.Unlock()
 		expectedServers--
-		if len(msg) == 0 {
-			// Skip empty responses until getting all the available servers.
-			if expectedServers > 0 {
-				s.mu.Unlock()
-				return
-			}
+		// Skip empty responses until getting all the available servers.
+		if len(msg) == 0 && expectedServers > 0 {
+			return
 		}
 		// Use the first valid response if there is still interest or
 		// one of the empty responses to signal that it was not found.
@@ -4271,7 +4266,6 @@ func (s *Server) fetch(res AccountResolver, name string, timeout time.Duration) 
 			default:
 			}
 		}
-		s.mu.Unlock()
 	}
 	s.sendInternalMsg(accountLookupRequest, replySubj, nil, []byte{})
 	quit := s.quitCh
--- a/server/accounts.go
+++ b/server/accounts.go
@@ -4082,12 +4082,10 @@ func (dr *DirAccResolver) Start(s *Server) error {
 		if theJWT, err := dr.DirJWTStore.LoadAcc(accName); err != nil {
 			if errors.Is(err, fs.ErrNotExist) {
 				s.Debugf("DirResolver - Could not find account %q", accName)
-				go func() {
-					// Reply with empty response to signal absence of JWT to others,
-					// but not too fast to mitigate beating actual responses with content.
-					time.Sleep(fetchTimeout / 10)
-					s.sendInternalMsgLocked(reply, _EMPTY_, nil, nil)
-				}()
+				// Reply with empty response to signal absence of JWT to others,
+				// but not too fast to mitigate beating actual responses with content.
+				time.Sleep(fetchTimeout / 10)
+				s.sendInternalMsgLocked(reply, _EMPTY_, nil, nil)
 			} else {
 				s.Errorf("DirResolver - Error looking up account %q: %v", accName, err)
 			}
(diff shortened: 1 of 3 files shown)


server/accounts.go
```diff
@@ -3990,20 +3991,23 @@ func (dr *DirAccResolver) Start(s *Server) error {
 	defer dr.Unlock()
 	dr.Server = s
 	dr.operator = opKeys
+	fetchTimeout := dr.fetchTimeout
 	dr.DirJWTStore.changed = func(pubKey string) {
 		if v, ok := s.accounts.Load(pubKey); ok {
 			if theJwt, err := dr.LoadAcc(pubKey); err != nil {
-				s.Errorf("update got error on load: %v", err)
+				s.Errorf("RESOLVER - Update got error on load: %v", err)
```
derekcollison, line 3998: Do we have cases where we all caps things similar? Maybe DirResolver?

server/accounts.go
```diff
@@ -4074,31 +4078,42 @@ func (dr *DirAccResolver) Start(s *Server) error {
...
 		}
-		if theJWT, err := dr.DirJWTStore.LoadAcc(tk[accReqAccIndex]); err != nil {
-			s.Errorf("Merging resulted in error: %v", err)
+		accName := tk[accReqAccIndex]
+		if theJWT, err := dr.DirJWTStore.LoadAcc(accName); err != nil {
+			if errors.Is(err, fs.ErrNotExist) {
+				s.Debugf("RESOLVER - Could not find account %q", accName)
+				go func() {
+					// Reply with empty response to signal absence of JWT to others,
+					// but not too fast to mitigate beating actual responses with content.
```
derekcollison, line 4087: This feels brittle and time sensitive based on rtt between entities in the system.

I think if the resolver is in partial or cache mode we stay silent, but if in full mode we all should have the same accounts so we should error (return nil) right away.

server/accounts.go
wallyqs: I used all caps so that it would look similar to `SYSTEM` logs, but I will update to `DirResolver` instead which would be clearer

wallyqs: Updated to DirResolver to label these instead, logs now look like this: https://gist.github.com/wallyqs/9c968143014f2e5fb61b7d99ec0f5326

server/accounts.go
wallyqs: Updated to be 1/10th of the configured fetch timeout, with the default of 1.9s that we have set [here](https://github.com/nats-io/nats-server/blob/main/server/const.go#L210) this means that the resolvers in full mode without all state would back off for 190ms before sending signal that they do not have it.

server/accounts.go
```diff
@@ -4074,31 +4078,42 @@ func (dr *DirAccResolver) Start(s *Server) error {
 		if len(tk) != accLookupReqTokens {
 			return
 		}
-		if theJWT, err := dr.DirJWTStore.LoadAcc(tk[accReqAccIndex]); err != nil {
-			s.Errorf("Merging resulted in error: %v", err)
+		accName := tk[accReqAccIndex]
+		if theJWT, err := dr.DirJWTStore.LoadAcc(accName); err != nil {
+			if errors.Is(err, fs.ErrNotExist) {
+				s.Debugf("DirResolver - Could not find account %q", accName)
+				go func() {
```
derekcollison, line 4085: So wondering if this should be a concern. I know why you are putting it into a separate go routine because of the sleep, but this could cause of an explosion of Go routines yes?

Like if we did a tight loop of clients trying to connect with a bad account we would be spinning these in every server for every lookup request no?

derekcollison: Maybe what we do for this version of the server, is wait for all servers if we receive a nil response. This way we can send a nil response immediately but on the receiving side we see the nil so wait for more responses. We know how many other servers are in the system internally so we can know when we are done.

derekcollison (changes requested):

server/accounts.go
wallyqs: sounds good, I will explore this idea and update. I was concerned about blocking but would now prefer to be cautious with spinning to many go routines here.

derekcollison: Maybe do a quick debug.PrintStack() there too to see if we are rooted in a route or GW read..

If we are we can not block for sure, but this approach might work.

wallyqs: I think that approach will work much better and can get faster responses for the happy path when servers online, I updated with this new heuristic.

server/accounts.go
```diff
@@ -4223,20 +4236,38 @@ func (s *Server) fetch(res AccountResolver, name string, timeout time.Duration)
...
 	replySubj := s.newRespInbox()
 	replies := s.sys.replies
+
 	// Store our handler.
-	replies[replySubj] = func(sub *subscription, _ *client, _ *Account, subject, _ string, msg []byte) {
+	replies[replySubj] = func(sub *subscription, _ *client, _ *Account, subject, reply string, msg []byte) {
 		clone := make([]byte, len(msg))
 		copy(clone, msg)
+
 		s.mu.Lock()
```
derekcollison, line 4257: Could have defer unlock here based on flow, or could grab ok from replies, release lock then do select and placement on the channel.

server/accounts.go
```diff
@@ -4223,20 +4236,38 @@ func (s *Server) fetch(res AccountResolver, name string, timeout time.Duration)
...
+		expectedServers--
+		if len(msg) == 0 {
```
derekcollison, line 4259: Can combine.

```go
if len(msg) == 00 && expectedServers > 0 
```

derekcollison (commented):
Some comments, we are close.

derekcollison (commented):
Some comments, we are close.

server/accounts.go
wallyqs: thanks, updated

server/accounts.go
```diff
@@ -4074,31 +4078,40 @@ func (dr *DirAccResolver) Start(s *Server) error {
...
 		}
-		if theJWT, err := dr.DirJWTStore.LoadAcc(tk[accReqAccIndex]); err != nil {
-			s.Errorf("Merging resulted in error: %v", err)
+		accName := tk[accReqAccIndex]
+		if theJWT, err := dr.DirJWTStore.LoadAcc(accName); err != nil {
+			if errors.Is(err, fs.ErrNotExist) {
+				s.Debugf("DirResolver - Could not find account %q", accName)
+				// Reply with empty response to signal absence of JWT to others,
+				// but not too fast to mitigate beating actual responses with content.
+				time.Sleep(fetchTimeout / 10)
```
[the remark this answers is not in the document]
wallyqs: removed the sleep to reply

server/accounts.go
```diff
@@ -4223,24 +4233,39 @@ func (s *Server) fetch(res AccountResolver, name string, timeout time.Duration)
 	respC := make(chan []byte, 1)
 	accountLookupRequest := fmt.Sprintf(accLookupReqSubj, name)
 	s.mu.Lock()
+	// Resolver will wait for available routes + gateways to reply
+	// before serving an error in case there weren't any found.
+	expectedServers := len(s.routes)
+	if s.gateway != nil {
```
derekcollison, line 4239: I think this is always non nil IIRC.

server/accounts.go
```diff
@@ -4223,24 +4233,39 @@ func (s *Server) fetch(res AccountResolver, name string, timeout time.Duration)
 	respC := make(chan []byte, 1)
 	accountLookupRequest := fmt.Sprintf(accLookupReqSubj, name)
 	s.mu.Lock()
+	// Resolver will wait for available routes + gateways to reply
+	// before serving an error in case there weren't any found.
+	expectedServers := len(s.routes)
```
derekcollison, line 4238: The server maps all other servers in a different data struct when it received remote statsz updates etc.
Could use that since if this is an expanded supercluster with leafnodes sharing system accounts this will not account for that.

server/accounts.go
```diff
@@ -4223,24 +4233,39 @@ func (s *Server) fetch(res AccountResolver, name string, timeout time.Duration)
...
 	if s.sys == nil || s.sys.replies == nil {
 		s.mu.Unlock()
 		return _EMPTY_, fmt.Errorf("eventing shut down")
 	}
 	replySubj := s.newRespInbox()
 	replies := s.sys.replies
+
 	// Store our handler.
 	replies[replySubj] = func(sub *subscription, _ *client, _ *Account, subject, _ string, msg []byte) {
 		clone := make([]byte, len(msg))
```
derekcollison, line 4251: Should we check for the nil responses here?
```go
var clone []byte
if len(msg) > 0 {
}
```

server/accounts.go
kozlovic: To keep in mind when merging to `dev`, one then should use s.numRoutes() because it accounts for pooled routes, per-account, etc...

server/accounts.go
kozlovic: Yes, instead the test could be:
```
if s.gateway.enabled {
..
```
There is currently no API to get the number of remotes, you could add one, but otherwise you should lock:
```
    s.gateway.RLock()
    expectedServers += len(s.gateway.remotes)
    s.gateway.RUnlock()
```

derekcollison: Could use ``` s.sys.servers```

wallyqs: Switched to use `s.sys.servers` to use the detected active servers instead.

server/accounts.go
wallyqs: Using s.sys.servers instead for the check now.

server/accounts.go
wallyqs: Added this check to do the copy when not empty.

server/accounts.go
```diff
@@ -4233,29 +4233,35 @@ func (s *Server) fetch(res AccountResolver, name string, timeout time.Duration)
...
-	expectedServers := len(s.routes)
-	if s.gateway != nil {
-		expectedServers += len(s.gateway.remotes)
-	}
+	var expectedServers int
 	if s.sys == nil || s.sys.replies == nil {
 		s.mu.Unlock()
 		return _EMPTY_, fmt.Errorf("eventing shut down")
 	}
+	if s.sys != nil && s.sys.servers != nil {
```
derekcollison, line 4243: s.sys check right above, and since you use len(s.sys.servers) it can be nil.

So just need 

```go
expectedServers = len(s.sys.servers)
```

server/accounts.go
```diff
@@ -4233,29 +4233,35 @@ func (s *Server) fetch(res AccountResolver, name string, timeout time.Duration)
...
+	}
 	replySubj := s.newRespInbox()
 	replies := s.sys.replies
 
 	// Store our handler.
 	replies[replySubj] = func(sub *subscription, _ *client, _ *Account, subject, _ string, msg []byte) {
-		clone := make([]byte, len(msg))
-		copy(clone, msg)
+		var clone []byte
+		var isEmpty bool
```
derekcollison, line 4252: You can just set this directly..

```go
isEmpty := len(msg)
```

server/accounts.go
wallyqs: good point, just updated this too

derekcollison (approved):
LGTM

PR #4152, opened 2023-05-12
[ADDED] Ability to drop partial wildcard tokens in some subject transforms

jnmoyne:
- [X] Tests added

There is currently a blanket requirement that subject transforms destinations MUST use ALL of the partial wildcards defined in the source of the transform. This is because the subject transformed defined for imports must be 'reversible' and therefore the destination transform must use all of the partial wildcard tokens defined in the source of the transform.

This reversing of a transform is only used for transforms used by imports, where in any case it doesn't make sense to use any transform other than Wildcard.

This PR:
- relaxes this requirement to only apply when the transform is used by an import, adding the ability to drop a wildcard token in transforms other than as part of an import.
- Improves transform reverse to support both legacy style wildcards $X and the new transform function {{Wildcard(X)}}- Improves reversible transform checking to only allow the use of wildcards in the destination.

### DIFF
--- a/server/subject_transform.go
+++ b/server/subject_transform.go
@@ -320,7 +320,7 @@ func transformUntokenize(subject string) (string, []string) {
 	var nda []string
 
 	for _, token := range strings.Split(subject, tsep) {
-		if args := getMappingFunctionArgs(wildcardMappingFunctionRegEx, token); (len(token) > 1 && token[0] == '$' && token[1] >= '1' && token[1] <= '9') || (args != nil && len(args) == 1 && args[0] != _EMPTY_) {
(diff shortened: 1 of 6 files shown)


server/subject_transform.go
```diff
@@ -69,7 +69,9 @@ type SubjectTransformer interface {
 	TransformTokenizedSubject(tokens []string) string
 }
 
-func NewSubjectTransform(src, dest string) (*subjectTransform, error) {
+func NewSubjectTransform(src, dest string, strict bool) (*subjectTransform, error) {
```
derekcollison, line 72: I would recommend leaving original and adding in new call that the old one calls with default of false.

```go
func NewSubjectTransform(src, dest string) (*subjectTransform, error) {
  return NewSubjectTransformWithStrict(src, dest, false)
}
```

derekcollison (approved):
LGTM

PR #4154, opened 2023-05-12
Use monotonic time for measuring time internally

wallyqs:

### DIFF
--- a/server/accounts.go
+++ b/server/accounts.go
@@ -3502,7 +3502,7 @@ func (s *Server) updateAccountClaimsWithRefresh(a *Account, ac *jwt.AccountClaim
 		a.jsLimits = nil
 	}
 
-	a.updated = time.Now().UTC()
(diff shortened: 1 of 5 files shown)


derekcollison (commented):
I think it was @ripienaar who wanted all times in UTC?

If this is internal only I think we are good, but want to make sure if we surface any of these externally they are UTC.

ripienaar:
Yes APIs responding in UTC is best. This light even be considered a breaking change?

wallyqs:
All these should not be front facing, they are used for the internal timers link ping interval etc... which if you call UTC then switch to being in wall clock mode so prone to clock skews.

ripienaar:
Ah all right then it’s fine. Sorry was just going off Derek’s comment (on my phone, code review hard)

derekcollison (approved):
LGTM

PR #4156, opened 2023-05-12
[FIXED] Allow user filtering on connz for other user types like nkeys etc.

### LINKED ISSUE #4149: Connz options doesn't support filtering by NKEY
Currently `ConnzOptions` only supports filtering by username (via the `user`) field. It would be great if this field could also filter by NKey, or if we could have another field to filter by user nkey.

derekcollison:
 
Resolves #4149

### DIFF
--- a/server/client.go
+++ b/server/client.go
@@ -5323,20 +5323,28 @@ func (c *client) doTLSHandshake(typ string, solicit bool, url *url.URL, tlsConfi
 	return false, err
 }
 
+// getRAwAuthUserLock returns the raw auth user for the client.
+// Will acquire the client lock.
+func (c *client) getRawAuthUserLock() string {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	return c.getRawAuthUser()
+}
+
 // getRAwAuthUser returns the raw auth user for the client.
 // Lock should be held.
 func (c *client) getRawAuthUser() string {
 	switch {
-	case c.opts.Nkey != "":
+	case c.opts.Nkey != _EMPTY_:
 		return c.opts.Nkey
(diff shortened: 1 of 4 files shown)


server/client.go
```diff
@@ -5323,32 +5323,40 @@ func (c *client) doTLSHandshake(typ string, solicit bool, url *url.URL, tlsConfi
 	return false, err
 }
 
+// getRAwAuthUserLock returns the raw auth user for the client.
```
wallyqs, line 5326:
```suggestion
// getRawAuthUserLock returns the raw auth user for the client.
```

wallyqs (approved):
LGTM

wallyqs:
Looks like `TestAccountReqMonitoring` needs to be updated too:
```
    events_test.go:1283: require "\"num_connections\":1," to be contained in "{\"server\":{\"name\":\"NDYE5VRDHGSRCBKIWXWMKJYKERUP44GEARVZNCWPPAKHADPWEBJ46NUD\",\"host\":\"127.0.0.1\",\"id\":\"NDYE5VRDHGSRCBKIWXWMKJYKERUP44GEARVZNCWPPAKHADPWEBJ46NUD\",\"cluster\":\"abc\",\"ver\":\"2.9.17-beta.7\",\"seq\":23,\"jetstream\":false,\"time\":\"2023-05-12T22:02:55.563428Z\"},\"data\":null,\"error\":{\"code\":500,\"description\":\"filter by user or account only allowed with auth option\"}}"
```

server/client.go
derekcollison: Fixed in both spots, thanks.

PR #4157, opened 2023-05-13
[FIXED] Allow sorting by rtt for connz.

### LINKED ISSUE #4150: Allow sorting Connz by RTT
It would be nice to be able to sort `Connz` results by their RTT descending, as filtering on larger numbers of connections makes it hard to discover connections that have a high RTT to the server.

derekcollison:

Resolves #4150


wallyqs (approved):
LGTM

PR #4159, opened 2023-05-14
[FIXED] Track all remote servers in a NATS system with different domains.

derekcollison:


wallyqs (approved):
LGTM

PR #4161, opened 2023-05-14
Add workflow for stale issues

bruth:
This adds a workflow to mark issues and PRs stale after the configured period of time, followed by closing the issue/PR after a subsequent period of time if there was no additional activity.

The `debug-only` option is so currently, so even when merged, it will do a dry-run and not perform any actions. Once we inspect the initial logs of the effect of an initial run (impacting existing issues), we can adjust accordingly and then follow-up with making it active.

For the debug logs to be enabled, we do need to add a repository secret named `ACTIONS_STEP_DEBUG` with a value set to `true` per [this instruction](https://github.com/marketplace/actions/close-stale-issues#debugging).

### DIFF
--- a/.github/workflows/stale-issues.yaml
+++ b/.github/workflows/stale-issues.yaml
@@ -1,0 +1,17 @@
+name: stale-issues
+on:
+  schedule:
+    - cron: "30 1 * * *"
+
+jobs:
+  - name: stale-issues
+    stale:
+      runs-on: ubuntu-latest
+      steps:
+        - uses: actions/stale@v8
+          with:
+            debug-only: true # Set until the behavior is tuned.
+            days-before-stale: 28 # Mark stale after 4 weeks (28 days) of inactivity
+            days-before-close: 7 # Close 1 week after marked `Stale` and without follow-up activity
+            exempt-all-milestones: true # Any issue/PR within a milestone will be omitted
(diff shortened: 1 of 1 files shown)


bruth:
Per some feedback from @ripienaar, I want to note the intention is to add a bit of automation for the purpose identifying stale issues having two main causes in mind:

- Author has not followed up to response/question from the maintainers
- Maintainers have not addressed the issue/PR or a follow-up from the author or other commenter

The goal is to improve the responsiveness from both parties and reduce the number of issues/PRs that are no longer relevant or the author no longer has interest.

Given this goal, there are a couple other tweaks to this workflow that can be made:

- Mark issues as stale, but disable the auto-close threshold. This is still useful for maintainers to observe the stale issues over time and follow-up.
- Related, but more _active_, is for a automated comment to be left on the issue/PR to state that it has been marked stale. This is more _active_ since it will trigger a notification for those involved/watching the issue.

bruth:
Current status: Updated the PR to disable auto-close and extended the staleness window to 8 weeks. If an issue/PR is marked stale (label), any follow-up activity will auto-remove the label.

It is still in `debug-only` mode for the merge so we can observe the behavior in the logs before we turn it on.

bruth:
CI failures are flappers..

philpennock (approved):

derekcollison (approved):
LGTM

bruth:
Thank you 😄

PR #4163, opened 2023-05-15
monitor: change account detail info back to utc when served

wallyqs:
Follow up from #4154 this one has to be in UTC when served via monitoring endpoint.


derekcollison (approved):
LGTM

derekcollison (commented):
LGTM

derekcollison (approved):
LGTM

PR #4164, opened 2023-05-15
[FIXED] Protect against out of bounds access on usage updates.

derekcollison:


wallyqs (approved):
LGTM

PR #4167, opened 2023-05-15
[ADDED] LeafNode: Support for s2 compression

kozlovic:
This is similar to PR #4115 but for LeafNodes.
Compression mode can be set on both side (the accept and in remotes).
```
leafnodes {
   port: 7422
   compression: s2_best
   remotes [
       {
         url: "nats://host2:74222"
         compression: s2_better
       }
   ]
}
```
Possible modes are similar than for routes (described in PR #4115), except that when not defined we default to `s2_auto`.


server/client.go
```diff
@@ -4711,8 +4737,8 @@ func (c *client) processPingTimer() {
 	now := time.Now()
 	needRTT := c.rtt == 0 || now.Sub(c.rttStart) > DEFAULT_RTT_MEASUREMENT_INTERVAL
 
-	// Do not delay PINGs for ROUTER, GATEWAY or spoke LEAF connections.
-	if c.kind == ROUTER || c.kind == GATEWAY || c.isSpokeLeafNode() {
+	// Do not delay PINGs for ROUTER, GATEWAY or LEAF connections.
+	if c.kind == ROUTER || c.kind == GATEWAY || c.kind == LEAF {
```
derekcollison, line 4741: We may want endpoints to be able to control how much traffic is being generated by pings.

server/leafnode.go
```diff
@@ -1093,22 +1111,112 @@ func (s *Server) createLeafNode(conn net.Conn, rURL *url.URL, remote *leafNodeCf
...
+	if !tlsRequired {
+		return false, nil
+	}
+
+	// If TLS required, peform handshake.
+	// Get the URL that was used to connect to the remote server.
+	rURL := remote.getCurrentURL()
 
+	// Perform the client-side TLS handshake.
+	if resetTLSName, err := c.doTLSClientHandshake("leafnode", rURL, tlsConfig, tlsName, tlsTimeout, opts.LeafNode.TLSPinnedCerts); err != nil {
```
derekcollison, line 1130: Make "leafnode" a const somewhere?

derekcollison (approved):
Few comments but LGTM.

server/leafnode.go
kozlovic: ok will do...

kozlovic: done

server/client.go
kozlovic: Yes, but is there any actionable item for me on that PR? Should we revert the change related to PING interval for LeafNode (if not also for routes)? Or is it a note-to-self for the future?

derekcollison: Does this change add to potential traffic to a large number of idle leafnode connections?

kozlovic: Adds a PING (from both sides) every 30 seconds IF the user configured `ping_interval` is more than that. In other words, this change caps the PING interval to 30 seconds for leafnode (and did the same for routes in the previous PR). As I said, we could revert those changes and let the s2_auto tuning be done at the normal PING interval (specified by user or 2 minutes if not set).

derekcollison: Let's revert back to the 2m for now and make sure we skip on hub side if activity is present.

kozlovic: Ok, will do.

kozlovic: I assume we leave it for ROUTER or do we also revert that (from previous PR)?

kozlovic: Yes, but:
```
..and make sure we skip on hub side if activity is present
```
that means that when the hub side has s2_auto (which remember is the default), then compression level will NEVER be re-evaluated if we always suppress PINGs because there is activity present.

kozlovic: Replace "NEVER" by every hour since we have some code that make sure we still send a PING in that case.

derekcollison: ok sounds good, and yes leave routes as is, no need to revert anything there imo.

kozlovic: Done in commit: https://github.com/nats-io/nats-server/pull/4167/commits/cf474d6333bae1b2b6a2b08845c6851de762a457

derekcollison (approved):
LGTM

PR #4177, opened 2023-05-17
[IMPROVED] Reset logic for streams

derekcollison:
When we detect conditions to reset streams, make sure we properly clean up old NRG nodes etc.


wallyqs (approved):
LGTM

PR #4186, opened 2023-05-22
Memory store compression

neilalexander:
This PR adds compression support to the memory store. It is much simpler than that of the file store as it compresses on a per-message level rather than a per-block level, therefore it's expected that savings will be modest by comparison.

As the memory store was previously tracking the stream-level `max_bytes` accounting by using the size of the `msg` buffer, I've had to wrap the `StoreMsg` inside another struct which contains the original buffer size as well as the compression algorithm used. Otherwise the accounting goes wrong and the limits are not enforced correctly.

It also moves various functions and types for both file and memory store compression into their own file for clarity, and updated the memory store unit tests to include a compressed permutation, as we already did with the filestore tests.

### DIFF
--- a/server/memstore.go
+++ b/server/memstore.go
@@ -40,7 +40,12 @@ type memStore struct {
 
 type memStoreMsg struct {
 	*StoreMsg
-	origMsgSz uint64 // Size of the message prior to compression
+	origMsgSz []byte // Uvarint, size of the message prior to compression
+}
+
+func (m *memStoreMsg) OrigMsgSz() uint64 {
+	sz, _ := binary.Uvarint(m.origMsgSz)
+	return sz
 }
 
 func newMemStore(cfg *StreamConfig) (*memStore, error) {
@@ -142,7 +147,7 @@ func (ms *memStore) storeRawMsg(subj string, hdr, msg []byte, seq uint64, ts int
 				ms.recalculateFirstForSubj(subj, ss.First, ss)
 			}
 			sm, ok := ms.msgs[ss.First]
-			if !ok || memStoreMsgSize(sm.subj, sm.hdr, sm.origMsgSz) < uint64(len(msg)+len(hdr)) {
+			if !ok || memStoreMsgSize(sm.subj, sm.hdr, sm.OrigMsgSz()) < uint64(len(msg)+len(hdr)) {
 				return ErrMaxBytes
 			}
 		}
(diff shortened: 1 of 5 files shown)


neilalexander:
Test `TestFileStoreCompactLastPlusOne` seems to be flaking on `dev` too.

server/memstore.go
```diff
@@ -35,6 +35,12 @@ type memStore struct {
 	receivedAny bool
 }
 
+type memStoreMsg struct {
+	*StoreMsg
+	compression StoreCompression // The compression algorithm used
```
derekcollison, line 40: These can't change per message can they? Rather not introduce additional memory usage per message unless needed.

server/memstore.go
```diff
@@ -879,6 +905,14 @@ func (ms *memStore) LoadLastMsg(subject string, smp *StoreMsg) (*StoreMsg, error
 		smp = new(StoreMsg)
 	}
 	sm.copy(smp)
+
+	// Decompress the message if needed.
+	if c, err := ms.cfg.Compression.Decompress(smp.msg); err != nil {
```
derekcollison, line 910: We call into this function for every load call, even if not compressed?

neilalexander: If not compressing then the slice just gets returned unmodified, so it stops us needing extra checks at the call site.

server/memstore.go
neilalexander: It's possible to change the compression mode of the stream config after creation, hence why we need to know what the compression was set to at the time the message was published. You can't set it on a per-message level otherwise.

server/memstore.go
```diff
@@ -35,6 +35,12 @@ type memStore struct {
 	receivedAny bool
 }
 
+type memStoreMsg struct {
+	*StoreMsg
+	compression StoreCompression // The compression algorithm used
+	origMsgSz   uint64           // Size of the message prior to compression
```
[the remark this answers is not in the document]
neilalexander: Can do, although currently I don't think there's a use for that outside of the memory store.

neilalexander:
@derekcollison Do you mind having another glance over the comments above? Wondering if there's any specific changes you would like.

server/memstore.go
derekcollison: Should we disallow on memory streams? If they change it what happens?

server/memstore.go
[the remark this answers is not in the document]
derekcollison: We may want to since we might introduce per message compression for the last block for filestore.

derekcollison (commented):
Let's find a time this week to walk through together.

server/memstore.go
derekcollison: I think we remove, just one compression choice for now.

neilalexander:
So far no apparent need or use-case has emerged for this, so will park for now.

PR #4190, opened 2023-05-24
De-race filestore shutdown

neilalexander:
There was a race condition that could result in lock contention when shutting down the filestore, potentially causing some goroutines to deadlock. With any luck this will also stop `TestFileStoreCompactLastPlusOne` from flaking.


derekcollison (commented):
Closed in favor of a simpler approach in #4193

PR #4193, opened 2023-05-24
When in async mode make sure to not flush or create new files after close.

derekcollison:
Some async tests were showing artifacts being left after a message block was being closed.


neilalexander (approved):
LGTM

PR #4199, opened 2023-05-30
Per-subject limits for MQTT retained messages

neilalexander:
By publishing retained messages on their own subject, we can apply max messages per subject limits to them. Also include the domain name in the subject to make consistent with the other MQTT streams.

### DIFF
--- a/server/mqtt.go
+++ b/server/mqtt.go
@@ -2330,6 +2330,7 @@ func (as *mqttAccountSessionManager) transferRetainedToPerKeySubjectStream(log *
 				break
 			}
 			log.Warnf("    Unable to load retained message with sequence %d: %s", smsg.Sequence, err)
+			errors++
 			return
 		}
 		// Unmarshal the message so that we can obtain the subject name.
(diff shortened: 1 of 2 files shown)


kozlovic:
@neilalexander I made some changes because running a manual test I noticed that messages would not be received after the transfer. See comments on my commit. 
As a general rule and experience from the client library, updating the stream is something in the code is risky because moving from older/newer/older versions could "erase" some new settings on the stream since an older version of a server would have a stream config that does not contain fields from a newer release. I am not sure if it is an issue here, but again, something to keep in mind. Maybe in those cases (where we need to update configs), we may want to make that external - as a tool or part of nats CLI).

kozlovic (approved):
Overall, change was good but there was an issue with not refreshing the stream info after transfer. I am posting an LGTM here for your code, but please review my changes too.

neilalexander:
@kozlovic Thanks for the review and the fixes! I've taken a look over your commit and it makes sense and looks fine to me — I had missed the need to call `lookupStream` again in order to keep the stream state up-to-date after the migration. Will rebase and merge.

PR #4202, opened 2023-05-31
[IMPROVED] Subject transform validation and error reporting

jnmoyne:
- [X] Tests added

1: Improves error reporting for weighted mappings:
As it was, any error in a weighted mapping would return a very unhelpfull error message.

e.g. `nats-server: mappingtest.cfg:38:39: interface conversion: interface {} is []interface {}, not string`

This was because the line `err := &configErr{tk, fmt.Sprintf("Error adding mapping for %q to %q : %v", subj, v.(string), err)}` would panic on the `v.(string)` since in weighted mapping that interface{} is actually a map[string]interface{} (since there's can be more than one mapping in weighted mappings).

Now returns the actual error:

e.g. `nats-server: mappingtest.cfg:40:3: Error adding mapping for "bla" : invalid mapping destination: wildcard index out of range in {{wildcard(1)}}`

2: improves subject transform checking and catches if the destination is using a mapping function and there are no partial wildcards in the source.

### DIFF
--- a/server/subject_transform.go
+++ b/server/subject_transform.go
@@ -110,13 +110,13 @@ func NewSubjectTransformWithStrict(src, dest string, strict bool) (*subjectTrans
 
 			if strict {
 				if tranformType != NoTransform && tranformType != Wildcard {
-					return nil, &mappingDestinationErr{token, ErrorMappingDestinationFunctionNotSupportedForImport}
+					return nil, &mappingDestinationErr{token, ErrMappingDestinationNotSupportedForImport}
 				}
 			}
 
 			if npwcs == 0 {
 				if tranformType != NoTransform {
-					return nil, &mappingDestinationErr{token, ErrorMappingDestinationFunctionWildcardIndexOutOfRange}
+					return nil, &mappingDestinationErr{token, ErrMappingDestinationIndexOutOfRange}
 				}
 			}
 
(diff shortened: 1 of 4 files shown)


server/subject_transform.go
```diff
@@ -114,6 +114,12 @@ func NewSubjectTransformWithStrict(src, dest string, strict bool) (*subjectTrans
 				}
 			}
 
+			if npwcs == 0 {
+				if tranformType != NoTransform {
+					return nil, &mappingDestinationErr{token, ErrorMappingDestinationFunctionWildcardIndexOutOfRange}
```
derekcollison, line 119: Got to be a shorter version of this no?

`ErrorMappingDestinationFunctionWildcardIndexOutOfRange`

server/subject_transform.go
```diff
@@ -140,6 +146,18 @@ func NewSubjectTransformWithStrict(src, dest string, strict bool) (*subjectTrans
...
+	} else {
+		// no wildcards used in the source: check that no transform functions are used in the destination
+		for _, token := range dtokens {
+			tranformType, _, _, _, err := indexPlaceHolders(token)
+			if err != nil {
+				return nil, err
+			}
+
+			if tranformType != NoTransform {
+				return nil, &mappingDestinationErr{token, ErrorMappingDestinationFunctionWildcardIndexOutOfRange}
```
derekcollison, line 158: Same

derekcollison (approved):
In general LGTM sans the length of that error.

neilalexander (approved):

derekcollison (approved):
LGTM

PR #4204, opened 2023-06-02
Annotate CPU and goroutine profiles

neilalexander:
This allows the CPU and goroutine profiles to be annotated with information that allows us to break down load based on accounts, streams and consumers. We could probably add more labels elsewhere for other purposes too. It makes it easier to spot whether there are certain assets that are responsible for heavy CPU usage, i.e. snapshotting certain stream states.

### DIFF
--- a/server/jetstream_cluster.go
+++ b/server/jetstream_cluster.go
@@ -833,7 +833,13 @@ func (js *jetStream) setupMetaGroup() error {
 	atomic.StoreInt32(&js.clustered, 1)
 	c.registerWithAccount(sacc)
 
-	js.srv.startGoRoutine(js.monitorCluster)
+	js.srv.startGoRoutine(
+		js.monitorCluster,
+		pprofLabels{
+			"type":    "metaleader",
+			"account": sacc.Name,
+		},
+	)
 	return nil
 }
 
@@ -3315,7 +3321,14 @@ func (js *jetStream) processClusterUpdateStream(acc *Account, osa, sa *streamAss
 			}
 			mset.monitorWg.Add(1)
 			// Start monitoring..
-			s.startGoRoutine(func() { js.monitorStream(mset, sa, needsNode) })
+			s.startGoRoutine(
+				func() { js.monitorStream(mset, sa, needsNode) },
+				pprofLabels{
+					"type":    "stream",
+					"account": mset.accName(),
+					"stream":  mset.name(),
+				},
+			)
 		} else if numReplicas == 1 && alreadyRunning {
 			// We downgraded to R1. Make sure we cleanup the raft node and the stream monitor.
 			mset.removeNode()
@@ -3538,7 +3551,14 @@ func (js *jetStream) processClusterCreateStream(acc *Account, sa *streamAssignme
 			if mset != nil {
 				mset.monitorWg.Add(1)
 			}
-			s.startGoRoutine(func() { js.monitorStream(mset, sa, false) })
+			s.startGoRoutine(
+				func() { js.monitorStream(mset, sa, false) },
+				pprofLabels{
+					"type":    "stream",
+					"account": mset.accName(),
+					"stream":  mset.name(),
+				},
+			)
 		}
 	} else {
 		// Single replica stream, process manually here.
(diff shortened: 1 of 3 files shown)


derekcollison:
Lots of tests failures that should be adressed.

neilalexander:
Yep, PR is still marked as draft as I haven't looked at those failures yet, but I've gone back to the consumer defaults for now.

server/jetstream_cluster.go
```diff
@@ -833,7 +833,11 @@ func (js *jetStream) setupMetaGroup() error {
 	atomic.StoreInt32(&js.clustered, 1)
 	c.registerWithAccount(sacc)
 
-	js.srv.startGoRoutine(js.monitorCluster)
+	js.srv.startGoRoutine(
+		js.monitorCluster,
+		"type", "metaleader",
```
derekcollison, line 838: Love this idea, but not sure it belongs here. And maybe if it does just supply a single optional map instead?

server/opts.go
```diff
@@ -986,6 +987,8 @@ func (o *Options) processConfigFileLine(k string, v interface{}, errors *[]error
 		o.PortsFileDir = v.(string)
 	case "prof_port":
 		o.ProfPort = int(v.(int64))
+	case "prof_tags":
```
derekcollison, line 990: I don't think this is needed, if you turn on profiling we will add in the tags.

server/server.go
```diff
@@ -190,6 +191,7 @@ type Server struct {
 	grTmpClients map[uint64]*client
 	grRunning    bool
 	grWG         sync.WaitGroup // to wait on various go routines
+	grTags       bool
```
derekcollison, line 194: No need for this imo.

server/server.go
```diff
@@ -2050,6 +2052,7 @@ func (s *Server) Start() {
 
 	s.grMu.Lock()
 	s.grRunning = true
+	s.grTags = opts.ProfTags
```
derekcollison, line 2055: Can remove.

server/server.go
```diff
@@ -3501,15 +3504,22 @@ func (s *Server) String() string {
...
 
-func (s *Server) startGoRoutine(f func()) bool {
+func (s *Server) startGoRoutine(f func(), labels ...string) bool {
 	var started bool
 	s.grMu.Lock()
+	defer s.grMu.Unlock()
 	if s.grRunning {
 		s.grWG.Add(1)
-		go f()
+		go func(applyLabels bool) {
```
derekcollison, line 3513: Wonder if we make this a simple server call that can be called from any go routine as needed to setup labeling vs as part of startGoRoutine?

neilalexander: It can be done either way yep. Doing it in `startGoRoutine` just seemed like it had the added benefit of giving context to the reader at the point of invocation, almost self-documenting.

server/opts.go
derekcollison: I can see an argument for both. If someone was sharing with us let's say would want all the information so that we could tie together any NATS cli triage, e.g. `nats stream info` with any profiling they may share.

neilalexander:
This was modified to just include the tags in all cases and removed the configuration option.

server/jetstream_cluster.go
```diff
@@ -833,7 +833,13 @@ func (js *jetStream) setupMetaGroup() error {
 	atomic.StoreInt32(&js.clustered, 1)
 	c.registerWithAccount(sacc)
 
-	js.srv.startGoRoutine(js.monitorCluster)
+	js.srv.startGoRoutine(
+		js.monitorCluster,
+		map[string]string{
```
derekcollison, line 838: Hmm yea this seems worse. Maybe back to strings but concat the key and value?

```go
js.srv.startGoRoutine( js.monitorCluster, "type: metaleader", ..)
```

neilalexander: Only thing with this approach is that it might end up ugly for other reasons, i.e. `fmt.Sprintf` littered everywhere:
```
js.srv.startGoRoutine( js.monitorCluster, "type:metaleader", fmt.Sprintf("stream:%s", streamName), ...)
```

derekcollison: Hmm, true.

derekcollison (approved):
LGTM

PR #4206, opened 2023-06-02
Only check ack floor if we are interest policy based.

derekcollison:
Saw performance issue with a user a limits based stream with large number of consumers.


Jarema (approved):
LGTM

wallyqs (approved):
LGTM

PR #4207, opened 2023-06-02
[FIXED] Daisy chained leafnodes sometimes would not propagate interest

derekcollison:
When we were optimizing for single cluster and large numbers of leafnodes we inadvertently broke a daisy chained scenario where a server was a spoke and a hub within a single hub server.

So interest on D would not propagate properly to server A as a publisher.

```
     B
   /    \
A       C -- D (SUB)
 |
PUB
```


wallyqs (approved):
LGTM

PR #4216, opened 2023-06-05
[IMPROVED] When R1 consumers were recreated with the same name when they became inactive.

derekcollison:
When consumers were R1 and the same name was reused, server restarts could try to cleanup old ones and effect the new ones. These changes allow consumer name reuse more effectively during server restarts.


Jarema (approved):
LGTM!

PR #4226, opened 2023-06-09
[FIXED] Performance issues with checkAckFloor with large first stream sequence.

derekcollison:
Bail early if new consumer, meaning stream sequence floor is 0.
Decide which linear space to scan.
Do no work if no pending and we just need to adjust which we do at the end.

Also realized some tests were named wrong and were not being run, or were in wrong file.


wallyqs (approved):
LGTM

neilalexander (approved):
Looks good to me!

PR #4228, opened 2023-06-09
Don't hold entire MQTT retained messages in memory

neilalexander:
This PR separates out the small amount of necessary metadata for retained messages (stream sequence, floor) from the message itself, instead accessing the messages themselves with KV-like access patterns.

This should save quite a bit of memory where there are lots of retained messages since we only now need to hold a small amount of metadata instead of the entire messages.

### DIFF
--- a/server/mqtt.go
+++ b/server/mqtt.go
@@ -3274,7 +3274,7 @@ func (s *Server) mqttCheckPubRetainedPerms() {
 		// Get all of the retained messages. Then we will sort them so
 		// that they are in sequence order, which should help the file
 		// store to not have to load out-of-order blocks so often.
-		asm.mu.Lock()
+		asm.mu.RLock()
 		rms = rms[:0] // reuse slice
 		for subj, rf := range asm.retmsgs {
 			rms = append(rms, retainedMsg{
@@ -3282,7 +3282,7 @@ func (s *Server) mqttCheckPubRetainedPerms() {
 				rmsg: rf,
 			})
 		}
-		asm.mu.Unlock()
+		asm.mu.RUnlock()
 		sort.Slice(rms, func(i, j int) bool {
 			return rms[i].rmsg.sseq < rms[j].rmsg.sseq
 		})
@@ -3319,8 +3319,10 @@ func (s *Server) mqttCheckPubRetainedPerms() {
 			// Not present or permissions have changed such that the source can't
 			// publish on that subject anymore: remove it from the map.
 			if u == nil {
+				asm.mu.Lock()
 				delete(asm.retmsgs, rf.subj)
 				asm.sl.Remove(rf.rmsg.sub)
+				asm.mu.Unlock()
 				deletes[rf.subj] = rf.rmsg.sseq
 			}
 		}
--- a/server/mqtt.go
+++ b/server/mqtt.go
@@ -221,13 +221,13 @@ type mqttSessionManager struct {
 
 type mqttAccountSessionManager struct {
 	mu         sync.RWMutex
-	sessions   map[string]*mqttSession     // key is MQTT client ID
-	sessByHash map[string]*mqttSession     // key is MQTT client ID hash
-	sessLocked map[string]struct{}         // key is MQTT client ID and indicate that a session can not be taken by a new client at this time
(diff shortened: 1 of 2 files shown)


server/mqtt.go
```diff
@@ -1873,11 +1875,11 @@ func (as *mqttAccountSessionManager) sendJSAPIrequests(s *Server, c *client, acc
 // or 0 if the record was added instead of updated.
 //
 // Lock not held on entry.
-func (as *mqttAccountSessionManager) handleRetainedMsg(key string, rm *mqttRetainedMsg) uint64 {
+func (as *mqttAccountSessionManager) handleRetainedMsg(key string, rm *mqttRetainedMsgRef) uint64 {
```
kozlovic, line 1878: It looks like with this (and previous changes you made turning it into a KV mode), we don't need to return the "old" sequence number, since we don't need to delete the message anymore. So should we change the signature?

server/mqtt.go
```diff
@@ -3263,7 +3263,15 @@ func (s *Server) mqttCheckPubRetainedPerms() {
 		perms := map[string]*perm{}
 		deletes := map[string]uint64{}
 		asm.mu.Lock()
-		for subject, rm := range asm.retmsgs {
+		for subject, rf := range asm.retmsgs {
+			jsm, err := asm.jsa.loadMsg(mqttRetainedMsgsStreamName, rf.sseq)
```
kozlovic, line 3267: I am a bit concerned that we are going to load messages (possibly lot of them) under the sm and asm locks. Also, since retmsgs is a map, iteration may be causing lots of read out-of-order (sequence wise) and cause blocks to be loaded excessively.
So I propose that we do the process in several passes. Maybe collect the "asm" (sessions) under sm's lock. Then, iterate over those collected sessions and under asm's lock, collect the sequences in an array, release lock, then do an ascending sort of those sequences. For each sequence, do what you have here. For the asm.sl.Remove() down there, maybe capture the asm's sl under the lock (when collecting sequences).

If this is not clear at all, let me know and I can try to code it up.

kozlovic (changes requested):
Could change signature of one function since we don't seem to need the old sequence anymore. But more importantly, I wonder if the mqttCheckPubRetainedPerms() function should not be refactored a little bit knowing that we now have to read messages (possibly from disk, and possibly a lot of them).

server/mqtt.go
neilalexander: I think there's one case where we still do: when handling an empty MQTT message, which is "equivalent" to a delete?

server/mqtt.go
neilalexander: Good point on the locks. I can review it when I'm back on Tuesday!

server/mqtt.go
kozlovic: No, I searched for all references to this function in your branch and we no longer use the returned value. The one you are referring to is called `handleRetainedMsgDel()` (notice the `Del` at the end).

server/mqtt.go
neilalexander: Done — if you don't mind, please check this is what you had in mind.

> For each sequence, do what you have here. For the asm.sl.Remove() down there, maybe capture the asm's sl under the lock (when collecting sequences).

`Remove` takes the lock itself in this case, so there's no need for me to do so.

kozlovic (commented):
Will review more later, I can't at the moment but just wanted to post that note about asm.sl

kozlovic (approved):
You could replace a write lock with a read lock, but other than that LGTM.

server/mqtt.go
```diff
@@ -3256,14 +3254,50 @@ func (s *Server) mqttCheckPubRetainedPerms() {
...
+	}
+
+	// For each session we will obtain a list of retained messages.
+	var _rms [128]retainedMsg
+	rms := _rms[:0]
+	for _, asm := range asms {
+		// Get all of the retained messages. Then we will sort them so
+		// that they are in sequence order, which should help the file
+		// store to not have to load out-of-order blocks so often.
+		asm.mu.Lock()
```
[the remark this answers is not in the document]
neilalexander: Good call. Thanks for your reviews on this!

PR #4229, opened 2023-06-09
Reduce messages in chaos tests

neilalexander:
It doesn't really appear as though, for what these tests are trying to prove, that an excessively large number of messages is required. Instead let's drop the count a little in the hope that they run a bit faster.


derekcollison (approved):

mprimi (approved):
Makes sense. 
We may even jettison these tests now, equivalent ones exist on Failground already

PR #4230, opened 2023-06-09
Fixed issues with leafnode compression negotiation

kozlovic:
When a server would send an asynchronous INFO to a remote server it would incorrectly contain compression information that could cause issues with one side thinking that the connection should be compressed while the other side was not.

It also caused the authentication timer to be incorrectly set which would cause a disconnect.

Signed-off-by: Ivan Kozlovic <ijkozlovic@gmail.com>

### DIFF
--- a/server/leafnode.go
+++ b/server/leafnode.go
@@ -696,6 +696,8 @@ func (s *Server) startLeafNodeAcceptLoop() {
 
 	tlsRequired := opts.LeafNode.TLSConfig != nil
 	tlsVerify := tlsRequired && opts.LeafNode.TLSConfig.ClientAuth == tls.RequireAndVerifyClientCert
+	// Do not set compression in this Info object, it would possibly cause
+	// issues when sending asynchronous INFO to the remote.
 	info := Info{
 		ID:            s.info.ID,
 		Name:          s.info.Name,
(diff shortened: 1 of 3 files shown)


server/leafnode.go
```diff
@@ -987,7 +984,11 @@ func (s *Server) createLeafNode(conn net.Conn, rURL *url.URL, remote *leafNodeCf
 		// Grab server variables
 		s.mu.Lock()
 		info = s.copyLeafNodeInfo()
-		info.Compression = opts.LeafNode.Compression.Mode
+		// For tests that want to simulate old servers, do not set the compression
+		// on the INFO protocol if configured with CompressionNotSupported.
+		if cm := opts.LeafNode.Compression.Mode; cm != CompressionNotSupported {
```
neilalexander, line 989: Is there any value in checking that the proposed compression mode is one we understand?

kozlovic: Not sure I understand the question. If it is "are we sure that `opts.LeafNode.Compression.Mode` is a valid value here, then yes, it has been already validated on startup (or reload). The point of this test here, is that if we configure as "CompressionNotSupported" (likely in tests), then we don't include the Compression field in the INFO protocol, which makes this server behaves as an old server (protocol wise), since this field won't be set/present in old servers.

derekcollison (approved):
LGTM - Thanks!

PR #4232, opened 2023-06-10
[IMPROVED] Last msg lookup (KV Get) when subject is a literal subject

### LINKED ISSUE #4221: KV Get message rate increases with message size
Please see [my NATS KV benchmark script](https://gist.github.com/jjthiessen/5d55da45ea4a73ff3ec5d4394b9f31a5#file-bench-sh) (just a wrapper around `nats bench`) along with transcripts/console output for two runs (one with [file storage on SSD](https://gist.github.com/jjthiessen/5d55da45ea4a73ff3ec5d4394b9f31a5#file-bench-ssd-log), the other with [file storage on a ramfs mount](https://gist.github.com/jjthiessen/5d55da45ea4a73ff3ec5d4394b9f31a5#file-bench-ram-log)) (all part of [the same gist](https://gist.github.com/jjthiessen/5d55da45ea4a73ff3ec5d4394b9f31a5)). I only included file storage results here, as memory backed KV appears to be fairly uniformly slow (much slower than file storage) at the moment.

The `Put`/publish rates and trends more-or-less make sense to me/aren't surprising.
However, the `Get`/subscribe message throughput rates seem to increase with message size.
This is rather surprising (for me, at least) — I would expect the data throughput rate to increase with message size (as it does), but for the message throughput rate to decrease.
The effect of this is that in order to get better KV read performance, one only needs to pad values to produce larger messages (even if the extra content is otherwise useless).

Does anybody know i) why this might be the case; and ii) whether I'm holding it wrong?

For additional context see [this thread](https://natsio.slack.com/archives/CM3T6T7JQ/p1684950034097139) on Slack.

derekcollison:
When messages were very small and the key space was very large the performance of last message gets in the store layer (both file and memory) would degrade.

If the subject is literal we can optimize and avoid sequence scans that are needed when multiple subject states need to be considered.

Resolves #4221

### DIFF
--- a/server/filestore.go
+++ b/server/filestore.go
@@ -4665,7 +4665,16 @@ func (fs *fileStore) loadLast(subj string, sm *StoreMsg) (lsm *StoreMsg, err err
 			mb.mu.Unlock()
 			return nil, err
 		}
-		_, _, l := mb.filteredPendingLocked(subj, wc, mb.first.seq)
+		var l uint64
+		// Optimize if subject is not a wildcard.
+		if !wc {
+			if ss := mb.fss[subj]; ss != nil {
+				l = ss.Last
+			}
+		}
+		if l == 0 {
+			_, _, l = mb.filteredPendingLocked(subj, wc, mb.first.seq)
+		}
 		if l > 0 {
 			if mb.cacheNotLoaded() {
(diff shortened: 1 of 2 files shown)


Jarema (approved):
LGTM

PR #4233, opened 2023-06-10
[IMPROVED] Only enable JetStream account updates in clustered mode.

derekcollison:
If we know we are in stand alone mode we do not need to run the updates for JetStream account resources updates.

Resolves #4227 (Partial)


wallyqs (approved):
LGTM

Issue #4234, opened 2023-06-11
How to config a single nats-server with jetstream in most resource-effective way


derekcollison:
### Discussed in https://github.com/nats-io/nats-server/discussions/4227

<div type='discussions-op-text'>

<sup>Originally posted by **tpihl** June  8, 2023</sup>
Hi,

I am making a PoC using nats-server (and go-based nats client) without network access on a battery-powered SBC.

However, the nats-server seems to keep itself rather busy (relative term but significant for battery time) and I hope for help in calming it down when the rest of the world is carm.

Changing PING interval was easy but it seems to have a very small effect (most activity seems to be related to
$SYS.>
$JSC.>

Are these really necessary in a single instance nats-server with a single client in the same machine. If they are, can we reduce the frequency of gathering/sending that info when there is no other activity?

I havent tried to run in MQTT mode even though you probably guess that it's MQTT server this PoC competes with. My argument is borrowed from how i remember a  nats.io recommendation, something like "use real nats/jetstream if possible, and that it should prove as effective in a comparable scenario with MQTT"

Any ideas/suggestions?

/T</div>

derekcollison:
Will only send out JetStream account updates if in clustered mode (https://github.com/nats-io/nats-server/commit/a5de25f2132da1e1a00d2f65c266df6c01ebf87d)

Also will only send statsz updates if there is external interest.

PR #4235, opened 2023-06-11
[IMPROVED] Optimize statsz locking and sending in standalone mode.

derekcollison:
If we know we are in stand alone mode only send out statsz updates if we know we have external interest.

Resolves: #4234


wallyqs (approved):
LGTM

PR #4238, opened 2023-06-12
GitHub Actions

bruth:
WIP.. trying out a parallel unit test CI process in GHA alongside Travis

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -4,7 +4,7 @@ on:
   # push:
   # pull_request:
   schedule:
-    - cron: "0 * * * *" # Hourly
+    - cron: "15 * * * *" # Hourly, 15 mins past the hour
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
@@ -31,14 +31,9 @@ jobs:
           skip-build-cache: true
           args: --timeout=5m --config=.golangci.yml
 
-  build:
-    name: Build (Go ${{ matrix.go }}, ${{ matrix.os }})
-    strategy:
-      fail-fast: false
-      matrix:
-        os: ["${{ vars.GHA_WORKER_SMALL }}"]
-        go: [stable, oldstable] # Stable is Go N, oldstable is Go N-1
-    runs-on: ${{ matrix.os }}
+  build-latest:
+    name: Build (Latest Go, ${{ matrix.os }})
+    runs-on: ${{ vars.GHA_WORKER_SMALL }}
     steps:
       - name: Checkout
         uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


neilalexander:
Currently set this up so that the workflow won't run for PRs or pushes, but will run hourly, so that we can monitor how the test reliability is for a bit.

Workers can be selected by tweaking the GHA variables in the repo settings.

.github/workflows/tests.yaml
```diff
@@ -0,0 +1,236 @@
+name: NATS Server Tests
+
+on:
+  # push:
+  # pull_request:
+  schedule:
+    - cron: "0 * * * *" # Hourly
```
philpennock, line 7: Might want to avoid 0 simply to avoid everyone else also using 0 and having to wait a long time for resources to become available.  Getting lost in the stampede ...

.github/workflows/tests.yaml
```diff
@@ -0,0 +1,236 @@
...
+    name: Lint
+    runs-on: ${{ vars.GHA_WORKER_SMALL }}
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Install Go
+        uses: actions/setup-go@v5
+        with:
+          go-version: stable
```
philpennock, line 24: This means that when a new Go minor releases, based on past experience we risk having the lint tests failing until 
 1. main is updated with new test data for certs or whatever else they've changed in crypto/tls for the default assumption of the-world-is-http
 2. the lint tools are updated to handle new syntax elements

And similarly for where the matrix is `[ stable, oldstable ]`.

It's fine (for `main`, not release branches), as long as you're _expecting_ this to periodically break and don't want this to be a var?

.github/workflows/tests.yaml
```diff
@@ -0,0 +1,236 @@
...
+          skip-build-cache: true
+          args: --timeout=5m --config=.golangci.yml
+
+  build:
+    name: Build (Go ${{ matrix.go }}, ${{ matrix.os }})
+    strategy:
+      fail-fast: false
+      matrix:
+        os: ["${{ vars.GHA_WORKER_SMALL }}"]
+        go: [stable, oldstable] # Stable is Go N, oldstable is Go N-1
```
philpennock, line 40: nit: 1.N, 1.N-1

philpennock (approved):
Do you not want a gather step which depends upon the others, and a Slack notification?

.github/workflows/tests.yaml
neilalexander: Good point, I think the best compromise here is to have two build steps: one that tests against latest (as our users will obviously do the same) and the other that tests against whatever our minimum supported version is set to in `go.mod`. That should mean that we can bump the minimum version without having to mess with the workflow.

.github/workflows/tests.yaml
neilalexander: Good point, set to `15` instead.

derekcollison (approved):
LGTM

neilalexander:
The time-based trigger was just so we could evaluate whether or not the tests were running OK on GHA for a few days with the workers being different sizes, we can remove that safely.

PR #4243, opened 2023-06-13
[CHANGED] MQTT: Support for topics with `.` character.

### LINKED ISSUE #1879: MQTT Clients compatibility issue with '.' char and use in Eclipse Tahu MQTT specification
There is an issue using NATS.io as a MQTT broker for with the Eclipse Tahu specification.

Unfortunately the root namespace used for the topic contains a '.' char. (See [page 12 of the specification](https://www.eclipse.org/tahu/spec/Sparkplug%20Topic%20Namespace%20and%20State%20ManagementV2.2-with%20appendix%20B%20format%20-%20Eclipse.pdf))

Given that the '/' topic delimiter chars are converted to NATS topic alternative, should the NATS topic delimiter chars be converted or escaped for MQTT clients? 

Could there be a known substitution used internally, like converting '.' to '_' and NATS clients would need to know about?

So MQTT client publish to `spBv1.0/group/message/node` and NATS client subscribe to `spBv1_0.group.message.node`?

### LINKED ISSUE #3482: allow . in mqtt topics
## Feature Request

- Allow dots as topic separator in mqtt.
- Eventually, use / as topic separator in NATS and MQTT, common topic structure could be convenient for applications using both NATS and MQTT protocols.

#### Use Case:

Allow nats mqtt server to be used with spark plug B :

[Sparkplug B specifications](https://www.eclipse.org/tahu/spec/Sparkplug%20Topic%20Namespace%20and%20State%20ManagementV2.2-with%20appendix%20B%20format%20-%20Eclipse.pdf)

Sparkplug uses spBv1.0 as topic prefix.

#### Proposed Change:

- add a parameter to allow dots in mqtt topics, eventually URLEncode the topic for the conversion to the NATS topic

- Users that would like to une NATS with Sparkplug B protocol
- Users that would like tu use nats in replacement of other mqtt workers, already having a working system that uses dots in the mqtt topics
- Myself :)

#### Alternative Approaches

Allow the use of / as separator in NATS, to harmonize the NATS and MQTT topics.

kozlovic:
The `.` character will be transformed to `//` in NATS subject. For instance an MQTT message published on `spBv1.0/plant1` would be received by a NATS subscriber as `spBv1//0.plant1`.

Conversely, a NATS message published on `spBv1//0.plant1` would be received by an MQTT subscriber as `spBv1.0/plant1`.

Resolves #1879
Resolves #3482


derekcollison (commented):
Is `//` ilegal in a MQTT topic today? Meaning we will not encounter it natively in an MQTT application yes?

kozlovic:
> Is // illegal in a MQTT topic today? Meaning we will not encounter it natively in an MQTT application yes?

It is legal actually, but the point of this PR is converting `.` from an MQTT topic into `//` in a NATS subject. The reason I had to select `//` as opposed to `/` is because of the need to convert valid MQTT topic such as `/myhome` in MQTT to `/.myhome` in NATS (as described here: https://docs.nats.io/running-a-nats-service/configuration/mqtt#mqtt-topics-and-nats-subjects). There would be cases where I would not be able to know how to translate a NATS subject back into MQTT (meaning, there would be a risk of converting to the wrong original MQTT topic).

Btw, once that is approved/merged, I will have to do a PR against the documentation mentioned above to update the table.

derekcollison:
So what happens if we encounter `//` natively in an MQTT subject that moves into the NATS space (say a microservice) and then comes back in? Will we handle properly on convert it to a `.`?

kozlovic:
> So what happens if we encounter // natively in an MQTT subject that moves into the NATS space (say a microservice) and then comes back in? Will we handle properly on convert it to a .?

No, if it comes from MQTT, say as `a//b`, following the table I pointed to, we would convert to `a././.b`, which on the way back would be translated to `a//b`. However, if MQTT topic is `a.b`, then it would be translated to `a//b`, and converted back to `a.b`.
I have test that check conversions from MQTT to NATS and back that handles `//` in MQTT (that has been extended in this PR).

kozlovic:
See nats.docs PR: https://github.com/nats-io/nats.docs/pull/601

derekcollison (approved):
LGTM

PR #4244, opened 2023-06-13
Support TLS 1.2 RSA signature with PKCS#1 v1.5 padding

tbeets:
Cert Store (aka wincert) feature wasn't properly handling TLS 1.2 handshake with TLS 1.2 clients that do not support RSA signature with PSS padding.

With this update, Cert Store will perform either PKCS#1 v1.5 or PSS padding for RSA signature depending on what type is negotiated by the TLS 1.2 client.

Issue surfaces with the NATS .NET v1 client which supports TLS 1.2 only (.NET 4.6.2 dependency) only when the client application was hosted on Windows 10 Enterprise LTSC 2019 (equivalent also to Windows 10 1809 and Windows Server 2019).  If the same client was executed on a more modern Windows 10 release, RSA signature with PSS padding was negotiated and succeeded normally.

The Go NATS client as well as any client operating at TLS 1.3 level would not exhibit the issue as TLS 1.3 requires PSS.

Fix tested good on Windows 10 Enterprise LTSC 2019 host and in confirmed fixed in user's Windows environment where the issue was originally detected.

### DIFF
--- a/server/certstore/certstore_windows.go
+++ b/server/certstore/certstore_windows.go
@@ -62,6 +62,7 @@ const (
 
 	winNcryptKeySpec = 0xFFFFFFFF // CERT_NCRYPT_KEY_SPEC
 
+	winBCryptPadPKCS1   uintptr = 0x2
 	winBCryptPadPSS     uintptr = 0x8 // Modern TLS 1.2+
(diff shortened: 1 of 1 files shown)


neilalexander (approved):
Looks reasonable!

neilalexander:
Oh, please sign off in the commit and PR description too!

PR #4256, opened 2023-06-20
Use `crypto/rand.Read` instead of `math/rand.Read`

neilalexander:
As of Go 1.20, `math/rand.Read` is deprecated. In addition to that, it also isn't recommended for use in combination with anything cryptographic.

I haven't replaced all `math/rand` imports with `crypto/rand` imports because there are still some legitimate uses for the `math/rand` package in some places.

### DIFF
--- a/server/filestore.go
+++ b/server/filestore.go
@@ -18,7 +18,6 @@ import (
 	"bytes"
 	"crypto/aes"
 	"crypto/cipher"
-	"crypto/rand"
 	"crypto/sha256"
 	"encoding/binary"
 	"encoding/hex"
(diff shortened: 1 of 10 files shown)


derekcollison (commented):
Let's target for dev branch.

Also mrand not deprecated per se, just top level functions and auto-seeding.

neilalexander:
> Let's target for dev branch.

Done :-)

> Also mrand not deprecated per se, just top level functions and auto-seeding

Yep, it's just `mrand.Read` specifically that's deprecated, not the entire package. Anywhere that `Int31()` etc are being used are still valid.

wallyqs (approved):
LGTM

derekcollison (approved):
LGTM

PR #4259, opened 2023-06-21
[CHANGED] LeafNode: remotes from same server binding to same hub account

kozlovic:
Previously, the server would reject a second remote leafnode connection from the same server if it was binding to the same account on the hub even if the remote was using different local accounts.

### DIFF
--- a/server/leafnode_test.go
+++ b/server/leafnode_test.go
@@ -2185,65 +2185,80 @@ func TestLeafNodeLoopDetectedDueToReconnect(t *testing.T) {
 	checkLeafNodeConnected(t, sl)
 }
 
-func TestLeafNodeTwoRemotesBindToSameAccount(t *testing.T) {
+func TestLeafNodeTwoRemotesBindToSameHubAccount(t *testing.T) {
 	opts := DefaultOptions()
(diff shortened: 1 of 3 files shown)


server/leafnode_test.go
```diff
@@ -2185,65 +2185,80 @@ func TestLeafNodeLoopDetectedDueToReconnect(t *testing.T) {
...
-	}
-	lopts.NoLog = false
-	ln, err := NewServer(lopts)
-	if err != nil {
-		t.Fatalf("Error creating server: %v", err)
-	}
-	defer ln.Shutdown()
-	l := &captureErrorLogger{errCh: make(chan string, 10)}
-	ln.SetLogger(l, false, false)
+			lopts, err := ProcessConfigFile(lconf)
```
derekcollison, line 2226: Does this denote formatting spaces or something?

derekcollison (approved):
LGTM - Just asked about formatting.

server/leafnode_test.go
kozlovic: This is because I transformed this test from a single to a t.Run() to cover cases where it should work (different local accounts) and should not (same local account). So the "old" body was shifted to the right.

PR #4264, opened 2023-06-21
Additional check for TLS required in WebSocket `INFO`

neilalexander:
This should stop us reporting `tls_required` when we aren't reporting `tls_available`.


derekcollison (approved):
LGTM

wallyqs (approved):

PR #4268, opened 2023-06-22
Cert Store (aka wincert)

tbeets:
For NATS Server on Windows, provide option for TLS certificate and handshake signature to be provided by the Windows Certificate Store instead of PEM files.


server/certstore/certstore_other.go
```diff
@@ -0,0 +1,46 @@
...
+}
+
+// Public always returns nil public key since this is a stub on non-supported platform
+func (k otherKey) Public() crypto.PublicKey {
+	return nil
+}
+
+// Sign always returns a nil signature since this is a stub on non-supported platform
+func (k otherKey) Sign(rand io.Reader, digest []byte, opts crypto.SignerOpts) (signature []byte, err error) {
+	_, _, _ = rand, digest, opts
```
derekcollison, line 41: What is this?

derekcollison (approved):
LGTM - One comment

server/certstore/certstore_other.go
tbeets: Stub implementation for non-Windows ties back to L26 comment and root avoidance of creating a Windows platform copy of `opts.go` since that is such a _special_ source code file to keep maintained.

L40-41 in particular could be alternately phrased as:

```golang
func TLSConfig(_ StoreType, _ MatchByType, _ string, _ *tls.Config) error {
	// _, _, _, _ = certStore, certMatchBy, certMatch, config
	return ErrOSNotCompatCertStore
}
```

tbeets: Whoops.  Gave an alternate phrasing example for L29-32, not L40-41 but same thing.  Linters complaining about unused variables...

tbeets: For more background, this code in `opts.go` is cross-platform:

https://github.com/nats-io/nats-server/blob/f854e95c1101b3894dddb7044a7eb3962136c866/server/opts.go#L4342

The options and interface potentially directly usable if we want to tie into other 3rd-party stores in future...

derekcollison: I think the better pattern is to change the args to the functions as `_`

PR #4269, opened 2023-06-23
[IMPROVED] Consumer create time when stream has large number of blks.

derekcollison:
When creating a consumer on a stream with a very large number of msg blks, calculating numPending could be slow.

This aims to optimize a bit, more work to be done on streams with a very large (> 200k) number of msg blks.


neilalexander (approved):
LGTM

PR #4272, opened 2023-06-27
Adjust nightly name if on a branch

philpennock:
If we're on a branch other than main or dev, then when building "nightly", build it with the name of the branch instead.  Overrideable via an env var.

It's a bit ugly because of limitations of goreleaser templating, we can't haul the expression out to be defined once, but it works.

Can build a custom Docker image with:

    goreleaser release --snapshot -p 2 -f .goreleaser-nightly.yml --clean


wallyqs (approved):
LGTM

PR #4278, opened 2023-06-29
Add support for filtering Connections by client kind (Client, Leafnode etc)

codegangsta:
This PR adds support for `kind` in `*ConnzOptions`, which allows a `Connz` user to filter returned connections by it's `kind` field, namely `Client` or `Leafnode`.

This should help monitoring solutions better separate leaf node and regular client connections.

### DIFF
--- a/server/monitor.go
+++ b/server/monitor.go
@@ -384,7 +384,7 @@ func (s *Server) Connz(opts *ConnzOptions) (*Connz, error) {
 				}
 
 				// Do client kind filtering
-				if kind != _EMPTY_ && client.kindString() != kind {
+				if kind != _EMPTY_ && !strings.EqualFold(client.kindString(), kind) {
 					continue
 				}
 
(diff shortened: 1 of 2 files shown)


server/monitor.go
```diff
@@ -363,6 +380,12 @@ func (s *Server) Connz(opts *ConnzOptions) (*Connz, error) {
 				if mqttCID != _EMPTY_ && client.getMQTTClientID() != mqttCID {
 					continue
 				}
+
+				// Do client kind filtering
+				if kind != _EMPTY_ && client.kindString() != kind {
```
neilalexander, line 385: It may be worth `strings.ToLower`-ing both `client.kindString()` and `kind` to add case-insensitivity for those who are crafting their own API calls.

codegangsta: Great! Will add this in shortly

derekcollison:
This for main?

server/monitor.go
derekcollison: Can we get this one revived and merged in to main?

codegangsta: Yup. I can take a look at this now

codegangsta:
@derekcollison Rebased and fixed feedback from @neilalexander. Once tests pass feel free to merge this one 😄

PR #4284, opened 2023-07-01
Add in support for segmented binary stream snapshots.

derekcollison:
Streams with many interior deletes was causing issues due to the fact that the interior deletes were represented as a sorted []uint64. This could cause snapshots for streams R>1 and with lots of interior deletes to take up more memory and cpu then we want.

This new approach introduces 3 sub types of delete blocks, an avl bitmask tree, a run length encoding, and the legacy format above. We also take into account large interior deletes such that on receiving a snapshot we can skip things we already know about.


server/events.go
```diff
@@ -193,18 +193,53 @@ type ServerID struct {
 	ID   string `json:"id"`
 }
 
+// Type for out server capabilities.
```
neilalexander, line 196: `our`?

neilalexander (approved):
LGTM!

server/avl/seqset.go
```diff
@@ -250,9 +276,19 @@ var (
...
+	if len(buf) < minLen || buf[0] != magic {
+		return nil, -1, ErrBadEncoding
+	}
+
+	// For v1 version with 64 buckets per node.
+	if buf[1] == 1 {
+		return decodev1(buf)
+	}
+
+	if buf[1] != version {
```
bruth, line 290: Would it be worth splitting out the below logic to `decodev2` and change this to a switch statement?

server/events.go
```diff
@@ -193,18 +193,53 @@ type ServerID struct {
...
-	JetStream bool      `json:"jetstream"`
-	Time      time.Time `json:"time"`
+	Name    string   `json:"name"`
+	Host    string   `json:"host"`
+	ID      string   `json:"id"`
+	Cluster string   `json:"cluster,omitempty"`
+	Domain  string   `json:"domain,omitempty"`
+	Version string   `json:"ver"`
+	Tags    []string `json:"tags,omitempty"`
+	// Whether JetStream is enabled (deprecated).
```
bruth, line 208:
```suggestion
	// Whether JetStream is enabled (deprecated in favor of the `ServerCapability` flag).
```

server/events.go
```diff
@@ -193,18 +193,53 @@ type ServerID struct {
...
+	JetStream bool `json:"jetstream"`
+	// Generic capability flags
+	Flags ServerCapability
+	// Sequence and Time from the remote server for this message.
+	Sequence uint64    `json:"seq"`
```
bruth, line 213: This is an exported type from the `server` package, so if users are embedding and using this field for any reason, it would break their app.

server/avl/seqset.go
```diff
@@ -250,9 +276,19 @@ var (
 	ErrSetNotEmpty = errors.New("ss: set not empty")
 )
 
-func Decode(buf []byte) (*SequenceSet, error) {
-	if len(buf) < minLen || buf[0] != magic || buf[1] != version {
-		return nil, ErrBadEncoding
+// Decode returns the sequence set and number of bytes read from the buffer on success.
+func Decode(buf []byte) (*SequenceSet, int, error) {
```
bruth, line 280: This is an exported function, so it is technically a breaking change for embedded usage. Low probability of being used, but noting..

bruth: ☝️ Ignore my comment.. this package (`avl`) is new for 2.10.. so nothing breaking here.

server/events.go
derekcollison: Will fix.

server/avl/seqset.go
derekcollison: Sure..

server/events.go
derekcollison: Not following.. How would it break there app?

server/avl/seqset.go
derekcollison: It can decode both versions and I assume there will be these out there already so made sure that would work.

bruth: I was referring to the signature changing, but since it's new to 2.10 it is not breaking.

server/events.go
bruth: Renaming an existing field. If the server was embedded and the program used the field, updating to 2.10 would result in the field being renamed.

derekcollison: Ah, ok I see now what you are referring to. Should I move it back in your opinion?

bruth: Yes

derekcollison:
I squashed the commits, good to merge?

bruth (approved):

PR #4290, opened 2023-07-06
Make make server embedded easier

nanjj:
Put usage in server package instead of main, and make it extendable, and make LoadX509Keypair and ReadCreds configurable.
 - [x] Tests added

Resolves #

 -
 -
 -

### DIFF
--- a/main.go
+++ b/main.go
@@ -1,4 +1,4 @@
-// Copyright 2012-2022 The NATS Authors
+// Copyright 2012-2023 The NATS Authors
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
 // You may obtain a copy of the License at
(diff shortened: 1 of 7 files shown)


neilalexander (commented):
I'm a bit uncertain about exporting overloads for things like `LoadX509KeyPair` and `ReadCreds`. Can I ask exactly what problem you are trying to solve here?

Moving the usage flags into their own file is probably fine but it is unusual for us to need them in an embedded application. Instead you would just pass in the relevant options through config structs when starting NATS.

nanjj:
> 

Private key file and creds file need to be encrypted when saving to config file，to load then need to decrypted firstly. So I export the the method out for overloading.

For the flags, now nats-server keep usage message in `main` package while these flags definination in the `server` package. I think they should be keep in same place, `main` or `server` is OK.

neilalexander:
There's probably a bigger piece of work for embedding compatibility here, but closing for now as old/stale.

PR #4292, opened 2023-07-06
[FIXED] Stream config idempotency

jnmoyne:

Fixes a behavior where idempotency of re-defining the same stream more than once (with the same attributes) was broken due to the DeepEqual failing due to the StreamSource struct received from the client app not having a value for the `iname` structure field (as it's internal) but the StreamSource struct return from `mset.config()` would have it set.


neilalexander (approved):
LGTM. On the whole I'm not a fan of `reflect.DeepEqual` because it can act in surprising ways (and has been broken by Go releases before) but that's a future issue.

PR #4295, opened 2023-07-10
Do not start up if failing to open `_META_` store or unable to decrypt encrypted stream meta files

neilalexander:
It's probably better to do this when we can't decrypt (i.e. because the encryption config has disappeared from under us) so that we don't lose data by recreating assets.


derekcollison (commented):
Should be do this in main?

neilalexander:
Superseded by #4301 for now.

PR #4298, opened 2023-07-10
[ADDED] $SYS server request to 'kick' or 'LDM' a client connection

### LINKED ISSUE #1556: Client connections load balancing among nats cluster nodes
Hello! Currently the mechanism of load balancing of client connections to nats cluster is based on random node selection. The problem (already discussed in some of the issues, like https://github.com/nats-io/nats-server/issues/1359) is when we restart nats cluster. To be able to serve clients we restart cluster one node after another and this will lead us to approximately the following distribution (step-by-step example for 3 node cluster):
1) original connection distribution is 24 / 24 / 24
2) after node1 restart : 0 / 36 / 36
3) after node2 restart:  18 / 0 / 54
4) after node3 restart: 45 / 27 / 0

As you can see original uniform distribution 1/3 | 1/3 | 1/3 tends to 2/3 | 1/3 | 0 . I suggest you to add some modifications that will allow developers to get around this problem (for example, we use nats cluster in 24/7 fashion and never stop it completely but sometimes need to upgrade server version or our hardware). There can be several solutions:
1) Implement clever load balancing (when nats cluster node tells client to try to connect to another known node which have less active clients)
2) Provide a command which can be sent to cluster and cause it to disconnect existing clients - we can use it after cluster restart is finished to redistribute client connections among nodes.

jnmoyne:

Resolves #1556

Adds tw new $SYS server API endpoints:

- `$SYS.REQ.SERVER.%s.KICK` (where %s is the server_id) which 'kicks' (effectiveley 'rebalance' as the client application reconnects itself right away (potentially to another server in the cluster)). The service takes a JSON payload containing either an "id" or a "name" field. "id" disconnects the client connection id, "name" disconnects _all_ of the clients connected to the server with that name.

- `$SYS.REQ.SERVER.%s.LDM` (where %s is the server_id) and takes a JSON payload containing either an "id" or a "name" field. "id" sends an LDM Info message to the client connection id, "name" sends an LDM Info message to _all_ of the clients connected to the server with that name.

This features allow administrators to manually 're-balance' client connections between the servers in the cluster (e.g. after a rolling upgrade of the servers where one server ends up with no client connections after the upgrade), by kicking some of the client connections from one of the 'overloaded' (in comparison to other servers) servers in the cluster, causing them to re-estalibsh their connection to (hopefully) another server.

### DIFF
--- a/server/events.go
+++ b/server/events.go
@@ -57,6 +57,8 @@ const (
 	accConnsEventSubjNew      = "$SYS.ACCOUNT.%s.SERVER.CONNS"
 	accConnsEventSubjOld      = "$SYS.SERVER.ACCOUNT.%s.CONNS" // kept for backward compatibility
 	shutdownEventSubj         = "$SYS.SERVER.%s.SHUTDOWN"
+	clientKickReqSubj         = "$SYS.REQ.SERVER.%s.KICK"
+	clientLDMReqSubj          = "$SYS.REQ.SERVER.%s.LDM"
 	authErrorEventSubj        = "$SYS.SERVER.%s.CLIENT.AUTH.ERR"
 	authErrorAccountEventSubj = "$SYS.ACCOUNT.CLIENT.AUTH.ERR"
 	serverStatsSubj           = "$SYS.SERVER.%s.STATSZ"
@@ -1227,6 +1229,17 @@ func (s *Server) initEventTracking() {
 	subject = fmt.Sprintf(serverReloadReqSubj, s.info.ID)
 	if _, err := s.sysSubscribe(subject, s.noInlineCallback(s.reloadConfig)); err != nil {
 		s.Errorf("Error setting up server reload handler: %v", err)
+	}
+
+	// Client connection kick
+	subject = fmt.Sprintf(clientKickReqSubj, s.info.ID)
+	if _, err := s.sysSubscribe(subject, s.noInlineCallback(s.kickClient)); err != nil {
+		s.Errorf("Error setting up client kick service: %v", err)
+	}
+	// Client connection LDM
+	subject = fmt.Sprintf(clientLDMReqSubj, s.info.ID)
+	if _, err := s.sysSubscribe(subject, s.noInlineCallback(s.ldmClient)); err != nil {
+		s.Errorf("Error setting up client LDM service: %v", err)
 	}
 }
 
@@ -2711,6 +2724,49 @@ func (s *Server) reloadConfig(sub *subscription, c *client, _ *Account, subject,
 	s.zReq(c, reply, hdr, msg, optz, optz, func() (interface{}, error) {
 		// Reload the server config, as requested.
 		return nil, s.Reload()
+	})
+}
+
+type KickClientReq struct {
+	CID uint64 `json:"cid"`
+}
+
+type LDMClientReq struct {
+	CID uint64 `json:"cid"`
+}
+
+func (s *Server) kickClient(_ *subscription, c *client, _ *Account, subject, reply string, hdr, msg []byte) {
+	if !s.eventsRunning() {
+		return
+	}
+
+	var req KickClientReq
+	if err := json.Unmarshal(msg, &req); err != nil {
+		s.sys.client.Errorf("Error unmarshalling kick client request: %v", err)
+		return
+	}
+
+	optz := &EventFilterOptions{}
+	s.zReq(c, reply, hdr, msg, optz, optz, func() (interface{}, error) {
+		return nil, s.DisconnectClientByID(req.CID)
+	})
+
+}
+
+func (s *Server) ldmClient(_ *subscription, c *client, _ *Account, subject, reply string, hdr, msg []byte) {
+	if !s.eventsRunning() {
+		return
+	}
+
+	var req LDMClientReq
+	if err := json.Unmarshal(msg, &req); err != nil {
+		s.sys.client.Errorf("Error unmarshalling kick client request: %v", err)
+		return
+	}
+
+	optz := &EventFilterOptions{}
+	s.zReq(c, reply, hdr, msg, optz, optz, func() (interface{}, error) {
+		return nil, s.LDMClientByID(req.CID)
 	})
 }
 
--- a/server/server.go
+++ b/server/server.go
@@ -4252,3 +4252,36 @@ func (s *Server) changeRateLimitLogInterval(d time.Duration) {
 	default:
 	}
 }
+
+// DisconnectClientByID disconnects a client by connection ID
+func (s *Server) DisconnectClientByID(id uint64) error {
+	client := s.clients[id]
+	if client != nil {
+		client.closeConnection(Kicked)
+		return nil
+	}
+	return errors.New("no such client id")
+}
+
+// LDMClientByID sends a Lame Duck Mode info message to a client by connection ID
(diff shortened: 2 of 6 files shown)


server/server.go
```diff
@@ -4169,3 +4169,42 @@ func (s *Server) changeRateLimitLogInterval(d time.Duration) {
...
+// Disconnects a client by cid
+func (s *Server) DisconnectClientByID(id uint64) {
+	client := s.clients[id]
+	if client != nil {
+		client.closeConnection(Kicked)
+	}
+}
+
+// Disconnects a client by name
+func (s *Server) DisconnectClientsByName(name string) {
```
wallyqs, line 4182: maybe take this one out from the first pass, now I'm thinking that you could have same names under different accounts which could be an issue

derekcollison:
Let's think a bit more about all of this for 2.11, lot more moving parts and want to have a cohesive answer across the board.

server/events.go
```diff
@@ -2679,6 +2691,42 @@ func (s *Server) nsubsRequest(sub *subscription, c *client, _ *Account, subject,
 	s.sendInternalMsgLocked(reply, _EMPTY_, nil, nsubs)
 }
 
+type kickClientReq struct {
+	Name string `json:"name"`
+	Id   uint64 `json:"id"`
```
wallyqs, line 2696:
```suggestion
	ID   uint64 `json:"id"`
```

jnmoyne: Changed it to ID for both kick and ldm requests types

server/server.go
jnmoyne: Took out potentially too powerful 'by name' functionality for both kick and ldm

neilalexander (approved):
On the whole I think this looks good, but it needs at least a unit test to prove it works (probably using nats.go's `LameDuckModeHandler`) and preferably a second unit test to prove it's only reachable on the system account.

ripienaar (changes requested):
This doesnt look like it gives feedback about what was done, if a server disconnected a client or not or which handled these?

Looking at the recently added `reloadConfig` example it uses the `zReq()` helper to craft responses for which servers handled it and errors are sent back to the caller.

From the perspective of using this (CLI) I cant see how to make this a friendly UX, getting the server info and affected changes back means i can tell a user success/fail etc

server/events.go
```diff
@@ -2712,6 +2725,36 @@ func (s *Server) reloadConfig(sub *subscription, c *client, _ *Account, subject,
 	})
 }
 
+type kickClientReq struct {
```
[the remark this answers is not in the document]
jnmoyne: Changed to CID

ripienaar (approved):
LGTM - docs on exported structs and funcs would be nice, in go a function / struct doc starts with its name

server/server.go
```diff
@@ -4223,3 +4223,29 @@ func (s *Server) changeRateLimitLogInterval(d time.Duration) {
 	default:
 	}
 }
+
+// Disconnects a client by cid
+func (s *Server) DisconnectClientByID(id uint64) error {
```
wallyqs, line 4228: Should it be CID to be consistent with that meaning connection ID?

```suggestion
func (s *Server) DisconnectClientByCID(id uint64) error {
```

server/server.go
```diff
@@ -4223,3 +4223,29 @@ func (s *Server) changeRateLimitLogInterval(d time.Duration) {
...
+	client := s.clients[id]
+	if client != nil {
+		client.closeConnection(Kicked)
+		return nil
+	}
+	return errors.New("no such client id")
+}
+
+func (s *Server) LDMClientByID(id uint64) error {
```
wallyqs, line 4237:
```suggestion
// LDMClientByID disconnects a client by connection ID.
func (s *Server) LDMClientByID(id uint64) error {
```

server/server.go
```diff
@@ -4223,3 +4223,29 @@ func (s *Server) changeRateLimitLogInterval(d time.Duration) {
...
+	info := s.copyInfo()
+	info.LameDuckMode = true
+
+	c := s.clients[id]
+	if c != nil && c.opts.Protocol >= ClientProtoInfo &&
+		c.flags.isSet(firstPongSent) {
+		// sendInfo takes care of checking if the connection is still
+		// valid or not, so don't duplicate tests here.
+		s.Debugf("sending LDM info to cid=%d", id)
```
wallyqs, line 4246: Can we use the client logger instead here maybe? That would include the cid in the logger fields already.
```suggestion
		c.Debugf("sending lame duck mode INFO to client")
```

jnmoyne: Done

server/server.go
jnmoyne: Done

server/server.go
```diff
@@ -4223,3 +4223,29 @@ func (s *Server) changeRateLimitLogInterval(d time.Duration) {
 	default:
 	}
 }
+
+// Disconnects a client by cid
```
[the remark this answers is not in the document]
jnmoyne: Done

jnmoyne:
Ok should be all good now. @neilalexander please merge if you are good with those last changes and the new test.

neilalexander:
Test & race fix look good, please squash down to one commit and I'll hit merge in the morning!

neilalexander (approved):
LGTM!

dw612:
Hi what would be the nats version to have this feature?

derekcollison:
2.10.x will have that feature.

dw612:
Hi I used 2.10.3-alpine image for my nats statefulset, and used nats-pub tool to send this message "Published [$SYS.REQ.SERVER.NAVK5EQIMAJ4S7SMDSERITSXY2Q4MCDAMSFJYCURWDQC4OQTVVJ6AFFS.KICK] : '{"name":"rebalance-worker"}'" 
The server does not seem to be kicking any connections. What did I miss?

derekcollison:
This is an internal feature and works by CID.

dw612:
Oh that means it has no released image yet? Is there a estimated time for the release?

derekcollison:
It's in there. You have to supply CID.

dw612:
Sorry could you elaborate? CID do you mean by container id? I was following the above "$SYS.REQ.SERVER.%s.KICK (%s is the server_id)"

jnmoyne:
The payload of the request should be 
{"cid": xxx}

Where xxx is the connection id (a number) for that connection to the server. 

The initial version of the PR also allowed to kick by name but that was removed before the PR merge as it's too much of a potential footgun (since it would kick all of the connections with the name)

ripienaar:
You can use “nats server request kick” and its help will ask the required info. Add —trace to see how it works

ripienaar:
Just shut down the server? If you don’t want connection on it at all there is no point in hanging around. So shut it with the LDM signal to the process

derekcollison:
Since these are system wide operations, the controls at an individual server level will be very low level and raw. The plan is to provide system level directives through the NATS cli to balance things like connections and JetStream leaders.

04116:
hi team,
as proposal, the solution here is provide a way to disconnect and let client reconnect? Depend on st like proxy to balance number of clients on server?

Do we handle when client still reconnect to the previous server?

ripienaar:
Clients who have a list of servers - either in co fig or using topology discovery - will connect to a random next server but not the one they were just on.

PR #4299, opened 2023-07-11
[FIXED] Duplicate queue subs over leafnodes with hub stream import/export with possible dataloss

derekcollison:
When a queue subscriber was updated multiple times over a leafnode connection we added in more shadow subscriptions which could become zombies when the connection went away.

In a case where a leafnode server had multiple queue subscribers on the same queue group, the hub server would add in multiple shadow subs. These subs would not be properly cleaned up and could lead to stale connections being associated with them.


Jarema (approved):
LGTM!

neilalexander (approved):

PR #4302, opened 2023-07-12
Evaluate service imports when publishing advisories

### LINKED ISSUE #4275: Allow messages published on $JS.EVENT.ADVISORY.API to be exported via a service
### Use case:

I want to aggregate JetStream advisory events from multiple accounts on one aggregate account. This would allow `nats-surveyor` to maintain a single connection with the server to gather JetStream advisory metrics from multiple accounts.

I have the following config:
```
jetstream: {max_mem_store: 64GB, max_file_store: 10TB}
accounts {
	A {
		jetstream: enabled
		users: [ {user: pp, password: foo} ]
		imports [
			{ service: { account: AGG, subject: '$JS.EVENT.ADVISORY.ACC.A.>' }, to: '$JS.EVENT.ADVISORY.>' }
		]
	}
	AGG {
		users: [ {user: agg, password: foo} ]
		exports: [
			{ service: '$JS.EVENT.ADVISORY.ACC.*.>', response: Singleton, account_token_position: 5 }
		]
	}
}
```

**Intended behavior**
1. On JetStream events, server publishes messages on `$JS.EVENT.ADVISORY.>` in account `A`.
2. Account `AGG` exports a service with `account_token_position` on which the advisory events should be forwarded
3. All advosory messages from account `A` should be available on `$JS.EVENT.ADVISORY.ACC.A.>`

**Actual behavior**
Advisory events published on subjects other than `$JS.EVENT.ADVISORY.API` (e.g. `$JS.EVENT.ADVISORY.STREAM.CREATED.ORDERS`) are mapped correctly and available on account `AGG`. Messages published on `$JS.EVENT.ADVISORY.API` (the audit messages) are not available on `AGG`.

### Additional information

- Behavior is the same whether or not mapping is used, e.g. exporting `$JS.EVENT.>` -> `$JS.EVENT.>` does not fix the issue
- Intended behavior **can** be achieved using stream export from account `A` to account `AGG` and in that case it works fine (I can subscribe on `$JS.EVENT.ADVOSORY.ACC.A.>` and all events are available. However, using service exports would be a desirable solution since with services I can utilize `account_token_position` and have a single export on account `AGG` instead of an import per account.
- This seems to be caused by some advisory events being published using `Server.sendInternalAccountMsg()` (those are not forwarded) - https://github.com/nats-io/nats-server/blob/main/server/jetstream_api.go#L4332C29-L4332C29
- The same behavior is present when attempting to export service latency subjects (which also uses `sendInternalAccountMsg()`)
- I am not sure whether this behavior is intentional or not, but seems to be inconsistent given it differs depending on whether stream export or service export is used.

I added a test to show the issue: https://github.com/nats-io/nats-server/commit/3b19e16ac804192ef919d04d6c4e6be6ceebb25b
The first test (`TestJetStreamAccountImportJSAdvisoriesAsService`) uses service exports (and fails) and the second uses stream exports (which succeeds). Both tests have the same scenario:
1. On account `JS`, subscribe to `$JS.EVENT.ADVISORY.>`
2. On account `AGG`, subscribe to `$JS.EVENT.ADVISORY.ACC.JS.>`
3. Create a stream on account `JS`
4. Expect two messages on both subscriptions:
  - an action event on `$JS.EVENT.ADVISORY.STREAM.CREATED.ORDERS`
  - an api audit event on `$JS.EVENT.ADVISORY.API`

neilalexander:
This should fix #4275 by allowing advisories to be copied into other accounts via service imports.

### DIFF
--- a/server/client.go
+++ b/server/client.go
@@ -3127,8 +3127,8 @@ var needFlush = struct{}{}
 // deliverMsg will deliver a message to a matching subscription and its underlying client.
 // We process all connection/client types. mh is the part that will be protocol/client specific.
 func (c *client) deliverMsg(prodIsMQTT bool, sub *subscription, acc *Account, subject, reply, mh, msg []byte, gwrply bool) bool {
-	// Check sub client and check echo. Only do this if not an internal sub.
-	if sub.client == nil || (c == sub.client && !sub.client.echo && sub.icb == nil) {
+	// Check sub client and check echo. Only do this if not a service import.
+	if sub.client == nil || (c == sub.client && !sub.client.echo && !sub.si) {
 		return false
 	}
 
--- a/server/client.go
+++ b/server/client.go
@@ -3127,19 +3127,13 @@ var needFlush = struct{}{}
 // deliverMsg will deliver a message to a matching subscription and its underlying client.
 // We process all connection/client types. mh is the part that will be protocol/client specific.
 func (c *client) deliverMsg(prodIsMQTT bool, sub *subscription, acc *Account, subject, reply, mh, msg []byte, gwrply bool) bool {
-	// Check sub client and check echo
-	if sub.client == nil || c == sub.client && !sub.client.echo {
+	// Check sub client and check echo. Only do this if not an internal sub.
+	if sub.client == nil || (c == sub.client && !sub.client.echo && sub.icb == nil) {
 		return false
 	}
 
 	client := sub.client
 	client.mu.Lock()
-
-	// Check echo
-	if c == client && !client.echo {
-		client.mu.Unlock()
-		return false
-	}
 
 	// Check if we have a subscribe deny clause. This will trigger us to check the subject
 	// for a match against the denied subjects.
--- a/server/jetstream_test.go
+++ b/server/jetstream_test.go
@@ -10723,6 +10723,223 @@ func TestJetStreamAccountImportBasics(t *testing.T) {
 	}
 }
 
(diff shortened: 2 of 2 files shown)


server/jetstream_events.go
```diff
@@ -30,6 +30,21 @@ func (s *Server) publishAdvisory(acc *Account, subject string, adv interface{})
 		err = s.sendInternalAccountMsg(acc, subject, ej)
 		if err != nil {
 			s.Warnf("Advisory could not be sent for account %q: %v", acc.Name, err)
+			return
+		}
+
+		// To allow advisories to be captured by service imports/exports,
+		// we will need to check the subs separately.
+		for _, sub := range acc.sl.Match(subject).psubs {
```
derekcollison, line 38: Can Match() return nil?

server/jetstream_events.go
```diff
@@ -30,6 +30,21 @@ func (s *Server) publishAdvisory(acc *Account, subject string, adv interface{})
...
+			if im, ok := acc.imports.services[string(sub.subject)]; ok {
```
derekcollison, line 39: Should this logic be in sendInternalAccountMsg?

server/jetstream_events.go
derekcollison, line 38: Why not use processServiceImport from client.go?

derekcollison (approved):
LGTM

derekcollison (approved):
LGTM

PR #4307, opened 2023-07-13
Added: reload server config by sending it a message

levb:
 - [x] Tests added

Resolves #

- Reload server's configuration by sending a message to `$SYS.REQ.SERVER.{server-id}.RELOAD`


neilalexander:
Just as a heads up, CI build phase is red because of `server/events_test.go:2586: File is not `gofmt`-ed with `-s` (gofmt)`.

Looks like the right direction for sure, not sure what makes sense for a return value here. Maybe Derek has an opinion.

Can you please extend the test to also check that a non-system account can't hit this?

derekcollison:
I will take a look this weekend. 

Definitely get your editor setup to auto run go-fmt on save for sure.

levb:
oops about `gofmt`, @neilalexander will add the test, there's probably one already for `sysSubscribe`, will look.

derekcollison:
Should return the error of reload to the caller.

levb:
@neilalexander I added a test to ensure that a regular user can not reload. The failure is kinda silent (`NextMsg` fails with "no responder available"), not sure if it's something that we'd care to change and how to go about it. For other (monitoring) APIs it's appropriate.

server/events.go
```diff
@@ -2679,6 +2686,19 @@ func (s *Server) nsubsRequest(sub *subscription, c *client, _ *Account, subject,
...
 }
 
+// accountClaimUpdate will receive claim updates for accounts.
+func (s *Server) reloadConfig(sub *subscription, c *client, _ *Account, subject, reply string, hdr, msg []byte) {
+	if !s.eventsRunning() {
+		return
+	}
+
+	optz := &EventFilterOptions{}
+	s.zReq(c, reply, hdr, msg, optz, optz, func() (interface{}, error) {
```
derekcollison, line 2696: This mapping exists in a table IIRC that has been setup, best to do the same with this one I think.

server/events_test.go
```diff
@@ -2535,6 +2535,73 @@ func TestServerEventsStatszSingleServer(t *testing.T) {
...
+		}
+		ping_interval: "100ms"
+	`))
+	opts := LoadConfig(conf)
+	opts.Trace = true
+	opts.Debug = true
+	opts.TraceVerbose = true
+	s := RunServer(opts)
+	defer s.Shutdown()
+	subject := fmt.Sprintf("$SYS.REQ.SERVER.%s.RELOAD", s.info.ID)
```
derekcollison, line 2553: Use const here vs string literal?

server/events.go
```diff
@@ -60,6 +60,7 @@ const (
 	serverDirectReqSubj       = "$SYS.REQ.SERVER.%s.%s"
 	serverPingReqSubj         = "$SYS.REQ.SERVER.PING.%s"
 	serverStatsPingReqSubj    = "$SYS.REQ.SERVER.PING"             // use $SYS.REQ.SERVER.PING.STATSZ instead
+	serverReloadReqSubj       = "$SYS.REQ.SERVER.%s.RELOAD"        // with server ID
```
derekcollison, line 63: Should we have one that will trigger all servers? We use this pattern a bunch where one subject is for all servers and then you can target specific ones as well.

server/events_test.go
```diff
@@ -2535,6 +2535,73 @@ func TestServerEventsStatszSingleServer(t *testing.T) {
...
+	// subscription interest that we will receive updates.
+	nc, _ := jsClientConnect(t, s, nats.UserInfo("admin", "p1d"))
+	defer nc.Close()
+
+	// Request the server to reload and wait for the response.
+	// subject := fmt.Sprintf("$SYS.REQ.SERVER.%s.RELOAD", s.info.ID)
+	reply := nc.NewRespInbox()
+	sub, err = nc.SubscribeSync(reply)
+	require_NoError(t, err)
+	err = nc.PublishRequest(subject, reply, nil)
```
derekcollison, line 2590: We should have a test case for when reload fails as well.

server/events_test.go
levb: Hardcoding the `"$SYS..."` formats seems to be the standard operating procedure in events_test; 1/5 tests should have these strings independently defined (hard-coded?) anyway.

server/events.go
levb: 1/5 Reloading configs on all servers at the same time may not be great for the SLA? Maybe introduce this as one server at a time and add a general Reload later if requested?

server/events_test.go
derekcollison: If we have a const for doing the template we should use that..

server/events_test.go
levb: @derekcollison added.

server/events_test.go
```diff
@@ -2596,10 +2594,36 @@ func TestServerEventsReload(t *testing.T) {
...
+	require_NoError(t, err)
+	msg, err = sub.NextMsg(time.Second)
+	require_NoError(t, err)
+
+	apiResp = ServerAPIResponse{}
+	err = json.Unmarshal(msg.Data, &apiResp)
+	require_NoError(t, err)
+
+	require_True(t, apiResp.Data == nil)
+	require_True(t, apiResp.Error != nil)
```
derekcollison, line 2621: There is a require_Error() that takes var args for type of error that we expect..

like the following..

```go
require_Error(t, err, errors.New("nats: maximum messages per subject exceeded"))
// or
require_Error(t, err, NewJSReplicasCountCannotBeNegativeError())

```

derekcollison: We also have requre_NotNil() IIRC.

levb: changed. Any objections to using "github.com/stretchr/testify/require"?

bruth:
@levb I will handle documentation and ask questions if I have any.

Also, in case there are more changes coming.. if you rebase against main then you will only have the PR Travis build now 😄

server/events_test.go
derekcollison: Prefer to not introduce that dependency, we have a bunch of our own that work.

levb:
The consistent test failures were (both) caused by test data mismatches; fixed now. It still took a few runs to turn green, but now no consistent failures, so I think I got them all.

server/events_test.go
```diff
@@ -2535,6 +2535,95 @@ func TestServerEventsStatszSingleServer(t *testing.T) {
...
+	// Request the server to reload and wait for the response.
+	reply = nc.NewRespInbox()
+	sub, err = nc.SubscribeSync(reply)
+	require_NoError(t, err)
+	err = nc.PublishRequest(subject, reply, nil)
+	require_NoError(t, err)
+	msg, err = sub.NextMsg(time.Second)
+	require_NoError(t, err)
+
+	apiResp = ServerAPIResponse{}
```
derekcollison, line 2616: For this pattern just do `var apiResp ServerAPIResponse`

server/events_test.go
```diff
@@ -2535,6 +2535,95 @@ func TestServerEventsStatszSingleServer(t *testing.T) {
...
+		}
+		ping_interval: "200ms"
+	`), 0666)
+	require_NoError(t, err)
+
+	// Request the server to reload and wait for the response.
+	reply := nc.NewRespInbox()
+	sub, err = nc.SubscribeSync(reply)
+	require_NoError(t, err)
+	err = nc.PublishRequest(subject, reply, nil)
```
derekcollison, line 2588: Can just do an nc.Request()..

derekcollison (changes requested):

server/events_test.go
```diff
@@ -2535,6 +2535,84 @@ func TestServerEventsStatszSingleServer(t *testing.T) {
...
+	conf := createConfFile(t, []byte(`
+		listen: "127.0.0.1:-1"
+		accounts: {
+			$SYS { users [{user: "admin", password: "p1d"}]}
+			test { users [{user: "foo", password: "bar"}]}
+		}
+		ping_interval: "100ms"
+	`))
+	opts := LoadConfig(conf)
+	opts.Trace = true
```
derekcollison, line 2548: Does this test need these to be set to true? Trace, Debug and Verbose?

derekcollison (approved):
LGTM - One minor question

server/events.go
```diff
@@ -1207,6 +1208,12 @@ func (s *Server) initEventTracking() {
 	if _, err := s.sysSubscribeInternal(accSubsSubj, s.noInlineCallback(s.debugSubscribers)); err != nil {
 		s.Errorf("Error setting up internal debug service for subscribers: %v", err)
 	}
+
+	// Listen for requests to reload the server configuration.
+	subject = fmt.Sprintf(serverReloadReqSubj, s.info.ID)
+	if _, err := s.sysSubscribe(subject, s.noInlineCallback(s.reloadConfig)); err != nil {
+		s.Errorf("Error setting up internal tracking: %v", err)
```
ripienaar, line 1215: I think this error log can probably be improved, I dont think `internal tracking` will tell people what this is about.

ripienaar (changes requested):

server/events_test.go
levb: @derek removed

ripienaar (approved):
LGTM

PR #4309, opened 2023-07-14
Remap ports in leaf node tests to hopefully reduce flakes

neilalexander:
Multiple tests were using the same port numbers and it seems that the NATS Servers were not always shutting down quickly enough or the test is reusing a port that was already assigned ephemerally, resulting in `listen tcp 127.0.0.1:50554: bind: address already in use` failures.


wallyqs (approved):
LGTM

PR #4322, opened 2023-07-19
Add `first_seq` to `StreamConfig`

neilalexander:
This allows specifying `first_seq` in the stream config when creating a stream. 

### DIFF
--- a/server/filestore.go
+++ b/server/filestore.go
@@ -417,12 +417,11 @@ func newFileStoreWithCreated(fcfg FileStoreConfig, cfg StreamConfig, created tim
 	}
 
 	// If the stream has an initial sequence number then make sure we
-	// have purged up until that point. Checking the number of blocks
-	// is a cheap way of checking if the stream looks new-ish, rather
-	// than re-running a potentially expensive set of dirty closes on
-	// an existing stream. A new stream should have 1 block (as the
-	// lmb has already been created) but account for 0 just in case.
-	if len(fs.blks) <= 1 && cfg.FirstSeq > 0 {
+	// have purged up until that point. We will do this only if the
+	// recovered first sequence number is before our configured first
+	// sequence.
+	fs.FastState(&fs.state)
+	if cfg.FirstSeq > 0 && fs.state.FirstSeq <= cfg.FirstSeq {
 		if _, err := fs.purge(cfg.FirstSeq); err != nil {
 			return nil, err
 		}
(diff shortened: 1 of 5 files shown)


server/filestore.go
```diff
@@ -416,6 +416,18 @@ func newFileStoreWithCreated(fcfg FileStoreConfig, cfg StreamConfig, created tim
 		return nil, err
 	}
 
+	// If the stream has an initial sequence number then make sure we
+	// have purged up until that point. Checking the number of blocks
+	// is a cheap way of checking if the stream looks new-ish, rather
+	// than re-running a potentially expensive set of dirty closes on
+	// an existing stream. A new stream should have 1 block (as the
+	// lmb has already been created) but account for 0 just in case.
+	if len(fs.blks) <= 1 && cfg.FirstSeq > 0 {
```
derekcollison, line 425: After we recover we should just make sure that the first sequence, if set, if less than or equal to the recovered first sequence.

server/memstore.go
```diff
@@ -666,11 +671,19 @@ func (ms *memStore) PurgeEx(subject string, sequence, keep uint64) (purged uint6
...
+	return ms.purge(first)
+}
+
+func (ms *memStore) purge(fseq uint64) (uint64, error) {
 	ms.mu.Lock()
 	purged := uint64(len(ms.msgs))
 	cb := ms.scb
 	bytes := int64(ms.state.Bytes)
-	ms.state.FirstSeq = ms.state.LastSeq + 1
+	ms.state.FirstSeq = fseq
```
derekcollison, line 685: You need to formalize this in case first is not > last.

So stream of 1-100, purge to 50..

derekcollison: And in that case you can't just wipe the ms.msgs. Or you could error in memory store for partial purge.

neilalexander: For now I'll go with erroring on partial purge, as I think the only alternative is to iterate through the entire `ms.msgs` map and check for lower sequences.

derekcollison (approved):
LGTM

ripienaar:
Is this something we want user facing or keep for development?

ie. Should I add it to the CLI?

derekcollison:
Customer requested, so probably yes to add in CLI.

ripienaar:
I think we should prevent mirrors being added with custom first sequences as if you make it bigger than the seq of the source stream it will not work - I guess till the source stream reaches that seq?

wdyt?

jnmoyne:
It should fail if you try to use that option on a stream that is a mirror IMHO.

PR #4323, opened 2023-07-19
Don't require TLS for in-process connection

neilalexander:
This should fix a bug where in-process connections expect TLS over the `net.Pipe` if TLS is configured.

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -2617,6 +2617,13 @@ func (s *Server) createClientEx(conn net.Conn, inProcess bool) *client {
 		info.AuthRequired = false
 	}
 
+	// Check to see if this is an in-process connection with tls_required.
+	// If so, set as not required, but available.
+	if inProcess && info.TLSRequired {
+		info.TLSRequired = false
+		info.TLSAvailable = true
+	}
+
 	s.totalClients++
 	s.mu.Unlock()
 
@@ -2667,7 +2674,7 @@ func (s *Server) createClientEx(conn net.Conn, inProcess bool) *client {
 	}
 	s.clients[c.cid] = c
 
-	tlsRequired := info.TLSRequired && !inProcess
+	tlsRequired := info.TLSRequired
 	s.mu.Unlock()
 
 	// Re-Grab lock
--- a/server/server.go
+++ b/server/server.go
@@ -2217,7 +2217,7 @@ func (s *Server) AcceptLoop(clr chan struct{}) {
 func (s *Server) InProcessConn() (net.Conn, error) {
 	pl, pr := net.Pipe()
 	if !s.startGoRoutine(func() {
-		s.createClient(pl)
+		s.createClientInProcess(pl)
 		s.grWG.Done()
 	}) {
 		pl.Close()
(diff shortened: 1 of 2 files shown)


neilalexander:
@caleblloyd Please test against this commit, should fix your issue.

server/server.go
```diff
@@ -2571,7 +2571,7 @@ func (c *tlsMixConn) Read(b []byte) (int, error) {
 	return c.Conn.Read(b)
 }
 
-func (s *Server) createClient(conn net.Conn) *client {
+func (s *Server) createClient(conn net.Conn, inProcess bool) *client {
```
derekcollison, line 2574: Prefer a new func called createInProcessClient(), keep the create client the way it is and add in another that takes the bool flag..

derekcollison: We will be formalizing alot on clients, so might be worth it imo.

neilalexander: Sure, have done.

server/server.go
```diff
@@ -2572,6 +2572,14 @@ func (c *tlsMixConn) Read(b []byte) (int, error) {
...
 
 func (s *Server) createClient(conn net.Conn) *client {
+	return s._createClient(conn, false)
+}
+
+func (s *Server) createClientInProcess(conn net.Conn) *client {
+	return s._createClient(conn, true)
+}
+
+func (s *Server) _createClient(conn net.Conn, inProcess bool) *client {
```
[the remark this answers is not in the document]
neilalexander: Fixed

derekcollison (approved):
LGTM

caleblloyd:
@sethjback can you test the latest commit to make sure TLS is not required now?

caleblloyd:
Getting a new error.  Before this change, was seeing:

```
Error: tls: failed to verify certificate: x509: certificate signed by unknown authority
```

After this change, am seeing:

```
Error: tls: first record does not look like a TLS handshake
```

derekcollison:
I will wait to merge until I hear from @neilalexander on the issue above.

neilalexander:
Have pushed another commit that should resolve this without having to set `AllowNonTLS` in the config file, which would apply to all connections. Please give it a try.

caleblloyd:
`tls_required: true` is still being sent in Server Info to the client, so the Client is attempting to upgrade to TLS...  Think we need to detect/change that for the in-process client to send to `tls_required: false, tls_available: true`

caleblloyd:
I think https://github.com/nats-io/nats-server/commit/4b064ff8eaf90346372c137927247eeb889a40c0 works, feel free to cherry-pick that as it is based on this branch already

neilalexander:
OK, have added a unit test this time.

caleblloyd (approved):
Looks good!

test/tls_test.go
```diff
@@ -72,6 +72,26 @@ func TestTLSConnection(t *testing.T) {
 	}
 }
 
+// TestTLSInProcessConnection checks that even if TLS is enabled on the server,
+// that an in-process connection that does *not* use TLS still connects successfully.
+func TestTLSInProcessConnection(t *testing.T) {
+	srv, opts := RunServerWithConfig("./configs/tls.conf")
+	defer srv.Shutdown()
+
+	nc, err := nats.Connect("", nats.InProcessServer(srv), nats.UserInfo(opts.Username, opts.Password))
```
derekcollison, line 81: Small nit, _EMPTY_ vs ""

derekcollison (approved):
LGTM

test/tls_test.go
derekcollison: Let me know if you fix or not and will merge after.

neilalexander: `_EMPTY_` isn't defined in the `test` package, only in `server`.

PR #4325, opened 2023-07-19
v2.11: Add digest to represent configuration state

wallyqs:
Makes the server generate a digest representing the expanded version of the parsed configuration:

```
$ nats-server -c foo.conf -t
nats-server: configuration file configs/foo.conf is valid (sha256:e14d9fd914ee4a83c481c6bf97b44d7bc23d8f475f7f30af0c4e96ab4e5ad00c)
```

It can be inspected via `/varz`:

```
  "config_load_time": "2023-07-19T14:28:28.77128Z",
  "config_digest": "sha256:6a383afa5fa6e00262105fbf061cabf48db892d1aa677f491dbabe7894393413",
  "system_account": "$SYS"
```

And it is also logged at runtime making it possible to check whether the reload had side effects:
```
[1935] 2023/07/19 02:57:29.245020 [INF] Using configuration file: configs/foo.conf (sha256:6a383afa5fa6e00262105fbf061cabf48db892d1aa677f491dbabe7894393413)
...
[1935] 2023/07/19 02:58:06.993874 [DBG] Trapped "hangup" signal
...
[1935] 2023/07/19 02:58:06.996113 [INF] Reloaded server configuration (sha256:6a383afa5fa6e00262105fbf061cabf48db892d1aa677f491dbabe7894393413)
...
[1935] 2023/07/19 02:58:32.895492 [INF] Reloaded server configuration (sha256:a622fb453f3464b041de668b82229aef155175824f3c3a6d126a33268d172987)
[1935] 2023/07/19 02:58:44.615983 [DBG] Trapped "hangup" signal
...
[1935] 2023/07/19 02:58:44.617870 [INF] Reloaded server configuration (sha256:6a383afa5fa6e00262105fbf061cabf48db892d1aa677f491dbabe7894393413)
```

 - [x] Reloader tests
 - [x] Monitoring test

### DIFF
--- a/conf/parse.go
+++ b/conf/parse.go
@@ -26,6 +26,7 @@ package conf
 // see parse_test.go for more examples.
 
 import (
+	"crypto/sha256"
 	"encoding/json"
 	"fmt"
 	"os"
@@ -35,6 +36,8 @@ import (
 	"time"
 	"unicode"
 )
+
+const _EMPTY_ = ""
 
 type parser struct {
 	mapping map[string]any
@@ -108,6 +111,44 @@ func ParseFileWithChecks(fp string) (map[string]any, error) {
 	return p.mapping, nil
 }
 
+// cleanupUsedEnvVars will recursively remove all already used
+// environment variables which might be in the parsed tree.
+func cleanupUsedEnvVars(m map[string]interface{}) {
+	for k, v := range m {
+		t := v.(*token)
+		if t.usedVariable {
+			delete(m, k)
+			continue
+		}
+		// Cleanup any other env var that is still in the map.
+		if tm, ok := t.value.(map[string]interface{}); ok {
+			cleanupUsedEnvVars(tm)
+		}
+	}
+}
+
+// ParseWithDigest returns the processed config and a digest
+// that represents the configuration.
+func ParseWithDigest(fp string) (map[string]interface{}, string, error) {
+	data, err := os.ReadFile(fp)
+	if err != nil {
+		return nil, _EMPTY_, err
+	}
+	p, err := parse(string(data), fp, true)
+	if err != nil {
+		return nil, _EMPTY_, err
+	}
+	// Filter out any environment variables before taking the digest.
+	cleanupUsedEnvVars(p.mapping)
+	digest := sha256.New()
+	e := json.NewEncoder(digest)
+	err = e.Encode(p.mapping)
+	if err != nil {
+		return nil, _EMPTY_, err
+	}
+	return p.mapping, fmt.Sprintf("sha256:%x", digest.Sum(nil)), nil
+}
+
 type token struct {
 	item         item
 	value        any
@@ -116,10 +157,15 @@ type token struct {
 }
 
 func (t *token) MarshalJSON() ([]byte, error) {
+	// Do not duplicate variable contents since
+	// they already have been expanded at this point.
+	if t.usedVariable {
+		return []byte("null"), nil
+	}
 	return json.Marshal(t.value)
 }
 
-func (t *token) Value() any {
+func (t *token) Value() interface{} {
 	return t.value
 }
 
--- a/conf/parse.go
+++ b/conf/parse.go
@@ -113,7 +113,7 @@ func ParseFileWithChecks(fp string) (map[string]any, error) {
 
 // cleanupUsedEnvVars will recursively remove all already used
 // environment variables which might be in the parsed tree.
-func cleanupUsedEnvVars(m map[string]interface{}) {
+func cleanupUsedEnvVars(m map[string]any) {
 	for k, v := range m {
 		t := v.(*token)
 		if t.usedVariable {
@@ -121,7 +121,7 @@ func cleanupUsedEnvVars(m map[string]interface{}) {
 			continue
 		}
 		// Cleanup any other env var that is still in the map.
-		if tm, ok := t.value.(map[string]interface{}); ok {
+		if tm, ok := t.value.(map[string]any); ok {
 			cleanupUsedEnvVars(tm)
 		}
 	}
@@ -129,7 +129,7 @@ func cleanupUsedEnvVars(m map[string]interface{}) {
 
 // ParseFileWithChecksDigest returns the processed config and a digest
 // that represents the configuration.
-func ParseFileWithChecksDigest(fp string) (map[string]interface{}, string, error) {
+func ParseFileWithChecksDigest(fp string) (map[string]any, string, error) {
 	data, err := os.ReadFile(fp)
 	if err != nil {
 		return nil, _EMPTY_, err
@@ -157,15 +157,10 @@ type token struct {
 }
 
 func (t *token) MarshalJSON() ([]byte, error) {
-	// Do not duplicate variable contents since
-	// they already have been expanded at this point.
-	if t.usedVariable {
-		return []byte("null"), nil
-	}
 	return json.Marshal(t.value)
 }
 
-func (t *token) Value() interface{} {
+func (t *token) Value() any {
 	return t.value
 }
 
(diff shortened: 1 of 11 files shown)


derekcollison:
What about includes? ENV vars?

wallyqs:
> What about includes? ENV vars?

@derekcollison they are supported: https://github.com/nats-io/nats-server/blob/e2ab3e16de341fc922daa863237781103d3c875d/conf/parse_test.go#L441-L474

derekcollison:
Where do you see this being useful? Juts curious..

wallyqs:
@derekcollison I think it would help in a few ways:

- When troubleshooting the server since can know from the logs the state of the config that was applied from the digest 

- When refactoring the configurations since using a single file or dividing it into multiple ones with `include`s or env vars also results in the same digest.

- Check whether the server is running with the expected config, for example can first generate the config and get the digest:
```
$ nats-server -c nats.conf -t
nats-server: configuration file nats.conf is valid (sha256:e14d9fd914ee4a83c481c6bf97b44d7bc23d8f475f7f30af0c4e96ab4e5ad00c)
```
Then via monitoring can assert whether the server is running with that version or not yet.

neilalexander:
In principle this looks good, but what I'm wondering is what you could compare the hash to.

For example, if you have a config file with environment variables or includes, then you can't just sha256sum the file from your terminal, so you might be able to tell from repeated runs that the config changed, but you might not be able to tell that it did what you wanted.

Does it match up against some hash that Kubernetes exposes?

ripienaar:
Perhaps add a utility in the server binary to load a config and expand all the env, includes etc - and print the hash?

Of course I can also do that in the nats cli as these functions are exported - thats a better idea.

conf/parse.go
```diff
@@ -98,13 +100,42 @@ func ParseFileWithChecks(fp string) (map[string]interface{}, error) {
 	return p.mapping, nil
 }
 
+// ParseWithDigest returns the processed config and a digest
+// that represents the configuration.
+func ParseWithDigest(fp string) (map[string]interface{}, string, error) {
+	data, err := os.ReadFile(fp)
+	if err != nil {
+		return nil, "", err
```
derekcollison, line 108: _EMPTY_

conf/parse.go
```diff
@@ -98,13 +100,42 @@ func ParseFileWithChecks(fp string) (map[string]interface{}, error) {
...
+	}
+	p, err := parse(string(data), fp, true)
+	if err != nil {
+		return nil, "", err
```
derekcollison, line 112: Same here..

conf/parse.go
```diff
@@ -98,13 +100,42 @@ func ParseFileWithChecks(fp string) (map[string]interface{}, error) {
...
+	}
+	digest := sha256.New()
+	e := json.NewEncoder(digest)
+	err = e.Encode(p.mapping)
+	if err != nil {
+		return nil, "", err
```
derekcollison, line 118: Same

derekcollison (changes requested):

wallyqs:
The current branch already supports env vars and includes so before making a deploy, you can check the digests using the `-t` test flag for each one of the configs for example:

```
$ POD_NAME=nats-0 nats-server -c server.conf -t
nats-server: configuration file server.conf is valid (sha256:dce6bc041e1953df8d4685ae8de7451cdef39ee7c7f295d11f4a486097a86941)

$ POD_NAME=nats-1 nats-server -c server.conf -t
nats-server: configuration file server.conf is valid (sha256:f3f0933a0e6fcdaa7c791e6f9e63abf69c2bcbcc8a98ac90c085de0c714e7d75)

$ POD_NAME=nats-2 nats-server -c server.conf -t
nats-server: configuration file server.conf is valid (sha256:8fc1b68163b6603b4cc260da597ce5654bbcc851e10744751242920480871c98)
```

conf/parse.go
```diff
@@ -98,13 +102,42 @@ func ParseFileWithChecks(fp string) (map[string]interface{}, error) {
...
 	value        interface{}
 	usedVariable bool
 	sourceFile   string
 }
 
+func (t *token) MarshalJSON() ([]byte, error) {
+	// Do not duplicate variable contents since
+	// they already have been expanded at this point.
+	if t.usedVariable {
+		return []byte("null"), nil
```
derekcollison, line 136: What is []byte("null") here?

wallyqs: it is so that the environment variables that have been already expanded are not duplicated, so with this for example:

```hcl
BAR_USERS = [{user: a}]

accounts {
  A { users = $BAR_USERS }
  B { users = [{user: b}]}
}
```

when expanded it becomes in the map:

```hcl
{"BAR_USERS":[{"user":"a"}],"accounts":{"A":{"users":[{"user":"a"}]},"B":{"users":[{"user":"b"}]}}}
```

so added this check to skip that with a `null` value when creating the expanded version:

```
{"BAR_USERS":null,"accounts":{"A":{"users":[{"user":"a"}]},"B":{"users":[{"user":"b"}]}}}
```

the right way though would be to skip including the env vars into the result after they have been expanded, not just leaving them undefined...

derekcollison: Feels brittle to me, what do you think? Should we try to remove them and just leave expanded? If we ship with them in we can never remove them and have the hash match.

wallyqs: Updated to clear the env vars before taking the digest so they would now match:

```go
		{
			`# substitutions
                        BAR_USERS = { users = [ {user = "bar"} ]}
                        hello = 'world'
                        accounts {
                          QUUX_USERS = [ { user: quux }]
                          bar = $BAR_USERS
                          quux = { users = $QUUX_USERS }
                        }
                        very { nested { env { VAR = 'NESTED', quux = $VAR }}}
                        `,
			nil,
			"sha256:34f8faf3f269fe7509edc4742f20c8c4a7ad51fe21f8b361764314b533ac3ab5",
		},
		{
			`# substitutions, same as previous one without env vars.
                        hello = 'world'
                        accounts {
                          bar  = { users = [ { user = "bar" } ]}
                          quux = { users = [ { user: quux   } ]}
                        }
                        very { nested { env { quux = 'NESTED' }}}
                        `,
			nil,
			"sha256:34f8faf3f269fe7509edc4742f20c8c4a7ad51fe21f8b361764314b533ac3ab5",
		},
```

wallyqs:
rebased and addressed the review comments ✅

conf/parse.go
```diff
@@ -98,13 +102,60 @@ func ParseFileWithChecks(fp string) (map[string]interface{}, error) {
...
+		// Cleanup any other env var that is still in the map.
+		if tm, ok := t.value.(map[string]interface{}); ok {
+			cleanupUsedEnvVars(tm)
+		}
+	}
+}
+
+// ParseWithDigest returns the processed config and a digest
+// that represents the configuration.
+func ParseWithDigest(fp string) (map[string]interface{}, string, error) {
```
neilalexander, line 123: Minor thing but let's make this naming clearer based on the other functions it is replacing, i.e. `ParseFileWithChecksDigest`.

server/reload.go
```diff
@@ -1589,6 +1589,8 @@ func (s *Server) diffOptions(newOpts *Options) ([]option, error) {
 			diffOpts = append(diffOpts, &ocspOption{newValue: newValue.(*OCSPConfig)})
 		case "ocspcacheconfig":
 			diffOpts = append(diffOpts, &ocspResponseCacheOption{newValue: newValue.(*OCSPResponseCacheConfig)})
+		case "configdigest":
```
neilalexander, line 1592: Bit unclear what this is for, aren't we supposed to be populating `ConfigDigest` automatically in `ProcessConfigFile`? Is there a reason why we'd want to take the digest and override it from the config input?

wallyqs: good point, updated to skip it since already done as you mention. I just have to add something here since otherwise it would complain with `config reload not supported` for field but a `continue` would suffice.

derekcollison (approved):
LGTM - Nice work!

PR #4327, opened 2023-07-20
[FIXED] Server reload with highly active accounts with service imports could cause panic or dataloss

derekcollison:
When service imports were reloaded on active accounts with lots of traffic the server could panic or lose data.


wallyqs (approved):
LGTM

neilalexander (approved):
LGTM

PR #4329, opened 2023-07-20
[FIXED] Leaking memory on usage of getHash()

### LINKED ISSUE #4289: Leak memory when creating a consumer name with jetstream.
I have problem leaking mem in sync.(*Map).Store in function github.com/nats-io/nats-server/v2/server.getHashSize.
I debug using go tool pprof.  Help me please!

derekcollison:
If we created lots of hashes, beyond server names, like for consumer or stream NRG group names etc, these maps would grow and not release memory. Performance hit is ~300ns per call, and we can use string intern trick if need be at a future date since it is GC friendly.

Resolves #4289

### DIFF
--- a/server/events.go
+++ b/server/events.go
@@ -843,35 +843,15 @@ func getHash(name string) string {
 	return getHashSize(name, sysHashLen)
 }
 
-var nameToHashSize8 = sync.Map{}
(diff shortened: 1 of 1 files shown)


wallyqs (approved):
LGTM

PR #4330, opened 2023-07-20
Track slow consumers per connection type

wallyqs:
Adds a new `slow_consumer_stats` field to varz to get more details about the types of connections that are becoming slow consumers:

```
"slow_consumer_stats": {
    "clients": 0,
    "routes": 0,
    "gateways": 0,
    "leafs": 0
  }
```

- [X] Client Tests
- [X] Routes Tests
- [x] Gateway tests
- [x] Leafnode tests

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -122,6 +122,7 @@ type Server struct {
 	// How often user logon fails due to the issuer account not being pinned.
 	pinnedAccFail uint64
 	stats
+	scStats
 	mu                  sync.RWMutex
 	kp                  nkeys.KeyPair
 	xkp                 nkeys.KeyPair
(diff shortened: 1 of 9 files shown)


neilalexander (approved):
Interesting idea, LGTM

server/server.go
```diff
@@ -330,6 +330,10 @@ type stats struct {
 	inBytes       int64
 	outBytes      int64
 	slowConsumers int64
+	scClients     int64
```
derekcollison, line 333: If these are all atomic accessed, we can now use atomic.Uint64..

server/server.go
```diff
@@ -330,6 +330,10 @@ type stats struct {
 	inBytes       int64
 	outBytes      int64
 	slowConsumers int64
```
derekcollison, line 332: Maybe make its own struct, slowConsumers and inside is total, client, routes, etc..

derekcollison (changes requested):

server/server.go
wallyqs: put the slow consumer counters in their own struct using atomic.Uint64 instead, thanks for the suggestion 👍

wallyqs: ah but atomic.Uint64 is only available in +Go 1.19... and we still have Go 1.18 in Travis right now...

wallyqs: rebased and updated Go 1.19 as the minimum supported version for dev branch.

PR #4331, opened 2023-07-22
Include `logtime_utc` option into v2.9 series

wallyqs:
Includes option to log in UTC with commits from following PRs:

https://github.com/nats-io/nats-server/pull/3833
https://github.com/nats-io/nats-server/pull/3920

### DIFF
--- a/logger/log.go
+++ b/logger/log.go
@@ -38,12 +38,36 @@ type Logger struct {
 	fl         *fileLogger
 }
 
-// NewStdLogger creates a logger with output directed to Stderr
-func NewStdLogger(time, debug, trace, colors, pid bool) *Logger {
+type LogOption interface {
+	isLoggerOption()
+}
+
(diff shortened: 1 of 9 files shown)


logger/log.go
```diff
@@ -38,13 +38,37 @@ type Logger struct {
 	fl         *fileLogger
 }
 
-// NewStdLogger creates a logger with output directed to Stderr
-func NewStdLogger(time, debug, trace, colors, pid bool) *Logger {
+type LogOption interface {
```
derekcollison, line 41: Why is type public but not the method?

wallyqs: That makes it so that the `LogOption` can be used in the `NewStdLogger` func, but limiting the interface to only be implemented by code within the `nats-server/logger` package.

derekcollison (approved):
LGTM

PR #4336, opened 2023-07-25
Add Go 1.20 to Travis and Nightly images

wallyqs:
Picks up https://github.com/nats-io/nats-server/pull/4297 into main.  Includes: 

- Using Go 1.20 for the nightly images and Travis tests
- Drop Go 1.18
- Updates to GitHub Actions
- Upgrade to golang-ci

### DIFF
--- a/.github/workflows/go-test.yaml
+++ b/.github/workflows/go-test.yaml
@@ -5,9 +5,7 @@ jobs:
   test:
     strategy:
       matrix:
-        # This should be quoted or use .x, but should not be unquoted.
-        # Remember that a YAML bare float drops trailing zeroes.
-        go: ['1.19']
+        go: ["1.20"]
 
     env:
       GOPATH: /home/runner/work/nats-server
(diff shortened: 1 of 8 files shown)


.github/workflows/go-test.yaml
```diff
@@ -19,15 +19,15 @@ jobs:
...
         with:
           go-version: ${{matrix.go}}
 
       - name: Install deps
         shell: bash --noprofile --norc -x -eo pipefail {0}
         run: |
-          go get -u honnef.co/go/tools/cmd/staticcheck
-          go get -u github.com/client9/misspell/cmd/misspell
+          go install honnef.co/go/tools/cmd/staticcheck@latest
+          go install github.com/client9/misspell/cmd/misspell@latest
```
neilalexander, line 30: I think we should remove these and use `golangci-lint`, as we already do in `runTestsOnTravis.sh`. That way the single source of linting truth is `.golangci.yml` for all.

wallyqs: golangci includes the misspell too?

neilalexander: It does

wallyqs: sounds good to me, ok changing to this:

```sh
      - name: Install deps
        shell: bash --noprofile --norc -x -eo pipefail {0}
        run: |
          go install github.com/golangci/golangci-lint/cmd/golangci-lint@v1.53.3

      - name: Lint
        shell: bash --noprofile --norc -x -eo pipefail {0}
        run: |
          golangci-lint run
```

bruth:
Sorry I missed this before! Looks like it seems a rebase.

wallyqs:
rebased and included fix for this behavior change from go 1.20 #4373

bruth (approved):

PR #4339, opened 2023-07-27
Report correct consumer count in paged list response

ripienaar:
Previously the Total in paged responses would always equal the size of the first response this would stall paged clients after the first page.

Now correctly sets the total so paging continues, improves the test to verify these aspects of the report


derekcollison (commented):
LGTM

derekcollison (approved):
LGTM

PR #4344, opened 2023-07-28
[FIXED] Stream recovery with corrupt msg block with sequence gaps.

derekcollison:
This is a fix for a bad msg blk detected in the field that had sequence holes.

The stream had max msgs per subject of one and only one subject but had lots of messages. The stream did not recover correctly, and upon further inspection determined that a msg blk had holes, which should not be possible.

We now detect the holes and deal with the situation appropriately. Heavily tested on the data dump from the field.

### DIFF
--- a/server/filestore.go
+++ b/server/filestore.go
@@ -275,6 +275,8 @@ const (
 	wiThresh = int64(30 * time.Second)
 	// Time threshold to write index info for non FIFO cases
 	winfThresh = int64(2 * time.Second)
+	// Checksum size for hash for msg records.
+	recordHashSize = 8
 )
 
 func newFileStore(fcfg FileStoreConfig, cfg StreamConfig) (*fileStore, error) {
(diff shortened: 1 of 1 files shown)


server/filestore.go
```diff
@@ -4055,23 +4090,39 @@ func (mb *msgBlock) indexCacheBuf(buf []byte) error {
 		dlen := int(rl) - msgHdrSize
 
 		// Do some quick sanity checks here.
-		if dlen < 0 || int(slen) > dlen || dlen > int(rl) || index+rl > lbuf || rl > 32*1024*1024 {
+		if dlen < 0 || int(slen) > (dlen-8) || dlen > int(rl) || index+rl > lbuf || rl > rlBadThresh {
```
neilalexander, line 4093: Might be worth commenting what the `8` is for, otherwise it's just a magic number

neilalexander (approved):
LGTM

server/filestore.go
derekcollison: Will add in const.

PR #4345, opened 2023-07-28
Prevent configuring `first_seq` on mirrors

neilalexander:
This prevents configurations where mirrors are mixed with `first_seq`s of greater than zero.


derekcollison (approved):
LGTM

PR #4346, opened 2023-07-28
[FIXED] Do not panic when re-calculating first sequence for SimpleState when fseq moves ahead of old first.

derekcollison:
When a lazy simple state has an outdated first that needs to be updated, if fseq had moved past it would panic.
This was not common but with latest fix prior in can become more common, hence why it showed up.


wallyqs (approved):
LGTM

PR #4347, opened 2023-07-28
config: make parsing configurations without usable values invalid

wallyqs:
This makes configuration files that are empty, or read and processed by the parser but with no detected values now return an error.


derekcollison (approved):
LGTM

PR #4348, opened 2023-07-30
[FIXED] Seqset encode bug that could cause bad stream state snapshots

derekcollison:
This would cause decode to spin and lock up the server.

This fixes the encode bug (a delete and clear node bug in avl/seqset). Also addressed receiving a bad snapshot, etc.


Jarema (approved):
LGTM

Issue #3244, opened 2022-07-06
Support MQTT QoS=2


robinsmidsrod:
## Feature Request

I'm using NATS as an MQTT server together with Home Assistant, and the HASS.Agent Windows companion application. The HASS.Agent application uses QoS=2 when it wants to enable commands to run on the Windows machine from Home Assistant. When this is enabled, the MQTT connection is disconnected and a message about QoS=2 not supported is logged.

#### Proposed Change:

Support MQTT QoS=2.

#### Who Benefits From The Change(s)?
People using Home Assistant and HASS.Agent with NATS MQTT service.

#### Alternative Approaches
None at the moment, because HASS.Agent doesn't support setting QoS level.

kozlovic:
Thank you for your interest. Unfortunately, there is no plan to implement QoS2 in NATS Server at this stage.

robinsmidsrod:
@kozlovic You shouldn't by any chance have a link to a document that describes why you have no current plans of supporting it? I'm going to assume it's quite tricky with how MQTT is implemented in NATS, but it could also be as trivial that you don't have time to prioritize it.

kozlovic:
@robinsmidsrod There is no document per-se. I have written the MQTT code in the NATS Server and I am concerned with the added complexity of QoS2 and that we don't have much time with all other priorities. I will, however, re-open this issue and leave it as enhancement request in case the community wants to contribute with a nice PR ;-)

pablitovicente:
I think there could be many of us wanting to use NATS for it's features and also as a full MQTT broker. What kind of help would be needed for this effort?

@kozlovic still haven't dig into the NATS code but do you consider it would be doable to support the full MQTT spec?

TomasVojacek:
It seem that [robinsmidsrod](https://github.com/robinsmidsrod) does not need QoS2. Problem is if message arrive with qos2 connection is closed. Would it be possible add configuration option downGradeQos2ToQos1 that will allow treat QoS2 messages like qos1 messages?

robinsmidsrod:
@TomasVojacek Your observation is correct. I just need the MQTT broker (NATS) to stay connected, even if it doesn't support QoS=2. If there is a "force downgrade" option in the protocol that HASS.Agent will be able to tolerate and continue working with then that is a workaround for me. But if HASS.Agent ignores that and expects QoS=2 behavior then we're no further in terms of solving my integration compatibility issue. I think QoS=2 support in NATS would be awesome, though.

[...]

PR #4349, opened 2023-07-31
[Added] MQTT: QoS2 support

levb:
@derekcollison @neilalexander @kozlovic 

#### Summary

Adds MQTT QoS2 support

 - [X] Resolves https://github.com/nats-io/nats-server/issues/3244
 - [X] Tests added

### DIFF
--- a/server/client.go
+++ b/server/client.go
@@ -4223,7 +4223,7 @@ func (c *client) processServiceImport(si *serviceImport, acc *Account, msg []byt
 	c.pa.reply = nrr
 
 	if changed && c.isMqtt() && c.pa.hdr > 0 {
-		c.srv.mqttStoreQoS1MsgForAccountOnNewSubject(c.pa.hdr, msg, siAcc.GetName(), to)
+		c.srv.mqttStoreQoSMsgForAccountOnNewSubject(c.pa.hdr, msg, siAcc.GetName(), to)
 	}
 
 	// FIXME(dlc) - Do L1 cache trick like normal client?
--- a/server/mqtt.go
+++ b/server/mqtt.go
@@ -68,7 +68,8 @@ const (
 	mqttPubFlagRetain = byte(0x01)
 	mqttPubFlagQoS    = byte(0x06)
 	mqttPubFlagDup    = byte(0x08)
-	mqttPubQos1       = byte(0x2) // 1 << 1
+	mqttPubQos1       = byte(0x1 << 1)
+	mqttPubQoS2       = byte(0x2 << 1)
 
 	// Subscribe flags
 	mqttSubscribeFlags = byte(0x2)
@@ -104,28 +105,34 @@ const (
 	// MQTT subscriptions to start with this.
 	mqttSubPrefix = "$MQTT.sub."
 
-	// MQTT Stream names prefix.
-	mqttStreamNamePrefix = "$MQTT_"
-
 	// Stream name for MQTT messages on a given account
-	mqttStreamName          = mqttStreamNamePrefix + "msgs"
+	mqttStreamName          = "$MQTT_msgs"
 	mqttStreamSubjectPrefix = "$MQTT.msgs."
 
 	// Stream name for MQTT retained messages on a given account
-	mqttRetainedMsgsStreamName    = mqttStreamNamePrefix + "rmsgs"
+	mqttRetainedMsgsStreamName    = "$MQTT_rmsgs"
 	mqttRetainedMsgsStreamSubject = "$MQTT.rmsgs."
 
 	// Stream name for MQTT sessions on a given account
-	mqttSessStreamName          = mqttStreamNamePrefix + "sess"
+	mqttSessStreamName          = "$MQTT_sess"
 	mqttSessStreamSubjectPrefix = "$MQTT.sess."
 
 	// Stream name prefix for MQTT sessions on a given account
-	mqttSessionsStreamNamePrefix = mqttStreamNamePrefix + "sess_"
+	mqttSessionsStreamNamePrefix = "$MQTT_sess_"
 
-	// Normally, MQTT server should not redeliver QoS 1 messages to clients,
-	// except after client reconnects. However, NATS Server will redeliver
-	// unacknowledged messages after this default interval. This can be
-	// changed with the server.Options.MQTT.AckWait option.
+	// Stream name and subject for incoming MQTT QoS2 messages
+	mqttQoS2IncomingMsgsStreamName          = "$MQTT_qos2in"
+	mqttQoS2IncomingMsgsStreamSubjectPrefix = "$MQTT.qos2.in."
+
+	// Stream name and subjects for outgoing MQTT QoS2 PUBREL messages
+	mqttQoS2PubRelStreamName            = "$MQTT_qos2out"
+	mqttQoS2PubRelStoredSubjectPrefix   = "$MQTT.qos2.out."
+	mqttQoS2PubRelDeliverySubjectPrefix = "$MQTT.qos2.delivery."
+
+	// As per spec, MQTT server may not redeliver QoS 1 and 2 messages to
+	// clients, except after client reconnects. However, NATS Server will
+	// redeliver unacknowledged messages after this default interval. This can
+	// be changed with the server.Options.MQTT.AckWait option.
 	mqttDefaultAckWait = 30 * time.Second
 
 	// This is the default for the outstanding number of pending QoS 1
@@ -159,9 +166,6 @@ const (
 	mqttJSARetainedMsgDel = "RD"
 	mqttJSAStreamNames    = "SN"
 
-	// Name of the header key added to NATS message to carry mqtt PUBLISH information
-	mqttNatsHeader = "Nmqtt-Pub"
-
 	// This is how long to keep a client in the flappers map before closing the
 	// connection. This prevent quick reconnect from those clients that keep
 	// wanting to connect with a client ID already in use.
@@ -181,7 +185,6 @@ var (
 	mqttPingResponse  = []byte{mqttPacketPingResp, 0x0}
 	mqttProtoName     = []byte("MQTT")
 	mqttOldProtoName  = []byte("MQIsdp")
-	mqttNatsHeaderB   = []byte(mqttNatsHeader)
 	mqttSessJailDur   = mqttSessFlappingJailDur
 	mqttFlapCleanItvl = mqttSessFlappingCleanupInterval
 	mqttJSAPITimeout  = 4 * time.Second
@@ -206,6 +209,7 @@ var (
 	errMQTTTopicIsEmpty             = errors.New("topic cannot be empty")
 	errMQTTPacketIdentifierIsZero   = errors.New("packet identifier cannot be 0")
 	errMQTTUnsupportedCharacters    = errors.New("characters ' ' and '.' not supported for MQTT topics")
+	errMQTTInvalidSession           = errors.New("invalid MQTT session")
 )
 
 type srvMQTT struct {
@@ -262,17 +266,36 @@ type mqttRetMsgDel struct {
 }
 
 type mqttSession struct {
-	mu       sync.Mutex
-	id       string // client ID
-	idHash   string // client ID hash
-	c        *client
-	jsa      *mqttJSA
-	subs     map[string]byte
-	cons     map[string]*ConsumerConfig
-	seq      uint64
-	pending  map[uint16]*mqttPending      // Key is the PUBLISH packet identifier sent to client and maps to a mqttPending record
-	cpending map[string]map[uint64]uint16 // For each JS consumer, the key is the stream sequence and maps to the PUBLISH packet identifier
-	ppi      uint16                       // publish packet identifier
+	mu                     sync.Mutex
+	id                     string // client ID
+	idHash                 string // client ID hash
+	c                      *client
+	jsa                    *mqttJSA
+	subs                   map[string]byte // Key is MQTT SUBSCRIBE filter, value is the subscription QoS
+	cons                   map[string]*ConsumerConfig
+	pubRelConsumer         *ConsumerConfig
+	pubRelSubscribed       bool
+	pubRelDeliverySubject  string
+	pubRelDeliverySubjectB []byte
+	pubRelSubject          string
+	seq                    uint64
+
+	// pendingPublish maps packet identifiers (PI) to JetStream ACK subjects for
+	// QoS1 and 2 PUBLISH messages pending delivery to the session's client.
+	pendingPublish map[uint16]*mqttPending
+
+	// pendingPubRel maps PIs to JetStream ACK subjects for QoS2 PUBREL
+	// messages pending delivery to the session's client.
+	pendingPubRel map[uint16]*mqttPending
+
+	// cpending maps delivery attempts (that come with a JS ACK subject) to
+	// existing PIs.
+	cpending map[string]map[uint64]uint16 // composite key: jsDur, sseq
+
+	// "Last used" publish packet identifier (PI). starting point searching for the next available.
+	last_pi uint16
+
+	// Maximum number of pending acks for this session.
 	maxp     uint16
 	tmaxack  int
 	clean    bool
@@ -285,6 +308,7 @@ type mqttPersistedSession struct {
 	Clean  bool                       `json:"clean,omitempty"`
 	Subs   map[string]byte            `json:"subs,omitempty"`
 	Cons   map[string]*ConsumerConfig `json:"cons,omitempty"`
+	PubRel *ConsumerConfig            `json:"pubrel,omitempty"`
 }
 
 type mqttRetainedMsg struct {
@@ -322,9 +346,9 @@ type mqtt struct {
 }
 
 type mqttPending struct {
-	sseq    uint64 // stream sequence
-	subject string // the ACK subject to send the ack to
-	jsDur   string // JS durable name
+	sseq         uint64 // stream sequence
+	jsAckSubject string // the ACK subject to send the ack to
+	jsDur        string // JS durable name
 }
 
 type mqttConnectProto struct {
@@ -374,6 +398,28 @@ type mqttPublish struct {
 	sz      int
 	pi      uint16
 	flags   byte
+}
+
+// When we re-encode incoming MQTT PUBLISH messages for NATS delivery, we add
+// the following headers:
+//   - "Nmqtt-Pub" (*always) indicates that the message originated from MQTT, and
+//     contains the original message QoS.
+//   - "Nmqtt-Subject" contains the original MQTT subject from mqttParsePub.
+//   - "Nmqtt-Mapped" contains the mapping during mqttParsePub.
+//
+// When we submit a PUBREL for delivery, we add a "Nmqtt-PubRel" header that
+// contains the PI.
+const (
+	mqttNatsHeader        = "Nmqtt-Pub"
+	mqttNatsPubRelHeader  = "Nmqtt-PubRel"
+	mqttNatsHeaderSubject = "Nmqtt-Subject"
+	mqttNatsHeaderMapped  = "Nmqtt-Mapped"
+)
+
+type mqttParsedPublishNATSHeader struct {
+	qos     byte
+	subject []byte
+	mapped  []byte
 }
 
 func (s *Server) startMQTT() {
@@ -667,6 +713,39 @@ func (c *client) mqttParse(buf []byte) error {
 		}
 
 		switch pt {
+		// Packets that we receive back when we act as the "sender": PUBACK,
+		// PUBREC, PUBCOMP.
+		case mqttPacketPubAck:
+			var pi uint16
+			pi, err = mqttParsePIPacket(r, pl)
+			if trace {
+				c.traceInOp("PUBACK", errOrTrace(err, fmt.Sprintf("pi=%v", pi)))
+			}
+			if err == nil {
+				err = c.mqttProcessPubAck(pi)
+			}
+
+		case mqttPacketPubRec:
+			var pi uint16
+			pi, err = mqttParsePIPacket(r, pl)
+			if trace {
+				c.traceInOp("PUBREC", errOrTrace(err, fmt.Sprintf("pi=%v", pi)))
+			}
+			if err == nil {
+				err = c.mqttProcessPubRec(pi)
+			}
+
+		case mqttPacketPubComp:
+			var pi uint16
+			pi, err = mqttParsePIPacket(r, pl)
+			if trace {
+				c.traceInOp("PUBCOMP", errOrTrace(err, fmt.Sprintf("pi=%v", pi)))
+			}
+			if err == nil {
+				c.mqttProcessPubComp(pi)
+			}
+
+		// Packets where we act as the "receiver": PUBLISH, PUBREL, SUBSCRIBE, UNSUBSCRIBE.
 		case mqttPacketPub:
 			pp := c.mqtt.pp
 			pp.flags = b & mqttPacketFlagMask
@@ -678,23 +757,19 @@ func (c *client) mqttParse(buf []byte) error {
 				}
 			}
 			if err == nil {
-				err = s.mqttProcessPub(c, pp)
+				err = s.mqttProcessPub(c, pp, trace)
 			}
-			if err == nil && pp.pi > 0 {
-				c.mqttEnqueuePubAck(pp.pi)
-				if trace {
-					c.traceOutOp("PUBACK", []byte(fmt.Sprintf("pi=%v", pp.pi)))
-				}
-			}
-		case mqttPacketPubAck:
+
+		case mqttPacketPubRel:
 			var pi uint16
-			pi, err = mqttParsePubAck(r, pl)
+			pi, err = mqttParsePIPacket(r, pl)
 			if trace {
-				c.traceInOp("PUBACK", errOrTrace(err, fmt.Sprintf("pi=%v", pi)))
+				c.traceInOp("PUBREL", errOrTrace(err, fmt.Sprintf("pi=%v", pi)))
 			}
 			if err == nil {
-				c.mqttProcessPubAck(pi)
+				err = s.mqttProcessPubRel(c, pi, trace)
 			}
+
 		case mqttPacketSub:
 			var pi uint16 // packet identifier
 			var filters []*mqttFilter
@@ -713,6 +788,7 @@ func (c *client) mqttParse(buf []byte) error {
 				c.mqttEnqueueSubAck(pi, filters)
 				c.mqttSendRetainedMsgsToNewSubs(subs)
 			}
+
 		case mqttPacketUnsub:
 			var pi uint16 // packet identifier
 			var filters []*mqttFilter
@@ -729,6 +805,8 @@ func (c *client) mqttParse(buf []byte) error {
 			if err == nil {
 				c.mqttEnqueueUnsubAck(pi)
 			}
+
+		// Packets that we get both as a receiver and sender: PING, CONNECT, DISCONNECT
 		case mqttPacketPing:
 			if trace {
 				c.traceInOp("PINGREQ", nil)
@@ -737,6 +815,7 @@ func (c *client) mqttParse(buf []byte) error {
 			if trace {
 				c.traceOutOp("PINGRESP", nil)
 			}
+
 		case mqttPacketConnect:
 			// It is an error to receive a second connect packet
 			if connected {
@@ -770,6 +849,7 @@ func (c *client) mqttParse(buf []byte) error {
 					rd = cp.rd
 				}
 			}
+
 		case mqttPacketDisconnect:
 			if trace {
 				c.traceInOp("DISCONNECT", nil)
@@ -784,8 +864,7 @@ func (c *client) mqttParse(buf []byte) error {
 			s.mqttHandleClosedClient(c)
 			c.closeConnection(ClientClosed)
 			return nil
-		case mqttPacketPubRec, mqttPacketPubRel, mqttPacketPubComp:
-			err = fmt.Errorf("protocol %d not supported", pt>>4)
+
 		default:
 			err = fmt.Errorf("received unknown packet type %d", pt>>4)
 		}
@@ -900,15 +979,12 @@ func (s *Server) mqttGetJSAForAccount(acc string) *mqttJSA {
 	return jsa
 }
 
-func (s *Server) mqttStoreQoS1MsgForAccountOnNewSubject(hdr int, msg []byte, acc, subject string) {
+func (s *Server) mqttStoreQoSMsgForAccountOnNewSubject(hdr int, msg []byte, acc, subject string) {
 	if s == nil || hdr <= 0 {
 		return
 	}
-	nhv := getHeader(mqttNatsHeader, msg[:hdr])
-	if len(nhv) < 1 {
-		return
-	}
-	if qos := nhv[0] - '0'; qos != 1 {
+	h := mqttParsePublishNATSHeader(msg[:hdr])
+	if h == nil || h.qos == 0 {
 		return
 	}
 	jsa := s.mqttGetJSAForAccount(acc)
@@ -916,6 +992,35 @@ func (s *Server) mqttStoreQoS1MsgForAccountOnNewSubject(hdr int, msg []byte, acc
 		return
 	}
 	jsa.storeMsg(mqttStreamSubjectPrefix+subject, hdr, msg)
+}
+
+func mqttParsePublishNATSHeader(headerBytes []byte) *mqttParsedPublishNATSHeader {
+	if len(headerBytes) == 0 {
+		return nil
+	}
+
+	pubValue := getHeader(mqttNatsHeader, headerBytes)
+	if len(pubValue) == 0 {
+		return nil
+	}
+	return &mqttParsedPublishNATSHeader{
+		qos:     pubValue[0] - '0',
+		subject: getHeader(mqttNatsHeaderSubject, headerBytes),
+		mapped:  getHeader(mqttNatsHeaderMapped, headerBytes),
+	}
+}
+
+func mqttParsePubRelNATSHeader(headerBytes []byte) uint16 {
+	if len(headerBytes) == 0 {
+		return 0
+	}
+
+	pubrelValue := getHeader(mqttNatsPubRelHeader, headerBytes)
+	if len(pubrelValue) == 0 {
+		return 0
+	}
+	pi, _ := strconv.Atoi(string(pubrelValue))
+	return uint16(pi)
 }
 
 // Returns the MQTT sessions manager for a given account.
@@ -1146,6 +1251,46 @@ func (s *Server) mqttCreateAccountSessionManager(acc *Account, quitCh chan struc
 		}
 	}
 
+	if si, err := lookupStream(mqttQoS2IncomingMsgsStreamName, "QoS2 incoming messages"); err != nil {
+		return nil, err
+	} else if si == nil {
+		// Create the stream for the incoming QoS2 messages that have not been
+		// PUBREL-ed by the sender. Subject is
+		// "$MQTT.qos2.<session>.<PI>", the .PI is to achieve exactly
+		// once for each PI.
+		cfg := &StreamConfig{
+			Name:          mqttQoS2IncomingMsgsStreamName,
+			Subjects:      []string{mqttQoS2IncomingMsgsStreamSubjectPrefix + ">"},
+			Storage:       FileStorage,
+			Retention:     LimitsPolicy,
+			Discard:       DiscardNew,
+			MaxMsgsPer:    1,
+			DiscardNewPer: true,
+			Replicas:      replicas,
+		}
+		if _, _, err := jsa.createStream(cfg); isErrorOtherThan(err, JSStreamNameExistErr) {
+			return nil, fmt.Errorf("create QoS2 incoming messages stream for account %q: %v", accName, err)
+		}
+	}
+
+	if si, err := lookupStream(mqttQoS2PubRelStreamName, "QoS2 outgoing PUBREL"); err != nil {
+		return nil, err
+	} else if si == nil {
+		// Create the stream for the incoming QoS2 messages that have not been
+		// PUBREL-ed by the sender. NATS messages are submitted as
+		// "$MQTT.pubrel.<session hash>"
+		cfg := &StreamConfig{
+			Name:      mqttQoS2PubRelStreamName,
+			Subjects:  []string{mqttQoS2PubRelStoredSubjectPrefix + ">"},
+			Storage:   FileStorage,
+			Retention: InterestPolicy,
+			Replicas:  replicas,
+		}
+		if _, _, err := jsa.createStream(cfg); isErrorOtherThan(err, JSStreamNameExistErr) {
+			return nil, fmt.Errorf("create QoS2 outgoing PUBREL stream for account %q: %v", accName, err)
+		}
+	}
+
 	// This is the only case where we need "si" after lookup/create
 	si, err := lookupStream(mqttRetainedMsgsStreamName, "retained messages")
 	if err != nil {
@@ -1346,6 +1491,16 @@ func (jsa *mqttJSA) newRequestEx(kind, subject string, hdr int, msg []byte, time
 		return nil, fmt.Errorf("timeout for request type %q on %q (reply=%q)", kind, subject, reply)
 	}
 	return i, nil
+}
+
+func (jsa *mqttJSA) sendAck(ackSubject string) {
+	if ackSubject == _EMPTY_ {
+		return
+	}
+
+	// We pass -1 for the hdr so that the send loop does not need to
+	// add the "client info" header. This is not a JS API request per se.
+	jsa.sendq.push(&mqttJSPubMsg{subj: ackSubject, hdr: -1})
 }
 
 func (jsa *mqttJSA) createConsumer(cfg *CreateConsumerRequest) (*JSApiConsumerCreateResponse, error) {
@@ -2068,8 +2223,8 @@ func (as *mqttAccountSessionManager) processSubs(sess *mqttSession, c *client,
 	var err error
 	subs := make([]*subscription, 0, len(filters))
 	for _, f := range filters {
-		if f.qos > 1 {
-			f.qos = 1
+		if f.qos > 2 {
+			f.qos = 2
 		}
 		subject := f.filter
 		sid := subject
@@ -2085,11 +2240,15 @@ func (as *mqttAccountSessionManager) processSubs(sess *mqttSession, c *client,
 		// Note that if a subscription already exists on this subject,
 		// the existing sub is returned. Need to update the qos.
 		asAndSessLock()
-		sub, err := c.processSub([]byte(subject), nil, []byte(sid), mqttDeliverMsgCbQos0, false)
+		sub, err := c.processSub([]byte(subject), nil, []byte(sid), mqttDeliverMsgCbQoS0, false)
 		if err == nil {
 			setupSub(sub, f.qos)
 		}
+		if f.qos == 2 {
+			err = sess.ensurePubRelConsumerSubscription(c)
+		}
 		asAndSessUnlock()
+
 		if err == nil {
 			// This will create (if not already exist) a JS consumer for subscriptions
 			// of QoS >= 1. But if a JS consumer already exists and the subscription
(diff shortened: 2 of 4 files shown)


server/mqtt.go
```diff
@@ -3144,7 +3154,52 @@ func mqttPubTrace(pp *mqttPublish) string {
 //
 // Runs from the client's readLoop.
 // No lock held on entry.
-func (s *Server) mqttProcessPub(c *client, pp *mqttPublish) error {
+func (s *Server) mqttProcessPub(c *client, pp *mqttPublish, trace bool) error {
+	qos := mqttGetQoS(pp.flags)
+
```
kozlovic, line 3159: nit: you could capture `sess := c.mqtt.sess` to avoid repeating `c.mqtt.sess` (which is being used quite a bit in this function).

server/mqtt.go
```diff
@@ -3144,7 +3154,52 @@ func mqttPubTrace(pp *mqttPublish) string {
...
+	if qos == 2 {
+		c.mqtt.sess.mu.Lock()
+		if len(c.mqtt.sess.pendingQOS2Received) > 0 {
+			_, drop = c.mqtt.sess.pendingQOS2Received[pp.pi]
+		}
+		c.mqtt.sess.mu.Unlock()
+	}
+
+	if !drop {
+		if err := s.mqttForwardPub(c, pp); err != nil {
```
kozlovic, line 3171: If I read this right, for QoS2, you seem to store message (for the first PUBLISH received), but that will also initiate delivery, then send the PUBREC then "store" the PI.  I write "store" because this is just in memory and not really persisted anywhere. 
The spec seem to indicate that there are methods A or B, and this looks to me a mix of the two. (see figure 4.3 [here](https://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html#_Toc398718102)).

Since we don't really "store" the PI, then if the server were to crash after "mqttForwardPub" (which stores and initiates delivery), but before sending the PUBREC, then if the sender were to call again PUBLISH with same PI, we would violate I believe this rule: `Until it has received the corresponding PUBREL packet, the Receiver MUST acknowledge any subsequent PUBLISH packet with the same Packet Identifier by sending a PUBREC. It MUST NOT cause duplicate messages to be delivered to any onward recipients in this case`.

server/mqtt.go
```diff
@@ -3451,6 +3506,40 @@ func (c *client) mqttProcessPubAck(pi uint16) {
...
+	}
+
+	isPending := false
+	sess.mu.Lock()
+	if sess.c != c {
+		sess.mu.Unlock()
+		return
+	}
+
+	if len(sess.pendingQOS2Received) > 0 {
```
kozlovic, line 3527: Again, since this is not persisted, if the server were to crash after sending the PUBREC and the sender sends the PUBREL, we would not reply with a PUBCOMP here (as we should). So I wonder if the send of the PUBCOMP should not be done regardless if the PI is found to be pending or not (if that is the case, then it could be simplified to a delete (not lookup and delete).

server/mqtt_test.go
```diff
@@ -2100,7 +2086,7 @@ func TestMQTTParsePub(t *testing.T) {
 	}
 }
 
-func TestMQTTParsePubAck(t *testing.T) {
```
kozlovic, line 2103: You may want to modify the test above (`TestMQTTParsePub`) to update the `flags` value to `0x6` and the expected error message for the `qos not supported` test case, from `not supported` to your new error message `is invalid in MQTT`.

kozlovic (changes requested):

server/mqtt.go
levb: Right. For PUBACKs, Is this accomplished (in reverse) with `cpending`? I have not understood that part of the code yet, nor the session persistence/management code yet.

server/client.go
```diff
@@ -4222,6 +4222,7 @@ func (c *client) processServiceImport(si *serviceImport, acc *Account, msg []byt
 	c.pa.reply = nrr
 
 	if changed && c.isMqtt() && c.pa.hdr > 0 {
+		// <>/<> handle QoS2
```
levb, line 4225: Need to understand the purpose of this, but may be no change is needed since QOS2 messages themselves (as opposed to their PUBRELs) are queued up jus like before.

server/mqtt.go
```diff
@@ -35,6 +35,19 @@ import (
 
 // References to "spec" here is from https://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.pdf
 
+// # QoS support, receiving messages.
+//
+//	- QoS 0: A message on a subject foo can be
+//
+// ### Streams
```
levb, line 42: WIP, finish

server/mqtt.go
```diff
@@ -376,6 +396,32 @@ type mqttPublish struct {
...
 }
 
+// Names of the header keys added to NATS message to carry mqtt PUBLISH
+// information.
+//   - "Nmqtt-Pub" indicates that the message originated from MQTT, and
+//     contains the original message QoS.
+//   - "Nmqtt-Subject" contains the original MQTT subject from mqttParsePub.
+//   - "Nmqtt-Mapped" contains the mapping during mqttParsePub.
+const (
+	mqttNatsHeader        = "Nmqtt-Pub"
```
levb, line 406: Need to encode `retain` and use it once initiating delivery (for QoS2)?

server/mqtt.go
```diff
@@ -904,11 +980,8 @@ func (s *Server) mqttStoreQoS1MsgForAccountOnNewSubject(hdr int, msg []byte, acc
 	if s == nil || hdr <= 0 {
 		return
 	}
-	nhv := getHeader(mqttNatsHeader, msg[:hdr])
-	if len(nhv) < 1 {
-		return
-	}
-	if qos := nhv[0] - '0'; qos != 1 {
+	h := mqttParseInternalNATSHeader(msg[:hdr])
```
levb, line 983: What is the use-case of `mqttStoreQoS1MsgForAccountOnNewSubject`?

server/mqtt.go
```diff
@@ -3835,6 +4264,86 @@ func (sess *mqttSession) cleanupFailedSub(c *client, sub *subscription, cc *Cons
...
+	if maxAckPending == 0 {
+		maxAckPending = mqttDefaultMaxAckPending
+	}
+
+	cc = &ConsumerConfig{
+		DeliverSubject: deliverSubj,
+		Durable:        sess.idHash + "_pubrel",
+		AckPolicy:      AckExplicit,
+		DeliverPolicy:  DeliverNew,
+		FilterSubject:  sess.pubRelSubject(),
```
levb, line 4318: What will make this globally unique for this session?

server/mqtt.go
```diff
@@ -3914,7 +4423,6 @@ func (sess *mqttSession) processJSConsumer(c *client, subject, sid string,
 			c.Errorf("Unable to add JetStream consumer for subscription on %q: err=%v", subject, err)
 			return nil, nil, err
 		}
-		sess.tmaxack += maxAckPending
```
levb, line 3917: Hmm... not sure why it is taken out, review

server/mqtt_test.go
```diff
@@ -1699,39 +1699,6 @@ func TestMQTTDontSetPinger(t *testing.T) {
 	testMQTTPublish(t, mc, r, 0, false, false, "foo", 0, []byte("msg"))
 }
 
-func TestMQTTUnsupportedPackets(t *testing.T) {
```
levb, line 1702: Will put back, corrected

server/mqtt.go
levb: Re-did using subject-based deduplication. Resolving the comment as no longer applicable.

server/mqtt.go
levb: ditto

server/mqtt.go
```diff
@@ -3140,44 +3298,231 @@ func mqttPubTrace(pp *mqttPublish) string {
...
+	// Best attempt to delete the message from the QoS2 stream.
+	_ = asm.jsa.deleteMsg(mqttQOS2IncomingMsgsStreamName, stored.Sequence, true)
+
+	// only MQTT QoS2 messages should be here, and they must have a subject.
+	h := mqttParseInternalNATSHeader(stored.Header)
+	if h == nil || h.packetType != mqttPacketPub || h.qos != 2 || len(h.subject) == 0 {
+		return
+	}
+
+	// msgToSend := append(stored.Header, stored.Data...)
```
levb, line 3511: remove

server/mqtt.go
```diff
@@ -3140,44 +3298,231 @@ func mqttPubTrace(pp *mqttPublish) string {
...
+		subject: h.subject,
+		mapped:  h.mapped,
+		msg:     stored.Data,
+		sz:      -1, // <>/<> TODO what should sz be?
+		pi:      pi,
+		// <>/<> TODO store and restore retain flag
+		flags: h.qos << 1,
+	}
+
+	_ = s.mqttInitiateMsgDelivery(c, pp)
```
levb, line 3523: don't do the `_`s, just ignore.

server/mqtt.go
```diff
@@ -3835,6 +4262,86 @@ func (sess *mqttSession) cleanupFailedSub(c *client, sub *subscription, cc *Cons
...
+	if cQOS2 > 0 && sess.pubrelConsumer == nil {
+		create = true
+	}
+	cc = sess.pubrelConsumer
+	sess.mu.Unlock()
+
+	deliverSubj := sess.pubRelDeliverySubject()
+	deliverSubjB := []byte(deliverSubj)
+	if ccDelete != nil {
+		sess.deleteConsumer(ccDelete)
```
levb, line 4294: check the order, session clear

levb: split `process` into a create and delete (only in clear session).

server/mqtt.go
```diff
@@ -3140,44 +3298,231 @@ func mqttPubTrace(pp *mqttPublish) string {
...
+	// <>/<> what is addToPCD?
+	// We need to call flushClients now since this we may have called c.addToPCD
+	// with destination clients (possibly a route). Without calling flushClients
+	// the following call may then be stuck waiting for a reply that may never
+	// come because the destination is not flushed (due to c.out.fsp > 0,
+	// see addToPCD and writeLoop for details).
+	c.flushClients(0)
+
+	// Store the message in JetStream as "$MQTT.msgs.<delivery-subject>"
+	_, err := c.mqtt.sess.jsa.storeMsg(mqttStreamSubjectPrefix+string(pp.subject), headerLen, natsMsg)
```
kozlovic, line 3450: I checking to see if there is not cases where we would come with c.pa.subject being something and that being changed in processInboundClientMsg. If that is the case, we would store here with the wrong subject (original code referenced c.pa.subject).
We do possibly change c.pa.subject in case of gateways, so again, not sure if the refactoring you did is safe...

server/mqtt.go
```diff
@@ -3140,44 +3298,231 @@ func mqttPubTrace(pp *mqttPublish) string {
...
 	return err
 }
 
+func (s *Server) mqttStoreQOS2Msg(c *client, pp *mqttPublish) error {
+	// `true` means encode the MQTT PUBLISH packet in the NATS message header.
+	natsMsg, headerLen := mqttNewDeliverableMessage(pp, true)
+
+	// Do not broadcast the message until it has been deduplicated and
+	// released by the sender. Instead store this QoS2 message as
+	// "$MQTT.qos2.<domain>.<account-name>.<client-id>.<PI>". If the message
```
kozlovic, line 3460: Domain should be removed, and all this runs under a given account already, so really client-id and PI.

server/mqtt.go
```diff
@@ -3140,44 +3298,231 @@ func mqttPubTrace(pp *mqttPublish) string {
...
+	clientID := c.mqtt.cid
+
+	c.mu.Lock()
+	acc := c.acc
+	c.mu.Unlock()
+	accName := acc.GetName()
+
+	return strings.Join([]string{
+		mqttQOS2IncomingMsgsStreamSubjectPrefixNoSep,
+		accName,
```
kozlovic, line 3479: Don't think we need the account since all MQTT streams are per account already.

kozlovic (changes requested):
Temporary review (to not risk losing pending comments ;-))

server/mqtt.go
```diff
@@ -3679,10 +4065,12 @@ func mqttDeliverMsgCbQos1(sub *subscription, pc *client, _ *Account, subject, re
 		return
 	}
 
-	// This is a QoS1 message for a QoS1 subscription, so get the pi and keep
-	// track of ack subject.
-	pQoS := byte(1)
-	pi, dup := sess.trackPending(pQoS, reply, sub)
+	qos := h.qos
+	// <>/<>  TODO do I need to lock the sub? Can the qos change?
```
kozlovic, line 4069: You are protected by the session's lock.

server/mqtt.go
```diff
@@ -3703,8 +4091,31 @@ func mqttDeliverMsgCbQos1(sub *subscription, pc *client, _ *Account, subject, re
...
+	// This is immutable
+	sess := cc.mqtt.sess
+
+	sess.mu.Lock()
+	if sess.c != cc {
+		sess.mu.Unlock()
+		return
+	}
+
+	pc.mqttEnqueuePubRelTo(cc, h.pi)
```
kozlovic, line 4118: Unlock session, but then here you should have say `cc.mqttEnqueuePubRel(h.pi)`. We don't need addToPCD and therefore we don't need `pc` here. See comment in existing mqttEnqueuePubRel for details.

server/mqtt.go
```diff
@@ -3754,12 +4168,25 @@ func (c *client) mqttDeliver(cc *client, sub *subscription, pi uint16, dup, reta
 		}
 		cc.traceOutOp("PUBLISH", []byte(mqttPubTrace(&pp)))
 	}
+}
+
+func (c *client) mqttEnqueuePubRelTo(cc *client, pi uint16) {
```
kozlovic, line 4173: Remove `cc *client`.

server/mqtt.go
kozlovic: Agreed

kozlovic (changes requested):
Finished my review but there are lots of tests that are failing, so clearly some issues. I also need to check that when we "store" (a QoS1/2) a message, I believe that the delivery to JS consumer is automatically initiated, which I think would break the MQTT workflow with PUBREC/REL/COMP. Also note that once a JS message is stored (and delivered to JS consumer), it may be redelivered automatically by JetStream when no ack is received. Again, not sure if that throws a wrench into the compliance.

server/mqtt.go
```diff
@@ -3140,44 +3298,231 @@ func mqttPubTrace(pp *mqttPublish) string {
...
+	}
+
+	// <>/<> what is addToPCD?
+	// We need to call flushClients now since this we may have called c.addToPCD
+	// with destination clients (possibly a route). Without calling flushClients
+	// the following call may then be stuck waiting for a reply that may never
+	// come because the destination is not flushed (due to c.out.fsp > 0,
+	// see addToPCD and writeLoop for details).
+	c.flushClients(0)
+
```
kozlovic, line 3448: You should return here if `mqttGetQoS(pp.flags) == 0` otherwise you would be storing QoS0 messages!

server/mqtt.go
```diff
@@ -3703,8 +4091,31 @@ func mqttDeliverMsgCbQos1(sub *subscription, pc *client, _ *Account, subject, re
 	}
 
 	topic := natsSubjectToMQTTTopic(strippedSubj)
+	retained := false // <>/<>
```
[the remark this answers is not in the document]
kozlovic: Actually I have re-read the spec and I think that are some issues with the value of the RETAIN flag sent by server to client (at least on v3.1.1), but in that specific case, I believe it will have to be set to 0 (so false here would be correct). I need to spend a bit more time on that, create/update tests and fix the issue, but in a separate PR.

server/mqtt.go
```diff
@@ -1135,8 +1235,11 @@ func (s *Server) mqttCreateAccountSessionManager(acc *Account, quitCh chan struc
 	} else if si == nil {
 		// Create the stream for the messages.
 		cfg := &StreamConfig{
-			Name:      mqttStreamName,
-			Subjects:  []string{mqttStreamSubjectPrefix + ">"},
+			Name: mqttStreamName,
+			Subjects: []string{
+				mqttStreamSubjectPrefix + ">",
+				mqttQOS2PubRelStoredSubjectPrefix + ">",
```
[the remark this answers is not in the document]

server/mqtt.go
levb: Oops

server/mqtt.go
```diff
@@ -3595,12 +3989,12 @@ func mqttDeliverMsgCbQos0(sub *subscription, pc *client, _ *Account, subject, re
...
-	// This is an MQTT publisher directly connected to this server.
 	if pc.isMqtt() {
-		// If the MQTT subscription is QoS1, then we bail out if the published
-		// message is QoS1 because it will be handled in the other callack.
-		if subQos == 1 && mqttGetQoS(pc.mqtt.pp.flags) > 0 {
+		// This is an MQTT publisher directly connected to this server.
+
+		msgQOS := mqttGetQoS(pc.mqtt.pp.flags)
+		// This callback delivers only QOS0 messages to QOS0 subscriptions.
+		if subQOS > 0 || msgQOS > 0 {
```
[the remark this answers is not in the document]
levb: `&&`, done, thx!

server/mqtt.go
levb: Restored the logic and got rid of the `sess.createConsumer` wrapper; use `sess.jsa.createConsumer` since sess is always locked when we do it.

server/mqtt.go
kozlovic: Not sure, at least not for MQTT v3.1.1

server/mqtt.go
```diff
@@ -2165,7 +2304,7 @@ func (as *mqttAccountSessionManager) serializeRetainedMsgsForSub(sess *mqttSessi
 		pi := sess.getPubAckIdentifier(mqttGetQoS(rm.Flags), sub)
 		// Need to use the subject for the retained message, not the `sub` subject.
 		// We can find the published retained message in rm.sub.subject.
-		flags := mqttSerializePublishMsg(prm, pi, false, true, []byte(rm.Topic), rm.Msg)
+		flags := mqttSerializePublishMsg(prm, pi, sub.mqtt.qos, false, true, []byte(rm.Topic), rm.Msg)
```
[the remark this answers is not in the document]
kozlovic: As I previously said, I think that there is currently (read before this PR) an issue with handling of RETAIN flag. I may work on that on a separate PR (created of your branch or after this PR is merged).

server/mqtt.go
```diff
@@ -3050,7 +3207,8 @@ func (s *Server) mqttHandleWill(c *client) {
 		pp.flags |= mqttPubFlagRetain
 	}
 	c.mu.Unlock()
-	s.mqttProcessPub(c, pp)
+	// <>/<> restore
+	// s.mqttForwardPub(c, pp)
```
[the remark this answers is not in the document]
kozlovic: Replace with `s.mqttInitiateMsgDelivery(c, pp)` and remove `c.flushClients(0)` (since that is done by mqttInitiateMsgDelivery)

server/mqtt_test.go
```diff
@@ -1699,39 +1699,6 @@ func TestMQTTDontSetPinger(t *testing.T) {
 	testMQTTPublish(t, mc, r, 0, false, false, "foo", 0, []byte("msg"))
 }
```
kozlovic, line 1700: Need to change test: `TestMQTTSubAck` to have `expected` slice change the second `1` to a `2` now since QoS2 is supported.

Also change `TestMQTTSubDups` so that we don't test the downgraded from QoS2 to 1 in this test:
```
-       // Ask for QoS 2 but server will downgrade to 1
-       testMQTTSub(t, 1, mc, r, []*mqttFilter{{filter: "bar", qos: 2}}, []byte{1})
+       // Ask for QoS 1
+       testMQTTSub(t, 1, mc, r, []*mqttFilter{{filter: "bar", qos: 1}}, []byte{1})
```

kozlovic (changes requested):

server/mqtt.go
```diff
@@ -3652,65 +4061,87 @@ func mqttDeliverMsgCbQos1(sub *subscription, pc *client, _ *Account, subject, re
...
+
+	// Check the subscription's QoS. This needs to be protected because
+	// the client may change an existing subscription at any time.
+	sess.mu.Lock()
+	if qos > sub.mqtt.qos {
+		qos = sub.mqtt.qos
+	}
+	jsDur := sub.mqtt.jsDur
 	sess.mu.Unlock()
+	if qos == 0 {
```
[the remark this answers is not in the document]
kozlovic: Or at least something like:
```
// We lock to check some of the subscription's fields and if we need to
+       // keep track of pending acks, etc..
+       sess.mu.Lock()
+       if sess.c != sub.client || sub.mqtt == nil {
+               sess.mu.Unlock()
                return
        }
 
        // Check for reserved subject violation.
        strippedSubj := string(subject[len(mqttStreamSubjectPrefix):])
        if mqttIsReserved(sub, strippedSubj) {
+               sess.jsa.sendq.push(&mqttJSPubMsg{subj: reply, hdr: -1})
+               sess.mu.Unlock()
                return
        }
```
which avoid allocating and then releasing a PI, but will "ack" that message.

server/mqtt.go
levb: OK. I am going to add encoding the flag into the header, so it's easily available; otherwise will try to leave the retained handling as is.

server/mqtt.go
```diff
@@ -2442,30 +2600,38 @@ func (sess *mqttSession) save() error {
...
-	if l := len(sess.cons); l > 0 {
-		durs = make([]string, 0, l)
-		for sid, cc := range sess.cons {
-			delete(sess.cons, sid)
-			durs = append(durs, cc.Durable)
+	// <>/<> Can I just add the pubRelConsumer to the cons map and get rid of
+	// the field?
+	for sid, cc := range sess.cons {
+		delete(sess.cons, sid)
+		if _, err := sess.jsa.deleteConsumer(mqttStreamName, cc.Durable); isErrorOtherThan(err, JSConsumerNotFoundErr) {
```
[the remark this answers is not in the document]
levb: > The "jsa" calls are request-replies with the JetStream sub-system and you don't want to hold the session lock there.

That's is exactly how I understood the idea, but... it was already inside the lock, see https://github.com/nats-io/nats-server/blob/b839c53abcecf5c3309bf17df1fecc97d04ebb3c/server/mqtt.go#L2458. And it appeared to me that it was deliberately moved inside the lock with    https://github.com/nats-io/nats-server/pull/2775.

I actually dug in and did not see a reason for why the session should be locked at all for using jsa. I'll move this out of the lock, but should prob. review other instances of using jsa requests.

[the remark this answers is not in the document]
kozlovic: The PR 2775 was different, this was about replacing go channels with "ipq" for intra-process communication. So moving under the lock was to not reference the queue's pointer outside of the lock, but that's it.

But I am not arguing that there are no other cases where we do actual "jsa" req/reply operations under the session lock, but we should try to minimize those.

server/mqtt.go
```diff
@@ -3415,23 +3770,21 @@ func mqttParsePubAck(r *mqttReader, pl int) (uint16, error) {
...
+// sess.pending[pi] so this works for both the message and the PUBREL consumers.
 //
-// Runs from the client's readLoop.
-// No lock held on entry.
-func (c *client) mqttProcessPubAck(pi uint16) {
+// Runs from the client's readLoop. No lock held on entry.
+func (c *client) mqttProcessPubAckAndComp(pi uint16) {
 	sess := c.mqtt.sess
-	if sess == nil {
+	if sess == nil || sess.c != c {
```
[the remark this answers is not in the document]
levb: Oops, I missed the reassignment of existing session's `c`, thought the pointer was immutable. Thx, will move back.

server/mqtt.go
```diff
@@ -3440,15 +3793,77 @@ func (c *client) mqttProcessPubAck(pi uint16) {
 		if len(sess.pending) == 0 {
 			sess.ppi = 0
 		}
-		ackSubject = ack.subject
+		jsAckSubject = ack.jsAckSubject
 	}
-	// Send the ack if applicable.
-	if ackSubject != _EMPTY_ {
```
levb, line 3446: @kozlovic why were lines 3446-3449 inside the lock? As I understand it, sending the ack is async anyway, and the send queue has its own synchronization?

kozlovic: True, but the "push" should be fast and non blocking also, so you would not gain too much by doing the processing outside, and you will introduce more changes in this PR than needed. You could have a later pass at code to try to reduce time under lock, but I would keep changes in this PR to a strict minimum to implement QoS2.
All that being said, yes, sess.jsa is immutable (so is the sess.jsa.sendq) and therefore you could that outside of the lock - given that this function will be invoked by the readLoop routine...

server/mqtt.go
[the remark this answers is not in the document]
levb: > we should try to minimize [other cases where we do actual "jsa" req/reply operations under the session lock]

Yes, but like you said in the other comment, let me restore the closest to original, and then do another PR to move them out, maybe also switch to RWMutex if warranted.

server/mqtt.go
```diff
@@ -3652,65 +4061,87 @@ func mqttDeliverMsgCbQos1(sub *subscription, pc *client, _ *Account, subject, re
...
 		return
 	}
+
+	// Check the subscription's QoS. This needs to be protected because
+	// the client may change an existing subscription at any time.
+	sess.mu.Lock()
+	if qos > sub.mqtt.qos {
+		qos = sub.mqtt.qos
+	}
+	jsDur := sub.mqtt.jsDur
```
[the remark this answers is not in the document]
levb: I tried to remove most differences, just finished up the code, test(s) are still failing, so more changes to come here.

server/mqtt.go
levb: @kozlovic I am going to resolve all "retained QOS" comments for now, I left a TODO at the top to finalize that,

server/mqtt_test.go
levb: We now support all these packets

server/mqtt.go
```diff
@@ -3140,44 +3309,229 @@ func mqttPubTrace(pp *mqttPublish) string {
...
+	// JetStream under "$MQTT.msgs.<delivery-subject>"
+	if qos := mqttGetQoS(pp.flags); qos == 0 {
+		return nil
+	}
+	_, err := c.mqtt.sess.jsa.storeMsg(mqttStreamSubjectPrefix+string(pp.subject), headerLen, natsMsg)
+	if err != nil {
+		return err
+	}
+
+	return nil
```
kozlovic, line 3469: Simplify with a `return err`?

server/mqtt.go
```diff
@@ -3140,44 +3309,229 @@ func mqttPubTrace(pp *mqttPublish) string {
...
+}
+
+func (s *Server) mqttStoreQOS2Msg(c *client, pp *mqttPublish) error {
+	// `true` means encode the MQTT PUBLISH packet in the NATS message header.
+	natsMsg, headerLen := mqttNewDeliverableMessage(pp, true)
+
+	// Do not broadcast the message until it has been deduplicated and
+	// released by the sender. Instead store this QoS2 message as
+	// "$MQTT.qos2.<account-name>.<client-id>.<PI>". If the message
```
kozlovic, line 3478: Comment can be updated here to remove `<account-name>`

server/mqtt.go
```diff
@@ -3440,15 +3812,73 @@ func (c *client) mqttProcessPubAck(pi uint16) {
...
+	if len(sess.pending) == 0 {
+		sess.ppi = 0
 	}
+
+	// Move the PI tracking to the PUBREL consumer. It will not have a
+	// sequence nor an ACK subject until we get a PUBREL callback on the
+	// subscription. See <>/<>
+	sess.pending[pi].jsDur = sess.pubRelConsumer.Durable
+	sess.pending[pi].jsAckSubject = ""
+	sess.pending[pi].sseq = 0
```
kozlovic, line 3874: Not sure if the compiler optimizes that, but anyway, you already have `ack, ok := sess.pending[pi]` above, so use `ack.jsDur ...` instead here.

server/mqtt.go
```diff
@@ -3584,23 +4015,23 @@ func mqttDeliverMsgCbQos0(sub *subscription, pc *client, _ *Account, subject, re
...
 	// the client may change an existing subscription at any time.
 	sess.mu.Lock()
-	subQos := sub.mqtt.qos
-	isReserved := mqttCheckReserved(sub, subject)
+	subQOS := sub.mqtt.qos
 	sess.mu.Unlock()
 
 	// We have a wildcard subscription and this subject starts with '$' so ignore per Spec [MQTT-4.7.2-1].
-	if isReserved {
+	if mqttIsReserved(sub, subject) {
```
kozlovic, line 4022: No, see comment on that function. It needs the lock. So keep the `isReserved := mqttIsReserved(sub, subject)` inside the lock few lines above.

server/mqtt.go
```diff
@@ -3679,20 +4098,29 @@ func mqttDeliverMsgCbQos1(sub *subscription, pc *client, _ *Account, subject, re
 		return
 	}
 
-	// This is a QoS1 message for a QoS1 subscription, so get the pi and keep
-	// track of ack subject.
-	pQoS := byte(1)
-	pi, dup := sess.trackPending(pQoS, reply, sub)
+	// In this callback we handle only QOS-published messages to QOS
+	// subscriptions. Ignore if either is 0, will be delivered by the other
+	// callback, mqttDeliverMsgCbQos1.
```
kozlovic, line 4103: You meant `..CbQos0`?

server/mqtt.go
```diff
@@ -3835,6 +4299,102 @@ func (sess *mqttSession) cleanupFailedSub(c *client, sub *subscription, cc *Cons
...
+			return err
+		}
+		sess.pubRelConsumer = &ccr.Config
+		sess.tmaxack += maxAckPending
+	}
+
+	return nil
+}
+
+// func (sess *mqttSession) countQOS2Subs() int {
```
kozlovic, line 4365: So you are going to invoke that in another commit? If not then remove the code (copy somewhere if needed)

kozlovic (changes requested):

levb:
@kozlovic sorry about all the debugging in `mqtt_test.go`. It is _not_ yet ready for review, I am actively working on it, I'll re-request your review on it soon, but thx!!!

server/mqtt.go
levb: Still uncertain if there's ever a use-case for removing the PubRel consumer once there are no QoS2 subs. Probably not since there may be pending PUBRELs there, will remove.

server/mqtt.go
levb: Oops I missed this one, fixed QoS1 only. Thx!!!

server/mqtt.go
```diff
@@ -2448,19 +2616,32 @@ func (sess *mqttSession) clear() error {
...
+	sess.pending = nil
+	sess.cpending = nil
+	sess.pubRelConsumer = nil
+	sess.seq = 0
+	sess.tmaxack = 0
+	sess.mu.Unlock()
+
 	for _, dur := range durs {
 		if _, err := sess.jsa.deleteConsumer(mqttStreamName, dur); isErrorOtherThan(err, JSConsumerNotFoundErr) {
-			sess.mu.Unlock()
```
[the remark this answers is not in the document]
levb: Exactly, that's why I got confused about the jsa.sendq.push usage elsewhere, this seemed deliberate. Just to be clear, are you good with the current proposed implementation?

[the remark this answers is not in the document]
kozlovic: Yes. But then should we go a step further and capture the pubRelConsumer.Durable (if present) and then perform the delete outside of the lock?

server/mqtt.go
levb: I am confused. 

Line 3872 changes the consumer name for this PI from "messages" (which is what it was before, and what's in the ack) to "pubrel", where I do not yet have the actual ack subject; that would be in `mqttDeliverPubRelCb` where I get it.

kozlovic: What I meant is that all those lines with `sess.pending[pi]...=` mean that there is a map lookup, then assignment to the returned value by the lookup. The lookup was done earlier when you had `ack, ok := sess.pending[pi]`. I don't see anywhere that you are changing `pi`, so those calls are actually modifying the structure that `ack` points to. If that is not what you intended, then the code needs to be changed.

server/mqtt.go
```diff
@@ -2641,6 +2638,13 @@ func (sess *mqttSession) clear() error {
 			return fmt.Errorf("unable to delete consumer %q for session %q: %v", dur, sess.id, err)
 		}
 	}
+	if pubRelDur != "" {
+		_, err := sess.jsa.deleteConsumer(mqttQoS2PubRelStreamName, pubRelDur)
+		if isErrorOtherThan(err, JSConsumerNotFoundErr) {
+			sess.mu.Unlock()
```
kozlovic, line 2644: Copy-paste error here.

kozlovic:
@levb Do you need me to do the final review or are you still working on it?

levb:
@kozlovic not yet, I am hoping to have it for you tomorrow.

server/mqtt.go
```diff
@@ -3440,15 +3814,73 @@ func (c *client) mqttProcessPubAck(pi uint16) {
...
+	// Initiate PUBREL delivery
+	natsMsg, headerLen := mqttNewDeliverablePubRel(pi)
+	_, err := sess.jsa.storeMsg(sess.pubRelSubject(), headerLen, natsMsg)
+	if err != nil {
+		return err
+	}
+
+	sess.mu.Lock()
+	if sess.c != c {
+		return nil
```
kozlovic, line 3852: session unlock missing here!

kozlovic (changes requested):

server/mqtt.go
```diff
@@ -3440,15 +3814,74 @@ func (c *client) mqttProcessPubAck(pi uint16) {
...
-		ackSubject = ack.subject
+		jsAckSubject = ack.jsAckSubject
+	}
+	sess.mu.Unlock()
+
+	// Send the ack to JS to remove the pending message from the consumer.
+	sess.jsa.sendAck(jsAckSubject)
+}
+
+func (s *mqttSession) pubRelSubject() string {
```
levb, line 3825: Move to a sess (const) variable

server/mqtt.go
kozlovic: So what I would suggest is that since `mqttDeliverNATS` (the function that invokes process inbound but has a defer that clears all `c.pa.xxx` fields (therefore `c.pa.subject`) is only invoked - so far - in `mqttInitiateMsgDelivery`, you could do the defer that clears c.pa fields in that function (before calling `mqttDeliverNATS`) and use `c.pa.subject` here instead of `pp.subject`, or do not clear `c.pa.subject` has part of the defer inside `mqttDeliverNATS`, but then it means still have a defer here to make sure you clear it since there are several returns here.

(Also, since this is always invoked from the client's readLoop, and we set c.pa.subject=pp.subject at the beginning of the processing, we *could* simply not do the reset in the defer. But I wanted to do that because pp.subject may sometime reference the read buffer, so I wanted to make sure that this would not cause the compiler to have to make a copy because the reference would still be held after a read loop iteration).

server/mqtt.go
```diff
@@ -1146,6 +1249,30 @@ func (s *Server) mqttCreateAccountSessionManager(acc *Account, quitCh chan struc
 		}
 	}
 
+	if si, err := lookupStream(mqttQOS2IncomingMsgsStreamName, "QoS2 incoming messages"); err != nil {
+		return nil, err
+	} else if si == nil {
+		// Create the stream for the incoming QoS2 messages that have not been
+		// PUBREL-ed by the sender. Subject is
+		// "$MQTT.qos2.<domain>.<session>.<PI>", the .PI is to achieve exactly
```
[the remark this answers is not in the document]
neilalexander: If we want to remove it then let's do so, but let's do it sooner rather than later (before any release candidates).

server/mqtt.go
levb: @kozlovic Restored the exact(?) logic from before, including an explicit `flushClient` in `mqttHandleWill`.

server/mqtt.go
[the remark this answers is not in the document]
levb: @kozlovic @neilalexander  I'll submit a separate PR against dev shortly.

levb:
@kozlovic I think I addressed previous comments and this PR is now ready for a final review. I'll make a separate PR to eliminate `tkDomain`.

server/mqtt.go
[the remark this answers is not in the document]
levb: https://github.com/nats-io/nats-server/pull/4440

server/mqtt.go
```diff
@@ -2336,10 +2336,17 @@ func (as *mqttAccountSessionManager) serializeRetainedMsgsForSub(sess *mqttSessi
...
 		}
 		prm := sub.mqtt.prm
-		pi := sess.getPubAckIdentifier(mqttGetQoS(rm.Flags), sub)
+		var pi uint16
+		qos := mqttGetQoS(rm.Flags)
+		if qos > sub.mqtt.qos {
+			qos = sub.mqtt.qos
+		}
+		if qos > 0 {
+			pi = sess.trackPublishRetained()
```
kozlovic, line 2333: So in this case we know that we should get a `pi > 0` (in that it is QoS > 0 for both message and subscription). Here I wonder if `pi == 0` if we should not skip delivery of that retained message - since we can't find a suitable PI. Not sure of the implications there, and maybe going with PI==0 is better than nothing, but just wanted to raise the issue.

server/mqtt.go
```diff
@@ -2743,6 +2738,40 @@ func (sess *mqttSession) bumpPI() uint16 {
...
+	}
+	if sess.cpending == nil {
+		sess.cpending = make(map[string]map[uint64]uint16)
+	}
+
+	pi := sess.bumpPI()
+	if pi == 0 {
+		return 0
+	}
+	sess.pendingPublish[pi] = &mqttPending{}
```
kozlovic, line 2749: I feel like you are missing setting the fields of `mqttPending`. I am not sure how you will be able to "clean up" when processing the PUBACK (in `untrackPublish()`).

kozlovic (changes requested):

server/mqtt.go
levb: Hmmm, good point @kozlovic. Passing PI 0 down the line is not a good idea anyway, may be a place in the QOS code where it is assumed to be non-0. Maybe we could (should?) set the (delivery) `qos` to 0 for that message? That'd be a 1-line change...

server/mqtt.go
```diff
@@ -3371,6 +3371,7 @@ func (s *Server) mqttHandleWill(c *client) {
 	}
 	c.mu.Unlock()
 	s.mqttInitiateMsgDelivery(c, pp)
+	c.flushClients(0)
```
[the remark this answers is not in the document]
levb: In my previous version there was a change relative to the original code; I effectively added a `flushClients` into the QoS0 path where it was not before. I am not sure if it was consequential, but restored the original anyway.

derekcollison (commented):
In general for PRs we rebase, not merge.

levb:
@derekcollison can we squash-merge this? There were so many "substantial" (as opposed to merging from dev) commits here that keeping the commit history would hardly make sense. I can squash everything in a single commit myself and force-push once we have all conversations resolved? LMK what is the right thing to do.

server/mqtt.go
levb: See https://github.com/nats-io/nats-server/pull/4349/commits/a2452aa844ff4609728a223c3c76d7d67927d922

server/mqtt.go
kozlovic: But you do set something is cpending:
```
    jsDur := mqttRetainedMsgsStreamName
    sseq := uint64(pi)
    sseqToPi, ok := sess.cpending[jsDur]
    if !ok {
        sseqToPi = make(map[uint64]uint16)
	sess.cpending[jsDur] = sseqToPi
    }
```

kozlovic:
@levb What Derek said is that when you want to update your PR's branch with the `dev` branch, you could have done `git rebase dev` instead of the merge to have all your changes "sit" on top of latest from dev.

But for sure, once you get the green light for this PR, I would recommend that you squash everything to a single commit and push-force and let Travis re-run all tests, then if all is OK, you will merge into dev.

server/mqtt.go
levb: Oh, sorry will remove. This was in preparation for persisting the PI mappings which I was doing out of `cpending` (for compactness). It does not belong in this PR.

kozlovic (approved):
LGTM

levb:
@derekcollison rebased/squashed

derekcollison:
Will let you two handle review and merge.

server/mqtt.go
```diff
@@ -2503,106 +2710,217 @@ func (sess *mqttSession) update(filters []*mqttFilter, add bool) error {
...
-	_, err := sess.jsa.createConsumer(cfg)
-	return err
+
+	delete(sess.pendingPublish, pi)
+	if len(sess.pendingPublish) == 0 {
+		sess.last_pi = 0
+	}
+
+	if len(sess.cpending) != 0 && ack.jsDur != _EMPTY_ {
+		if sseqToPi := sess.cpending[ack.jsDur]; sseqToPi != nil {
```
neilalexander, line 2858: When using `map[x]map[y]z` you need to make sure that you clean up the top-level key if the second-level map is now `len(...) == 0`, otherwise we can end up with a map pointing to a bunch of empty maps.

Better approach if you are only dealing with base types is to define a composite key type struct:
```
type foo struct {
  bar string
  baz uint64
}
```
... and have a single-level `map[foo]qux`, so that you don't have to worry about nesting and top-level deletes.

server/mqtt.go
```diff
@@ -2503,106 +2710,217 @@ func (sess *mqttSession) update(filters []*mqttFilter, add bool) error {
...
+		}
+		return
+	}
+
+	sseq, _, _ := ackReplyInfo(jsAckSubject)
+
+	var sseqToPi map[uint64]uint16
+	if sess.cpending == nil {
+		sess.cpending = make(map[string]map[uint64]uint16)
+	} else if sseqToPi = sess.cpending[jsDur]; sseqToPi == nil {
```
neilalexander, line 2894: Above would clean up quite a lot of things like this.

neilalexander (commented):
Few minor things but otherwise the bulk of the PR looks great!

server/mqtt.go
levb: Yes, I even started the change and then decided against it. Maybe a future refactoring, with benchmarks?

server/mqtt.go
```diff
@@ -2503,106 +2710,217 @@ func (sess *mqttSession) update(filters []*mqttFilter, add bool) error {
...
+		pi = sseqToPi[sseq]
 	}
-	if pi == 0 {
+
+	if pi != 0 {
+		// There is a possible race between a PUBLISH re-delivery calling us,
+		// and a PUBREC received already having submitting a PUBREL into JS . If
+		// so, indicate no need for (re-)delivery by returning a PI of 0.
+		_, usedForPubRel := sess.pendingPubRel[pi]
+		if /*dup && */ usedForPubRel {
```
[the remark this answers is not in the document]
levb: 👍 Yes, I wanted to assert it. Will do a separate small PR for that, later.

server/mqtt.go
neilalexander: Yep, let's keep it for a future change then.

server/mqtt.go
```diff
@@ -2503,106 +2710,217 @@ func (sess *mqttSession) update(filters []*mqttFilter, add bool) error {
 	return err
 }
 
-// If both pQos and sub.mqtt.qos are > 0, then this will return the next
-// packet identifier to use for a published message.
+func (sess *mqttSession) bumpPI() uint16 {
+	var avail bool
+	next := sess.last_pi
+	for i := 0; i < 0xFFFF; i++ {
```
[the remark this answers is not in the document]
levb: prior code, left unchanged.

neilalexander (approved):
LGTM

PR #4354, opened 2023-08-01
[ADDED] Full StreamSource (filters, transforms) functionality to stream mirror

jnmoyne:
- [X] Tests added

Follow up to #4276 extending to Mirror the full StreamSource functionality.

### DIFF
--- a/server/stream.go
+++ b/server/stream.go
@@ -481,7 +481,7 @@ func (a *Account) addStreamWithAssignment(config *StreamConfig, fsConfig *FileSt
 				if st.Destination != _EMPTY_ {
 					if _, err = NewSubjectTransform(st.Source, st.Destination); err != nil {
 						jsa.mu.Unlock()
-						return nil, fmt.Errorf("subject transform from '%s' to '%s' for the mirror %w", st.Source, st.Destination, err)
+						return nil, fmt.Errorf("subject transform from '%s' to '%s' for the mirror: %w", st.Source, st.Destination, err)
 					}
 				}
 			}
--- a/server/stream.go
+++ b/server/stream.go
@@ -458,7 +458,37 @@ func (a *Account) addStreamWithAssignment(config *StreamConfig, fsConfig *FileSt
 		}
 	}
 
-	// Setup our internal indexed names here for sources and check if the transform (if any) is valid.
+	// If mirror, check if the transforms (if any) are valid.
+	if cfg.Mirror != nil {
+		if len(cfg.Mirror.SubjectTransforms) == 0 {
+			if cfg.Mirror.FilterSubject != _EMPTY_ && !IsValidSubject(cfg.Mirror.FilterSubject) {
+				jsa.mu.Unlock()
+				return nil, fmt.Errorf("subject filter '%s' for the mirror %w", cfg.Mirror.FilterSubject, ErrBadSubject)
+			}
+			if cfg.Mirror.SubjectTransformDest != _EMPTY_ {
+				if _, err = NewSubjectTransform(cfg.Mirror.FilterSubject, cfg.Mirror.SubjectTransformDest); err != nil {
+					jsa.mu.Unlock()
+					return nil, fmt.Errorf("subject transform from '%s' to '%s' for the mirror %w", cfg.Mirror.FilterSubject, cfg.Mirror.SubjectTransformDest, err)
+				}
+			}
+		} else {
+			for _, st := range cfg.Mirror.SubjectTransforms {
+				if st.Source != _EMPTY_ && !IsValidSubject(st.Source) {
+					jsa.mu.Unlock()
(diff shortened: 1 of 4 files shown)


server/stream.go
```diff
@@ -458,7 +458,37 @@ func (a *Account) addStreamWithAssignment(config *StreamConfig, fsConfig *FileSt
...
+			for _, st := range cfg.Mirror.SubjectTransforms {
+				if st.Source != _EMPTY_ && !IsValidSubject(st.Source) {
+					jsa.mu.Unlock()
+					return nil, fmt.Errorf("subject filter '%s' for the mirror %w", st.Source, ErrBadSubject)
+				}
+				// check the transform, if any, is valid
+				if st.Destination != _EMPTY_ {
+					if _, err = NewSubjectTransform(st.Source, st.Destination); err != nil {
+						jsa.mu.Unlock()
+						return nil, fmt.Errorf("subject transform from '%s' to '%s' for the mirror %w", st.Source, st.Destination, err)
```
Jarema, line 484: I do not like the `%w` without colon, as it's pretty hard to read for me, but that's only one opinion.
We should agree how we format server errors in general.

server/stream.go
```diff
@@ -2343,7 +2404,19 @@ func (mset *stream) setupMirrorConsumer() error {
...
 		req.Config.FilterSubject = mset.cfg.Mirror.FilterSubject
+		if mset.cfg.Mirror.SubjectTransformDest != _EMPTY_ {
+			mirror.tr, _ = NewSubjectTransform(mset.cfg.Mirror.FilterSubject, mset.cfg.Mirror.SubjectTransformDest)
+		}
+	}
+
+	var filters []string
+	for _, tr := range mset.cfg.Mirror.SubjectTransforms {
+		// will not fail as already checked before that the transform will work
+		subjectTransform, _ := NewSubjectTransform(tr.Source, tr.Destination)
```
Jarema, line 2415: I know you check for this error, and that right now it's fine, but I feel it's not safe against refactors and future changes that might alter the flow.

I think that price for single error check is pretty small to avoid possible future risk.

Jarema (commented):
Looks good!

Just few comments.

server/stream.go
```diff
@@ -458,7 +458,37 @@ func (a *Account) addStreamWithAssignment(config *StreamConfig, fsConfig *FileSt
...
+				if _, err = NewSubjectTransform(cfg.Mirror.FilterSubject, cfg.Mirror.SubjectTransformDest); err != nil {
+					jsa.mu.Unlock()
+					return nil, fmt.Errorf("subject transform from '%s' to '%s' for the mirror %w", cfg.Mirror.FilterSubject, cfg.Mirror.SubjectTransformDest, err)
+				}
+			}
+		} else {
+			for _, st := range cfg.Mirror.SubjectTransforms {
+				if st.Source != _EMPTY_ && !IsValidSubject(st.Source) {
+					jsa.mu.Unlock()
+					return nil, fmt.Errorf("subject filter '%s' for the mirror %w", st.Source, ErrBadSubject)
```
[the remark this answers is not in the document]
jnmoyne: Yes, same difference really. For example if you just want to filter messages and not transform them then you only provide the source part of the transform. Notice it's the same error message if you provide an invalid value for the single subject filter field. So I kept the error the same. Let me know if that bothers you and you would like different error messages in each case.

server/stream.go
jnmoyne: Sure, done.

server/stream.go
jnmoyne: This is because it's not obvious to me how to handle an error at that point. That's why it does check for an error earlier in the code path. If you check, although the signature of `setupMirrorConsumer()` returns and `error` the code actually always does `return nil` and the only couple of error conditions it handles are for potentially transient errors where it re-schedules re-trying to set the consumer later, while if `NewSubjectTransform` returns an error scheduling a retry later will not help because the transform destination is not going to suddenly fix itself. 

So I don't mind checking the error is, but what would you suggest as to what to do if it does return an error?

jnmoyne: Added the logging of an error if it happens

server/stream.go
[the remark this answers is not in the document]
Jarema: One of important things when debugging why your client/app does not work, is a hint where in config look for the error. Hence my concern that having exactly the same error might be confusing. Technically it's correct, but it would be nice to tell the users what the messed up.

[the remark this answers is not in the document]
jnmoyne: Improved the error message

server/stream.go
jnmoyne: Ok to resolve?

neilalexander (approved):
Overall LGTM

Jarema (approved):
LGTM!

neilalexander (approved):

Issue #3773, opened 2023-01-09
OCSP mode assumes only one CA cert, for the current issuer


philpennock:
#### Versions of `nats-server` and affected client libraries used:

nats-server 2.9.11, nats-cli current HEAD

#### OS/Container environment:

FreeBSD Jail, but reproduces anywhere

#### Steps or code to reproduce the issue:

Have a TLS block with a `ca_file` with more than one cert in it, and setup an OCSP block.

```
tls {
    ca_file:   /etc/ssl/cert.pem
    cert_file: /etc/nats/pkix/nats.home.pennock.cloud/tls.crt
    key_file:  /etc/nats/pkix/nats.home.pennock.cloud/tls.key
}
ocsp {
    mode: always
    url: "http://r3.o.lencr.org"
}
```

#### Expected result:

The server starts up, requests and obtains an OCSP staple, and I can query it.

#### Actual result:

If the `ca_file` bundle contains just one certificate, the intermediate used for issuing the current certificate, then things work.  This is problematic, because at a bare minimum with a CA such as Let's Encrypt, you should be configuring both R3 and R4 (or E1 and E2) together, so that LE can fail-over to their stand-by.  The `ca_file` should be a list of all certificate authorities trusted for verifying peer identities, not a statement of which specific intermediate issued the current server certificate.

If the `ca_file` contains a list of all the normal system CA certs, then with OCSP enabled the nats-server start-up errors out with:

```
% /usr/local/sbin/nats-server -DV -c /etc/nats/nats.conf -l /dev/stdout 
nats-server: CN=Thawte Universal CA Root,OU=Thawte Universal CA Root,O=Thawte invalid ca basic constraints: is not ca
```

If the `ca_file` contains `/etc/ssl/bundle-lets-encrypt.pem` which is a bundle of _just_ the Let's Encrypt certs (E1, E2, R3, R4), then nats-server errors out with:

```
% /usr/local/sbin/nats-server -DV -c /etc/nats/nats.conf -l /dev/stdout  
nats-server: bad OCSP status update for certificate at '/etc/nats/pkix/nats.home.pennock.cloud/tls.crt': failed to get remote status: ocsp: error from server: unauthorized
```

which is OCSP-speak in this scenario for "sent the wrong cert details".

If I strip `ca_file` down to contain _only_ the R3 certificate, then I get working OCSP; running the new natscli command:

```
nats account tls --ocsp --no-pem
```

correctly yields:

```
[...]

PR #4355, opened 2023-08-01
Fixed local issuer determination for OCSP Staple, issue #3773

tbeets:
Resolves problems of [issue #3773](https://github.com/nats-io/nats-server/issues/3773).  

With this fix, NATS Server will locally determine it's own certificate's issuer from either the configured server certificate (bundle of leaf cert plus optional intermediate CA certs) or from the configured server CA trust store, as follows: 

1. The operator may provide the server's certificate issuer in the second position of the server's certificate configuration (typically `cert_file` but may be `cert_store` on the Windows platform).  If a candidate issuer is found here it is PKI validated as the actual issuer of the server's cert else a hard error.

2. If not found in [1], NATS Server will seek to create at least one verified chain with its configured trust store (typically `ca_file` but could by the system trust store if not configured).  It will derive the issuer from the first verified chain.  If no verified chain can be formed it is a hard error.

### DIFF
--- a/test/ocsp_test.go
+++ b/test/ocsp_test.go
@@ -3454,7 +3454,6 @@ func TestOCSPLocalIssuerDetermination(t *testing.T) {
 	serverCert := "configs/certs/ocsp_peer/mini-ca/server1/TestServer1_cert.pem"
 
 	var (
-		// errExpectedNoStaple = fmt.Errorf("expected no staple")
 		errMissingStaple = fmt.Errorf("missing OCSP Staple from server")
 	)
 
--- a/test/ocsp_test.go
+++ b/test/ocsp_test.go
@@ -2305,20 +2305,14 @@ func TestOCSPGateway(t *testing.T) {
 }
 
 func TestOCSPGatewayIntermediate(t *testing.T) {
-	const (
-		caCert       = "configs/certs/ocsp/desgsign/ca-cert.pem"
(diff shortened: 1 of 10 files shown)


tbeets:
This fix also includes unit tests that verifies that both OCSP Staple and OCSP Peer features can be configured and used concurrently.

test/ocsp_test.go
```diff
@@ -3450,3 +3439,259 @@ func TestOCSPSuperCluster(t *testing.T) {
...
+	intermediateCA1ResponderURL := fmt.Sprintf("http://%s", intermediateCA1Responder.Addr)
+	defer intermediateCA1Responder.Shutdown(ctx)
+
+	// Test constants
+	ocspURL := intermediateCA1ResponderURL
+	clientTrustBundle := "configs/certs/ocsp_peer/mini-ca/misc/trust_config1_bundle.pem"
+	serverCert := "configs/certs/ocsp_peer/mini-ca/server1/TestServer1_cert.pem"
+
+	var (
+		// errExpectedNoStaple = fmt.Errorf("expected no staple")
```
wallyqs, line 3457: leftover comment?

wallyqs (approved):
LGTM!

test/ocsp_test.go
tbeets: Yep. Fixed.

PR #4356, opened 2023-08-01
apply Go 1.21 loopvar fixes

wallyqs:
Applies some of the loopvar fixes that Go 1.21 will find:

```sh
GOEXPERIMENT=loopvar go1.21rc2 build -gcflags=all=-d=loopvar=2 
server/accounts.go:3235:10: loop variable m now per-iteration, stack-allocated
server/server.go:3666:9: loop variable i now per-iteration, stack-allocated
server/server.go:3699:10: loop variable i now per-iteration, stack-allocated
server/filestore.go:5141:10: loop variable mb now per-iteration, stack-allocated
server/reload.go:1100:10: loop variable a now per-iteration, stack-allocated
server/reload.go:2195:10: loop variable an now per-iteration, stack-allocated
server/reload.go:2342:10: loop variable an now per-iteration, stack-allocated
```

### DIFF
--- a/server/accounts.go
+++ b/server/accounts.go
@@ -3233,6 +3233,7 @@ func (s *Server) updateAccountClaimsWithRefresh(a *Account, ac *jwt.AccountClaim
 	for sub, wm := range ac.Mappings {
 		mappings := make([]*MapDest, len(wm))
 		for i, m := range wm {
+			m := m
 			mappings[i] = &MapDest{
 				Subject: string(m.Subject),
 				Weight:  m.GetWeight(),
--- a/server/filestore.go
+++ b/server/filestore.go
@@ -5139,6 +5139,7 @@ func (fs *fileStore) State() StreamState {
 		cur := fs.state.FirstSeq
 
 		for _, mb := range fs.blks {
+			mb := mb
 			mb.mu.Lock()
 			fseq := mb.first.seq
 			// Account for messages missing from the head.
--- a/server/reload.go
+++ b/server/reload.go
@@ -1098,6 +1098,7 @@ func imposeOrder(value interface{}) error {
 			return value[i].Name < value[j].Name
 		})
 		for _, a := range value {
+			a := a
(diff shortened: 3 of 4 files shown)


server/accounts.go
```diff
@@ -3233,6 +3233,7 @@ func (s *Server) updateAccountClaimsWithRefresh(a *Account, ac *jwt.AccountClaim
 	for sub, wm := range ac.Mappings {
 		mappings := make([]*MapDest, len(wm))
 		for i, m := range wm {
+			m := m
```
derekcollison, line 3236: Why the heck is this needed?

wallyqs: to make `m` a different instance per iteration, in Go 1.22 that will be default behavior but in Go 1.21 can already try to find the for loops that will be auto fixed to test in advance: https://github.com/golang/go/wiki/LoopvarExperiment#if-the-proposal-is-accepted-how-will-the-change-be-deployed || https://go.googlesource.com/proposal/+/master/design/60078-loopvar.md#a-decade-of-experience-shows-the-cost-of-the-current-semantics

derekcollison: ok I will read up on it later but the fix, m := m looks kinda silly (not you just in general).. So rather find a better solution than that..

derekcollison: I think you read it wrong, they are trying to fix the closure capture problem..

oderwat:
@wallyqs I know I am just some "outsider" but I am curious about the loopvar problem in general. Do these changes imply that all those locations in the code are currently bugged?

bruth:
@oderwat The output @wallyqs showed indicates that these variables are heap allocated rather than stack (in loop) prior to Go 1.21.

After a quick review, all references are being used as inline lookups or methods are being accessed on them which would yield stack-scope values within the loop. So the existing behavior was not buggy.

wallyqs:
Thanks for looking @bruth, I will close this one for now since the behavior can be confirmed anyway by using `GOEXPERIMENT=loopvar` in a build.

PR #4357, opened 2023-08-02
[FIXED] Accounting drifts

derekcollison:
Three issues were found and resolved.

    1. Some purge replays after recovery  could execute full purge.
    2. Callback was registered without lock, which  could lead to skew.
    3. Cluster reset could stop stream store and recreate it, which could lead to double accounting.


wallyqs (approved):
LGTM

PR #4358, opened 2023-08-02
config: make parsing configurations without usable values invalid (backport)

### LINKED ISSUE #4343: Binary file accepted as valid configuration
## Defect

I mistakenly fed `nats-server` a binary rather than a configuration file, and `-t` tells me it is a valid configuration.

Noticed this on `2.9.21-RC.1`, but this has not changed recently.

Not relevant

Pass the path of a binary (or other non-textfile, like PDF) as configuration argument:
```
 ./nats-server -t -c ./nats-server
```

#### Expected result:
Non zero-exit code, and a message telling me I'm doing it wrong

#### Actual result:
`nats-server: configuration file ./nats-server is valid`

#### Details

Stepping through the configuration parsing steps, it seems like the entire binary is treated treated like as `itemKey`, and accepted.

Even setting `pedantic`, there are no errors and no warnings.

wallyqs:
This makes configuration files that are empty, or read and processed by the parser but with no detected values now return an error.

Fixes #4343 
Backport from dev branch (https://github.com/nats-io/nats-server/pull/4347)

### DIFF
--- a/conf/parse.go
+++ b/conf/parse.go
@@ -146,7 +146,9 @@ func parse(data, fp string, pedantic bool) (p *parser, err error) {
 			return nil, err
 		}
 	}
-
+	if len(p.mapping) == 0 {
+		return nil, fmt.Errorf("config has no values or is empty")
+	}
 	return p, nil
 }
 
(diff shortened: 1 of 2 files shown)


derekcollison (approved):
LGTM

PR #4361, opened 2023-08-02
Allow switching from limits-based to interest-based retention in stream update

neilalexander:
This should make it possible to switch from limits-based retention to interest-based retention on an existing stream.

### DIFF
--- a/server/stream.go
+++ b/server/stream.go
@@ -1495,9 +1495,11 @@ func (jsa *jsAccount) configUpdateCheck(old, new *StreamConfig, s *Server) (*Str
 	if cfg.Storage != old.Storage {
 		return nil, NewJSStreamInvalidConfigError(fmt.Errorf("stream configuration update can not change storage type"))
 	}
-	// Can't change retention.
+	// Can only change retention from limits to interest or back, not to/from work queue for now.
 	if cfg.Retention != old.Retention {
-		return nil, NewJSStreamInvalidConfigError(fmt.Errorf("stream configuration update can not change retention policy"))
+		if old.Retention == WorkQueuePolicy || cfg.Retention == WorkQueuePolicy {
+			return nil, NewJSStreamInvalidConfigError(fmt.Errorf("stream configuration update can not change retention policy to/from workqueue"))
+		}
 	}
 	// Can not have a template owner for now.
 	if old.Template != _EMPTY_ {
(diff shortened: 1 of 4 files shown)


server/stream.go
```diff
@@ -1488,9 +1488,9 @@ func (jsa *jsAccount) configUpdateCheck(old, new *StreamConfig, s *Server) (*Str
 		return nil, NewJSStreamInvalidConfigError(fmt.Errorf("stream configuration update can not change storage type"))
 	}
 	// Can't change retention.
-	if cfg.Retention != old.Retention {
-		return nil, NewJSStreamInvalidConfigError(fmt.Errorf("stream configuration update can not change retention policy"))
-	}
+	// if cfg.Retention != old.Retention {
```
derekcollison, line 1491: Remove?

server/stream.go
```diff
@@ -1780,6 +1780,27 @@ func (mset *stream) updateWithAdvisory(config *StreamConfig, sendAdvisory bool)
 	// Now update config and store's version of our config.
 	mset.cfg = *cfg
 
+	// Check for changes to the retention policy. If we are moving to interest-based
+	// retention then we will need to check the ack floor of all consumers.
+	if ocfg.Retention != cfg.Retention && cfg.Retention == InterestPolicy {
+		toUpdate := make([]*consumer, 0, len(mset.consumers))
+		for _, c := range mset.consumers {
```
derekcollison, line 1787: One important piece, if the consumer is retention interest, then its replication factor must be same as stream, so need to adjust that here and add in an R1 consumer to the test above etc.

server/stream.go
```diff
@@ -1780,6 +1780,27 @@ func (mset *stream) updateWithAdvisory(config *StreamConfig, sendAdvisory bool)
...
+				toUpdate = append(toUpdate, c)
+			}
+		}
+		mset.mu.Unlock()
+		for _, c := range toUpdate {
+			c.mu.Lock()
+			c.retention = cfg.Retention
+			c.mu.Unlock()
+			if c.retention == InterestPolicy {
+				c.checkStateForInterestStream()
```
derekcollison, line 1798: Nice! I saw this and was confused for a second then realized what you did here..

server/stream.go
```diff
@@ -1780,6 +1780,27 @@ func (mset *stream) updateWithAdvisory(config *StreamConfig, sendAdvisory bool)
 	// Now update config and store's version of our config.
 	mset.cfg = *cfg
 
+	// Check for changes to the retention policy. If we are moving to interest-based
+	// retention then we will need to check the ack floor of all consumers.
+	if ocfg.Retention != cfg.Retention && cfg.Retention == InterestPolicy {
+		toUpdate := make([]*consumer, 0, len(mset.consumers))
+		for _, c := range mset.consumers {
+			if c.retention != cfg.Retention {
```
derekcollison, line 1788: This should always be true yes?

server/test_test.go
```diff
@@ -112,17 +111,17 @@ func require_Error(t *testing.T, err error, expected ...error) {
...
+		t.Fatalf("require %T equal, but got: %v != %v", a, a, b)
 	}
 }
 
-func require_NotEqual(t *testing.T, a, b [32]byte) {
+func require_NotEqual[T comparable](t *testing.T, a, b T) {
 	t.Helper()
-	if bytes.Equal(a[:], b[:]) {
-		t.Fatalf("require not equal, but got: %v != %v", a, b)
+	if a == b {
```
derekcollison, line 123: Will this do the right thing for 2 different byte slices that have same contents?

derekcollison (changes requested):

server/test_test.go
neilalexander: The function was previously taking `a, b [32]byte` and fixed-size arrays can be compared using `==`, so it will continue to work for what it was being used for.

Trying to pass in slices instead of fixed-size arrays will result in a compile-time error as they are not `comparable`.

server/stream.go
```diff
@@ -1495,9 +1495,11 @@ func (jsa *jsAccount) configUpdateCheck(old, new *StreamConfig, s *Server) (*Str
 	if cfg.Storage != old.Storage {
 		return nil, NewJSStreamInvalidConfigError(fmt.Errorf("stream configuration update can not change storage type"))
 	}
-	// Can't change retention.
+	// Can only change retention from limits to interest or back, not to/from work queue for now.
 	if cfg.Retention != old.Retention {
-		return nil, NewJSStreamInvalidConfigError(fmt.Errorf("stream configuration update can not change retention policy"))
+		if old.Retention == WorkQueuePolicy || cfg.Retention == WorkQueuePolicy {
+			return nil, NewJSStreamInvalidConfigError(fmt.Errorf("stream configuration update can not change retention policy to/from work queue"))
```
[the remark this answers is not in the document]
neilalexander: Fixed, thanks!

wallyqs (approved):
LGTM

PR #4362, opened 2023-08-02
OCSP backports and adds

tbeets:
This PR backports the OCSP Peer feature option (as in 2.10 train) and includes two fixes for the existing OCSP Staple feature.  

OCSP Staple: 

1. Fixed and clarified how NATS Server determines its own Issuer CA when obtaining and validating an OCSP Response for subsequent staple
2. Eliminated problematic assumption that all node peers are issued by same CA when NATS Server validates ROUTE and GATEWAY peer nodes
3. Added OCSP Response effectivity checks on ROUTE and GATEWAY peer-presented staple

Note for #3: Allowed host clock skew between node peers set at 30-seconds.  If the OCSP Response contains an empty assertion for NextUpdate, NATS Server will default to 1-hour validity (after ThisUpdate).  It is recommended that CA OCSP Responder should assert NextUpdate.


wallyqs (approved):
LGTM

derekcollison (approved):
LGTM

PR #4365, opened 2023-08-03
[IMPROVED] When taking over make sure to sync and reset clfs for clustered streams.

derekcollison:
If the failed state of clfs drifts between leaders and followers, replicas could discard and skip messages possibly incorrectly. This will force sync if we have a non-zero clfs state when a leader takes over.


wallyqs (approved):
LGTM

PR #4370, opened 2023-08-05
Match --signal PIDs with globular-style expression.

jevolk:
When multiple instances are running on the machine a PID argument suffixed with a '*' character will signal all matching PIDs.

Example: `nats-server --signal reload=*`

 - [X] Tests added
 - [X] Branch rebased on top of current ~~main~~ dev

### DIFF
--- a/server/signal.go
+++ b/server/signal.go
@@ -82,54 +82,82 @@ func (s *Server) handleSignals() {
 
 // ProcessSignal sends the given signal command to the given process. If pidStr
 // is empty, this will send the signal to the single running instance of
-// nats-server. If multiple instances are running, it returns an error. This returns
-// an error if the given process is not running or the command is invalid.
-func ProcessSignal(command Command, pidStr string) error {
-	var pid int
-	if pidStr == "" {
-		pids, err := resolvePids()
-		if err != nil {
-			return err
-		}
-		if len(pids) == 0 {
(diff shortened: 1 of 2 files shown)


server/signal.go
```diff
@@ -82,54 +82,81 @@ func (s *Server) handleSignals() {
...
-			return fmt.Errorf("invalid pid: %s", pidStr)
+		if err = kill(pid, signum); err != nil {
+			errStr += fmt.Sprintf("\nsignal %q %d: %s", command, pid, err)
 		}
-		pid = p
 	}
+	if errStr != "" {
+		return errors.New(errStr)
+	}
+	return err
```
bruth, line 140: This is a nit, but it appears safe to `return nil` here since in all other cases above, an non-nil `err` value was returned appropriately. IMO, it helps with readability indicating that there is not fact _not_ an error value that could propagate down this far.

bruth (approved):
Other than the minor comment, LGTM!. Thanks for the contribution!

jevolk:
Thanks for taking the time to review this so quickly. I have made the requested change. I have also added another test to ensure coverage for partial matches with a trailing glob.

neilalexander (approved):
LGTM, thanks!

Just one final thing, would appreciate if you can add a `Signed-off-by:` to the commit message for your Apache 2 sign-off.

bruth (approved):

PR #4371, opened 2023-08-05
[IMPROVED] Delete blocks performance

derekcollison:
Track deleted with single avl.SeqSet dmap for now vs old method for memory store.

For fileStore, we were trying to be too smart to save space at the expense of encoding time, so revert back to simple version that is much 100x faster.
 
Size of encoding may be a bit bigger then we wanted, but we want to prefer speed over size.


neilalexander (approved):
LGTM!

PR #4381, opened 2023-08-09
test: fix TestAccountImportCycle flake

wallyqs:
Add extra flushes to make test more precise and try to avoid timeouts

```
=== RUN   TestAccountImportCycle
    accounts_test.go:3447: require no error, but got: nats: timeout
--- FAIL: TestAccountImportCycle (1.01s)
```


neilalexander (approved):
LGTM

PR #4394, opened 2023-08-14
config: allow empty configs, but prevent bad configs

wallyqs:
Allows again empty configs and improves support for detecting some types of invalid configs:

- Adds reporting the line with the bad key position that makes the config invalid.

```
nats-server: config is invalid (foo.conf:1:2)
nats-server: error parsing include file 'included.conf', config is invalid (included.conf:2:2)
```

- Fixes a few tests with trailing braces which were being handled as keys and ignored before.

### DIFF
--- a/conf/parse.go
+++ b/conf/parse.go
@@ -137,17 +137,19 @@ func parse(data, fp string, pedantic bool) (p *parser, err error) {
 	}
 	p.pushContext(p.mapping)
 
+	var prevItem itemType
 	for {
 		it := p.next()
 		if it.typ == itemEOF {
+			if prevItem == itemKey {
+				return nil, fmt.Errorf("config is invalid (%s:%d:%d)", fp, it.line, it.pos)
+			}
 			break
 		}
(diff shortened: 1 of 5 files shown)


derekcollison (commented):
Should we target main vs dev?

conf/parse.go
```diff
@@ -137,18 +137,20 @@ func parse(data, fp string, pedantic bool) (p *parser, err error) {
...
+				return nil, fmt.Errorf("config is invalid (%s:%d:%d)", fp, it.line, it.pos)
+			}
 			break
 		}
+		prevItem = it.typ
 		if err := p.processItem(it, fp); err != nil {
 			return nil, err
 		}
 	}
-	if len(p.mapping) == 0 {
```
ripienaar, line 149: We could still warn on these?

wallyqs:
changed branch to main and added a check to report loading empty configs as a warning only

neilalexander (approved):
LGTM

PR #4400, opened 2023-08-15
Fix stream config update of source transforms

jnmoyne:

Fixes potential out of range access during some stream source transform configuration updates and tiny clean up
Fixes stream sourcing message header parsing for multi-subject transform in sources


neilalexander (approved):
LGTM

PR #4402, opened 2023-08-16
Don't set block profile rate

neilalexander:
We rarely benefit from block profiles and in many cases a mutex profile will tell us what we need to know. Additionally, setting the block profile rate to `1` has the special meaning of capturing every single blocking event, which can have a fairly significant negative impact on publish performance.


wallyqs (approved):
LGTM

derekcollison (approved):
LGTM

PR #4403, opened 2023-08-16
[IMPROVED] StreamInfo reflecting subject transforms with just a filter and no transformation for Sources

jnmoyne:

Adds sfs to sourceInfo such that transforms with just a subject filter (and no transformation, meaning that the transform pointer in streamInfo is nil) can still be reflected in SourceInfo, which is important since the filtering is still happening, just no transformation as well.

### DIFF
--- a/server/stream.go
+++ b/server/stream.go
@@ -304,6 +304,7 @@ type sourceInfo struct {
 	wg    sync.WaitGroup
 	sf    string // subject filter
 	tr    *subjectTransform
+	sfs   []string            // subject filters
 	trs   []*subjectTransform // subject transforms
 }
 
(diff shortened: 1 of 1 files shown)


server/stream.go
```diff
@@ -2450,17 +2455,20 @@ func (mset *stream) setupMirrorConsumer() error {
 		}
 	}
 
-	var filters []string
+	var sfs []string
```
neilalexander, line 2458: Minor nit, but in the case of the length being already known (i.e. `len(mset.cfg.Mirror.SubjectTransforms)`) then it's better to preallocate backing capacity so that `append` won't result in reallocations:
```
sfs := make([]string, 0, len(mset.cfg.Mirror.SubjectTransforms))
```

server/stream.go
```diff
@@ -2013,10 +2016,12 @@ func (mset *stream) sourceInfo(si *sourceInfo) *StreamSourceInfo {
 	}
 
 	var trConfigs []SubjectTransformConfig
```
neilalexander, line 2018: Same here.

server/stream.go
```diff
@@ -3265,15 +3273,17 @@ func (mset *stream) startingSequenceForSources() {
 				mset.srv.Errorf("Unable to get subject transform for source: %v", err)
 			}
 		} else {
+			var sfs []string
```
neilalexander, line 3276: ... and here.

neilalexander (approved):
Otherwise LGTM.

server/stream.go

server/stream.go

server/stream.go

PR #4404, opened 2023-08-17
Tweak consumer replica scaling

neilalexander:
This should hopefully catch some consumer scaling situations more reliably. I've also added a test to test whether filtered consumers will scale properly even when the stream subject orphans them, which is what led me down this path to begin with.

### DIFF
--- a/server/consumer.go
+++ b/server/consumer.go
@@ -213,14 +213,13 @@ var (
 
 // Calculate accurate replicas for the consumer config with the parent stream config.
 func (consCfg ConsumerConfig) replicas(strCfg *StreamConfig) int {
-	if consCfg.Replicas == 0 {
+	if consCfg.Replicas == 0 || consCfg.Replicas > strCfg.Replicas {
 		if !isDurableConsumer(&consCfg) && strCfg.Retention == LimitsPolicy {
 			return 1
 		}
 		return strCfg.Replicas
-	} else {
-		return consCfg.Replicas
 	}
+	return consCfg.Replicas
 }
 
 // Consumer is a jetstream consumer.
--- a/server/jetstream_cluster.go
+++ b/server/jetstream_cluster.go
@@ -5996,11 +5996,11 @@ func (s *Server) jsClusteredStreamUpdateRequest(ci *ClientInfo, acc *Account, su
 		// Need to remap any consumers.
 		for _, ca := range osa.consumers {
 			// Ephemerals are R=1, so only auto-remap durables, or R>1, unless stream is interest or workqueue policy.
-			numPeers := len(ca.Group.Peers)
-			if ca.Config.Durable != _EMPTY_ || numPeers > 1 || cfg.Retention != LimitsPolicy {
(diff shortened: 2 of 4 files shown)


server/consumer.go
```diff
@@ -213,14 +213,13 @@ var (
 
 // Calculate accurate replicas for the consumer config with the parent stream config.
 func (consCfg ConsumerConfig) replicas(strCfg *StreamConfig) int {
-	if consCfg.Replicas == 0 {
+	if consCfg.Replicas == 0 || consCfg.Replicas > strCfg.Replicas {
 		if !isDurableConsumer(&consCfg) && strCfg.Retention == LimitsPolicy {
 			return 1
```
derekcollison, line 218: We only want this, default to 1, for the old school ephemerals, meaning replicas should be 0.

If it is not zero, but larger then the stream now, we should assume they picked it and we should not return 1 here, but what the new stream replicas is..

server/consumer.go
```diff
@@ -213,14 +213,13 @@ var (
 
 // Calculate accurate replicas for the consumer config with the parent stream config.
 func (consCfg ConsumerConfig) replicas(strCfg *StreamConfig) int {
-	if consCfg.Replicas == 0 {
+	if consCfg.Replicas == 0 || consCfg.Replicas > strCfg.Replicas {
 		if !isDurableConsumer(&consCfg) && strCfg.Retention == LimitsPolicy {
```
derekcollison, line 217: Can add this conditional at the end. `consCfg.Replicas == 0 `

derekcollison (changes requested):

server/jetstream_cluster.go
```diff
@@ -5996,11 +5996,11 @@ func (s *Server) jsClusteredStreamUpdateRequest(ci *ClientInfo, acc *Account, su
 		// Need to remap any consumers.
 		for _, ca := range osa.consumers {
 			// Ephemerals are R=1, so only auto-remap durables, or R>1, unless stream is interest or workqueue policy.
-			numPeers := len(ca.Group.Peers)
-			if ca.Config.Durable != _EMPTY_ || numPeers > 1 || cfg.Retention != LimitsPolicy {
+			replicas := ca.Config.replicas(cfg)
```
derekcollison, line 5999: Why the switch here?

neilalexander: It seemed clearer and less error-prone to check the intended number of replicas from the assignments rather than the Raft state. I expect the overall effect to be about the same but might be more robust during a Raft recovery.

derekcollison: The real number of peers is best represented by len(ca.Group.Peers).

derekcollison: I might suggest reverting this one unless there is a real bug that we have identified.

neilalexander: OK reverted!

derekcollison (approved):
LGTM

PR #4405, opened 2023-08-18
Send shutdown event on LDM

neilalexander:
If we send an event when entering lame duck mode, other nodes will mark the server as offline immediately, therefore R1 assets will not be placed onto that node. This is not a problem with R3 or higher because an LDM server operates as a Raft observer only and therefore cannot take the leadership role from an election, but R1 assets can in theory be placed onto any node that is not marked as offline.

A final shutdown event will still be sent when the server actually shuts down so there is no change there.

### DIFF
--- a/server/events.go
+++ b/server/events.go
@@ -533,6 +533,19 @@ RESET:
 	}
 }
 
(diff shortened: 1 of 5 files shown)


derekcollison (approved):
LGTM

PR #4407, opened 2023-08-18
Add filesystem locking on jetstream fileStore instances.

jevolk:
 - [X] Tests added
 - [X] Branch rebased on top of current ~main~ dev
 - [X] You have certified that the contribution is your original work and that you license the work to the project

### DIFF
--- a/server/filestore.go
+++ b/server/filestore.go
@@ -174,6 +174,7 @@ type fileStore struct {
 	hh          hash.Hash64
 	qch         chan struct{}
 	cfs         []ConsumerStore
+	lfd         *os.File
 	sips        int
 	closed      bool
 	fip         bool
@@ -385,6 +386,10 @@ func newFileStoreWithCreated(fcfg FileStoreConfig, cfg StreamConfig, created tim
 		prf:    prf,
 		oldprf: oldprf,
 		qch:    make(chan struct{}),
+	}
+
+	if err := fs.lockFileSystem(); err != nil {
+		return nil, err
 	}
 
 	// Set flush in place to AsyncFlush which by default is false.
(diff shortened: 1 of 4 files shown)


derekcollison (commented):
Possible to get a test written that shows correct behavior?

Also where to we clear the lock on successful shutdown of the filestore?

jevolk:
Thanks for taking the time to review this. 

> Also where to we clear the lock on successful shutdown of the filestore?

The lock is cleared when the file descriptor at `fileStore.lfd` is garbage collected, or when the process exits. I have since added an explicit `Close()` call in `fileStore.Stop()`. I have not added matching unlock `fcntl`'s but can do so at your request.

> Possible to get a test written that shows correct behavior?

POSIX record locks used here are "process-associated" which means testing within the same process or thread group will likely be insufficient. The `flock(2)` alternative is superior but not available on all platforms. AFAICT, the testing infrastructure here is limited to instances within the same address-space, and does not fork into sub-processes. I'm not yet sure how to test this here.

Let me know your thoughts, thanks.

jevolk:
Also, the CI appears limited to Linux only; I have not verified if the windows overload is effective.

derekcollison:
If you added the close to Stop() you can do a Stop() and make sure you can recreate the filestore with the same storedir..

jevolk:
By switching to `flock(2)` I have been able to add the proper tests, one affirmative (`TestFileStoreLockFileSystem`), one negative (`TestFileStoreLockFileSystemConflict`), which will work with the single address-space testing infrastructure here. This approach is superior because it can catch conflicts from two FileStore instances within the same server process as well as between two separate servers processes.

It appears that `syscall.Flock()` is made available by Go on all relevant targets. The caveat is that the behavior may be different depending on whether it's wrapping `fcntl(2)`. On systems with a wrapper, the negative test might fail, since the POSIX record lock (via `fcntl(2)`) does not contend within the same process; two FileStores instantiated within the same process will not exclude each other. So while `flock(2)` is superior if available, it might not actually be available.

Volumes have been written about how UNIX filesystem locking is a notorious shitshow, so please bear with me here.

PR #4411, opened 2023-08-21
[FIXED] Fixed deadlock when checkAndSync was being called as part of storing message

derekcollison:
We violated the locking pattern, so we now make sure we do this in a separate Go routine and put checks to only run it once.

### DIFF
--- a/server/jetstream.go
+++ b/server/jetstream.go
@@ -148,6 +148,9 @@ type jsAccount struct {
 	// From server
 	sendq *ipQueue[*pubMsg]
 
+	// For limiting only running one checkAndSync at a time.
+	sync atomic.Bool
+
 	// Usage/limits related fields that will be protected by usageMu
 	usageMu    sync.RWMutex
 	limits     map[string]JetStreamAccountLimits // indexed by tierName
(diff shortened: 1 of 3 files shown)


wallyqs (approved):
LGTM

server/jetstream.go
```diff
@@ -1811,6 +1814,12 @@ func (jsa *jsAccount) remoteUpdateUsage(sub *subscription, c *client, _ *Account
 // When we detect a skew of some sort this will verify the usage reporting is correct.
 // No locks should be held.
 func (jsa *jsAccount) checkAndSyncUsage(tierName string, storeType StorageType) {
+	// This will run in a separate go routine, so check that we are only running once.
+	if !jsa.sync.CompareAndSwap(false, true) {
+		return
+	}
+	defer jsa.sync.Store(true)
```
neilalexander, line 1821: Should the `defer` be back to `false`?

derekcollison: Yes good catch, thanks! Will update.

derekcollison: Fixed

neilalexander (approved):
LGTM!

PR #4412, opened 2023-08-21
[FIXED] Durable pull consumers could get cleaned up incorrectly on leader change.

derekcollison:
Fix for a bug that would allow old leaders of pull based durables to delete a consumer from an inactivity threshold timer inadvertently.


neilalexander (approved):
LGTM

PR #4414, opened 2023-08-22
Fix discarding explicit routes while removing duplicate ones

wallyqs:
In the new clustering logic for v2.10, sometimes the `TestStressChainedSolicitWorks` test would flake because a node would end up with only implicit routes. In this change, we stamp that one of the remotes is configured so that the nodes at least have one explicit configured remote node.


neilalexander (approved):
LGTM

derekcollison:
Asked @kozlovic to take a quick peek.

PR #4418, opened 2023-08-22
Fix JSON compatibility in conf format

wallyqs:
Fixes a side effect from #4394 which was breaking support for JSON configs.


derekcollison (approved):
LGTM

PR #4427, opened 2023-08-24
[FIXED] Consumer ack pending > max ack pending on restart or leader change

derekcollison:
When a consumer reached a max delivered condition, we did not properly synchronize the state such that on a restore or leader switch the ack pending could jump and be higher than max ack pending and block the consumer.

This propagates a delivered update and we updated the store state engine to do the right thing when the condition is reached.


wallyqs (approved):
LGTM

PR #4430, opened 2023-08-25
Fix shutdown deadlock in TestJetStreamClusterMemLeaderRestart

wallyqs:
While shutting down a server an error during purge from a memory stream would cause a deadlock sometimes, this would sometimes show up in the `TestJetStreamClusterMemLeaderRestart` while tearing down the cluster.

This was introduced in https://github.com/nats-io/nats-server/commit/4d8d01949b04e20d6ec167e195504a2dcc3a87e2 so only relates to v2.10.


Jarema (approved):
LGTM!

PR #4431, opened 2023-08-25
[FIX] PurgeEx with keep and deleted bug 

derekcollison:
Fix for purge with keep bug with user deletes and improved search for large number of blocks.


wallyqs (approved):
LGTM

PR #4432, opened 2023-08-25
[FIXED] NRG layer could loop if leader asked for an item and we have none.

derekcollison:
If a leader is asked for an item and we have no items left, make sure to also step-down.


wallyqs (approved):
LGTM

PR #4433, opened 2023-08-25
Run tests using Go 1.21

wallyqs:
Flips the order to test with ~~Go 1.20~~ Go 1.21 instead of Go 1.19


derekcollison (approved):
LGTM

wallyqs:
Updated to test with Go 1.21 mainly instead.

PR #103, opened 2015-08-14
Add JSONP support for monitoring routes

cmfatih:
See #100

PR #4437, opened 2023-08-26
Fix Content-Type header in /healthz when status is not 200 OK

### LINKED ISSUE #4436: HTTP headers are not set for the monitoring server when the status is not 200 OK

nats-server: v2.9.21

Linux amd64

All the monitoring server endpoints return `http.StatusOK` except the `/healthz` endpoint when the health check fails.

Requests made to the `/healthz` endpoint when the health check fails will receive responses with the default HTTP headers, for example the `Content-Type` header will be set to the value `text/plain; charset=utf-8` instead of `application/json` and `application/javascript` for JSONP.

This is happening because setting the headers on an `http.ResponseWriter` has no effect after setting the status code with `ResponseWriter.WriteHeader`.

The function [`ResponseHandler()`](https://github.com/mdawar/nats-server/blob/5b18e80d424926eca48091c6eb2e2a56fdffdc5d/server/monitor.go#L2305) sets the headers, but when the health check fails, the function [`HandleHealthz()`](https://github.com/mdawar/nats-server/blob/5b18e80d424926eca48091c6eb2e2a56fdffdc5d/server/monitor.go#L3080) calls `WriteHeader()` before calling `ResponseHandler()` which causes this issue.

I will add tests that reproduce this issue in a pull request.

The HTTP headers for the `/healthz` endpoint should be set properly when the health check fails.

Setting the headers has no effect after calling `ResponseWriter.WriteHeader()`.

mdawar:
- Added a new internal function `handleResponse` that accepts the HTTP status code and sets it after setting the headers
 - Added tests for the `/healthz` endpoint for the `ok`, `error` and `unavailable` statuses
 - Changed the IETF API health check URL to https://datatracker.ietf.org/doc/html/draft-inadarei-api-health-check (Previous link broken)

The function [`ResponseHandler`](https://github.com/mdawar/nats-server/blob/5b18e80d424926eca48091c6eb2e2a56fdffdc5d/server/monitor.go#L2305) that handles the monitoring server responses is exported so I had to add another internal function to prevent a breaking change to the API.
I don't know if dependent code should use this function, maybe in the next major version this function should be unexported because it's only used for the monitoring endpoints.

Resolves #4436

### DIFF
--- a/server/monitor.go
+++ b/server/monitor.go
@@ -2301,19 +2301,26 @@ func (s *Server) HandleAccountStatz(w http.ResponseWriter, r *http.Request) {
 	ResponseHandler(w, r, b)
 }
 
-// ResponseHandler handles responses for monitoring routes
+// ResponseHandler handles responses for monitoring routes.
 func ResponseHandler(w http.ResponseWriter, r *http.Request, data []byte) {
+	handleResponse(http.StatusOK, w, r, data)
+}
+
+// handleResponse handles responses for monitoring routes with a specific HTTP status code.
+func handleResponse(code int, w http.ResponseWriter, r *http.Request, data []byte) {
 	// Get callback from request
 	callback := r.URL.Query().Get("callback")
 	// If callback is not empty then
 	if callback != "" {
 		// Response for JSONP
 		w.Header().Set("Content-Type", "application/javascript")
+		w.WriteHeader(code)
 		fmt.Fprintf(w, "%s(%s)", callback, data)
 	} else {
 		// Otherwise JSON
 		w.Header().Set("Content-Type", "application/json")
 		w.Header().Set("Access-Control-Allow-Origin", "*")
+		w.WriteHeader(code)
 		w.Write(data)
 	}
 }
@@ -3048,7 +3055,7 @@ type HealthStatus struct {
 	Error  string `json:"error,omitempty"`
 }
 
-// https://tools.ietf.org/id/draft-inadarei-api-health-check-05.html
+// https://datatracker.ietf.org/doc/html/draft-inadarei-api-health-check
 func (s *Server) HandleHealthz(w http.ResponseWriter, r *http.Request) {
 	s.mu.Lock()
 	s.httpReqStats[HealthzPath]++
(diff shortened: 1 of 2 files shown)


bruth (approved):
LGTM, thanks for the contribution!

server/monitor.go
```diff
@@ -2301,19 +2301,26 @@ func (s *Server) HandleAccountStatz(w http.ResponseWriter, r *http.Request) {
 	ResponseHandler(w, r, b)
 }
 
-// ResponseHandler handles responses for monitoring routes
+// ResponseHandler handles responses for monitoring routes.
 func ResponseHandler(w http.ResponseWriter, r *http.Request, data []byte) {
+	handleResponse(http.StatusOK, w, r, data)
```
wallyqs, line 2306: Note that we currently do not use this handler in most monitoring APIs when serving error responses (one of the reasons why we missed setting header right in /healthz) https://github.com/nats-io/nats-server/blob/e5836fc98dd7e5cada52ca23339bd3d5fcb38eaf/server/monitor.go#L728-L732

server/monitor.go
```diff
@@ -2301,19 +2301,26 @@ func (s *Server) HandleAccountStatz(w http.ResponseWriter, r *http.Request) {
 	ResponseHandler(w, r, b)
 }
 
-// ResponseHandler handles responses for monitoring routes
+// ResponseHandler handles responses for monitoring routes.
```
wallyqs, line 2304: It was not necessary to have this exported but it has been public for quite some time already, introduced in https://github.com/nats-io/nats-server/pull/103/files

server/monitor_test.go
```diff
@@ -4775,3 +4775,119 @@ func TestMonitorAccountszMappingOrderReporting(t *testing.T) {
...
+func expectHealthStatus(t *testing.T, url string, statusCode int, wantStatus string) {
+	t.Helper()
+
+	// First check for regular requests.
+	body := readBodyEx(t, url, statusCode, appJSONContent)
+	checkHealthStatus(t, body, wantStatus)
+
+	// Another check for JSONP requests.
+	jsonpURL := createCallbackURL(t, url) // Adds a callback query param.
+	jsonpBody := readBodyEx(t, jsonpURL, statusCode, appJSContent)
```
wallyqs, line 4821: as you may have found `readBodyEx` currently uses misses the line of the error which can be fixed by using `t.Helper`, fixing that in a separate PR when including this change

wallyqs (approved):
LGTM, thanks for the contribution. I will make a follow up PR to include this into v2.10 dev branch here: https://github.com/nats-io/nats-server/compare/dev...healthz-mon-fixes

server/monitor.go
mdawar: BTW shouldn't error responses be also returned as JSON for more consistency? Maybe in the next major version? Because this change might break the current consumers of the API.

PR #4439, opened 2023-08-27
[FIXED] Expire on recover could not update global per subject map

derekcollison:
When expiring complete blocks on recover make sure to update global subject index psim.


wallyqs (approved):
LGTM

PR #4440, opened 2023-08-28
[FIXED] MQTT: Removed the use of tkDomain from retained msg subjects

levb:
(Partially?) addresses https://github.com/nats-io/nats-server/pull/4349#discussion_r1306576048

@kozlovic @neilalexander I did not remove the use of `domainTk` in the session subject since it seems to have significance to it; removing it failed `TestMQTTSessionsDifferentDomains` and I did not understand the specifics of the issue enough. Please let me know your thoughts.

### DIFF
--- a/server/mqtt.go
+++ b/server/mqtt.go
@@ -1154,7 +1154,7 @@ func (s *Server) mqttCreateAccountSessionManager(acc *Account, quitCh chan struc
 		// Create the stream for retained messages.
 		cfg := &StreamConfig{
 			Name:       mqttRetainedMsgsStreamName,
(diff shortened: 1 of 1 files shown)


kozlovic (approved):
LGTM. But to be perfectly clear, leaving the domain here does not "break" anything (just makes the subjects longer). However this change (remove the domain) may break users that have used the "dev" branch against existing MQTT streams - and are using domains. I would assume that those users did not do that against production data, but that is something to keep in mind.

neilalexander (approved):
LGTM

PR #4443, opened 2023-08-29
MQTT: Cleanup code regarding retain flag and add test

kozlovic:
As per specification MQTT-3.3.1-8, we are now setting the RETAIN flag when delivering to new subscriptions and clear the flag in all other conditions.

### DIFF
--- a/server/mqtt.go
+++ b/server/mqtt.go
@@ -2341,6 +2341,7 @@ func (as *mqttAccountSessionManager) serializeRetainedMsgsForSub(sess *mqttSessi
 
 		// Need to use the subject for the retained message, not the `sub` subject.
 		// We can find the published retained message in rm.sub.subject.
+		// Set the RETAIN flag: [MQTT-3.3.1-8].
 		flags := mqttSerializePublishMsg(prm, pi, qos, false, true, []byte(rm.Topic), rm.Msg)
 		if trace {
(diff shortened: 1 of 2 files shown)


derekcollison:
This only for dev or should be for main too?

kozlovic:
@derekcollison There have been "many" other fixes in Lev's PR that was against dev branch that we felt that this fix is ok to be in dev only. That being said, I can make it against "main" branch but there would be conflict merging into "dev" due to recent changes.

derekcollison:
If applicable to main yes let's do that..

kozlovic:
> If applicable to main yes let's do that..

Ok will do so in a bit.

kozlovic:
@derekcollison @levb After further review, there was no bad behavior in that the retain flag was cleared (as expected) when processing a retained message, which means that the code in the "QoS0" callback that used the retain flag value to serialize the message to an existing subscription would have always be "false" (as it should). For the "QoS1" callback, the retain boolean was declared but never set, so again the end-result was correct behavior.

I have changed the title of this PR to make it clear that this is more of a code cleanup, no defect is being fixed. I therefore propose to leave this against "dev" and not have a PR for "main" since there was no defect per-se.

derekcollison (approved):
LGTM

PR #4444, opened 2023-08-29
Fixed a route pooling flapper

kozlovic:

### DIFF
--- a/server/routes_test.go
+++ b/server/routes_test.go
@@ -3129,7 +3129,7 @@ func TestRoutePoolWithOlderServerConnectAndReconnect(t *testing.T) {
 	// We could get some, but it should settle.
 	checkRepeatConnect := func() {
 		t.Helper()
-		tm := time.NewTimer(2 * DEFAULT_ROUTE_CONNECT)
+		tm := time.NewTimer(4 * routeConnectDelay)
 		var last time.Time
 		for done := false; !done; {
 			select {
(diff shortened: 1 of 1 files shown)


derekcollison (approved):
LGTM - Thanks!

kozlovic:
@wallyqs Here is my attempt at fixing the flapper you have mentioned to me. I increased the number of attempts we let the test settle but use routeConnectDelay, which in test is 15ms instead of the constant that is 1sec. I did not see it flap with these changes on my machine.

wallyqs (approved):
LGTM, thanks Ivan!!

kozlovic:
@derekcollison @wallyqs I have seen TestRoutePoolConnectRace fail on Travis (not on my machine so far). Having a look on that one too. Could we keep this PR opened a bit to see if I have an update?

kozlovic:
@derekcollison @wallyqs I modified the test that was failing. The number of duplicate logs was too small given the size of the cluster and pool size. I have verified that if I make the code "fail" (by not preventing the reconnect when detected as duplicate), the test will fail even with the new limit of 20. With correct code, I have seen in my tests that the number of "duplicate" log messages could go to 16 or so in some cases but will then stop. So the new limit of 20 seem ok but could be increased easily since it would not prevent detecting the bug (of constant reconnect).

wallyqs (approved):
LGMT, thanks!

PR #4446, opened 2023-08-29
[FIXED] Max msgs per subject config update to lower values

### LINKED ISSUE #4445: A KeyValue stream's max messages per subject isn't being respected after updating the stream's max messages per subject to be smaller than before

`2.9.21`

Mac M1

Yes, this is reproducible. Run a nats server with jetstream enabled and then run the following test:
```
package history_test

import (
	"encoding/json"
	"strconv"
	"testing"

	"github.com/nats-io/nats.go"
	"github.com/stretchr/testify/require"
)

func TestHistory(t *testing.T) {
	// Start server with `nats-server --js` to make sure jetstream is turned on
	nc, err := nats.Connect(nats.DefaultURL)
	require.Nil(t, err)

	// Create jetstream and KeyValue
	js, err := nc.JetStream()
	require.Nil(t, err)
	bucketName := "HistoryTest"
	kv, err := js.CreateKeyValue(&nats.KeyValueConfig{
		Bucket:      bucketName,
		Description: "Test for number of historical values",
		History:     10,
	})
	require.Nil(t, err)
	key := "mykey"

	// Write to the same key 50 times
	for j := 0; j < 50; j++ {
		value := strconv.Itoa(j)
		marshalledBytes, err := json.Marshal(value)
		require.Nil(t, err)
		_, err = kv.Put(key, marshalledBytes)
		require.Nil(t, err)
	}
	// Verify we have the correct amount of history
	info, err := js.StreamInfo("KV_" + bucketName)
	require.Nil(t, err)
	histories, err := kv.History(key)
	require.Nil(t, err)
	require.Len(t, histories, int(info.Config.MaxMsgsPerSubject))

	// Update the stream to store half the amount of history as before
	newConfig := info.Config
	newConfig.MaxMsgsPerSubject = newConfig.MaxMsgsPerSubject / 2
	info, err = js.UpdateStream(&newConfig)
	require.Nil(t, err)

	// Write 50 more values
	for j := 0; j < 50; j++ {
		value := strconv.Itoa(50 + j)
		marshalledBytes, err := json.Marshal(value)
		require.Nil(t, err)
		_, err = kv.Put(key, marshalledBytes)
		require.Nil(t, err)
	}
	// Verify that the history has the updated number of max messages per subject
	histories, err = kv.History(key)
	require.Nil(t, err)
	require.Len(t, histories, int(info.Config.MaxMsgsPerSubject)) // We fail here...
}

```

My expectation is that after updating the Max Messages Per Subject for the Stream, the corresponding KeyValue should contain the updated amount of history

After updating the stream associated with the KV, it is storing more history than it should be storing.

derekcollison:
We were not recalculating first correctly since we were not considering seq < mb.first.seq.

Resolves #4445

### DIFF
--- a/server/filestore.go
+++ b/server/filestore.go
@@ -451,7 +451,7 @@ func (fs *fileStore) UpdateConfig(cfg *StreamConfig) error {
 		fs.ageChk = nil
 	}
 
-	if cfg.MaxMsgsPer > 0 && cfg.MaxMsgsPer < old_cfg.MaxMsgsPer {
+	if fs.cfg.MaxMsgsPer > 0 && fs.cfg.MaxMsgsPer < old_cfg.MaxMsgsPer {
 		fs.enforceMsgPerSubjectLimit()
 	}
 	fs.mu.Unlock()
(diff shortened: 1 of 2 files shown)


wallyqs (approved):
LGTM

PR #4450, opened 2023-08-30
Updates to the way meta indexing is handled for filestore.

derekcollison:
Historically we kept indexing information, either by sequence or by subject, as a per msg block operation. These were the "*.idx" and "*.fss" indexing files. When streams became very large this could have an impact on recovery time. Also, for encryption the fast path for determining if the indexing was current would require loading and decrypting the complete block.

This design moves to a more traditional WAL and snapshot approach. The snapshots for the complete stream, including summary information, global per subject information maps (PSIM) and per msg block details including summary and dmap, are processed asynchronously. The snapshot includes the msg block and has for the last record hash that was considered in the snapshot. On recovery the snapshot is read and processed and any additional records past the point of the snapshot itself are processed. To this end, any non-system removal of a message has to be expressed as a delete tombstone that is always added the the fs.lmb file. These are processed on recovery and our indexing layer knows to skip them.

Changing to this method drastically improves startup and recovery times, and has simplified the code. Some normal performance benefits have been seen as well.

### DIFF
--- a/server/filestore.go
+++ b/server/filestore.go
@@ -18,6 +18,7 @@ import (
 	"bytes"
 	"crypto/aes"
 	"crypto/cipher"
+	"crypto/rand"
 	"crypto/sha256"
 	"encoding/binary"
 	"encoding/hex"
@@ -34,8 +35,6 @@ import (
 	"sync"
 	"sync/atomic"
 	"time"
-
-	crand "crypto/rand"
 
 	"github.com/klauspost/compress/s2"
 	"github.com/minio/highwayhash"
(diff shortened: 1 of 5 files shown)


server/filestore.go
```diff
@@ -415,9 +413,64 @@ func newFileStoreWithCreated(fcfg FileStoreConfig, cfg StreamConfig, created tim
 		}
 	}
 
-	// Recover our message state.
-	if err := fs.recoverMsgs(); err != nil {
-		return nil, err
+	// Attempt to receover our state.
```
neilalexander, line 416: `recover`

neilalexander (approved):
LGTM, I really like the approach here.

Some items for 2.11 out of this:
1. Clean-up of accounting, quite a lot of duplicated code
2. Clean-up of key loading from disk

wallyqs (approved):
LGTM

server/filestore.go
derekcollison: Fixed thx

PR #4452, opened 2023-08-30
[IMPROVED] Allow 2.10 tombstones to be skipped and allow us to recover on downgrade

derekcollison:
Also fixed small bug that could set bad first sequence for subject tracking info.


wallyqs (approved):
LGTM

PR #4454, opened 2023-08-30
[FIXED] potential message duplication from sources when downgrading back from 2.10

jnmoyne:

2.10 adds a couple space separated fields to the sourcing message header from 2 to 4 but the current 2.9 code is too strict of checking the number of fields is exactly 2 rather than at least 2


derekcollison (approved):
LGTM

PR #4458, opened 2023-08-31
Fix lock issue in filestore

neilalexander:
This should hopefully fix a panic on unlock of unlocked mutex in the file store.

### DIFF
--- a/server/filestore.go
+++ b/server/filestore.go
@@ -3509,7 +3509,7 @@ func (fs *fileStore) removeMsg(seq uint64, secure, viaLimits, needFSLock bool) (
 	}
 
 	cb := fs.scb
-	fs.mu.Unlock()
+	fsUnlock()
 
 	// Storage updates.
 	if cb != nil {


server/filestore.go
```diff
@@ -3509,7 +3509,7 @@ func (fs *fileStore) removeMsg(seq uint64, secure, viaLimits, needFSLock bool) (
 	}
 
 	cb := fs.scb
-	fs.mu.Unlock()
+	fsUnlock()
```
ripienaar, line 3512: It's worth calling fsLock() on `3524` for consistency and to avoid introducing future bugs should the `fsLock` and `fsUnlock` pair get new behaviors, one would easily assume updating those functions would update the entire func behavior.

ripienaar (changes requested):
small change

server/filestore.go
ripienaar: Oh, I see the logic there is actually different. So ignore this comment sorry.

ripienaar (approved):
LGTM

derekcollison:
Let me take a look, for some reason the fix does not look right, but just getting up.

derekcollison (approved):
LGTM

derekcollison:
Actually introduced a deadlock of a more serious form, a direct pushed to dev to fix.

PR #4461, opened 2023-08-31
[FIXED] Unlock needed to be guarded, could deadlock filestore

derekcollison:
Needed to check guard for unlock here.


wallyqs (approved):

neilalexander (approved):
LGTM

PR #4464, opened 2023-08-31
Fix leaf connection missing LS+ sometimes

wallyqs:
`TestNoRaceLeafNodeSmapUpdate` could occasionally fail with missing `LS+` commands due not capturing all the inflight SUB commands as they were being processed outside the client lock.


derekcollison (approved):
LGTM! Thanks to you both.

PR #3326, opened 2022-08-03
Added param options to /healthz endpoint

mfaizanse:
- [x] Link to issue, e.g. `Resolves #NNN`
 - [x] Documentation added (if applicable)
 - [x] Tests added
 - [x] Branch rebased on top of current main (`git pull --rebase origin main`)
 - [x] Changes squashed to a single commit (described [here](http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html))
 - [x] Build is green in Travis CI
 - [x] You have certified that the contribution is your original work and that you license the work to the project under the [Apache 2 license](https://github.com/nats-io/nats-server/blob/main/LICENSE)

Related to # https://github.com/nats-io/k8s/issues/541
Resolves # The JetStream portion of https://github.com/nats-io/nats-server/issues/2687

### Changes proposed in this pull request:

 - Added `/healthz?js-enabled=true` to return error if JetStream is disabled.
 - Added `/healthz?js-server-only=true` to skip checking of JetStream accounts, streams and consumers.

/cc @nats-io/core

PR #4467, opened 2023-08-31
Fixes to /healthz response (v2.10)

wallyqs:
Follow up from #4437 content-type fix for v2.9.22, some fixes to the response from `/healthz` for dev:

- In #[3326](https://github.com/nats-io/nats-server/pull/4097) it was changed to return 500 status when before we used to return 503 so this changes it back.
- Also as part of #3326 we started to return `status_code` in the healthz response (e.g `{"status":"ok","status_code":200}`) so this removes it for http responses just relying on the http header.

### DIFF
--- a/server/monitor.go
+++ b/server/monitor.go
@@ -2339,7 +2339,6 @@ func ResponseHandler(w http.ResponseWriter, r *http.Request, data []byte) {
 func handleResponse(code int, w http.ResponseWriter, r *http.Request, data []byte) {
 	// Get callback from request
 	callback := r.URL.Query().Get("callback")
-	// If callback is not empty then
 	if callback != "" {
 		// Response for JSONP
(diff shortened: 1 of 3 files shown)


server/monitor.go
```diff
@@ -2339,7 +2339,6 @@ func ResponseHandler(w http.ResponseWriter, r *http.Request, data []byte) {
 func handleResponse(code int, w http.ResponseWriter, r *http.Request, data []byte) {
 	// Get callback from request
 	callback := r.URL.Query().Get("callback")
-	// If callback is not empty then
 	if callback != "" {
```
derekcollison, line 2342: _EMPTY_

derekcollison (approved):
LGTM

Small nit, and make sure tests can be green.

PR #4468, opened 2023-08-31
[FIXED] Sending too fast to have replicas be caught up enough to register directs.

derekcollison:


wallyqs (approved):
LGTM

PR #4472, opened 2023-09-01
[IMPROVED] Consumer failing to deliver re-adjusts delivered count and any waiting request.

derekcollison:
When we fail to deliver a message for a consumer, either through didNotDeliver() or LoadMsg() failure re-adjust delivered count and waitingRequest accounting.

### DIFF
--- a/server/consumer.go
+++ b/server/consumer.go
@@ -3530,11 +3530,13 @@ func (o *consumer) loopAndGatherMsgs(qch chan struct{}) {
 			if err == ErrStoreMsgNotFound || err == errDeletedMsg || err == ErrStoreEOF || err == errMaxAckPending {
 				goto waitForMsgs
 			} else if err == errPartialCache {
-				s.Warnf("Unexpected partial cache error looking up message for consumer")
+				s.Warnf("Unexpected partial cache error looking up message for consumer '%s > %s > %s'",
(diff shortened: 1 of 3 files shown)


server/consumer.go
```diff
@@ -3508,8 +3527,12 @@ func (o *consumer) loopAndGatherMsgs(qch chan struct{}) {
 			if err == ErrStoreEOF {
 				o.checkNumPendingOnEOF()
 			}
-			if err == ErrStoreMsgNotFound || err == errDeletedMsg || err == ErrStoreEOF || err == errMaxAckPending || err == errPartialCache {
+			if err == ErrStoreMsgNotFound || err == errDeletedMsg || err == ErrStoreEOF || err == errMaxAckPending {
+				goto waitForMsgs
+			} else if err == errPartialCache {
+				s.Warnf("Unexpected partial cache error looking up message for consumer")
```
neilalexander, line 3533: Can we put the consumer name in this error?

neilalexander (approved):
Otherwise LGTM

server/consumer.go
derekcollison: Good idea, should we do same for err below?

derekcollison: ok pushed second CL for that. Take a look.

neilalexander: Yep looks good!

PR #4478, opened 2023-09-02
[FIXED] Fix for a bug that would make normal streams use the wrong block size.

derekcollison:


neilalexander (approved):
LGTM

PR #4481, opened 2023-09-03
Filestore tweaks and improvements

derekcollison:
Optimized for startup restore time.

### DIFF
--- a/server/filestore.go
+++ b/server/filestore.go
@@ -433,7 +433,9 @@ func newFileStoreWithCreated(fcfg FileStoreConfig, cfg StreamConfig, created tim
 		// Check if our prior remember a last past where we can see.
 		if fs.ld != nil && prior.LastSeq > fs.state.LastSeq {
 			fs.state.LastSeq, fs.state.LastTime = prior.LastSeq, prior.LastTime
-			if _, err = fs.newMsgBlockForWrite(); err != nil {
+			if lmb, err := fs.newMsgBlockForWrite(); err == nil {
+				lmb.writeTombstone(prior.LastSeq, prior.LastTime.UnixNano())
+			} else {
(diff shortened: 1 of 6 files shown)


server/filestore.go
```diff
@@ -1586,54 +1593,42 @@ func (fs *fileStore) recoverFullState() (rerr error) {
...
-
 	// First let's check the happy path, open the blk file that was the lmb when we created the full state.
 	// See if we have the last block available.
 	var matched bool
 	var mb *msgBlock
 	if mb = fs.bim[blkIndex]; mb != nil {
-		matched = bytes.Equal(mb.lastChecksum(), lchk[:])
-		if matched && blkIndex == blkFiles.maxIndex {
-			return nil
+		if _, err := os.Stat(mb.mfn); err != nil && os.IsNotExist(err) {
```
neilalexander, line 1604: Is it safe to drop other `err` here?

server/filestore.go
```diff
@@ -1586,54 +1593,42 @@ func (fs *fileStore) recoverFullState() (rerr error) {
...
-	start, stop := blkIndex, blkFiles.maxIndex
+	// We may need to check other blocks. Even if we matched last checksum we will see if there is another block.
+	// If we did not match we re-process the last block.
+	start := blkIndex
 	if matched {
 		start++
 	}
 
-	for bi := start; bi <= stop; bi++ {
+	for bi := start; ; bi++ {
```
neilalexander, line 1625: Minor nit but `for bi := start; true; bi++ {` might be clearer.

derekcollison: I believe that is the way Go recommends it so will leave as is.

neilalexander (approved):
LGTM

PR #4483, opened 2023-09-04
Added ability to control sync intervals and sync always

derekcollison:

### DIFF
--- a/server/filestore.go
+++ b/server/filestore.go
@@ -52,6 +52,8 @@ type FileStoreConfig struct {
 	CacheExpire time.Duration
 	// SyncInterval is how often we sync to disk in the background.
 	SyncInterval time.Duration
+	// SyncAlways is when the stream should sync all data writes.
+	SyncAlways bool
 	// AsyncFlush allows async flush to batch write operations.
 	AsyncFlush bool
 	// Cipher is the cipher to use when encrypting.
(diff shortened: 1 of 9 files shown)


server/filestore.go
```diff
@@ -186,42 +188,44 @@ type fileStore struct {
...
+	ctmr       *time.Timer
+	werr       error
+	dmap       avl.SequenceSet
+	fch        chan struct{}
+	qch        chan struct{}
+	lchk       [8]byte
+	loading    bool
+	flusher    bool
+	noTrack    bool
+	needSync   bool
```
neilalexander, line 226: AFAICT `needSync` is never not `false`?

ripienaar:
Be good if this setting is surfaced somewhere like `JetStreamVarz.Config` fetched via varz

server/filestore.go
derekcollison: Thanks good catch, was getting late..

Fixed in separate CL but once LGTM I will squash.

derekcollison:
Will make sure its returned.

neilalexander (approved):
LGTM

derekcollison:
ok added in the jsz / varz and additional test to watch needSync flag etc.

neilalexander (approved):
LGTM

PR #4484, opened 2023-09-04
[FIXED] MQTT: more generic names for outgoing stream, etc.

levb:
- [x] Branch rebased on top of current main (`git pull --rebase origin dev`)

Removed `qos2` reference from the MQTT outgoing stream name and subjects.

Rationale: As I was reviewing the documentation and thinking how to move forward with MQTT, I am inclined to use the `$MQTT_out` stream and session-specific consumers (as opposed to subscription-specific) for all QOS-related packet delivery and PI persistence. I think I will be able to repurpose it "as is", so proposing the generic name change.

Since these (stream, consumer, subscriptions) are newly introduced in https://github.com/nats-io/nats-server/pull/4349 and have not yet been released, this is a safe change.


derekcollison (approved):
LGTM

PR #4486, opened 2023-09-04
Use `TempDir` in more tests

neilalexander:
This uses `TempDir` in more tests. Some were trying to manage the directory manually, others were just not managing a directory at all. There may be more that need doing.


wallyqs (approved):
LGTM

PR #4487, opened 2023-09-04
v2.9.x: Change CI and nightlies to use Go 1.20

wallyqs:
Includes fix for test `TestGatewayTLSMixedIPAndDNS ` that was breaking in Go 1.20.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,6 @@ vm:
 language: go
 go:
   - "1.20.x"
-  - "1.19.12"
 
 go_import_path: github.com/nats-io/nats-server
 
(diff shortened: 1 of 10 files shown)


.travis.yml
```diff
@@ -6,7 +6,9 @@ vm:
 
 language: go
 go:
+  - "1.20.x"
   - "1.19.12"
```
derekcollison, line 10: Do we want to double our builds here?

derekcollison (changes requested):
I would replace with Go 1.20, since 1.19 deprecated would not run 2 builds.
Otherwise LGTM.

.travis.yml
wallyqs: Updated to removed v1.19 from the build

derekcollison (approved):
LGTM

PR #4490, opened 2023-09-06
v2.10: Use Go 1.21 in nightlies, update rand usage

wallyqs:
- Use Go 1.21 in nightlies
- Both rand.Seed and rand.Read are both deprecated, remove its use to fix staticcheck errors

```go
server/client.go:95:2: SA1019: rand.Seed has been deprecated since Go 1.20 and an alternative 
  has been available since Go 1.0: As of Go 1.20 there is no reason to call Seed with a random value. 
  Programs that call Seed with a known value to get a specific sequence of results should use 
  New(NewSource(seed)) to obtain a local random generator. (staticcheck)

server/jetstream_test.go:20399:2: SA1019: rand.Read has been deprecated since Go 1.20 
  because it shouldn't be used: For almost all use cases, crypto/rand.Read is more appropriate. (staticcheck)
```

### DIFF
--- a/.github/workflows/cov.yaml
+++ b/.github/workflows/cov.yaml
@@ -23,7 +23,7 @@ jobs:
       - name: Set up Go
         uses: actions/setup-go@v4
         with:
-          go-version: "1.20.x"
+          go-version: "1.21.x"
 
       - name: Run code coverage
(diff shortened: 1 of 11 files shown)


neilalexander (approved):
LGTM

PR #4492, opened 2023-09-06
Fix race condition in hashing stream snapshots

neilalexander:
The `streamSnapshot` function was not holding the filestore lock while resetting and using the hasher. This created a data race and also resulted in flakes in `TestFileStoreSnapshot`.


derekcollison (approved):
LGTM

PR #4495, opened 2023-09-06
Revert "Enables 0s deduplication window duration when the stream has sources"

wallyqs:
This reverts commit db96238ad9b18c9930e63c404d1ae288f8855374 which was causing `TestNoRaceJetStreamSuperClusterSources` to fail sometimes with duplicate messages around shutdown.


derekcollison (approved):
LGTM

PR #4498, opened 2023-09-07
Use write lock for memory store filtered state

neilalexander:
Calls to `filteredStateLocked` can mutate the per-subject state by recalculating the first sequences. Doing so under a read-only lock can lead to race conditions, so switch the call sites out to write locks instead.


wallyqs (approved):
LGTM

PR #4501, opened 2023-09-07
Don't log error in legitimate nkey auth callout config case (4479).

### LINKED ISSUE #4479: Authorization callout user not valid (when having users in nats options)

Latest from main as of 2023-09-02.

dev box Ubuntu

Yes. Take a look at auth.go:293
```go
for _, u := range opts.AuthCallout.AuthUsers {
			// Check for user in users and nkeys since this is server config.
			var found bool
			if len(s.users) > 0 {
				_, found = s.users[u]
			} else if len(s.nkeys) > 0 && !found {
				_, found = s.nkeys[u]
			}
			if !found {
				s.Errorf("Authorization callout user %q not valid: %v", u, err)
			}
		}
```
The statement:
> else if len(s.nkeys) > 0 && !found {

is never run when len(s.users) > 0.

This should simply be an "if len()..." without the else.

No errors should be reported if the callout user exists in nkeys

I'm observing an incorrect error in the log. The callout nkey does exist, but the check doesn't check the nkeys.

tbeets:
`Resolves #4479 `

In AuthCallout server-config system, a spurious error message would be generated on server reload in the edge case that the AuthCallout user is an NKEY and both User and NKEY account members are defined in the config.

### DIFF
--- a/server/auth.go
+++ b/server/auth.go
@@ -28,10 +28,11 @@ import (
 	"sync/atomic"
 	"time"
 
+	"golang.org/x/crypto/bcrypt"
+
 	"github.com/nats-io/jwt/v2"
 	"github.com/nats-io/nats-server/v2/internal/ldap"
 	"github.com/nats-io/nkeys"
-	"golang.org/x/crypto/bcrypt"
 )
 
 // Authentication is an interface for implementing authentication
@@ -279,7 +280,7 @@ func (s *Server) configureAuthorization() {
 	// Check for server configured auth callouts.
 	if opts.AuthCallout != nil {
 		s.mu.Unlock()
-		// Make sure we have a valid account and auth_users.
+		// Give operator log entries if not valid account and auth_users.
 		_, err := s.lookupAccount(opts.AuthCallout.Account)
 		s.mu.Lock()
 		if err != nil {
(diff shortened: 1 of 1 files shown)


aricart (approved):
LGTM

neilalexander (approved):
LGTM

server/auth.go
```diff
@@ -28,10 +28,11 @@ import (
 	"sync/atomic"
 	"time"
 
+	"golang.org/x/crypto/bcrypt"
+
```
derekcollison, line 32: Remove newline here.

derekcollison (commented):
LGTM - one nit, then we can merge.

server/auth.go
tbeets: I'm using the `goimports -local github.com/nats.io/` style where it [groups our project imports separately from stdlib and everything else](https://stackoverflow.com/questions/38704769/golang-import-grouping-by-package).

Do you specifically just want the stdlib and everything else (including our stuff) style?

derekcollison: ok usually we put these at bottom with a single line space.

derekcollison (approved):
LGTM

PR #4505, opened 2023-09-08
Allow more time in `TestNoRaceSeqSetEncodeLarge`

neilalexander:
In CI this is typically < 700uS, but quite often > 500uS.


wallyqs (approved):
LGTM

PR #4515, opened 2023-09-12
Allow more time in `TestFileStoreNumPendingLargeNumBlks`

neilalexander:
This should hopefully de-flake `TestFileStoreNumPendingLargeNumBlks` a bit. By adding a new `require_LessThan` helper function, we also will produce a more meaningful log line that tells us what the bad values were if the test fails.


server/test_test.go
```diff
@@ -133,6 +135,13 @@ func require_Len(t *testing.T, a, b int) {
 	}
 }
 
+func require_LessThan[T constraints.Ordered](t *testing.T, a, b T) {
```
wallyqs, line 138: In a future release when Go 1.22 is out we should change this to be `cmp.Ordered` which would be included since Go 1.21? https://pkg.go.dev/cmp@master#Ordered

wallyqs (approved):
LGTM

PR #4517, opened 2023-09-12
[FIXED] Increased MQTT test R/W timeout from 4 to 5s (TestMQTTSubPropagation)

levb:
Tracing the connect (ack?) read times in `TestMQTTSubPropagation` showed that they come in the 2-3s range during normal execution, and it appears that they occasionally exceed the 4s timeout.

I am not sure exactly why MQTT CONNECT takes such a long time, but as the name of the test suggests, perhaps it has to do with session propagation in a cluster.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,29 +20,29 @@ addons:
 
 jobs:
   include:
-    # - name: "Compile and various other checks"
-    #   env: TEST_SUITE=compile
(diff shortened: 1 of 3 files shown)


wallyqs (approved):
LGTM

levb:
@wallyqs I have 3 commits to run different (repeat) combinations of tests, please squash merge when done.

PR #4518, opened 2023-09-12
[FIXED] Increased AckWait in TestMQTTQoS2RetriesPublish, TestMQTTQoS2RetriesPubRel

levb:
TestMQTTQoS2RetriesPublish to 100ms, and in TestMQTTQoS2RetriesPubRel to 50ms.

A lesser value caused another PUBLISH to be fired while the test was still processing the final QoS2 flow. Reduced the number of retries we wait for to make the test a little quicker.


neilalexander (approved):
LGTM

PR #4519, opened 2023-09-12
[FIXED] Move test fix - TestJetStreamSuperClusterMovingStreamsAndConsumers/R3

derekcollison:
The default timeout for JetStream API calls is 10s, so in the case where we determine that we are the leader, but the stream info endpoint has not registered with the server we are connected to, the stream info call could fail and we would exhaust the whole checkFor since we would stay in one call for 10s.

Fix is to override and make multiple attempts possible for the checkFor loops.


wallyqs (approved):
LGTM

PR #4520, opened 2023-09-12
[FIXED] Fix for panic from a bug in selecting a filestore block in new fs code.

derekcollison:
When num blocks > 32 and we used new binary search in NumPending() we could return -1, nil. If sequence is inclusive this should always return valid index and mb.

The reason we could return -1 would be that we were not accounting for gaps as mb.first.seq can move ahead as first is removed. The panic could orphan held locks for filestore, consumer and possibly stream which would lock up a system, leading to memory growth and unstable behaviors.

### DIFF
--- a/server/filestore.go
+++ b/server/filestore.go
@@ -2452,7 +2452,10 @@ func (fs *fileStore) NumPending(sseq uint64, filter string, lastPerSubject bool)
 
 	// See if we need to figure out starting block per sseq.
 	if sseq > fs.state.FirstSeq {
-		seqStart, _ = fs.selectMsgBlockWithIndex(sseq)
+		// This should not, but can return -1, so make sure we check to avoid panic below.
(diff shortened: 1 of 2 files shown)


server/filestore.go
```diff
@@ -4765,6 +4769,13 @@ func (fs *fileStore) selectMsgBlockWithIndex(seq uint64) (int, *msgBlock) {
 		if seq > last {
 			low = mid + 1
 		} else if seq < first {
+			// A message block's first sequence can change here meaning we could find a gap.
+			// We want to behave like above, which if inclusive (we check at start) should
+			// awlays return an index and a valid mb.
```
wallyqs, line 4774:
```suggestion
			// always return an index and a valid mb.
```

derekcollison: Thx will fix. I do run spellcheck but I guess its not catching much these days..

wallyqs (approved):
LGTM

PR #4532, opened 2023-09-14
Fix leaking timers in stream sources

neilalexander:
Repeated calls to `scheduleSetSourceConsumerRetry` could end up creating multiple timers for the same source, which would eventually schedule even more timers, which would result in runaway CPU usage. This PR instead bounds to one timer per source per stream.


derekcollison (approved):
L˝GTM

PR #4544, opened 2023-09-15
AuthCallout request should include TLS data when client is NATS WS client

tbeets:
Make sure the client handshake flag is set when TLS handshake is made as part of WebSocket connection/upgrade (notionally HTTPS) rather than as part of the NATS protocol TLS initiation chain.  AuthCallout tests the flag when building the data for the AuthCallout service request.

Added AuthCallout unit test for NATS WS client auth that requires the TLS data.

### DIFF
--- a/server/auth_callout_test.go
+++ b/server/auth_callout_test.go
@@ -155,7 +155,14 @@ func (at *authTest) Connect(clientOptions ...nats.Option) *nats.Conn {
 func (at *authTest) WSNewClient(clientOptions ...nats.Option) (*nats.Conn, error) {
 	pi := at.srv.PortsInfo(10 * time.Millisecond)
 	require_False(at.t, pi == nil)
-	conn, err := nats.Connect(strings.Replace(pi.WebSocket[0], "127.0.0.1", "localhost", 1), clientOptions...)
+
+	// test cert is SAN to DNS localhost, not local IPs returned by server in test environments
+	wssUrl := strings.Replace(pi.WebSocket[0], "127.0.0.1", "localhost", 1)
+
+	// Seeing 127.0.1.1 in some test environments...
+	wssUrl = strings.Replace(wssUrl, "127.0.1.1", "localhost", 1)
+
+	conn, err := nats.Connect(wssUrl, clientOptions...)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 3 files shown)


server/auth_callout_test.go
```diff
@@ -1423,3 +1447,82 @@ func TestAuthCalloutOperator_AnyAccount(t *testing.T) {
...
+				# Needs to be a public account nkey, will work for both server config and operator mode.
+				issuer: "ABJHLOVMPA4CI6R5KLNGOB4GSLNIY7IOUPAJC4YFNDLQVIOBYQGUWVLA"
+				account: AUTH
+				auth_users: [ auth ]
+			}
+		}
+	`
+	handler := func(m *nats.Msg) {
+		user, si, ci, _, ctls := decodeAuthRequest(t, m.Data)
+		require_True(t, si.Name == "T")
```
neilalexander, line 1489: Please use `require_Equal` instead of `require_True` in these cases, otherwise the test logging doesn't say anything useful (like what the values were) when the condition fails.

neilalexander (changes requested):

server/auth_callout_test.go
tbeets: Oh man, you're brutal.  I just copied @derekcollison existing test verbatim (on purpose) with only change the nature of the client as NATS WS TLS instead of NATS TLS.

neilalexander: Sorry :D It's on my list to go through at some point and tidy up these things in the tests, as we've had an awful lot of lacking logging output when debugging flakes recently, but I'd rather not make the problem worse in the meantime.
```
git grep "require_True" | grep "==" | wc -l
     647

git grep "require_True" | grep "!=" | wc -l
      61
```

tbeets: I like to tease.  :-)

I am seeing that fixing the issue of this PR surfaced a WS compression test artifact that overrides the natural underlying type of the *net.Conn for compression testing purposes and thus causes server panic in monitor.go when the *tls.Conn is mysteriously changed to a *server.testWSWrappedConn.  Will fix with better type check handling in monitor.go.

neilalexander (approved):
LGTM

PR #4548, opened 2023-09-15
Add logfile_max_num feature

tbeets:

NATS Server 2.9 has `logfile_size_limit` option which allows the operator to set an optional byte limit on the NATS Server log file which when met causes a "rotation" such that the current log file is renamed (original file name appended with a time stamp to nanosecond accuracy) and a new log file is instantiated.

This PR is a new `logfile_max_num` companion option (alias `log_max_num`) which allows the operator to designate that the server should prune the **total number of log files** -- the currently active log file plus backups -- to the maximum setting. 

A max value of `0` (the implicit default) or a negative number has meaning of unlimited log files (no maximum) as this is an opt-in feature.

A max value of `1` is effectively a truncate-only logging pattern as any backup made at rotation will subsequently be purged.

A max value of `2` will maintain the active log file plus the latest backup.  And so on...

> The currently active log file is never purged.  Only backups are purged.  

When enabled, backup log deletion is evaluated inline after each successful rotation event.  To be considered for log deletion, backup log files MUST adhere to the file naming format used in log rotation as well as agree with the current `logfile` name and location.  Any other files or sub-directories in the log directory will be ignored.  E.g. if an operator makes a manual copy of the log file to `logfile.bak` that file will not be evaluated as a backup.

### Typical use case:

This feature is useful in a constrained hosting environment for NATS Server, for example an embedded, edge-compute, or IoT device scenario, in which _more featureful_ platform or operating system log management features do not exist or the complexity is not required.

### DIFF
--- a/logger/log.go
+++ b/logger/log.go
@@ -19,7 +19,6 @@ import (
 	"log"
 	"os"
 	"path/filepath"
-	"sort"
 	"strings"
 	"sync"
 	"sync/atomic"
@@ -131,14 +130,14 @@ type fileLogger struct {
 	out       int64
 	canRotate int32
 	sync.Mutex
-	l            *Logger
-	f            writerAndCloser
-	limit        int64
-	olimit       int64
-	pid          string
-	time         bool
-	closed       bool
-	archiveLimit int
+	l           *Logger
+	f           writerAndCloser
+	limit       int64
+	olimit      int64
+	pid         string
+	time        bool
+	closed      bool
+	maxNumFiles int
 }
 
 func newFileLogger(filename, pidPrefix string, time bool) (*fileLogger, error) {
@@ -173,9 +172,9 @@ func (l *fileLogger) setLimit(limit int64) {
 	}
 }
 
-func (l *fileLogger) setArchiveLimit(limit int) {
+func (l *fileLogger) setMaxNumFiles(max int) {
 	l.Lock()
-	l.archiveLimit = limit
+	l.maxNumFiles = max
 	l.Unlock()
 }
 
@@ -200,9 +199,7 @@ func (l *fileLogger) logDirect(label, format string, v ...interface{}) int {
 	return len(entry)
 }
 
-func (l *fileLogger) archivePurge(fname string) {
-	// Evaluate number of saved backups for purge
-	// l readlock held
+func (l *fileLogger) logPurge(fname string) {
 	var backups []string
 	lDir := filepath.Dir(fname)
 	lBase := filepath.Base(fname)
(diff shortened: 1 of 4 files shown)


logger/log.go
```diff
@@ -190,6 +200,46 @@ func (l *fileLogger) logDirect(label, format string, v ...interface{}) int {
...
+		}
+	}
+	archives := len(backups)
+	if archives > l.archiveLimit {
+		// Oldest to latest
+		sort.Slice(backups, func(i, j int) bool {
+			return backups[i] < backups[j]
+		})
+		for i := 0; i < archives-l.archiveLimit; i++ {
+			if err := os.Remove(fmt.Sprintf("%s%s%s", lDir, string(os.PathSeparator), backups[i])); err != nil {
```
derekcollison, line 233: Use filepath join here.

logger/log.go
```diff
@@ -190,6 +200,46 @@ func (l *fileLogger) logDirect(label, format string, v ...interface{}) int {
...
+	if err != nil {
+		l.logDirect(l.l.errorLabel, "Unable to read directory %q for log purge (%v), will attempt next rotation", lDir, err)
+		return
+	}
+	for _, entry := range entries {
+		if entry.IsDir() || entry.Name() == lBase || !strings.HasPrefix(entry.Name(), lBase) {
+			continue
+		}
+		stamp, found := strings.CutPrefix(entry.Name(), fmt.Sprintf("%s%s", lBase, "."))
+		if found {
```
derekcollison, line 219: Can make on line combined with above.

`if stamp, found := strings.CutPrefix(entry.Name(), fmt.Sprintf("%s%s", lBase, "."))l found {`

logger/log.go
```diff
@@ -190,6 +200,46 @@ func (l *fileLogger) logDirect(label, format string, v ...interface{}) int {
 	return len(entry)
 }
 
+func (l *fileLogger) archivePurge(fname string) {
```
derekcollison, line 203: Is fname a dir?

logger/log.go
```diff
@@ -190,6 +200,46 @@ func (l *fileLogger) logDirect(label, format string, v ...interface{}) int {
...
+	// Evaluate number of saved backups for purge
+	// l readlock held
+	var backups []string
+	lDir := filepath.Dir(fname)
+	lBase := filepath.Base(fname)
+	entries, err := os.ReadDir(lDir)
```
derekcollison, line 209: I think this lexicographically sorts already IIRC.

logger/log.go
```diff
@@ -190,6 +200,46 @@ func (l *fileLogger) logDirect(label, format string, v ...interface{}) int {
...
+	if err != nil {
+		l.logDirect(l.l.errorLabel, "Unable to read directory %q for log purge (%v), will attempt next rotation", lDir, err)
+		return
+	}
+	for _, entry := range entries {
```
derekcollison, line 214: Easier way to collect these?

logger/log.go
```diff
@@ -190,6 +200,46 @@ func (l *fileLogger) logDirect(label, format string, v ...interface{}) int {
...
+			_, err := time.Parse("2006:01:02:15:04:05.999999999", strings.Replace(stamp, ".", ":", 5))
+			if err == nil {
+				backups = append(backups, entry.Name())
+			}
+		}
+	}
+	archives := len(backups)
+	if archives > l.archiveLimit {
+		// Oldest to latest
+		sort.Slice(backups, func(i, j int) bool {
```
derekcollison, line 229: I think it might already be sorted from above.

logger/log.go
```diff
@@ -128,13 +131,14 @@ type fileLogger struct {
...
-	time   bool
-	closed bool
+	l            *Logger
+	f            writerAndCloser
+	limit        int64
+	olimit       int64
+	pid          string
+	time         bool
+	closed       bool
+	archiveLimit int
```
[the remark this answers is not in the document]
tbeets: To me name `maxNumFiles` on the logger implies the current log file _plus_ backups, where the option value is centric to how many older log _segments_ (aka archives) to keep not including the actual current log file that is open.  `maxNumArchives`?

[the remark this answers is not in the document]
derekcollison: Seems more natural to me to say how many total, vs understanding +1 all the time.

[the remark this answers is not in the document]
tbeets: Agree. I will change throughout. Will also allow `log_max_files: 1` to essentially mean truncation and don't keep any rotated logs.

logger/log.go
tbeets: Fully qualified name of the active log file.  So both the base logfile name and the directory where all the backups are being stashed derived from that.

logger/log.go
tbeets: You're right.  os.ReadDir() underneath does the identical lexical sort.

logger/log.go
tbeets: Confirmed.

logger/log.go
tbeets: Good.

logger/log.go
tbeets: Good.

logger/log.go
tbeets: Don't think so.  Since we allow the operator to put the log file in an arbitrary location, need to test each entry in the directory in several ways to make sure it is a legitimate log backup that we created.

derekcollison (approved):
LGTM

PR #4553, opened 2023-09-18
Various fixes and improvements to tombstone and buffer gaps.

derekcollison:
We fixed a few bugs in tombstone handling, and formalized support for holes in the underlying buffers. Due to customer data from the field we also now use holes during compaction.


wallyqs (approved):
LGTM

PR #4556, opened 2023-09-18
Fix to stop forwarding proposals in consumers after scaling down a stream

wallyqs:
Sometimes when scaling down a stream, a raft node could continue forwarding proposals after already being closed, in the debug logs this can be confirmed by many entries logging `Direct proposal ignored, not leader (state: CLOSED)`.

### DIFF
--- a/server/consumer.go
+++ b/server/consumer.go
@@ -1122,6 +1122,16 @@ func (o *consumer) isLeader() bool {
 	return true
 }
 
+func (o *consumer) clearLoopAndForward() {
+	o.mu.Lock()
+	defer o.mu.Unlock()
+	if o.qch != nil {
+		close(o.qch)
+		// Note can not close pch here.
+		o.qch, o.pch = nil, nil
+	}
+}
+
 func (o *consumer) setLeader(isLeader bool) {
 	o.mu.RLock()
 	mset := o.mset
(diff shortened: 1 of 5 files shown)


derekcollison (approved):
LGTM - How long does test take?

server/consumer.go
```diff
@@ -1122,6 +1122,16 @@ func (o *consumer) isLeader() bool {
 	return true
 }
 
+func (o *consumer) clearLoopAndForward() {
+	o.mu.Lock()
+	defer o.mu.Unlock()
+	if o.qch != nil {
+		close(o.qch)
+		// Note can not close pch here.
```
neilalexander, line 1130: Is this because of a panic when writing to the channel elsewhere?

wallyqs:
Currently it takes about 40s, I think it can be made shorter

server/jetstream_cluster.go
```diff
@@ -4264,6 +4264,8 @@ func (js *jetStream) processClusterCreateConsumer(ca *consumerAssignment, state
 		} else {
 			// Check for scale down to 1..
 			if rg.node != nil && len(rg.Peers) == 1 {
+				// Need to pop loopAndForward by closing qch and nil out both qch and pch.
```
neilalexander, line 4267: Would be nice if we can expand this comment just to explain what happens if we don't, i.e. that it results in errant proposals.

neilalexander (approved):
Otherwise LGTM

server/jetstream_cluster.go
wallyqs: Updated this part

neilalexander (approved):
LGTM

PR #4557, opened 2023-09-19
Remove `subject_transform_dest`

jnmoyne:
Removes the single subject transform destination field any subject transformation in StreamSources must now be done using the SubjectTransforms array instead.

### DIFF
--- a/server/stream.go
+++ b/server/stream.go
@@ -180,25 +180,23 @@ type PeerInfo struct {
 
 // StreamSourceInfo shows information about an upstream stream source.
 type StreamSourceInfo struct {
-	Name                 string                   `json:"name"`
-	External             *ExternalStream          `json:"external,omitempty"`
-	Lag                  uint64                   `json:"lag"`
-	Active               time.Duration            `json:"active"`
-	Error                *ApiError                `json:"error,omitempty"`
-	FilterSubject        string                   `json:"filter_subject,omitempty"`
-	SubjectTransformDest string                   `json:"-"`
-	SubjectTransforms    []SubjectTransformConfig `json:"subject_transforms,omitempty"`
+	Name              string                   `json:"name"`
+	External          *ExternalStream          `json:"external,omitempty"`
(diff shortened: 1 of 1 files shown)


server/stream.go
```diff
@@ -180,22 +180,24 @@ type PeerInfo struct {
...
-	OptStartSeq          uint64                   `json:"opt_start_seq,omitempty"`
-	OptStartTime         *time.Time               `json:"opt_start_time,omitempty"`
-	FilterSubject        string                   `json:"filter_subject,omitempty"`
+	Name          string     `json:"name"`
+	OptStartSeq   uint64     `json:"opt_start_seq,omitempty"`
+	OptStartTime  *time.Time `json:"opt_start_time,omitempty"`
+	FilterSubject string     `json:"filter_subject,omitempty"`
+	// Do not expose SubjectTransformDest in the client libraries
+	// it is here because it was enabled for a while in the 2.10 beta but is now removed
+	// and there is now a check to error out if it is provided in the request
```
wallyqs, line 200: If it is an unused field and not in v2.9.22, I think now the best chance to remove it...

jnmoyne: That's exactly what this PR is about :)

wallyqs: ok we can as is for now but remove in v2.11

jnmoyne: If it makes it in 2.10 then has to be maintained afterwards if we want to keep backwards compatibility, hence why right now is the last chance to remove it while it's never been officially part of a 'main' release (even though it was on dev for a very long time)

wallyqs (approved):
LGTM

wallyqs (approved):
LGTM

derekcollison (approved):
LGTM - if tests pass.

neilalexander (approved):
LGTM now that the `TestJetStreamFilteredSubjectUsesNewConsumerCreateSubject` test exists and passes.

PR #4558, opened 2023-09-19
Ignore `subject_transform_dest` in stream sourcing

neilalexander:
This is a safer (less lines of code touched) alternative to #4557 for now, which simply ignores the `subject_transform_dest` field in the API and the stream assignments. We'll still look to merge the other PR to clean up but will do so post-release when we have more time to test it.


Jarema (approved):
LGTM!

This is much safer approach for now.

bruth (approved):

PR #4564, opened 2023-09-19
Use new consumer create subject when single subject filter specified in `SubjectFilters`

neilalexander:
This fixes an issue where specifying a single subject filter, i.e. in `SubjectFilters` or `SubjectTransforms`, instead of using `SubjectFilter` would result in the old consumer create subject being incorrectly used.

### DIFF
--- a/server/jetstream_test.go
+++ b/server/jetstream_test.go
@@ -21853,3 +21853,78 @@ func TestJetStreamSyncInterval(t *testing.T) {
 		})
 	}
 }
+
+func TestJetStreamFilteredSubjectUsesNewConsumerCreateSubject(t *testing.T) {
+	s := RunBasicJetStreamServer(t)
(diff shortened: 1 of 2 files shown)


jnmoyne:
LGTM

derekcollison (approved):
LGTM

tbeets:
LGTM - let's get those test flappers fixed and light the candle.

derekcollison:
This test (at least) is truly broke, so we need to fix it.

```
TestJetStreamSourceBasics
```

derekcollison:
The change introduced a breaking behavior.

neilalexander:
Have added a comment explaining what is happening and rebased to top of `main`.

Thanks @jnmoyne for the fix!

derekcollison (approved):
LGTM

PR #4565, opened 2023-09-20
[FIXED] LeafNode: TLS Handshake when remote does not have a tls{} block

kozlovic:
If a leafnode remote configuration does not have a tls{} block but connect to a hub that requires TLS, the handshake between the two servers will fail. A simple workaround is to add in the remote configuration an empty tls{} block.

This issue was introduced in v2.10.0 due to some refactoring in order to support compression.


derekcollison (approved):
LGTM - Thanks.

PR #4571, opened 2023-09-22
[FIXED] We will panic since we unlock the mb in removeMsgBlock

derekcollison:


wallyqs (approved):
LGTM

kozlovic:
Looks like there is one more place where `fs.removeMsgBlock()` is invoked without the mb's lock, is that ok? I am thinking of  `rebuildFirst()`..

derekcollison:
Thanks will take a look.

PR #4575, opened 2023-09-22
[FIXED] Protect against going upside down on mb.msgs.

derekcollison:


wallyqs (approved):
LGTM

PR #4579, opened 2023-09-24
In lameduck mode shutdown jetstream at start, do not leave running

derekcollison:


wallyqs (approved):
LGTM

PR #4582, opened 2023-09-25
Report observer mode in replica state

neilalexander:
Currently we don't surface anything through the JS API that reports whether a node is in observer state or not (i.e. because of LDM or because of `js_cluster_migrate`). In part this is because nodes don't currently track whether other nodes in the group are observers or not.

This branch adds a new proposal type to the Raft layer to signal when a node enters or leaves observer mode. We then use this to bubble up the observer state through the stream and consumer state, which the CLI can use to render the state. This should make it possible to debug whether a clustered asset is leaderless and/or not voting because all of the nodes are observers.

**This PR has a one or two things to clean up before merge** — namely we need to merge changes to nats.go first to remove the `replace` from `go.mod`, the JS API schema and CLI also need changes for this too.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,5 +1,7 @@
 module github.com/nats-io/nats-server/v2
 
+replace github.com/nats-io/nats.go => github.com/nats-io/nats.go v1.30.1-0.20230925101945-66fd3c30ab7b
+
 go 1.20
 
 require (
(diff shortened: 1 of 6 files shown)


neilalexander:
Closing as we can now see this via `/raftz`.

PR #4585, opened 2023-09-25
[IMPROVED] Move some contended locks to atomic.Bools

derekcollison:


wallyqs (approved):
LGTM

PR #4586, opened 2023-09-25
[IMPROVED] The func subjectIsSubsetMatch() is heavy so do without the account lock.

derekcollison:

### DIFF
--- a/server/accounts.go
+++ b/server/accounts.go
@@ -1716,6 +1716,11 @@ func (a *Account) checkForReverseEntries(reply string, checkInterest, recursed b
 
 	var _rs [64]string
 	rs := _rs[:0]
+
+	if n := len(a.imports.rrMap); n > cap(rs) {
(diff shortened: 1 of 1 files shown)


server/accounts.go
```diff
@@ -1717,14 +1717,15 @@ func (a *Account) checkForReverseEntries(reply string, checkInterest, recursed b
 	var _rs [64]string
```
neilalexander, line 1717: It will likely now make sense to just do `rs := make([]string, 0, len(a.imports.rrMap))` to prevent reallocations.

wallyqs (approved):
LGTM

neilalexander (approved):
Otherwise LGTM!

server/accounts.go
derekcollison: Fixed, please take a quick look.

PR #4587, opened 2023-09-25
Add `prof_block_rate` option for enabling/configuring the block profile

neilalexander:
The new `prof_block_rate` configuration option allows the block profiler to be enabled on demand after it was previously disabled in #4402. The option is also reloadable so that it can be changed after startup.

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -2098,6 +2098,10 @@ func (s *Server) Start() {
 	// Pprof http endpoint for the profiler.
 	if opts.ProfPort != 0 {
 		s.StartProfiler()
+	} else {
+		// It's still possible to access this profile via a SYS endpoint, so set
+		// this anyway. (Otherwise StartProfiler would have called it.)
+		s.setBlockProfileRate(opts.ProfBlockRate)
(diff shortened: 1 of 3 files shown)


server/server.go
```diff
@@ -2098,6 +2098,10 @@ func (s *Server) Start() {
 	// Pprof http endpoint for the profiler.
 	if opts.ProfPort != 0 {
 		s.StartProfiler()
+	} else {
+		// It's still possible to access this profile via a SYS endpoint, so set
+		// this anyway. (Otherwise StartProfiler would have called it.)
+		s.setBlockProfileRate(opts.ProfBlockRate)
```
derekcollison, line 2104: I thought this was bad to set by default no?

derekcollison: Is it zero by default? Meaning does not set anything penal?

derekcollison (approved):
LGTM

server/server.go
neilalexander: Yep, zero by default, which disables the block profiler.

PR #4589, opened 2023-09-25
[FIXED] JetStream: stream assignment data race

kozlovic:
Two go routines could possibly execute the stream assignment at the same time. A WaitGroup was used to prevent that, but an issue caused the data race and possible concurrent execution.


wallyqs (approved):
LGTM

derekcollison (approved):

PR #4591, opened 2023-09-27
[IMPROVED] ServiceImport Reply Optimizations

derekcollison:
We added  some small performance tweak to the func checkForReverseEntries. In addition, we move the shutdown bool for the server to an atomic so we could efficiently check it when doing unsubs. If the server is going away there is really no need since the other side will do its own thing when the connection goes away. And finally we do not have to range over the account rrMap if the subscription going away is a reserved reply.


wallyqs (approved):
LGTM

neilalexander (approved):
LGTM

PR #4592, opened 2023-09-27
[FIXED] Race condition in some leader failover scenarios leading to messages being potentially sourced more than once.

jnmoyne:
- [X] Tests added

Fixes a race condition in some leader failover scenarios leading to messages being potentially sourced more than once.

In some failure scenarios where the current leader of a stream sourcing from other stream(s) gets shutdown while publications are happening on the stream(s) being sourced leads to `setLeader(true)` being called on the new leader for the sourcing stream before all the messages having been sourced by the previous leader are completely processed such that when the new leader does it's reverse scan from the last message in it's view of the stream in order to know what sequence number to start the consumer for the stream being sourced from, such that the last message(s) sourced by the previous leader get sourced again, leading to some messages being sourced more than once.

The existing `TestNoRaceJetStreamSuperClusterSources` test would sidestep the issue by relying on the deduplication window in the sourcing stream. Without deduplication the test is a flapper.

This avoid the race condition by adding a small delay before scanning for the last message(s) having been sourced and starting the sources' consumer(s). Now the test (without using the deduplication window) never fails because more messages than expected have been received in the sourcing stream.

(Also adds a guard to give up if `setupSourceConsumers()` is called and we are no longer the leader for the stream (that check was already present in `setupMirrorConsumer()` so assuming it was forgotten for `setupSourceConsumers()`)

### DIFF
--- a/server/norace_test.go
+++ b/server/norace_test.go
@@ -1874,7 +1874,6 @@ func TestNoRaceJetStreamSuperClusterSources(t *testing.T) {
 			msg := fmt.Sprintf("R-MSG-%d", i+1)
 			for _, sname := range []string{"foo", "bar", "baz"} {
 				m := nats.NewMsg(sname)
-				m.Header.Set(nats.MsgIdHdr, sname+"-"+msg)
 				m.Data = []byte(msg)
 				if _, err := js.PublishMsg(m); err != nil {
 					t.Errorf("Unexpected publish error: %v", err)
@@ -1891,7 +1890,7 @@ func TestNoRaceJetStreamSuperClusterSources(t *testing.T) {
 	sc.clusterForName("C3").waitOnStreamLeader("$G", "MS2")
 	<-doneCh
 
-	checkFor(t, 15*time.Second, 100*time.Millisecond, func() error {
+	checkFor(t, 15*time.Second, time.Second, func() error {
 		si, err := js2.StreamInfo("MS2")
 		if err != nil {
 			return err
--- a/server/stream.go
+++ b/server/stream.go
@@ -246,8 +246,9 @@ type stream struct {
 	mirror *sourceInfo
 
 	// Sources
-	sources       map[string]*sourceInfo
-	sourceRetries map[string]*time.Timer
+	sources              map[string]*sourceInfo
+	sourceRetries        map[string]*time.Timer
+	sourcesConsumerSetup *time.Timer
 
 	// Indicates we have direct consumers.
 	directs int
@@ -821,6 +822,11 @@ func (mset *stream) setLeader(isLeader bool) error {
 			return err
 		}
 	} else {
+		// cancel timer to create the source consumers if not fired yet
+		if mset.sourcesConsumerSetup != nil {
+			mset.sourcesConsumerSetup.Stop()
+			mset.sourcesConsumerSetup = nil
+		}
 		// Stop responding to sync requests.
 		mset.stopClusterSubs()
 		// Unsubscribe from direct stream.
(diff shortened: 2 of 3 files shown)


server/norace_test.go
```diff
@@ -1891,7 +1890,7 @@ func TestNoRaceJetStreamSuperClusterSources(t *testing.T) {
 	sc.clusterForName("C3").waitOnStreamLeader("$G", "MS2")
 	<-doneCh
 
-	checkFor(t, 15*time.Second, 100*time.Millisecond, func() error {
+	checkFor(t, 15*time.Second, 1000*time.Millisecond, func() error {
```
derekcollison, line 1893: time.Second

server/stream.go
```diff
@@ -249,6 +249,8 @@ type stream struct {
 	sources       map[string]*sourceInfo
 	sourceRetries map[string]*time.Timer
 
+	sourcesMirrorConsumerSetup *time.Timer
```
derekcollison, line 252: Looks like java ;) Is this for sources as well as mirrors? 

`remoteConsumerSetup`

server/stream.go
```diff
@@ -820,7 +822,30 @@ func (mset *stream) setLeader(isLeader bool) error {
...
+				}
+			} else if len(mset.cfg.Sources) > 0 {
+				if err := mset.setupSourceConsumers(); err != nil {
+					mset.srv.Logger().Errorf("source consumer setup error: %v", err)
+				}
+			}
+
+			mset.mu.Unlock()
+		})
+		mset.sourcesMirrorConsumerSetup = setupConsumersTimer
```
derekcollison, line 843: Why not set direct above?

server/stream.go
jnmoyne: Outaded after new rewrite

server/norace_test.go
jnmoyne: Done

server/stream.go
```diff
@@ -2389,7 +2395,7 @@ func (mset *stream) scheduleSetupMirrorConsumerRetryAsap() {
 	}
 	// To make *sure* that the next request will not fail, add a bit of buffer
 	// and some randomness.
-	next += time.Duration(rand.Intn(50)) + 10*time.Millisecond
+	next += time.Duration(rand.Intn(10000000)) + 10*time.Millisecond
```
derekcollison, line 2398: Let's make it clear here.

`rand.Intn(10*time.Millisecond)` with appropriate casts here.

server/stream.go
```diff
@@ -3439,13 +3457,36 @@ func (mset *stream) subscribeToStream() error {
 	}
 	// Check if we need to setup mirroring.
 	if mset.cfg.Mirror != nil {
-		if err := mset.setupMirrorConsumer(); err != nil {
-			return err
+		// setup the initial mirror sourceInfo
+		mset.mirror = &sourceInfo{name: mset.cfg.Mirror.Name}
+		mirror := mset.mirror
```
derekcollison, line 3462: Just use mset.mirror below, will be clearer.

derekcollison (changes requested):

server/stream.go

server/stream.go

server/stream.go
```diff
@@ -2389,7 +2395,7 @@ func (mset *stream) scheduleSetupMirrorConsumerRetryAsap() {
 	}
 	// To make *sure* that the next request will not fail, add a bit of buffer
 	// and some randomness.
-	next += time.Duration(rand.Intn(50)) + 10*time.Millisecond
+	next += time.Duration(rand.Intn(10*int(time.Millisecond))) + 10*time.Millisecond
```
derekcollison, line 2398: Put 10 inside, so int(10*time.Mliisecond)

jnmoyne: Done. I also updated the scheduleSourceConsumerRetryASAP to use the same backoff computation (it was still using the 10ms with 0 to 50uS jitter)

derekcollison:
If this runs green are we good to pull in?

jnmoyne:
Yes

derekcollison (approved):
LGTM

PR #4593, opened 2023-09-27
Fix `TestNoRaceJetStreamStreamInfoSubjectDetailsLimits` for changes in nats.go

neilalexander:
There are changes in recent versions of nats.go that seemingly increase the size of the stream info and cause this test to fail consistently with `norace_test.go:4259: require no error, but got: nats: maximum payload exceeded`. Fix the test to use larger limits and payloads so we are not sensitive to this when nats.go is upgraded.


Jarema (approved):
LGTM

PR #4594, opened 2023-09-27
Don't take sublist write lock in `match` if sublist cache disabled

neilalexander:
We may be creating unnecessary lock contention on the sublist when the cache is disabled by taking the write lock anyway.


derekcollison (approved):
LGTM

wallyqs (approved):
LGTM

PR #4600, opened 2023-09-28
[IMPROVED] Concurrent stream creation of the same stream could return not found

derekcollison:
Here we know that if we can't find the stream but have the stream assignment, this is a distinct possibility. So we wait, since not processed inline, to see if it appears.

Fixes TestJetStreamClusterParallelStreamCreation as well that could flap.


wallyqs (approved):
LGTM

PR #4601, opened 2023-09-28
[IMPROVED] Additional markers for dirty state

derekcollison:
Under certain circumstances we could delay recovery if the state file pointed to an absent msg block.
Found additional places to mark dirty and optionally kick the flusher.


wallyqs (approved):
LGTM

PR #4603, opened 2023-09-28
systemd: use SIGUSR2 for shutdown, for LDM

philpennock:

 - tell systemd to use SIGUSR2 for shutdown, to trigger Lame Duck Mode in nats-server

### DIFF
--- a/util/nats-server-hardened.service
+++ b/util/nats-server-hardened.service
@@ -9,6 +9,9 @@ ExecReload=/bin/kill -s HUP $MAINPID
 ExecStop=/bin/kill -s SIGINT $MAINPID
 User=nats
 Group=nats
+# The nats-server uses SIGUSR2 to trigger using Lame Duck Mode (LDM) shutdown
+KillSignal=SIGUSR2
+# You might want to adjust TimeoutStopSec too.
 
 # Hardening
 CapabilityBoundingSet=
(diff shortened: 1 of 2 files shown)


wallyqs (approved):

derekcollison (approved):
LGTM - Question should we leave commented out though? This would change behavior yes?

philpennock:
Someone would have to actually use this util/ script from the repo, as-is, automatically deploying it.  Do such setups exist?  AFAIK this is just a source for people to copy from during setup, if they so choose.

derekcollison:
Oh yes they do. Without any edits.

util/nats-server-hardened.service
```diff
@@ -9,6 +9,9 @@ ExecReload=/bin/kill -s HUP $MAINPID
 ExecStop=/bin/kill -s SIGINT $MAINPID
 User=nats
 Group=nats
+# The nats-server uses SIGUSR2 to trigger using Lame Duck Mode (LDM) shutdown
+KillSignal=SIGUSR2
+# You might want to adjust TimeoutStopSec too.
```
JulienVdG, line 14: Hello, sorry for the late message, I only saw the change when browsing the release notes.

Shouldn't the ExecStop command also be updated to send SIGUSR2 ?

PR #4604, opened 2023-09-28
[FIXED] Flapping TestMQTTLockedSession

levb:
A test-only fix.

I can not reproduce the flapping behavior, but did see a race during debugging suggesting that the CONNACK is delivered to the test before `mqttProcessConnect` finishes and releases the record.

### DIFF
--- a/server/mqtt_test.go
+++ b/server/mqtt_test.go
@@ -4654,7 +4654,8 @@ func TestMQTTLockedSession(t *testing.T) {
 	s := testMQTTRunServer(t, o)
 	defer testMQTTShutdownServer(s)
 
-	ci := &mqttConnInfo{clientID: "sub", cleanSess: false}
+	subClientID := "sub"
+	ci := &mqttConnInfo{clientID: subClientID, cleanSess: false}
 	c, r := testMQTTConnect(t, ci, o.MQTT.Host, o.MQTT.Port)
 	defer c.Close()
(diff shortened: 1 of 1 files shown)


server/mqtt_test.go
```diff
@@ -4667,6 +4667,19 @@ func TestMQTTLockedSession(t *testing.T) {
 		t.Fatalf("account session manager not found")
 	}
 
+	// It is possible, however unlikely, to have received CONNACK while
+	// mqttProcessConnect is still running, and the session remains locked. Wait
+	// for it to finish.
+	stillLocked := true
+	for stillLocked {
```
derekcollison, line 4674:
```
for stillLocked := true; stillLocked; {
}
```

server/mqtt_test.go
```diff
@@ -4667,6 +4667,19 @@ func TestMQTTLockedSession(t *testing.T) {
 		t.Fatalf("account session manager not found")
 	}
 
+	// It is possible, however unlikely, to have received CONNACK while
+	// mqttProcessConnect is still running, and the session remains locked. Wait
+	// for it to finish.
+	for stillLocked := true; stillLocked; {
```
derekcollison, line 4673: You can also use a checkFor() block..

derekcollison (changes requested):
This is what checkFor blocks are for..

server/mqtt_test.go
levb: even better, TY, didn't know!

derekcollison (approved):
LGTM

PR #4605, opened 2023-09-28
[FIXED] Do not bypass authorization blocks when turning on $SYS account access

derekcollison:
Only setup auto no-auth for $G account iff no authorization block was defined.

Resolves #4535


neilalexander (approved):
LGTM

PR #4613, opened 2023-09-30
[IMPROVED] Reduce contention for high connections in a JetStream enabled account with high API usage.

derekcollison:
Several strategies are used which are listed below.

1. Checking a RaftNode to see if it is the leader now uses atomics.
2. Checking if we are the JetStream meta leader from the server now uses an atomic.
3. Accessing the JetStream context no longer requires a server lock, uses atomic.Pointer.
4. Filestore syncBlocks would hold msgBlock locks during sync, now does not.

### DIFF
--- a/server/consumer.go
+++ b/server/consumer.go
@@ -1556,7 +1556,7 @@ func (o *consumer) deleteNotActive() {
 		}
 	}
 
-	s, js := o.mset.srv, o.mset.srv.js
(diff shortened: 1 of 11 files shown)


bruth (approved):
LGTM! Good amount of subtle clean-up.

wallyqs (approved):
LGTM

PR #4618, opened 2023-10-03
[FIX] remove extra system headers on direct gets

### LINKED ISSUE #4573: [JS] direct get apis include old `Nats-` headers creating duplicates

2.10.1

any

To reproduce the error:

Save this program to `direct_issue.ts`
```typescript
import { parse } from "https://deno.land/std@0.202.0/flags/mod.ts";
import { connect, StorageType, nuid } from "https://raw.githubusercontent.com/nats-io/nats.deno/main/src/mod.ts";
import { JetStreamManagerImpl } from 'https://raw.githubusercontent.com/nats-io/nats.deno/main/jetstream/jsm.ts'
import { MsgHdrsImpl } from '../nats-base-client/headers.ts'

const opts = parse(Deno.args);

if (opts.h || opts.help) {
  console.log("deno run -A direct_issue.ts --server hostport --help");
}
const servers = opts.server || opts.s || "localhost:4222";

const nc = await connect({ servers });

const jsm = await nc.jetstreamManager() as JetStreamManagerImpl;

const name = nuid.next();


await jsm.streams.add({
  name: name,
  subjects: ["A.>"],
  storage: StorageType.Memory,
  republish: {
    src: ">",
    dest: `$KV.${name}.>`,
  },
});

const js = nc.jetstream();
const B = await js.views.kv(name);

nc.publish("A.orange", "hey");
await js.publish("A.tomato", "hello");
await B.put("A.potato", "yo");

console.log("direct gets");
for (let i = 1; i <= 3; i++) {
  const m = await jsm.direct.getMessage(`KV_${name}`, { seq: i });
  console.log((m.header as MsgHdrsImpl).toString());
  console.log("-------------");
}
console.log("stream gets");
for (let i = 1; i <= 3; i++) {
  const m = await jsm.streams.getMessage(`KV_${name}`, { seq: i });
  console.log((m.header as MsgHdrsImpl).toString());
  console.log("-------------");
}

await jsm.streams.delete(name);
await B.destroy()
await nc.close()
```

Start a nats server with JetStream

```bash
> deno run -A direct_issue.ts

direct gets
NATS/1.0
Nats-Stream: Z0EZD0S7VBF6EL9SL4SN5I
Nats-Stream: KV_Z0EZD0S7VBF6EL9SL4SN5I
Nats-Subject: A.orange
Nats-Subject: $KV.Z0EZD0S7VBF6EL9SL4SN5I.A.orange
Nats-Sequence: 1
Nats-Sequence: 1
Nats-Last-Sequence: 0
Nats-Time-Stamp: 2023-09-22T11:35:36.861641Z


-------------
NATS/1.0
Nats-Stream: Z0EZD0S7VBF6EL9SL4SN5I
Nats-Stream: KV_Z0EZD0S7VBF6EL9SL4SN5I
Nats-Subject: A.tomato
Nats-Subject: $KV.Z0EZD0S7VBF6EL9SL4SN5I.A.tomato
Nats-Sequence: 2
Nats-Sequence: 2
Nats-Last-Sequence: 0
Nats-Time-Stamp: 2023-09-22T11:35:36.86172Z


-------------
NATS/1.0
Nats-Stream: KV_Z0EZD0S7VBF6EL9SL4SN5I
Nats-Subject: $KV.Z0EZD0S7VBF6EL9SL4SN5I.A.potato
Nats-Sequence: 3
Nats-Time-Stamp: 2023-09-22T11:35:36.862229Z


-------------
stream gets
NATS/1.0
Nats-Stream: Z0EZD0S7VBF6EL9SL4SN5I
Nats-Subject: A.orange
Nats-Sequence: 1
Nats-Last-Sequence: 0


-------------
NATS/1.0
Nats-Stream: Z0EZD0S7VBF6EL9SL4SN5I
Nats-Subject: A.tomato
Nats-Sequence: 2
Nats-Last-Sequence: 0
-------------

```

The headers from the added by the stream republish onboarding into KV shouldn't be reported. The side effect of that is Get operations trying to figure out the name of the key will fail, as there are multiple `Nats-Subject` entries (as well as sequence etc.).

```
Nats-Subject: A.tomato
Nats-Subject: $KV.1RHFTJKE86C02I6P51LXR3.A.tomato
```
that

`Nats-` headers shouldn't have duplicates.

aricart:
[FIX] direct get APIs can contain duplicate `Nats-Stream`, `Nats-Sequence`, `Nats-Time-Stamp` and `Nats-Subject` headers because it simply appends JSON bytes to existing headers. If the message was onboarded on a republish, this will contain the above system headers. Since headers are not ordered and can contain multiple values for the same header, this can break KV clients and create ambiguity on the stream that yielded the message.

 - [X] Tests added

Resolves #4573

 - remove extra `Nats-Stream`, `Nats-Sequence`, `Nats-Time-Stamp` and `Nats-Subject`  that may be present on the stream message when using direct get apis

### DIFF
--- a/server/stream.go
+++ b/server/stream.go
@@ -341,6 +341,7 @@ const (
 
 // Headers for republished messages and direct gets.
 const (
+	JSHeaderPrefix = "Nats-"
 	JSStream       = "Nats-Stream"
 	JSSequence     = "Nats-Sequence"
 	JSTimeStamp    = "Nats-Time-Stamp"
@@ -4051,6 +4052,8 @@ func (mset *stream) getDirectRequest(req *JSApiMsgGetRequest, reply string) {
 		hdr = []byte(fmt.Sprintf(ht, name, sm.subj, sm.seq, ts.Format(time.RFC3339Nano)))
 	} else {
 		hdr = copyBytes(hdr)
+		// streams that onboarded via republish could contain the original stream headers
+		hdr = removeStreamIdentityHeaders(hdr)
 		hdr = genHeader(hdr, JSStream, name)
 		hdr = genHeader(hdr, JSSubject, sm.subj)
 		hdr = genHeader(hdr, JSSequence, strconv.FormatUint(sm.seq, 10))
@@ -4397,6 +4400,11 @@ func (mset *stream) processJetStreamMsg(subject, reply string, hdr, msg []byte, 
 		if sm, _ := store.LoadLastMsg(subject, &smv); sm != nil {
 			tlseq = sm.seq
 		}
+	}
+
+	// if we are NOT doing subject remapping, this is a message to on-board it shouldn't store Nats-* headers
+	if tsubj == _EMPTY_ {
+		hdr = removeStreamIdentityHeaders(hdr)
 	}
 
 	// Store actual msg.
--- a/server/client.go
+++ b/server/client.go
@@ -3968,6 +3968,19 @@ func (c *client) setupResponseServiceImport(acc *Account, si *serviceImport, tra
 	return rsi
 }
 
+func removeStreamIdentityHeaders(hdr []byte) []byte {
+	if hdr == nil {
+		return hdr
+	}
+	if idx := bytes.Index(hdr, []byte(JSHeaderPrefix)); idx == -1 {
+		return hdr
+	}
+	for _, h := range []string{JSStream, JSSequence, JSTimeStamp, JSSubject} {
+		hdr = removeHeaderIfPresent(hdr, h)
+	}
+	return hdr
+}
+
 // Will remove a header if present.
 func removeHeaderIfPresent(hdr []byte, key string) []byte {
 	start := bytes.Index(hdr, []byte(key))
--- a/server/jetstream_test.go
+++ b/server/jetstream_test.go
@@ -21931,3 +21931,114 @@ func TestJetStreamFilteredSubjectUsesNewConsumerCreateSubject(t *testing.T) {
 		},
 	})
 }
+
+func TestJetStreamNoDuplicateHeadersOnDirectGet(t *testing.T) {
+	s := RunBasicJetStreamServer(t)
+	defer s.Shutdown()
+
+	nc, _ := jsClientConnect(t, s)
+	defer nc.Close()
+
+	name := nuid.Next()
+
+	req := StreamConfig{
+		Name:     name,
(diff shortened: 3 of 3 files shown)


neilalexander (approved):
LGTM

derekcollison (commented):
What is the performance impact on direct get performance due to these?

Should we make one function that removes them all at once?

aricart:
> What is the performance impact on direct get performance due to these?

We would have to scan at least once, but a better question is perhaps we should be doing this on the on-board? The issue is republishing is causing a recapture by the stream. If the stream is guaranteed not to have the header, then no cleanup would be necessary.

aricart:
@derekcollison discussed with @neilalexander an alternative strategy - remove these headers before onboarding into the stream. That keeps direct as fast as possible and prevents serialization of any `Nats-` header into the stream saving some space, and removing the need to check/erase/reset these headers downstream.

aricart:
(and yes, I added one function to remove all in one shot)

aricart:
Made a change to constrain the actual headers that are removed as some are handled further down during the Store() operation.

server/stream.go
```diff
@@ -4399,6 +4400,11 @@ func (mset *stream) processJetStreamMsg(subject, reply string, hdr, msg []byte,
 		}
 	}
 
+	// if we are NOT doing subject remapping, this is a message to on-board it shouldn't store Nats-* headers
+	if tsubj == _EMPTY_ {
+		hdr = removeStreamIdentityHeaders(hdr)
```
derekcollison, line 4405: I think this is correct, but big change since we have always stored the message and headers as we received them. Be good to discuss with @wallyqs and @bruth

aricart: More importantly - this doesn't fix the issue for direct get - it simply solves it for the on-boarding. We would still need to have the initial fix on the outbound to ensure that clients won't get duplicate headers.

derekcollison: Yes since they are stored with headers today..

aricart:
@derekcollison this is possibly the correct fix, if we think that we want to reduce the surface area, we can keep the on-boarding of the the stream identity headers, but this is the best we can do. The direct API checks if there's a `Nats-` entry in the headers, and then attempts the removal.

derekcollison:
I am conflicted, I think we need to retain what the app sent us.

PR #4620, opened 2023-10-03
[IMPROVED] Memory growth on compressed websocket connections.

derekcollison:
Holding onto the compressor and not recycling the internal byte slice could cause havoc with GC.

This needs to be improved but this at least should allow the GC to cleanup more effectively.


neilalexander (approved):
LGTM!

PR #4621, opened 2023-10-03
[IMPROVED] Bumped inflight updates to 16 and move one lock to rlock.

derekcollison:


wallyqs (approved):
LGTM

PR #4623, opened 2023-10-04
Reduce allocations in WebSockets

neilalexander:
This PR primarily does two things:

1. ~~Updates the buffer pool to allow reclaiming buffers that have been grown beyond their original capacity by slicing them up into capacities we can make use of~~ (reverted for now);
2. Updates the WebSocket code to remove extra copies, instead drawing straight from the compression buffer, which is reset on each call.

With `nats bench ... --pub=1 --sub=3 --msgs=100000 --size=1024` over non-TLS WebSockets, we have gone from:
* `174,688 msgs/sec ~ 170.59 MB/sec` up to `834,240 msgs/sec ~ 814.69 MB/sec`
* 40GB of `wsCollapsePtoNB` allocations down to 10MB
* 13GB of `flate.NewWriter` allocations down to 3.63MB

### DIFF
--- a/server/client.go
+++ b/server/client.go
@@ -346,23 +346,20 @@ func nbPoolGet(sz int) []byte {
 	}
 }
 
-func nbPoolPut(in []byte) {
-	ca := cap(in)
-	for in = in[:ca]; ca >= nbPoolSizeSmall; ca = cap(in) {
-		switch {
-		case ca >= nbPoolSizeLarge:
-			b := (*[nbPoolSizeLarge]byte)(in[0:nbPoolSizeLarge:nbPoolSizeLarge])
-			nbPoolLarge.Put(b)
-			in = in[nbPoolSizeLarge:]
-		case ca >= nbPoolSizeMedium:
-			b := (*[nbPoolSizeMedium]byte)(in[0:nbPoolSizeMedium:nbPoolSizeMedium])
-			nbPoolMedium.Put(b)
-			in = in[nbPoolSizeMedium:]
-		case ca >= nbPoolSizeSmall:
-			b := (*[nbPoolSizeSmall]byte)(in[0:nbPoolSizeSmall:nbPoolSizeSmall])
-			nbPoolSmall.Put(b)
-			in = in[nbPoolSizeSmall:]
-		}
+func nbPoolPut(b []byte) {
+	switch cap(b) {
+	case nbPoolSizeSmall:
+		b := (*[nbPoolSizeSmall]byte)(b[0:nbPoolSizeSmall])
+		nbPoolSmall.Put(b)
+	case nbPoolSizeMedium:
+		b := (*[nbPoolSizeMedium]byte)(b[0:nbPoolSizeMedium])
+		nbPoolMedium.Put(b)
+	case nbPoolSizeLarge:
+		b := (*[nbPoolSizeLarge]byte)(b[0:nbPoolSizeLarge])
+		nbPoolLarge.Put(b)
+	default:
+		// Ignore frames that are the wrong size, this might happen
+		// with WebSocket/MQTT messages as they are framed
 	}
 }
 
--- a/server/client.go
+++ b/server/client.go
@@ -346,20 +346,23 @@ func nbPoolGet(sz int) []byte {
 	}
 }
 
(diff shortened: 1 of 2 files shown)


server/client.go
```diff
@@ -346,20 +346,23 @@ func nbPoolGet(sz int) []byte {
...
-		nbPoolMedium.Put(b)
-	case nbPoolSizeLarge:
-		b := (*[nbPoolSizeLarge]byte)(b[0:nbPoolSizeLarge])
-		nbPoolLarge.Put(b)
-	default:
-		// Ignore frames that are the wrong size, this might happen
-		// with WebSocket/MQTT messages as they are framed
+func nbPoolPut(in []byte) {
+	ca := cap(in)
+	for in = in[:ca]; ca >= nbPoolSizeSmall; ca = cap(in) {
```
derekcollison, line 351: This is hard to follow. What is this doing?

server/websocket.go
```diff
@@ -1295,7 +1296,13 @@ func (c *client) wsCollapsePtoNB() (net.Buffers, int64) {
...
 		}
 		buf := bytes.NewBuffer(nbPoolGet(usz))
-		cp, _ := flate.NewWriter(buf, flate.BestSpeed)
+		cp := c.ws.compressor
+		if cp == nil {
+			c.ws.compressor, _ = flate.NewWriter(buf, flate.BestSpeed)
+			cp = c.ws.compressor
+		} else {
+			cp.Reset(buf)
+		}
```
derekcollison, line 1305: The issue with this is that we hold onto the compressor and hence the bytes.Buffer and hence the []byte underneath.

server/client.go
neilalexander: At this point the input `in` length might not match the capacity, so we size the slice back up to the full capacity. Then the loop will start checking if the remaining capacity is at least as big as one of the pool sizes and will return 

The idea being that if we've used `append` a lot and the buffer capacity is no longer one of our three standard sizes, like the compressor will with the `bytes.Buffer`, that we can still reclaim the underlying arrays afterwards by chunking it back down into one of our three sizes.

server/websocket.go
neilalexander: The compressor will be held yes, as will the `bytes.Buffer` (which is only a few bytes holding the underlying slice reference), but the actual underlying `[]byte` itself will not be held indefinitely.

Now that we aren't copying from that buffer anymore, the underlying memory is reused and queued up in `queueOutbound` and returned to the pool when either `flushOutbound` or `flushAndClose` completes. On the next pass, we reset the compressor with a new `bytes.Buffer` with new underlying memory from the pool.

server/websocket.go
```diff
@@ -1335,16 +1340,9 @@ func (c *client) wsCollapsePtoNB() (net.Buffers, int64) {
...
 				wsMaskBuf(key, p)
 			}
-			bufs = append(bufs, h)
-			for len(p) > 0 {
-				new := nbPoolGet(len(p))
-				n := copy(new[:cap(new)], p)
-				bufs = append(bufs, new[:n])
-				p = p[n:]
-			}
+			bufs = append(bufs, h, p)
```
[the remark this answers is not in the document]
neilalexander: Done.

server/client.go
derekcollison: I am concerned that might cause GC problems with interior pointers and causing it not to release anything.

server/websocket.go
derekcollison: But unless we do a reset on the compressor it will hold a reference all the way down to the byte slice.

server/client.go
neilalexander: I've reverted the changes to `nbPoolPut` for now, it seems that we don't generate that many slices larger than the largest pool size so we maintain most of the performance increase.

server/websocket.go
neilalexander: The reference is already held anyway by the `queueOutbound` because we're no longer copying, but I've added a `cp.Reset(nil)` to force it to drop the reference as we always end up calling `cp.Reset(buf)` before the next use anyway. That way the GC should be able to do its work post-flush rather than having to wait for the next write.

derekcollison (approved):
LGTM

PR #4631, opened 2023-10-06
[FIXED] Setting initial min on dmap caused subtle bugs with dmap. 

derekcollison:
Under heavy load with max msgs per subject of 1 the dmap, when considered empty and resetting the initial min, could cause lookup misses that would lead to excess messages in a stream and longer restore issues.


neilalexander (approved):
LGTM

PR #4644, opened 2023-10-10
Backport JetStream benchmarks improvements to 2.9.x

mprimi:
Backporting of JetStream benchmarks that landed in `main` (post 2.10.x) and `dev` branches back to 2.9.

These improvements are to the benchmark code and not specific to 2.10. 
Therefore it seems appropriate to back port them, so that benchmarks more accurately compare 2.9 to 2.10 and beyond.

Notice that I cherry-picked the original commits to preserve attribution (and make review easier). However because the back porting can be considered an 'atomic' change, and these have already been reviewed I could also squash into a single commit.

Signed-off-by: Marco Primi <marco@synadia.com>


wallyqs (approved):
LGTM

PR #4645, opened 2023-10-11
[FIXED] Compaction with compression and added out of band compaction

derekcollison:
This will also reclaim more space for streams with lots of interior deletes.

### DIFF
--- a/server/filestore.go
+++ b/server/filestore.go
@@ -4758,7 +4758,7 @@ func (fs *fileStore) syncBlocks() {
 		}
 
 		// Check if we need to sync. We will not hold lock during actual sync.
-		needSync := mb.needSync
+		needSync, fn := mb.needSync, mb.mfn
 		if needSync {
 			// Flush anything that may be pending.
 			mb.flushPendingMsgsLocked()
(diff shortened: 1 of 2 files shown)


server/filestore.go
```diff
@@ -4722,24 +4750,33 @@ func (fs *fileStore) syncBlocks() {
...
 		}
 		mb.mu.Unlock()
 
 		// Check if we need to sync.
 		// This is done not holding any locks.
-		if fn != _EMPTY_ {
-			if fd, _ := os.OpenFile(fn, os.O_RDWR, defaultFilePerms); fd != nil {
-				fd.Sync()
+		if needSync {
+			if fd, _ := os.OpenFile(mb.mfn, os.O_RDWR, defaultFilePerms); fd != nil {
```
neilalexander, line 4771: This might be OK if the only place `mb.mfn` can be changed after init is `dirtyCloseWithRemove`, but it stood out to me so just want to make sure this is safe to access without a lock held?

neilalexander (approved):
LGTM

server/filestore.go
derekcollison: I can grab a copy from above when I have the lock. Will update.

neilalexander (approved):
LGTM, thanks!

PR #4646, opened 2023-10-11
Fixed a crash in MQTT outgoing PUBREL

levb:
This really was a cut/paste/typo error, the `else` should not have been there. Came up in my testing.

The effect was that when there was a pending `PUBREL` in JetStream, and a matching client connects - we would sometimes attempt to deliver the PUBREL immediately once connected. `cpending` was already initialized, but the pubrel map was not (yet).


derekcollison (approved):
LGTM

Issue #4653, opened 2023-10-12
Updating a consumer on a work queue stream with overlapping subjects passes without an error


mdawar:
### What version were you using?

nats-server: `v2.10.2`
nats.go: `v1.30.2`

### What environment was the server running in?

Linux amd64

### Is this defect reproducible?

On a work queue stream, consumers with overlapping subjects are not allowed and on consumer creation we get the error `10100`:

```
filtered consumer not unique on workqueue stream
```

But if we try to update an existing consumer with overlapping subjects we don't get the same error, the update passes and now consumers start receiving the same messages on the overlapping subjects.

I've created some tests that show this bug in this [repo](https://github.com/mdawar/jetstream-api-demos/tree/main/consumer-subjects).

Steps to reproduce:

1. Clone the repo:

```sh
git clone git@github.com:mdawar/jetstream-api-demos.git
```

2. Change directory:

```sh
cd jetstream-api-demos/consumer-subjects
```

3. Install dependencies:

```
go mod download
```

4. Run the tests:

```
go test
```

### Given the capability you are leveraging, describe your expectation?

The expected behavior is to get an error if we try to update an existing consumer with overlapping subjects on a work queue stream.

### Given the expectation, what is the defect you are observing?

For work queue streams, overlapping consumer subjects are checked on create and not on update.

PR #4654, opened 2023-10-12
Fix updating a non unique consumer on workqueue stream not returning an error

mdawar:
This is a possible fix for #4653.

Changes made:
1. Added tests for creating and updating consumers on a work queue stream with overlapping subjects.
2. Check for overlapping subjects before [updating](https://github.com/nats-io/nats-server/blob/a25af02c7348d1e39cce66a5c275e1740fb0ca81/server/consumer.go#L770) the consumer config.
3. Changed [`func (*stream).partitionUnique(partitions []string) bool`](https://github.com/nats-io/nats-server/blob/a25af02c7348d1e39cce66a5c275e1740fb0ca81/server/stream.go#L5269) to accept the consumer name being checked so we can skip it while checking for overlapping subjects (Required for [`FilterSubjects`](https://github.com/nats-io/nats-server/blob/a25af02c7348d1e39cce66a5c275e1740fb0ca81/server/consumer.go#L75) updates), wasn't needed before because the checks were made on creation only. 

There's only 1 thing that I'm not sure about.

In the [current work queue stream conflict checks](https://github.com/nats-io/nats-server/blob/a25af02c7348d1e39cce66a5c275e1740fb0ca81/server/consumer.go#L796), the consumer config `Direct` is being checked if `false`, should we also make this check before the update?

Signed-off-by: Pierre Mdawar <pierre@mdawar.dev>

### DIFF
--- a/server/consumer.go
+++ b/server/consumer.go
@@ -767,6 +767,13 @@ func (mset *stream) addConsumerWithAssignment(config *ConsumerConfig, oname stri
 			if action == ActionCreate && !reflect.DeepEqual(*config, eo.config()) {
 				return nil, NewJSConsumerAlreadyExistsError()
 			}
+			// Check for overlapping subjects.
+			if mset.cfg.Retention == WorkQueuePolicy {
+				subjects := gatherSubjectFilters(config.FilterSubject, config.FilterSubjects)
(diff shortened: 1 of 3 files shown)


derekcollison (approved):
LGTM! Thanks for the contribution as always.

Issue #4643, opened 2023-10-10
KV Get/Watch fails even though key/value is present


anthonyjacques20:
### What version were you using?

`2.10.1`

### What environment was the server running in?

Docker image `2.10.1` was where we originally saw it but able to reproduce it running `nats-server` binary directly on mac M2

### Is this defect reproducible?

Yes the overall issue is reproducible but the specific keys that are erroring is not the same every time.

1. nats-server `2.9.21` running in production (for months)
2. `nats stream backup` (or `nats account backup`, both appeared to cause the same problem)
3. Spin up brand new nats-server `2.10.1`
4. `nats stream restore` (or `nats account restore`) with the data from the old server to the new server
5. Change number of history values in the `kv` with `nats stream edit {bucket} --max-msgs-per-subject=1`
6. We have one KV that is pretty large (~3GB) and we were getting `nats key not found` errors for keys that we know are in the KV
7. The following were ways that we _could_ access the data:
* `nats kv ls {bucket}` would include the "missing" key
* `nats kv history {bucket} {key}` would show the "missing" key where the last entry was a `PUT` (see screenshot)
* `nats kv get {bucket} {key} --revision={revision}`
8. The following did **not** work:
* `nats kv get {bucket} {key}` would give `nats key not found` errors
* `nats kv watch {bucket} {key}`

Some notes:
* Once the server was in this "error" state, the `nats kv get` would fail every time, it wasn't intermittent
* Unable to provide the backups that caused this due to data contents (redacted some info in screenshot)
* This only happened to _some_ of the keys, it wasn't _every_ key
![NatsKVGetFailure](https://github.com/nats-io/nats-server/assets/12755023/289848b3-6373-471e-ac9b-1959ddc53a5a)

### Given the capability you are leveraging, describe your expectation?

We expect that a `KV Get` would return the value stored in that key.

### Given the expectation, what is the defect you are observing?

Some of the keys were showing `nats key not found` error even though the following were true:
* `nats kv ls` showed the key in list
* `nats kv history {bucket} {key}` showed the last entry was a `PUT` (so it should exist)
* `nats kv get {bucket} {key} --revision={revision}` returned correct value

anthonyjacques20:
Also worth noting that I used `nats cli` as an example in the bug but we were seeing the same behavior from the `go client`, so I don't think it's specific to the `nats cli`

ripienaar:
What version CLI? And if next time this happens can you please do the failing get with —trace

[...]

PR #4656, opened 2023-10-12
[FIXED] Stream / KV lookups fail after decreasing history size.

derekcollison:
Fixed a bug that was not correctly selecting next first because it was not properly skipping new dbit entries.
This could result in lookups failing, e.g. after a change in max msgs per subject to a lower value.

Also fixed a bug that would not properly update our psim during compact when throwing away the whole block and a subject had more than one message.

Resolves: #4643


wallyqs (approved):
LGTM

PR #4657, opened 2023-10-12
[FIXED] Only mark fs as dirty vs full write on mb compaction.

derekcollison:
On streams that were constantly removing items, like KVs, this could become over active when not needed. Simply mark the store as dirty for next check.


wallyqs (approved):
LGTM

PR #4660, opened 2023-10-13
Add fan-in/out benchmarks

ReubenMathew:
Benchmarks for NATS core fan-in and fan-out pattern workloads. 

Signed-off-by: Reuben Ninan <reuben@nats.io>

### DIFF
--- a/server/core_benchmarks_test.go
+++ b/server/core_benchmarks_test.go
@@ -19,6 +19,8 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"strconv"
+	"strings"
 	"sync"
 	"testing"
 	"time"
@@ -116,7 +118,7 @@ func BenchmarkCoreTLSFanOut(b *testing.B) {
 
 	// Custom error handler that ignores ErrSlowConsumer.
 	// Lots of them are expected in this benchmark which indiscriminately publishes at a rate higher
-	// than what the server can fan-out to subscribers.
+	// than what the server can relay to subscribers.
 	ignoreSlowConsumerErrorHandler := func(conn *nats.Conn, s *nats.Subscription, err error) {
 		if errors.Is(err, nats.ErrSlowConsumer) {
 			// Swallow this error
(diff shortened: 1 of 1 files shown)


derekcollison (commented):
We have these already, why the need for new things here? Just curious.

mprimi:
@derekcollison the ones we have are 1) "historical" (as in: we don't dare to touch them) and 2) microbenchmarks (e.g. write straight into the socket without a client)

These new ones are more in line with the rest of the benchmarks we've added into the `server`: they use a client, measure end-to-end time to complete some semi-realistic workload, they are multi-threaded, multi-connection, etc.

So in summary, exercise the same core feature, but at higher abstraction layer.

One bonus of the 2 new benchmark in this PR is that they allow us to monitor "fairness".
e.g. single producer fanout to 10 consumers, we don't want 1 consumer to receive 100k messages and another one to receive 100, right? If NATS gets more "unfair" in this dimension, this benchmark will slow down!

Similarly, 10 producers fanning into a single consumer, we don't want any single producer to saturate and the remaining 9 to be discarded. It's not the case now, and if that's the case in the future this benchmark will slow down.

derekcollison:
Please fix..

server/core_benchmarks_test.go:511:23: `succesfully` is a misspelling of `successfully` (misspell)

derekcollison (approved):
LGTM

server/core_benchmarks_test.go
```diff
@@ -249,3 +251,317 @@ func BenchmarkCoreTLSFanOut(b *testing.B) {
...
+								})
+								if err != nil {
+									b.Fatalf("failed to subscribe: %s", err)
+								}
+								err = sub.SetPendingLimits(maxPendingMessages, maxPendingBytes)
+								if err != nil {
+									b.Fatalf("failed to set pending limits: %s", err)
+								}
+								defer sub.Unsubscribe()
+								if err != nil {
```
mprimi, line 335: This is a redundant error check. Strange that IDE/compiler didn't warn you about it..

server/core_benchmarks_test.go
```diff
@@ -249,3 +251,317 @@ func BenchmarkCoreTLSFanOut(b *testing.B) {
...
+							go publish()
+							// Wait for all subscribers to have delivered b.N messages
+							wg.Wait()
+							// Stop the clock
+							b.StopTimer()
+
+							// Stop publisher
+							quitCh <- true
+
+							b.ReportMetric(float64(errorCount), "errors")
```
mprimi, line 385: Error rate would be more informative (`100 * publish errors / publish attempts`) since we don't really know how many times publish gets called.

server/core_benchmarks_test.go
```diff
@@ -249,3 +251,317 @@ func BenchmarkCoreTLSFanOut(b *testing.B) {
...
+							errors := make([]int, numPubs)
+							// track messages received by each publisher
+							counters := make([]int, numPubs)
+							// quit signals for each publisher
+							quitChs := make([]chan bool, numPubs)
+							for i := range quitChs {
+								quitChs[i] = make(chan bool, 1)
+							}
+
+							// TODO rename
```
mprimi, line 462: Leftover TODO comment

server/core_benchmarks_test.go
```diff
@@ -249,3 +251,317 @@ func BenchmarkCoreTLSFanOut(b *testing.B) {
...
+							completedPublishersCount := 0
+
+							var benchCompleteWg sync.WaitGroup
```
mprimi, line 465: Could use a comment, e.g., decremented after subscriber received at least b.N messages from each of the publishers

server/core_benchmarks_test.go
```diff
@@ -249,3 +251,317 @@ func BenchmarkCoreTLSFanOut(b *testing.B) {
...
+							benchCompleteWg.Add(1)
+
+							ncSub.Subscribe(fmt.Sprintf("%s.*", subjectBaseName), func(msg *nats.Msg) {
+								// split subject by "." and get the publisher id
+								pubIdx, err := strconv.Atoi(strings.Split(msg.Subject, ".")[1])
```
mprimi, line 470: Since you know the length of subjectBaseName, you could have also directly gotten the sub-slice without calling `Split`

server/core_benchmarks_test.go
```diff
@@ -249,3 +251,317 @@ func BenchmarkCoreTLSFanOut(b *testing.B) {
...
+								if err != nil {
+									b.Fatal(err)
+								}
+
+								counters[pubIdx] += 1
+								if counters[pubIdx] == b.N {
```
mprimi, line 476: Comment: subscriber has received b.N messages from producer `pubIdx`

server/core_benchmarks_test.go
```diff
@@ -249,3 +251,317 @@ func BenchmarkCoreTLSFanOut(b *testing.B) {
...
+									completedPublishersCount++
+									if completedPublishersCount == numPubs {
+										benchCompleteWg.Done()
```
mprimi, line 479: Comment: subscriber has received b.N messages from all producers

server/core_benchmarks_test.go
```diff
@@ -249,3 +251,317 @@ func BenchmarkCoreTLSFanOut(b *testing.B) {
...
+							publishersReadyWg.Add(numPubs + 1)
+
+							// wait group to ensure all publishers have been torn down
+							var finishedPublishersWg sync.WaitGroup
+							finishedPublishersWg.Add(numPubs)
+
+							// create N publishers
+							for i := 0; i < numPubs; i++ {
+								// create publisher connection
+								ncPub, err := nats.Connect(clientUrl, opts...)
```
mprimi, line 499: Would have pulled all `Connect` into a separate for loop, and only move on to fork go subroutines after all connections are successfully established

server/core_benchmarks_test.go
```diff
@@ -249,3 +251,317 @@ func BenchmarkCoreTLSFanOut(b *testing.B) {
...
+							finishedPublishersWg.Wait()
+
+							// sum errors from all publishers
+							totalErrors := 0
+							for _, err := range errors {
+								totalErrors += err
+							}
+
+							// report errors
+							b.ReportMetric(float64(totalErrors), "errors")
```
mprimi, line 561: Same as above, rate would be more informative, this is just a number that would be hard to make sense of.

mprimi (commented):
@ReubenMathew next time let's keep iterating privately before opening a PR against server. This got merged before i had a chance to review (and wanted to review results together too).

PR #4662, opened 2023-10-15
Record details about the asset a particular raft stream belongs to

ripienaar:
I often encounter sisutation where one of the Raft streams are growing out of control or streams that just wont recover correctly, being able to walk backward from these streams back to their managed asset will help a lot with debugging activities


ripienaar:
@wallyqs please let me know your thoughts

wallyqs (approved):
LGTM, agree this would be convenient to check to which stream/consumer a raft group belongs too.

ripienaar:
Added server to approval list now after Wally as Derek asked specifically for Wallys input here.

PR #4665, opened 2023-10-16
[FIXED]  #4664 mqtt retained messages memory leak

### LINKED ISSUE #4664: MQTT - memory leak when server handle a retained message publish in handleRetainedMsg

2.10.2

All environments affected

STR:
- publish in mqtt protocol some messages with retain flag enabled, incrementing the `sseq` of messages
- keep the same subject for each message
- heaps grows indefinitely as messages are published, and fill the RAM completely

I setup a test to prove the leak in the function `handleRetainedMsg`. I expect that the handler creates a internal subscription only 1 time per subject (if a new messages comes with same subject, then  the same subscription is used). I also expect that when `handleRetainedMsgDel` is called, then subscription related to that subject is deleted
```
func TestMQTTRetainedMsgDel(t *testing.T) {
	o := testMQTTDefaultOptions()
	s := testMQTTRunServer(t, o)
	defer testMQTTShutdownServer(s)
	mc, _ := testMQTTConnect(t, &mqttConnInfo{clientID: "sub", cleanSess: true}, o.MQTT.Host, o.MQTT.Port)
	defer mc.Close()

	c := testMQTTGetClient(t, s, "sub")
	asm := c.mqtt.asm
	var i uint64
        // 3 messages with increasing sseq are published
	for i = 0; i < 3; i++ {
		rf := &mqttRetainedMsgRef{sseq: i}
                //
		asm.handleRetainedMsg("subject", rf)
	}
	asm.handleRetainedMsgDel("subject", 2)
	if asm.sl.count > 0 {
		t.Fatalf("all retained messages subs should be removed, but %d still present", asm.sl.count)
	}
	fmt.Printf("a")
}
```

in `handleRetainedMsg` invokation, additional subscriptios are created even if the message subject is the same of an already mapped message. The result is that 
- the subscription list grows with a lot of subscriptions containing the same subject
- the heap is filled

example handling 1000 messages with same subject and increasing `sseq`:
image

pricelessrabbit:
fixes #4664

### DIFF
--- a/server/mqtt.go
+++ b/server/mqtt.go
@@ -2029,9 +2029,6 @@ func (as *mqttAccountSessionManager) sendJSAPIrequests(s *Server, c *client, acc
 // Add/Replace this message from the retained messages map.
 // If a message for this topic already existed, the existing record is updated
 // with the provided information.
-// This function will return the stream sequence of the record before its update,
-// or 0 if the record was added instead of updated.
-//
 // Lock not held on entry.
 func (as *mqttAccountSessionManager) handleRetainedMsg(key string, rm *mqttRetainedMsgRef) {
 	as.mu.Lock()
@@ -2040,7 +2037,7 @@ func (as *mqttAccountSessionManager) handleRetainedMsg(key string, rm *mqttRetai
 		as.retmsgs = make(map[string]*mqttRetainedMsgRef)
 		as.sl = NewSublistWithCache()
 	} else {
-		// Check if we already had one. If so, update the existing one.
+		// Check if we already had one retained message. If so, update the existing one.
 		if erm, exists := as.retmsgs[key]; exists {
 			// If the new sequence is below the floor or the existing one,
 			// then ignore the new one.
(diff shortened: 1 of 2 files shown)


server/mqtt.go
```diff
@@ -2029,35 +2029,32 @@ func (as *mqttAccountSessionManager) sendJSAPIrequests(s *Server, c *client, acc
...
-			erm.sseq = rm.sseq
-			// Clear the floor
-			erm.floor = 0
-			// If sub is nil, it means that it was removed from sublist following a
-			// network delete. So need to add it now.
-			if erm.sub == nil {
-				erm.sub = &subscription{subject: []byte(key)}
-				as.sl.Insert(erm.sub)
-			}
+	}
```
levb, line 2039: 1/5 keep the `else`, just add the missing `return`; just a very slight optimization, no need to lookup in the new (empty) map?

levb (changes requested):
@pricelessrabbit thank you for tracking this down and the contribution!

one small nit, otherwise LGTM!

server/mqtt.go
pricelessrabbit: sure, i removed it to make more readable but you're right for the optmization

neilalexander:
Thanks for the contribution but you need to sign off before we can merge. Please take a look at [CONTRIBUTING.md](https://github.com/nats-io/nats-server/blob/main/CONTRIBUTING.md#sign-off) for more info.

neilalexander (approved):
Thanks so much, LGTM!

PR #4669, opened 2023-10-17
Fix panic if a store error occurs when requesting consumer info

neilalexander:

### DIFF
--- a/server/consumer.go
+++ b/server/consumer.go
@@ -2543,14 +2543,15 @@ func (o *consumer) infoWithSnapAndReply(snap bool, reply string) *ConsumerInfo {
 
 	// If we are replicated and we are not the leader we need to pull certain data from our store.
 	if rg != nil && rg.node != nil && !o.isLeader() && o.store != nil {
-		if state, err := o.store.BorrowState(); err == nil {
-			info.Delivered.Consumer, info.Delivered.Stream = state.Delivered.Consumer, state.Delivered.Stream
-			info.AckFloor.Consumer, info.AckFloor.Stream = state.AckFloor.Consumer, state.AckFloor.Stream
-			info.NumAckPending = len(state.Pending)
-			info.NumRedelivered = len(state.Redelivered)
-		} else {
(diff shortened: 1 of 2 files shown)


derekcollison (commented):
LGTM

Question, should we make sure if this is the case that the user sees consumer not found?

server/consumer.go
```diff
@@ -2543,7 +2543,12 @@ func (o *consumer) infoWithSnapAndReply(snap bool, reply string) *ConsumerInfo {
 
 	// If we are replicated and we are not the leader we need to pull certain data from our store.
 	if rg != nil && rg.node != nil && !o.isLeader() && o.store != nil {
-		state, _ := o.store.BorrowState()
+		state, err := o.store.BorrowState()
+		if err != nil {
+			o.mu.Unlock()
+			o.srv.Warnf("Error retrieving consumer state from store: %s", err)
```
derekcollison, line 2549: This can happen naturally as a simple race from a consumer being deleted vs consumer info call. Should not be a warn IMO, actually would not even log anything here. This can be normal.

server/jetstream_api.go
```diff
@@ -4120,7 +4120,9 @@ func (s *Server) jsConsumerListRequest(sub *subscription, c *client, _ *Account,
 	}
 
 	for _, o := range obs[offset:] {
-		resp.Consumers = append(resp.Consumers, o.info())
+		if cinfo := o.info(); cinfo != nil {
```
derekcollison, line 4123: This means consumer got deleted yes? So we should remove entry?

derekcollison (changes requested):

server/consumer.go
neilalexander: Fair enough, will remove the log line.

server/jetstream_api.go
neilalexander: This just stops `nil` entries ending up in the consumer list response yep.

derekcollison (approved):
LGTM

PR #4670, opened 2023-10-17
Update dates in licence headers

piotrpio:
Updated dates on files with outdated licenses.

Signed-off-by: Piotr Piotrowski <piotr@synadia.com>

### DIFF
--- a/conf/fuzz.go
+++ b/conf/fuzz.go
@@ -1,4 +1,4 @@
-// Copyright 2020 The NATS Authors
+// Copyright 2020-2023 The NATS Authors
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
 // You may obtain a copy of the License at
--- a/conf/lex.go
+++ b/conf/lex.go
@@ -1,4 +1,4 @@
-// Copyright 2013-2018 The NATS Authors
+// Copyright 2013-2023 The NATS Authors
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
 // You may obtain a copy of the License at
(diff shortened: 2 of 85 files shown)


ripienaar:
Typically the rule is you update a date in the header only when a particular file changes - not the entire repo at a time.  That's how I understood the "rule" anyway?

So we'd totally have files here that should not change.

piotrpio:
@ripienaar all the files changed in this PR had outdated license, I did not touch the ones that were up to date. The only thing I was not sure about were the ones where e.g. header said `2019-2020`, and the file was changed in 2021 - I updated those to 2023 as well, but maybe they should have `2019-2021`? 🤷

ripienaar:
That's what I mean - a file last edited in 2021 should have a header saying 2021 not 2023.  

At least thats how I understand it - but really I think this is a bad idea so I dont know the exact rule, so take my comment rather as we should ask someone who knows rather than you should not do it :)

piotrpio:
@ripienaar Yeah, not sure about the rules as well. My idea here was to keep the license date consistent with last commit date - so when I'm updating a file now because the last commit year > license year, I'm putting 2023, so that it's in line with this exact commit (for future reference).

Again, I did not touch files that were fine, e.g. `server/auth_test.go` has `Copyright 2012-2022` and was last updated in 2022, so it stays without any changes.

ripienaar:
Oh, @piotrpio I missed that you didnt update those files, sorry for the noise in that case I think we're good.

ripienaar:
Be great if someone could write a linter for this stuff that we can just catch them on every PR!

derekcollison:
You update the copyright on changes.. If no changes, leave as is.

piotrpio:
@derekcollison I did that - all those files had outdated license.

piotrpio:
@ripienaar Yeah, linter would be nice, I'll give it a go.

neilalexander:
We tidied a lot of these up recently with a new script in `scripts/updateCopyrights.sh`, so closing.

PR #4676, opened 2023-10-18
systemd service files: use correct special target

philpennock:
Per systemd.special(7) and
<https://www.freedesktop.org/wiki/Software/systemd/NetworkTarget/> the target we want for ordering starting up the nats server is `network-online`, not `network`.

We want to be able to listen for incoming connections, and possibly connect outwards for setting up routes or connecting to external account resolvers.

Without this change, nats server start-up is subject to system-dependent race-conditions for working networking.


neilalexander (approved):
LGTM

Issue #4674, opened 2023-10-18
Account service import duplicates MSG as HMSG with remapping subject


izwerg:
### What version were you using?

nats-server: v2.10.3
configuration:
```
accounts: {
  tenant1: {
    jetstream: enabled
    users: [
      { user: "one", password: "one" }
    ]
    exports: [
      { stream: "DW.>" }
      { service: "DW.>" }
    ]
  }
  global: {
    jetstream: enabled
    users: [
      { user: "global", password: "global" }
    ]
    imports: [
      { stream: { account: tenant1, subject: "DW.>" }, prefix: tenant1 }
      { service: { account: tenant1, subject: "DW.>" }, to: "tenant1.DW.>" }
    ]
  }
}

no_auth_user: global
```

### What environment was the server running in?

- Darwin Kernel Version 22.6.0 arm64
- docker.io/bitnami/nats:2.9.17-debian-11-r1

### Is this defect reproducible?

Subscribe in one terminal:
`nats --user=one --password=one  sub '>'`

Publish in another terminal:
`nats --user=one --password=one pub DW.test.123 'test'`

### Given the capability you are leveraging, describe your expectation?

Receive only one MSG:
```
[#1] Received on "DW.test.123"
test
```

### Given the expectation, what is the defect you are observing?

I receive one MSG and one HMSG duplicate with strange subject without one token (DW.test.123 -> DW.123):
```
[#1] Received on "DW.test.123"
test

[#2] Received on "DW.123"
Nats-Request-Info: {"acc":"tenant1","rtt":2675625}

test
```

Part of nats-server trace log:
```
[17346] 2023/10/18 08:52:54.543902 [TRC] 127.0.0.1:58872 - cid:8 - <<- [SUB >  1]
[17346] 2023/10/18 08:53:04.042908 [TRC] 127.0.0.1:58910 - cid:9 - <<- [PUB DW.test.123 4]
[17346] 2023/10/18 08:53:04.042914 [TRC] 127.0.0.1:58910 - cid:9 - <<- MSG_PAYLOAD: ["test"]
[17346] 2023/10/18 08:53:04.042922 [TRC] 127.0.0.1:58872 - cid:8 - ->> [MSG DW.test.123 1 4]
[17346] 2023/10/18 08:53:04.043297 [TRC] 127.0.0.1:58872 - cid:8 - ->> [HMSG DW.123 1 64 68]
```

[...]

PR #4678, opened 2023-10-19
[FIXED] Avoid adding n shadow subs if the origin sub is a service import.

derekcollison:

Resolves: #4674


wallyqs (approved):
LGTM

PR #4679, opened 2023-10-19
[CHANGED] MQTT no longer requires server name to be set

levb:
It was required because it was required by JetStream, but JetStream now auto-generates a default name.

### DIFF
--- a/server/mqtt.go
+++ b/server/mqtt.go
@@ -197,6 +197,7 @@ var (
 	errMQTTTopicFilterCannotBeEmpty = errors.New("topic filter cannot be empty")
 	errMQTTMalformedVarInt          = errors.New("malformed variable int")
 	errMQTTSecondConnectPacket      = errors.New("received a second CONNECT packet")
+	errMQTTServerNameMustBeSet      = errors.New("mqtt requires server name to be explicitly set")
 	errMQTTUserMixWithUsersNKeys    = errors.New("mqtt authentication username not compatible with presence of users/nkeys")
 	errMQTTTokenMixWIthUsersNKeys   = errors.New("mqtt authentication token not compatible with presence of users/nkeys")
 	errMQTTAckWaitMustBePositive    = errors.New("ack wait must be a positive value")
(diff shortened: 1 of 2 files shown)


bruth (approved):
LGTM, thanks!

server/mqtt.go
```diff
@@ -598,11 +597,6 @@ func validateMQTTOptions(o *Options) error {
 	if mo.Port == 0 {
 		return nil
 	}
-	// We have to force the server name to be explicitly set. There are conditions
-	// where we need a unique, repeatable name.
-	if o.ServerName == _EMPTY_ {
```
derekcollison, line 603: This might be ok for single server mode, but not clustered mode. In clustred mode this is required.

derekcollison: Peer names, which need to be consistent, are a derived hash from server names. So in clustered mode this condition has to hold. So we need to check if we are in single server mode, or not using JetStream, although I thought it was required for MQTT.

levb: JS is required for MQTT; I am 0/5 if it makes any sense to remove the name requirement for single server only. @bruth?

derekcollison: There is a js helper function to test if in single server mode.

bruth: @levb The observation was that enabling jetstream (without MQTT) in single-server mode does not require an explicit server name, so it felt unnecessary only because MQTT is enabled.

levb: K, pushed a fix, please review. @derekcollison this is unfortunately in the context of options, so no convenience methods. Can you please see if I need to add `(... || o.Gateway.Port != 0 || o.LeafNode.Port != 0)`?

derekcollison: Look at what the js helper uses to determine..

derekcollison: Let's do same thing.

levb: I did, unfortunately it just checks that the run-time object is not nil, that it has initialized. This one, https://github.com/nats-io/nats-server/blob/main/server/jetstream_cluster.go#L724?

derekcollison:
```
// Determines if this server is in standalone mode, meaning no routes or gateways.
func (s *Server) standAloneMode() bool {
	opts := s.getOpts()
	return opts.Cluster.Port == 0 && opts.Gateway.Port == 0
}
```

levb: ah! thx!

levb: @derekcollison Can you please confirm

derekcollison (approved):
LGTM

PR #4684, opened 2023-10-20
Raft state machine tweaks

neilalexander:
This PR does the following:

1. Replaces `state` with an atomic, which can be accessed lock-free instead of needing to take the group mutex
2. Ensures that `runAsFollower`, `runAsLeader`, `runAsCandidate` can only continue running as long as the `state` is correct (they will stop looping and hand back to `run` if not)
3. Removes the `isLeader` atomic, as it is now unnecessary with the `state` atomic
4. Adds a unit test to prove that we can now recover from the situation of all nodes being forced into a leaderless follower state, which now passes with the above changes

### DIFF
--- a/server/raft.go
+++ b/server/raft.go
@@ -1499,11 +1499,13 @@ func (n *raft) Delete() {
 }
 
 func (n *raft) shutdown(shouldDelete bool) {
-	if n.State() == Closed {
+	// Needs to be a Swap operation in order to prevent races if
+	// more than one caller hits shutdown() at once, the returned
+	// value will be the previous state.
+	if n.state.Swap(int32(Closed)) == int32(Closed) {
 		return
 	}
 
-	n.state.Store(int32(Closed))
 	n.Lock()
 
 	close(n.quit)
--- a/server/raft.go
+++ b/server/raft.go
@@ -651,12 +651,11 @@ func (s *Server) transferRaftLeaders() bool {
 // Propose will propose a new entry to the group.
 // This should only be called on the leader.
 func (n *raft) Propose(data []byte) error {
-	n.RLock()
 	if state := n.State(); state != Leader {
-		n.RUnlock()
 		n.debug("Proposal ignored, not leader (state: %v)", state)
 		return errNotLeader
 	}
+	n.RLock()
 	// Error if we had a previous write error.
 	if werr := n.werr; werr != nil {
 		n.RUnlock()
@@ -672,12 +671,11 @@ func (n *raft) Propose(data []byte) error {
 // ProposeDirect will propose entries directly.
 // This should only be called on the leader.
 func (n *raft) ProposeDirect(entries []*Entry) error {
-	n.RLock()
 	if state := n.State(); state != Leader {
-		n.RUnlock()
 		n.debug("Direct proposal ignored, not leader (state: %v)", state)
 		return errNotLeader
 	}
+	n.RLock()
 	// Error if we had a previous write error.
 	if werr := n.werr; werr != nil {
 		n.RUnlock()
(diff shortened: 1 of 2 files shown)


server/raft.go
```diff
@@ -654,7 +652,7 @@ func (s *Server) transferRaftLeaders() bool {
 // This should only be called on the leader.
 func (n *raft) Propose(data []byte) error {
 	n.RLock()
-	if state := n.state; state != Leader {
+	if state := n.State(); state != Leader {
```
derekcollison, line 655: If atomic can check this before getting readlock.

server/raft.go
```diff
@@ -675,9 +673,9 @@ func (n *raft) Propose(data []byte) error {
 // This should only be called on the leader.
 func (n *raft) ProposeDirect(entries []*Entry) error {
 	n.RLock()
-	if n.state != Leader {
+	if state := n.State(); state != Leader {
```
derekcollison, line 676: Same here.

server/raft.go
```diff
@@ -707,7 +705,7 @@ func (n *raft) ForwardProposal(entry []byte) error {
 // ProposeAddPeer is called to add a peer to the group.
 func (n *raft) ProposeAddPeer(peer string) error {
 	n.RLock()
-	if n.state != Leader {
+	if n.State() != Leader {
```
derekcollison, line 708: Same.

server/raft.go
```diff
@@ -795,7 +793,7 @@ func (n *raft) AdjustBootClusterSize(csz int) error {
 // Must be the leader.
 func (n *raft) AdjustClusterSize(csz int) error {
 	n.Lock()
-	if n.state != Leader {
+	if n.State() != Leader {
```
derekcollison, line 796: Same.

server/raft.go
```diff
@@ -1788,7 +1784,7 @@ func (n *raft) processAppendEntries() {
 }
 
 func (n *raft) runAsFollower() {
-	for {
+	for n.State() == Follower {
```
derekcollison, line 1787: This just pedantic or did you see a case where we could be running here not as a follower?

neilalexander: In particular it seems we can get stuck in `runAsLeader`. I added the check to all three states just to be sure.

derekcollison: Meaning we are in runAsLeader but not the leader? Do we have a test to reproduce?

neilalexander: I am not sure if indefinitely or whether it's just taking a while for `runAsLeader` to stop and switch to another, but changing `for n.State() == Leader {` back to `for {` causes the new test in the PR to fail.

derekcollison: ok thanks for the explanation.

neilalexander:
Currently investigating:
```
=== RUN   TestJetStreamClusterDetectOrphanNRGs
panic: close of closed channel
goroutine 72044 [running]:
github.com/nats-io/nats-server/v2/server.(*raft).shutdown(0xc014fc6300, 0x0)
	/home/travis/gopath/src/github.com/nats-io/nats-server/server/raft.go:1509 +0xb2
github.com/nats-io/nats-server/v2/server.(*raft).Stop(0xc012ca67a0?)
	/home/travis/gopath/src/github.com/nats-io/nats-server/server/raft.go:1494 +0x29
github.com/nats-io/nats-server/v2/server.(*jetStream).monitorConsumer(0xc0043a9000, 0xc00024b900, 0xc00171da70)
	/home/travis/gopath/src/github.com/nats-io/nats-server/server/jetstream_cluster.go:4603 +0x1533
github.com/nats-io/nats-server/v2/server.(*jetStream).processClusterCreateConsumer.func1()
	/home/travis/gopath/src/github.com/nats-io/nats-server/server/jetstream_cluster.go:4295 +0x47
github.com/nats-io/nats-server/v2/server.(*Server).startGoRoutine.func1()
	/home/travis/gopath/src/github.com/nats-io/nats-server/server/server.go:3700 +0x5a
created by github.com/nats-io/nats-server/v2/server.(*Server).startGoRoutine in goroutine 71984
	/home/travis/gopath/src/github.com/nats-io/nats-server/server/server.go:3698 +0x1bb
FAIL	github.com/nats-io/nats-server/v2/server	370.200s
FAIL
```

derekcollison (approved):
LGTM

derekcollison (approved):
LGTM

PR #4686, opened 2023-10-21
[IMPROVED] Routing: reduce account lookup for pinned accounts

kozlovic:
Small optimization that bypass lookup of account for a pinned account's route since the account is known for this route. The lookup would occur for every inbound message whose exact destination was not found in the cache.
Also, since configuration reload no long swaps account's sublist, we can do without the account's locking to get the sublist pointer.


derekcollison (approved):
LGTM!

PR #4687, opened 2023-10-22
[IMPROVED] WriteFullState Allocations

derekcollison:
Avoid additional allocations when doing a writeFullState for large state.
Track the total subject length of all subjects in the PSIM structure.
This is to optimize the writeFullState when the state is large due to a large PSIM with large subjects.


neilalexander (approved):
LGTM

PR #4690, opened 2023-10-23
Ensure Windows signal handler is stopped when shutting down via other means

neilalexander:
We weren't monitoring `quitCh` on Windows so we could leak goroutines if it wasn't a signal that caused us to shut down.


derekcollison (approved):
LGTM

PR #4692, opened 2023-10-24
[IMPROVED] Updated how we recover mid-update blocks on recoverFullState.

derekcollison:
Before we could encounter a message in the new message block that was deleted and that would cause a full rebuild, which this approach avoids. This could happen on very active KeyValue stores.


neilalexander (approved):
LGTM

Issue #4685, opened 2023-10-20
For K/V with history, WatchAll does not return all keys if jetstream.MemoryStorage is used


a-h:
### What version were you using?

github.com/nats-io/nats-server/v2 v2.10.3
github.com/nats-io/nats.go v1.31.0

### What environment was the server running in?

Linux, x86-64

### Is this defect reproducible?

Yes, I've created a full reproduction here.

https://github.com/a-h/natswatchalltest

### Given the capability you are leveraging, describe your expectation?

I expect that if I...

Put 1 item into a K/V bucket that has history enabled, update it, then add 4 more items... then `WatchAll` should a goroutine which will return 5 items (error handling elided for brevity).

```go
// Add 5 items to table.
_, err = kv.Put(ctx, "user1", []byte("value: a"))

// If I comment this update out, I get the expected results...
// But if I leave it in, the iteration stops at user 4.
_, err = kv.Update(ctx, "user1", []byte("value: a"), 1)

_, err = kv.Put(ctx, "user2", []byte("value: b"))
_, err = kv.Put(ctx, "user3", []byte("value: c"))
_, err = kv.Put(ctx, "user4", []byte("value: d"))
_, err = kv.Put(ctx, "user5", []byte("value: e"))

watcher, err := kv.WatchAll(ctx)
for update := range watcher.Updates() {
	if update == nil {
		break
	}
	log.Printf("Got %s\n", update.Key())
}
watcher.Stop()
```

### Given the expectation, what is the defect you are observing?

I only receive an output of 4 items.

```
2023/10/20 15:18:19 Got user1
2023/10/20 15:18:19 Got user2
2023/10/20 15:18:19 Got user3
2023/10/20 15:18:19 Got user4
```

a-h:
It worked fine with the latest CLI against NATS server v2.9.11 (the one that's in nixpkgs), and the latest server (v2.10.3).

```
nats kv add --history=10 test_kv
nats kv put test_kv user1 "a"
nats kv update test_kv user1 "a" 1
nats kv put test_kv user2 "b"
nats kv put test_kv user3 "c"
nats kv put test_kv user4 "d"
nats kv put test_kv user5 "e"
nats kv watch test_kv
```

I read through the source code of the NATS CLI, and it uses the same Go modules I've used, in the same way. So I worked out that it was a problem with the server, and realised that the issue went away if I removed the configuration to use `MemoryStorage`.

See reproduction a thttps://github.com/a-h/natswatchalltest

[...]

PR #4693, opened 2023-10-24
[FIXED] Fix for a bug that would adjust down num pending when a valid msg was present.

derekcollison:
When a message was detected out of range for filteredState we would mark an adjustment when we encountered a partial. This was not properly taking into account for lastBySubject when another message on that subject was valid past the first sequence.

Resolves: #4685 


neilalexander (approved):
LGTM

PR #4697, opened 2023-10-24
Extend the term advisory with a reason

ripienaar:
* Support getting a reason from the client doing a term
 * Supply some reasons when terminating messages during limits processing

### DIFF
--- a/server/consumer.go
+++ b/server/consumer.go
@@ -272,6 +272,10 @@ var (
 	AckNext = []byte("+NXT")
 	// Terminate delivery of the message.
 	AckTerm = []byte("+TERM")
+
+	// reasons to supply when terminating messages using limits
+	ackTermLimitsReason        = "Message deleted by stream limits"
+	ackTermUnackedLimitsReason = "Unacknowledged message was deleted"
 )
 
 // Calculate accurate replicas for the consumer config with the parent stream config.
@@ -1991,8 +1995,12 @@ func (o *consumer) processAck(subject, reply string, hdr int, rmsg []byte) {
 		o.processNak(sseq, dseq, dc, msg)
 	case bytes.Equal(msg, AckProgress):
 		o.progressUpdate(sseq)
-	case bytes.Equal(msg, AckTerm):
-		o.processTerm(sseq, dseq, dc)
+	case bytes.HasPrefix(msg, AckTerm):
+		var reason string
+		if buf := msg[len(AckTerm):]; len(buf) > 0 {
+			reason = string(bytes.TrimSpace(buf))
+		}
+		o.processTerm(sseq, dseq, dc, reason)
 	}
 
 	// Ack the ack if requested.
@@ -2319,7 +2327,7 @@ func (o *consumer) processNak(sseq, dseq, dc uint64, nak []byte) {
 }
 
 // Process a TERM
-func (o *consumer) processTerm(sseq, dseq, dc uint64) {
+func (o *consumer) processTerm(sseq, dseq, dc uint64, reason string) {
 	// Treat like an ack to suppress redelivery.
 	o.processAckMsg(sseq, dseq, dc, false)
 
@@ -2338,6 +2346,7 @@ func (o *consumer) processTerm(sseq, dseq, dc uint64) {
 		ConsumerSeq: dseq,
 		StreamSeq:   sseq,
 		Deliveries:  dc,
+		Reason:      reason,
 		Domain:      o.srv.getOpts().JetStreamDomain,
 	}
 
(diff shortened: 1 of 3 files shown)


derekcollison (commented):
I like the concept of a reason, but not sure of approach of simply tacking a random string onto the term msg itself.

server/consumer.go
```diff
@@ -5152,7 +5163,7 @@ func (o *consumer) decStreamPending(sseq uint64, subj string) {
 	if wasPending {
 		// We could have lock for stream so do this in a go routine.
 		// TODO(dlc) - We should do this with ipq vs naked go routines.
-		go o.processTerm(sseq, p.Sequence, rdc)
+		go o.processTerm(sseq, p.Sequence, rdc, "")
```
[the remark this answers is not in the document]
derekcollison: Could leave empty or message was deleted.

ripienaar:
> I like the concept of a reason, but not sure of approach of simply tacking a random string onto the term msg itself.

Took my queue from AckNext which takes a simple form and a JSON form for more complex stuff, if you have a prefered alternative approach happy to change here

server/consumer.go
```diff
@@ -1988,8 +1994,12 @@ func (o *consumer) processAck(subject, reply string, hdr int, rmsg []byte) {
 		o.processNak(sseq, dseq, dc, msg)
 	case bytes.Equal(msg, AckProgress):
 		o.progressUpdate(sseq)
-	case bytes.Equal(msg, AckTerm):
-		o.processTerm(sseq, dseq, dc)
+	case bytes.HasPrefix(msg, AckTerm):
+		var reason string
+		if len(msg) > ackTermLen+1 {
```
derekcollison, line 1999: Let's use same pre-processing logic as AckNext. 

```go
if buf := msg[len(AckTerm):]; len(buf) > 0 {
  reason := string(bytes.TrimSpace(buf))
}
```

server/consumer.go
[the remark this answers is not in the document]
derekcollison: This happens when we have an unacked message that is no longer available.

ripienaar:
Updated and rebased

server/consumer.go
```diff
@@ -272,6 +272,10 @@ var (
 	AckNext = []byte("+NXT")
 	// Terminate delivery of the message.
 	AckTerm = []byte("+TERM")
+
+	// reasons to supply when terminating messages using limits
+	ackTermLimitsReason        = "Message deleted by stream limits"
+	ackTermUnackedLimitsReason = "Unacknowledged message deleted by stream limits"
```
derekcollison, line 278: Could be user initiated, think purge. So would just do ""Unacknowledged message was deleted"

derekcollison (changes requested):
LGTM - minor nit then we should be good.

server/consumer.go
ripienaar: OK pushed an update

derekcollison (approved):
LGTM

server/consumer.go
```diff
@@ -1991,8 +1995,12 @@ func (o *consumer) processAck(subject, reply string, hdr int, rmsg []byte) {
 		o.processNak(sseq, dseq, dc, msg)
 	case bytes.Equal(msg, AckProgress):
 		o.progressUpdate(sseq)
-	case bytes.Equal(msg, AckTerm):
-		o.processTerm(sseq, dseq, dc)
+	case bytes.HasPrefix(msg, AckTerm):
+		var reason string
+		if buf := msg[len(AckTerm):]; len(buf) > 0 {
```
ripienaar, line 2000: @scottf had a good question here, do you think we should limit the length of the reason @derekcollison ?

server/consumer.go
```diff
@@ -1991,8 +1995,12 @@ func (o *consumer) processAck(subject, reply string, hdr int, rmsg []byte) {
 		o.processNak(sseq, dseq, dc, msg)
 	case bytes.Equal(msg, AckProgress):
 		o.progressUpdate(sseq)
-	case bytes.Equal(msg, AckTerm):
```
ripienaar, line 1994: I didnot realize this will land in 2.10.

Concern here is that if on a < 2.10.4 the term ack is sent it will be ignored - as in the entire ack will be ignored not just the reason -  since this check here is `Equal` and term with reason wont be equal.

So I'd be curous on thoughts here on when do we surface it in clients @derekcollison @Jarema or any thoughts on how to do this while being backward compat

neilalexander: This doesn't have to go into 2.10, we can just not cherry-pick it into the 2.10.x release branches.

ripienaar: Unfortunately its already in 2.10.4 by the looks of things :(

server/consumer.go
scottf: Also, the content? Like should we restrict it to printable, should we encode it? Maybe we should wrap in in json {"reason":"blah whatever b/c it's json encoded"}

derekcollison: Will be limited by max msg payload. 
Right now see as a closure that R.I. thought would be important.

ripienaar: I am not attached to this format. I just made it the same as the others so if someone has a suggestion for change we can do that

PR #4698, opened 2023-10-24
Don't purge entire stream for sequence 1

### LINKED ISSUE #4696: `purge` command purges entire stream with `sequenceId=1`
### Observed behavior

When purging a stream (either via CLI or node JSM client), a `sequenceId` parameter value of 1 will cause the entire stream to be purged.

State prior to purge:
```
State:
    Messages: 2
    Bytes: 249 B
    FirstSeq: 1 @ 2023-10-24T12:49:25 UTC
    LastSeq: 2 @ 2023-10-24T12:49:25 UTC
    Active Consumers: 0
    Number of Subjects: 2
```

Either of the following will produce the behavior:
- CLI Command: `nats stream purge stream_name --seq=1`
- Node Client: `await jetstreamManager.streams.purge(stream, {seq: 1})`

State after purge:
```
State:

    Messages: 0
    Bytes: 0 B
    FirstSeq: 3 @ 0001-01-01T00:00:00 UTC
    LastSeq: 2 @ 2023-10-24T12:49:25 UTC
    Active Consumers: 0
```

### Expected behavior

The node client docs state that the `sequenceId` parameter has the following behavior:
> Purge all messages up to but not including the message with this sequence.

So a value of 1 should leave the stream unchanged

### Server and client version

Server: 2.9.23
Node client: 2.13.1
CLI client: 0.0.35

### Host environment

_No response_

### Steps to reproduce

- Create a limits-retention stream
- Publish a few messages to it
- Run `nats stream purge stream_name --seq=1`

neilalexander:
Fixes #4696.

### DIFF
--- a/server/filestore.go
+++ b/server/filestore.go
@@ -5989,7 +5989,7 @@ func compareFn(subject string) func(string, string) bool {
 // Will return the number of purged messages.
 func (fs *fileStore) PurgeEx(subject string, sequence, keep uint64) (purged uint64, err error) {
 	if subject == _EMPTY_ || subject == fwcs {
-		if keep == 0 && (sequence == 0 || sequence == 1) {
+		if keep == 0 && sequence == 0 {
 			return fs.Purge()
 		}
 		if sequence > 1 {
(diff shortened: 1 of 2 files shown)


derekcollison (commented):
I feel this must be a reason that conditional was added. Tests pass, so maybe an artifact of something in the past that is not longer applicable?

derekcollison (approved):
LGTM

PR #4699, opened 2023-10-24
[IMPROVED] Recycle buffers on rebuild and write out our full state prior to snapshotting

derekcollison:


wallyqs (approved):
LGTM

Issue #4681, opened 2023-10-19
Partially cut off nats-server in a cluster of three nats-servers


maxmoehl:
### Observed behavior

We observed that one nats-server in a cluster of three was no longer receiving / sending all messages. This led to messages from producers connected to that nats-server being lost.

Our messages are sent in intervals of 20s, we could see that instance 1 was receiving messages which it then relayed only to instance 2. Subscribers on instance 3 did not receive the messages:

1 <-> 2 ✅
1 <-> 3 ❌
2 <-> 3 ✅

We checked this via the `/routez` endpoint:
```
$ curl -i 10.5.1.3:8223/routez # this is instance 3
[...]
    { # instance 2, other connections were inactive
      "ip": "10.2.1.3",
      "port": 4225,
      "in_msgs": 4099266,
      "out_msgs": 1269,
      "in_bytes": 3096262046,
      "out_bytes": 893231,
      "subscriptions": 5,
      "compression": "off"
    }
[...]
    { # instance 1, all connections inactive
       "ip": "10.1.1.3",
      "port": 58410,
      "start": "2023-10-13T05:25:37.293102354Z",
      "last_activity": "2023-10-13T05:25:37.3046772Z",
      "rtt": "1.560618ms",
      "uptime": "4d4h16m14s",
      "idle": "4d4h16m14s",
      "pending_size": 0,
      "in_msgs": 0,
      "out_msgs": 0,
      "in_bytes": 0,
      "out_bytes": 0,
      "subscriptions": 0,
      "compression": "off"
    },
[...]
```

We are not yet sure what caused this issue but we upgraded to 2.10.x about a day before we started seeing this issue and never had issues like this before. Could it be related to the new connection pooling feature introduced with #4001?

The issue occurred on Azure and are aware that there was some service degradation in that timeframe but it is not yet clear if this issue is related to it.

### Expected behavior

All messages are reliably delivered from all producers to all consumers.

If a connection breaks, nats notices and falls back to another connection or re-establishes the connection.

### Server and client version

nats-server v2.10.1
nats.go multiple

### Host environment

Provider: Microsoft Azure
Host OS: Ubuntu 22.04
Architecture: amd64
Deployed via [bosh](https://bosh.io/docs/), compiled [like this](https://github.com/cloudfoundry/nats-release/blob/dcd16e1c373a289222f7d824f0de8a96942588c9/packages/nats-server/packaging#L15):
```
[...]

PR #4701, opened 2023-10-24
[IMPROVED] Routing: upgrade non solicited routes if present in config

kozlovic:
When a server accepts a route, that route would not be marked as solicited or configured while it should if it was present in the configuration.

Prior to pooling, if server A pointed to server B and vice-versa, one of the server will have the route as solicited and the server that accepted the route would have upgraded it as solicited. With pooling code, this was not always happening.

Maybe related to #4681 


derekcollison (approved):
LGTM - Will wait on tests green before merge.

derekcollison (approved):
LGTM

PR #4703, opened 2023-10-25
[FIXED] Avoid concurrent consumer setLeader calls to run multiple leaders.

derekcollison:
Be more stringent on setting the leader. Use atomic swap to make sure only one is running as leader.
Had a situation where the consumer processed multiple setLeader calls and had multiple loopAndGather routines running.


wallyqs (approved):
LGTM

PR #4705, opened 2023-10-25
[ADDED] Config options to disable MQTT QOS2 support

levb:
Added MQTT config options `reject_qos2_publish` and `downgrade_qos2_subscriptions`. When used together, fully disable QOS2 support (effectively going back to v2.9 level of QOS support).

### DIFF
--- a/server/mqtt.go
+++ b/server/mqtt.go
@@ -345,6 +345,14 @@ type mqtt struct {
 	asm  *mqttAccountSessionManager // quick reference to account session manager, immutable after processConnect()
 	sess *mqttSession               // quick reference to session, immutable after processConnect()
 	cid  string                     // client ID
+
+	// rejectQoS2Pub tells the MQTT client to not accept QoS2 PUBLISH, instead
+	// error and terminate the connection.
+	rejectQoS2Pub bool
+
+	// downgradeQOS2Sub tells the MQTT client to downgrade QoS2 SUBSCRIBE
+	// requests to QoS1.
+	downgradeQoS2Sub bool
 }
 
 type mqttPending struct {
@@ -475,7 +483,11 @@ func (s *Server) createMQTTClient(conn net.Conn, ws *websocket) *client {
 	}
 	now := time.Now()
 
-	c := &client{srv: s, nc: conn, mpay: maxPay, msubs: maxSubs, start: now, last: now, mqtt: &mqtt{}, ws: ws}
+	mqtt := &mqtt{
+		rejectQoS2Pub:    opts.MQTT.rejectQoS2Pub,
+		downgradeQoS2Sub: opts.MQTT.downgradeQoS2Sub,
+	}
+	c := &client{srv: s, nc: conn, mpay: maxPay, msubs: maxSubs, start: now, last: now, mqtt: mqtt, ws: ws}
 	c.headers = true
 	c.mqtt.pp = &mqttPublish{}
 	// MQTT clients don't send NATS CONNECT protocols. So make it an "echo"
@@ -2226,6 +2238,10 @@ func (as *mqttAccountSessionManager) processSubs(sess *mqttSession, c *client,
 		if f.qos > 2 {
 			f.qos = 2
 		}
+		if c.mqtt.downgradeQoS2Sub && f.qos == 2 {
+			c.Warnf("Downgrading subscription QoS2 to QoS1 for %q, as configured", f.filter)
+			f.qos = 1
+		}
 		subject := f.filter
 		sid := subject
 
@@ -2328,6 +2344,10 @@ func (as *mqttAccountSessionManager) serializeRetainedMsgsForSub(sess *mqttSessi
 		qos := mqttGetQoS(rm.Flags)
 		if qos > sub.mqtt.qos {
 			qos = sub.mqtt.qos
+		}
+		if c.mqtt.rejectQoS2Pub && qos == 2 {
+			c.Warnf("Rejecting retained message with QoS2 for subscription %q, as configured", sub.subject)
+			continue
 		}
 		if qos > 0 {
 			pi = sess.trackPublishRetained()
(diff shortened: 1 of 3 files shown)


server/opts.go
```diff
@@ -4631,6 +4639,11 @@ func parseMQTT(v interface{}, o *Options, errors *[]error, warnings *[]error) er
 		case "consumer_inactive_threshold", "consumer_auto_cleanup":
 			o.MQTT.ConsumerInactiveThreshold = parseDuration("consumer_inactive_threshold", tk, mv, errors, warnings)
 
+		case "reject_qos2_publish":
+			o.MQTT.rejectQoS2Pub, _ = mv.(bool)
```
kozlovic, line 4643: Don't need the `, _` here and below.

server/mqtt_test.go
```diff
@@ -2160,6 +2203,28 @@ func TestMQTTPublish(t *testing.T) {
...
+	defer nc.Close()
+
+	mcp, mpr := testMQTTConnect(t, &mqttConnInfo{cleanSess: true}, o.MQTT.Host, o.MQTT.Port)
+	defer mcp.Close()
+	testMQTTCheckConnAck(t, mpr, mqttConnAckRCConnectionAccepted, false)
+
+	testMQTTPublish(t, mcp, mpr, 1, false, false, "foo", 1, []byte("msg"))
+
+	testMQTTPublishNoAcks(t, mcp, 2, false, false, "foo", 2, []byte("msg"))
+	_, err := mpr.readByte("failed attempt")
```
kozlovic, line 2222: Wonder if you could replace this and below with:
```
testMQTTExpectDisconnect(t, mcp)
```
since we expect a disconnect, no?

server/opts.go
```diff
@@ -562,6 +562,14 @@ type MQTTOpts struct {
 
 	// Snapshot of configured TLS options.
 	tlsConfigOpts *TLSConfigOpts
+
+	// rejectQoS2Pub tells the MQTT client to not accept QoS2 PUBLISH, instead
+	// error and terminate the connection.
+	rejectQoS2Pub bool
```
kozlovic, line 568: Unless this is temporary, the option names need to start with a upper-case: `RejectQoS2Pub` to be exported, so that users can set this option when importing NATS Server code (embedding it). It works for you because the test belongs to the same `server` package. It would not from the `test` package for instance because these fields would then not be visible.

server/mqtt_test.go
```diff
@@ -2160,6 +2203,28 @@ func TestMQTTPublish(t *testing.T) {
...
+	if err == nil || !strings.Contains(err.Error(), "error reading failed attempt: EOF") {
+		t.Fatalf("Expected error about QoS 2 publish rejected, got %v", err)
+	}
+}
+
```
kozlovic, line 2227: Should we also add a test for a QoS2 will message in connect (and make sure that the connection fails)?

kozlovic (changes requested):

server/opts.go
levb: This was deliberate - I am considering this a temporary fix, we can always make it public and documented later if we decide it worthwhile.

server/mqtt_test.go
levb: 👍 ditto for dropping retained QoS2 messages; I was just trying to get this done quickly

derekcollison:
Where are we on this?

server/opts.go
kozlovic: Ok, understood.

kozlovic (approved):
Understood for the naming of the options. You can do the other changes (the `, _` and the expect disconnect) without adding the new tests, or merge as-is if pressed by time.

levb:
@derekcollison finishing the extra tests that @kozlovic requested (I agree, they are needed), give me a few minutes

server/mqtt_test.go
levb: @kozlovic Can you please review the added tests?

server/mqtt.go
```diff
@@ -85,6 +85,7 @@ const (
 	mqttConnAckRCServerUnavailable           = byte(0x3)
 	mqttConnAckRCBadUserOrPassword           = byte(0x4)
 	mqttConnAckRCNotAuthorized               = byte(0x5)
+	mqttConnAckRCQoS2WillRejected            = byte(0x10)
```
levb, line 88: @kozlovic I added this RC, non-continuos with the rest of them because again, hopefully it's temporary; otherwise the CONNECT was dropped and EOF returned instead of a CONNACK with RC. The new test revealed it.

server/mqtt_test.go
```diff
@@ -3920,6 +3918,35 @@ func TestMQTTWill(t *testing.T) {
 	}
 }
 
+func TestMQTTQoS2WillReject(t *testing.T) {
+	o := testMQTTDefaultOptions()
+	o.MQTT.rejectQoS2Pub = true
+	s := testMQTTRunServer(t, o)
+	defer testMQTTShutdownServer(s)
+
+	nc := natsConnect(t, s.ClientURL())
```
kozlovic, line 3927: Not sure you need that, but ok to leave.

server/mqtt_test.go
```diff
@@ -4152,6 +4179,39 @@ func TestMQTTPublishRetain(t *testing.T) {
...
+	mc1, rs1 := testMQTTConnect(t, &mqttConnInfo{cleanSess: true}, o.MQTT.Host, o.MQTT.Port)
+	testMQTTCheckConnAck(t, rs1, mqttConnAckRCConnectionAccepted, false)
+	testMQTTPublish(t, mc1, rs1, 2, false, true, "foo", 1, []byte("qos2 failed"))
+	testMQTTPublish(t, mc1, rs1, 1, false, true, "bar", 2, []byte("qos1 retained"))
+	testMQTTFlush(t, mc1, nil, rs1)
+	testMQTTDisconnect(t, mc1, nil)
+	mc1.Close()
+	s.Shutdown()
+
+	// restart the server with MQTT disabled; we should be using the same
```
kozlovic, line 4194: Not sure the comment is related to what we do here?

kozlovic (approved):
Small copy/paste issues related to NATS connection and comment, but other than that, LGTM.

server/mqtt_test.go
levb: Added/clarified the comment, s/MQTT/MQTT QoS2/

kozlovic (approved):
LGTM

PR #4709, opened 2023-10-25
Trim unnecessary dependencies for `ordered` generic constraint.

misterpickypants:
A generic test function was introduced in #4515 using `constraints.Ordered` from golang.org/x/exp/constaints. This unfortunately introduces further transitive dependencies on packages such as golang.org/x/tools which has versions with various breaking changes. This makes it more difficult to pull in nats-server in downstream projects.

By inlining the trivial `ordered` definition, we can trim this dependency.

An alternative to this change would be to downgrade the required version of golang.org/x/exp since the constraints package has been there (and unchanged) since 2022-02-02.

Signed-off-by: Augusto Roman <augusto.roman@tanium.com>


wallyqs (approved):
LGTM

Issue #4680, opened 2023-10-19
Object store - List bucket's content is very slow


StanEgo:
### Observed behavior

Listing items in the specific bucket of the object store is very slow, even if there is only one item.

### Expected behavior

Expected to get a list of the items faster.

### Server and client version

nats-server 2.10.3
natscli 0.1.1 (that is recommended here https://docs.nats.io/nats-concepts/what-is-nats/walkthrough_setup)
NATS.Client for .NET 1.1.0

### Host environment

Microsoft Windows NT 10.0.19045.0
CPU: AMD Ryzen 9 5950X
NVMe: ADATA SX8200PNP
```
[68776] 2023/10/19 13:47:51.238565 [INF] ---------------- JETSTREAM ----------------
[68776] 2023/10/19 13:47:51.238565 [INF]   Max Memory:      95.94 GB
[68776] 2023/10/19 13:47:51.238565 [INF]   Max Storage:     1.00 TB
[68776] 2023/10/19 13:47:51.238565 [INF]   Store Directory: "e:\Temp\nats\jetstream"
[68776] 2023/10/19 13:47:51.238565 [INF] -------------------------------------------
```

### Steps to reproduce

I have created a bucket as in docs: `nats obj add testbucket`, and added a big file to it (~18G): `nats obj add testbucket somefile.mp4`. When trying to get the list of the objects in the bucket: `nats obj ls testbucket` I have received a message "nats: error: context deadline exceeded" with some delay.

In the server log I got this message:
`[68776] 2023/10/19 14:48:10.303340 [WRN] Internal subscription on "$JS.API.CONSUMER.CREATE.OBJ_testbucket.32td3LZE.$O.testbucket.M.>" took too long: 6.8698428s`

When upgraded timeout to 10s: `/nats obj ls testbucket --timeout=10s` it was ok.

I supposed that this is client-side problems and created a simple app using the latest NATS.Client for .NET 1.1.0:
```
using NATS.Client;
using NATS.Client.JetStream;
using NATS.Client.ObjectStore;

ConnectionFactory factory = new();
IConnection connection = factory.CreateConnection();

var jso = JetStreamOptions
    .Builder()
    .WithOptOut290ConsumerCreate(true)
    .Build()
;

IObjectStore os = connection.CreateObjectStoreContext("testbucket", ObjectStoreOptions.Builder(jso).Build());

DateTime start = DateTime.Now;
var response = os.GetList();
Console.WriteLine(DateTime.Now - start);
```

[...]

PR #4712, opened 2023-10-26
[IMPROVED] Improved time to select skip list and starting sequence number for deliver last by subject

derekcollison:
Previously we would need to move through all msg blocks and load all of the blocks into memory. We optimized this regardless to flush the cache if we loaded it like other places in the code, but for this specifically we load subject's totals which is already in memory and for each matching subject we now simply load the last message for that subject and add the sequence number to the skip list.

This drastically improves consumers that want last per subject, like KV watchers and object store lists.

Resolves: #4680

### DIFF
--- a/server/consumer.go
+++ b/server/consumer.go
@@ -4657,16 +4657,6 @@ type lastSeqSkipList struct {
 	seqs   []uint64
 }
 
-// Will create a skip list for us from a store's subjects state.
-func createLastSeqSkipList(mss map[string]SimpleState) []uint64 {
-	seqs := make([]uint64, 0, len(mss))
-	for _, ss := range mss {
-		seqs = append(seqs, ss.Last)
-	}
-	sort.Slice(seqs, func(i, j int) bool { return seqs[i] < seqs[j] })
-	return seqs
-}
-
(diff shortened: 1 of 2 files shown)


wallyqs (approved):
LGTM

server/consumer.go
```diff
@@ -4697,38 +4687,36 @@ func (o *consumer) selectStartingSeqNo() {
...
-							resume: state.LastSeq,
-							seqs:   createLastSeqSkipList(mss),
-						}
-						o.sseq = o.lss.seqs[0]
-					} else {
-						// If no mapping info just set to last.
-						o.sseq = state.LastSeq
+					filters = append(filters, o.cfg.FilterSubject)
+				} else {
+					for _, filter := range o.subjf {
```
neilalexander, line 4696: Could potentially preallocate `filters` to the length of `o.subjf` here.

neilalexander (approved):
LGTM

server/consumer.go
derekcollison: This will be a small number or 1 in majority of cases.

PR #4713, opened 2023-10-27
Make sure if we have lots of subjects to fallback to subjects state

derekcollison:
This is optimization for last change, the get last works but not if number of subjects is very large, for that fall back to subjects state.


wallyqs (approved):
LGTM

PR #4714, opened 2023-10-27
[IMPROVED] Optimize on last by subject if mmps is one

derekcollison:
Also on SubjectsState only loadMsgs if the mb.fss is nil.


wallyqs (approved):
LGTM

PR #4717, opened 2023-10-28
Enable stale issue/PR workflow

bruth:
This will mark an issue/PR as stale after 8 weeks
of complete inactivity (including comments). If/when new activity occurs, the stale label will be removed.

The purpose of this is to make it simplier for triagers to filter on stale issues to follow-up on.

### DIFF
--- a/.github/workflows/stale-issues.yaml
+++ b/.github/workflows/stale-issues.yaml
@@ -14,8 +14,8 @@ jobs:
     steps:
       - uses: actions/stale@v8
         with:
-          debug-only: true # Set until the behavior is tuned.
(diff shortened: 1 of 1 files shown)


bruth:
@derekcollison Worth considering? There is no auto-closing, merely adding a `stale` label if there is no activity for 8 weeks. Any new activity will auto-remove the label when this job runs. Could be useful as a quick way to identify/filter them.

The alternative would be to rely on sorting issues by "least recently updated" which could be sufficient, but not automatically managed by the job (in terms of adding/removing the `stale` label).

derekcollison (approved):
LGTM

PR #4718, opened 2023-10-28
[FIXED] MQTT sess.clear: bring deleteConsumers back under the sess lock

levb:
A customer mentioned in a Zendesk ticket, "_It is also not possible to clean sessions._" (in their cluster configuration).

One of the things that changed in 2.10 session cleaning was moving the JS calls to delete Subscription's consumers outside the session lock, see 2.9.x implementation [here](https://github.com/nats-io/nats-server/blob/release/v2.9.23/server/mqtt.go#L2338).

Not 100% sure why it would matter with a cluster, but bringing the JS calls back under the lock, to be safe.

### DIFF
--- a/server/mqtt.go
+++ b/server/mqtt.go
@@ -2676,19 +2676,22 @@ func (sess *mqttSession) clear() error {
 	sess.pubRelConsumer = nil
 	sess.seq = 0
 	sess.tmaxack = 0
-	sess.mu.Unlock()
 
 	for _, dur := range durs {
 		if _, err := sess.jsa.deleteConsumer(mqttStreamName, dur); isErrorOtherThan(err, JSConsumerNotFoundErr) {
+			sess.mu.Unlock()
 			return fmt.Errorf("unable to delete consumer %q for session %q: %v", dur, sess.id, err)
 		}
 	}
 	if pubRelDur != "" {
 		_, err := sess.jsa.deleteConsumer(mqttOutStreamName, pubRelDur)
 		if isErrorOtherThan(err, JSConsumerNotFoundErr) {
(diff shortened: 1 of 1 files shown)


kozlovic (approved):
LGTM

server/mqtt.go
```diff
@@ -2676,20 +2676,23 @@ func (sess *mqttSession) clear() error {
 	sess.pubRelConsumer = nil
 	sess.seq = 0
 	sess.tmaxack = 0
-	sess.mu.Unlock()
 
 	for _, dur := range durs {
 		if _, err := sess.jsa.deleteConsumer(mqttStreamName, dur); isErrorOtherThan(err, JSConsumerNotFoundErr) {
+			sess.mu.Unlock()
```
derekcollison, line 2682: This means we are holding the lock across and request/reply boundary, which will be hopping between servers and routes. So do not think we should do that, unless I am mistaken on deleteConsumer. But that is also why we do not hold below for deleteMsg.

derekcollison (changes requested):
I don't think we should do this.

What was it seeming to solve?

server/mqtt.go
levb: @derekcollison Understood, that's why I made the change to begin with, in https://github.com/nats-io/nats-server/pull/4349/files#diff-ef6be3a4d0880e5771af5ea89b210064b3f988fdde90076a188b3b20360d8a13R2664 but it was not really related to QoS2; creeped in.

This PR would revert to how it was in v2.9

Anyway, we can hold off on this PR, see if https://github.com/nats-io/nats-server/pull/4734 addresses session `clear()` concerns.

PR #4720, opened 2023-10-29
chore: Update GitHub actions

reneleonhardt:
* goreleaser/goreleaser-action 3 -> 5
* actions/checkout 3 -> 4

Also bumped default tag for nightly.

Signed-off-by: Rene Leonhardt <65483435+reneleonhardt@users.noreply.github.com>

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,10 @@
+version: 2
+updates:
+  - package-ecosystem: "gomod"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
--- a/.github/workflows/rc_nightly.yaml
+++ b/.github/workflows/rc_nightly.yaml
@@ -15,10 +15,10 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout code
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
         with:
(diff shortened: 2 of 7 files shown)


.github/workflows/rc_nightly.yaml
bruth, line 1: @wallyqs Looks like this workflow was [explicitly disabled](https://github.com/nats-io/nats-server/actions/workflows/rc_nightly.yaml) a month ago, so it appears safe to remove.

.github/dependabot.yml
```diff
@@ -0,0 +1,10 @@
+version: 2
+updates:
+  - package-ecosystem: "gomod"
```
bruth, line 3: The dependabot docs don't give much insight, but I am going to assume this does not attempt to automatically propose new Go versions, only the dependency updates?

bruth (commented):
Nice, we appreciate the contribution @reneleonhardt! A couple questions, but otherwise looks good.

.github/dependabot.yml
reneleonhardt: As far as I know only dependencies and not a major golang upgrade... but even if such a PR would be created you wouldn't have to merge it 😉

bruth: For sure and agreed 😄

.github/workflows/rc_nightly.yaml
bruth: We can remove this file post-PR. Don't want to block this one.

bruth (approved):

PR #4724, opened 2023-10-30
Improve `raft.go` comments

neilalexander:
This PR mostly just tries to comment what is going on inside `raft.go` and does odd little bits of housekeeping/cleanup in places.

### DIFF
--- a/server/raft.go
+++ b/server/raft.go
@@ -142,8 +142,8 @@ type raft struct {
 	qn    int             // Number of nodes needed to establish quorum
 	peers map[string]*lps // Other peers in the Raft group
 
-	removed map[string]struct{}            //
-	acks    map[uint64]map[string]struct{} //
+	removed map[string]struct{}            // Peers that were removed from the group
+	acks    map[uint64]map[string]struct{} // Append entry responses/acks, map of entry index -> peer ID
 	pae     map[uint64]*appendEntry        // Pending append entries
 
 	elect  *time.Timer // Election timer, normally accessed via electTimer
@@ -859,7 +859,7 @@ func (n *raft) PauseApply() error {
 		n.stepdown.push(noLeader)
 	}
 
-	n.debug("Pausing our apply queue")
+	n.debug("Pausing our apply channel")
 	n.paused = true
 	n.hcommit = n.commit
 	// Also prevent us from trying to become a leader while paused and catching up.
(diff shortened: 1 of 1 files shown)


server/raft.go
```diff
@@ -122,108 +122,105 @@ func (state RaftState) String() string {
...
-	// Random generator, used to generate inboxes for instance
-	prand *rand.Rand
+	hh       hash.Hash64  // Highwayhash, used for snapshots
+	snapfile string       // Snapshot filename
+
+	csz   int             // Cluster size
+	qn    int             // Number of nodes needed to establish quorum
+	peers map[string]*lps // Other peers in the Raft group
+
+	removed map[string]struct{}            //
```
derekcollison, line 145: Missing comment?

server/raft.go
```diff
@@ -380,7 +376,6 @@ func (s *Server) startRaftNode(accName string, cfg *RaftConfig, labels pprofLabe
 		track:    cfg.Track,
 		csz:      ps.clusterSize,
 		qn:       ps.clusterSize/2 + 1,
-		hash:     hash,
```
derekcollison, line 383: Why removed? Never accessed?

server/raft.go
```diff
@@ -819,7 +859,7 @@ func (n *raft) PauseApply() error {
 		n.stepdown.push(noLeader)
 	}
 
-	n.debug("Pausing our apply channel")
+	n.debug("Pausing our apply queue")
```
derekcollison, line 862: These are tough, sometimes easy to see a better log line but if its out there folks could depend on it, so I usually shy away from changing.

server/raft.go
```diff
@@ -837,7 +879,7 @@ func (n *raft) ResumeApply() {
 		return
 	}
 
-	n.debug("Resuming our apply channel")
+	n.debug("Resuming our apply queue")
```
derekcollison, line 882: Same.

server/raft.go
neilalexander: Forgot to fill those in, have done so now.

server/raft.go
neilalexander: Yep, it was never used.

server/raft.go
neilalexander: Good point, I've put them back to what they were.

derekcollison (approved):
LGTM

PR #4725, opened 2023-10-30
[FIXED] Remove the state check in the runAs loops except for runAsLeader.

derekcollison:
The runAsFollower would disable observer state working at all.


wallyqs (approved):

neilalexander:
Just for future reference that observer state wasn't actually broken by #4684 after all. We've decided to keep the revert of these two loop conditions as per this PR because they didn't contribute directly to a fix like the `runAsLeader` one did.

PR #4727, opened 2023-10-31
Remove unused `Observer` const, add unit test to check observer applies

neilalexander:
The `Observer` raft state was actually unused and is a red herring. Instead, observer state is tracked using a separate bool.

This PR also adds a unit test that confirms that applies are still happening when followers are observers.

Signed-off-by: Neil Twigg <neil@synadia.com>

### DIFF
--- a/server/raft.go
+++ b/server/raft.go
@@ -103,7 +103,6 @@ const (
 	Follower RaftState = iota
 	Leader
 	Candidate
-	Observer
 	Closed
 )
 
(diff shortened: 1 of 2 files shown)


derekcollison (commented):
Feels like we should do more to formalize vs remove here.

neilalexander:
I do think bringing back and formalising the `Observer` state in the future is a good idea but it is quite a bit riskier to do. For now it would just make the code less misleading.

PR #4731, opened 2023-11-01
[IMPROVED] Throttle writeFullState from separate go routine. 

derekcollison:
Limit to at most once every 10s.
We will schedule timer so that the signal is not lost.
Also after first run through of writeFullState track the avg dmap encoded len for subsequent runs.


neilalexander (approved):
LGTM

PR #4734, opened 2023-11-02
[FIXED] MQTT: rapid load-balanced (re-)CONNECT to cluster causes races

levb:
The tests explain the condition. TL;DR: a rapid, **load-balanced** sequence of connects/disconnects from the same client to a cluster was causing failures.

This PR:
- eliminates the need to load the session message from JS when processing a persist notification from another server. The clientID that was previously obtained from the loaded message is now available in the ACK subject.
- eliminates the separate goroutine/ipQueue for processing session persist messages; now in-line
- ignores deleteMsg errors if caused by outdated cached session data
- fixes incorrect ACK subject in transfering retained messages 2.9->2.10

@derekcollison please note a **highly** variable test run times O(1ms) - O(100ms), including targeting the same server. Not sure what's going on there. (`go test -v ./server --run TestMQTTClusterConnectDisconnect` prints them out):
image
 
Same test failing in `main`:
image

### DIFF
--- a/server/mqtt.go
+++ b/server/mqtt.go
@@ -155,6 +155,7 @@ const (
 	// while "$MQTT.JSA.<node id>.SL.<number>" is for a stream lookup, etc...
 	mqttJSAIdTokenPos     = 3
 	mqttJSATokenPos       = 4
+	mqttJSAClientIDPos    = 5
 	mqttJSAStreamCreate   = "SC"
 	mqttJSAStreamUpdate   = "SU"
 	mqttJSAStreamLookup   = "SL"
@@ -237,10 +238,9 @@ type mqttAccountSessionManager struct {
 	sl         *Sublist                       // sublist allowing to find retained messages for given subscription
 	retmsgs    map[string]*mqttRetainedMsgRef // retained messages
 	jsa        mqttJSA
-	rrmLastSeq uint64           // Restore retained messages expected last sequence
-	rrmDoneCh  chan struct{}    // To notify the caller that all retained messages have been loaded
-	sp         *ipQueue[uint64] // Used for cluster-wide processing of session records being persisted
-	domainTk   string           // Domain (with trailing "."), or possibly empty. This is added to session subject.
+	rrmLastSeq uint64        // Restore retained messages expected last sequence
+	rrmDoneCh  chan struct{} // To notify the caller that all retained messages have been loaded
+	domainTk   string        // Domain (with trailing "."), or possibly empty. This is added to session subject.
 }
 
 type mqttJSA struct {
@@ -1109,7 +1109,6 @@ func (s *Server) mqttCreateAccountSessionManager(acc *Account, quitCh chan struc
 			nuid:   nuid.New(),
 			quitCh: quitCh,
 		},
-		sp: newIPQueue[uint64](s, qname+"sp"),
 	}
 	// TODO record domain name in as here
 
@@ -1170,14 +1169,15 @@ func (s *Server) mqttCreateAccountSessionManager(acc *Account, quitCh chan struc
 	// This is a subscription that will process all JS API replies. We could split to
 	// individual subscriptions if needed, but since there is a bit of common code,
 	// that seemed like a good idea to be all in one place.
-	if err := as.createSubscription(jsa.rplyr+"*.*",
+	if err := as.createSubscription(jsa.rplyr+">",
 		as.processJSAPIReplies, &sid, &subs); err != nil {
 		return nil, err
 	}
 
 	// We will listen for replies to session persist requests so that we can
 	// detect the use of a session with the same client ID anywhere in the cluster.
-	if err := as.createSubscription(mqttJSARepliesPrefix+"*."+mqttJSASessPersist+".*",
+	//   `$MQTT.JSA.{js-id}.SP.{client-id-hash}.{uuid}`
+	if err := as.createSubscription(mqttJSARepliesPrefix+"*."+mqttJSASessPersist+".*.*",
 		as.processSessionPersist, &sid, &subs); err != nil {
 		return nil, err
 	}
@@ -1201,12 +1201,6 @@ func (s *Server) mqttCreateAccountSessionManager(acc *Account, quitCh chan struc
 	s.startGoRoutine(func() {
 		defer s.grWG.Done()
 		as.sendJSAPIrequests(s, c, accName, closeCh)
-	})
-
-	// Start the go routine that will handle network updates regarding sessions
-	s.startGoRoutine(func() {
-		defer s.grWG.Done()
-		as.sessPersistProcessing(closeCh)
 	})
 
 	lookupStream := func(stream, txt string) (*StreamInfo, error) {
@@ -1454,7 +1448,7 @@ func (s *Server) mqttDetermineReplicas() int {
 //////////////////////////////////////////////////////////////////////////////
 
 func (jsa *mqttJSA) newRequest(kind, subject string, hdr int, msg []byte) (interface{}, error) {
-	return jsa.newRequestEx(kind, subject, hdr, msg, mqttJSAPITimeout)
+	return jsa.newRequestEx(kind, subject, _EMPTY_, hdr, msg, mqttJSAPITimeout)
 }
 
 func (jsa *mqttJSA) prefixDomain(subject string) string {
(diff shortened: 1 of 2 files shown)


levb:
Actually, still failed the new test when I run it with `--count 50`, investigating.

levb:
While testing with --count=X I saw that persistent sessions were hitting a similar race condition when the session persist notification was not processed in-time; it was causing invalid seq errors. 

The processing of persist notification was in a separate go-routine fed from ipQueue so it could load session messages for inspection. However, the only data we need from the loaded message was the client ID (hash) which we can add to the JS ACK subject and use directly... This is what the last commit does.

This does not improve the latency of CONNECT/CONNACK itself, but makes the inter-server messaging more robust, allowing the sessions to stay in sync better.

server/mqtt.go
```diff
@@ -2553,7 +2536,8 @@ func (as *mqttAccountSessionManager) transferRetainedToPerKeySubjectStream(log *
 		}
 		// Store the message again, this time with the new per-key subject.
 		subject := mqttRetainedMsgsStreamSubject + rmsg.Subject
-		if _, err := jsa.storeMsgWithKind(mqttJSASessPersist, subject, 0, smsg.Data); err != nil {
+
```
levb, line 2539: This was likely a mistake. I think using `jsa.storeMsgWithKind(mqttJSASessPersist,...` here was causing unnecessary JS ACK processing. @neilalexander Please confirm sice it came from https://github.com/nats-io/nats-server/pull/4199

neilalexander: Not sure I'm following, this is only for the migration from the old subject? If the migration has completed already then I wouldn't expect us to get this far.

levb: Correct, I think this is unrelated to the customer issue; just checking that the intention is to just save the message, not the `SP` ack broadcast. Like you said, it's a one-time migration.

server/mqtt.go
```diff
@@ -1170,14 +1169,15 @@ func (s *Server) mqttCreateAccountSessionManager(acc *Account, quitCh chan struc
...
+	if err := as.createSubscription(jsa.rplyr+">",
 		as.processJSAPIReplies, &sid, &subs); err != nil {
 		return nil, err
 	}
 
 	// We will listen for replies to session persist requests so that we can
 	// detect the use of a session with the same client ID anywhere in the cluster.
-	if err := as.createSubscription(mqttJSARepliesPrefix+"*."+mqttJSASessPersist+".*",
+	//   `$MQTT.JSA.{js-id}.SP.{client-id-hash}.{uuid}`
+	if err := as.createSubscription(mqttJSARepliesPrefix+"*."+mqttJSASessPersist+".*.*",
```
kozlovic, line 1180: So does that mean that server with this fix will not be able to co-operate with other servers (say current v2.10.3)? That is, a current server that would persist a session with a reply subject on `$MQTT.JSA.<serverId>.SP.<nuid>` would not be received by a server with this fix. Maybe it's ok, but I am just raising this to make sure that you thought about it.

server/mqtt.go
```diff
@@ -1454,7 +1448,7 @@ func (s *Server) mqttDetermineReplicas() int {
 //////////////////////////////////////////////////////////////////////////////
 
 func (jsa *mqttJSA) newRequest(kind, subject string, hdr int, msg []byte) (interface{}, error) {
-	return jsa.newRequestEx(kind, subject, hdr, msg, mqttJSAPITimeout)
+	return jsa.newRequestEx(kind, subject, "", hdr, msg, mqttJSAPITimeout)
```
kozlovic, line 1451: nit: Derek likes that we use `_EMPTY_` for that.

server/mqtt.go
```diff
@@ -1467,19 +1461,24 @@ func (jsa *mqttJSA) prefixDomain(subject string) string {
...
+	var sb strings.Builder
 	jsa.mu.Lock()
 	// Either we use nuid.Next() which uses a global lock, or our own nuid object, but
 	// then it needs to be "write" protected. This approach will reduce across account
 	// contention since we won't use the global nuid's lock.
-	var sb strings.Builder
 	sb.WriteString(jsa.rplyr)
 	sb.WriteString(kind)
 	sb.WriteByte(btsep)
+	if cidHash != "" {
```
kozlovic, line 1473: `_EMPTY_`

server/mqtt.go
```diff
@@ -2479,7 +2464,7 @@ func (as *mqttAccountSessionManager) transferUniqueSessStreamsToMuxed(log *Serve
 	}()
 
 	jsa := &as.jsa
-	sni, err := jsa.newRequestEx(mqttJSAStreamNames, JSApiStreams, 0, nil, 5*time.Second)
+	sni, err := jsa.newRequestEx(mqttJSAStreamNames, JSApiStreams, "", 0, nil, 5*time.Second)
```
kozlovic, line 2467: `_EMPTY_`

server/mqtt.go
kozlovic: Good catch. I think that this was an issue when implementing the move of retained messages with code copied from the move of session records.

server/mqtt.go
```diff
@@ -2691,8 +2675,14 @@ func (sess *mqttSession) clear() error {
...
 
 	if seq > 0 {
-		if err := sess.jsa.deleteMsg(mqttSessStreamName, seq, true); err != nil {
-			return fmt.Errorf("unable to delete session %q record at sequence %v", id, seq)
+		err := sess.jsa.deleteMsg(mqttSessStreamName, seq, true)
+		// Ignore the various errors indicating that the message (or sequence)
+		// is already deleted, can happen in a cluster.
+		if isErrorOtherThan(err, JSSequenceNotFoundErrF) {
+			if !IsNatsErr(err, JSStreamMsgDeleteFailedF) ||
+				!strings.HasSuffix(err.Error(), "no message found (10057)") {
```
kozlovic, line 2683: Not sure about that one, it feels a bit brittle. You exclude errors related to sequence not found (`isErrorOtherThan(err, JSSequenceNotFoundErrF)`), so here, I think what you are trying to exclude is the specific error about message being not found. The server would return a `JSStreamMsgDeleteFailedF` error, which is more generic, but you want to ignore only if the "sub-reason" is message not found, correct? If so, you should test only for the `JSNoMessageFoundErr` error, and I believe it could be done in one check:
```
if isErrorOtherThan(err, JSSequenceNotFoundErrF) && isErrorOtherThan(err, JSNoMessageFoundErr) {
    return fmt.Errorf(...
}
```

kozlovic (changes requested):
Question about interoperability, some not and recommendation about error checking.

server/mqtt.go
levb: Unfortunately, the underlying error was an `errors.stringError` with the formatted string. I did not chase it deep enough to see where it got converted, but I agree that ^^ is too fragile. Will try to see if I can fix it.

server/mqtt.go
levb: Right, thanks for raising it. I did consider the compatibility issue and kinda punted on it, because of the "edge condition" nature of the use-case. I considered serializing the clientID into the uuid token using a different separator, but that felt too hacky for a permanent solution to a temporary edge-case. Nothing else that I could think of would make this PR backwards-operable, i.e. broadcasting to the 2.10.(N-1) servers in a way that they'd understand. I could easily add another listening subscription to pick up their messages, but that'd be 1-way only.

All in all, 1/5 leave as is and require that all servers in an MQTT cluster are upgraded/downgraded at approximately the same time. (Note for others, this is not affecting the session store itself, just the ACK change notifications.)

server/mqtt.go
levb: "kinda" fixed now, please LMK if there already is a function to map those errors. @kozlovic please see changes in stream.go

server/stream.go
```diff
@@ -2029,7 +2029,12 @@ func (mset *stream) deleteMsg(seq uint64) (bool, error) {
 	}
 	mset.mu.RUnlock()
 
-	return mset.store.RemoveMsg(seq)
+	y, err := mset.store.RemoveMsg(seq)
```
levb, line 2032: @kozlovic @neilalexander is there an existing utility to map Store -> JS API errors?

server/mqtt.go
levb: Having said that, I think using say, `$MQTT.JSA.{js-id}.SP.{client-id-hash}_{uuid}` would work just fine.

server/mqtt.go
```diff
@@ -2678,11 +2678,8 @@ func (sess *mqttSession) clear() error {
 		err := sess.jsa.deleteMsg(mqttSessStreamName, seq, true)
 		// Ignore the various errors indicating that the message (or sequence)
 		// is already deleted, can happen in a cluster.
-		if isErrorOtherThan(err, JSSequenceNotFoundErrF) {
-			if !IsNatsErr(err, JSStreamMsgDeleteFailedF) ||
-				!strings.HasSuffix(err.Error(), "no message found (10057)") {
-				return fmt.Errorf("unable to delete session %q record at sequence %v: %v", id, seq, err)
-			}
+		if isErrorOtherThan(err, JSSequenceNotFoundErrF) && isErrorOtherThan(err, JSNoMessageFoundErr) {
```
kozlovic, line 2681: Ok, I see what you are saying regarding low level store API not returning the JS error we want. So maybe revert, but I would then change the `!strings.HasSuffix(...` with `!strings.Contains(err.Error(), ErrStoreMsgNotFound.Error())`, so at least if that error message text were to change, this code would still catch this.

server/stream.go
kozlovic: I would revert changes to stream.go. See my other comment.

kozlovic (changes requested):

kozlovic (approved):
LGTM

server/mqtt.go
levb: @kozlovic you agree with ^^?

kozlovic: The advantage is that you could revert some of the create subscription to keep the same number of tokens. But you would need to do more processing to extract the client ID from the last token. Up to you.

PR #4735, opened 2023-11-02
NRG: Drop append entries when upper layer is overloaded

neilalexander:
This adds a mechanism for the Raft layer to detect when the upper layer isn't keeping up with applies. When the overloaded condition is detected, we'll start dropping newly incoming append entries instead so that the apply queue doesn't build up endlessly.

Still need to figure out what threshold makes sense here, or how it makes sense to configure it.

### DIFF
--- a/server/raft.go
+++ b/server/raft.go
@@ -73,6 +73,7 @@ type RaftNode interface {
 	LeadChangeC() <-chan bool
 	QuitC() <-chan struct{}
 	Created() time.Time
+	Overloaded() bool
 	Stop()
 	Delete()
 	Wipe()
@@ -239,14 +240,15 @@ const (
 )
 
 var (
-	minElectionTimeout   = minElectionTimeoutDefault
-	maxElectionTimeout   = maxElectionTimeoutDefault
-	minCampaignTimeout   = minCampaignTimeoutDefault
-	maxCampaignTimeout   = maxCampaignTimeoutDefault
-	hbInterval           = hbIntervalDefault
-	lostQuorumInterval   = lostQuorumIntervalDefault
-	lostQuorumCheck      = lostQuorumCheckIntervalDefault
-	observerModeInterval = observerModeIntervalDefault
+	minElectionTimeout          = minElectionTimeoutDefault
+	maxElectionTimeout          = maxElectionTimeoutDefault
+	minCampaignTimeout          = minCampaignTimeoutDefault
+	maxCampaignTimeout          = maxCampaignTimeoutDefault
+	hbInterval                  = hbIntervalDefault
+	lostQuorumInterval          = lostQuorumIntervalDefault
+	lostQuorumCheck             = lostQuorumCheckIntervalDefault
+	observerModeInterval        = observerModeIntervalDefault
+	overloadThreshold    uint64 = 1024 * 1024 * 8
 )
 
 type RaftConfig struct {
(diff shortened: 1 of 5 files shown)


server/raft.go
```diff
@@ -2836,6 +2856,13 @@ func (n *raft) runAsCandidate() {
 
 // handleAppendEntry handles an append entry from the wire.
 func (n *raft) handleAppendEntry(sub *subscription, c *client, _ *Account, subject, reply string, msg []byte) {
+	// If we are overwhelmed, i.e. the upper layer is not applying entries
+	// fast enough and our apply queue is building up, start to drop new
+	// append entries instead.
+	if n.Overloaded() {
```
derekcollison, line 2862: Do not want to acquire lock on network packets, we worked hard to try to get rid of that as much as possible. So let's consider atomic.Bool here that is set once we determine the state elsewhere.

server/raft.go
```diff
@@ -1322,6 +1326,24 @@ func (n *raft) Healthy() bool {
...
+
+// Pushes to the apply queue and updates the overloaded state. Lock must be held.
+func (n *raft) pushToApply(ce *CommittedEntry) {
+	n.apply.push(ce)
+	n.updateOverloadState()
+}
+
+// Updates the overloaded state. Lock must be held.
+func (n *raft) updateOverloadState() {
+	n.overload.Store(n.apply.len() >= overloadThreshold || n.commit-n.applied >= uint64(overloadThreshold))
```
derekcollison, line 1344: More about memory then number of elements I think, WDYT?

server/raft.go
```diff
@@ -1322,6 +1326,24 @@ func (n *raft) Healthy() bool {
...
+func (n *raft) Overloaded() bool {
+	if n == nil {
+		return false
+	}
+	return n.overload.Load()
+}
+
+// Pushes to the apply queue and updates the overloaded state. Lock must be held.
+func (n *raft) pushToApply(ce *CommittedEntry) {
+	n.apply.push(ce)
```
derekcollison, line 1338: I think this function returns pending, so could short circuit some work here.

server/raft.go
```diff
@@ -2939,6 +2961,13 @@ func (n *raft) runAsCandidate() {
 // handleAppendEntry handles an append entry from the wire. This function
 // is an internal callback from the "asubj" append entry subscription.
 func (n *raft) handleAppendEntry(sub *subscription, c *client, _ *Account, subject, reply string, msg []byte) {
+	// If we are overwhelmed, i.e. the upper layer is not applying entries
+	// fast enough and our apply queue is building up, start to drop new
+	// append entries instead.
+	if n.Overloaded() {
```
derekcollison, line 2967: Just check n.apply.len()? Always present, so do not need to lock the raft group, and ipq underneath has its own which we use here anyway.

neilalexander: `Overloaded()` doesn't take a lock but now just checks the apply queue size.

derekcollison: It takes the ipq lock, so we are now taking it twice in that function I think.

derekcollison (commented):
Does the ipq size tracking effect performance?

neilalexander:
Haven't benched it yet but I will do so.

neilalexander:
I may pick this one back up at some point but closing for now as I'm not 100% convinced how safe it really is.

PR #4742, opened 2023-11-04
[IMPROVED] Memory usage with lots of subjects in filestores.

derekcollison:
Prior to 2.10 we depended on per block subject metadata, the fss files. We used subject interning to cut down on the number of duplicate strings.

With 2.10, we have a builtin version of this with fs.psim that tracks all in use subjects for the filestore. We can just use that, which should be friendlier on memory buildup vs the global subject interning.

### DIFF
--- a/server/filestore.go
+++ b/server/filestore.go
@@ -157,6 +157,7 @@ type psi struct {
 	total uint64
 	fblk  uint32
 	lblk  uint32
+	subj  string
 }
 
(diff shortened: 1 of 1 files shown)


neilalexander (approved):
LGTM, nice approach!

derekcollison:
Let's see how it fares but should prevent some of bloat up we have been seeing on NGS.

PR #4743, opened 2023-11-04
[IMPROVED] Estimation on full state allocations to avoid reallocations.

derekcollison:
When deleted entries went from none to some on small full states we would see reallocations.
Also optionally use 4k stack and avoid any allocation, and always check if deleted > 0 and adml is still 0.


ripienaar (approved):
LGTM, tested on a system that had the problem

PR #4744, opened 2023-11-05
[FIXED] Make access to mb's first and last sequence to consistently use atomics.

derekcollison:
Fixes data races.

### DIFF
--- a/server/filestore.go
+++ b/server/filestore.go
@@ -1093,11 +1093,12 @@ func (fs *fileStore) rebuildStateLocked(ld *LostStreamData) {
 		mb.mu.RLock()
 		fs.state.Msgs += mb.msgs
 		fs.state.Bytes += mb.bytes
-		if fs.state.FirstSeq == 0 || mb.first.seq < fs.state.FirstSeq {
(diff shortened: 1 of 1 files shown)


neilalexander (commented):
In principle LGTM, but wonder if changing `first.seq` and `last.seq` to an `atomic.Uint64` type might be less error-prone than potentially forgetting `atomic.LoadUint64`/`atomic.StoreUint64` somewhere in the future and may also avoid any potential alignment issues?

derekcollison:
Might be, could not remember when they introduced those but looking now is G0 1.19, so we should be good if we want to switch. Eventually I want the top level block selection to not need atomics or locks, but more for 2.11.

What alignment issues to you forsee?

derekcollison:
I think we are ok alignment wise, but understand what you are saying.

I think for now will leave but agree if we keep this direction we should consider. But I think there is a better way for the upper layer indexing and block selection.

neilalexander (approved):
Yep agreed, LGTM!

wallyqs (approved):
LGTM

PR #4752, opened 2023-11-06
[FIXED] MQTT race condition when subscribing to `#` with retained messages

levb:
Attempting to subscribe to `#` was causing deadlocks when retained messages were present.

This PR takes a rather conservative approach, does not make locking any more granular than it was before, but eliminates deadlocks that were occurring between JSAPI and `mqttDeliverMsgCbQoS0`. The delivery callback (more recent) was getting invoked for the JSAPI subject, before passing the request to JS.

@derekcollison we can't easily pre-load the retained messages before `processSub` because we need `sub.shadow` which is set up there.

@neilalexander has a work-in-progress attempt to make the locking more granular for `mqttProcessSub`

@neilalexander @derekcollison: I have not yet looked into batch-fetching the retained message.

### DIFF
--- a/server/client.go
+++ b/server/client.go
@@ -2657,12 +2657,12 @@ func (c *client) parseSub(argo []byte, noForward bool) error {
 }
 
 func (c *client) processSub(subject, queue, bsid []byte, cb msgHandler, noForward bool) (*subscription, error) {
-	return c.processSubEx(subject, queue, bsid, cb, noForward, false, false)
+	return c.processSubEx(subject, queue, bsid, cb, noForward, false, false, nil)
 }
 
-func (c *client) processSubEx(subject, queue, bsid []byte, cb msgHandler, noForward, si, rsi bool) (*subscription, error) {
+func (c *client) processSubEx(subject, queue, bsid []byte, cb msgHandler, noForward, si, rsi bool, mqtt *mqttSub) (*subscription, error) {
 	// Create the subscription
-	sub := &subscription{client: c, subject: subject, queue: queue, sid: bsid, icb: cb, si: si, rsi: rsi}
+	sub := &subscription{client: c, subject: subject, queue: queue, sid: bsid, icb: cb, si: si, rsi: rsi, mqtt: mqtt}
 
 	c.mu.Lock()
 
--- a/server/mqtt.go
+++ b/server/mqtt.go
@@ -331,7 +331,7 @@ type mqttRetainedMsgRef struct {
 }
 
 type mqttSub struct {
-	ready atomic.Value
+	ready atomic.Bool
 
 	qos byte
 	// Pending serialization of retained messages to be sent when subscription is registered
@@ -4184,7 +4184,7 @@ func mqttDeliverMsgCbQoS0(sub *subscription, pc *client, _ *Account, subject, re
 	}
 
 	// The subcription may still be initializing.
-	if ready := sub.mqtt.ready.Load(); ready == nil || !ready.(bool) {
+	if ready := sub.mqtt.ready.Load(); !ready {
 		return
 	}
 
@@ -4259,7 +4259,7 @@ func mqttDeliverMsgCbQoS12(sub *subscription, pc *client, _ *Account, subject, r
 	}
 
 	// The subcription may still be initializing.
-	if ready := sub.mqtt.ready.Load(); ready == nil || !ready.(bool) {
+	if ready := sub.mqtt.ready.Load(); !ready {
 		return
 	}
 
--- a/server/mqtt.go
+++ b/server/mqtt.go
@@ -2187,8 +2187,7 @@ func (sess *mqttSession) setupSub(c *client, subject, sid, jsDur string, qos byt
 	sub.mqtt.qos = qos
 
 	if qos == 2 {
-		err = sess.ensurePubRelConsumerSubscription(c)
-		if err != nil {
+		if err = sess.ensurePubRelConsumerSubscription(c); err != nil {
 			return nil, err
 		}
 	}
--- a/server/mqtt.go
+++ b/server/mqtt.go
@@ -27,6 +27,7 @@ import (
 	"strconv"
 	"strings"
 	"sync"
+	"sync/atomic"
 	"time"
 	"unicode/utf8"
 
@@ -330,6 +331,8 @@ type mqttRetainedMsgRef struct {
 }
 
 type mqttSub struct {
+	ready atomic.Value
+
 	qos byte
 	// Pending serialization of retained messages to be sent when subscription is registered
 	prm *mqttWriter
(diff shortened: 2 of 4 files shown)


server/mqtt.go
```diff
@@ -2161,6 +2164,37 @@ func (as *mqttAccountSessionManager) removeSession(sess *mqttSession, lock bool)
...
+	if err != nil {
+		return nil, err
+	}
+
+	// Update QOS again since if there was an existing subscription, it was
+	// returned unchanged.
+	sub.mqtt.qos = qos
+
+	if qos == 2 {
+		err = sess.ensurePubRelConsumerSubscription(c)
```
derekcollison, line 2190: Can do in one line.

server/mqtt.go
```diff
@@ -2297,6 +2309,12 @@ func (as *mqttAccountSessionManager) processSubs(sess *mqttSession, c *client,
 		subs = append(subs, sub)
 		addJSConsToSess(sid, jscons)
 	}
+
+	// Mark subscriptions as ready so the callbacks deliver messages.
+	for _, sub := range subs {
+		sub.mqtt.ready.Store(true)
```
derekcollison, line 2315: Make bool.

derekcollison (changes requested):

server/client.go
```diff
@@ -2657,12 +2657,12 @@ func (c *client) parseSub(argo []byte, noForward bool) error {
 }
 
 func (c *client) processSub(subject, queue, bsid []byte, cb msgHandler, noForward bool) (*subscription, error) {
-	return c.processSubEx(subject, queue, bsid, cb, noForward, false, false)
+	return c.processSubEx(subject, queue, bsid, cb, noForward, false, false, nil)
 }
 
-func (c *client) processSubEx(subject, queue, bsid []byte, cb msgHandler, noForward, si, rsi bool) (*subscription, error) {
+func (c *client) processSubEx(subject, queue, bsid []byte, cb msgHandler, noForward, si, rsi bool, mqtt *mqttSub) (*subscription, error) {
```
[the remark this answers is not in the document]
levb: We need to pre-initialize `mqtt` there. I can add another cut&paste function, don't really see what else I can do.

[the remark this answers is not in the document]
derekcollison: Were we hanging mqtt context off of subs before just not passing to processSubEx of course?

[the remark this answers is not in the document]
levb: yes, we were setting it up "right after" the call to `c.processSub` but it was racy

server/mqtt.go
```diff
@@ -330,6 +331,8 @@ type mqttRetainedMsgRef struct {
 }
 
 type mqttSub struct {
+	ready atomic.Value
```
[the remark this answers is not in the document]
levb: `cause I was not thinking straight? Fixed.

server/mqtt.go
```diff
@@ -2161,6 +2164,37 @@ func (as *mqttAccountSessionManager) removeSession(sess *mqttSession, lock bool)
...
+}
+
+func (sess *mqttSession) setupSub(c *client, subject, sid, jsDur string, qos byte, isReserved bool, h msgHandler) (*subscription, error) {
+	msub := &mqttSub{
+		qos:      qos,
+		reserved: isReserved,
+		jsDur:    jsDur,
+	}
+	msub.ready.Store(false)
+	sub, err := c.processSubEx([]byte(subject), nil, []byte(sid), h, false, false, false, msub)
```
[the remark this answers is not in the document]
levb: resolving, see the other comment thread

server/client.go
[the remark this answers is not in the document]
derekcollison: ok, we need to be taking things away from default sub structure, but that can be for another time.

server/mqtt.go
```diff
@@ -2199,25 +2232,6 @@ func (as *mqttAccountSessionManager) processSubs(sess *mqttSession, c *client,
...
 	}
 
-	// Helper that sets the sub's mqtt fields and possibly serialize retained messages.
-	// Assumes account manager and session lock held.
-	setupSub := func(sub *subscription, qos byte) {
-		subs := []*subscription{sub}
-		if len(sub.shadow) > 0 {
-			subs = append(subs, sub.shadow...)
-		}
-		for _, sub := range subs {
```
kozlovic, line 2209: I am not too sure of your change. We used to create a `&mqtt{}` per shadow sub, now you rely on processSub() to have the shadow be a copy of the sub, but I am pretty sure that this does not make a deep copy of the mqtt{} struct. Which means that things in mqtt{} will be shared between the original sub and its shadow subscriptions. I honestly don't know if this is an issue or not, but given the `prm` which is a writer for serializing retained messages, I feel like sharing this is wrong.

kozlovic (changes requested):
Not sure about the copy done when adding the shadow sub...

kozlovic:
@levb But in general, is that not a moot point and the deadlock will still occur if an existing subscription is about to receive a message, but the setup of a new subscription triggers the call to getRetainedPublishMsgs which will then try to send a JS API to load the retained message. The first call in JSAPI will still be processing the callback and be waiting for the session lock, while the processing of the sub will try to send a JSAPI request...

server/mqtt.go
levb: Right, I can debate how to fix this, but your other point I think leaves this one moot indeed if the entire session needs to be locked, and I think you are right, it would be.

Going to look some more into doing everything JS in advance of `c.processSub`. Need to get the shadow subjects without actually effecting the sub.

kozlovic: Or maybe for the serialization of retained messages, we then go to more fine grained locking of the session, which may mean simply releasing/reacquiring the lock before loading the message (which I usually don't like to do, but, something to think about).

levb:
@kozlovic Replied in the thread so we can keep the notes there and resolve when done. I think you are correct. I had to move setting the subs to `ready` to after the filters loop because the multi-filter use-case was causing deadlocks, basically just like you're saying.

server/mqtt.go
levb: I am wary of either, TBH, because I am not sure how to interpret `Any existing retained messages matching the Topic Filter MUST be re-sent, but the flow of publications MUST NOT be interrupted [MQTT-3.8.4-3].`

One interpretation would be that we could even deliver them in a separate go-routine if we so desired, as long as new incoming PUBLISH delivered as soon as we send out SUBACK. Another interpretation would be that the _order_ must not be interrupted, so no new messages should be delivered until all retained (that are updated close to real-time) are sent out. The intention of the current code was the latter.

Anyway, looking into doing all necessary JS before `processSub`. Besides retained messages, we also need to create the PubRel consumer/subscription (1-time/sess).

kozlovic: Yes, and thank you for reminded me of another reason why locking was held for that whole duration. Exactly what you said about interpretation of delivery of messages.

levb:
Closing, replaced with https://github.com/nats-io/nats-server/pull/4760

PR #4753, opened 2023-11-06
Adjust MQTT lock granularity

neilalexander:
We were holding locks for probably longer than we needed to.

The unit test is borrowed from #4752.

### DIFF
--- a/server/mqtt.go
+++ b/server/mqtt.go
@@ -2199,11 +2199,13 @@ func (as *mqttAccountSessionManager) processSubs(sess *mqttSession, c *client,
 			subs = append(subs, sub.shadow...)
 		}
 		for _, sub := range subs {
+			sess.mu.Lock()
 			if sub.mqtt == nil {
 				sub.mqtt = &mqttSub{}
 			}
 			sub.mqtt.qos = qos
 			sub.mqtt.reserved = isMQTTReservedSubscription(string(sub.subject))
+			sess.mu.Unlock()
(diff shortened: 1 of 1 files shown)


neilalexander:
@kozlovic Can you see any potential issues with more granular locking like this? It's not really clear to me why we need to hold both the account and session locks in some of these places but I may be missing something.

kozlovic:
@neilalexander @levb Well there was a reason to the madness, but the code has changed quite a bit so not sure if this is ok now or not. I know that if I had the setup of a new sub with both account manager and session, there was a reason. The fact that the sub's mqtt was set outside of processSub() (which I suggest changing at the time - passing a sub object fully initialized - but that was rejected) could cause an early delivery. Attempting to get the lock inside the callback would block until the setup was fully complete, which would ensure that sub.mqtt is properly set when processing the rest of the callback.
The changes that is causing problem is the fact that we now load retained messages when processing a new subscription. I understand that my early approach was wrong (keeping retained messages in memory), but the problem of JS apis and session deadlock would not occur at that time.
So this approach in this PR is - I believe - going to cause problems since there is no guarantee that the sub.mqtt be fully initialized when the callback is invoked, and @levb PR still has issues with deadlocks (if my hunch is correct) in presence of existing subscriptions.

levb:
@neilalexander close in favor of https://github.com/nats-io/nats-server/pull/4760?

PR #4755, opened 2023-11-06
[FIXED] Websocket: partial writes may lead to disconnect

kozlovic:
The application would disconnect due to garbage data.

This issue was introduced when refactoring flushOutbound with introduction of c.out.wnb. We no longer need to handle specifically for websocket (we were incorrectly adding c.out.nb to websocket frames...)


derekcollison (approved):
LGTM

PR #4756, opened 2023-11-07
[IMPROVED] Remove places where we were using time.After could cause gc pressure.

derekcollison:

### DIFF
--- a/server/consumer.go
+++ b/server/consumer.go
@@ -3731,7 +3731,8 @@ func (o *consumer) processInboundAcks(qch chan struct{}) {
 	// How often we will check for ack floor drift.
 	// Spread these out for large numbers on a server restart.
 	delta := time.Duration(rand.Int63n(int64(time.Minute)))
-	var ackFloorCheck = time.Minute + delta
+	ticker := time.NewTicker(time.Minute + delta)
+	defer ticker.Stop()
 
 	for {
 		select {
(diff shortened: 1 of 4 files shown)


server/consumer.go
```diff
@@ -3731,7 +3731,8 @@ func (o *consumer) processInboundAcks(qch chan struct{}) {
 	// How often we will check for ack floor drift.
 	// Spread these out for large numbers on a server restart.
 	delta := time.Duration(rand.Int63n(int64(time.Minute)))
-	var ackFloorCheck = time.Minute + delta
+	ticker := time.NewTicker(time.Minute + delta)
```
wallyqs, line 3734: I think that to replace this timer we have to do it inside of the `for` before `select` like:

```go
	for {
		ticker := time.NewTimer(time.Minute + delta)
		select {
		case <-o.ackMsgs.ch:
			ticker.Stop()
			// ...
		case <-ticker.C:
			o.checkAckFloor()
		case <-qch:
			t.Stop()
			return
		case <-s.quitCh:
			t.Stop()
			return
		}
```

derekcollison: I made it into a Ticker vs one shot timers. Always should be firing and checking so made single Ticker outside for loop.

wallyqs (approved):
LGTM

PR #4757, opened 2023-11-07
[IMPROVED] Better check for standalone mode when determining to send statsz.

derekcollison:
When determining if we should check for interest for sending statsz, take into consideration leafnodes that may extend the system account.


neilalexander (approved):
LGTM

ripienaar (approved):
LGTM, confirmed on the docker compose repro environment that this resolves the issue

PR #4760, opened 2023-11-07
[FIXED] MQTT: deadlock between JS API and mqttDeliverMsgCbQoS0

levb:
Re-ordered the interactions with JetStream outside the locks.

2 specific use-cases covered in the tests
- Attempt to subscribe to `#` was causing deadlocks when retained messages were present.
- Attempt to subscribe to anything with relevant retained messages if a `#` (or any other subscription matching internal subjects, say `+/+/.../+`) was already set up.

### DIFF
--- a/server/mqtt.go
+++ b/server/mqtt.go
@@ -4584,6 +4584,7 @@ func (sess *mqttSession) ensurePubRelConsumerSubscription(c *client) error {
 	pubRelConsumer := sess.pubRelConsumer
 	tmaxack := sess.tmaxack
 	idHash := sess.idHash
+	id := sess.id
 	sess.mu.Unlock()
 
 	// Subscribe before the consumer is created so we don't loose any messages.
@@ -4622,7 +4623,7 @@ func (sess *mqttSession) ensurePubRelConsumerSubscription(c *client) error {
 			ccr.Config.InactiveThreshold = opts.MQTT.ConsumerInactiveThreshold
 		}
 		if _, err := sess.jsa.createConsumer(ccr); err != nil {
-			c.Errorf("Unable to add JetStream consumer for PUBREL for client %q: err=%v", sess.id, err)
+			c.Errorf("Unable to add JetStream consumer for PUBREL for client %q: err=%v", id, err)
 			return err
 		}
 		pubRelConsumer = &ccr.Config
--- a/server/mqtt.go
+++ b/server/mqtt.go
@@ -2165,6 +2165,27 @@ func (as *mqttAccountSessionManager) removeSession(sess *mqttSession, lock bool)
 	}
 }
 
+// Helpers that sets the sub's mqtt fields and possibly serialize
+// (pre-loaded) retained messages.
+// Account manager and session lock held.
+func (sess *mqttSession) processSub(c *client, subject, sid []byte, isReserved bool, qos byte, jsDurName string, h msgHandler) (*subscription, error) {
+	sub, err := c.processSub(subject, nil, sid, h, false)
+
+	if err != nil {
+		// c.processSub already called c.Errorf(), so no need here.
+		return nil, err
+	}
+	for _, ss := range append([]*subscription{sub}, sub.shadow...) {
+		if ss.mqtt == nil {
+			ss.mqtt = &mqttSub{}
+		}
+		ss.mqtt.qos = qos
+		ss.mqtt.reserved = isReserved
+		ss.mqtt.jsDur = jsDurName
+	}
+	return sub, nil
+}
+
 // Process subscriptions for the given session/client.
 //
 // When `fromSubProto` is false, it means that this is invoked from the CONNECT
@@ -2271,39 +2292,6 @@ func (as *mqttAccountSessionManager) processSubs(sess *mqttSession, c *client,
 		sess.cons[sid] = cc
 	}
 
-	// Helpers that sets the sub's mqtt fields and possibly serialize
-	// (pre-loaded) retained messages. Assumes account manager and session lock
-	// held.
-	initSub := func(sub *subscription, qos byte) {
-		if sub.mqtt == nil {
-			sub.mqtt = &mqttSub{}
-		}
-		sub.mqtt.qos = qos
-		sub.mqtt.reserved = isMQTTReservedSubscription(string(sub.subject))
-	}
-
-	lockAndInitializeSub := func(subject, sid []byte, qos byte, h msgHandler) (*subscription, error) {
-		as.mu.Lock()
-		defer as.mu.Unlock()
-		sess.mu.Lock()
-		defer sess.mu.Unlock()
-
-		sub, err := c.processSub(subject, nil, sid, h, false)
-
-		if err != nil {
-			// c.processSub already called c.Errorf(), so no need here.
-			return nil, err
-		}
-		initSub(sub, qos)
-		for _, ss := range sub.shadow {
-			initSub(ss, qos)
-		}
-		if fromSubProto {
-			as.serializeRetainedMsgsForSub(rms, sess, c, sub, trace)
-		}
-		return sub, nil
-	}
-
 	var err error
 	subs := make([]*subscription, 0, len(filters))
 	for _, f := range filters {
@@ -2322,7 +2310,15 @@ func (as *mqttAccountSessionManager) processSubs(sess *mqttSession, c *client,
 
 		// Note that if a subscription already exists on this subject, the
 		// existing sub is returned. Need to update the qos.
-		sub, err := lockAndInitializeSub(bsubject, bsid, f.qos, mqttDeliverMsgCbQoS0)
+		as.mu.Lock()
+		sess.mu.Lock()
+		sub, err := sess.processSub(c, bsubject, bsid,
+			isMQTTReservedSubscription(subject), f.qos, _EMPTY_, mqttDeliverMsgCbQoS0)
+		if err == nil {
+			as.serializeRetainedMsgsForSub(rms, sess, c, sub, trace)
+		}
+		sess.mu.Unlock()
+		as.mu.Unlock()
 		if err != nil {
 			f.qos = mqttSubAckFailure
 			sess.cleanupFailedSub(c, sub, jscons, jssub)
@@ -2347,7 +2343,15 @@ func (as *mqttAccountSessionManager) processSubs(sess *mqttSession, c *client,
 			var fwcsub *subscription
 
 			// See note above about existing subscription.
-			fwcsub, err = lockAndInitializeSub([]byte(fwcsubject), []byte(fwcsid), f.qos, mqttDeliverMsgCbQoS0)
+			as.mu.Lock()
+			sess.mu.Lock()
+			fwcsub, err = sess.processSub(c, []byte(fwcsubject), []byte(fwcsid),
+				isMQTTReservedSubscription(subject), f.qos, _EMPTY_, mqttDeliverMsgCbQoS0)
+			if err == nil {
+				as.serializeRetainedMsgsForSub(rms, sess, c, fwcsub, trace)
+			}
+			sess.mu.Unlock()
+			as.mu.Unlock()
 			if err != nil {
 				// c.processSub already called c.Errorf(), so no need here.
 				f.qos = mqttSubAckFailure
@@ -4438,11 +4442,11 @@ func mqttIsReservedSub(sub *subscription, subject string) bool {
 
 // Check if a sub is a reserved wildcard. E.g. '#', '*', or '*/" prefix.
 func isMQTTReservedSubscription(subject string) bool {
-	if len(subject) == 1 && subject[0] == fwc || subject[0] == pwc {
+	if len(subject) == 1 && (subject[0] == fwc || subject[0] == pwc) {
 		return true
 	}
 	// Match "*.<>"
-	if len(subject) > 1 && subject[0] == pwc && subject[1] == btsep {
+	if len(subject) > 1 && (subject[0] == pwc && subject[1] == btsep) {
 		return true
 	}
 	return false
@@ -4556,9 +4560,6 @@ func (sess *mqttSession) cleanupFailedSub(c *client, sub *subscription, cc *Cons
 
 // Make sure we are set up to deliver PUBREL messages to this QoS2-subscribed
 // session.
-//
-// Locks the session for the entire duration. There is no deadlock potential
-// because we subscribe mqttDeliverPubRelCb to the delivery subject only.
 func (sess *mqttSession) ensurePubRelConsumerSubscription(c *client) error {
 	opts := c.srv.getOpts()
 	ackWait := opts.MQTT.AckWait
@@ -4635,15 +4636,22 @@ func (sess *mqttSession) ensurePubRelConsumerSubscription(c *client) error {
 // When invoked with a QoS of 0, looks for an existing JS durable consumer for
 // the given sid and if one is found, delete the JS durable consumer and unsub
 // the NATS subscription on the delivery subject.
+//
 // With a QoS > 0, creates or update the existing JS durable consumer along with
 // its NATS subscription on a delivery subject.
 //
-// Lock not held on entry, but session is in the locked map.
+// Session lock is acquired and released as needed. Session is in the locked
+// map.
 func (sess *mqttSession) processJSConsumer(c *client, subject, sid string,
 	qos byte, fromSubProto bool) (*ConsumerConfig, *subscription, error) {
 
-	// Check if we are already a JS consumer for this SID.
+	sess.mu.Lock()
 	cc, exists := sess.cons[sid]
+	tmaxack := sess.tmaxack
+	idHash := sess.idHash
+	sess.mu.Unlock()
+
+	// Check if we are already a JS consumer for this SID.
 	if exists {
 		// If current QoS is 0, it means that we need to delete the existing
 		// one (that was QoS > 0)
@@ -4652,7 +4660,11 @@ func (sess *mqttSession) processJSConsumer(c *client, subject, sid string,
 			// the form: mqttSubPrefix + <nuid>. It is also used as the sid
 			// for the NATS subscription, so use that for the lookup.
 			sub := c.subs[cc.DeliverSubject]
+
+			sess.mu.Lock()
 			delete(sess.cons, sid)
+			sess.mu.Unlock()
+
 			sess.deleteConsumer(cc)
 			if sub != nil {
 				c.processUnsub(sub.sid)
@@ -4688,12 +4700,12 @@ func (sess *mqttSession) processJSConsumer(c *client, subject, sid string,
 		}
 
 		// Check that the limit of subs' maxAckPending are not going over the limit
-		if after := sess.tmaxack + maxAckPending; after > mqttMaxAckTotalLimit {
+		if after := tmaxack + maxAckPending; after > mqttMaxAckTotalLimit {
 			return nil, nil, fmt.Errorf("max_ack_pending for all consumers would be %v which exceeds the limit of %v",
 				after, mqttMaxAckTotalLimit)
 		}
 
-		durName := sess.idHash + "_" + nuid.Next()
+		durName := idHash + "_" + nuid.Next()
 		ccr := &CreateConsumerRequest{
 			Stream: mqttStreamName,
 			Config: ConsumerConfig{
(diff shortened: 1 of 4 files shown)


server/mqtt.go
```diff
@@ -4160,7 +4160,25 @@ func mqttSubscribeTrace(pi uint16, filters []*mqttFilter) string {
 // message and this is the callback for a QoS1+ subscription because in that
 // case, it will be handled by the other callback. This avoid getting duplicate
 // deliveries.
+func mqttDoNotDeliverSubject(subject string) bool {
```
kozlovic, line 4163: Not sure if this function should be under this comment (that belongs to the QoS0 callback). But more importantly, that may not work for JS Api prefix that use domain, so would it be simpler to just reject anything that starts with `$JS.` and `$MQTT.`?
We could also maybe optimize with something like (maybe we need to check that subject len > 0, but I don't think we would be invoked in a callback with an empty subject)
```
if subject[0] != '$' {
    return false
}
if strings.HasPrefix(subject, "$JS.") || strings.HasPrefix(subject, mqttJSARepliesPrefix) {
    return true
}
return false
```

kozlovic (changes requested):
Of course, this is not ideal, we need to rethink all that, but as a stop gap, it could be ok.

server/mqtt.go
levb: RE: Comment - oops

RE: Optimize - sure, SG.

levb: image

server/mqtt.go
```diff
@@ -4142,6 +4142,41 @@ func mqttSubscribeTrace(pi uint16, filters []*mqttFilter) string {
...
+	if subject[3] != '.' && subject[3] != 'T' {
+		return false
+	}
+	if l < 6 {
+		return false
+	}
+	if subject[4] != 'T' {
+		return false
+	}
+	if subject[5] != '.' {
```
kozlovic, line 4174: Not just `$MQTT.` because otherwise I think you will fail some QoS2 tests. It has to be the `mqttJSARepliesPrefix`.

kozlovic (changes requested):

server/mqtt.go
levb: Right.

server/mqtt_test.go
```diff
@@ -7633,3 +7735,34 @@ func BenchmarkMQTT_QoS1_PubSub2_256b_Payload(b *testing.B) {
...
 }
+
+func BenchmarkMQTTDoNotDeliverSubject(b *testing.B) {
+	buf := make([]byte, 32)
+	jsSubjects := make([]string, 100)
+	mqttSubjects := make([]string, 100)
+	noSubjects := make([]string, 1000)
+
+	for i := 0; i < 100; i++ {
+		_, _ = rand.Read(buf)
```
kozlovic, line 7746: Seems like you can't use `rand.` since Go 1.20 ;-)

server/mqtt.go
levb: (oops, left in `-race` there, but the relative perf. is similar, but MUCH faster of course)

derekcollison (commented):
Does this PR not change holding the session lock while calling processSubEx?

server/mqtt.go
levb: It wasn't even right to begin with, added a test.

levb:
@derekcollison it does not. It merely avoids the deadlocks on known use cases.

kozlovic:
@derekcollison @levb This is worse than I thought. We cannot do any JSA requests with the session lock, and it will be very important if a message comes from a route and triggers one of the callback (qos0 or 12) and tries to grab the session lock while a subscription is created and we try to create a consumer for qos2 (currently under the session lock) or load the retained messages (also under session lock, but more importantly, if messages come from the route - say leader is on different server) then we won't get the message since it will be backed-up behind the route message that triggered the callback.

derekcollison:
We should not be doing any IO while holding locks unless the use case is extremely well understood.

server/mqtt_test.go
```diff
@@ -7633,3 +7737,34 @@ func BenchmarkMQTT_QoS1_PubSub2_256b_Payload(b *testing.B) {
...
```
kozlovic, line 7748: I don't think using rand here is a good thing because it could generate invalid subject, no? what is there something like "ancd..dwre"? So I would revert the mathrand change and go with something actually less random here (which also will add reproducibility in the bench)

kozlovic (changes requested):

server/mqtt_test.go
levb: There's a 1000+ of them there, so it shouldn't be much variance? And we don't look beyond 8 bytes anyway? But yes, I see your point. I can just do a handful of static strings, sure.

neilalexander: Alternatively just hex-encode or something to avoid invalid characters.

levb:
@kozlovic @derekcollison removing the IO from inside the locks,  still pretty raw but passed the tests.

server/mqtt.go
```diff
@@ -2037,27 +2037,28 @@ func (as *mqttAccountSessionManager) handleRetainedMsg(key string, rm *mqttRetai
...
-			// Clear the floor
-			erm.floor = 0
-			// If sub is nil, it means that it was removed from sublist following a
-			// network delete. So need to add it now.
-			if erm.sub == nil {
-				erm.sub = &subscription{subject: []byte(key)}
-				as.sl.Insert(erm.sub)
-			}
+	}
+
```
levb, line 2041: I think I did the right thing here, @kozlovic?

server/mqtt.go
```diff
@@ -4476,33 +4570,43 @@ func (sess *mqttSession) ensurePubRelConsumerSubscription(c *client) error {
...
+			c.Errorf("Unable to create subscription for JetStream consumer on %q: %v", pubRelDeliverySubject, err)
 			return err
 		}
-		sess.pubRelSubscribed = true
+		pubRelSubscribed = true
 	}
 
 	// Create the consumer if needed.
-	if sess.pubRelConsumer == nil {
+	if pubRelConsumer == nil {
```
levb, line 4595: Need to handle the error that the consumer may already exost

server/mqtt.go
```diff
@@ -2202,97 +2271,99 @@ func (as *mqttAccountSessionManager) processSubs(sess *mqttSession, c *client,
...
+		sess.mu.Lock()
+		defer sess.mu.Unlock()
+
+		sub, err := c.processSub(subject, nil, sid, h, false)
+
+		if err != nil {
+			// c.processSub already called c.Errorf(), so no need here.
+			return nil, err
 		}
+		initSub(sub, qos)
```
kozlovic, line 2297: We could "inline" the content of initSub here by simply doing something like:
```
subs := []*subscription{sub}
subs = append(subs, sub.shadow...)
for _, sub := range subs {
    if sub.mqtt == nil {
        sub.mqtt = &mqttSub{}
    }
    sub.mqtt.qos = qos
    sub.mqtt.reserved = isMQTTReservedSubscription(string(sub.subject))
}
```

server/mqtt.go
```diff
@@ -2367,23 +2442,35 @@ func (as *mqttAccountSessionManager) serializeRetainedMsgsForSub(sess *mqttSessi
...
+		loadSubject := mqttRetainedMsgsStreamSubject + subject
+		jsm, err := as.jsa.loadLastMsgFor(mqttRetainedMsgsStreamName, loadSubject)
 		if err != nil || jsm == nil {
+			log.Warnf("failed to load retained message for subject %q: %v", loadSubject, err)
 			continue
 		}
 		var rm mqttRetainedMsg
 		if err := json.Unmarshal(jsm.Data, &rm); err != nil {
-			continue
+			log.Warnf("failed to decode retained message for subject %q: %v", loadSubject, err)
```
kozlovic, line 2471: Still should have the `continue` after the warn, otherwise you are going to store an empty struct.

server/mqtt.go
```diff
@@ -4231,6 +4318,13 @@ func mqttDeliverMsgCbQoS0(sub *subscription, pc *client, _ *Account, subject, re
 // the message contains a NATS/MQTT header that indicates that this is a
 // published QoS1+ message.
 func mqttDeliverMsgCbQoS12(sub *subscription, pc *client, _ *Account, subject, reply string, rmsg []byte) {
+	// This is for consistency with mqttDeliverMsgCbQoS0. This callback does not
```
kozlovic, line 4321: Remove.

server/mqtt.go
```diff
@@ -4463,8 +4557,8 @@ func (sess *mqttSession) cleanupFailedSub(c *client, sub *subscription, cc *Cons
 // Make sure we are set up to deliver PUBREL messages to this QoS2-subscribed
 // session.
 //
-// Session lock held on entry. Need to make sure no other subscribe packet races
-// to do the same.
+// Locks the session for the entire duration. There is no deadlock potential
```
kozlovic, line 4560: The comment here does not match the code. You are not locking for the entire duration...

server/mqtt.go
```diff
@@ -4476,33 +4570,43 @@ func (sess *mqttSession) ensurePubRelConsumerSubscription(c *client) error {
...
+	pubRelConsumer := sess.pubRelConsumer
+	tmaxack := sess.tmaxack
+	idHash := sess.idHash
+	sess.mu.Unlock()
+
 	// Subscribe before the consumer is created so we don't loose any messages.
-	if !sess.pubRelSubscribed {
-		_, err := c.processSub(sess.pubRelDeliverySubjectB, nil, sess.pubRelDeliverySubjectB,
+	if !pubRelSubscribed {
+		_, err := c.processSub(pubRelDeliverySubjectB, nil, pubRelDeliverySubjectB,
```
kozlovic, line 4585: You are not calling this under the session lock. I know it is important for the other subs, so I assume that you understand that it is not needed for `mqttDeliverPubRelCb`.

server/mqtt.go
```diff
@@ -4515,10 +4619,16 @@ func (sess *mqttSession) ensurePubRelConsumerSubscription(c *client) error {
 			c.Errorf("Unable to add JetStream consumer for PUBREL for client %q: err=%v", sess.id, err)
 			return err
 		}
-		sess.pubRelConsumer = &ccr.Config
-		sess.tmaxack += maxAckPending
+		pubRelConsumer = &ccr.Config
+		tmaxack += maxAckPending
```
kozlovic, line 4623: We have the same "issue" in `processJSConsumer()` where in the `else` of the `if exists {` statement, we access sess.tmaxack, idHash without locking and bump tmaxack at the end also without locking (lines 4691, 4696 and 4718). Would you mind also fixing that?

server/mqtt.go
levb: Right. This is neither an "mqtt" sub (doesn't need to be initialized as such), nor are there sync issues with other messages.

The main reason I took out the overarching session lock was because of creating the consumer in the step later. That one MUST be out of a lock since there may be a sub listening on `#` and such.

kozlovic: Correct. Call to processSub() could be again, if there is no need (synchronization with callback, etc..), then please don't. Hence my comment about the comment at the top of the function stating that we lock for the entire duration that seem wrong...

server/mqtt.go
levb: Please review https://github.com/nats-io/nats-server/pull/4760/commits/8d6a98f4ffb7bfad8dad4243c449aa6e438db0e1

server/mqtt.go
levb: Right, removed. The code is right.

server/mqtt.go
levb: Sorry, took the comment out, it was wrong/leftover from the past.

server/mqtt.go
```diff
@@ -2202,97 +2271,99 @@ func (as *mqttAccountSessionManager) processSubs(sess *mqttSession, c *client,
...
-			setupSub(sub, f.qos)
-		}
-		if f.qos == 2 {
-			err = sess.ensurePubRelConsumerSubscription(c)
+		// Note that if a subscription already exists on this subject, the
+		// existing sub is returned. Need to update the qos.
+		sub, err := lockAndInitializeSub(bsubject, bsid, f.qos, mqttDeliverMsgCbQoS0)
+		if err != nil {
+			f.qos = mqttSubAckFailure
+			sess.cleanupFailedSub(c, sub, jscons, jssub)
```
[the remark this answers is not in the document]
levb: 3/5 I think it's more consistent and more intention-revealing this way.

server/mqtt.go
levb: Done in https://github.com/nats-io/nats-server/pull/4760/commits/8d6a98f4ffb7bfad8dad4243c449aa6e438db0e1

server/mqtt.go
```diff
@@ -2322,7 +2310,15 @@ func (as *mqttAccountSessionManager) processSubs(sess *mqttSession, c *client,
 
 		// Note that if a subscription already exists on this subject, the
 		// existing sub is returned. Need to update the qos.
-		sub, err := lockAndInitializeSub(bsubject, bsid, f.qos, mqttDeliverMsgCbQoS0)
+		as.mu.Lock()
+		sess.mu.Lock()
+		sub, err := sess.processSub(c, bsubject, bsid,
+			isMQTTReservedSubscription(subject), f.qos, _EMPTY_, mqttDeliverMsgCbQoS0)
+		if err == nil {
+			as.serializeRetainedMsgsForSub(rms, sess, c, sub, trace)
```
kozlovic, line 2321: I think this is wrong, meaning that we used to do that for each sub (including shadow sub), here we are doing only for the main sub. Moreover, this needs to be done only when called with fromSubProto==true.

server/mqtt.go
```diff
@@ -2347,7 +2343,15 @@ func (as *mqttAccountSessionManager) processSubs(sess *mqttSession, c *client,
 			var fwcsub *subscription
 
 			// See note above about existing subscription.
-			fwcsub, err = lockAndInitializeSub([]byte(fwcsubject), []byte(fwcsid), f.qos, mqttDeliverMsgCbQoS0)
+			as.mu.Lock()
+			sess.mu.Lock()
+			fwcsub, err = sess.processSub(c, []byte(fwcsubject), []byte(fwcsid),
+				isMQTTReservedSubscription(subject), f.qos, _EMPTY_, mqttDeliverMsgCbQoS0)
+			if err == nil {
+				as.serializeRetainedMsgsForSub(rms, sess, c, fwcsub, trace)
```
kozlovic, line 2354: Same comment than above.

server/mqtt.go
```diff
@@ -4711,29 +4723,28 @@ func (sess *mqttSession) processJSConsumer(c *client, subject, sid string,
...
+
+		tmaxack += maxAckPending
 	}
+
 	// This is an internal subscription on subject like "$MQTT.sub.<nuid>" that is setup
 	// for the JS durable's deliver subject.
 	sess.mu.Lock()
-	sub, err := c.processSub([]byte(inbox), nil, []byte(inbox), mqttDeliverMsgCbQoS12, false)
+	sess.tmaxack = tmaxack
+	sub, err := sess.processSub(c, []byte(inbox), []byte(inbox),
```
kozlovic, line 4736: I have a problem with that in that the original code dealt only with `sub` where calling this now makes it process the shadow subs. Unless we understand that this was a bug, for the sake of minimizing behavior changes, I would not do that here.

kozlovic (changes requested):

server/mqtt.go
levb: ok

server/mqtt.go
levb: Resolving this comment since already have new threads

server/mqtt.go
levb: See https://github.com/nats-io/nats-server/pull/4760/commits/a5e47c99aaa7ec2b01a5e7bb4fe0daa3b765a422

server/mqtt.go
```diff
@@ -2181,14 +2182,82 @@ func (as *mqttAccountSessionManager) removeSession(sess *mqttSession, lock bool)
 func (as *mqttAccountSessionManager) processSubs(sess *mqttSession, c *client,
 	filters []*mqttFilter, fromSubProto, trace bool) ([]*subscription, error) {
 
-	// Helpers to lock/unlock both account manager and session.
-	asAndSessLock := func() {
-		as.mu.Lock()
-		sess.mu.Lock()
+	// Helper to determine if we need to create a separate top-level
+	// subscription for a wildcard.
+	fwc := func(subject string) (bool, string, string) {
```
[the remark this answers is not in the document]
levb: I ❤️ keeping them inside the func unless they are re-usable or testable; 1/5 it's currently intention-revealing. If anything, I'd be inclined to do the opposite and make more "value" helpers. 2/5 they should be moved to the top of their parent functions and be parameterized for volatile data. But reusing `c`, `as`, `sess`, etc. makes it typographically nicer IMO.

server/mqtt.go
levb: @kozlovic I do not seem to be getting no error when creating a consumer twice with identical configs. `JSConsumerAlreadyExists` is only returned if the config differs. Does it mean that only one gets created for the unique config, and I don't need to worry about this issue?

kozlovic: It should be fine. Moreover, you are already getting the sess.pubRelConsumer when restoring a session. If the JS consumer was already created, you won't try to create again (you simply recreate the NATS subscription that receives messages from the JS consumer). 

The only think that I am noticing now, is that technically maybe we should do the tmaxack+maxAckPending check in the `if !pubRelSubscribed {` check because although we may not need to recreate the JS consumer, but we do the same in processJSConsumer(), so I think it is fine living like this.

server/mqtt.go
```diff
@@ -4515,25 +4625,38 @@ func (sess *mqttSession) ensurePubRelConsumerSubscription(c *client) error {
 			c.Errorf("Unable to add JetStream consumer for PUBREL for client %q: err=%v", sess.id, err)
```
kozlovic, line 4625: Maybe capture sess.id under lock at beginning of function and use sessId here?

server/mqtt.go
```diff
@@ -2164,6 +2164,30 @@ func (as *mqttAccountSessionManager) removeSession(sess *mqttSession, lock bool)
 	}
 }
 
+// Helpers that sets the sub's mqtt fields and possibly serialize
+// (pre-loaded) retained messages.
+// Account manager and session lock held.
```
kozlovic, line 2169: Not true for "account manager" in all cases (when called in processJSConsumer, only session). So could be changed to "Session lock held. Account manager lock possibly held".

kozlovic (changes requested):
Small few changes still needed.

server/mqtt.go
levb: Account lock does not matter to this function; changed the comment.

kozlovic (approved):
LGTM

derekcollison (changes requested):
I think we want to check len of subject in the fwc helper function. Otherwise LGTM.

server/mqtt.go
```diff
@@ -4476,33 +4576,44 @@ func (sess *mqttSession) ensurePubRelConsumerSubscription(c *client) error {
 		maxAckPending = mqttDefaultMaxAckPending
 	}
 
+	sess.mu.Lock()
```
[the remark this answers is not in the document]
levb: It's a `sync.Mutex` historically; will leave for another day.

server/mqtt.go
```diff
@@ -2181,14 +2205,85 @@ func (as *mqttAccountSessionManager) removeSession(sess *mqttSession, lock bool)
...
-		as.mu.Lock()
-		sess.mu.Lock()
+	// Helper to determine if we need to create a separate top-level
+	// subscription for a wildcard.
+	fwc := func(subject string) (bool, string, string) {
+		if !mqttNeedSubForLevelUp(subject) {
+			return false, _EMPTY_, _EMPTY_
+		}
+		// Say subject is "foo.>", remove the ".>" so that it becomes "foo"
+		fwcsubject := subject[:len(subject)-2]
```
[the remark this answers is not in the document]
levb: Yes, see https://github.com/nats-io/nats-server/blob/78d90274117d86951a6daaa309188a855cec4dad/server/mqtt.go#L4922

server/mqtt.go
```diff
@@ -4605,25 +4733,23 @@ func (sess *mqttSession) processJSConsumer(c *client, subject, sid string,
 			return nil, nil, err
 		}
 		cc = &ccr.Config
-		sess.tmaxack += maxAckPending
+
```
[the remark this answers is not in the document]

derekcollison (approved):
LGTM

PR #4764, opened 2023-11-07
[IMPROVED] Resiliency when doing lots of conditional updates to a KV and restarting servers.

derekcollison:
We would attempt to sync with stream state and reset our CLFS tracking which would cause issues, so we do not reset CLFS or send snapshots upon being elected leader.

There is still a bug that I can see that I am chasing, but this solves many of the issues we would see in other tests.


wallyqs (approved):
LGTM

PR #4770, opened 2023-11-08
Fix subpath used for ws leaf remote connect url

### LINKED ISSUE #4765: Leafnode websocket connections: subpath in wrong order
### Observed behavior

When connecting leafnodes by websockets via a reverse proxy (e.g. in Kubernetes), the subpath for websocket is attached after the static `/leafnode` path instead of before it.

Currently, if in a leafnode configuration the following config is given:
```
leafnodes {
    remotes = [
        {
          url: "wss://user:pass@server.host.org:443/nats-leafnode/",
          account: "ACC"
        },
    ]
}
```

The websocket tries to connect to the following URL:
`server.host.org:443/leafnodenats-leafnode/` instead of `server.host.org:443/nats-leafnode/leafnode`

### Expected behavior

The websocket should connect to
`server.host.org:443/nats-leafnode/leafnode` instead of `server.host.org:443/leafnodenats-leafnode/`

### Server and client version

Both at 2.10.4

### Host environment

_No response_

### Steps to reproduce

Deploy a NATS server after an NGINX with websocket support, which rewrites the websocket connection URL from somehost.org/nats/leafnode to somehost.org/leafnode.

Have a NATS server with leafnode support running behind the proxy and a NATS server acting as a leafnode connecting through the reverse proxy with following config:
```
leafnodes {
    remotes = [
        {
          url: "ws://user:pass@server.host.org:443/nats/",
          account: "ACC"
        },
    ]
}
```

# Additional Info

The following line probably causes this issue:
https://github.com/nats-io/nats-server/blob/9c2e109f9f770a2c74e33d2c2a675e120b0065c4/server/leafnode.go#L2792
and should probably be changed to something like:
```
path = curPath + path
```

wallyqs:
This fixes the concatenated path used by remote leafnodes.

Fixes #4765

### DIFF
--- a/server/leafnode.go
+++ b/server/leafnode.go
@@ -25,6 +25,7 @@ import (
 	"net/http"
 	"net/url"
 	"os"
+	"path"
 	"reflect"
 	"regexp"
 	"runtime"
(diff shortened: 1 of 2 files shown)


derekcollison (approved):
LGTM - Let's make sure tests are green.

derekcollison:
Need to fix..

image

PR #4773, opened 2023-11-08
[FIXED] Datarace and panic when accessing psim from syncBlocks.

derekcollison:
When compacting in the syncBlocks make sure to grab fs lock in case we need to load msgs and reference psim.


levb (approved):
LGTM

wallyqs (approved):
LGTM

PR #4774, opened 2023-11-08
Check for /leafnode suffix path on leaf ws connect

wallyqs:
This changes the`/leafnode` and `/mqtt` checks on a ws connection to detect that it is a leafnode connection over the websocket port.

This way given a remote like this:

```hcl
remotes = [
     {
       url: "ws://user:pass@127.0.0.1:8080/nats-leafnode",
       account: "ACC"
      }
 ]
```

Then the nginx LB configuration would match the same sub path as the configured remote leaf URL:

```hcl
upstream websocket {
        server 127.0.0.1:8888;
}
location /nats-leafnode/leafnode {
  proxy_pass http://websocket;
}
```


derekcollison (approved):
LGTM

PR #4776, opened 2023-11-09
Make sure compact bool localized.

derekcollison:
Once set was set for all blks, so this localizes per blk.


wallyqs (approved):
LGTM

PR #4777, opened 2023-11-09
[IMPROVED] General stability and consistency improvements for clustered streams with failure offsets during server restarts.

derekcollison:
When a message is rejected due to constraints, like CAS operations on KeyValue, we track offsets between the NRG (Raft) layer and the stream. During heavy usage and many msgs being rejected due to constraint violations and constantly restarting servers we saw some inconsistencies.


neilalexander (approved):
LGTM

PR #4779, opened 2023-11-09
Fix panic in `fileStore.Stop()`

neilalexander:
The lock was briefly released to run `writeFullState`, but `fs.closed` wasn't set by that point, so another `Stop()` call could run and panic on closing the `qch` which was already closed.

### DIFF
--- a/server/filestore.go
+++ b/server/filestore.go
@@ -7212,6 +7212,10 @@ func (fs *fileStore) Stop() error {
 		return ErrStoreClosed
 	}
 
+	// Mark as closed. Do before releasing the lock to writeFullState
+	// so we don't end up with this function running more than once.
+	fs.closed = true
+
 	fs.checkAndFlushAllBlocks()
 	fs.closeAllMsgBlocks(false)
 
@@ -7229,8 +7233,7 @@ func (fs *fileStore) Stop() error {
 	fs.writeFullState()
 	fs.mu.Lock()
 
-	// Mark as closed.
-	fs.closed = true
+	// Needs to be cleared after writeFullState has completed.
 	fs.lmb = nil
 
 	// We should update the upper usage layer on a stop.
--- a/server/filestore.go
+++ b/server/filestore.go
@@ -187,6 +187,7 @@ type fileStore struct {
 	cfs         []ConsumerStore
 	sips        int
 	dirty       int
+	closing     bool
 	closed      bool
 	fip         bool
 	receivedAny bool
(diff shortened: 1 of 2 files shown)


server/filestore.go
```diff
@@ -7212,6 +7212,10 @@ func (fs *fileStore) Stop() error {
 		return ErrStoreClosed
 	}
 
+	// Mark as closed. Do before releasing the lock to writeFullState
+	// so we don't end up with this function running more than once.
+	fs.closed = true
```
derekcollison, line 7217: This makes writeFullState a no-op which is not what we want.

derekcollison (commented):
We should have a test that shows the index.db is written properly. Since if it is not could be a big deal.

server/filestore_test.go
```diff
@@ -6512,6 +6512,41 @@ func TestFileStoreLargeFullStateMetaCleanup(t *testing.T) {
...
+	fs, err := newFileStore(
+		FileStoreConfig{StoreDir: sd},
+		StreamConfig{Name: "zzz", Subjects: []string{">"}, Storage: FileStorage})
+	require_NoError(t, err)
+
+	subj, _ := "foo.bar.baz", bytes.Repeat([]byte("ABC"), 33) // ~100bytes
+	for i := 0; i < 1000; i++ {
+		fs.StoreMsg(subj, nil, nil)
+	}
+
```
derekcollison, line 6526: Maybe when you set dirty to true, also remove the existing one. Then you would just need to check for existence.

server/filestore_test.go
```diff
@@ -6512,6 +6512,41 @@ func TestFileStoreLargeFullStateMetaCleanup(t *testing.T) {
...
+	fs.mu.Lock()
+	fs.dirty = 1
+	fs.mu.Unlock()
+	stoptime := time.Now()
+	fs.Stop()
+	<-fs.fsld
```
derekcollison, line 6532: If doing checkFor below not sure this is needed.

server/filestore_test.go
```diff
@@ -6512,6 +6512,40 @@ func TestFileStoreLargeFullStateMetaCleanup(t *testing.T) {
...
+	fs.dirty = 1
+	if err := os.Remove(idxFile); err != nil && !errors.Is(err, os.ErrNotExist) {
+		t.Fatal(err)
+	}
+	fs.mu.Unlock()
+
+	fs.Stop()
+
+	checkFor(t, time.Second, 50*time.Millisecond, func() error {
+		if !fs.closed {
```
derekcollison, line 6539: Need to lock.

derekcollison (changes requested):
We are close. One minor nit.

server/filestore_test.go
neilalexander: Actually I think we probably don't even need to check `fs.closed` at all, as `Stop` blocks until it is set. Will remove it.

derekcollison (approved):
LGTM

PR #4780, opened 2023-11-09
[IMPROVED] When we are linear scanning for first message, check if start < mb.first.seq.

derekcollison:
In cases where msg blocks have msgs but many have been deleted from the front, this speeds that up significantly.


neilalexander (approved):
LGTM, nice fix

PR #4782, opened 2023-11-09
[IMPROVED] When removing very large streams, do fs cleanup in separate go routine.

derekcollison:
This matches what fs.Delete() (and fs.Purge) were already doing.


neilalexander (approved):
LGTM

PR #4783, opened 2023-11-09
[IMPROVED] Move removal of fs.Delete files to a go routine

derekcollison:
Will will remove by doing meta inline in case there are perm errors.


wallyqs (approved):
LGTM

PR #4792, opened 2023-11-13
[FIXED]  #4789 to allow a restart of a shutdown server instance.

### LINKED ISSUE #4789: Panic when calling server.Start() after shutting down the server.
### Observed behavior

Panics when running calling server.Start() again after a successful shutdown.

### Expected behavior

Shouldn't panic when running calling server.Start() again after a successful shutdown.

### Server and client version

v2.10.5

### Host environment

Mac

### Steps to reproduce

```go
import natsserver "github.com/nats-io/nats-server/v2/test"

func TestPaics(t *testing.T) {
	s := natsserver.RunServer(&natsserver.DefaultTestOptions)

	s.Shutdown()

	assert.PanicsWithError(t, "close of closed channel", func() {
		s.Start()
	})
}
```

derekcollison:
Resolves: #4789 

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -2083,6 +2083,11 @@ func (s *Server) Start() {
 
 	defer s.Noticef("Server is ready")
 
+	// Check is this may be a restart.
+	if s.startupComplete == nil {
+		s.startupComplete = make(chan struct{})
+	}
+
 	// Check for insecure configurations.
 	s.checkAuthforWarnings()
 
@@ -2340,7 +2345,10 @@ func (s *Server) Start() {
 	}
 
 	// We've finished starting up.
-	close(s.startupComplete)
+	if s.startupComplete != nil {
+		close(s.startupComplete)
+		s.startupComplete = nil
+	}
 
 	// Wait for clients.
 	if !opts.DontListen {
--- a/server/server_test.go
+++ b/server/server_test.go
@@ -2107,3 +2107,14 @@ func TestServerAuthBlockAndSysAccounts(t *testing.T) {
 	_, err = nats.Connect(s.ClientURL())
 	require_Error(t, err, nats.ErrAuthorization, errors.New("nats: Authorization Violation"))
 }
+
+// https://github.com/nats-io/nats-server/issues/4789
+func TestServerRestartAfterShutdown(t *testing.T) {
+	opts := DefaultTestOptions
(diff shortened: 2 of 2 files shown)


server/server.go
```diff
@@ -2083,6 +2083,11 @@ func (s *Server) Start() {
 
 	defer s.Noticef("Server is ready")
 
+	// Check is this may be a restart.
+	if s.startupComplete == nil {
```
kozlovic, line 2087: I think one should ask if we want to support calling s.Start() after a s.Shutdown(). If yes, then I think it is a bit more than that, that needs to be done. I believe that there are others things that are created/setup in NewServer() whose state may be changed during Shutdown() and that Start() will not take into account.

server/server_test.go
```diff
@@ -2107,3 +2107,14 @@ func TestServerAuthBlockAndSysAccounts(t *testing.T) {
...
+
+// https://github.com/nats-io/nats-server/issues/4789
+func TestServerRestartAfterShutdown(t *testing.T) {
+	opts := DefaultTestOptions
+	opts.Port = -1
+	s := RunServer(&opts)
+	s.Shutdown()
+	s.WaitForShutdown()
+	s.Start()
+	defer s.Shutdown()
```
kozlovic, line 2119: Case in point, would s.Shutdown() not panic because channel `shutdownComplete` has already been closed in previous Shutdown()? My point is that I believe that there way more state that would need to be recreated/reset on a Start() after a Shutdown() that we can think of.

kozlovic: If it does not panic, it means that the second Shutdown() is actually not shutting down, likely because the code thinks it is already shutdown...

derekcollison: Meaning we should expand the test to make sure the server is usable? Or are you suggesting that the second start should fail and you can not re-use servers that have been shutdown?

kozlovic: The latter. I think that there is a lot of "state" that would need to be recreated/reset if we wanted to support a Start() after a Shutdown() (and therefore subsequent Shutdown() calls..). The only issue I have with that is Start() does not return an error, so that would leave the user not knowing that it is not started again. We could add a StartServer() or something like that, that would return an error (similar to what we did with NewServer vs New).

neilalexander:
This one worth rebasing?

PR #4804, opened 2023-11-20
[IMPROVED] Added in internal filestore state checks on write and recover.

derekcollison:
This led to some bug fixes around compaction where the states diverged. 
Also made all filestore tests that used permutations helper actually do encryption and be more standard.


neilalexander (approved):
LGTM

PR #4810, opened 2023-11-21
[FIXED] MQTT: Another potential lock contention in deliverMsgCbQoS0

levb:
Added a `[sess.]subsMu sync.RWMutex` to use in deliverMsgCbQoS0 and avoid the sess.mu.Lock() there. With a `#` subscription, deliverMsgCbQoS0 is getting invoked on route messages, and could stall them during subscription processing.

Test: WORK IN PROGRESS, submitting for an early code review. I do not yet have a definitive "before and after" test. I have been testing with a `nats-server` go test (everything in 1 process), and the Paho client to manage the state of all connections, subscriptions, etc. Unfortunately, the paho client is inefficient, opaque, and creating too much "go noise" - allocations, goroutines, etc. I need to refine the tests.


server/mqtt.go
```diff
@@ -1090,7 +1102,7 @@ func (s *Server) mqttCreateAccountSessionManager(acc *Account, quitCh chan struc
 	c := s.createInternalAccountClient()
 	c.acc = acc
 
-	id := getHash(s.Name())
+	id := s.NodeName()
```
levb, line 1105: This is synonymous; took the liberty of putting the nit here.

server/mqtt.go
```diff
@@ -1466,20 +1478,21 @@ func (jsa *mqttJSA) prefixDomain(subject string) string {
 
 func (jsa *mqttJSA) newRequestEx(kind, subject, cidHash string, hdr int, msg []byte, timeout time.Duration) (interface{}, error) {
 	var sb strings.Builder
-	jsa.mu.Lock()
 	// Either we use nuid.Next() which uses a global lock, or our own nuid object, but
 	// then it needs to be "write" protected. This approach will reduce across account
 	// contention since we won't use the global nuid's lock.
+	jsa.mu.Lock()
```
levb, line 1484: Nit, not related to the fix. Doesn't seem that anything other than getting the uuid needs to be under the lock.

server/mqtt.go
```diff
@@ -2164,10 +2177,16 @@ func (as *mqttAccountSessionManager) removeSession(sess *mqttSession, lock bool)
...
-// (pre-loaded) retained messages.
-// Session lock held on entry.
+// Helpers that sets the sub's mqtt fields and possibly serialize (pre-loaded)
+// retained messages.
+//
+// Session lock held on entry. Acquires the session subs lock and holds it for
+// the duration. Non-MQTT messages coming into mqttDeliverMsgCbQoS0 will be
+// waiting.
 func (sess *mqttSession) processSub(c *client, subject, sid []byte, isReserved bool, qos byte, jsDurName string, h msgHandler, initShadow bool) (*subscription, error) {
+	sess.subsMu.Lock()
```
levb, line 2187: This is the main fix. The only `sub.mqtt` data we need for QOS0 delivery is qos; so make a shorter-span `sess.subsMu` to protect it.

server/mqtt.go
```diff
@@ -2179,10 +2198,15 @@ func (sess *mqttSession) processSub(c *client, subject, sid []byte, isReserved b
 	}
 	for _, ss := range subs {
 		if ss.mqtt == nil {
-			ss.mqtt = &mqttSub{}
+			// reserved is set only once and once the subscription has been
+			// created it can be considered immutable.
+			ss.mqtt = &mqttSub{
+				reserved: isReserved,
```
levb, line 2204: This is a nit; we are still accessing `reserved` under a lock at all times.

server/mqtt.go
```diff
@@ -2415,10 +2439,12 @@ func (as *mqttAccountSessionManager) serializeRetainedMsgsForSub(rms map[string]
 		if !ok {
 			continue
 		}
+		c.mu.Lock()
```
levb, line 2442: @kozlovic I am 0/5 on this. I noticed that everywhere else we do `c.mu.Lock` to change `sub.mqtt.prm`, so I added it here; not 100% sure if it's necessary but doesn't appear to hurt?

server/mqtt.go
```diff
@@ -4271,32 +4297,32 @@ func mqttDeliverMsgCbQoS0(sub *subscription, pc *client, _ *Account, subject, re
...
 	// This is immutable
 	sess := cc.mqtt.sess
 
-	// Check the subscription's QoS. This needs to be protected because
-	// the client may change an existing subscription at any time.
-	sess.mu.Lock()
+	// Lock here, otherwise we may be called with sub.mqtt == nil. Ignore
+	// wildcard subscriptions if this subject starts with '$', per Spec
+	// [MQTT-4.7.2-1].
+	sess.subsMu.RLock()
```
levb, line 4309: The 2nd part of "the fix", replacing sess.mu with sess.subsMu here. We need to ensure that `sub.mqtt` has been set up, not a nil; and we need the QOS.

server/mqtt.go
kozlovic: The idea was that when we set that up here (for a new subscription), the callback - even if firing early - would not have access to the `prm` field because it would be blocked by the session lock (which this function runs under). But yes, I guess it is more correct to lock here.

server/mqtt.go
kozlovic: ok

server/mqtt.go
kozlovic: Technically there is `jsa.rplyr` but this is immutable, so should be fine done outside.

server/mqtt.go
kozlovic: Are you sure? We need to have retained messages prepped and in the `prm` writer prior to a callback be able to proceed. I don't think to recall that retained messages do not apply to QoS0 callback, at least I don't see that we don't do retained when qos==0 in `serializeRetainedMsgsForSub()`

server/mqtt.go
kozlovic: But it would not prevent the callback to proceed even if retained messages are not fully serialized. When calling `mqttEnqueuePublishMsgTo()` at the end of that function, there will be a check to see if the `prm` writer has something and then write it out before the messages. But that relied on the fact that they would be done under the session lock to prevent this "race" of the callback firing before we are ready.

kozlovic (changes requested):
I don't think this is correct with regards to retained messages and the QoS0 callback, but maybe I am wrong.

server/mqtt.go
levb: FWIW, I still think that the dedicated granular RWMutex for QOS0 makes sense since it's the "hot" callback; but yes retained messages need to be queued for delivery first, mosquitto does it, I checked. I'll add a test for this, it's easy on a cluster since there are most definitely many "internal" NATS messages in flight while processing an MQTT subscription.

@derekcollison's thought was that maybe the MQTT code (the QOS0 callback, what else?) is slowing down the route message processing. I will re-synchronize the writing out of the retained messages and benchmark it.

0/5 not sure if it's worth it, but maybe make mqttWriter a [][]byte with a `queueToClientOutbound()` that would queue up chunk by chunk (it all gets re-buffered again there anyway?). Or would it be too expensive to call `c.queueOutbound` for small chunks? Ditto in https://github.com/nats-io/nats-server/blob/e8772d592cab4d503fe9744ca73e5af2ffd155d0/server/mqtt.go#L4466, is there a reason for re-buffering the message and calling `c.queueOutbound` once, rather than separately for the header, and then for  msg itself? We may be copying a lot of bytes there, with a `#` sub.

Another 0/5. I understand that we consider NATS->MQTT QOS0 to be an essential feature, but should we put some constraints on it? I have tried naive blacklisting attempts, maybe it's a direction worth exploring more? Are there more kinds of (cluster-specific?) internal messages that we should "firewall out" of the MQTT code altogether?

server/mqtt.go
levb: (resolving this in favor of the other thread)

server/mqtt.go
kozlovic: You mean start to queueOutbound retained messages one by one instead of using the `prm`? That we can't because we want the retained messages to go before incoming messages, but they have to go after the SUBACK. This is why I used this buffer. 
As to make the writer a [][]byte instead of a single writer, maybe, but not sure how this is going to change much. But you can experiment and see... Also, to see any difference my guess is that you would need to get a LOT of retained messages (that is, lot of different topics, etc..).

levb: > You mean start to queueOutbound retained messages one by one instead of using the prm? That we can't because we want the retained messages to go before incoming messages, but they have to go after the SUBACK. This is why I used this buffer.

No, that's understood, and understood well now. I was just suggesting that I keep the new `subsMu` lock specifically for QoS0 delivery, and move the rms->prm serialization under it, potentially with the [][] optimization so we don't need to copy the bytes while holding the lock. 

> you would need to get a LOT of retained messages (that is, lot of different topics, etc..).

Agreed. But lots of messages can happen with wildcard subscriptions, and their size can be considerable. Hard to estimate without the use-case (and IMO it'd be a sub-optimally architected app to do this kind of thing with large messages, but I don't know...)

kozlovic: If serialization is under subsMu, then it could work, but then I am not sure that will minimize by that much, but only experimentation will tell.

server/mqtt.go
```diff
@@ -382,7 +400,7 @@ type mqttReader struct {
 }
 
 type mqttWriter struct {
-	bytes.Buffer
+	*bytes.Buffer
```
levb, line 403: It's an extra indirection, but makes initializing it so much easier. Consider avoiding the extra pointer.

server/mqtt.go
```diff
@@ -3837,37 +3921,44 @@ func (c *client) mqttHandlePubRetain() {
 	// but should still be delivered as a normal message.
 	if pp.sz == 0 {
 		if seqToRemove := asm.handleRetainedMsgDel(key, 0); seqToRemove > 0 {
-			asm.deleteRetainedMsg(seqToRemove)
-			asm.notifyRetainedMsgDeleted(key, seqToRemove)
+			go func() {
```
levb, line 3924: 2/5 `mqttHandlePubRetain` is opportunistic to begin with, as I understand it we only needed to block to get the sequence numbers updated in the session cache; can be done in a goroutine, and unblock `c.processIncomingMsg`?

0/5 doing this with an unbounded number of goroutines

server/mqtt.go
levb: I changed it now so that `subsMu` covers just the initialization of the (mqtt) subs, and there is a per-sub WaitGroup indicating that its `prm` is ready.

levb: I am going to resolve this thread since there have been a lot of changes since, and the core issue (serializing and sending prm to a sub before other messages) should have been addressed.

server/mqtt.go
```diff
@@ -270,6 +274,7 @@ type mqttRetMsgDel struct {
 
 type mqttSession struct {
 	mu                     sync.Mutex
+	subsMu                 sync.RWMutex
```
levb, line 277: NS: 1/5 consider making this a sync.WaitGroup as well. Need to research vs RWMutex

server/mqtt.go
```diff
@@ -329,13 +334,26 @@ type mqttRetainedMsgRef struct {
...
+	// quickly accessed using sess.subsMu.RLock, or under the main session lock.
+	qos   byte
 	jsDur string
-	// If this subscription needs to be checked for being reserved. E.g. # or * or */
+
+	// Pending serialization of retained messages to be sent when subscription
+	// is registered. The sub's delivery callbacks must wait until `prm` is
+	// ready (can block on sess.mu for that, too).
+	prm      [][]byte
+	prmReady *sync.WaitGroup
```
levb, line 352: NS: check that gets (re-)set properly when re-subscribing and re-delivering retained (with the new QOS).

levb: NS: do we, should we re-deliver retained if there's a re-subscribe with the same QOS as before?

server/mqtt.go
```diff
@@ -183,6 +183,10 @@ const (
 
 	// For Websocket URLs
 	mqttWSPath = "/mqtt"
+
+	mqttInitialPRMCapacity = 32 * 1024
+	mqttInitialPubHeader   = 16 // An overkill, should need 7 bytes max
+	mqttProcessSubTooLong  = 100 * time.Millisecond
```
levb, line 189: NS: lower?

server/mqtt.go
levb: Also, the order of execution will be different here; perhaps best to funnel through a channel/queue.

server/mqtt.go
levb: I looked into pre-serialization, and I think I ran into a dependency I couldn't resolve easily. I'll look into it one more time.

And yes, on while this change would unblock other subs on the session, it would still be holding the ROUTE, etc. messages coming into the sub.

server/mqtt.go
```diff
@@ -3666,6 +3746,46 @@ func mqttNewDeliverablePubRel(pi uint16) (natsMsg []byte, headerLen int) {
 	return buf.Bytes(), buf.Len()
 }
 
+// Composes a NATS message for a storeable mqttRetainedMsg.
+func mqttEncodeRetainedMessage(rm *mqttRetainedMsg) (natsMsg []byte, headerLen int) {
```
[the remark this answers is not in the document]
levb: Currently, we encode the message payload as a []byte in JSON, turning every byte into 2 hex characters. This change's intention is to store the JSON (metadata only) in the header, while the message body is the unencoded original. The decode function is backwards-compatible; it decodes the message body if the header is missing. 

This is not a 100% backwards-compatible change; rollback would be impossible, etc.

server/mqtt_test.go
```diff
@@ -333,6 +331,7 @@ func testMQTTRunServer(t testing.TB, o *Options) *Server {
 	}
 	l := &DummyLogger{}
 	s.SetLogger(l, true, true)
+	// s.ConfigureLogger()
```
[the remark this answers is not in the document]
levb: yy, this commented out line snuck in, will remove.

server/mqtt.go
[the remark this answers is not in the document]
levb: I'll make this into a separate PR.

server/mqtt.go
```diff
@@ -4271,32 +4375,39 @@ func mqttDeliverMsgCbQoS0(sub *subscription, pc *client, _ *Account, subject, re
...
 
-	// We have a wildcard subscription and this subject starts with '$' so ignore per Spec [MQTT-4.7.2-1].
-	if isReservedSub {
+	if ignore {
 		return
 	}
 
+	// Do not proceed until  all retained messages for the sub have been
+	// serialized.
+	if prmReady != nil {
```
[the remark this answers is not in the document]
kozlovic: Re: c.addToPCD(). This is because this map is meant to be used only in the client's readLoop. So there is no locking used there. If we divert messages from incoming readLoop (say pc.kind == ROUTER, etc..), then we can no longer use addToPCD() and will have to find a way to know which "destination" clients we need to flushOutbound/signal_the_writeLoop.
I am not sure at the moment how we will do that (efficiently). One simple approach would be to flush or signal each `cc` in mqttEnqueuePublishMsgTo().

server/mqtt.go
```diff
@@ -3837,37 +3957,43 @@ func (c *client) mqttHandlePubRetain() {
...
 		}
 	} else {
 		// Spec [MQTT-3.3.1-5]. Store the retained message with its QoS.
 		// When coming from a publish protocol, `pp` is referencing a stack
 		// variable that itself possibly references the read buffer.
-		rm := &mqttRetainedMsg{
+
+		// TODO: (levb) consider storing the JSON in the header, and pp.msg
+		// directly; no need to encode it to JSON.
+		rmBytes, hdr := mqttEncodeRetainedMessage(&mqttRetainedMsg{
```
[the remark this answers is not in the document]
levb: Resolving in favor of https://github.com/nats-io/nats-server/pull/4810#discussion_r1406627253

server/mqtt.go
```diff
@@ -5178,15 +5292,27 @@ func (w *mqttWriter) WriteBytes(bs []byte) {
 }
 
 func (w *mqttWriter) WriteVarInt(value int) {
+	w.Write(mqttSerializeVarInt(value))
+}
+
+func mqttSerializeVarInt(value int) []byte {
```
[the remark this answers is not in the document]
levb: yes, it was for trying to resolve what I perceived as an inconsistency with `mqttWritePublish`. I'll tidy this up, I understand your point about reusing it.

server/mqtt.go
levb: Hm, ok, I am going to defend my approach some more.

1. The "serialization" no longer involves copying the message bytes, it now takes 
  - (a) a second `as.sl` scan for each subject (can be cached from loadRetainedMessages if we wanted to)
  - (b) serializing the header (QOS, PI) that is specific to the sub/message
 Both of these should already be quite quick.

2. Using a separate (from `sess.mu`), cheaper lock for initializing `sub.mqtt` just so we can even engage in the (QOS0) callback - makes sense, thus `subsMu`. 

3. 0/5 about `sub.mqtt.prmReady` - it can be removed, because (a). We can use sess.subsMu (just like sess.mu before) to block for the duration of building the `prm`.

server/mqtt.go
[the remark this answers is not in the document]
levb: https://github.com/nats-io/nats-server/pull/4825

server/mqtt.go
```diff
@@ -2297,12 +2375,6 @@ func (as *mqttAccountSessionManager) processSubs(sess *mqttSession, c *client,
 		sess.cons[sid] = cc
 	}
 
-	serializeRMS := func(sub *subscription) {
```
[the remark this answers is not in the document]
levb: So, the act of serializing the header depends on the sub (QOS), and requires the session lock (PIs).

That's why I kept it where it was, but I removed the payload from there, we now just move the []byte pointer.

Like I mentioned in my other comment, we scan the RM "subs" (as.cl) again, that can be cached; otherwise I am not sure what else is there to do. The PI allocation must be done under the session lock, of course.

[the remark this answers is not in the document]
levb: (resolving this in favor of https://github.com/nats-io/nats-server/pull/4810/files?diff=split&w=1#r1406463798)

server/mqtt.go
```diff
@@ -2315,18 +2387,23 @@ func (as *mqttAccountSessionManager) processSubs(sess *mqttSession, c *client,
...
 		var jscons *ConsumerConfig
 		var jssub *subscription
 
 		// Note that if a subscription already exists on this subject, the
 		// existing sub is returned. Need to update the qos.
+		var sub *subscription
+		var err error
 		as.mu.Lock()
 		sess.mu.Lock()
-		sub, err := sess.processSub(c, bsubject, bsid,
```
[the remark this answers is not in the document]
levb: Resolving, see https://github.com/nats-io/nats-server/pull/4810/files?diff=split&w=1#r1406463798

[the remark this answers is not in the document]
levb: Resolving in favor of https://github.com/nats-io/nats-server/pull/4810/files?diff=split&w=1#r1406463798.

That aside, there's a syntactical difference; I thought it'd be more intention-reveaing (even if less DRY) to have the `if fromSubProto` higher up.

server/mqtt.go
levb: Removed. 

Going back to processing messages coming from ROUTER and such, isn't it possible that there are mqtt (retain=true) messages coming over from ROUTER, and wouldn't we be then holding the ROUTER client read loop? That'd seem bad?.. Or am I confused and this is only invoked for locally received MQTT messages?

server/mqtt.go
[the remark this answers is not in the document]
levb: (keeping this thread for the discussion of handling all special clients asynchronously)

server/mqtt.go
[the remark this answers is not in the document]
levb: it is no longer used, I restored the original code

server/mqtt_test.go
```diff
@@ -7500,7 +7464,7 @@ func mqttBenchPubQoS1(b *testing.B, subject, payload string, numSubs int) {
 	b.StartTimer()
 	for i := 0; i < b.N; i++ {
 		if pi <= maxpi {
-			mqttWritePublish(w, 1, false, false, subject, pi, []byte(payload))
+			testMQTTWritePublishPacket(b, w, 1, false, false, subject, pi, []byte(payload))
```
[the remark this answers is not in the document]
levb: K, should be fixed now.

levb:
@kozlovic please review. New changes:
- took out the goroutines in mqttHandlePubRetain
- tidied up the mqttWritePublish/test refactoring to address the overhead concern
- moved the de-JSON-ification of stored retained messages into a separate PR.

I answered the "serialize-before-we-lock" question in the thread; the short is - we effectively are already doing it in this PR.

I'd like to further understand the `ROUTER` etc. message handling, and whether we want to queue up messages for delivery. (And if so - which messages).

levb:
Based on the comments and a zoom with @derekcollison, @kozlovic and @neilalexander, the following TODO:

- [ ] Add a timed cache of retained messages so we don't re-load multiple times
- [ ] Fan out the JS API calls to load RMS
- [ ] Take out `prmReady`, do the RMS "serialization" inside `subsMu.Lock`
- [ ] Make `sess.processSub` a single function again
- [ ] In initialization, if MaxPerSubject is already 1, do not `transferRetained...`
- [ ] (@levb NS) Investigate if we're receiving/processing RM updates from "self", see if should/can be avoided like we do with sessions.

derekcollison:
If you load all retained messages be better to do ordered consumer internally I think.

server/mqtt.go
```diff
@@ -2535,22 +2592,30 @@ func (as *mqttAccountSessionManager) serializeRetainedMsgsForSub(rms map[string]
...
 	if len(result.psubs) == 0 {
-		return
+		return 0
 	}
+	numToDeliver := 0
 	for _, sub := range result.psubs {
 		subject := string(sub.subject)
-		if rms[subject] != nil {
-			continue // already loaded
+		ref := as.retmsgs[subject]
```
kozlovic, line 2621: This is wrong, since you can't use that map without the `as`'s lock.

server/mqtt.go
```diff
@@ -2535,22 +2592,30 @@ func (as *mqttAccountSessionManager) serializeRetainedMsgsForSub(rms map[string]
...
+		if ref == nil {
+			// If we don't have anything for the subject - skip, but this should
+			// not happen.
+			continue
+		}
+		if ref.loaded != nil && ref.loaded.expiresFromCache.After(time.Now()) {
```
kozlovic, line 2627: That would be wrong too and be racy since this is not protected by the `as`'s lock. Also, if we want to use the `ref.loaded` we should either bump the expiration or turnoff expiration for that message until we are done with it.

server/mqtt.go
```diff
@@ -2563,8 +2628,14 @@ func (as *mqttAccountSessionManager) loadRetainedMessagesForSubject(rms map[stri
 			log.Warnf("failed to decode retained message for subject %q: %v", loadSubject, err)
 			continue
 		}
-		rms[subject] = &rm
+
+		// Add the loaded retained message to the cache.
+		ref.loaded = &rm
+		ref.loaded.expiresFromCache = time.Now().Add(mqttRetainedCacheTTL)
```
kozlovic, line 2645: Again, this should be protected, but also assumes that this message will not expire by the time we need it.

kozlovic (changes requested):
Just reviewed commit "Add RM cache" and this needs some fixing (locking issues).

server/mqtt.go
```diff
@@ -2764,55 +2775,65 @@ func (as *mqttAccountSessionManager) transferUniqueSessStreamsToMuxed(log *Serve
...
 }
 
-func (as *mqttAccountSessionManager) transferRetainedToPerKeySubjectStream(log *Server) bool {
+func (as *mqttAccountSessionManager) transferRetainedToPerKeySubjectStream(log *Server) error {
 	jsa := &as.jsa
-	var count, errors int
+	var processed int
+	var transferred int
 
+	const timeoutAfter = 1 * time.Second
```
kozlovic, line 2783: This is I think way too low to transfer possibly large amount of retained messages. Knowing that now the client connection will fail, we need to make sure that this does not lead to too many times application not able to connect, or worse never able to since and error in this function is considered "fatal". 
I am thinking now that this is maybe too much of a risk: being unable to connect any client because an error transferring messages. For instance with your other PR changing the JSON encoding, you could imagine a situation similar in the future and that would prevent the server with this code to ever accept a client connection.

server/mqtt.go
```diff
@@ -2764,55 +2775,65 @@ func (as *mqttAccountSessionManager) transferUniqueSessStreamsToMuxed(log *Serve
...
 		// Unmarshal the message so that we can obtain the subject name.
 		var rmsg mqttRetainedMsg
-		if err := json.Unmarshal(smsg.Data, &rmsg); err != nil {
+		if err = json.Unmarshal(smsg.Data, &rmsg); err == nil {
+			// Store the message again, this time with the new per-key subject.
+			subject := mqttRetainedMsgsStreamSubject + rmsg.Subject
+			if _, err = jsa.storeMsg(subject, 0, smsg.Data); err != nil {
+				log.Errorf("    Unable to transfer the retained message with sequence %d: %v", smsg.Sequence, err)
+			}
+			transferOK = true
```
kozlovic, line 2809: Well not really if there was a failure. But here a transferred++ would have worked, no need for transferOK variable. Also, since we continue, we will delete the original message.

levb:
> If you load all retained messages be better to do ordered consumer internally I think.

Oops, I first put them in a separate `sync.Map` but later got greedy for "simplicity" and moved to the already existing map, missed the locking. Will fix now.

server/mqtt.go
```diff
@@ -2764,55 +2775,65 @@ func (as *mqttAccountSessionManager) transferUniqueSessStreamsToMuxed(log *Serve
...
-				break
-			}
-			log.Warnf("    Unable to load retained message from '$MQTT.rmsgs': %s", err)
-			errors++
+		if IsNatsErr(err, JSNoMessageFoundErr) {
+			// We've ran out of messages to transfer, done.
 			break
 		}
+		if err != nil {
+			log.Warnf("    Unable to transfer a retained message: failed to load from '$MQTT.rmsgs': %s", err)
```
kozlovic, line 2796: We have seen a situation where the load returns a timeout (did not figure out how can that be, but it happened). So that means that if this error (timeout) were to repeat, uses would not be able to connect any client. This is maybe a bit too risky.

kozlovic (changes requested):
Not so sure about timing this process out, and the risk of getting a failure that would prevent clients to connect. I would say maybe we don't time it out, and try to transfer them all, but then make it best effort with proceeding with the update of the stream (and possibly deleting the then single message in `$MQTT.rmsg` subject.
But writing that (about trying to transferring them all), if the issue is a timeout, and repeated for each message, this is also pretty bad. Not sure what to do here.

server/mqtt.go
levb: 👍  to your overall comment - make 1-time best effort, proceed with the update, attempt to delete the possible one remaining message.

levb: I did not remove the timeout altogether, upped it to 5s, but maybe still not enough?

@derekcollison  suggested using an ordered consumer to load them quicker, if I understood correctly. Have not gotten to that yet.

server/mqtt.go
```diff
@@ -2138,20 +2138,20 @@ func (as *mqttAccountSessionManager) handleRetainedMsg(key string, rm *mqttRetai
...
 			// Update the in-memory retained message cache but only for messages
 			// that are already in the cache, i.e. have been (recently) used.
-			if erm.loaded != nil {
-				erm.loaded = rm.loaded
-				erm.loaded.expiresFromCache = time.Now().Add(mqttRetainedCacheTTL)
+			if rm != nil {
+				if _, ok := as.rmsCache.Load(key); ok {
+					toStore := *rm
+					toStore.expiresFromCache = time.Now().Add(mqttRetainedCacheTTL)
+					as.rmsCache.Store(rm.Subject, toStore)
```
kozlovic, line 2153: Why not use `key` here?

server/mqtt.go
```diff
@@ -242,6 +242,7 @@ type mqttAccountSessionManager struct {
 	flapTimer  *time.Timer                    // Timer to perform some cleanup of the flappers map
 	sl         *Sublist                       // sublist allowing to find retained messages for given subscription
 	retmsgs    map[string]*mqttRetainedMsgRef // retained messages
+	rmsCache   sync.Map                       // map[string(subject)]mqttRetainedMsg
```
kozlovic, line 245: Should we not store pointers (`*mqttRetainedMsg`)?

server/mqtt.go
kozlovic: I understand that we need to make a copy and store again, but should we store a `*mqttRetainedMsg` instead?

server/mqtt.go
```diff
@@ -2551,16 +2551,15 @@ func (as *mqttAccountSessionManager) serializeRetainedMsgsForSub(sess *mqttSessi
...
-		if rm == nil {
+
+		strsubj := string(psub.subject)
+		rmv, _ := as.rmsCache.Load(strsubj)
+		if rmv == nil {
 			// This should not happen since we pre-load messages into the cache
 			// before calling serialize.
 			continue
 		}
+		rm := rmv.(mqttRetainedMsg)
```
kozlovic, line 2570: See my previous comments about storing pointers instead.

server/mqtt.go
```diff
@@ -2624,9 +2623,13 @@ func (as *mqttAccountSessionManager) loadRetainedMessagesForSubject(topSubject [
...
 		}
-		if ref.loaded != nil && ref.loaded.expiresFromCache.After(time.Now()) {
-			numToDeliver++
-			continue // already loaded and still good
+		rmv, ok := as.rmsCache.Load(subject)
+		if ok {
+			rm := rmv.(mqttRetainedMsg)
+			if rm.expiresFromCache.After(time.Now()) {
+				numToDeliver++
+				continue // already loaded and still good
```
kozlovic, line 2639: But again, it could be that a SUB starts and needs a retained message that was loaded/stored in cache by another subscription some time ago. It could be valid at this time but be just few milliseconds away from expiring. So by the time this function returns and we are going to use that map, the message may have been deleted.

server/mqtt.go
```diff
@@ -2624,9 +2623,13 @@ func (as *mqttAccountSessionManager) loadRetainedMessagesForSubject(topSubject [
 			// not happen.
 			continue
 		}
-		if ref.loaded != nil && ref.loaded.expiresFromCache.After(time.Now()) {
```
kozlovic: Few lines above, you still reference `as.retmsgs[subject]` outside of the lock, and you can't.

kozlovic (changes requested):
I am not convinced that this is still ok. See my comment about a message that may be in the cache but about to expire.

server/mqtt.go
```diff
@@ -1374,22 +1374,28 @@ func (s *Server) mqttCreateAccountSessionManager(acc *Account, quitCh chan struc
...
-			return nil, err
+			return
 		}
-
 		// We need another lookup to have up-to-date si.State values in order
 		// to load all retained messages.
 		si, err = lookupStream(mqttRetainedMsgsStreamName, "retained messages")
 		if err != nil {
-			return nil, err
+			return
```
kozlovic, line 1389: Well, for this, a failure is bad because it means that `si`, which is expected to reflect the latest state, will be not the correct one.

kozlovic (changes requested):

server/mqtt.go
levb: 0/5 I chose to store a copy to simplify dealing with concurrent access. What would be the advantage of storing a pointer to the cloned object, as opposed to the object itself if either way they are wrapped as an interface{}?

server/mqtt.go
levb: Right. Thx!

server/mqtt.go
levb: Understood. I missed the "expiring from having been fetch before" use-case.  I 3/5 don't want to "touch" the expiry and risk old values getting stuck. I like having a yet another lock even less.

1/5 Will build a separate list of `rm`s locally and hold on to it for serialization, sort of like before when it was a 1-time loaded map.

server/mqtt.go
kozlovic: Yes, ok.

server/mqtt.go
kozlovic: Yes, so in case where a SUB starts and we need to "load" retained messages, I would have used the same `rms` local map than before, but the loading part should then first check the cache, and if in the cache, set the message into the `rms` map. If not, do the actual load, set in `rms` map and in cache.

server/mqtt.go
kozlovic: I guess if you go back to use `rms`, this will be removed.

server/mqtt.go

server/mqtt.go
```diff
@@ -183,6 +183,11 @@ const (
 
 	// For Websocket URLs
 	mqttWSPath = "/mqtt"
+
+	mqttInitialPRMCapacity = 32 * 1024
```
kozlovic, line 187: Looks like this is no longer used.

server/mqtt.go
levb: should be resolved now - ignore the first time (since will retry anyway), fail the connection on the 2nd, if errors.

server/mqtt.go
```diff
@@ -1338,30 +1373,48 @@ func (s *Server) mqttCreateAccountSessionManager(acc *Account, quitCh chan struc
...
 		}
+		needToTransfer = false
+		return nil
 	}
+
+	// Attempt to transfer all "single subject" retained messages to new
+	// subjects. It may fail, will log its own error; ignore it the first time
+	// and proceed to updating MaxMsgsPer. Then we invoke transferRMS() again,
+	// which will get another chance to resolve the error; if not we bail there.
+	transferRMS()
```
kozlovic, line 1398: We must check the error here, and return if not nil. The reason is that if the transfer is ok, but the lookup failed, the `si` state would be wrong. I guess that since original si.Config.MaxMsgsPer would likely be != 1, we would update the stream (which updates `si`) and we do check for error at the last call to transferRMS(). Still, I think it would be correct to return error if lookup failed here.

server/mqtt.go
```diff
@@ -1797,9 +1851,12 @@ func (as *mqttAccountSessionManager) processRetainedMsg(_ *subscription, c *clie
 	seq, _, _ := ackReplyInfo(reply)
 
 	// Handle this retained message
-	rf := &mqttRetainedMsgRef{}
+	rf := &mqttRetainedMsgRef{
+		sseq: seq,
+	}
+
 	rf.sseq = seq
```
kozlovic, line 1858: Either remove here, or undo the change of initialization done above.

server/mqtt.go
```diff
@@ -2616,55 +2778,61 @@ func (as *mqttAccountSessionManager) transferUniqueSessStreamsToMuxed(log *Serve
 	retry = false
 }
 
-func (as *mqttAccountSessionManager) transferRetainedToPerKeySubjectStream(log *Server) bool {
+func (as *mqttAccountSessionManager) transferRetainedToPerKeySubjectStream(log *Server) error {
```
kozlovic, line 2781: We no longer check for error, so you could change that.

server/mqtt.go
```diff
@@ -4460,17 +4611,15 @@ func isMQTTReservedSubscription(subject string) bool {
...
-	w := &sw
-
-	flags := mqttSerializePublishMsg(w, pi, qos, dup, false, topic, msg)
+	flags, headerBytes := mqttMakePublishHeader(pi, qos, dup, false, topic, len(msg))
 
 	cc.mu.Lock()
-	if sub.mqtt.prm != nil {
-		cc.queueOutbound(sub.mqtt.prm.Bytes())
-		sub.mqtt.prm = nil
+	for _, data := range sub.mqtt.prm {
```
kozlovic, line 4617: Although code works if `sub.mqtt.prm` is `nil`, I think having the `if` not nil and setting it to `nil` under the `if` makes more sense:
```
if sub.mqtt.prm != nil {
    for _, data := range sub.mqtt.prm {
        cc.queueOutbound(data)
    }
    sub.mqtt.prm = nil
}
```

kozlovic (changes requested):

server/mqtt.go
levb: But if we return the error here, we would then get into the risky situation you were describing earlier, no? I thought we were trying to apply the MaxPerSubject update even on transfer error (we'd _always_ do it if we are transferring); the update would then reduce the "old" messages to 1, and then the second attempt will either clear it successfully (and update si), or fail altogether.

server/mqtt.go
levb: 0/5 prefer to keep since it makes sense for the function to return it. The caller is free to ignore, perhaps we do something else in the future with the error?

server/mqtt.go
levb: Also, maybe there is a way to delete that last remaining message without loading it first, I just don't know how; just deleting it would probably be cleaner (2nd pass).

kozlovic: Different that what we discussed. In this case, it is the lookupStream only that can return an error. That I would be fine with the idea that if we can't lookup the stream, we ought to fail. More over, if we have a failure looking up the stream:
```
		si, err = lookupStream(mqttRetainedMsgsStreamName, "retained messages")
		if err != nil {
			return err
		}
```
then `si` will be set to `nil`. So in that case we would get a panic at the next call:
```
	if si.Config.MaxMsgsPer != 1 {
		si.Config.MaxMsgsPer = 1
...
```
since `si` would have been set to nil due to failed `lookupStream()` call.

kozlovic (changes requested):
You need to check error from transferRMS(), even the first one, or use temp variables for  lookupStream and set `si` only on success, otherwise code will panic. See https://github.com/nats-io/nats-server/pull/4810#discussion_r1410049202

server/mqtt.go
levb: You're right. Updated both to return on error.

server/mqtt.go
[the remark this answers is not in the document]
levb: ((If we do the special handling, it'd be a separate PR, resolving this thread)

kozlovic (approved):
LGTM

levb:
@derekcollison @neilalexander this is ready for your final review. I am going to make a separate PR for the concurrent fetching of retained messages, it should not affect any of the changes here.

server/mqtt.go
```diff
@@ -1327,7 +1356,12 @@ func (s *Server) mqttCreateAccountSessionManager(acc *Account, quitCh chan struc
 				return nil, err
 			}
 		}
+		needToTransfer = false
+
+	default:
+		needToTransfer = si.Config.MaxMsgsPer != 1
```
derekcollison, line 1362: I would do > 1 here.

server/mqtt.go
```diff
@@ -1338,30 +1372,49 @@ func (s *Server) mqttCreateAccountSessionManager(acc *Account, quitCh chan struc
...
+		// We will need an up-to-date si, so don't use local variable here.
+		if si, err = jsa.updateStream(&si.Config); err != nil {
+			return nil, fmt.Errorf("failed to update stream config: %w", err)
 		}
 	}
 
+	// If we failed the first time, there is now at most one lingering message
+	// in the old subject. Try again (it will be a NO-OP if succeeded the first
+	// time).
+	if err = transferRMS(); err != nil {
```
derekcollison, line 1414: Once the update to maxMsgsPer is 1, what does it mean to call into here now?

server/mqtt.go
```diff
@@ -1797,9 +1851,11 @@ func (as *mqttAccountSessionManager) processRetainedMsg(_ *subscription, c *clie
 	seq, _, _ := ackReplyInfo(reply)
 
 	// Handle this retained message
-	rf := &mqttRetainedMsgRef{}
-	rf.sseq = seq
-	as.handleRetainedMsg(rm.Subject, rf)
+	rf := &mqttRetainedMsgRef{
```
derekcollison, line 1854: ok to do these in one line if you want.

server/mqtt.go
```diff
@@ -1937,6 +1993,41 @@ func (as *mqttAccountSessionManager) createSubscription(subject string, cb msgHa
...
 }
 
+// A timer loop to cleanup up expired cached retained messages for a given MQTT account.
+// The closeCh is used by the caller to be able to interrupt this routine
+// if the rest of the initialization fails, since the quitCh is really
+// only used when the server shutdown.
+//
+// No lock held on entry.
+func (as *mqttAccountSessionManager) cleaupRetainedMessageCache(s *Server, closeCh chan struct{}) {
+	tt := time.NewTicker(mqttRetainedCacheTTL)
```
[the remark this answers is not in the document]
derekcollison: Good catch.

server/mqtt.go
```diff
@@ -1937,6 +1993,41 @@ func (as *mqttAccountSessionManager) createSubscription(subject string, cb msgHa
...
+	for {
+		select {
+		case <-tt.C:
+			// Set a limit to the number of retained messages to scan since we
+			// lock as for it. Since the map enumeration gives random order we
+			// should eventually clean up everything.
+			i := 0
```
derekcollison, line 2010: Can do all these on one line.

derekcollison (commented):
A couple of questions / nits, but otherwise LGTM.
Make sure to defer ticker.

server/mqtt.go
neilalexander: `< 1` or `> 0` you mean?

derekcollison: Actually it is correct since we control it, will either be 1 or default which could be 0 or -1 but in this case will be 0 if we have not transferred.

server/mqtt.go
levb: By now there _may_ be a single message remaining, and we want to delete it. I do not know how to delete it without obtaining the sequence, and that's what the 2nd 1transferRMS` accomplishes. It only errors if it can't get the stream info.

server/mqtt.go
```diff
@@ -2475,6 +2622,16 @@ func (as *mqttAccountSessionManager) loadRetainedMessagesForSubject(rms map[stri
 		if rms[subject] != nil {
 			continue // already loaded
 		}
+
+		// See if we have the retained message in the cache.
+		if rmv, _ := as.rmsCache.Load(subject); rmv != nil {
+			rm := rmv.(mqttRetainedMsg)
```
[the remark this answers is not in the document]
derekcollison: ok I just meant if someone updates and existing rm whether we should bump the expiration because they updated it.

server/mqtt.go
```diff
@@ -2616,55 +2777,61 @@ func (as *mqttAccountSessionManager) transferUniqueSessStreamsToMuxed(log *Serve
...
 }
 
-func (as *mqttAccountSessionManager) transferRetainedToPerKeySubjectStream(log *Server) bool {
+func (as *mqttAccountSessionManager) transferRetainedToPerKeySubjectStream(log *Server) error {
 	jsa := &as.jsa
-	var count, errors int
+	var processed int
+	var transferred int
+
+	const timeoutAfter = 5 * time.Second
```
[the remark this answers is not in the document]
levb: 10s now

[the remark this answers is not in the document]
derekcollison: 10s better I think.

PR #4814, opened 2023-11-23
callout: try to renew jwt when expire

ramonberrutti:
When a jwt with expiration expires, we are going to request request a new one or disconnect the client.

### DIFF
--- a/server/auth_callout.go
+++ b/server/auth_callout.go
@@ -289,7 +289,7 @@ func (s *Server) processClientOrLeafCallout(c *client, opts *Options) (authorize
 		}
 
 		// Check if we need to set an auth timer if the user jwt expires.
-		c.setExpiration(arc.Claims(), expiration)
+		c.setRenewal(arc.Claims(), expiration)
 
 		respCh <- _EMPTY_
 	}
--- a/server/client.go
+++ b/server/client.go
@@ -254,6 +254,7 @@ type client struct {
 	darray     []string
 	pcd        map[*client]struct{}
 	atmr       *time.Timer
+	rtmr       *time.Timer // renew timer.
 	expires    time.Time
 	ping       pinfo
 	msgb       [msgScratchSize]byte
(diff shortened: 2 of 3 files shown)


server/client.go
```diff
@@ -4902,13 +4925,52 @@ func (c *client) setExpirationTimer(d time.Duration) {
...
+	c.setRenewalTimerUnlocked(d)
+	c.mu.Unlock()
+}
+
+func (c *client) setRenewalTimerUnlocked(d time.Duration) {
+	// Stop any previous timer
+	if c.rtmr != nil {
+		c.rtmr.Stop()
+	}
+	c.rtmr = time.AfterFunc(d, func() {
```
derekcollison, line 4950: Let's make this a formal function.

server/auth_callout.go
```diff
@@ -289,7 +289,7 @@ func (s *Server) processClientOrLeafCallout(c *client, opts *Options) (authorize
 		}
 
 		// Check if we need to set an auth timer if the user jwt expires.
-		c.setExpiration(arc.Claims(), expiration)
+		c.setRenewal(arc.Claims(), expiration)
```
derekcollison, line 292: This should have a buffer imo. So if expiration is say 30mins, we should ask for a renewal at some point before 30mins in case auth callout service is slow to respond. So some percentage, say 10-15% (so for 30mins would be 3mins, so timer would be 27m not 30m). We should also set a max offset for smaller ttls.

derekcollison (changes requested):

server/auth_callout.go
ramonberrutti: Do we want to retry for slow responds?

derekcollison: I do not believe so, at least to start. Just have the buffer so we are not cutting things too close.

derekcollison:
Any updates here or are we good for final review?

neilalexander:
Closing as old/stale.

PR #4820, opened 2023-11-27
Only gather subject filters if we need them

svenfoo:
This Just moves an assignment inside the if branch where it is used.  It's a micro optimization that's probably not worth making at all, but I came across this when reading the code and thought I could as well change it.


neilalexander (approved):
LGTM, thanks!

PR #4821, opened 2023-11-27
[ADDED] MQTT: test/bench using external client

levb:
This PR adds a number of MQTT-specific benchmarks, executed with an external test client (see https://github.com/ConnectEverything/mqtt-test/pull/1). This came out of my work on MQTT message delivery issues in 2.10, and is now packaged as a `go test --bench`, and added to CI.

The test client uses the Eclipse `paho` library, and can be executed standalone, pointed at an MQTT server. It measures the timing of its operations, and outputs it as JSON of reportable metrics.

The benchmarks here run the client, and report back the metrics it outputs. Memory allocations are measured server-only. 

Github action `benchmark` is used to compare the benchmark results to the base ref of the PR, or the prior result in a branch. If it fails a 120% threshold, the failure is reported as a comment on the commit(!), and the build fails.

To measure the effect of parallel-fetching retained messages (`SUBRET`), added `disable_retained_message_cache` MQTT option to the server. 

The following benchmarks are run, covering standalone/cluster, QOS, message sizes to 100K, as well as concurrent clients.
- `PUB` measures just publishing
- `PUBRET` measures publishing with the retained flag on
- `PUBSUB` measures message delivery times from publisher to subscriber
- `SUBRET` measures delivery of 100* retained messages to new subscriptions

Example of a failure (I rolled back the retained message "fanout" commit):
image
if you click on the comment link, you get:
image

###### TODO
- `CON` measure CONNECT to CONNACK, with stored sessions
- `SUB`  measuring SUBSCRIBE to SUBACK

### DIFF
--- a/.github/workflows/MQTT_test.yaml
+++ b/.github/workflows/MQTT_test.yaml
@@ -1,15 +1,15 @@
-name: MQTT Compliance 
+name: MQTTEx
 on: [push, pull_request]
+
+permissions:
+  pull-requests: write # to comment on PRs
+  contents: write # to comment on commits (to upload artifacts?)
 
 jobs:
   test:
-    strategy:
-      matrix:
-        go: ["1.21"]
     env:
       GOPATH: /home/runner/work/nats-server
       GO111MODULE: "on"
-
     runs-on: ubuntu-latest
     steps:
       - name: Checkout code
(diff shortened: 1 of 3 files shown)


.github/workflows/MQTT_test.yaml
```diff
@@ -20,18 +16,48 @@ jobs:
...
+
+      - name: Setup Go
+        uses: actions/setup-go@v4
+        with:
+          go-version: 1.21
+
+      - name: Install external test tools
+        shell: bash --noprofile --norc -x -eo pipefail {0}
+        run: |
+          go install github.com/levb/mqtt-test/...@f776ebe06dc0a813af3b11067515e4a4a544051e
```
levb, line 56: CHANGEME

derekcollison (commented):
So this adds in an option to disable the retained msg cache. When would one want to do this? Why are we exposing it I guess is my question.

PR looks fine, just looking for some clarifications.

levb:
@derekcollison the option is used only for testing, and is exposed because we use config files in the test utility functions  to initialize servers/clusters.

I could move the flag into the `(server.)srvMQTT` struct, set it after the server(s) have been started, would that be preferable? If there's a better mechanism of passing "test options" into a server, please advise.

derekcollison:
Let's see what @kozlovic has to say, he has done a few of these testing flags / switches for things.

server/mqtt.go
```diff
@@ -1142,6 +1142,9 @@ func (s *Server) mqttCreateAccountSessionManager(acc *Account, quitCh chan struc
 			quitCh: quitCh,
 		},
 	}
+	if !opts.MQTT.disableRetainedMessageCache {
```
kozlovic, line 1145: If we don't want to add an option (as to not expose in option parsing), what you could do is in this file, declare this boolean:
```
var testDisableRMSCache bool
```
And here, you would do:
```
	if !testDisableRMSCache {
		as.rmsCache = &sync.Map{}
	}
```
Now see test to know how to activate.

server/mqtt_ex_test.go
```diff
@@ -0,0 +1,421 @@
...
+		"--n", "1",
+		"--size", "100",
+		"--num-subscribers", "1")
+}
+
+func (bc *mqttBenchContext) startServer(b *testing.B, disableRMSCache bool) func() {
+	b.Helper()
+	b.StopTimer()
+	o := testMQTTDefaultOptions()
+	o.MQTT.disableRetainedMessageCache = disableRMSCache
```
kozlovic, line 274: Would remove that of course, instead, at the top of the function add:
```
testDisableRMSCache = true
```
and in the defer function for the shutdown:
```
return func() {
    testMQTTShutdownServer(s)
    testDisableRMSCache = false
}
```

But, this would work only if the tests are running sequentially, not in parallel (at least from the same test suite process), because then it could cause races and behave unexpectedly).

server/opts.go
```diff
@@ -572,6 +572,10 @@ type MQTTOpts struct {
 	// downgradeQOS2Sub tells the MQTT client to downgrade QoS2 SUBSCRIBE
 	// requests to QoS1.
 	downgradeQoS2Sub bool
+
+	// disableRetainedMessageCache disables the caching of retained images,
+	// used only for testing.
+	disableRetainedMessageCache bool
```
kozlovic, line 578: This could be removed if you go the `var testDisableRMSCache bool` route.

server/opts.go
```diff
@@ -4713,6 +4717,8 @@ func parseMQTT(v interface{}, o *Options, errors *[]error, warnings *[]error) er
 			o.MQTT.rejectQoS2Pub = mv.(bool)
 		case "downgrade_qos2_subscribe":
 			o.MQTT.downgradeQoS2Sub = mv.(bool)
+		case "disable_retained_message_cache":
```
kozlovic, line 4720: This could be removed if you go the `var testDisableRMSCache bool` route.

kozlovic (changes requested):
Assuming the proposed change does not cause races (tests not running in parallel), it could be a better way since we would not introduce an non exposed option.

server/mqtt.go
levb: Oh silly me, I forgot that even the cluster runs all in the same process! Of course, thank you. Raciness is not an issue since this is only used in benchmarks (no parallel execution)

.github/workflows/MQTT_test.yaml
```diff
@@ -20,18 +20,48 @@ jobs:
...
 
-      - name: Install deps
-        shell: bash --noprofile --norc -x -eo pipefail {0}
+      - name: Set up testing tools and environment
+        shell: bash --noprofile --norc -eo pipefail {0}
+        id: setup
         run: |
           wget https://github.com/hivemq/mqtt-cli/releases/download/v4.20.0/mqtt-cli-4.20.0.deb
           sudo apt install ./mqtt-cli-4.20.0.deb
+          # <>/<> FIXME: use @latest
```
levb, line 31: NS: use `@latest` once mqtt-test is merged

.github/workflows/MQTT_test.yaml
```diff
@@ -1,15 +1,15 @@
-name: MQTT Compliance 
+name: MQTTEx
 on: [push, pull_request]
```
levb, line 2: Arguably, these should be run on PR events only, there is little benefit to run on push. For `main` this can be run as part of nightly? @wallyqs @derekcollison?

wallyqs (approved):
LGTM

PR #4825, opened 2023-11-28
[IMPROVED] MQTT: do not JSON-encode retained message bodies

levb:
#### BREAKING
This change is not backwards-compatible. It will break in a mixed environment and/or when downgrading.

if a server runs this code, and the cluster/server is reverted to an older version or there are servers in the cluster that are still running older versions, the retained messages will fail to be loaded. 

#### SUMMARY

We used to JSON-encode the payloads of retained messages when storing to JetStream. This PR moves the JSON metadata into the NATS header, and stores the message bytes unencoded in the NATS message body.


server/mqtt.go
```diff
@@ -2481,15 +2482,55 @@ func (as *mqttAccountSessionManager) loadRetainedMessagesForSubject(rms map[stri
...
-		rms[subject] = &rm
+		rms[subject] = rm
 	}
 }
 
+// Composes a NATS message for a storeable mqttRetainedMsg.
+func mqttEncodeRetainedMessage(rm *mqttRetainedMsg) (natsMsg []byte, headerLen int) {
+	// Overallocate, 2K for all the header, JSON overhead
+	buf := bytes.NewBuffer(make([]byte, 0,
+		len(rm.Msg)+len(rm.Subject)+len(rm.Origin)+len(rm.Topic)+len(rm.Source)+2*1024))
```
kozlovic, line 2498: Seems a bit excessive, especially that the buffer can grow if needed.

kozlovic (approved):
LGTM, but let's be clear (for server release process purposes), that if a server runs this code, and the cluster/server is reverted to an older version, the retained messages will fail to be loaded.

server/mqtt.go
levb: You're right, the "overhead" is fixed already; adjusted it down to ~140 bytes; still too much but at this point doesn't matter, no need to be overly precise.

PR #4827, opened 2023-11-28
[IMPROVED] We can check CAS operations in stage1

derekcollison:
If and only if sequence is not zero and less than last seq.


neilalexander (approved):
LGTM

PR #4828, opened 2023-11-29
Tweak consumer monitor logic to avoid potential race

neilalexander:
Beforehand the wait group was not adjusted until the goroutine had started, whereas the `inMonitor` state was adjusted before, so it may have been possible to race here. Moving these two things together should hopefully avoid this and I've also renamed a function for clarity.

This should hopefully also fix a potential panic on wait group reuse in some chaos scenarios.

### DIFF
--- a/server/consumer.go
+++ b/server/consumer.go
@@ -5257,23 +5257,30 @@ func gatherSubjectFilters(filter string, filters []string) []string {
 	return filters
 }
 
-// Will check if we are running in the monitor already and if not set the appropriate flag.
-func (o *consumer) checkInMonitor() bool {
+// shouldStartMonitor will return true if we should start a monitor
+// goroutine or will return false if one is already running.
+func (o *consumer) shouldStartMonitor() bool {
 	o.mu.Lock()
 	defer o.mu.Unlock()
 
 	if o.inMonitor {
-		return true
+		return false
 	}
+	o.monitorWg.Add(1)
 	o.inMonitor = true
-	return false
+	return true
 }
 
-// Clear us being in the monitor routine.
+// Clear the monitor running state. The monitor goroutine should
+// call this in a defer to clean up on exit.
 func (o *consumer) clearMonitorRunning() {
 	o.mu.Lock()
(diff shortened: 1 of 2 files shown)


server/consumer.go
```diff
@@ -5257,23 +5257,30 @@ func gatherSubjectFilters(filter string, filters []string) []string {
 	return filters
 }
 
-// Will check if we are running in the monitor already and if not set the appropriate flag.
-func (o *consumer) checkInMonitor() bool {
+// shouldStartMonitor will return true if we should start a monitor
```
derekcollison, line 5260: return true if we did start the monitor routine.

server/consumer.go
```diff
@@ -5257,23 +5257,30 @@ func gatherSubjectFilters(filter string, filters []string) []string {
...
-// Clear us being in the monitor routine.
+// Clear the monitor running state. The monitor goroutine should
+// call this in a defer to clean up on exit.
 func (o *consumer) clearMonitorRunning() {
 	o.mu.Lock()
 	defer o.mu.Unlock()
-	o.inMonitor = false
+
+	if o.inMonitor {
+		o.monitorWg.Done()
```
derekcollison, line 5281: This is better done when actually exiting the go routine imo.

derekcollison: Meaning make it explicit, but this could be ok.

server/consumer.go
neilalexander: At this point we haven't actually started it yet, the caller to `shouldStartMonitor` is where the `startGoRoutine` call is.

server/consumer.go
neilalexander: I think the main thing is just to make sure `monitorWg` and `inMonitor` are updated at the same time, so they aren't subject to differing goroutine scheduling.

server/consumer.go
derekcollison: Apologies you are correct.

derekcollison (approved):
LGTM

PR #4829, opened 2023-11-29
Remove `"ocsp_peer_cache"` from varz response when not needed

neilalexander:


derekcollison (approved):
LGTM - Thanks

PR #4830, opened 2023-11-29
[IMPROVED] Avoid resetting WAL in RAFT layer if we already processed msg.

derekcollison:
When the raft layer receives an index less than ours, if all else ok simply ack.


kozlovic (approved):
LGTM

PR #4835, opened 2023-11-30
[IMPROVED] MQTT: concurrently fetch retained messages for a new SUB

levb:
Speed up sending out retained messages to a sub by loading them from JetStream concurrently.

- added `jsa.newRequestExMulti` that fans out multiple requests (on the same subject, at least for now) and gathers all responses
- further changed the processing of retained messages in `as.processSubs`: now we 
  - a) find all messages that we need and that are not in `rmsCache` first
  - (b) load them to `rms` with one call
  - (c) serialize to subs/filters and fail the filter if it misses messages from (a) in `rms`.


server/mqtt_test.go
```diff
@@ -3277,7 +3306,8 @@ func TestMQTTRetainedMsgMigration(t *testing.T) {
 	require_NoError(t, err)
 
 	// Publish some retained messages on the old "$MQTT.rmsgs" subject.
-	for i := 0; i < 100; i++ {
+	const N = 100
```
levb, line 3309: This was an accidental change from debugging (this test was failing at one point) but it seems worth leaving

server/mqtt.go
```diff
@@ -2416,21 +2485,54 @@ func (as *mqttAccountSessionManager) processSubs(sess *mqttSession, c *client,
...
+				if len(sub.shadow) > 0 {
+					subs = append(subs, sub.shadow...)
+				}
+				for _, sub := range subs {
+					subject := string(sub.subject)
+					// Best-effort loading the messages, logs on errors (to c.srv), loads
+					// once for subject.
+					as.addRetainedSubjectsForSubject(rmSubjects, subject)
+					for _, ss := range sub.shadow {
+						as.addRetainedSubjectsForSubject(rmSubjects, string(ss.subject))
```
levb, line 2510: 0/5 NS: if it makes sense to process shadow subs for retained messages, since those live in the mqtt topic space? (or is it NATS subject namespace?) check.

server/mqtt.go
```diff
@@ -2833,6 +2960,30 @@ func (as *mqttAccountSessionManager) transferRetainedToPerKeySubjectStream(log *
...
 }
 
+func (as *mqttAccountSessionManager) getCachedRetainedMsg(subject string) *mqttRetainedMsg {
+	v, ok := as.rmsCache.Load(subject)
+	if !ok {
+		return nil
+	}
+	rm := v.(mqttRetainedMsg)
+	if rm.expiresFromCache.Before(time.Now()) {
+		as.rmsCache.Delete(subject)
```
neilalexander, line 2970: If we never request that retained message again, then we will never delete the cache entry, so we maybe either need something that periodically fires to check for expired messages, or we just need to loop through all of them here and evacuate all dead entries.

neilalexander (commented):
Overall approach looks OK but a few minor things.

server/mqtt.go
```diff
@@ -1523,51 +1528,108 @@ func (jsa *mqttJSA) prefixDomain(subject string) string {
...
+	return r.value, nil
+}
 
-	ch := make(chan interface{}, 1)
-	jsa.replies.Store(reply, ch)
+// newRequestExMulti sends multiple messages on the same subject and waits for
+// all responses. Unless it times out, it returns the same number of responses
+// in the same order as msgs parameter.
+func (jsa *mqttJSA) newRequestExMulti(kind, subject, cidHash string, hdrs []int, msgs [][]byte, timeout time.Duration) ([]*mqttJSAResponse, error) {
+	if len(msgs) == 0 {
```
kozlovic, line 1553: If anything, then I would consider that an error, so return `nil, fmt.Errorf("...")`. But those checks that you have added are more like a dev error, not runtime, so either should be a panic, or simply removed.

server/mqtt.go
```diff
@@ -1523,51 +1528,108 @@ func (jsa *mqttJSA) prefixDomain(subject string) string {
...
+		case r := <-responseCh:
+			i := r2i[r.reply]
+			responses[i] = r
+			c++
+			if c == len(msgs) {
+				return responses, nil
+			}
+
+		case <-jsa.quitCh:
+			return nil, ErrServerNotRunning
```
kozlovic, line 1617: Would return `responses, ErrServerNotRunning` here.

server/mqtt.go
```diff
@@ -1523,51 +1528,108 @@ func (jsa *mqttJSA) prefixDomain(subject string) string {
...
+			var reply string
+			now := time.Now()
+			for reply = range r2i { // preserve the last value for Errorf
+				jsa.replies.Delete(reply)
+			}
+
+			if len(msgs) == 1 {
+				return nil, fmt.Errorf("timeout after %v: request type %q on %q (reply=%q)", now.Sub(start), kind, subject, reply)
+			} else {
+				return nil, fmt.Errorf("timeout after %v: request type %q on %q: got %d out of %d", now.Sub(start), kind, subject, c, len(msgs))
```
kozlovic, line 1629: Would return `responses` here too.

server/mqtt.go
```diff
@@ -1523,51 +1528,108 @@ func (jsa *mqttJSA) prefixDomain(subject string) string {
...
+
+		case <-t.C:
+			var reply string
+			now := time.Now()
+			for reply = range r2i { // preserve the last value for Errorf
+				jsa.replies.Delete(reply)
+			}
+
+			if len(msgs) == 1 {
+				return nil, fmt.Errorf("timeout after %v: request type %q on %q (reply=%q)", now.Sub(start), kind, subject, reply)
```
kozlovic, line 1627: Would return `responses` here, although when only 1 response, it could be `nil` here.

server/mqtt.go
```diff
@@ -1667,6 +1729,32 @@ func (jsa *mqttJSA) loadLastMsgFor(streamName string, subject string) (*StoredMs
...
+		bb, err := json.Marshal(mreq)
+		if err != nil {
+			return nil, err
+		}
+		marshaled = append(marshaled, bb)
+		headerBytes = append(headerBytes, 0)
+	}
+
+	all, err := jsa.newRequestExMulti(mqttJSAMsgLoad, fmt.Sprintf(JSApiMsgGetT, streamName), _EMPTY_, headerBytes, marshaled, mqttJSAPITimeout)
+	if err != nil {
```
kozlovic, line 1746: So here, I would not check error since `all` would always be returned with possibly some responses present, some nil.
```
responses := make([]*JSApiMsgGetResponse, len(all))
for i, v := range all {
	if v != nil {
		responses[i] = v.value.(*JSApiMsgGetResponse)
	}
}
return responses, err
```

server/mqtt.go
```diff
@@ -2416,21 +2487,55 @@ func (as *mqttAccountSessionManager) processSubs(sess *mqttSession, c *client,
...
+				if err := c.addShadowSubscriptions(acc, sub, false); err != nil {
+					return err
+				}
+
+				subs := []*subscription{sub}
+				if len(sub.shadow) > 0 {
+					subs = append(subs, sub.shadow...)
+				}
+				for _, sub := range subs {
+					subject := string(sub.subject)
```
kozlovic, line 2507: nit: Derek introduced bytesToString() which avoid mem alloc. I believe that in this case, since we are not going to store/etc.. the subject and use it just for the reverse match, it could be safe to do so.

server/mqtt.go
```diff
@@ -2416,21 +2487,55 @@ func (as *mqttAccountSessionManager) processSubs(sess *mqttSession, c *client,
...
+					// Best-effort loading the messages, logs on errors (to c.srv), loads
```
kozlovic, line 2508: Comment may not be up-to-date, since we are not loading anything here.

server/mqtt.go
```diff
@@ -2591,65 +2697,86 @@ func (as *mqttAccountSessionManager) serializeRetainedMsgsForSub(rms map[string]
...
 // Account session manager NOT lock held on entry.
-func (as *mqttAccountSessionManager) loadRetainedMessagesForSubject(rms map[string]*mqttRetainedMsg, topSubject []byte, log Logger) {
+func (as *mqttAccountSessionManager) addRetainedSubjectsForSubject(list map[string]struct{}, topSubject string) bool {
 	as.mu.RLock()
 	if len(as.retmsgs) == 0 {
 		as.mu.RUnlock()
-		return
+		return false
 	}
 	result := as.sl.ReverseMatch(string(topSubject))
```
kozlovic, line 2724: Since `topSubject` is now a string, we can remove the casting here.

server/mqtt.go
```diff
@@ -2591,65 +2697,86 @@ func (as *mqttAccountSessionManager) serializeRetainedMsgsForSub(rms map[string]
...
+	return added
+}
+
+type warner interface {
+	Warnf(format string, v ...any)
+}
+
+// Loads a list of retained messages given a list of stored message subjects.
+func (as *mqttAccountSessionManager) loadRetainedMessages(subjects map[string]struct{}, w warner) map[string]*mqttRetainedMsg {
+	rms := make(map[string]*mqttRetainedMsg)
```
kozlovic, line 2746: It is assumed that `rms` length will be of `subjects` length. So you could initialize with that, same for `ss`.

server/mqtt.go
```diff
@@ -2591,65 +2697,86 @@ func (as *mqttAccountSessionManager) serializeRetainedMsgsForSub(rms map[string]
...
-		loadSubject := mqttRetainedMsgsStreamSubject + subject
-		jsm, err := as.jsa.loadLastMsgFor(mqttRetainedMsgsStreamName, loadSubject)
-		if err != nil || jsm == nil {
-			log.Warnf("failed to load retained message for subject %q: %v", loadSubject, err)
+	if len(ss) == 0 {
+		return rms
+	}
+
+	results, err := as.jsa.loadLastMsgForMulti(mqttRetainedMsgsStreamName, ss)
+	if err != nil {
```
kozlovic, line 2761: Would warn if error, but proceed with what we got:
```
// If an error occurred, warn, but then proceed with what we got.
if err != nil {
    w.Warnf("error loading retained messages: %v", err)
}
for i, result := range results {
	if result == nil {
		continue
	}
	if err := result.ToError(); err != nil {
		w.Warnf("failed to load retained message for subject %q: %v", ss[i], err)
		continue
	}
       ....
```

server/mqtt.go
kozlovic: There is a long-running go routine that takes care of expiration, see `cleaupRetainedMessageCache`

server/mqtt.go
```diff
@@ -2833,6 +2960,30 @@ func (as *mqttAccountSessionManager) transferRetainedToPerKeySubjectStream(log *
...
+	}
+	rm := v.(mqttRetainedMsg)
+	if rm.expiresFromCache.Before(time.Now()) {
+		as.rmsCache.Delete(subject)
+		return nil
+	}
+	return &rm
+}
+
+func (as *mqttAccountSessionManager) setCachedRetainedMsg(subject string, rm mqttRetainedMsg, onlyReplace bool) {
```
kozlovic, line 2976: The fact that `rm` is not a pointer causes a copy of the whole struct when this function is invoked. So the caller is already doing `setCachedRetainedMsg(key, *rm, true)`, then a copy is made by the runtime to pass as a parameter. Does not seem too efficient. Also here it is actually what you want, but modifying `rm` would not modify the `rm` that is passed to the function.

kozlovic (changes requested):

server/mqtt.go
```diff
@@ -1523,51 +1528,108 @@ func (jsa *mqttJSA) prefixDomain(subject string) string {
...
-	reply := sb.String()
+	return r.value, nil
+}
 
-	ch := make(chan interface{}, 1)
-	jsa.replies.Store(reply, ch)
+// newRequestExMulti sends multiple messages on the same subject and waits for
+// all responses. Unless it times out, it returns the same number of responses
+// in the same order as msgs parameter.
+func (jsa *mqttJSA) newRequestExMulti(kind, subject, cidHash string, hdrs []int, msgs [][]byte, timeout time.Duration) ([]*mqttJSAResponse, error) {
```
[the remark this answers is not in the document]
levb: Done for timeout, doesn't apply elsewhere; also updated the comment.

server/mqtt.go
levb: Removed

server/mqtt.go
levb: I made it return `responses` on timeout, not sure what is the benefit returning values if the server is shutting down?

server/mqtt.go
```diff
@@ -1523,51 +1528,108 @@ func (jsa *mqttJSA) prefixDomain(subject string) string {
...
-		sb.WriteByte(btsep)
+func (jsa *mqttJSA) newRequestEx(kind, subject, cidHash string, hdr int, msg []byte, timeout time.Duration) (any, error) {
+	responses, err := jsa.newRequestExMulti(kind, subject, cidHash, []int{hdr}, [][]byte{msg}, timeout)
+	if err != nil {
+		return nil, err
+	}
+	if len(responses) != 1 {
+		return nil, fmt.Errorf("unreachable: invalid number of responses (%d)", len(responses))
+	}
+	r := responses[0]
```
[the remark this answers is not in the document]
levb: The value coming back may contain a higher-level (JS?) error, and this call is supposed to return it as such. Am missing something? I shortened the code a bit though.

server/mqtt.go
levb: I'd then argue that the same logic should apply, so returning a "sparsely populated" array of 1 `nil` is appropriate (and done now)

server/mqtt.go
levb: And `loadRetainedMessages` ignores the error now, loading and then delivering whatever messages came through.

server/mqtt.go
levb: A bit of a mess up on my part, the intention was to avoid storing the (extra) pointers in the map, but it is simply not worth the added complexity. Storing (and passing) pointers now

server/mqtt.go
```diff
@@ -1523,51 +1528,112 @@ func (jsa *mqttJSA) prefixDomain(subject string) string {
...
-		sb.WriteByte(btsep)
+func (jsa *mqttJSA) newRequestEx(kind, subject, cidHash string, hdr int, msg []byte, timeout time.Duration) (any, error) {
+	responses, err := jsa.newRequestExMulti(kind, subject, cidHash, []int{hdr}, [][]byte{msg}, timeout)
+	if err != nil {
+		return nil, err
+	}
+	if len(responses) != 1 {
+		return nil, fmt.Errorf("unreachable: invalid number of responses (%d)", len(responses))
+	}
+	v := responses[0].value
```
kozlovic, line 1539: Again, no, if newRequestExMulti() returns an error, we would have already returned at line 1533. For instance, here is how `loadNextMsgFor()` looks like:
```
lmri, err := jsa.newRequest(mqttJSAMsgLoad, fmt.Sprintf(JSApiMsgGetT, streamName), 0, req)
if err != nil {
    return nil, err
}
lmr := lmri.(*JSApiMsgGetResponse)
return lmr.Message, lmr.ToError()
```
And before your changes, `newRequestEx()` would have returned nil and an error on timeout. With your changes, you return the responses and the error, and if there is a timeout, the error will have been handled line 1533 as I said. If the error is part of the response object, it is up to the caller to use it. See above, we cast the response to `JSApiMsgGetResponse` and then return the message part and the error part (that may be nil).

Again, I fail to see why returning `responses[0].value, nil` here is not enough, and consistent with previous behaviour.

server/mqtt.go
kozlovic: The idea was that on error, we still may have gotten some valid responses and we should not assume what the caller wants/can do with those responses even in error condition. In the context that you are introducing this (retained messages), yes, I don't think that user could do really anything, so I guess it is fine either way.

server/mqtt.go
kozlovic: Not really the same because if only 1 response and we got the timeout, it means that responses[0] will always be nil anyway. And again, since the caller gets an error anyway, it will bail right after this call:
```
responses, err := jsa.newRequestExMulti(kind, subject, cidHash, []int{hdr}, [][]byte{msg}, timeout)
if err != nil {
    return nil, err
}
...
```
Meaning that the fact that we return `responses, <some error>`, the `responses` won't be used, but its content would be `nil` anyway.

server/mqtt.go
```diff
@@ -2833,6 +2959,30 @@ func (as *mqttAccountSessionManager) transferRetainedToPerKeySubjectStream(log *
...
+	}
+	rm := v.(*mqttRetainedMsg)
+	if rm.expiresFromCache.Before(time.Now()) {
+		as.rmsCache.Delete(subject)
+		return nil
+	}
+	return rm
+}
+
+func (as *mqttAccountSessionManager) setCachedRetainedMsg(subject string, rm *mqttRetainedMsg, onlyReplace bool) {
```
kozlovic, line 2975: Now I see that in the case of `handleRetainedMsg()` we call this with `rm` that is created and reference `pp.msg` that is no longer copied, while in this case I believe it should (see comment about `pp` in that function), or at least make it a copy in `handleRetainedMsg()` when storing if `rm != nil`...

server/mqtt.go
```diff
@@ -2833,6 +2959,30 @@ func (as *mqttAccountSessionManager) transferRetainedToPerKeySubjectStream(log *
 	return nil
 }
 
+func (as *mqttAccountSessionManager) getCachedRetainedMsg(subject string) *mqttRetainedMsg {
+	v, ok := as.rmsCache.Load(subject)
+	if !ok {
+		return nil
+	}
+	rm := v.(*mqttRetainedMsg)
```
kozlovic, line 2967: I think that there is still a place where you cast to `(mqttRetainedMsg)` without the pointer. That would be in `cleaupRetainedMessageCache()`. Not sure if Go does the right thing, but for consistency, I would change the casting there.

kozlovic (changes requested):

server/mqtt.go
levb: You're right, I over-complicated this. Fixing.

server/mqtt.go
levb: Right, It was coded correctly without the wrappers before. Thanks for catching!

server/mqtt.go
levb: fixed, thx! not sure how to add a test for the cleanup, would've been nice.

kozlovic:
@levb I just pushed a change that adds a test for the retained messages cache cleanup, that would have caught the panic during the interface conversion. Going to finish reviewing your recent changes now.

server/mqtt.go
```diff
@@ -2972,15 +2966,17 @@ func (as *mqttAccountSessionManager) getCachedRetainedMsg(subject string) *mqttR
...
-		if _, ok := as.rmsCache.Load(subject); ok {
-			as.rmsCache.Store(subject, rm)
+		if _, ok := as.rmsCache.Load(subject); !ok {
+			return
 		}
-	} else {
-		as.rmsCache.Store(subject, rm)
 	}
+	if copyBytesToCache {
+		rm.Msg = append([]byte(nil), rm.Msg...)
```
kozlovic, line 2978: Could be replaced with `rm.Msg = copyBytes(rm.Msg)`. I just bench'ed it, and it is generally faster and also allocates only what you need, as opposed to append that may create a bigger capacity than what is copied. Run with size of 5. Notice how the append will use 8 bytes.
```
Benchmark_Ivan_CopyBytes
Benchmark_Ivan_CopyBytes-16    	128584900	         8.200 ns/op	       5 B/op	       1 allocs/op
Benchmark_Ivan_Append
Benchmark_Ivan_Append-16       	74689119	        15.79 ns/op	       8 B/op	       1 allocs/op
```
Run with size of 64.
```
Benchmark_Ivan_CopyBytes
Benchmark_Ivan_CopyBytes-16    	57017350	        18.97 ns/op	      64 B/op	       1 allocs/op
Benchmark_Ivan_Append
Benchmark_Ivan_Append-16       	48468579	        24.95 ns/op	      64 B/op	       1 allocs/op
```

kozlovic (changes requested):
nit: could change to use copyBytes().

kozlovic (approved):
LGTM!

Issue #4842, opened 2023-12-04
Corrupted Jetstream subjects when regenerating stream index.db


oscarwcl:
### Observed behavior

Some subjects stored in a Jetstream filestore are corrupted when for a stream when regenerating a stream's index.db. This was initially observed when updating to 2.10.6 and restoring a stream which logged:
```
Stream state encountered internal inconsistency on recover
```
It can be reproduced with a fresh stream by manually deleting index.db.

Have only reproduced with a stream that contains large messages (close to 1 MB).

### Expected behavior

No corruption.

### Server and client version

nats-server 2.10.6, still occurs on `main`

Bisected to 66b0b5a566d95497ef28e75e161ec919910612a0

Observed via nats-cli 0.1.1

### Host environment

Debian 12 x86_64

### Steps to reproduce

<details>
<summary>
Stream config
</summary>

```
{
    "name": "test",
    "subjects": [
      "test.*"
    ],
    "retention": "limits",
    "max_consumers": -1,
    "max_msgs_per_subject": -1,
    "max_msgs": -1,
    "max_bytes": -1,
    "max_age": 604800000000000,
    "max_msg_size": -1,
    "storage": "file",
    "discard": "old",
    "num_replicas": 1,
    "duplicate_window": 120000000000,
    "sealed": false,
    "deny_delete": false,
    "deny_purge": false,
    "allow_rollup_hdrs": false,
    "allow_direct": false,
    "mirror_direct": false
}
```
</details>

```
# publish large messages to a bunch of subjects
for i in $(seq 1000); do yes test | head -c 900000 | nats publish "test.$i"; done

# observe correct subject names
nats stream subjects test

╭─────────────────────────────────────────────────────────╮
│               1000 Subjects in stream test              │
├──────────┬───────┬───────────┬───────┬──────────┬───────┤
│ Subject  │ Count │ Subject   │ Count │ Subject  │ Count │
├──────────┼───────┼───────────┼───────┼──────────┼───────┤
│ test.52  │ 1     │ test.981  │ 1     │ test.459 │ 1     │
│ test.838 │ 1     │ test.409  │ 1     │ test.48  │ 1     │
│ test.440 │ 1     │ test.672  │ 1     │ test.483 │ 1     │
│ test.315 │ 1     │ test.750  │ 1     │ test.303 │ 1     │
│ test.703 │ 1     │ test.171  │ 1     │ test.695 │ 1     │
[...]

PR #4851, opened 2023-12-05
[FIXED] Corruption of subjects on recover of bad or missing index.db

derekcollison:
When populating the block's fss during a manual recovery, we could hoist the fss to the global PSIM.
This could lead to data corruption since we were not copying. This revert's that change with a test.

The corruption was all internal at the PSIM layer and was not reflected on disk, so a server upgrade will fix with no data corruption.

Resolves: #4842 


wallyqs (approved):
LGTM

Issue #4850, opened 2023-12-05
NATS Server nil pointer when sourcing from one memory stream to another


andreib1:
### Observed behavior

Exiting the application after NATS has apparently shut down results in a NATS server panic.

``` console
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x60 pc=0xc9767c]

goroutine 611 [running]:
github.com/nats-io/nats-server/v2/server.(*memStore).GetSeqFromTime(0xc0005d4000, {0x0?, 0xc00199b998?, 0x0?})
	/home/andreib/go/pkg/mod/github.com/nats-io/nats-server/v2@v2.10.6/server/memstore.go:287 +0x15c
github.com/nats-io/nats-server/v2/server.(*consumer).selectStartingSeqNo(0xc001d9d400)
	/home/andreib/go/pkg/mod/github.com/nats-io/nats-server/v2@v2.10.6/server/consumer.go:4742 +0x25d
github.com/nats-io/nats-server/v2/server.(*stream).addConsumerWithAssignment(0xc000305880, 0xc00196b8d0, {0x0, 0x0}, 0x0, 0x0, 0x0)
	/home/andreib/go/pkg/mod/github.com/nats-io/nats-server/v2@v2.10.6/server/consumer.go:947 +0x1b5c
github.com/nats-io/nats-server/v2/server.(*stream).addConsumerWithAction(...)
	/home/andreib/go/pkg/mod/github.com/nats-io/nats-server/v2@v2.10.6/server/consumer.go:694
github.com/nats-io/nats-server/v2/server.(*Server).jsConsumerCreateRequest(0xc000016d80, 0x0?, 0xc000c33300, 0x3d?, {0xc00196d100, 0x40}, {0xc000c27ed8, 0x11}, {0xc0019685a0, 0x1d6, ...})
	/home/andreib/go/pkg/mod/github.com/nats-io/nats-server/v2@v2.10.6/server/jetstream_api.go:3901 +0x14b4
github.com/nats-io/nats-server/v2/server.(*jetStream).apiDispatch(0xc000070700, 0xc0000cf2c0, 0xc000c33300, 0xc00028c780, {0xc00196d100, 0x40}, {0xc000c27ed8, 0x11}, {0xc0019685a0, 0x1d6, ...})
[...]

PR #4853, opened 2023-12-06
[FIXED] If a memstore had a last sequence that was deleted, GetSeqFromTime could panic.

derekcollison:
Resolves: #4850 


wallyqs (approved):
LGTM

PR #4857, opened 2023-12-06
[FIXED] Protect against bad consumer state with rollback.

derekcollison:
If we had a rollback this could consume lots of memory during checkStateForInterestStream for followers since it will create pre-register acks incorrectly.

### DIFF
--- a/server/consumer.go
+++ b/server/consumer.go
@@ -2460,8 +2460,11 @@ func (o *consumer) setStoreState(state *ConsumerState) error {
 	if state == nil || o.store == nil {
 		return nil
 	}
-	o.applyState(state)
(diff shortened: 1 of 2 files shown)


server/consumer.go
```diff
@@ -5312,11 +5315,16 @@ func (o *consumer) checkStateForInterestStream() {
 		return
 	}
 
+	asflr := state.AckFloor.Stream
+	// Protect ourselves against rolling backwards.
+	if asflr&(1<<63) != 0 {
+		return
```
neilalexander, line 5321: Is there some use in logging if this happens?

neilalexander (approved):
Otherwise LGTM

PR #4862, opened 2023-12-07
Fix reloading message for prof_block_rate

wallyqs:
Also correcting a few typos in tests that have accumulated.


derekcollison (approved):
LGTM

PR #4867, opened 2023-12-09
[WIP] fix for JS limits being calculated by multiplying by the number of replicas

aricart:
[FIX] store usage for a message was calculated by multiplying by the number of replicas - this made limits be a distributed aggregate rather than the maximum for the resource


derekcollison (commented):
Thanks, also triggers incorrectly on simply setting max_bytes, so will look into both. May extend this or create a new one with proper tests.

derekcollison:
Was bit more involved, so did a separate branch and will post new PR. Thanks though!

PR #4868, opened 2023-12-10
[FIXED] When calculating limits exceeded, treat replicated tier limits as flat

derekcollison:
We had an issue with tiered accounts, where limits are flat based on type of storage. Meaning 100GB on R3 means 100GB of R3 storage, or 300GB in total. If we have limits only on an empty string tier, then consider limit as absolute, otherwise scale by tier replication.

Resolves issues with NGS account for HA / R3 limits.


aricart (approved):
LGTM

PR #4880, opened 2023-12-13
Move tokenizing of subject after early returns

svenfoo:
We do not always need the tokenized subject here.


neilalexander (approved):
LGTM, thanks!

svenfoo:
The call to tokenizeSubjectIntoSlice() showed up in the profile. Moving it after the early returns avoids it in some cases. Probably not a huge win, but it's an easy one.

derekcollison (approved):
LGTM

PR #4881, opened 2023-12-13
Move helper subjectInfo() to the file where it's used

svenfoo:
This method is only used from `subject_transform.go`, should it then live there as well?


derekcollison (approved):
LGTM

PR #4884, opened 2023-12-14
[IMPROVED] JSAPI internal routing and reporting and Source and Mirror setup

derekcollison:
A customer had an issue with JS api requests backing up in the internal queue we use to avoid blocking non-clients. The reason for the buildup was known due to time spent in NumPending calculations for R1 consumers with wildcard filtered subjects and large subject/key space. These consumers were being created from a large fan in stream with many sources.

1. We now properly track the queue depth of this internal queue, warn if we are getting backed up and make sure to properly report under JS API usage stats - Inflight.
2. Use multiple Go routines in a pool to process this API queue vs 1.
3. When we fail to create a consumer for a mIrror or a source, back off retries.
4. When we fail do immediate cleanup, do not wait.
5. Only create internal interest for the consumer once we get a response from the source stream.

### DIFF
--- a/server/jetstream_api.go
+++ b/server/jetstream_api.go
@@ -21,6 +21,7 @@ import (
 	"math/rand"
 	"os"
 	"path/filepath"
+	"runtime"
 	"sort"
 	"strconv"
 	"strings"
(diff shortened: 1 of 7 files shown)


neilalexander (approved):
LGTM

PR #4885, opened 2023-12-14
[IMPROVED] Do not load all blocks for NumPending when delivery is LastPerSubject.

derekcollison:
Update NumPending to use PSIM since block based fss is no longer present in >= 2.10.x
This was causing us to load all blocks to calculate NumPending for a new consumer with LastBySubject delivery.
Note this will be drastically improved with the new stree(ART) data structure in a future PR.


derekcollison:
I see that `TestJetStreamKVReductionInHistory` is failing due to this change. Will fix tomorrow.

wallyqs (approved):
LGTM

PR #4889, opened 2023-12-15
Add 'certs' option to TLS block for multi-certs support

wallyqs:
This adds a `certs` option to the `tls` block to support loading multiple certs:
    
```hcl
    tls {
      certs = [
        {
          cert_file: "./configs/certs/srva-cert.pem"
          key_file:  "./configs/certs/srva-key.pem"
        },
        {
          cert_file: "./configs/certs/srvb-cert.pem"
          key_file:  "./configs/certs/srvb-key.pem"
        }
      ]
      ca_file: "./configs/certs/ca.pem"
      verify:  true
      timeout: 2
    }
```

Follow up from https://github.com/nats-io/nats-server/pull/2029

### DIFF
--- a/server/opts.go
+++ b/server/opts.go
@@ -4681,16 +4681,28 @@ func GenTLSConfig(tc *TLSConfigOpts) (*tls.Config, error) {
 	case tc.CertFile == _EMPTY_ && tc.KeyFile != _EMPTY_:
 		return nil, fmt.Errorf("missing 'cert_file' in TLS configuration")
 	case tc.CertFile != _EMPTY_ && tc.KeyFile != _EMPTY_:
-		// Now load in cert and private key
-		cert, err := tls.LoadX509KeyPair(tc.CertFile, tc.KeyFile)
-		if err != nil {
-			return nil, fmt.Errorf("error parsing X509 certificate/key pair: %v", err)
+		// To support multiple certs (for SANs) without rearchitecting our
+		// configuration, we accept a list, separated by the platform list
+		// separator; both specs must contain the same count of entries.
+		// (So, colon-separated on Unix, semi-colon on Windows).
+		certFiles := strings.Split(tc.CertFile, string(os.PathListSeparator))
+		keyFiles := strings.Split(tc.KeyFile, string(os.PathListSeparator))
+		if len(certFiles) != len(keyFiles) {
+			return nil, fmt.Errorf("TLS configuration has %d entries in cert_file but %d in key_file", len(certFiles), len(keyFiles))
 		}
-		cert.Leaf, err = x509.ParseCertificate(cert.Certificate[0])
-		if err != nil {
-			return nil, fmt.Errorf("error parsing certificate: %v", err)
+		config.Certificates = make([]tls.Certificate, len(certFiles))
+		// Now load in cert and private keys
+		for i := range certFiles {
+			cert, err := tls.LoadX509KeyPair(certFiles[i], keyFiles[i])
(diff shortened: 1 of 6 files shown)


server/opts.go
```diff
@@ -4381,10 +4388,46 @@ func parseTLS(v interface{}, isClientCtx bool) (t *TLSConfigOpts, retErr error)
...
+					if !ok {
+						return nil, &configErr{tk, fmt.Sprintf("error parsing certificates config: unsupported type %T", vv)}
+					}
+					switch k {
+					case "cert_file":
+						certPair.CertFile = file
+					case "key_file":
+						certPair.KeyFile = file
+					default:
+						return nil, &configErr{tk, fmt.Sprintf("error parsing tls certs config, unknown field [%q]", k)}
```
philpennock, line 4416: tiny nit, `%q` will already quote, so the square brackets seem redundant

philpennock (approved):
What happens if given a config block with a present-but-empty list of certificates?

```
tls {
  certs: []
}
```

wallyqs:
> What happens if given a config block with a present-but-empty list of certificates?
> 
> ```
> tls {
>   certs: []
> }
> ```

that would have been the same as doing this:

```
tls {
}
```
which also makes the server start with TLS 'configured' but actually cannot be used.  I think we should change to prevent this?
In releases so far doing `tls {}` would result in this:
```
[86149] 2023/12/15 11:27:44.329219 [INF] TLS required for client connections
[86149] 2023/12/15 11:27:44.329509 [INF] Server is ready
[86149] 2023/12/15 11:27:59.796999 [ERR] 127.0.0.1:62606 - cid:5 - TLS handshake error: tls: no certificates configured
```

server/opts.go
```diff
@@ -4696,6 +4739,20 @@ func GenTLSConfig(tc *TLSConfigOpts) (*tls.Config, error) {
 		if err != nil {
 			return nil, err
 		}
+	case tc.Certificates != nil:
+		// Multiple certificate support.
+		config.Certificates = make([]tls.Certificate, len(tc.Certificates))
+		for i, certPair := range tc.Certificates {
+			cert, err := tls.LoadX509KeyPair(certPair.CertFile, certPair.KeyFile)
+			if err != nil {
+				return nil, fmt.Errorf("error parsing X509 certificate/key pair: %v", err)
```
philpennock, line 4748: For human operator debugging of their configs, both this one and the next could usefully gain `i+1`  indexing of the array, something like:

```go
return nil, fmt.Errorf("error parsing X509 certificate/key pair %d/%d: %v", i+1, len(tc.Certificates), err)
```

philpennock (approved):

server/opts.go
```diff
@@ -4381,10 +4388,46 @@ func parseTLS(v interface{}, isClientCtx bool) (t *TLSConfigOpts, retErr error)
...
+					switch k {
+					case "cert_file":
+						certPair.CertFile = file
+					case "key_file":
+						certPair.KeyFile = file
+					default:
+						return nil, &configErr{tk, fmt.Sprintf("error parsing tls certs config, unknown field %q", k)}
+					}
+				}
+				if certPair.CertFile == "" || certPair.KeyFile == "" {
```
derekcollison, line 4419: Use _EMPTY_ vs ""

derekcollison (approved):
LGTM - Minor nit

server/opts.go
wallyqs: good idea, I applied this change

PR #4890, opened 2023-12-15
[IMPROVED] Detect corrupt psim subjects during recovery of index.db.

derekcollison:
We had a bug that could cause internal memory corruption of psim subjects/keys in filestore. This could make it to disk via index.db. This checks subjects for corruption on reading in index.db.


wallyqs (approved):
LGTM

PR #4895, opened 2023-12-18
[IMPROVED] Do not allow large memory buildup in the apply queue for NRGs during startup.

derekcollison:
Previously we processed this inline with startRaftNode. Normally this was fine, however if the existing WAL entries were very large, this would consume lots of memory since the upper state machine will not be running. 

Now we will track how much memory we have been building up in the apply queue, and if it exceeds our threshold we will pause the apply queue and resume from the run() go routine.


neilalexander (approved):
LGTM

PR #4898, opened 2023-12-19
[IMPROVED] Only call into stream update() iff the config has changed.

derekcollison:


neilalexander (approved):
LGTM

PR #4901, opened 2023-12-19
Use `fastrand` instead of `prand`s

neilalexander:
In order to avoid top-level lock contention in the `math/rand` package, we were previously initialising our own `prand` sources in various structs. However these still needed to be protected by locks and they also just weren't that fast.

This PR removes those `prand`s and replaces those calls with calls to the Go runtime's `fastrand`, which is a couple hundred times faster and is, more importantly, lock-free:
```
goos: darwin
goarch: arm64
pkg: github.com/nats-io/nats-server/v2/server/fastrand
BenchmarkFastRand32
BenchmarkFastRand32-10          1000000000               0.2564 ns/op
BenchmarkFastRand64
BenchmarkFastRand64-10          1000000000               0.5019 ns/op
BenchmarkDefaultRand32
BenchmarkDefaultRand32-10       11997840               104.4 ns/op
BenchmarkDefaultRand64
BenchmarkDefaultRand64-10       11745747               105.8 ns/op
PASS
ok      github.com/nats-io/nats-server/v2/server/fastrand       3.768s
```

### DIFF
--- a/internal/fastrand/LICENSE
+++ b/internal/fastrand/LICENSE
@@ -1,0 +1,27 @@
+Copyright (c) 2011 The LevelDB-Go Authors. All rights reserved.
+
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions are
+met:
+
+   * Redistributions of source code must retain the above copyright
+notice, this list of conditions and the following disclaimer.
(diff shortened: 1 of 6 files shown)


derekcollison (approved):
LGTM

PR #4903, opened 2023-12-19
[FIXED] Fix unsupported MQTT topic characters error message

tomhollingworth:
Previously ' ' and '.' were not supported for MQTT topics until [#4243](https://github.com/nats-io/nats-server/pull/4243) brought in support for '.'. However the error message for unsupported characters appears to not have been updated. This simply updates the error message.

Had an example of a MQTT client trying to connect to `spBv1.0/Field Devices` and the error message indicated that both '.' and ' '  were unsupported which was misleading as only ' '  is unsupported.

Signed-off-by: Tom Hollingworth <tom.hollingworth@spruiktec.com>


derekcollison (approved):
LGTM - Thanks

PR #4905, opened 2023-12-20
[IMPROVED] Make sure to flush cache during PurgeEx to avoid memory bloat.

derekcollison:


wallyqs (approved):
LGTM

PR #4909, opened 2023-12-24
[FIXED] Fixes for JetStream accounting for replicas. 

derekcollison:
Also added in reporting for account info for reserved storage.


wallyqs (approved):
LGTM

PR #4910, opened 2023-12-25
[IMPROVED] Return account not enabled for JetStream when accessing via system account

derekcollison:


wallyqs (approved):
LGTM

PR #4915, opened 2023-12-29
Fix return value in func

korovindenis:
**Changes Proposed**
This pull request introduces a modification to the GetStatusAssertionStr

This change ensures that if no suitable value is found in the mapping (StatusAssertionIntToVal), the function will now default to returning "unknown" (from ocsp.Unknown).

**Motivation**
The motivation behind this change is to enhance the robustness of the GetStatus Assertion Str function. Previously, if the provided status assertion value did not exist in the map, the function return ocsp.Good.

**Unit Tests**
Unit tests have been added to cover various scenarios, including valid status assertion values for "good," "revoked," and "unknown," as well as an invalid status assertion value.

Please review this pull request 

Signed-off-by: Denis Korovin @korovindenis

### DIFF
--- a/server/certidp/certidp.go
+++ b/server/certidp/certidp.go
@@ -53,7 +53,13 @@ var (
 )
 
 func GetStatusAssertionStr(sa int) string {
-	return StatusAssertionValToStr[StatusAssertionIntToVal[sa]]
+	v, ok := StatusAssertionIntToVal[sa]
+	if !ok {
+		// set unknown as fallback
+		v = ocsp.Unknown
+	}
+
+	return StatusAssertionValToStr[v]
 }
 
 func (sa StatusAssertion) MarshalJSON() ([]byte, error) {
--- a/server/certidp/certidp.go
+++ b/server/certidp/certidp.go
@@ -54,12 +54,10 @@ var (
 
 // GetStatusAssertionStr returns the corresponding string representation of the StatusAssertion.
 func GetStatusAssertionStr(sa int) string {
-	/*
-	 If the provided status assertion value is not found in the map (StatusAssertionIntToVal),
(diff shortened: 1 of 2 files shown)


server/certidp/certidp.go
```diff
@@ -53,7 +53,13 @@ var (
 )
 
 func GetStatusAssertionStr(sa int) string {
-	return StatusAssertionValToStr[StatusAssertionIntToVal[sa]]
+	v, ok := StatusAssertionIntToVal[sa]
+	if !ok {
+		// set unknown as fallback
```
neilalexander, line 58: It would be worth expanding this comment here to explain that this is necessary because `Good` is the default `iota` value (https://pkg.go.dev/golang.org/x/crypto/ocsp#pkg-constants) and that we don't want to default to `Good` if there's no map entry. 

For that matter, it's probably worth noting that in `MarshalJSON` and `UnmarshalJSON` too.

korovindenis: Okay, I have added comments to the functions.

server/certidp/certidp.go
```diff
@@ -52,7 +52,14 @@ var (
 	}
 )
 
+// GetStatusAssertionStr returns the corresponding string representation of the StatusAssertion.
 func GetStatusAssertionStr(sa int) string {
+	/*
```
derekcollison, line 57: Small nit but prefer // comments for consistency with the rest of the codebase.

derekcollison (commented):
Small nit on comment type but otherwise LGTM

server/certidp/certidp.go
korovindenis: Ok, fixed it

derekcollison (approved):
LGTM

PR #4916, opened 2023-12-29
[IMPROVED] Allow fss metadata to not expire when msg cache expires.

derekcollison:
We will hold onto fss data to prevent large blocks being loaded into memory during scans like PurgeEx.
Also we now track both forward and backward linear access for expiring msg caches.


wallyqs (approved):
LGTM

PR #4925, opened 2024-01-04
Optimise metalayer snapshots

neilalexander:
This PR optimises the marshalling of metalayer snapshots with a few techniques:

* Simplifies the custom marshalers for a number of types that appear often in the snapshotted config, as these were using reflection unnecessarily
* Simplifies the custom unmarshallers for the same types, as these were string-building more than necessary
* Attempts to reduce the number of reallocations that might happen in `metaSnapshot()` by preallocating slices

### DIFF
--- a/server/consumer.go
+++ b/server/consumer.go
@@ -133,6 +133,12 @@ const (
 	actionCreateOrUpdateJSONString = `""`
 )
 
+var (
+	actionUpdateJSONBytes         = []byte(actionUpdateJSONString)
+	actionCreateJSONBytes         = []byte(actionCreateJSONString)
+	actionCreateOrUpdateJSONBytes = []byte(actionCreateOrUpdateJSONString)
+)
+
 func (a ConsumerAction) String() string {
 	switch a {
 	case ActionCreateOrUpdate:
@@ -148,11 +154,11 @@ func (a ConsumerAction) String() string {
 func (a ConsumerAction) MarshalJSON() ([]byte, error) {
 	switch a {
 	case ActionCreate:
-		return []byte(actionCreateJSONString), nil
+		return actionCreateJSONBytes, nil
 	case ActionUpdate:
-		return []byte(actionUpdateJSONString), nil
+		return actionUpdateJSONBytes, nil
 	case ActionCreateOrUpdate:
-		return []byte(actionCreateOrUpdateJSONString), nil
+		return actionCreateOrUpdateJSONBytes, nil
 	default:
 		return nil, fmt.Errorf("can not marshal %v", a)
 	}
--- a/server/consumer.go
+++ b/server/consumer.go
@@ -128,30 +128,31 @@ const (
 )
 
 const (
-	actionUpdateString         = "update"
-	actionCreateString         = "create"
-	actionCreateOrUpdateString = ""
+	actionUpdateJSONString         = `"update"`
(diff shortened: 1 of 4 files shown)


server/jetstream_cluster.go
```diff
@@ -1457,18 +1457,22 @@ func (js *jetStream) clusterStreamConfig(accName, streamName string) (StreamConf
...
+	streams := make([]writeableStreamAssignment, 0, nsa)
 	for _, asa := range cc.streams {
 		for _, sa := range asa {
 			wsa := writeableStreamAssignment{
-				Client:  sa.Client,
-				Created: sa.Created,
-				Config:  sa.Config,
-				Group:   sa.Group,
-				Sync:    sa.Sync,
+				Client:    sa.Client,
```
derekcollison, line 1470: I wonder if we need client here for snapshots? Both for streams and consumers?

derekcollison (commented):
Looking good. I left a comment on whether we should remove client info from snapshots for streams and consumers. We don't need it at that point and probably should not store it.

How are performance results looking?

server/jetstream_cluster.go
neilalexander: I think we do, because there are quite a few places that stream assignments check for the service account and other fields once rehydrated from the snapshot.

derekcollison: You are correct we need account, but not much else I believe.

derekcollison:
Let me know if you want a review now for merge.

neilalexander:
Think this is fine for review to merge for now, looking into the `ClientInfo` separately.

server/consumer.go
```diff
@@ -128,42 +128,43 @@ const (
...
 	}
-	return actionCreateOrUpdateString
+	return actionCreateOrUpdateJSONString
 }
+
 func (a ConsumerAction) MarshalJSON() ([]byte, error) {
 	switch a {
 	case ActionCreate:
-		return json.Marshal(actionCreateString)
+		return []byte(actionCreateJSONString), nil
```
derekcollison, line 151: These create copies, worth a global var?

server/store.go
```diff
@@ -548,23 +543,23 @@ func (st StorageType) String() string {
 func (st StorageType) MarshalJSON() ([]byte, error) {
 	switch st {
 	case MemoryStorage:
-		return json.Marshal(memoryStorageString)
+		return []byte(memoryStorageJSONString), nil
```
derekcollison, line 546: Same here.

server/store.go
```diff
@@ -667,19 +662,19 @@ func (p *DeliverPolicy) UnmarshalJSON(data []byte) error {
 func (p DeliverPolicy) MarshalJSON() ([]byte, error) {
 	switch p {
 	case DeliverAll:
-		return json.Marshal(deliverAllPolicyString)
+		return []byte(deliverAllPolicyJSONString), nil
```
derekcollison, line 665: And here, not sure it would make huge difference.

server/consumer.go
neilalexander: Just tested locally, it does reduce allocations very slightly but seems to have no meaningful impact on the time taken otherwise. If you think it's worth doing, I'll push the extra commit up!

derekcollison: Let's do it..

PR #4927, opened 2024-01-08
Don't allow stream or consumer snapshots before recovery completed

neilalexander:
Since `doSnapshot` in both cases can be called from a maintenance timer or on the quit signal, this may cause us to snapshot before recovery has completed.

### DIFF
--- a/server/jetstream_cluster.go
+++ b/server/jetstream_cluster.go
@@ -2197,9 +2197,13 @@ func (js *jetStream) monitorStream(mset *stream, sa *streamAssignment, sendSnaps
 	var lastState SimpleState
 	var lastSnapTime time.Time
 
+	// Don't allow the upper layer to install snapshots until we have
+	// fully recovered from disk.
+	isRecovering := true
+
 	// Should only to be called from leader.
 	doSnapshot := func() {
-		if mset == nil || isRestore || time.Since(lastSnapTime) < minSnapDelta {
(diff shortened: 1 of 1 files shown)


server/jetstream_cluster.go
```diff
@@ -2197,8 +2197,17 @@ func (js *jetStream) monitorStream(mset *stream, sa *streamAssignment, sendSnaps
...
 	var lastSnapTime time.Time
 
+	// Don't allow the upper layer to install snapshots until we have
+	// fully recovered from disk.
+	isRecovering := true
+
 	// Should only to be called from leader.
 	doSnapshot := func() {
+		// Only allow snapshots if we fully recovered from disk.
+		if isRecovering {
```
derekcollison, line 2207: Just add to below check IMO.

server/jetstream_cluster.go
```diff
@@ -4552,7 +4560,16 @@ func (js *jetStream) monitorConsumer(o *consumer, ca *consumerAssignment) {
 	var lastSnap []byte
 	var lastSnapTime time.Time
 
+	// Don't allow the upper layer to install snapshots until we have
+	// fully recovered from disk.
+	recovering := true
+
 	doSnapshot := func(force bool) {
+		// Only allow snapshots if we fully recovered from disk..
+		if recovering {
```
derekcollison, line 4569: Same just add to check below.

server/jetstream_cluster.go
neilalexander: Done!

PR #4929, opened 2024-01-09
[IMPROVED] Optimize catchups for replicas and mirrors with large interior deletes.

derekcollison:
Under scenarios where interior deletes were orders of magnitude larger than the msgs themselves, the system would need to transfer all interior deletes as skip msgs. This change adopts the ability to send DeleteRanges on streamSyncRequests, and for R1 mirrors to locally use the new store.SkipMsgs to optimize speed.

This CL also has a bug fix for memory store state with delete details, and several general improvements to the filestore impl.


wallyqs (approved):
LGTM

PR #4932, opened 2024-01-09
Only send snapshots on leader change once recovery has finished

neilalexander:
If the JetStream upper layer is still processing recovery items on the apply queue but the underlying Raft group becomes the leader, it's possible that the leader change notification can interleave with the apply queue, causing us to send a snapshot to the rest of the group before recovery has completed.

This delays sending snapshots on leader change notifications while recovery is taking place and will instead wait to do so after receiving the "nil guard" on the apply queue.


neilalexander:
Closing as this is probably not the right solution.

PR #4933, opened 2024-01-09
[IMPROVED] Avoid stream locks for certain operations that could block routes/gateways.

derekcollison:
Avoid needing the lock for mset subscriptions to avoid possibly blocking routes/gateways.
Also increase timeout for consumer creation for sources since might take a bit longer.


wallyqs (approved):
LGTM

PR #4935, opened 2024-01-10
Improve finding the last sourced message for each source of a stream

jnmoyne:
Improves how the sequence number of the last sourced message is found for all the stream's sources (in a single pass rather that individual passes for each source).

### DIFF
--- a/server/stream.go
+++ b/server/stream.go
@@ -1773,6 +1773,8 @@ func (mset *stream) updateWithAdvisory(config *StreamConfig, sendAdvisory bool) 
 		// Check for Sources.
 		if len(cfg.Sources) > 0 || len(ocfg.Sources) > 0 {
 			currentIName := make(map[string]struct{})
+			needsStartingSeqNum := make(map[string]struct{})
+
 			for _, s := range ocfg.Sources {
 				currentIName[s.iname] = struct{}{}
 			}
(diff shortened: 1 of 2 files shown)


server/stream.go
```diff
@@ -1806,18 +1808,21 @@ func (mset *stream) updateWithAdvisory(config *StreamConfig, sendAdvisory bool)
...
 				} else {
 					// source already exists
 					delete(currentIName, s.iname)
 				}
 			}
-			// What is left in cuurentIName needs to be deleted.
+			// What is left in curentIName needs to be deleted.
 			for iName := range currentIName {
 				mset.cancelSourceConsumer(iName)
 				delete(mset.sources, iName)
```
derekcollison, line 1820: Should we delete from needsStartingSeqNum here?

derekcollison (approved):
LGTM - Check my one comment.

server/stream.go
jnmoyne: No need, it's not in needsStartingSeqNum. currentIName is initially populated with all the sources and here we delete if not needed, while needsStartingSeqNum starts empty and gets added only if needed.

derekcollison:
You have some basic go fmt issues etc you need to fix..

derekcollison:
`server/stream.go:1823: File is not `gofmt`-ed with `-s` (gofmt)
			for iname, _ := range needsStartingSeqNum {`

server/stream.go
```diff
@@ -1806,18 +1808,25 @@ func (mset *stream) updateWithAdvisory(config *StreamConfig, sendAdvisory bool)
...
-					mset.setStartingSequenceForSource(s.iname, s.External)
-					mset.setSourceConsumer(s.iname, si.sseq+1, time.Time{})
+					needsStartingSeqNum[s.iname] = struct{}{}
 				} else {
 					// source already exists
 					delete(currentIName, s.iname)
 				}
 			}
-			// What is left in cuurentIName needs to be deleted.
+			// What is left in curentIName needs to be deleted.
```
derekcollison, line 1817: Two r's

server/stream.go
```diff
@@ -3360,10 +3374,24 @@ func (mset *stream) setStartingSequenceForSource(iName string, external *Externa
...
+
+		if _, ok := iNames[indexName]; ok {
+			si := mset.sources[indexName]
 			si.sseq = sseq
 			si.dseq = 0
-			return
+			delete(iNames, indexName)
+		} else if indexName == _EMPTY_ && streamName != _EMPTY_ {
+			for iName := range iNames {
+				if streamName == mset.sources[iName].name || (mset.streamSource(iName).External != nil && streamName == mset.sources[iName].name+":"+getHash(mset.streamSource(iName).External.ApiPrefix)) {
```
derekcollison, line 3385: Break this line up I think.

server/stream.go
jnmoyne: Corrected

server/stream.go

PR #4936, opened 2024-01-10
Add a pre-commit hook for the repo

philpennock:
This is Derek's list of desired checks, but wrapped up with update policy checks etc to be more appropriate for a git-hook which is used in an repository with commits from various places.

### DIFF
--- a/scripts/githook.pre-commit
+++ b/scripts/githook.pre-commit
@@ -24,9 +24,6 @@ declare -i failures=0
 warn() { stderr "$@"; failures+=1; }
 have_command() { command -v "$1" >/dev/null 2>&1; }
 
-## FIXME: check for diffs HOOK/REPO
-## FIXME: ensure in top-dir
-
 for c in git go; do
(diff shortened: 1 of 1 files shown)


wallyqs (approved):
LGTM

neilalexander:
Pretty sure all of these are already covered by our `golangci-lint` config, which should integrate with the Go language server in compatible editors and be runnable by hand with `golangci-lint run` if not. The CI pipeline also does this in the compile & test step. For consistency we probably should be using that.

kozlovic:
I would agree with @neilalexander in that we are already running checks in CI and one should not make direct commits to the `main` branch anyway, so errors should be caught by the developer when submitting their PR.

derekcollison:
Let's rework, but no rush.

philpennock:
I think there's no need to keep this PR around.  This one was just to add the checks as requested, but done in a way safe for use with a repo with various committers.  If the goal is to switch to more standard checks, then that should be its own PR.

Issue #4934, opened 2024-01-09
Changes to subjects of interest due to subject mappings amendments are not propagated to leaf nodes


vinopm:
### Observed behavior

Subjects of interest which are communicated to leaf nodes are not updated accordingly when subject mappings are updated on the main NATS server.

Assume we have a main NATS node and a leaf node. Assume we have a publisher connected to the leaf node, who is publishing to 'source1'. Assume we also have a subscriber who is connected to the main NATS node and subscribing to 'target'.
e.g. if the mapping is set up to be `source1` -> `target`, and we then change the mapping to `source2` -> `target`.
A subscriber (connected to main NATS node) on 'target' should start receiving data from a publisher (connected) to leaf node that is publishing to 'source2'.

Starting up a subscriber (connected to the main NATS node) on the subject 'source2' (instead of 'target') forces data to start streaming from the publisher on the leaf node to the subscriber who is listening for 'target' subject data. No data is received on the subscriber to 'source2' as expected because the NATS server is mapping the messages from source2 to target. However, it appears that because there is a subscriber to 'source2', the data on the leaf node is actually published out (due to an interested party in this subject). When we remove the subscriber to 'source2', the subscriber to 'target' stops receiving data.

### Expected behavior

When the subject mapping in the server config is changed, and a reload signal is issued, the expected behaviour should be to inform the leaf nodes of the updated subjects of interest.

e.g. if the mapping is set up to be `source1` -> `target`, and we change the mapping to `source2` -> `target`.
A subscriber (connected to main NATS node) on target should start receiving data from a publisher (connected) to leaf node that is publishing to source2.

### Server and client version

Issue exists on v2.9.24, v2.10.7.

### Host environment

Container optimised OS
Repeated test locally on Mac OS 13.2.1.

### Steps to reproduce

1. Create a `server.conf` file:
```
port: 4222
server_name: test-server

debug: true
trace: true

leafnodes {
  port: 7422
}

mappings = {
	"source1": "target",
}
```

2. Create a `leaf.conf` file:
```
port: 12543
server_name: test-leaf

debug: true
trace: true

leafnodes {
	remotes = [
	    {
	      url: "nats://localhost:7422"
	    },
	]
}
```

3. Start up a nats server: `nats-server -c server.conf`
4. Start up a nats leaf node: `nats-server -c leaf.conf`
5. Verify existing subject mapping works end to end by
[...]

PR #4937, opened 2024-01-10
[FIXED] Reloads that contain mapping updates for global account with leafnodes.

derekcollison:
We should update the leafnode connection's smaps correctly.

Resolves: #4934 


wallyqs (approved):
LGTM

Issue #3196, opened 2022-06-16
sample config with leafnode and nkeys


TomasVojacek:
Is it possible to add to docs sample config with leaf nodes secured by nkey?
I have found ugly solution using credential file but it require pretense of JWT token. 
I kept JWT token from docs and put correct USER NKEY SEED there.
Is there better solution?

on leafnode `nats.conf`
```
leafnodes {
    remotes = [
        {
           url: "nats-leaf://@server_url:7422"
          account: someuser
         credentials: "server.creds"
        }
    ]
}
```
where `server.creds`
is copy of example creds from docs
```
**** this part is keept as it is in docs ****
-----BEGIN NATS USER JWT-----
eyJ0eXAiOiJqd3QiLCJhbGciOiJlZDI1NTE5In0.eyJqdGkiOiJUVlNNTEtTWkJBN01VWDNYQUxNUVQzTjRISUw1UkZGQU9YNUtaUFhEU0oyWlAzNkVMNVJBIiwiaWF0IjoxNTU4MDQ1NTYyLCJpc3MiOiJBQlZTQk0zVTQ1REdZRVVFQ0tYUVM3QkVOSFdHN0tGUVVEUlRFSEFKQVNPUlBWV0JaNEhPSUtDSCIsIm5hbWUiOiJvbWVnYSIsInN1YiI6IlVEWEIyVk1MWFBBU0FKN1pEVEtZTlE3UU9DRldTR0I0Rk9NWVFRMjVIUVdTQUY3WlFKRUJTUVNXIiwidHlwZSI6InVzZXIiLCJuYXRzIjp7InB1YiI6e30sInN1YiI6e319fQ.6TQ2ilCDb6m2ZDiJuj_D_OePGXFyN3Ap2DEm3ipcU5AhrWrNvneJryWrpgi_yuVWKo1UoD5s8bxlmwypWVGFAA
------END NATS USER JWT------

************************* IMPORTANT *************************
NKEY Seed printed below can be used to sign and prove identity.
NKEYs are sensitive and should be treated as secrets.

-----BEGIN USER NKEY SEED-----
REAL_USER_PRIVATE_KEY
------END USER NKEY SEED------

*************************************************************
```

Server config for reference:
```
leafnodes {
  port: 7422
  authorization: {
        users: [ {user: UDXU4RCSJNZOIQHZNWXHXORDPRTGNJAHAHFRGZNEEJCPQTT2M7NLCNF4,  account: someuser}]
  }
}
```

[...]

PR #4940, opened 2024-01-10
[FIXED] Allow leafnode authorization to support nkeys.

derekcollison:
Resolves: #3196 

### DIFF
--- a/server/opts.go
+++ b/server/opts.go
@@ -147,6 +147,7 @@ type LeafNodeOpts struct {
 	Port              int           `json:"port,omitempty"`
 	Username          string        `json:"-"`
 	Password          string        `json:"-"`
+	Nkey              string        `json:"-"`
 	Account           string        `json:"-"`
 	Users             []*User       `json:"-"`
 	AuthTimeout       float64       `json:"auth_timeout,omitempty"`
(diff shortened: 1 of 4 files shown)


server/opts.go
```diff
@@ -2351,6 +2355,12 @@ func parseLeafAuthorization(v interface{}, errors *[]error, warnings *[]error) (
 			auth.user = mv.(string)
 		case "pass", "password":
 			auth.pass = mv.(string)
+		case "nkey":
+			nk := mv.(string)
```
neilalexander, line 2359: Is this guaranteed to be `string` or do we want an `nk, ok :=` to avoid a possible panic?

server/opts.go
```diff
@@ -2496,7 +2506,24 @@ func parseRemoteLeafNodes(v interface{}, errors *[]error, warnings *[]error) ([]
...
 					continue
 				}
+				// Can't have both creds and nkey
+				if remote.Nkey != _EMPTY_ {
+					*errors = append(*errors, &configErr{tk, "Remote leafnode can not have both creds and nkey defined"})
+					continue
+				}
 				remote.Credentials = p
+			case "nkey", "seed":
+				nk := v.(string)
```
neilalexander, line 2516: Same here

neilalexander (approved):
LGTM

server/opts.go
derekcollison: Not guaranteed but like above do not test for failed cast, so I think its ok.

PR #4942, opened 2024-01-10
Formalize claims update response

ripienaar:
The previous response was attempting to look like a typical Api response but in the process invented its own way of representing error and more making it really hard for users to figure out how this is supposed to look.

Added a struct that retains the non standard behavior while working on CLI tooling


derekcollison (approved):
LGTM

PR #4947, opened 2024-01-10
Fix flapping in TestJetStreamMirrorBasics

jnmoyne:
It can take up to a couple of seconds for new source consumers to get established and all the message getting sourced in the test, the final tests had a timeout of 2 seconds that could sometimes be reached in Travis. Relaxing that timeout to something more generous fixes the flapping


wallyqs (approved):
LGTM

derekcollison (approved):
LGTM

PR #4950, opened 2024-01-11
[FIXED] setStartingSequenceForSources could panic in some specific scenarios

jnmoyne:
Fixed setStartingSequenceForSources line using the wrong variable name

### DIFF
--- a/server/stream.go
+++ b/server/stream.go
@@ -3382,11 +3382,12 @@ func (mset *stream) setStartingSequenceForSources(iNames map[string]struct{}) {
 			delete(iNames, indexName)
 		} else if indexName == _EMPTY_ && streamName != _EMPTY_ {
 			for iName := range iNames {
-				if streamName == mset.sources[iName].name ||
-					(mset.streamSource(iName).External != nil && streamName == mset.sources[iName].name+":"+getHash(mset.streamSource(iName).External.ApiPrefix)) {
-					si := mset.sources[indexName]
+				// TODO streamSource is a linear walk, to optimize later
+				if si := mset.sources[iName]; si != nil && streamName == si.name ||
+					(mset.streamSource(iName).External != nil && streamName == si.name+":"+getHash(mset.streamSource(iName).External.ApiPrefix)) {
 					si.sseq = sseq
 					si.dseq = 0
+					delete(iNames, iName)
 					break
 				}
 			}


server/stream.go
```diff
@@ -3384,7 +3384,7 @@ func (mset *stream) setStartingSequenceForSources(iNames map[string]struct{}) {
 			for iName := range iNames {
 				if streamName == mset.sources[iName].name ||
 					(mset.streamSource(iName).External != nil && streamName == mset.sources[iName].name+":"+getHash(mset.streamSource(iName).External.ApiPrefix)) {
-					si := mset.sources[indexName]
+					si := mset.sources[iName]
 					si.sseq = sseq
```
derekcollison, line 3388: Please put a guard around this regardless.

derekcollison (changes requested):

server/stream.go
derekcollison:
```go
if si := mset.sources[iName]; si != nil {
}
```

server/stream.go
```diff
@@ -3382,9 +3382,9 @@ func (mset *stream) setStartingSequenceForSources(iNames map[string]struct{}) {
 			delete(iNames, indexName)
 		} else if indexName == _EMPTY_ && streamName != _EMPTY_ {
 			for iName := range iNames {
-				if streamName == mset.sources[iName].name ||
-					(mset.streamSource(iName).External != nil && streamName == mset.sources[iName].name+":"+getHash(mset.streamSource(iName).External.ApiPrefix)) {
-					si := mset.sources[indexName]
+				if mset.sources[iName] != nil && (streamName == mset.sources[iName].name ||
```
derekcollison, line 3385: Pull this into an si first since you reference it several times.

derekcollison: so something like..

```go
if si := mset.sources[iName]; si != nil && (si.External != nil && streamName == si.name+":"+getHash(si.External.ApiPrefix)) {
    si.sseq = sseq
    si.dseq = 0
}
```

derekcollison (approved):
LGTM

server/stream.go
```diff
@@ -3382,9 +3382,8 @@ func (mset *stream) setStartingSequenceForSources(iNames map[string]struct{}) {
 			delete(iNames, indexName)
 		} else if indexName == _EMPTY_ && streamName != _EMPTY_ {
 			for iName := range iNames {
-				if streamName == mset.sources[iName].name ||
-					(mset.streamSource(iName).External != nil && streamName == mset.sources[iName].name+":"+getHash(mset.streamSource(iName).External.ApiPrefix)) {
-					si := mset.sources[indexName]
+				if si := mset.sources[iName]; si != nil && (streamName == si.name ||
+					(mset.streamSource(iName).External != nil && streamName == si.name+":"+getHash(mset.streamSource(iName).External.ApiPrefix))) {
```
derekcollison, line 3386: Actually this should be si.External != nil

derekcollison: Same with last one, I pasted exact code to use.

derekcollison (changes requested):

server/stream.go
derekcollison: Actually I am wrong here, but we should only call mset.streamSource(iName) once since its a linear walk.

derekcollison (approved):
LGTM

PR #4958, opened 2024-01-15
Fix create consumer's 'ActionCreate' functionality

jnmoyne:
Fix create consumer's 'ActionCreate' functionality and adds testing of the functionality to the TestJetStreamConsumerWithNameAndDurable test


jnmoyne:
Apologies had made the wrong assumption on the 'create' behavior, I see now that what I need is an insert action instead.

PR #4965, opened 2024-01-17
Don't unconditionally send consumer snapshot when becoming leader

neilalexander:
Feasibly a Raft node that is keeping up with its peers can become a leader regardless of whether or not the upper layer is caught up with its applies. Since the leader change notification to the upper state machine could interleave with the apply queue, then unconditionally sending a snapshot as a new append entry on consumer leader transitions could cause the rest of the group to revert to an earlier consumer state.

### DIFF
--- a/server/jetstream_cluster.go
+++ b/server/jetstream_cluster.go
@@ -4660,16 +4660,6 @@ func (js *jetStream) monitorConsumer(o *consumer, ca *consumerAssignment) {
 				js.setConsumerAssignmentRecovering(ca)
 			}
 
-			// Synchronize everyone to our state.
-			if isLeader && n != nil {
-				// Only send out if we have state.
-				if _, _, applied := n.Progress(); applied > 0 {
-					if snap, err := o.store.EncodedState(); err == nil {
-						n.SendSnapshot(snap)
-					}
-				}
-			}
-
 			// Process the change.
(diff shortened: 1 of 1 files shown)


server/jetstream_cluster.go
```diff
@@ -4660,16 +4660,6 @@ func (js *jetStream) monitorConsumer(o *consumer, ca *consumerAssignment) {
 				js.setConsumerAssignmentRecovering(ca)
 			}
 
-			// Synchronize everyone to our state.
```
derekcollison, line 4663: Should we remove all together or adjust to make sure our upper state is correct?

It is there for a reason :)

neilalexander: Do you remember what that reason was? I can't find anything written down as to why, the best I can find is from the commit message in cc77d66 which says:

> Also sync other consumers when taking over as leader but no need to process snapshots when we are in fact the leader.

As far as I can tell everything should work without this because the Raft machinery would be keeping the `ConsumerStore` up-to-date on the replicas in the usual way anyway, whereas sending snapshots from the upper layer directly opens up a potential for a race between the committed Raft node state and the applied upper layer state, but I'm not sure if I'm missing something here.

derekcollison: We were experiencing state skew, and I wanted us synched. But could have been from a prior bug that has since been corrected.

derekcollison (approved):
LGTM

PR #4968, opened 2024-01-18
Adds testing for backwards compatibility of pre 2.10 existing stream sources

jnmoyne:
Adds testing for backwards compatibility of pre 2.10 existing stream sources which used a previous message header format, the header is used to find the starting point of the source's consumer on leadership changes or stream config updates when sources are added.


derekcollison (approved):
LGTM - Thanks!

PR #4969, opened 2024-01-18
[IMPROVED] Do not load blocks when calling LoadNextMsg for the first time after restart or long inactivity

derekcollison:
If LoadNextMsg was called with start sequence == FirstSeq, we would walk all blocks, which if fss was loaded we could skip fairly quickly. However on a restart or no activity past a sync interval this could cause memory blocks to be loaded in just to be skipped.

https://github.com/nats-io/nats-server/discussions/4906


neilalexander (approved):
LGTM

PR #4973, opened 2024-01-18
Adds machine readable conninfo durations

ripienaar:
The server use a home-grown encoding format for durations that's hard to use programatically while pleasing to look at - see myUptime() function.

This adds matching time.Duration based ones for machines.

### DIFF
--- a/server/monitor.go
+++ b/server/monitor.go
@@ -116,8 +116,11 @@ type ConnInfo struct {
 	Stop           *time.Time     `json:"stop,omitempty"`
 	Reason         string         `json:"reason,omitempty"`
 	RTT            string         `json:"rtt,omitempty"`
+	RTTNanos       time.Duration  `json:"rtt_nanos,omitempty"`
 	Uptime         string         `json:"uptime"`
+	UptimeNanos    time.Duration  `json:"uptime_nanos"`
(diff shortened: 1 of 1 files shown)


derekcollison (commented):
Should this be opt-in? Should we consider the ability to form shapes by requesting only certain fields?

I am growing concerned watching SCP on NGS deal with the number of connections it already has, and thinking when then goes up multiple orders of magnitude.

Not against making it easier but concerned on longer term impact.

ripienaar:
Opt-in doesnt seem like a nice granular approach but for sure being able to extract subsets of fields would be handy.

derekcollison:
The string we use is directly parseable to a time.Duration, but that is in Go of course.

ripienaar:
> The string we use is directly parseable to a time.Duration, but that is in Go of course.

if thats an API promise then I'll go with that :)

derekcollison:
I think the only place we deviate is in uptime since parse only supports up to hours IIRC. Would need to double check.

derekcollison:
I think we should close if clients can parse ok, as to not add additional weight here to the conninfo.

ripienaar:
OK I will verify it parses correctly then close

neilalexander:
Closing as old/stale.

PR #4974, opened 2024-01-18
stree: Reduce allocations in `iter` and `match`

neilalexander:
The `_pre` and `cparts` copies are more often than not unnecessary and result in potentially gigabytes of allocations which can slow down functions like `NumPending`.

Passing `pre` and `nparts` down recursively without copies is usually safe, even where there are appends in those, because there is no concurrency and those appends will not modify the slice length back at the callsite. Instead we'll only reallocate either when `append` does so naturally (we've ran out of capacity) or when we know specifically that we're modifying something in-place (like in `matchParts`).

This improves performance of things like `NumPending` with high subject cardinality.


derekcollison (approved):
LGTM! 

Thanks, well done!

PR #4976, opened 2024-01-19
Faster payload generation in benchmarks

mprimi:
Two commits (for ease of review, but they can be squashed in merge):

1. Stop using the static (thread-safe) RNG provided by `math/rand`, the CoreFanIn benchmark is very slow due to the fact that producers are piling up to use a single shared instance
2. Switch all benchmarks to a faster message payload generation (which reuses the previous payload with minor mutations).

Signed-off-by: Your Name <marco@synadia.com>

### DIFF
--- a/server/core_benchmarks_test.go
+++ b/server/core_benchmarks_test.go
@@ -14,17 +14,18 @@
 package server
 
 import (
-	"crypto/rand"
 	"crypto/tls"
 	"errors"
 	"fmt"
-	mrand "math/rand"
+	"math"
+	"math/rand"
 	"os"
 	"strconv"
 	"sync"
 	"testing"
 	"time"
 
+	"github.com/nats-io/nats-server/v2/internal/fastrand"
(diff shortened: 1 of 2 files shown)


derekcollison (commented):
We have a faster one, check w/ @neilalexander

derekcollison (commented):
You don't need the ran.New calls anymore I believe..

mprimi:
> You don't need the ran.New calls anymore I believe..

@derekcollison the remaining calls to `rng.New` are outside the benchmark hot-path. For example filling up a stream before the benchmark begins. So even if they're a little slow, it doesn't matter, they don't count towards the benchmark duration.

Everywhere applicable is now using `fastrand`.

mprimi:
Travis is red but the failed test is unrelated to this change:

 > `FAIL: TestOCSPPeerNextUpdateUnset`

derekcollison (approved):
LGTM

PR #4977, opened 2024-01-19
stree: Reduce heap escapes in `iter`

neilalexander:
When using `make(x, y, z)`, there is a heap escape due to the non-constant size/capacity. Try to stay on the stack instead, reducing GC pressure.


derekcollison (approved):
LGTM

PR #4983, opened 2024-01-22
Use stree for per-subject tracking in memory store

neilalexander:
This updates the memory store to use the new subject tree from #4960 for per-subject tracking.


derekcollison (approved):
LGTM

I have not wanted to update fss under filsstore because they are usually small but great to be here!

PR #4985, opened 2024-01-23
Update `cacheAlreadyLoaded` to exclude deletions

neilalexander:
The `mb.msgs` count doesn't include deletes, so by including them here, we fail the `numEntries` check which causes us to reload the block from disk when we don't need to.


derekcollison (commented):
The cache.idx can not have empty slots, so for things that are deleted we will place markers in the idx for the sequence to mapping. So for the full cache to be loaded I believe the code is correct.

Do you have a test that shows different or was this just a hunch?

neilalexander:
Closing as have identified the correct fix, will open separately.

PR #4995, opened 2024-01-24
NRG: Log healthy status when no longer falling behind

neilalexander:
This makes it clear in the logs that the falling behind condition cleared up.

### DIFF
--- a/server/raft.go
+++ b/server/raft.go
@@ -154,11 +154,12 @@ type raft struct {
 	llqrt  time.Time   // Last quorum lost time
 	lsut   time.Time   // Last scale-up time
 
-	term    uint64 // The current vote term
-	pterm   uint64 // Previous term from the last snapshot
-	pindex  uint64 // Previous index from the last snapshot
-	commit  uint64 // Sequence number of the most recent commit
-	applied uint64 // Sequence number of the most recently applied commit
+	term     uint64 // The current vote term
+	pterm    uint64 // Previous term from the last snapshot
+	pindex   uint64 // Previous index from the last snapshot
+	commit   uint64 // Sequence number of the most recent commit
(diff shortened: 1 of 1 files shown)


server/raft.go
```diff
@@ -1269,8 +1270,18 @@ func (n *raft) isCurrent(includeForwardProgress bool) bool {
 		return false
 	}
 
+	// If we were previously logging about falling behind, also log when the problem
+	// was cleared.
+	clearBehindState := func() {
+		if n.hcbehind {
+			n.warn("Health check OK, no longer falling behind")
```
ripienaar, line 1277: can we add details here about what asset it is? also on the falling behind log

neilalexander: The `n.warn` is already prefixing with the Raft ID & group, I don't think `*raft` actually knows the real name of the asset though.

ripienaar: ah of course, good (well unfortunately always logging the raft group names not the asset but ok its something)

server/raft.go
```diff
@@ -1269,8 +1270,18 @@ func (n *raft) isCurrent(includeForwardProgress bool) bool {
...
+			n.hcbehind = false
+		}
+	}
+
 	// Make sure we are the leader or we know we have heard from the leader recently.
 	if n.State() == Leader {
+		clearBehindState()
```
derekcollison, line 1284: This not only applicable to leaders no?

neilalexander: I put this here just to cover the case of if you became leader in that time, to ensure the behind state gets cleared.

derekcollison: Gotcha..

derekcollison (approved):
LGTM

PR #5001, opened 2024-01-25
Wait for recovery to complete before sending snapshot on stream scale-up from R1

neilalexander:
This PR ensures that we don't try to generate and send a snapshot to catch up followers on a stream scale-up from an R1 until recovery has completed.


derekcollison (approved):
LGTM

PR #5004, opened 2024-01-26
Optimise filestore `writeFullState` for high subject cardinality

neilalexander:
This PR makes two changes to `writeFullState`:

1. Use `Match(">", ...)` instead of `Iter(...)` as it is faster and results in less heap escapes
2. Don't allocate a 4K buffer on the stack if it's obvious that we won't use it

### DIFF
--- a/server/filestore.go
+++ b/server/filestore.go
@@ -7329,11 +7329,12 @@ func (fs *fileStore) writeFullState() error {
 		binary.MaxVarintLen64 + 8 + 8 // last index + record checksum + full state checksum
 
 	// Do 4k on stack if possible.
-	var raw [4 * 1024]byte
+	const ssz = 4 * 1024
 	var buf []byte
 
-	if sz <= cap(raw) {
-		buf, sz = raw[0:2:cap(raw)], cap(raw)
+	if sz <= ssz {
+		var _buf [ssz]byte
+		buf, sz = _buf[0:2:ssz], ssz
 	} else {
 		buf = make([]byte, hdrLen, sz)
 	}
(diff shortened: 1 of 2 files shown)


server/filestore.go
```diff
@@ -7329,11 +7329,12 @@ func (fs *fileStore) writeFullState() error {
...
 
 	// Do 4k on stack if possible.
-	var raw [4 * 1024]byte
+	const ssz = 4 * 1024
 	var buf []byte
 
-	if sz <= cap(raw) {
-		buf, sz = raw[0:2:cap(raw)], cap(raw)
+	if sz <= ssz {
+		var _buf [ssz]byte
```
derekcollison, line 7336: This seems like it would make runtime work since you are betting it does not grow stack unless on conditional but it would then need to escape the block properly. Being only 4k I think its fine on main stack, I like the make branch.

derekcollison (approved):
LGTM - One nit

server/filestore.go
neilalexander: I can put it back to being unconditional, but I did check it with escape analysis and the escape paths are the same (either via `.Seal()` or a realloc with `append`), regardless of whether it's inside the conditional block or not. Just seems like a little way to save the allocation from happening at all if we know from the get-go that we're over 4K.

derekcollison: ok if you checked.. Hopefully makes sense though that the runtime seems like it would need to do more work to not decrease stack leaving that block since access to that escapes the block but not the func(). Stack growth is free unless it splits the stack being very big.

PR #5005, opened 2024-01-26
Use AVL seqset to track stream pre-acks

neilalexander:
In principle this should align the memory footprint of pre-acks more closely with the number of consumers rather than the range of acked sequences, the latter of which can be unexpectedly large when the stream is handling a lot of traffic.

### DIFF
--- a/server/stream.go
+++ b/server/stream.go
@@ -32,6 +32,8 @@ import (
 	"sync/atomic"
 	"time"
 
+	"github.com/nats-io/nats-server/v2/server/avl"
+
 	"github.com/klauspost/compress/s2"
 	"github.com/nats-io/nuid"
 )
@@ -271,7 +273,7 @@ type stream struct {
 
 	// For non limits policy streams when they process an ack before the actual msg.
 	// Can happen in stretch clusters, multi-cloud, or during catchup for a restarted server.
-	preAcks map[uint64]map[*consumer]struct{}
+	preAcks map[*consumer]*avl.SequenceSet
 
 	// TODO(dlc) - Hide everything below behind two pointers.
 	// Clustered mode.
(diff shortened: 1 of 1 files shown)


wallyqs:
This improves the memory utilization quite a bit as it is, on a restart recovery from interest based streams that would consume 5GBs memory is now reduced to about 140MBs for me.

derekcollison:
Will review tomorrow. Might pull into 2.10.10, we will see.

server/stream.go
```diff
@@ -32,6 +32,8 @@ import (
 	"sync/atomic"
 	"time"
 
+	"github.com/nats-io/nats-server/v2/server/avl"
+
```
derekcollison, line 36: Remove this line.

server/stream.go
```diff
@@ -5513,15 +5516,30 @@ func (mset *stream) hasAllPreAcks(seq uint64, subj string) bool {
 // Check if we have all consumers pre-acked.
 // Write lock should be held.
 func (mset *stream) clearAllPreAcks(seq uint64) {
-	delete(mset.preAcks, seq)
+	for o, ss := range mset.preAcks {
```
derekcollison, line 5519: This is now O(n) vs O(1) depending on number of consumers. The interest policy usually has lower number of consumers but the general trend we have been observing is more and more consumers.

server/stream.go
```diff
@@ -5513,15 +5516,30 @@ func (mset *stream) hasAllPreAcks(seq uint64, subj string) bool {
...
 }
 
 // Clear all preAcks below floor.
 // Write lock should be held.
 func (mset *stream) clearAllPreAcksBelowFloor(floor uint64) {
-	for seq := range mset.preAcks {
-		if seq < floor {
-			delete(mset.preAcks, seq)
+	for o, ss := range mset.preAcks {
+		min, max := ss.MinMax()
```
derekcollison, line 5530: This is dynamically calculated on the fly and not maintained.

server/stream.go
```diff
@@ -5513,15 +5516,30 @@ func (mset *stream) hasAllPreAcks(seq uint64, subj string) bool {
...
+		if max < floor {
+			delete(mset.preAcks, o)
+			continue
+		}
+		if min > floor {
```
derekcollison, line 5535: This could be eliminated and just let the for loop on line 5538 take care of it. But with this and 5538 we do not cover min == floor, so something to resolve.

server/stream.go
neilalexander: Yes, not clear exactly what the right balance is here. Certainly a lot of memory is saved which prevents OOMs but does risk being marginally slower.

server/stream.go
neilalexander: The check before was `seq < floor` so this check might want fixing but 5538 should be fine?

server/stream.go
neilalexander: Good point, I wonder if this outweighs the benefit of the following two checks...

server/stream.go
derekcollison: Yes, but again preAck state was meant to be small. So better way to fix the exploding state space I think.

neilalexander:
Closing for now as this may make some hot paths worse.

PR #5006, opened 2024-01-26
Republish before interest retention checks on streams

### LINKED ISSUE #5003: Republish does not work for a stream with no consumers and Interest retention policy
### Observed behavior

Configuration:
streamA(source) -> streamB ---republish---> streamC1,C2,C3

Configured supercluster with jetstreams.
streamA is a source for streamB.
streamB has no consumers but republishes messages to streams streamC1,C2,C3.

If set retention policy for streamB to Interest then messages written to streamA are received by streamB but dropped and not republished to streams streamC1,C2,C3.
(With Limits retention policy on streamB everything works as expected).

### Expected behavior

If a stream sources messages from another stream, has no consumers, has retention policy Interest, republishes messages to other streams - republishing should work and messages should not be dropped before they are republished.

### Server and client version

embedded nats server v2.10.9

### Host environment

_No response_

### Steps to reproduce

_No response_

neilalexander:
This should fix #5003 by moving the republishing to before the interest checks in `processJetStreamMsg`. Added a unit test which adds a stream but no consumers, which fails on `main` but passes on this branch.

### DIFF
--- a/server/jetstream_test.go
+++ b/server/jetstream_test.go
@@ -17326,6 +17326,45 @@ func TestJetStreamStreamRepublishHeadersOnly(t *testing.T) {
 	}
 }
 
+func TestJetStreamStreamRepublishNoInterest(t *testing.T) {
+	s := RunBasicJetStreamServer(t)
+	defer s.Shutdown()
+
+	nc, js := jsClientConnect(t, s)
+	defer nc.Close()
+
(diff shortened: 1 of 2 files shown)


neilalexander:
Closing as this isn't right for sequencing/timestamps.

PR #5008, opened 2024-01-27
[FIXED] Acking a redelivered msg with more pending outstanding that should move the ack floor.

derekcollison:
A prior fix to deal with larger consumer delivery numbers due to redeliveries caused the ack floor calculations to not properly work when a gap was filled by an ack of a redelivered msg.

The ack floor would move and fix itself when the pending went to zero. This fixes the bug introduced and goes back to the pending sequence always being the original consumer delivery sequence.


wallyqs (approved):
LGTM

neilalexander (approved):
LGTM

PR #5009, opened 2024-01-27
[FIXED] Fix for a panic calculating record length for slotInfo when followed by dbit markers in idx.

derekcollison:
This was reported by @piotrpio when trying to do a secure erase of a message the was followed by dbit markers which occur  after compaction removes records from the raw msg block. The idx then expands via dbit markers for the idx slice.


neilalexander (approved):
LGTM

PR #5011, opened 2024-01-29
[FIXED] Memory growth on interest stream reset with high sequence numbers.

derekcollison:
If an interest stream has been reset with consumers that have very high ack floors, we could explode the pre-ack state.

The pre-ack state should be minimal, and in this case, when the stream's last sequence is below the ackfloor, we should not process pre-acks during startup.

Also fixed a bug that was not properly doing snapshots when the quick check state was a purged state.

### DIFF
--- a/server/consumer.go
+++ b/server/consumer.go
@@ -5334,32 +5334,41 @@ func (o *consumer) isMonitorRunning() bool {
 	return o.inMonitor
 }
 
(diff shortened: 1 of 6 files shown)


wallyqs (approved):
LGTM

neilalexander (approved):
LGTM

PR #5012, opened 2024-01-29
Avoid panic when reloading during LDM

neilalexander:
If a reload happened during the LDM grace period, it was possible for the `resetCh` to have been closed between the loader fetching the channel, releasing the lock and then sending to it. By making sure that closing the channel happens after `s.sys` is invalidated, we can avoid this panic.


derekcollison (approved):
LGTM

PR #5017, opened 2024-01-30
[CHANGED] MQTT RMS consumer: instead of deleting/creating, create a u…

levb:
Every time MQTT was initialized for an account on a server, it was re-creating the server-specific RMS consumer by deleting it and creating again, with the same name. @derek though it might be causing issues and suggested this change.

Now, always create a new consumer with a unique name, and a 5 minute inactivity threshold. Kept the code to delete the "legacy"-named consumer, even though it'd get executed at most once, errors ignored.

### DIFF
--- a/server/mqtt.go
+++ b/server/mqtt.go
@@ -1439,11 +1439,10 @@ func (s *Server) mqttCreateAccountSessionManager(acc *Account, quitCh chan struc
 		}
 	}
 
-	// Delete the old (legacy) consumer, from v2.10.10 and before.
+	// Opportunistically delete the old (legacy) consumer, from v2.10.10 and
+	// before. Ignore any errors that might arise.
 	rmLegacyDurName := mqttRetainedMsgsStreamName + "_" + jsa.id
-	if _, err := jsa.deleteConsumer(mqttRetainedMsgsStreamName, rmLegacyDurName); isErrorOtherThan(err, JSConsumerNotFoundErr) {
-		return nil, err
-	}
+	jsa.deleteConsumer(mqttRetainedMsgsStreamName, rmLegacyDurName)
 
 	// Using ephemeral consumer is too risky because if this server were to be
 	// disconnected from the rest for few seconds, then the leader would remove
--- a/server/mqtt.go
+++ b/server/mqtt.go
@@ -1439,26 +1439,32 @@ func (s *Server) mqttCreateAccountSessionManager(acc *Account, quitCh chan struc
 		}
 	}
 
+	// Delete the old (legacy) consumer, from v2.10.10 and before.
+	rmLegacyDurName := mqttRetainedMsgsStreamName + "_" + jsa.id
+	if _, err := jsa.deleteConsumer(mqttRetainedMsgsStreamName, rmLegacyDurName); isErrorOtherThan(err, JSConsumerNotFoundErr) {
+		return nil, err
+	}
+
 	// Using ephemeral consumer is too risky because if this server were to be
 	// disconnected from the rest for few seconds, then the leader would remove
 	// the consumer, so even after a reconnect, we would no longer receive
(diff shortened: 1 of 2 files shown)


server/mqtt.go
```diff
@@ -1439,26 +1439,32 @@ func (s *Server) mqttCreateAccountSessionManager(acc *Account, quitCh chan struc
 		}
 	}
 
+	// Delete the old (legacy) consumer, from v2.10.10 and before.
+	rmLegacyDurName := mqttRetainedMsgsStreamName + "_" + jsa.id
+	if _, err := jsa.deleteConsumer(mqttRetainedMsgsStreamName, rmLegacyDurName); isErrorOtherThan(err, JSConsumerNotFoundErr) {
```
derekcollison, line 1444: I don't think this should bubble up error.

server/mqtt_test.go
```diff
@@ -3026,8 +3026,10 @@ func TestMQTTCluster(t *testing.T) {
 				for _, sn := range streams {
 					cl.waitOnStreamLeader(globalAccountName, sn)
 				}
-				cl.waitOnConsumerLeader(globalAccountName, mqttRetainedMsgsStreamName, "$MQTT_rmsgs_esFhDys3")
-				cl.waitOnConsumerLeader(globalAccountName, mqttRetainedMsgsStreamName, "$MQTT_rmsgs_z3WIzPtj")
+				// <>/<> TODO: need to find a way to wait for the consumer
```
derekcollison, line 3029: Can do a consumer list and match based on prefix `$MQTT_rmsgs` and suffix being our server hash.

levb: exactly, just wasn't sure of the easiest way in the test; will finish in the afternoon.

server/mqtt.go
levb: 2/5 we shouldn't continue initializing MQTT - what else would we return?

derekcollison: Should not matter if we get an error trying to delete legacy consumer IMO. Should not prevent us from proceeding.

kozlovic (commented):
Change look ok to me, but not sure what the issue was and how is this going to be different.

derekcollison:
I am concerned by what I saw at Schaeffler that might be triggered by quick succession of delete and add with same name but different NRG assignment.

server/mqtt_test.go
levb: I ignored the prefix, since there aren't any "other" consumers in this stream anyway

levb:
@derekcollison I can't get past the build error after several retries, 
image
Can't see how this failure may possibly be related to my change, and this branch is up to date with `main`. Not sure what's going on. Will retry again.

kozlovic:
I have seen this test starting to flap on Travis. I can have a look say tomorrow, but for now, if it is a blocker, add a `t.Skip()` to skip this test.

server/mqtt.go
```diff
@@ -1439,26 +1439,31 @@ func (s *Server) mqttCreateAccountSessionManager(acc *Account, quitCh chan struc
...
-	if _, err := jsa.deleteConsumer(mqttRetainedMsgsStreamName, rmDurName); isErrorOtherThan(err, JSConsumerNotFoundErr) {
-		return nil, err
-	}
+	// retained messages.
+	//
+	// So we use a durable consumer, and create a new one each time we start.
+	// The old one should expire and get deleted due to inactivity. The name for
+	// the durable is $MQTT_rmsgs_{uuid}_{server-name}, the server name is just
+	// for readability.
+	rmDurName := mqttRetainedMsgsStreamName + "_" + nuid.Next() + "_" + s.String()
```
derekcollison, line 1456: Not really durable, just the name, and we only really expect it to live for the lifetime of the running server.

derekcollison (approved):
LGTM - Small comment but not blocking.

PR #5021, opened 2024-01-31
RAFT node responds to VoteRequest with outdated Term

mprimi:
A node's Term may be increased during handling of a VoteRequest. When that was the case, the node was responding with a stale Term.

Example trace:

```
[DBG] RAFT [S1Nunr6R - S-R3F-5xgoWjQW] Received a voteRequest &{term:27 lastTerm:26 lastIndex:46 candidate:yrzKKRBu reply:$NRG.R.OHF7VRUO}
[DBG] RAFT [S1Nunr6R - S-R3F-5xgoWjQW] Stepping down from leader, detected higher term: 27 vs 26
[DBG] RAFT [S1Nunr6R - S-R3F-5xgoWjQW] Stepping down
[DBG] RAFT [S1Nunr6R - S-R3F-5xgoWjQW] Switching to follower
[DBG] RAFT [S1Nunr6R - S-R3F-5xgoWjQW] Sending a voteResponse &{term:26 peer:S1Nunr6R granted:true} -> "$NRG.R.OHF7VRUO"
```
This is handled correctly except VoteResponse in the last log line should be sending back Term 27, not 26.

### DIFF
--- a/server/raft.go
+++ b/server/raft.go
@@ -3953,6 +3953,10 @@ func (n *raft) processVoteRequest(vr *voteRequest) error {
 			n.resetElect(randCampaignTimeout())
 		}
 	}
+
+	// Term might have changed, make sure response has the most current
+	vresp.term = n.term
+
 	n.Unlock()
 
 	n.sendReply(vr.reply, vresp.encode())
--- a/server/raft_test.go
+++ b/server/raft_test.go
@@ -273,15 +273,13 @@ func TestNRGSimpleElection(t *testing.T) {
 		re := decodeVoteResponse(msg.Data)
 		require_True(t, re != nil)
 
-		// The new term hasn't started yet, so the vote responses
-		// should contain the term from before the election. It is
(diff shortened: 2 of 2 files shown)


derekcollison:
Seems this test not always fails - `TestNRGSimpleElection`

Make sure you can get a green run on branch before posting PR.

neilalexander (commented):
~~Is this the right thing to do? The vote response term should surely always correspond with the vote request, otherwise the vote responses may appear to be unsolicited for the new term.~~

~~Additionally, nodes receiving a vote response when in candidate state will simply ignore the response if the term is less than the current term, see L2943: `if vresp.granted && nterm >= vresp.term {`~~

Have checked this against another Raft implementation and think it's OK, will fix the test.

derekcollison:
@mprimi feel free to convert from draft and will merge. Thanks!

derekcollison (approved):
LGTM

PR #5022, opened 2024-02-01
[IMPROVED] Utilize our IO gated semaphore more comprehensively.

derekcollison:
In extended stress testing we ran into more issues with concurrent blocking IO and the Go runtime. We had the gated semaphore already in use but not everywhere.


wallyqs (approved):
LGTM

PR #5024, opened 2024-02-01
NRG: Don't send metaleader snapshot as normal entry on leader change

neilalexander:
Sending meta snapshots as normal entries on leader change might cause a snapshot to be generated before the apply queue has been processed fully.


derekcollison (approved):
LGTM

PR #5027, opened 2024-02-01
[FIXED] Use dios gate internally and do not wrap whole function.

derekcollison:
Was not released, was introduced in dios work.


kozlovic (approved):
LGTM

PR #5028, opened 2024-02-02
[IMPROVED] Revamped fallback cleanup of consumer raft directory.

derekcollison:
We used to more broadly try to cleanup directories under a consumer. This could cause issues with reports of missing "tav.idx" etc.

This only scopes to then the account has been removed from the server.


wallyqs (approved):
LGTM

PR #5029, opened 2024-02-02
Add support to customize cluster ping_interval

wallyqs:
Adds support to customize the `ping_interval` for routes to be different from `ping_interval` from clients:

```hcl
port = 4222
http = 8222

server_name = 'A'

cluster {
  name = 'ABC'
  port = 6222
  routes = [
    nats://localhost:6222
    nats://localhost:6223
    nats://localhost:6224
  ]
  ping_interval = '5s'
  ping_max = 6
}

ping_interval = '30s'
```

### DIFF
--- a/server/client.go
+++ b/server/client.go
@@ -4795,7 +4795,11 @@ func (c *client) processPingTimer() {
 
 	var sendPing bool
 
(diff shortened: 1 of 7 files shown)


derekcollison (approved):
LGTM

PR #5032, opened 2024-02-02
[FIXED] Make sure to not restart streams or consumers when they are deleted or creation time is too soon.

derekcollison:
Would lead to zombies, most common in consumers.


wallyqs (approved):
LGTM

PR #5036, opened 2024-02-03
Improved batch direct get.

derekcollison:
Updated direct get batch to always have num pending in headers. Also added in last sequence to detect any message loss to all messages and the EOB marker. Will also allow pipeline fills, to better set us up for using this as even lighter weight replacement for any consumer.

### DIFF
--- a/server/stream.go
+++ b/server/stream.go
@@ -4104,10 +4104,18 @@ func (mset *stream) getDirectRequest(req *JSApiMsgGetRequest, reply string) {
 
 	seq := req.Seq
 	wc := subjectHasWildcard(req.NextFor)
+	// For tracking num pending if we are batch.
+	var np, lseq uint64
+	var isBatchRequest bool
 	batch := req.Batch
 	if batch == 0 {
 		batch = 1
+	} else {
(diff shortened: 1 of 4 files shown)


server/stream.go
```diff
@@ -4174,8 +4200,8 @@ func (mset *stream) getDirectRequest(req *JSApiMsgGetRequest, reply string) {
 
 	// If batch was requested send EOB.
 	if batch > 1 {
-		np, _ := store.NumPending(seq, req.NextFor, false)
-		hdr := []byte(fmt.Sprintf("NATS/1.0 204 EOB\r\nNats-Pending-Messages: %d\r\n\r\n", np))
+		// TODO(dlc) - Should we recalculate num pending here to account for new messages?
```
ripienaar, line 4203: yeah would feel weird if these are wrong

ripienaar (approved):
LGTM

server/stream.go
derekcollison: Chance os low but not zero. I could re-calculate on EOB marker but again those calls aren't free. Or I could add ValidThrough like we do with some lower level calls.

PR #5038, opened 2024-02-05
[FIX] [authcallout] ensure authcallout in operator mode sends an auth error instead of timeout

aricart:
For operator mode, the server was not sending an auth violation. Added additional tests checking for this on both conf and operator mode.


derekcollison (approved):
LGTM

PR #5039, opened 2024-02-05
Variable-length prefix slices in `stree`

neilalexander:
This replaces the fixed-size prefix arrays in the node meta struct with variable-length slices.

Beforehand, every single node in the tree had a 60-byte prefix array embedded in the node meta struct, regardless of whether the prefix was anywhere near that long, which will be a contributing factor to why `stree` is so prominent in the in-use heap memory on systems where subject cardinality is very high.

This does add a few allocations during the insert phase, however it removes various others from many other paths as we no longer need to create new slice headers from the `prefixLen` on each call to `path`, `newNode4`, `newNode16`, `newNode256` or `matchParts`.

As a result, as well as saving memory, there's actually a slight performance improvement in most cases.

On `main` with 100 million subjects:
```
=== RUN   TestSubjectTreeMatchAllPerf
    stree_test.go:602: Match ">" took 12.530168458s and matched 100000000 entries
    stree_test.go:602: Match "subj.>" took 12.511737208s and matched 100000000 entries
    stree_test.go:602: Match "subj.*.*" took 11.492906792s and matched 100000000 entries
    stree_test.go:602: Match "*.*.*" took 11.461866875s and matched 100000000 entries
    stree_test.go:602: Match "subj.1.*" took 115.148375ms and matched 999408 entries
    stree_test.go:602: Match "subj.1.>" took 125.542584ms and matched 999408 entries
    stree_test.go:602: Match "subj.*.1" took 14.5µs and matched 1 entries
    stree_test.go:602: Match "*.*.1" took 6.167µs and matched 1 entries
--- PASS: TestSubjectTreeMatchAllPerf (78.29s)
```

On this branch, also with 100 million subjects:
```
=== RUN   TestSubjectTreeMatchAllPerf
    stree_test.go:602: Match ">" took 11.358998084s and matched 100000000 entries
    stree_test.go:602: Match "subj.>" took 11.398956958s and matched 100000000 entries
    stree_test.go:602: Match "subj.*.*" took 10.474617958s and matched 100000000 entries
    stree_test.go:602: Match "*.*.*" took 10.43056825s and matched 100000000 entries
    stree_test.go:602: Match "subj.1.*" took 105.615875ms and matched 1002026 entries
    stree_test.go:602: Match "subj.1.>" took 115.68ms and matched 1002026 entries
    stree_test.go:602: Match "subj.*.1" took 22.083µs and matched 1 entries
    stree_test.go:602: Match "*.*.1" took 6.25µs and matched 1 entries
--- PASS: TestSubjectTreeMatchAllPerf (75.10s)
```

### DIFF
--- a/server/stree/dump.go
+++ b/server/stree/dump.go
@@ -38,7 +38,7 @@ func (t *SubjectTree[T]) dump(w io.Writer, n node, depth int) {
 	} else {
 		// We are a node type here, grab meta portion.
 		bn := n.base()
-		fmt.Fprintf(w, "%s %s Prefix: %q\n", dumpPre(depth), n.kind(), bn.prefix[:bn.prefixLen])
+		fmt.Fprintf(w, "%s %s Prefix: %q\n", dumpPre(depth), n.kind(), bn.prefix)
 		depth++
 		n.iter(func(n node) bool {
 			t.dump(w, n, depth)
--- a/server/stree/node.go
+++ b/server/stree/node.go
@@ -37,9 +37,15 @@ type node interface {
 // This also makes the meta base layer exactly 64 bytes, a normal L1 cache line.
 const maxPrefixLen = 60
 
-// 64 bytes total - an L1 cache line.
 type meta struct {
-	prefix    [maxPrefixLen]byte
-	prefixLen uint16
-	size      uint16
+	prefix []byte
+	size   uint16
 }
+
+func (m *meta) setPrefix(pre []byte) {
+	max := min(len(pre), maxPrefixLen)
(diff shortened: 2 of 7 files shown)


derekcollison:
The design was to avoid those allocations. We could make it smaller potentially but not a fan of introducing more allocations which I have seen take up to 10ms depending on load. So trying to optimize inserts as much as we can here.

Agree need to watch memory too, so maybe making the prefix smaller might be worth looking into.

Be good to be able to get the subject space from those machines as a test set.

neilalexander:
I had forgotten about this PR but this was pretty much done in #5339 in the end.

PR #5041, opened 2024-02-05
[ADDED] JetStream domain elected advisory for when we have a new meta leader.

derekcollison:

### DIFF
--- a/server/jetstream_cluster.go
+++ b/server/jetstream_cluster.go
@@ -5333,6 +5333,30 @@ func (js *jetStream) stopUpdatesSub() {
 	}
 }
 
+func (s *Server) sendDomainLeaderElectAdvisory() {
+	js, cc := s.getJetStreamCluster()
+	if js == nil || cc == nil {
+		return
+	}
+
+	js.mu.RLock()
+	node := cc.meta
+	js.mu.RUnlock()
+
+	adv := &JSDomainLeaderElectedAdvisory{
+		TypedEvent: TypedEvent{
+			Type: JSDomainLeaderElectedAdvisoryType,
+			ID:   nuid.Next(),
+			Time: time.Now().UTC(),
+		},
+		Leader:   node.GroupLeader(),
+		Replicas: s.replicas(node),
+		Domain:   s.getOpts().JetStreamDomain,
+	}
+
+	s.publishAdvisory(nil, JSAdvisoryDomainLeaderElected, adv)
+}
+
 func (js *jetStream) processLeaderChange(isLeader bool) {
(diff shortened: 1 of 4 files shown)


server/jetstream_cluster.go
```diff
@@ -5333,6 +5333,30 @@ func (js *jetStream) stopUpdatesSub() {
...
+
+	js.mu.RLock()
+	node := cc.meta
+	js.mu.RUnlock()
+
+	adv := &JSDomainLeaderElectedAdvisory{
+		TypedEvent: TypedEvent{
+			Type: JSDomainLeaderElectedAdvisoryType,
+			ID:   nuid.Next(),
+			Time: time.Now().UTC(),
```
philpennock, line 5350: Should this time stamp be obtained while the lock is held, so that we don't report that at time T the leader was in one place when it had actually already moved out, or do we go "it's second resolution, you can't rely on that for anything but informational purposes in the context of subscribing to a complete stream"?

philpennock (approved):

server/jetstream_events.go
```diff
@@ -193,10 +193,21 @@ const JSRestoreCompleteAdvisoryType = "io.nats.jetstream.advisory.v1.restore_com
...
-// JSStreamLeaderElectedAdvisoryType is sent when the system elects a leader for a stream.
+// JSClusterLeaderElectedAdvisoryType is sent when the system elects a new meta leader.
+const JSDomainLeaderElectedAdvisoryType = "io.nats.jetstream.advisory.v1.domain_leader_elected"
+
+// JSClusterLeaderElectedAdvisory indicates that a domain has elected a new leader.
+type JSDomainLeaderElectedAdvisory struct {
+	TypedEvent
+	Leader   string      `json:"leader"`
+	Replicas []*PeerInfo `json:"replicas"`
+	Domain   string      `json:"domain,omitempty"`
```
ripienaar, line 204: Maybe add `cluster`?

derekcollison: Not following that, what do you mean?

ripienaar: It would be good to know what cluster hosts the leader in this event.

ripienaar: stream is where it is, it'll never change until someone goes and specifically moves a stream so its not needed there I think.

But the leader moves around and one might want to react on the advisory should the leader move to a bad place, be good if the info needed to answer if its a bad place is all in the advisory.

derekcollison: ok done.

ripienaar (approved):
LGTM

PR #5045, opened 2024-02-06
Use IO gate when doing io.FullRead on block

wallyqs:
While troubleshooting a cluster recovering a large stream, it was spotted that this read call was taking long to complete.


derekcollison (approved):
LGTM

PR #5048, opened 2024-02-07
FIXED: MQTT retained message consumer creation

levb:
- Cleaned up the retained message consumer name so that it does not cause problems
- Per @derekcollison's recommendation, use an ephemeral consumer for retained messages

### DIFF
--- a/server/mqtt.go
+++ b/server/mqtt.go
@@ -1464,7 +1464,7 @@ func (s *Server) mqttCreateAccountSessionManager(acc *Account, quitCh chan struc
 			InactiveThreshold: 5 * time.Minute,
 		},
 	}
-	if _, err := jsa.createConsumer(ccfg); err != nil {
+	if _, err := jsa.createEphemeralConsumer(ccfg); err != nil {
 		return nil, fmt.Errorf("create retained messages consumer for account %q: %v", accName, err)
 	}
 
@@ -1644,7 +1644,7 @@ func (jsa *mqttJSA) sendAck(ackSubject string) {
 	jsa.sendq.push(&mqttJSPubMsg{subj: ackSubject, hdr: -1})
 }
 
-func (jsa *mqttJSA) createConsumer(cfg *CreateConsumerRequest) (*JSApiConsumerCreateResponse, error) {
+func (jsa *mqttJSA) createEphemeralConsumer(cfg *CreateConsumerRequest) (*JSApiConsumerCreateResponse, error) {
 	cfgb, err := json.Marshal(cfg)
 	if err != nil {
 		return nil, err
--- a/server/mqtt.go
+++ b/server/mqtt.go
@@ -1458,9 +1458,8 @@ func (s *Server) mqttCreateAccountSessionManager(acc *Account, quitCh chan struc
 	//
 	// So we use a durable consumer, and create a new one each time we start.
 	// The old one should expire and get deleted due to inactivity. The name for
-	// the durable is $MQTT_rmsgs_{uuid}_{server-name}, the server name is just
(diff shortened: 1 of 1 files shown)


server/mqtt.go
```diff
@@ -1658,12 +1649,21 @@ func (jsa *mqttJSA) createConsumer(cfg *CreateConsumerRequest) (*JSApiConsumerCr
...
+	ccr := ccri.(*JSApiConsumerCreateResponse)
+	return ccr, ccr.ToError()
+}
+
+func (jsa *mqttJSA) createDurableConsumer(cfg *CreateConsumerRequest) (*JSApiConsumerCreateResponse, error) {
+	cfgb, err := json.Marshal(cfg)
+	if err != nil {
+		return nil, err
 	}
+	subj := fmt.Sprintf(JSApiDurableCreateT, cfg.Stream, cfg.Config.Durable)
```
derekcollison, line 1666: Why is this durable and one above is not?

server/mqtt.go
```diff
@@ -1658,12 +1649,21 @@ func (jsa *mqttJSA) createConsumer(cfg *CreateConsumerRequest) (*JSApiConsumerCr
 	if err != nil {
 		return nil, err
 	}
-	var subj string
-	if cfg.Config.Durable != _EMPTY_ {
-		subj = fmt.Sprintf(JSApiDurableCreateT, cfg.Stream, cfg.Config.Durable)
-	} else {
-		subj = fmt.Sprintf(JSApiConsumerCreateT, cfg.Stream)
+	subj := fmt.Sprintf(JSApiConsumerCreateT, cfg.Stream)
+	ccri, err := jsa.newRequest(mqttJSAConsumerCreate, subj, 0, cfgb)
```
derekcollison, line 1653: How long will this wait in place?

server/mqtt.go
```diff
@@ -1658,12 +1649,21 @@ func (jsa *mqttJSA) createConsumer(cfg *CreateConsumerRequest) (*JSApiConsumerCr
...
+	if err != nil {
+		return nil, err
+	}
+	ccr := ccri.(*JSApiConsumerCreateResponse)
+	return ccr, ccr.ToError()
+}
+
+func (jsa *mqttJSA) createDurableConsumer(cfg *CreateConsumerRequest) (*JSApiConsumerCreateResponse, error) {
```
derekcollison, line 1661: How is this materially different from createConsumer above?

server/mqtt.go
levb: I broke out the function into 2 to make it more explicit. Do you prefer to keep it exactly as before, 1 function  triggering which API to use based on `.Durable` set?

server/mqtt.go
levb: `newRequest` waits for the response to arrive, times out at 4s, https://github.com/nats-io/nats-server/blob/main/server/mqtt.go#L1615

derekcollison (approved):
LGTM

kozlovic:
My 2 cents: I had this comment when I implemented MQTT and used a durable for the retained messages:
```
// Using ephemeral consumer is too risky because if this server were to be
// disconnected from the rest for few seconds, then the leader would remove
// the consumer, so even after a reconnect, we would no longer receive
// retained messages.
```
Granted that was at the very early stages of JetStream and things may be totally different now, but is that concern still valid or not? If still valid, then I wonder why switching to an ephemeral...

derekcollison:
Its not since we can control the inactivity threshold now, vs before it was fixed and fairly low, IIRC 5s.

kozlovic:
I see, great!

PR #5050, opened 2024-02-09
[IMPROVED] MQTT: do not JSON-encode retained message bodies (restore #4825)

levb:
This change is for 2.11 only. As noted in the original PR, downgrading to 2.10 and below would result in data loss for MQTT retained messages (and potential error handling issues).

https://github.com/nats-io/nats-server/pull/4825 was reverted from main, adding it back now, with some changes. Specifically, no longer use JSON for the header value; instead use 4 separate headers.

The revert was messy, not sure I can reproduce the issue seen in the original PR.


server/mqtt.go
```diff
@@ -1960,6 +1964,7 @@ func (as *mqttAccountSessionManager) processRetainedMsg(_ *subscription, c *clie
 	as.handleRetainedMsg(rm.Subject, &mqttRetainedMsgRef{sseq: seq}, rm, false)
 
 	// If we were recovering (lastSeq > 0), then check if we are done.
+	// TODO <>/<> does this work with ReplayImmediate?
```
levb, line 1967: @derekcollison @kozlovic Wouldn't the sequence check here potentially fire too early? Does `ReplayInstant` maintain the sequence order? Was not sure from the godoc. (there is no change in this PR, just curious)

https://github.com/nats-io/nats-server/blob/cca23cdacc66411905fe80fd7d17a9ccc8443401/server/mqtt.go#L1462

kozlovic: ReplayInstant just means that the messages are resent without preserving the gap in time that there may be between original ingress messages. So the order should be preserved.

server/mqtt.go
```diff
@@ -2772,20 +2776,94 @@ func (as *mqttAccountSessionManager) loadRetainedMessages(subjects map[string]st
...
+	buf.WriteByte(':')
+	buf.WriteString(rm.Topic)
+	buf.WriteString(_CRLF_)
+
+	buf.WriteString(mqttNatsRetainedMessageFlags)
+	buf.WriteByte(':')
+	buf.WriteString(strconv.FormatUint(uint64(rm.Flags), 16))
+	buf.WriteString(_CRLF_)
+
+	if rm.Origin != "" {
```
derekcollison, line 2821: Use `_EMPTY_` vs "" in server.

server/mqtt.go
```diff
@@ -2772,20 +2776,94 @@ func (as *mqttAccountSessionManager) loadRetainedMessages(subjects map[string]st
...
+		buf.WriteString(mqttNatsRetainedMessageOrigin)
+		buf.WriteByte(':')
+		buf.WriteString(rm.Origin)
+		buf.WriteString(_CRLF_)
+	}
+	if rm.Source != "" {
```
derekcollison, line 2827: Same

derekcollison (commented):
LGTM - a small nit. Will let @kozlovic take a proper look.

server/mqtt_test.go
```diff
@@ -7448,6 +7448,84 @@ func TestMQTTJetStreamRepublishAndQoS0Subscribers(t *testing.T) {
...
+			listen: 127.0.0.1:-1
+			consumer_inactive_threshold: %q
+		}
+
+		# For access to system account.
+		accounts { $SYS { users = [ { user: "admin", pass: "s3cr3t!" } ] } }
+	`
+	conf := createConfFile(t, []byte(fmt.Sprintf(tmpl, tdir, "0.2s")))
+	s, o := RunServerWithConfig(conf)
+
```
kozlovic, line 7471: You should have a `defer testMQTTShutdownServer(s)` here in case the test fails and does not leave a server running...

server/mqtt_test.go
```diff
@@ -7448,6 +7448,84 @@ func TestMQTTJetStreamRepublishAndQoS0Subscribers(t *testing.T) {
...
+	// Connect and publish a retained message, this will be in the "newer" form,
+	// with the metadata in the header.
+	mc, r := testMQTTConnectRetry(t, &mqttConnInfo{clientID: "test", cleanSess: true}, o.MQTT.Host, o.MQTT.Port, 5)
```
kozlovic, line 7474: Get into the habit of having a `defer mc.Close()` after connect in case the test fails before you explicitly call `mc.Close()` so that it does not leave a connection running and possibly attempting to reconnect (case of NATS clients, not really MQTT, but good habit to take).

server/mqtt_test.go
```diff
@@ -7448,6 +7448,84 @@ func TestMQTTJetStreamRepublishAndQoS0Subscribers(t *testing.T) {
...
+		Data:    jsonData,
+	})
+	if err != nil {
+		t.Fatalf("Error publishing retained message to JS directly: %v", err)
+	}
+
+	// Restart the server to see that it picks up both retained messages on restart.
+	s.Shutdown()
+	s = RunServer(o)
+
```
kozlovic, line 7502: Need a new `defer testMQTTShutdownServer(s)` here.

server/mqtt_test.go
```diff
@@ -7448,6 +7448,84 @@ func TestMQTTJetStreamRepublishAndQoS0Subscribers(t *testing.T) {
...
+	// Connect again, subscribe, and check that we get both messages.
+	mc, r = testMQTTConnectRetry(t, &mqttConnInfo{clientID: "test", cleanSess: true}, o.MQTT.Host, o.MQTT.Port, 5)
```
kozlovic, line 7504: `defer mc.Close()`

server/mqtt_test.go
```diff
@@ -7448,6 +7448,84 @@ func TestMQTTJetStreamRepublishAndQoS0Subscribers(t *testing.T) {
...
+	testMQTTCheckConnAck(t, r, mqttConnAckRCConnectionAccepted, false)
+	testMQTTSub(t, 1, mc, r, []*mqttFilter{{filter: "foo/+", qos: 0}}, []byte{0})
+	testMQTTCheckPubMsg(t, mc, r, "foo/1", mqttPubFlagRetain, []byte("msg1"))
+	testMQTTCheckPubMsg(t, mc, r, "foo/2", mqttPubFlagRetain, []byte("msg2"))
+	testMQTTExpectNothing(t, r)
+	mc.Close()
+
+	// Clear both retained messages.
+	mc, r = testMQTTConnectRetry(t, &mqttConnInfo{clientID: "test", cleanSess: true}, o.MQTT.Host, o.MQTT.Port, 5)
```
kozlovic, line 7513: `defer mc.Close()`

server/mqtt_test.go
```diff
@@ -7448,6 +7448,84 @@ func TestMQTTJetStreamRepublishAndQoS0Subscribers(t *testing.T) {
...
+	testMQTTCheckConnAck(t, r, mqttConnAckRCConnectionAccepted, false)
+	testMQTTPublish(t, mc, r, 0, false, true, "foo/1", 0, []byte{})
+	testMQTTPublish(t, mc, r, 0, false, true, "foo/2", 0, []byte{})
+	mc.Close()
+
+	// Connect again, subscribe, and check that we get nothing.
+	mc, r = testMQTTConnectRetry(t, &mqttConnInfo{clientID: "test", cleanSess: true}, o.MQTT.Host, o.MQTT.Port, 5)
```
kozlovic, line 7520: `defer mc.Close()`

server/mqtt_test.go
```diff
@@ -7448,6 +7448,84 @@ func TestMQTTJetStreamRepublishAndQoS0Subscribers(t *testing.T) {
...
+	testMQTTCheckConnAck(t, r, mqttConnAckRCConnectionAccepted, false)
+	testMQTTSub(t, 1, mc, r, []*mqttFilter{{filter: "foo/+", qos: 0}}, []byte{0})
+	testMQTTExpectNothing(t, r)
+	mc.Close()
+
+	testMQTTShutdownServer(s)
```
kozlovic, line 7526: No need if you have the proper defer.

kozlovic (changes requested):
Little changes needed in the test code only.

kozlovic (approved):
LGTM

PR #5054, opened 2024-02-09
[FIXED] KeyValue not found after server restarts

derekcollison:
Do not force sequence mismatches and possible dataloss on first sequence move past a catchup request.

When a server restarts and does a stream catchup where the snapshot from the leader is for messages it no longer has, we would create a sequence mismatch reset on the follower.
    
Also on a catchup if the stream state has moved past the updated request range from a follower, meaning first for the new state is past the last of the request, extend and continue catchup.


wallyqs (approved):
LGTM

PR #5057, opened 2024-02-11
Added OpenTelemetry support

### LINKED ISSUE #5052: Otel activation for message traces
### Proposed change

In https://github.com/nats-io/nats-server/pull/5014 a basic capability to do on-demand message tracing was introduced for core pub-sub messages.

This capability is activated using a NATS specific header.  The bigger picture is that users want to trigger traces based on trace headers from the likes of Open Telemetry etc.

There is prior-art in the server around [ADR-3](https://github.com/nats-io/nats-architecture-and-design/blob/main/adr/ADR-3.md) that already supports triggering based on the headers we need and there are already functions for parsing the headers and determining if it's a continuation of a trace that started outside of NATS. See the `shouldSample()` function.

This leaves the question of where should traces be delivered when these headers like `traceparent` are present, of course we don't know as previously we configured the delivery per message. The only option I can think of is to add an account scope config item that configures a per-account subject for these externally-activated traces to be delivered.  

Some other cases to consider, there might be more:

 * If the configuration is not set the account does not support this style of activation of tracing.
 * If the `Nats-Trace-Only` is present AND otel headers are set, we honor `Nats-Trace-Only` and do not deliver the message
 * If `Nats-Trace-Dest` is set AND otel headers are set, we honor `Nats-Trace-Dest` and deliver to that subject

/cc @kozlovic 

### Use case

Integration with external tracing systems on account level.

### Contribution

_No response_

kozlovic:
If the `Nats-Trace-Dest` header is not present, but `Traceparent` is and its last token is `01`, then message tracing is triggered. This also requires that the account be defined with a `trace_dest` subject so that traces can be sent there.
Note that `Nats-Trace-Only` is not applicable for `Traceparent`.

Addition to PR #5014
Resolves #5052

### DIFF
--- a/server/msgtrace.go
+++ b/server/msgtrace.go
@@ -575,12 +575,12 @@ func (t *msgTrace) disableTraceHeaders(c *client, msg []byte) []int {
 // Changes back the character at the given position `pos` in the `msg`
 // byte slice to the first character of the MsgTraceSendTo header.
 func (t *msgTrace) enableTraceHeaders(c *client, msg []byte, positions []int) {
-	headers := [2]string{MsgTraceSendTo, trcCtx}
+	firstChar := [2]byte{MsgTraceSendTo[0], trcCtx[0]}
 	for i, pos := range positions {
 		if pos == -1 {
 			continue
 		}
-		msg[pos] = headers[i][0]
+		msg[pos] = firstChar[i]
 	}
 }
 
--- a/server/msgtrace.go
+++ b/server/msgtrace.go
@@ -330,12 +330,25 @@ func (c *client) initMsgTrace() *msgTrace {
 		return nil
 	}
 	hdr := c.msgBuf[:c.pa.hdr]
+	var external bool
 	// Do not call c.parseState.getHeader() yet for performance reasons.
 	// We first do a "manual" search of the "send-to" destination's header.
 	// If not present, no need to lift the message headers.
 	td := getHeader(MsgTraceSendTo, hdr)
 	if len(td) <= 0 {
-		return nil
+		// If the NATS trace destination header is not present, look for
+		// the external trace header.
+		trcCtxHdrVal := getHeader(trcCtx, hdr)
+		if len(trcCtxHdrVal) <= 0 {
+			return nil
+		}
+		// See if we need to trigger the race based on the last token
+		// that should be set to "01".
+		tk := bytes.Split(trcCtxHdrVal, stringToBytes("-"))
+		if len(tk) != 4 || len([]byte(tk[3])) != 2 || !bytes.Equal(tk[3], stringToBytes("01")) {
+			return nil
+		}
+		external = true
 	}
 	// Now we know that this is a message that requested tracing, we
 	// will lift the headers since we also need to transfer them to
@@ -346,11 +359,14 @@ func (c *client) initMsgTrace() *msgTrace {
 	}
 	ct := getCompressionType(headers.Get(acceptEncodingHeader))
 	var traceOnly bool
-	if to := headers.Get(MsgTraceOnly); to != _EMPTY_ {
-		tos := strings.ToLower(to)
-		switch tos {
-		case "1", "true", "on":
-			traceOnly = true
+	// Check for traceOnly only if not external.
+	if !external {
+		if to := headers.Get(MsgTraceOnly); to != _EMPTY_ {
+			tos := strings.ToLower(to)
+			switch tos {
+			case "1", "true", "on":
+				traceOnly = true
+			}
 		}
 	}
 	var (
@@ -403,11 +419,26 @@ func (c *client) initMsgTrace() *msgTrace {
 		acc = c.acc
 		ian = acc.GetName()
 	}
+	// If external, we need to have the account's trace destination set,
+	// otherwise, we are not enabling tracing.
+	var dest string
+	if external {
+		if acc != nil {
+			acc.mu.RLock()
+			dest = acc.TraceDest
+			acc.mu.RUnlock()
+		}
+		if dest == _EMPTY_ {
+			return nil
+		}
+	} else {
+		dest = string(td)
+	}
 	c.pa.trace = &msgTrace{
 		srv:  c.srv,
 		acc:  acc,
 		oan:  oan,
-		dest: string(td),
+		dest: dest,
 		ct:   ct,
 		hop:  hop,
 		event: &MsgTraceEvent{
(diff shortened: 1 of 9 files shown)


server/server.go
```diff
@@ -1046,6 +1046,15 @@ func validateOptions(o *Options) error {
 		return fmt.Errorf("max_payload (%v) cannot be higher than max_pending (%v)",
 			o.MaxPayload, o.MaxPending)
 	}
+	// Check that account's trace_dest is a valid subject.
+	for _, acc := range o.Accounts {
+		if acc.TraceDest == _EMPTY_ {
+			continue
+		}
+		if !IsValidSubject(acc.TraceDest) {
```
derekcollison, line 1054: `IsValidPublishSubject`?

server/msgtrace.go
```diff
@@ -496,49 +527,61 @@ func (t *msgTrace) setHopHeader(c *client, msg []byte) []byte {
...
 }
 
 // Changes back the character at the given position `pos` in the `msg`
 // byte slice to the first character of the MsgTraceSendTo header.
-func (t *msgTrace) enableTraceHeader(c *client, msg []byte, pos int) {
-	if pos <= 0 {
-		return
+func (t *msgTrace) enableTraceHeaders(c *client, msg []byte, positions []int) {
+	headers := [2]string{MsgTraceSendTo, trcCtx}
+	for i, pos := range positions {
```
derekcollison, line 579: Can inline I think.

```go
for i, pos := range [2]string{MsgTraceSendTo, trcCtx}
```

server/opts.go
```diff
@@ -3025,6 +3025,14 @@ func parseAccounts(v interface{}, opts *Options, errors *[]error, warnings *[]er
 						*errors = append(*errors, err)
 						continue
 					}
+				case "trace_dest", "trace_destination", "trace_subject":
+					td := mv.(string)
+					if !IsValidSubject(td) {
```
derekcollison, line 3030: Same here?

`IsValidPublishSubject`

derekcollison (commented):
In general LGTM - A few nits/questions.

server/msgtrace.go
kozlovic: No, we are given `positions` which are the indexes in the `msg` byte slice where to put back the first character of the 2 given headers. But agreed that I can replace `headers` with a slice of the first character of the 2 headers.

server/server.go
kozlovic: That is a good point and will add to the test that I will reject wildcard subjects (which I assume `IsValidPublishSubject` does.

kozlovic:
@derekcollison Updated the PR based on your comments. But also would want @ripienaar to have a look at it tomorrow to see if that matches its expectations for otel support.

derekcollison:
Thanks @kozlovic, agree will wait on @ripienaar.

server/msgtrace.go
```diff
@@ -330,12 +330,25 @@ func (c *client) initMsgTrace() *msgTrace {
...
+		// If the NATS trace destination header is not present, look for
+		// the external trace header.
+		trcCtxHdrVal := getHeader(trcCtx, hdr)
+		if len(trcCtxHdrVal) <= 0 {
+			return nil
+		}
+		// See if we need to trigger the race based on the last token
+		// that should be set to "01".
+		tk := bytes.Split(trcCtxHdrVal, stringToBytes("-"))
+		if len(tk) != 4 || len([]byte(tk[3])) != 2 || !bytes.Equal(tk[3], stringToBytes("01")) {
```
ripienaar, line 348: I'll do some testing but just fyi, looks like the initial implemenation in the server was maybe a bit naive and I based my spec on that.

This isnt just a case of a `01` and if its `01` its traced, this is actually a hex encoded bit field - today there is only one value stored in this bit field so the naive check is ok, but we might need to revisit this in future.  And I'll try to get some example trace headers

ripienaar (changes requested):
This looks good to me but keeping in mind that one comment.

Bigger problem is that I dont think our otel support ever worked, when we wrote it not much was using the then-new standard :)

https://github.com/nats-io/nats-server/blob/b2d37d7a7d5dda2b54fba70763c6f4aeeaea9b6e/server/accounts.go#L2067

This will do `Traceparent` and all our checks are done case sensitively, but the header is exactly `traceparent` in the spec, we need to either check without case sensitivity as the http spec requires or at least make this one lowercase in `trcCtx`.

I've asked for example headers so will update when I got those to be 100% certain

So hand fixing that it works with lowercase `traceparent` but I see in the headers we send on its again upper cased

From a trace:
```
{"header":{"Nats-Trace-Hop":["2"],"Traceparent":.....
```

Not sure if this is avoidable because go headers class is not great but ideally we change the case of headers

kozlovic:
@ripienaar The original code with "Traceparent" is correct because the checks were done after "lifting" the message header into a http.Header that makes it canonical. 
The issue that I have is that I use low level "getHeader" to just find the string without having to lift the headers. That will be a problem since bytes.IndexOf() does not do case insensitive. Of course, in this case I could have a version of the header string that is all lower case, but if the app sends with a Go library, it will likely be canonical. So that is a huge problem, and maybe we unfortunately have to pay the price of lifting the headers?

server/msgtrace.go
kozlovic: Ok, thanks. I'll check the spec too and see what needs to be done there.

ripienaar:
> @ripienaar The original code with "Traceparent" is correct because the checks were done after "lifting" the message header into a http.Header that makes it canonical. The issue that I have is that I use low level "getHeader" to just find the string without having to lift the headers. That will be a problem since bytes.IndexOf() does not do case insensitive. Of course, in this case I could have a version of the header string that is all lower case, but if the app sends with a Go library, it will likely be canonical. So that is a huge problem, and maybe we unfortunately have to pay the price of lifting the headers?

Would it be cheaper to do `getHeader` on both the `traceparent` and then `Traceparent`? Probably unavoidable that'd we'd send it on as `Traceparent` though, I wonder if the std otel libraries will handle that well let me try and see

ripienaar:
> Would it be cheaper to do `getHeader` on both the `traceparent` and then `Traceparent`? Probably unavoidable that'd we'd send it on as `Traceparent` though, I wonder if the std otel libraries will handle that well let me try and see

Looks to me like we'd have to send it as `traceparent` in the headers contained in the traces, but probably we cant because of go:

https://github.com/nats-io/nats-server/blob/b2d37d7a7d5dda2b54fba70763c6f4aeeaea9b6e/server/msgtrace.go#L54

We might have to document that people will have to lowercase it to use it :(

ripienaar:
Apparently if one use the header like a map you can circumvent the uppercase nonsense https://go.dev/play/p/WnjqoLSLhe

kozlovic:
@ripienaar I made the requested changes, please check to see if that works for you.

ripienaar (approved):
LGTM, does what I think we need from a behaviour perspective. But lets hold off on the merge while we get some details from one user to confirm it's 100% the right track

ripienaar:
The only possible addition I could think of is to be able to force compression on those messages - but from what I saw compression isnt helping much on these anyway so no biggie.

kozlovic:
@ripienaar Your opinion on changing the constant name `MsgTraceSendTo` (which maps to `Nats-Trace-Dest` to `MsgTraceDest`? Since this PR already "breaks" application using main in that the RequestInfo.Header is no longer an http.Header, may as well take that opportunity?

PS: I made them public thinking of apps importing server code to do tracing, let me know if you think it was a bad idea.

ripienaar:
> @ripienaar Your opinion on changing the constant name `MsgTraceSendTo` (which maps to `Nats-Trace-Dest` to `MsgTraceDest`? Since this PR already "breaks" application using main in that the RequestInfo.Header is no longer an http.Header, may as well take that opportunity?
> 
> PS: I made them public thinking of apps importing server code to do tracing, let me know if you think it was a bad idea.

Yes go ahead, no1 is using this other than me - its quite hard to use without docs which I'll add later hehe

Feel free to change whatever you need.  Public is fine I think

ripienaar:
Functionality wise I am happy to merge this we have confirmation this will do what we intend.

So will wait for code review from @derekcollison

derekcollison (approved):
LGTM

kozlovic:
@derekcollison @ripienaar I would like to make sure of something before this gets into a future release and cannot be changed. When I added the account's trace destination, I made it a public field because the way options are defined. Meaning that if one want to programmatically have an account and configure it, one would do: 
```
acc := NewAccount("MyAccount")
acc.TraceDest = "my.trace.dest"
opts.Accounts = []*Account{acc}
...
s, err := NewServer(opts)
...
```
Hence my decision to make `TraceDest` public, but is it the right thing? Are users expected to be able to configure accounts programmatically? It does not seem to be many public fields in `Account`, so I am now wondering if I made a mistake? 

Again, it would not be too late to change since this has not been released yet.

Conversely, I think that TraceDest *needs* to be added to `Account` in the JWT library in order for JWT account updates can set/update this value (this is what I am working on right now, along with the AllowTrace in service/stream).

ripienaar:
Where I embed NATS I often wish I had more options to configure accounts programatically - very little one can do now and it can be awkward to do things with accounts.

That said, I dont know if thats a wider need or not, certainly people who embed the server isnt a huge demographic so I'll take it or leave it, no strong oppinions

derekcollison:
Can it be access without a lock? If not then would make private and maybe add setter and getter.

kozlovic:
> Can it be accessed without a lock? If not then would make private and maybe add setter and getter.

That's a good point. I will switch to private.

Issue #555, opened 2017-08-07
Provide binary for IBM Power LE (ppc64le)


Beilmann:
- [x] Feature Request

## Provide binary for IBM Power LE (ppc64le)

- Please compile the software on IBM Power LE (ppc64le) and provide the installable package for download.
- Please provide also docker image for IBM Power LE (ppc64le)

kozlovic:
Building the executable for IBM Power LE (ppc64le) with manual cross compile is easy:
```
env CGO_ENABLED=0 GOOS=linux GOARCH=ppc64le go build -ldflags "-s -w" -o gnatsd
```
However, I don't think we could provide a docker image for that platform since I don't think that there is an official base image for POWER. One can publish an official image only if it is based on an official image or from SCRATCH.

kozlovic:
We provide binaries as part of a release for hardware that we could test/debug on should an issue be reported. For that reason, we are not going to include an IBM Power LE binary in the next release.
Remember that you can always compile the binary yourself from source, either directly on the target platform or using cross-compilation as indicated above.

PR #5058, opened 2024-02-11
Support building for linux/ppc64le arch

bruth:
This request was raised in #555 (2017) and more recently in the nats-docker repo here: https://github.com/nats-io/nats-docker/issues/117.

Although @kozlovic noted in #555 that we don't have hardware readily available to test against if a platform-specific bug is raised, we are already building for other architectures that fall into this category: `mips64le` and `s390x`. If such an issue arises, we would rely on the community who are using those platforms to facilitate debugging.

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -32,6 +32,7 @@ builds:
   - 386
   - mips64le
   - s390x
+  - ppc64le
   goarm:
   - 6
   - 7


derekcollison (approved):
LGTM

bruth:
Looks like a test need to be kicked (apologies, should have done `[skip ci]`).

wallyqs (approved):
LGTM

bruth:
@derekcollison tests pass when you are ready to merge. This will be for 2.11

PR #5070, opened 2024-02-13
Make ping/pong responses update last activity

wallyqs:
Changes ping/pong interval to also make updates to LastActivity to better reflect possible stale connections via monitoring.

### DIFF
--- a/server/client.go
+++ b/server/client.go
@@ -2380,7 +2380,9 @@ func (c *client) processPing() {
 
 	// Record this to suppress us sending one if this
 	// is within a given time interval for activity.
-	c.lastIn = time.Now()
+	now := time.Now()
+	c.lastIn = now
+	c.last = now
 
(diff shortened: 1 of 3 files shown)


derekcollison (commented):
Not sure we want this, as discussed knowing a connection has been idle from an app perspective is quite helpful.

wallyqs:
As different part of metadata from the ConnInfo to track the ping/pong activity (`last_ping_activity`) maybe? Something like: https://github.com/wallyqs/nats-server/commit/4915087f21f78a27012d77b60ebb8dfa7ae5fe2d

derekcollison:
Possibly, but not sure if that info is useful in real world scenarios.. WDYT?

wallyqs:
Good point, it would be rare to monitor a live stale connection error as it happens and this would be stored as part of the closed connections anyway. I will close this one for now.

PR #5077, opened 2024-02-14
[FIXED] flapping TestMQTTDecodeRetainedMessage

levb:
The order of messages was always the same on my machine, but doesn't have to be.


wallyqs (approved):
LGTM

derekcollison (approved):
LGTM

PR #5078, opened 2024-02-14
Change Account.TraceDest to private and add getter/setter

kozlovic:
Update the tests accordingly.

This change is ok since the message tracing feature is targeted for v2.11.0 which has not been released yet.

Related to #5057

### DIFF
--- a/server/accounts.go
+++ b/server/accounts.go
@@ -54,7 +54,6 @@ type Account struct {
 	Name         string
 	Nkey         string
 	Issuer       string
-	TraceDest    string
 	claimJWT     string
 	updated      time.Time
 	mu           sync.RWMutex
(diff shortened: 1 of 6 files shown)


server/accounts.go
```diff
@@ -256,12 +256,28 @@ func (a *Account) String() string {
 	return a.Name
 }
 
+func (a *Account) setTraceDest(dest string) {
```
derekcollison, line 259: These need to be public for embedded servers no?

kozlovic: But since we can't really configure much an account programmatically (refer to Ari's previous comment), I figure that we don't make it public at all, at least for now. There was still a need for the getter/setter internally, this is why I added them.

kozlovic: Ari comment was:
```
Where I embed NATS I often wish I had more options to configure accounts programatically - very little one can do now and it can be awkward to do things with accounts.
```

derekcollison (approved):
LGTM

Issue #5071, opened 2024-02-13
JS stream count not balanced among the cluster nodes


kohlisid:
### Observed behavior

When creating a multiple streams (with replicas = 3) on a Jetstream cluster (with number of nodes > stream replica count), I have been observing a behaviour where the streams are not evenly distributed among the servers. 

Some of the server instances end up getting a large chunk of the stream replicas. 

```
skohli@macos-JQWR9T560R ~ % nats --context east-sys-ac server report jetstream
╭──────────────────────────────────────────────────────────────────────────────────────────────────╮
│                                         JetStream Summary                                        │
├────────┬─────────┬─────────┬───────────┬──────────┬───────┬────────┬──────┬─────────┬────────────┤
│ Server │ Cluster │ Streams │ Consumers │ Messages │ Bytes │ Memory │ File │ API Req │ API Err    │
├────────┼─────────┼─────────┼───────────┼──────────┼───────┼────────┼──────┼─────────┼────────────┤
│ n1-c1  │ C1      │ 28      │ 0         │ 0        │ 0 B   │ 0 B    │ 0 B  │ 55      │ 0          │
│ n2-c1* │ C1      │ 1       │ 0         │ 0        │ 0 B   │ 0 B    │ 0 B  │ 33      │ 2 / 6.060% │
│ n3-c1  │ C1      │ 28      │ 0         │ 0        │ 0 B   │ 0 B    │ 0 B  │ 30      │ 0          │
│ n4-c1  │ C1      │ 0       │ 0         │ 0        │ 0 B   │ 0 B    │ 0 B  │ 0       │ 0          │
│ n5-c1  │ C1      │ 27      │ 0         │ 0        │ 0 B   │ 0 B    │ 0 B  │ 63      │ 0          │
├────────┼─────────┼─────────┼───────────┼──────────┼───────┼────────┼──────┼─────────┼────────────┤
│        │         │ 84      │ 0         │ 0        │ 0 B   │ 0 B    │ 0 B  │ 181     │ 2          │
╰────────┴─────────┴─────────┴───────────┴──────────┴───────┴────────┴──────┴─────────┴────────────╯
```

On some testing, if we **wait** for some time (sleep = 3s) before creating consecutive streams we end up seeing a far balanced distribution. 

```
skohli@macos-JQWR9T560R ~ % nats --context east-sys-ac server report jetstream
╭─────────────────────────────────────────────────────────────────────────────────────────────────────╮
│                                          JetStream Summary                                          │
[...]

PR #5079, opened 2024-02-14
[IMPROVED] Stream placement in cluster with N > R

derekcollison:
Calculate peer group based on streams as well as usage and realtime data on HAAssets vs async.

Resolves: #5071 


wallyqs (approved):
LGTM

PR #5080, opened 2024-02-14
[FIXED] Stree match bug, filestore minor improvements.

derekcollison:
When encountering partial frags or parts we were not properly offsetting by si. This would cause certain stree setups to fail to match properly against wildcarded subjects that had to deal with partial frags and partial parts.

Also lowered the prefix from 60 to 24.

Improved NumPending in filestore by check if fssLoaded vs whole cache.


wallyqs (approved):
LGTM

PR #5082, opened 2024-02-14
Added tests for MQTT retained messages in various cluster/domain conf…

levb:
This is a Test- and CI-only PR. No server changes.

See also https://github.com/ConnectEverything/mqtt-test/pull/2.


.github/workflows/MQTT_test.yaml
```diff
@@ -29,39 +25,13 @@ jobs:
...
-          go install github.com/ConnectEverything/mqtt-test@latest
-
-      # - name: Download benchmark result for ${{ github.base_ref || github.ref_name }}
-      #   uses: dawidd6/action-download-artifact@v2
-      #   continue-on-error: true
-      #   with:
-      #     path: src/github.com/nats-io/nats-server/bench
-      #     name: bench-output-${{ runner.os }}
-      #     branch: ${{ github.base_ref || github.ref }}
+          go install github.com/ConnectEverything/mqtt-test@27cdbf032ee67b8d46015548fc360cf2eee4317a
```
levb, line 28: will change to a release ref once the mqtt-test is merged and tagged.

server/mqtt_ex_test_test.go
```diff
@@ -0,0 +1,569 @@
...
+		{
+			name:  "cluster with leafnode cluster no system account",
+			makef: mqttExMakeClusterWithLeafnodeCluster("HUBD", "LEAFD", false),
+		},
+	} {
+		t.Run(topo.name, func(t *testing.T) {
+			target := topo.makef(t)
+			t.Cleanup(target.Shutdown)
+
+			// TODO (levb): if we do not pre-initialize the streams and
```
levb, line 111: NS: cc @derekcollison @kozlovic This test uncovered a rather inconsistent behavior in multi-domain scenarios. If there are 2 domains, A and B, and the MQTT JetStream assets are first initialized in A, B will only receive and replay retained messages after its own retained message assets are created. Touching the servers with an MQTT connection here ensures the success of the tests; otherwise A<->B fail 1/2 of the time.

Not quite sure if this is by design.

kozlovic: I would think that if they have different domains, messages should not even be visible in the other domain, but I have to say I don't really understand that concept. What I notice is that you don't specify `js_domain: ` in the `mqtt{}` block of the config, but if I try to add those, then the test fails early since the lookup of a stream fails with a timeout since the JS APIs are rewritten with the domain name, like `$JS.HUBD.API.STREAM.INFO.$MQTT_sess` and for some reason, the server does not listen to those, even though you have specified `domain` in the JS config too. I don't really see in JS code where the server listens to that...

kozlovic (changes requested):
Some comments, but unfortunately regarding the domains, I am not sure how that works and can't really help there.

server/mqtt_ex_test_test.go
```diff
@@ -0,0 +1,569 @@
...
+					// publish numRMS retained messages, one at a time,
+					// round-robin across pub nodes.
+					iNode := 0
+					for i := 0; i < numRMS; i++ {
+						pubTopic := fmt.Sprintf("%s/%d", topic, i)
+						dial := tc.pub[iNode%len(tc.pub)]
+						mqttexRunTest(t, "pub", dial,
+							"--retain",
+							"--topic", pubTopic,
+							"--qos", "0",
```
kozlovic, line 138: Should it not be `"1"` here to persist a server restart (I tried, and still tests fail, but I still think it should be "1")

server/mqtt_ex_test_test.go
```diff
@@ -0,0 +1,569 @@
...
+							"--size", "100",
```
kozlovic, line 139: Should use `strNumRMS` here.

server/mqtt_ex_test_test.go
```diff
@@ -0,0 +1,569 @@
...
+	}
+	s = in
+	return u, p, s, c
+}
+
+func (t *mqttExTarget) Reload(tb testing.TB) {
+	tb.Helper()
+
+	for _, c := range t.clusters {
+		c.restartAllSamePorts()
```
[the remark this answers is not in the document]
kozlovic: Sorry, should call c.stopAll(), not c.shutdown() since in that case the config/store would be wiped out.

server/mqtt_ex_test_test.go
kozlovic: Actually, if I remember, we always store retained messages, even if they were QoS0. So my comment about qos needed to be "1" is likely wrong.

kozlovic (changes requested):

server/mqtt_ex_test_test.go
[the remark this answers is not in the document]
kozlovic: We need after the restart to add something like:
```
for _, stream := range []string{mqttSessStreamName, mqttStreamName, mqttQoS2IncomingMsgsStreamName, mqttOutStreamName, mqttRetainedMsgsStreamName} {
    c.waitOnStreamLeader("ONE", stream)
}
```
Without something like that, then it is possible that a connect fails because one of the stream lookup fails with a timeout (since after a restart, it may take some time to elect a leader for those streams).

server/mqtt_ex_test_test.go
```diff
@@ -0,0 +1,569 @@
...
+	all      []mqttExDial
+}
+
+type mqttExTestConfig struct {
+	name string
+	pub  []mqttExDial
+	sub  []mqttExDial
+}
+
+func TestMQTTExCompliance(t *testing.T) {
```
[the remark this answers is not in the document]
levb: I thought about it, but passed on a separate build tag. All the MQTTEx tests (should) do a `t.Skip` if `mqtt-test` is not installed in CI, so it's ok to invoke them in the main tag?

server/mqtt_ex_test_test.go
levb: Exactly. We persist them even for QOS0, but it may take longer than publishing them (since the publisher doesn't wait for acks). Originally, I coded a PUBRET single test that used the same QOS, and QOS0 was important to measure the receive performance without the acks.

For this test I broke PUB and SUB into separate steps, to test out delivery after a server restart, but I kept the QOS0 publishing + a simple SUB to wait for all publish's to be processed.

server/mqtt_ex_test_test.go
levb: No, it's the message size, 100 random bytes. I'll make it 128 instead, just to be more intention-revealing.

server/mqtt_ex_test_test.go
```diff
@@ -0,0 +1,569 @@
...
+func (t *mqttExTarget) Reload(tb testing.TB) {
+	tb.Helper()
+
+	for _, c := range t.clusters {
+		c.restartAllSamePorts()
+	}
+
+	for i, s := range t.servers {
+		o := s.getOpts()
+		s.Shutdown()
```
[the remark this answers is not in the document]
levb: It's the case at the moment, but you're right that it may not be the case in the future (a cluster with a single leaf node immediately comes to mind).

TBH, this was just a "minimal viable" implementation for the current requirements, I didn't want to cut/paste nor to go too abstract here (it's just an "extra" test for the time being). If we keep using MQTTEx going forward more then yes, the "target" implementation should be refactored.

server/mqtt_ex_test_test.go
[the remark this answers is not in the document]
levb: Ouch... fixing.

server/mqtt_ex_test_test.go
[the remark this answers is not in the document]
kozlovic: Right. I did not realize that this is what was happening. I installed mqtt-test, etc.. and added it to my path, but then when I was trying to run the "normal" MQTT tests, those 2 would run and I thought that was new.

kozlovic (changes requested):
Still need to make sure that the cluster is really restarted. And I have found then that tests are quite flappy. I think you need to have to maybe use testMQTTConnectRetry() with a clean session and dummy ID (or no ID specified) until it succeeds before having the real test in mqttexRunTest(). Just waiting on the cluster to be ready proved not enough in tests I was running last week.

levb:
> Still need to make sure that the cluster is really restarted. 

@kozlovic sorry my commit message was non-descriptive, I squashed from another branch. Anyway, I am still working on the restart a another potential issue I ran into, will mark you for review when done.

server/mqtt_ex_test_test.go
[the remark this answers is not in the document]
levb: ok, will separate.

kozlovic (approved):
LGTM

PR #5083, opened 2024-02-14
[IMPROVED] Expire cache when firstMatching does not match anything.

derekcollison:
When calling into firstMatching, if we do not find anything make sure to expire the cache if we need to load fss state.


neilalexander (approved):
LGTM

wallyqs (approved):
LGTM

PR #5085, opened 2024-02-14
[IMPROVED] Move reallocation warnings on psim to debug.

derekcollison:
They should not happen, but it is possible and not actionable by operators, so debug is better.


wallyqs (approved):
LGTM

PR #5088, opened 2024-02-15
[IMPROVED] Startup time for a start sequence by time filtered consumer.

derekcollison:
When a filtered consumer starts by time, there could be a large gap between the time to sequence and the real start.

This uses similar logic to starting from the beginning of a stream for a filtered consumer.


wallyqs (approved):
LGTM

PR #5089, opened 2024-02-15
[IMPROVED] Filtered consumers retrieval speed.

derekcollison:
When doing firstMatching we were picking wrong on linearScans. We also did not optimize the subjectIsSubsetMatch functionality.


wallyqs (approved):
LGTM

PR #5091, opened 2024-02-15
Fix flapping in TestJetStreamMirrorBasics

jnmoyne:
Fix test that could flap with the following kind  error in TestJetStreamMirrorBasics:
```
jetstream_test.go:11608: Expected messages with the transformed subject foo2
jetstream_test.go:11611: Expected 100 messageson  the transformed subject foo2 but got 13
```

Also saw a context timeout happening on a Stream Info in another part of the test, which could also flap sometimes as the timeout was pretty short at 250ms so relaxed it to 500ms and haven't seen that flap since.

### DIFF
--- a/server/jetstream_test.go
+++ b/server/jetstream_test.go
@@ -11437,7 +11437,7 @@ func TestJetStreamMirrorBasics(t *testing.T) {
 	}
 
 	// Faster timeout since we loop below checking for condition.
-	js2, err := nc.JetStream(nats.MaxWait(250 * time.Millisecond))
+	js2, err := nc.JetStream(nats.MaxWait(500 * time.Millisecond))
 	require_NoError(t, err)
 
 	checkFor(t, 5*time.Second, 250*time.Millisecond, func() error {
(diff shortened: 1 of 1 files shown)


server/jetstream_test.go
```diff
@@ -11605,25 +11605,25 @@ func TestJetStreamMirrorBasics(t *testing.T) {
 			si, err := js2.StreamInfo(streamName, &nats.StreamInfoRequest{SubjectsFilter: subject})
 			require_NoError(t, err)
 			if ss, ok := si.State.Subjects[subject]; !ok {
-				t.Logf("Expected messages with the transformed subject %s", subject)
+				return fmt.Errorf("expected messages with the transformed subject %s", subject)
 			} else {
-				if ss != subjectNumMsgs {
-					t.Fatalf("Expected %d messages on the transformed subject %s but got %d", subjectNumMsgs, subject, ss)
+				if ss > subjectNumMsgs || ss < subjectNumMsgs {
```
derekcollison, line 11610: This is just `ss !- subjectNumMsgs` like it was originally.

jnmoyne: Yes but it now properly returns an error (it wasn't using checkFor correctly by not returning an error which caused checkFor to stop trying again).

derekcollison: Yes saw that, but the if is non-sensical, its just `!=`

jnmoyne: Changed

wallyqs:
you need to rebase to get the mqtt fix

jnmoyne:
After restarting a couple of times the MQTT tests I was figuring out there would probably be a fix for those tests if I rebased, as confirmed with Wally's comment above. Now rebased.

derekcollison (approved):
LGTM

PR #5093, opened 2024-02-15
Made updates to message tracing code to support JWT updates

kozlovic:
Related to #5014


derekcollison (approved):
LGTM

ripienaar:
@kozlovic will these settings also support config reload or do we need to do more to support that?

kozlovic:
When configured locally, config reload is supported (see tests where I use `reloadUpdateConfig`). For JWTs, they are updated using the account resolver, and this is what this PR was about. So in short, config reload is supported and nothing else to do.

PR #5100, opened 2024-02-16
Switch to use fmt.Appendf in stream.go and consumer.go

wallyqs:
Present since Go v1.19, uses less allocs.


derekcollison (approved):
LGTM

PR #5107, opened 2024-02-19
[ADDED] Add support for multi-get directly from a stream.

derekcollison:
This allows a client to get multiple results based on multiple filters, each which can include a wildcard. You can optionally specify an UpToSeq or UpToTime.

The results will be returned in sequential order.

They will have headers matching direct get batch as well, with num pending being how many more results are coming. Currently we limit to 64 results maximum, and may revisit this in the future.

### DIFF
--- a/server/jetstream_api.go
+++ b/server/jetstream_api.go
@@ -649,6 +649,13 @@ type JSApiMsgGetRequest struct {
 	// This will make sure we limit how much data we blast out. If not set we will
 	// inherit the slow consumer default max setting of the server. Default is MAX_PENDING_SIZE.
 	MaxBytes int `json:"max_bytes,omitempty"`
+
+	// Multiple response support. Will get the last msgs matching the subjects. These can include wildcards.
+	MultiLastFor []string `json:"multi_last,omitempty"`
+	// Only return messages up to this sequence. If not set, will be last sequence for the stream.
+	UpToSeq uint64 `json:"up_to_seq,omitempty"`
+	// Only return messages up to this time.
+	UpToTime *time.Time `json:"up_to_time,omitempty"`
 }
 
 type JSApiMsgGetResponse struct {
--- a/server/jetstream_test.go
+++ b/server/jetstream_test.go
@@ -1,4 +1,4 @@
-// Copyright 2019-2023 The NATS Authors
+// Copyright 2019-2024 The NATS Authors
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
 // You may obtain a copy of the License at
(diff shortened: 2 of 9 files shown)


server/jetstream_test.go
```diff
@@ -22402,3 +22402,340 @@ func TestJetStreamSubjectFilteredPurgeClearsPendingAcks(t *testing.T) {
...
+	checkResponses(sub, "1", "foo.foo", "foo.bar", "foo.baz", _EMPTY_)
+
+	sub = sendRequest(&JSApiMsgGetRequest{Seq: 1, MultiLastFor: []string{"foo.*"}, UpToTime: &mid})
+	checkResponses(sub, "2", "foo.foo", "foo.bar", "foo.baz", _EMPTY_)
+
+	sub = sendRequest(&JSApiMsgGetRequest{Seq: 1, MultiLastFor: []string{"foo.*"}, UpToTime: &end})
+	checkResponses(sub, "3", "foo.foo", "foo.bar", "foo.baz", _EMPTY_)
+}
+
+func TestJetStreamDirectGetMultiMaxAllowed(t *testing.T) {
```
ripienaar, line 22612: I am curious why this behaviour, why would we ever want to fail in this way? Shouldnt it just behave as if batch was set always? If I ever ask more than is allowed just act as if I reached end of batch, we wouldnt ever say you exceeded the allowed responses, its just always paged?

That seems like it would be a bit more consistent

server/jetstream_test.go
```diff
@@ -22402,3 +22402,340 @@ func TestJetStreamSubjectFilteredPurgeClearsPendingAcks(t *testing.T) {
...
+		require_NoError(t, err)
+		return sub
+	}
+
+	// Setup variables that control procesPartial
+	start, seq, np, b, bsz := 1, 1, sent-1, 0, 128
+
+	processPartial := func(expected int) {
+		//		t.Helper()
+		sub := sendRequest(&JSApiMsgGetRequest{Seq: uint64(start), Batch: b, MultiLastFor: []string{"foo.*"}})
```
ripienaar, line 22699: as above comment in `TestJetStreamDirectGetMultiMaxAllowed` if this was just always batched behaviour we wouldnt need to also bring in `Batch` here into the fold, it seems redundant.  Multi just a batch that support multiple subjects.

ripienaar (commented):
Behaviour wise I find the batch/unbatched behaviour weird as pointed out in comments - it should just always be a batch imo.

To continue should we hit batch limits I think we need a bit more info.

Imagine I ask for `USER.1234.>` and this user has a ton of fields, more than 1000 or maxBytes, it would unexpectedly have to continue with another get to get the full user.

But how do I continue in a consistent way? I have no idea what the time or seq is to get it relative against? 

So imagine if the EOB message included the `sseq` pass to `UpToSeq` for the next multi get. 

In this case the `sseq` would be either the last message in the stream at the time that you gathered the seqs (as in the comment for MultiLastSeq in file store) or whatever the UpToSeq was or whatever UpToTime resolved the seq to.  This way if there is a EOB and I need to get more keys to get the whole thing I immediately know that the next Multi get should have `UpToSeq: sseq` and all the many reads needed to resolve the entire `USER.1234.>` would be done around a consistent time.

server/jetstream_api.go
```diff
@@ -641,6 +641,13 @@ type JSApiMsgGetRequest struct {
 	// This will make sure we limit how much data we blast out. If not set we will
 	// inherit the slow consumer default max setting of the server. Default is MAX_PENDING_SIZE.
 	MaxBytes int `json:"max_bytes,omitempty"`
+
+	// Multiple response support. Will get the last msgs matching the subjects. These can include wildcards.
+	MultiLastFor []string `json:"multi_last,omitempty"`
+	// Only return messages up to this sequence. If not set, will be last sequence for the stream.
```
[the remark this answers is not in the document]
derekcollison: You mean more so that line 645?

server/jetstream_test.go
```diff
@@ -22402,3 +22402,340 @@ func TestJetStreamSubjectFilteredPurgeClearsPendingAcks(t *testing.T) {
...
+						require_Equal(t, msg.Header.Get("Status"), "204")
+						// Check description is EOB
+						require_Equal(t, msg.Header.Get("Description"), "EOB")
+						// Check we have NumPending and its correct.
+						require_Equal(t, strconv.Itoa(np), msg.Header.Get(JSNumPending))
+					}
+				}
+			}
+
+			sub := sendRequest(&JSApiMsgGetRequest{Seq: 1, MultiLastFor: []string{"foo.*"}})
```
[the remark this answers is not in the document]
derekcollison: If it is a one-shot can be left alone and we start with 1, which is how far back should I look for a match. These is important one paged responses.

server/jetstream_test.go
derekcollison: What is stream has 10M msgs all are unique and you do a multi for ">". IMO we need to protect the system and not bloat up, so we fail here instead of allowing you to page through 10M msgs.

server/jetstream_test.go
derekcollison: This just shows we support batch in terms of limiting the one shot responses. So a user could say get multi but only give me 10 at a time, etc.

server/jetstream_test.go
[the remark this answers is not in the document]
derekcollison: Got what you meant, will fix.

server/jetstream_test.go
```diff
@@ -22740,3 +22740,341 @@ func TestJetStreamConsumerSurvivesRestart(t *testing.T) {
...
+	}
+
+	req, _ := json.Marshal(&JSApiMsgGetRequest{Seq: 1, MultiLastFor: []string{"foo.*"}})
+	msg, err := nc.Request("$JS.API.DIRECT.GET.TEST", req, time.Second)
+	require_NoError(t, err)
+
+	// Check for properly formatted Too Many Results error.
+	// Should have no body.
+	require_Equal(t, len(msg.Data), 0)
+	// We mark status as 204 - No Content
```
ripienaar, line 22982: mark it as `413` in fact with `Too Many Results`

ripienaar (approved):
Behaviour wise this looks great, so LGTM.  Minor comment on code, but get server team to review code.

ripienaar:
ADR update that attempts to capture the behaviour here https://github.com/nats-io/nats-architecture-and-design/pull/268

server/jetstream_test.go
derekcollison: Thanks - fixed.

PR #5109, opened 2024-02-20
ocsp: make ocsp request with POST method

wallyqs:
This makes the server send a POST request to the OCSP Server, then falling back to use GET as in previous versions.

### DIFF
--- a/server/ocsp.go
+++ b/server/ocsp.go
@@ -14,12 +14,14 @@
 package server
 
 import (
+	"bytes"
 	"crypto/sha256"
 	"crypto/tls"
 	"crypto/x509"
 	"encoding/asn1"
 	"encoding/base64"
 	"encoding/pem"
+	"errors"
 	"fmt"
 	"io"
 	"net/http"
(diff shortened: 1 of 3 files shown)


derekcollison:
Why do we need this functionality?

philpennock:
Derek: interoperability with various network setups, such that a GET request might not succeed and a POST gets a cleaner (HTTP-caching-wise) path to the origin and is less tampered with.  Also, RFC 6960 has size limits on the request (since the DER gets base64 encoded into the URL, which can have size limits) so using POST avoids those size limits and can work if using larger RSA keys.

derekcollison:
Should we just support POST vs this weird duality or is that just table stakes for OCSP ecosystem etc?

wallyqs:
POST is more common from what I've seen so we could change to that instead.

server/ocsp.go
```diff
@@ -162,15 +163,40 @@ func (oc *OCSPMonitor) getRemoteStatus() ([]byte, *ocsp.Response, error) {
...
+		start := time.Now()
+		resp, err := hc.Do(hreq)
+		if err != nil {
+			return nil, err
+		}
+		defer resp.Body.Close()
 
+		oc.srv.Debugf("Received OCSP response (method=POST, status=%v, url=%s, duration=%.3fs)",
+			resp.StatusCode, u, time.Since(start).Seconds())
+		if resp.StatusCode != http.StatusOK {
```
philpennock, line 197: Shouldn't this be `if resp.StatusCode > 299` to return an error and then perhaps if it's not 200 then log it as being of interest but not an error?

server/ocsp.go
```diff
@@ -195,12 +219,21 @@ func (oc *OCSPMonitor) getRemoteStatus() ([]byte, *ocsp.Response, error) {
...
 		u = strings.TrimSuffix(u, "/")
-		raw, err = getRequestBytes(fmt.Sprintf("%s/%s", u, reqEnc), hc)
+		// Prefer to make POST requests first.
+		raw, err = postRequestBytes(u, reqDER, hc)
 		if err == nil {
 			break
+		} else {
+			// Fallback to use a GET request.
+			reqEnc := base64.StdEncoding.EncodeToString(reqDER)
+			raw, err = getRequestBytes(fmt.Sprintf("%s/%s", u, reqEnc), hc)
```
philpennock, line 233: If the POST is more reliable and the GET fails too, this loses our POST error response.  Any chance we can use `errors.Join()` to return both errors, instead of risking masking a solid POST error response with a poor HTTP proxy GET error response which doesn't tell us why the POST failed in the first place?

philpennock (approved):

server/ocsp.go
wallyqs: I will make this change.

server/ocsp.go
wallyqs: good idea, I applied these fixes.

wallyqs:
Applied the changes so now ready.

server/ocsp.go
```diff
@@ -161,16 +163,43 @@ func (oc *OCSPMonitor) getRemoteStatus() ([]byte, *ocsp.Response, error) {
 	if config := opts.OCSPConfig; config != nil {
 		overrideURLs = config.OverrideURLs
 	}
-	getRequestBytes := func(u string, hc *http.Client) ([]byte, error) {
+	getRequestBytes := func(u string, reqDER []byte, hc *http.Client) ([]byte, error) {
```
derekcollison, line 166: Do we need Get as well here? Do all OCSP support POST? Should we just support that?

philpennock: In theory they should, but the problem is "do they really, and do the applicaiton firewalls permit it", inside corporate environments.  So since we have been using GET, I think the goal is to make sure that worst case scenario, for things we don't control, we don't cause a regression in system behavior.

wallyqs: They should support both, but we have used GETs only so far so in case there was a firewall or something depending on that behavior already we do the fallback just in case.

server/ocsp.go
```diff
@@ -195,18 +222,28 @@ func (oc *OCSPMonitor) getRemoteStatus() ([]byte, *ocsp.Response, error) {
...
-		if err == nil {
+		// Prefer to make POST requests first.
+		raw, postErr = postRequestBytes(u, reqDER, hc)
+		if postErr == nil {
 			break
+		} else {
+			// Fallback to use a GET request.
+			raw, getErr = getRequestBytes(u, reqDER, hc)
+			if getErr == nil {
+				break
```
philpennock, line 238: I think you need an `err = nil` before the break here: if the POST fails but the GET succeeds, we want to proceed.

philpennock (changes requested):

derekcollison (approved):
LGTM

derekcollison:
@philpennock you had a block on this so re-requested your review to make sure its cleared.

philpennock:
> @philpennock you had a block on this so re-requested your review to make sure its cleared.

Yes, the block is 6 minutes old because it's for a bug introduced in the latest push. :D

derekcollison:
@wallyqs ping me when ready and cleared.

server/ocsp.go
wallyqs: nice catch, it was also failing when trying multiple ocsp servers urls after one failed and the following one succeeded. This is fixed now.

derekcollison:
ok @philpennock let me know when you have LGTM'd.

philpennock (approved):

derekcollison:
Will wait on green then merge..

PR #5116, opened 2024-02-21
Fix for #5098

### LINKED ISSUE #5098: NATS JetStream slowing down over time
### Observed behavior

We see a constant degradation of E2E Latency over time (~24 test period) when having a constant load on the a single jetstream.
The degradation mainly can be seen in outliers - so the avg latency is not so much affected yet, but that will continue to degrade over time as well.

We see that, based on our latency Grafana DBs, that the degradtion comes from the producers side - The producer waits until a publish call is confirmed by the NATS cluster and then goes on. This duration is also monitored as a histogram as follows:
`lastPublish = System.nanoTime();`
`PublishAck pa = js.publish(msg, opts);`
`long deltaNano = System.nanoTime() - lastPublish;`

streamTimeout for publishoptions is set to 2 secs. deltaNano is monitored as a promehteus histogram.

Garafan DB as a proof: 
E2E Latency:
image
Producer Latency:
image

Note: In the middle (~18:00) we stopped prod & consumers to verify if the JetStream recovers when the load is gone. Apparently it did not as the latency outliers remained where the were when restarting producers/consumers. 
Note 2: We then started the exact same test again with another JetStream without restarting the NATS cluster. The latency was again fine in the beginning and followed the patterns as can be seen above. 
Note 3: Unfornately the producer latency histogram is capped in that test which is why the increase then is not visible any longer in the +INF bucket.

### Expected behavior

There is no continuous degration of E2E Latency over time. 

### Server and client version

Server. = 2.10.10
Client = Java v2.17.3

### Host environment

Openshift using NATS provided HELM charts

### Steps to reproduce

Single Jetstream on a 3 node cluster with R3.
10 Producers, 10 Consumers and 10 subjects (like test.1, test.2, test.10). Every producer/consumer form a pair (1:1 relation via subject). 
Producers provide messages in a constant rate of 100msg/sec. 

JetStream settings:
` StreamConfiguration sc = StreamConfiguration.builder()`
`                .name(streamName) ` 
`               .storageType(StorageType.Memory)`
`                .subjects(subjects)`
`                .retentionPolicy(RetentionPolicy.Limits)`
`                .discardPolicy(DiscardPolicy.Old)`
`                .replicas(3)`
`                .maxAge(Duration.ofMinutes(10))`
`                .maxBytes(500_000_000) //500MB`
`                .maxMessagesPerSubject(100_000)`
`                .build();`

Consumer Configuration:
` ConsumerConfiguration.Builder consumerBuilder = ConsumerConfiguration.builder()`
`                    .numReplicas(3)`
`                    .ackPolicy(AckPolicy.Explicit)`
`                    .name(exArgs.consumerName)`
`                    .ackWait(Duration.ofSeconds(5))`
`                    .inactiveThreshold(Duration.ofSeconds(60))`
`                    .filterSubject(exArgs.subject)`
`                    .deliverPolicy(DeliverPolicy.All)`
`                    .memStorage(true);`

Consumer is a pull consumer using "fetch"
`try (FetchConsumer consumer = consumerContext.fetch(consumeOptions)) {`

The paylod of messages is random and has the format:
SequenceNr + space + CurrentTimeMillis + space + String with 100 random characters 
UTF-coded to byte array.

stefanLeo:
This PR fixes #5098 and includes a testcase for it. 

Signed-off-by: Stefan Leonhartsberger <stefanleo>

### DIFF
--- a/server/memstore_test.go
+++ b/server/memstore_test.go
@@ -17,6 +17,7 @@ import (
 	"bytes"
 	"errors"
 	"fmt"
+	"math"
 	"math/rand"
 	"reflect"
 	"testing"
@@ -87,6 +88,67 @@ func TestMemStoreMsgLimit(t *testing.T) {
 	// Make sure we can not lookup seq 1.
 	if _, err := ms.LoadMsg(1, nil); err == nil {
 		t.Fatalf("Expected error looking up seq 1 but got none")
+	}
+}
+
+func TestMemStorePerformance(t *testing.T) {
+	subj, msg := "foo", make([]byte, 1000)
+	storedMsgSize := memStoreMsgSize(subj, nil, msg)
+
+	toStore := uint64(500_000)
+	maxBytes := storedMsgSize * toStore
+
+	ms, err := newMemStore(&StreamConfig{Storage: MemoryStorage, MaxBytes: int64(maxBytes)})
+	require_NoError(t, err)
+	defer ms.Stop()
+
+	for i := uint64(0); i < toStore; i++ {
+		ms.StoreMsg(subj, nil, msg)
+	}
+	state := ms.State()
+	if state.Msgs != toStore {
+		t.Fatalf("Expected %d msgs, got %d", toStore, state.Msgs)
+	}
+	if state.Bytes != storedMsgSize*toStore {
+		t.Fatalf("Expected bytes to be %d, got %d", storedMsgSize*toStore, state.Bytes)
+	}
+
+	iteration := 10
+	msgAddPerIteration := 1_000_000
+	maxTime := int64(0)
+	minTime := int64(math.MaxInt64)
+	maxIteration := 0
+	for j := 0; j < iteration; j++ {
+		// Now send 10 more and check that bytes limit enforced.
+		for i := 0; i < msgAddPerIteration; i++ {
+			if _, _, err := ms.StoreMsg(subj, nil, msg); err != nil {
+				t.Fatalf("Error storing msg: %v", err)
+			}
+		}
+		compactStart := toStore + uint64((j+1)*msgAddPerIteration) - 200_000
+		start := time.Now()
+		purged, errorVal := ms.Compact(compactStart)
+		end := time.Now()
+		elapsed := end.Sub(start)
+		fmt.Printf("Compact took: %d us in iteration %d. purged = %d. error: %s", elapsed.Microseconds(), j, purged, errorVal)
+		fmt.Println()
+		if purged != 299_999 {
+			t.Fatalf("unexpected value of purge. Expected %d vs. Received %d", 299_999, purged)
+		}
+
+		if maxTime < elapsed.Microseconds() {
+			maxTime = elapsed.Microseconds()
+			maxIteration = j
+		}
+		if minTime > elapsed.Microseconds() {
+			minTime = elapsed.Microseconds()
+		}
+	}
+	delta := (maxTime - minTime) / 1000.0
+	fmt.Printf("Compact took max: %d us on iteration %d out of %d. Delta between min/max %d msec", maxTime, maxIteration, iteration, delta)
+	fmt.Println()
+	if delta > 100 {
+		t.Fatalf("Compact duration varied to much - Max %d usec and Min %d usec duration have to be within 100 msec but is %d msec", maxTime, minTime, delta)
 	}
 }
 
--- a/server/memstore.go
+++ b/server/memstore.go
@@ -792,10 +792,11 @@ func (ms *memStore) Compact(seq uint64) (uint64, error) {
 			ms.mu.Unlock()
 			return 0, ErrStoreMsgNotFound
 		}
+		oldFirstSeq := ms.state.FirstSeq - 1
 		ms.state.FirstSeq = seq
 		ms.state.FirstTime = time.Unix(0, sm.ts).UTC()
 
-		for seq := seq - 1; seq > 0; seq-- {
+		for seq := seq - 1; seq > oldFirstSeq; seq-- {
 			if sm := ms.msgs[seq]; sm != nil {
 				bytes += memStoreMsgSize(sm.subj, sm.hdr, sm.msg)
 				purged++
--- a/server/memstore.go
+++ b/server/memstore.go
@@ -792,11 +792,11 @@ func (ms *memStore) Compact(seq uint64) (uint64, error) {
 			ms.mu.Unlock()
 			return 0, ErrStoreMsgNotFound
 		}
-		oldFirstSeq := ms.state.FirstSeq - 1
+		fseq := ms.state.FirstSeq
 		ms.state.FirstSeq = seq
 		ms.state.FirstTime = time.Unix(0, sm.ts).UTC()
 
-		for seq := seq - 1; seq > oldFirstSeq; seq-- {
+		for seq := seq - 1; seq >= fseq; seq-- {
 			if sm := ms.msgs[seq]; sm != nil {
 				bytes += memStoreMsgSize(sm.subj, sm.hdr, sm.msg)
 				purged++


server/memstore_test.go
```diff
@@ -90,6 +91,67 @@ func TestMemStoreMsgLimit(t *testing.T) {
 	}
 }
 
+func TestMemStorePerformance(t *testing.T) {
```
derekcollison, line 94: Let's move this to norace_test.go and follow correct naming conventions there.

server/memstore_test.go
```diff
@@ -90,6 +91,67 @@ func TestMemStoreMsgLimit(t *testing.T) {
...
+			if _, _, err := ms.StoreMsg(subj, nil, msg); err != nil {
+				t.Fatalf("Error storing msg: %v", err)
+			}
+		}
+		compactStart := toStore + uint64((j+1)*msgAddPerIteration) - 200_000
+		start := time.Now()
+		purged, errorVal := ms.Compact(compactStart)
+		end := time.Now()
+		elapsed := end.Sub(start)
+		fmt.Printf("Compact took: %d us in iteration %d. purged = %d. error: %s", elapsed.Microseconds(), j, purged, errorVal)
```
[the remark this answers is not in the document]
stefanLeo: I updated/moved the test. Let me know if this is what you expected.

server/memstore_test.go
stefanLeo: done

server/norace_test.go
```diff
@@ -9599,3 +9599,58 @@ func TestNoRaceJetStreamKVReplaceWithServerRestart(t *testing.T) {
...
+
+	toStoreOnTop := uint64(10_000_000)
+
+	//1st run: Load additional 10 Mio messages then compact
+	for i := uint64(0); i < toStoreOnTop; i++ {
+		ms.StoreMsg(subj, nil, msg)
+	}
+	startFirstRun := time.Now()
+	purgedFirstRun, _ := ms.Compact(10_300_000)
+	elapsedFirstRun := time.Now().Sub(startFirstRun).Microseconds()
```
neilalexander, line 9634: CI is currently failing on the linter because of these:
```
server/norace_test.go:9634:21: S1012: should use `time.Since` instead of `time.Now().Sub` (gosimple)
	elapsedFirstRun := time.Now().Sub(startFirstRun).Microseconds()
	                   ^
server/norace_test.go:9645:22: S1012: should use `time.Since` instead of `time.Now().Sub` (gosimple)
	elapsedSecondRun := time.Now().Sub(startSecondRun).Microseconds()
	                    ^
The command "./scripts/runTestsOnTravis.sh $TEST_SUITE" exited with 1.
```

stefanLeo: thx for pointing out. fixed.

server/norace_test.go
```diff
@@ -9599,3 +9599,58 @@ func TestNoRaceJetStreamKVReplaceWithServerRestart(t *testing.T) {
...
+
+	ms, err := newMemStore(&StreamConfig{Storage: MemoryStorage, MaxBytes: int64(maxBytes)})
+	require_NoError(t, err)
+	defer ms.Stop()
+
+	for i := uint64(0); i < toStore; i++ {
+		ms.StoreMsg(subj, nil, msg)
+	}
+	state := ms.State()
+	if state.Msgs != toStore {
```
derekcollison, line 9619: `require_Equal(t, state.Msgs, toStore)`

server/norace_test.go
```diff
@@ -9599,3 +9599,58 @@ func TestNoRaceJetStreamKVReplaceWithServerRestart(t *testing.T) {
...
+		t.Fatalf("Expected %d msgs, got %d", toStore, state.Msgs)
+	}
+	if state.Bytes != storedMsgSize*toStore {
+		t.Fatalf("Expected bytes to be %d, got %d", storedMsgSize*toStore, state.Bytes)
```
derekcollison, line 9623: Same here.

server/norace_test.go
```diff
@@ -9599,3 +9599,58 @@ func TestNoRaceJetStreamKVReplaceWithServerRestart(t *testing.T) {
...
+	}
+
+	toStoreOnTop := uint64(10_000_000)
+
+	//1st run: Load additional 10 Mio messages then compact
```
derekcollison, line 9628: Mio?

server/norace_test.go
```diff
@@ -9599,3 +9599,58 @@ func TestNoRaceJetStreamKVReplaceWithServerRestart(t *testing.T) {
...
+	for i := uint64(0); i < toStoreOnTop; i++ {
+		ms.StoreMsg(subj, nil, msg)
+	}
+	startFirstRun := time.Now()
+	purgedFirstRun, _ := ms.Compact(10_300_000)
+	elapsedFirstRun := time.Since(startFirstRun).Microseconds()
+	if purgedFirstRun != 299_999 {
```
derekcollison, line 9635: Again require_Equal is our standard for these.

server/norace_test.go
```diff
@@ -9599,3 +9599,58 @@ func TestNoRaceJetStreamKVReplaceWithServerRestart(t *testing.T) {
...
+	}
+
+	//2nd run: Load additional 10 Mio messages then compact
+	for i := uint64(0); i < toStoreOnTop; i++ {
+		ms.StoreMsg(subj, nil, msg)
+	}
+	startSecondRun := time.Now()
+	purgedSecondRun, _ := ms.Compact(20_300_000)
+	elapsedSecondRun := time.Since(startSecondRun).Microseconds()
+	if purgedSecondRun != 299_999 {
```
derekcollison, line 9646: require_Equal

server/norace_test.go
```diff
@@ -9599,3 +9599,58 @@ func TestNoRaceJetStreamKVReplaceWithServerRestart(t *testing.T) {
...
+		t.Fatalf("unexpected value of purge. Expected %d vs. Received %d", 299_999, purgedSecondRun)
+	}
+
+	//Calculate delta between runs and fail if it is too high
+	delta := (elapsedSecondRun - elapsedFirstRun) / 1000.0
```
derekcollison, line 9651: Use time helpers here and truncate to seconds.

server/norace_test.go
```diff
@@ -9599,3 +9599,58 @@ func TestNoRaceJetStreamKVReplaceWithServerRestart(t *testing.T) {
...
+	if delta > 100 {
```
derekcollison, line 9652: Compare against the time domain vs naked number.

derekcollison (changes requested):
We are getting there! 

Between first and second run, do a PurgeEx with a very high sequence number, say 10B or so, then put messages into that stream. That will really show the issue we are trying to fix. After the purge the first sequence will be 10B..

server/norace_test.go
stefanLeo: Understood & Fixed

server/norace_test.go
stefanLeo: Understood & Fixed

server/norace_test.go
stefanLeo: Understood & Fixed

server/norace_test.go
stefanLeo: Fixed

server/norace_test.go
stefanLeo: I will think about a lower numer. Maybe be just removing N-1 as initially stated by you. 
I just wanna make sure that this test never breaks as timing related tests have the tendency to do so if the build server is slightly slower as someone is doing an update etc... I.e., I want to make sure not to hand over a test that breaks randomly every now and then :-) 
With 1 mio entries I ran into false positives on my local machine already... 
With purgeex of course the results may be more stable - i will try this

server/norace_test.go
stefanLeo: changed to compare duration objects. Hope that's what you meant.

stefanLeo:
@neilalexander : Can you help me with the failing build? The failed tests are not at all related to my change?

neilalexander:
I kicked the failing tests, there are some flaky ones but looks green now.

server/norace_test.go
```diff
@@ -9599,3 +9599,47 @@ func TestNoRaceJetStreamKVReplaceWithServerRestart(t *testing.T) {
...
+
+	//1st run: Load additional 10 Mio messages then compact
+	for i := uint64(0); i < toStoreOnTopFirst; i++ {
+		ms.StoreMsg(subj, nil, msg)
+	}
+	startFirstRun := time.Now()
+	purgedFirstRun, _ := ms.Compact(toStore + toStoreOnTopFirst)
+	elapsedFirstRun := time.Since(startFirstRun)
+	require_Equal(t, expectedPurge, purgedFirstRun)
+
```
derekcollison, line 9633: Let's do a PurgeEx here as I have mentioned to bump the first sequence very high.

derekcollison (commented):
Once we have a purge to bump the first sequence before the second addition of messages and compact we should be good.

server/norace_test.go
stefanLeo: I changed it to what you have in mind (not using PurgeEx) as imho PurgeEx would anyhow only call compact. 
The test now has only one iteration, compacts the map and then recompacts it, which then has to be a noop.

derekcollison:
The PurgeEx call with a very high sequence beyond the last sequence of the stream will bump the first sequence to that value.

What I want is this test to very clearly show we do not iterate all the way to 0, so setting this extremely high but compacting the same amount on the second compact run will be a better test.

stefanLeo:
I think I got it now ... Sorry for these iterations ;-)

derekcollison (approved):
LGTM

PR #5121, opened 2024-02-22
Added sampling for account's trace destination

kozlovic:
If an account has a trace destination and an incoming message has the `traceparent` header with the proper sampled flag, a trace message would be triggered. The `sampling` field allows to trace a certain percentage of the traffic.

The field `trace_dest` or now `msg_trace` can be a simple string representing the destination, and in this case sampling is 100% or it can be a structure with the `dest` and `sampling` fields. Sampling values that are negative or above 100 will trigger an error on configuration parsing. A value of 0 is converted to 100. If the sampling is specified without an account trace destination, it is set to 0 and a warning is issued when parsing configuration.

There is similar support for the property set in a JWT account claim.

Relates to #5014

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -5,7 +5,7 @@ go 1.20
 require (
 	github.com/klauspost/compress v1.17.6
 	github.com/minio/highwayhash v1.0.2
-	github.com/nats-io/jwt/v2 v2.5.4
+	github.com/nats-io/jwt/v2 v2.5.5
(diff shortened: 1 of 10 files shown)


derekcollison (approved):
LGTM

derekcollison:
Will wait on @ripienaar for merge.

ripienaar (approved):
LGTM

PR #5125, opened 2024-02-23
[FIXED] MQTT: (in v2.11) header length calculation for stored retained messages

levb:
v2.11 only.


derekcollison (approved):
LGTM

kozlovic (approved):
LGTM

levb:
@derek there was another discrepancy there, resulting in an extra allocation; fixed that too.

derekcollison (approved):
LGTM

Issue #5117, opened 2024-02-22
improvements in leaf loop detection logic


borjaetxebarria:
### Observed behavior

loop detection logic in leaf nodes can be defeated in simple setups   
Setup:  
B --[leaf]--> A  
C --[leaf]--> A    
C --[leaf] --> B

with the right permissions and startup sequence, the loop is not detected

### Expected behavior

loop is detected

### Server and client version

2.10.11

### Host environment

docker compose

### Steps to reproduce

see attached docker compose files and instructions to reproduce the issue
[leaf-loops.tgz](https://github.com/nats-io/nats-server/files/14368289/leaf-loops.tgz)

derekcollison:
Appreciated, but I mean a server test. Running inside the system tests under server directory. Those are most actionable. In absence the server team needs to write them almost as a first step.

derekcollison:
I can write the test, will queue up for my weekend work.

derekcollison:
With a simple server test doing what you are doing, C detects the loop right away, so must be timing based. Will slow down the starting of the servers and see if that trips it.

derekcollison:
If I delay between B and A for 2s I start to see it.

derekcollison:
ok the bug is when C connect to A first before B does. I can use staggered reconnect intervals to force this.

PR #5126, opened 2024-02-25
[FIXED] Better loop detection on daisy chained leafnodes.

derekcollison:
Make sure to always update smap on LDS subjects to forward across multiple hops.
This way when a loop is created later on we can still detect it.
Also detect loops on leafnode connect if a cluster is already registered.

Resolves: #5117 

### DIFF
--- a/server/accounts.go
+++ b/server/accounts.go
@@ -928,6 +928,13 @@ func (a *Account) registerLeafNodeCluster(cluster string) {
 	a.leafClusters[cluster]++
 }
 
+// Check to see if we already have this cluster registered.
+func (a *Account) hasLeafNodeCluster(cluster string) bool {
+	a.mu.Lock()
+	defer a.mu.Unlock()
+	return a.leafClusters[cluster] > 0
+}
+
 // Check to see if this cluster is isolated, meaning the only one.
 // Read Lock should be held.
 func (a *Account) isLeafNodeClusterIsolated(cluster string) bool {
--- a/server/leafnode.go
+++ b/server/leafnode.go
@@ -2031,8 +2031,8 @@ func (s *Server) initLeafNodeSmapAndSendSubs(c *client) {
 		c.leaf.smap[oldGWReplyPrefix+"*.>"]++
 		c.leaf.smap[gwReplyPrefix+">"]++
 	}
-	// Detect loop by subscribing to a specific subject and checking
-	// if this is coming back to us.
+	// Detect loops by subscribing to a specific subject and checking
+	// if this sub is coming back to us.
 	c.leaf.smap[lds]++
 
 	// Check if we need to add an existing siReply to our map.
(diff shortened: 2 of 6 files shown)


server/leafnode.go
```diff
@@ -2125,11 +2125,11 @@ func (acc *Account) updateLeafNodes(sub *subscription, delta int32) {
...
 		// Check to make sure this sub does not have an origin cluster that matches the leafnode.
 		ln.mu.Lock()
-		skip := (cluster != _EMPTY_ && cluster == ln.remoteCluster()) || (delta > 0 && !ln.canSubscribe(subject))
 		// If skipped, make sure that we still let go the "$LDS." subscription that allows
-		// the detection of a loop.
-		if isLDS || !skip {
-			ln.updateSmap(sub, delta)
+		// the detection of loops as long as different cluster.
+		clusterDifferent := cluster != ln.remoteCluster()
+		if (isLDS && clusterDifferent) || (cluster == _EMPTY_ || clusterDifferent) && (delta <= 0 || ln.canSubscribe(subject)) {
```
kozlovic, line 2131: Did not review the whole PR, but the last `&& (delta <= 0 ...` is a bit confusing, meaning, the `if` statement looks like: `if A || B && C` and it would be nice to add parenthesis to make sure of the intent (is it `if A || (B && C)` or `if (A || B) && C`?

derekcollison: Its A || B && C. The negative of the original I changed just to inline and not have to evaluate everything.

neilalexander (approved):
LGTM

server/leafnode.go
derekcollison: I will make more clear for the reader.

server/accounts.go
```diff
@@ -928,6 +928,13 @@ func (a *Account) registerLeafNodeCluster(cluster string) {
 	a.leafClusters[cluster]++
 }
 
+// Check to see if we already have this cluster registered.
+func (a *Account) hasLeafNodeCluster(cluster string) bool {
+	a.mu.Lock()
```
kozlovic, line 933: Could this be read lock?

server/leafnode.go
```diff
@@ -2125,11 +2125,11 @@ func (acc *Account) updateLeafNodes(sub *subscription, delta int32) {
...
 		// Check to make sure this sub does not have an origin cluster that matches the leafnode.
 		ln.mu.Lock()
-		skip := (cluster != _EMPTY_ && cluster == ln.remoteCluster()) || (delta > 0 && !ln.canSubscribe(subject))
 		// If skipped, make sure that we still let go the "$LDS." subscription that allows
-		// the detection of a loop.
-		if isLDS || !skip {
-			ln.updateSmap(sub, delta)
+		// the detection of loops as long as different cluster.
+		clusterDifferent := cluster != ln.remoteCluster()
+		if (isLDS && clusterDifferent) || ((cluster == _EMPTY_ || clusterDifferent) && (delta <= 0 || ln.canSubscribe(subject))) {
```
kozlovic, line 2131: Yes, using parenthesis make it more clear.

kozlovic (changes requested):
Maybe rlock is enough?

server/accounts.go
derekcollison: Yes will update.

derekcollison: ok please check now and lmk..

kozlovic (approved):
LGTM

Issue #5047, opened 2024-02-07
Pass token via websockets cookie for auth callout


jonaskello:
### Proposed change

Make it possible to use a cookie to pass the token to the auth callout. 

Today if I try this config:

```
accounts {
  AUTH { users: [ { user: auth, password: auth } ] }
  SYS {}
  my-app {}
}

authorization {
  auth_callout {
    issuer: XXXXXXX
    account: AUTH
    users: [ auth ]
  }
}

system_account: SYS

# Client port of 4222 on all interfaces
port: 4222

# Limit the payload
max_payload: 20Mb

# HTTP monitoring port
monitor_port: 8222

# Websocket
websocket 
{
     port: 9222
     no_tls: true
     compression: true
     jwt_cookie: "my_jwt_cookie_name"
}
```

I get this error:

```
nats-server: trusted operators or trusted keys configuration is required for JWT authentication via cookie "my_jwt_cookie_name"
```

Today the javascript browser client connects like this and the `authToken` string is passed to the auth callout service:

```js
    const conn = await Nats.connect({ servers: servers, token: authToken });
```

What I would want is to skip the `token` parameter and have the value of the cookie named "my_jwt_cookie_name" passed as the token in this code:

```js
    const conn = await Nats.connect({ servers: servers }); // token is implicitly set from cookie value
```

### Use case

A browser based application that gets a cookie from a backend that contains a custom token that you want to validate in an auth callout service. Today we need to send the token value in the javascript client as parameter. We could still read it in the browser from a cookie but that means we cannot use a httpOnly cookie which will make it less secure.

### Contribution

The check seems to be done here:

https://github.com/nats-io/nats-server/blob/74a2fbda355633f441156d0a7ce3dd86cf3c5ebe/server/websocket.go#L1003

I will probably not be able to fix this myself.

aricart:
@jonaskello the server currently has the websocket configuration, that allows configuring the cookie as you have seen:
```
# Name for an HTTP cookie, that if present will be used as a client JWT.
    # If the client specifies a JWT in the CONNECT protocol, this option is ignored.
    # The cookie should be set by the HTTP server as described [here](https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies#restrict_access_to_cookies).
    # This setting is useful when generating NATS `Bearer` client JWTs as the
    # result of some authentication mechanism. The HTTP server after correct
    # authentication can issue a JWT for the user, that is set securely preventing
[...]

PR #5131, opened 2024-02-26
[FEAT] additional WS authentication cookie configurations

aricart:
[FEAT] this change adds `user_cookie`, `pass_cookie`, and `token_cookie` to the WS configuration block - these mirror the behavior of `jwt_cookie`. If the options are not specified by the client but the server is has them configured, a WS connection will lookup values for the corresponding options under the HTTP cookie. This simplies websocket applications by allowing the HTTP application to set cookies that provide authentication details.

Simplifies #5047 

Signed-off-by: Alberto Ricart <aricart@synadia.com>

### DIFF
--- a/server/client.go
+++ b/server/client.go
@@ -2025,10 +2025,26 @@ func (c *client) processConnect(arg []byte) error {
 		}
 	}
 
-	// If websocket client and JWT not in the CONNECT, use the cookie JWT (possibly empty).
-	if ws := c.ws; ws != nil && c.opts.JWT == "" {
-		c.opts.JWT = ws.cookieJwt
+	// if websocket client, maybe some options through cookies
+	if ws := c.ws; ws != nil {
+		// if JWT not in the CONNECT, use the cookie JWT (possibly empty).
+		if c.opts.JWT == _EMPTY_ {
+			c.opts.JWT = ws.cookieJwt
+		}
+		// if user not in the CONNECT, use the cookie user (possibly empty)
+		if c.opts.Username == _EMPTY_ {
+			c.opts.Username = ws.cookieUsername
+		}
+		// if pass not in the CONNECT, use the cookie password (possibly empty).
+		if c.opts.Password == _EMPTY_ {
+			c.opts.Password = ws.cookiePassword
+		}
+		// if token not in the CONNECT, use the cookie token (possibly empty).
+		if c.opts.Token == _EMPTY_ {
+			c.opts.Token = ws.cookieToken
+		}
 	}
+
 	// when not in operator mode, discard the jwt
 	if srv != nil && srv.trustedKeys == nil {
 		c.opts.JWT = _EMPTY_
--- a/server/opts.go
+++ b/server/opts.go
@@ -1,4 +1,4 @@
-// Copyright 2012-2023 The NATS Authors
+// Copyright 2012-2024 The NATS Authors
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
 // You may obtain a copy of the License at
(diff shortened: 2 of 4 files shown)


server/client.go
```diff
@@ -2029,6 +2029,22 @@ func (c *client) processConnect(arg []byte) error {
 	if ws := c.ws; ws != nil && c.opts.JWT == "" {
 		c.opts.JWT = ws.cookieJwt
 	}
+
+	// If websocket client and user not in the CONNECT, use the cookie user (possibly empty).
+	if ws := c.ws; ws != nil && c.opts.Username == "" {
```
derekcollison, line 2034: "" should be `_EMPTY_` in the server code base.

server/opts.go
```diff
@@ -434,6 +434,21 @@ type WebsocketOpts struct {
...
+
+	// Name of the cookie, which if present in WebSocket upgrade headers,
+	// will be treated as Password during CONNECT phase as long as
+	// "pass" specified in the CONNECT options is missing or empty.
+	PasswordCookie string
+
+	// Name of the cookie, which if present in WebSocket upgrade headers,
+	// will be treated as Token during CONNECT phase as long as
+	// "auth_token" specified in the CONNECT options is missing or empty.
+	TokenCookie string
```
derekcollison, line 450: Should we encourage this? I feel it should not be used and in my mind is deprecated. WDYT?

derekcollison (commented):
Some questions / comments. 

Also want @kozlovic to take a peek.

Looks good though and agree would be good addition.

server/opts.go
aricart: This is the perfect location for users to provide a non-nats JWT for callout to process.

aricart: (this is why I thought this was a good thing to do)

server/client.go
```diff
@@ -2029,6 +2029,22 @@ func (c *client) processConnect(arg []byte) error {
...
+		c.opts.Username = ws.cookieUsername
+	}
+
+	// If websocket client and pass not in the CONNECT, use the cookie password (possibly empty).
+	if ws := c.ws; ws != nil && c.opts.Password == "" {
+		c.opts.Password = ws.cookiePassword
+	}
+
+	// If websocket client and token not in the CONNECT, use the cookie token (possibly empty).
+	if ws := c.ws; ws != nil && c.opts.Token == "" {
```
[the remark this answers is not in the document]
derekcollison: Ah good point, you are correct.

server/opts.go
derekcollison: ok that make sense. So maybe we call that out in the comment, that this should not be a plain server wide token but  something external that folks want passed to the auth callout service etc.

kozlovic (changes requested):

server/websocket.go
```diff
@@ -812,6 +815,21 @@ func (s *Server) wsUpgrade(w http.ResponseWriter, r *http.Request) (*wsUpgradeRe
 				ws.cookieJwt = c.Value
 			}
 		}
+		if opts.Websocket.UsernameCookie != _EMPTY_ {
```
kozlovic, line 818: Since it looks like we are doing it more than once (above with JWTCookie), maybe we could use `r.Cookies` to avoid parsing the headers multiple times. Something like:
```
if cookies := r.Cookies(); len(cookies) > 0 {
	ows := &opts.Websocket
	for _, c := range cookies {
		if ows.JWTCookie == c.Name {
			ws.cookieJwt = c.Value
		} else if ows.UsernameCookie == c.Name {
			ws.cookieUsername = c.Value
		} else if ows.PasswordCookie == c.Name {
			ws.cookiePassword = c.Value
		} else if ows.TokenCookie == c.Name {
			ws.cookieToken = c.Value
		}
	}
}
```

kozlovic (changes requested):

server/opts.go
aricart: well, it can be anything they want, including a server-wide token even if that is discouraged.

aricart: (added a good comment line)

server/client.go
```diff
@@ -2026,9 +2026,25 @@ func (c *client) processConnect(arg []byte) error {
 	}
 
 	// If websocket client and JWT not in the CONNECT, use the cookie JWT (possibly empty).
-	if ws := c.ws; ws != nil && c.opts.JWT == "" {
+	if ws := c.ws; ws != nil && c.opts.JWT == _EMPTY_ {
```
kozlovic, line 2029: I am not sure if my earlier comment was properly posted. Trying again:

Why don't we consolidate everything under that first part of the if statement:
```
if ws := c.ws; ws != nil {
    if c.opts.JWT == _EMPTY_ {
        c.opts.JWT = ws.cookieJwt
    }
    if c.opts.Username == _EMPTY_ {
        c.opts.Username = ws.cookieUsername
    }
    ...
}
```
Basically, no need to check for `ws := c.ws; ws != nil` every time.

kozlovic (changes requested):

server/client.go
aricart: didn't complete it yet.. Now it is in

kozlovic (approved):
LGTM

server/opts.go
MamaMia777: What is proper way to access this cookie in auth_callout ?

aricart: > What is proper way to access this cookie in auth_callout ?

https://github.com/nats-io/nats-server/issues/5047#issuecomment-2607752031

aricart:
An important note is that when setting these cookies browser will be subject to CORS, which means that if the HTTP sets a cookie, it will need `Secure`, `SameSite`, (and possibly `HttpOnly`), which in turn requires https/wss connections.

PR #5132, opened 2024-02-26
[FIXED] Make sure to not remove mappings that just had their value changed.

derekcollison:


wallyqs (approved):
LGTM

PR #5135, opened 2024-02-27
Refactor `getHeader` to improve bounds checking

neilalexander:
There's now an additional bounds check when checking for CRLF and moved one or two other bits around.


derekcollison (approved):
LGTM

PR #5144, opened 2024-02-29
[FIXED] Slow ack for snapshots could cause slow consumer and client disconnect.

derekcollison:


wallyqs (approved):
LGTM

derekcollison:
Thanks @wallyqs, I want to make sure @ripienaar has a chance to double check the cli and the headers such that they do not affect anything. They shouldn't since we normally check len(m.Data) == 0 but want him to confirm.

ripienaar (approved):
LGTM, did some tests to add awareness of this to cli and nice to get the error.  But non breaking change for older CLI, so all good!

neilalexander (approved):
LGTM

PR #5145, opened 2024-02-29
[IMPROVED] Do not fire the store's expire timer too soon or too often since.

derekcollison:
Firing too often can effect ingest performance.

Resolves: #5136 


neilalexander (approved):
LGTM

Jarema (approved):
LGTM!

I wonder if it make sense to make this a bit more flexible and what is the sweet spot between burts of expires and too frequent expires.

derekcollison:
The value is actually tied into the resolution we will support in per msg ttls..

PR #5146, opened 2024-02-29
[FIXED] Data race and lock inversion on consumer delete.

derekcollison:
When we were checking if the consumer was assigned we were doing it with the consumer lock being held which is a lock inversion.

We also needed to lock the account when accessing the jsa to avoid a data race.


neilalexander (approved):
LGTM

PR #5149, opened 2024-02-29
NRG: Tweak vote response handling

neilalexander:
This should ensure that duplicate vote responses from the same node can't be handled and that vote response terms can't be mixed during the handling.

### DIFF
--- a/server/raft.go
+++ b/server/raft.go
@@ -2919,7 +2919,9 @@ func (n *raft) runAsCandidate() {
 	n.requestVote()
 
 	// We vote for ourselves.
-	votes := 1
+	votes := map[string]struct{}{
+		n.ID(): {},
+	}
 
 	for n.State() == Candidate {
 		elect := n.electTimer()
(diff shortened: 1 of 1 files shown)


server/raft.go
```diff
@@ -2947,11 +2949,11 @@ func (n *raft) runAsCandidate() {
 			nterm := n.term
 			n.RUnlock()
 
-			if vresp.granted && nterm >= vresp.term {
```
derekcollison, line 2950: Why not >=?

neilalexander: We should step down if we detect a higher term than the current candidate is campaigning for, so changing to `==` allows the `else` condition to kick in for that case regardless of granted status.

derekcollison: Right now with your change we do not handle that at all AFAIK.

neilalexander: I think that's right, we would handle the same term if it's a granted vote only, step down for a higher term and ignore a lower term. That way anything still in flight before a new campaign is properly ignored.

derekcollison: Each vote request has different inbox, so I do not believe there is a chance of other stuff reaching us from a prior campaign.

derekcollison: You are correct, it is the same. Hmm.. ok. We can revisit that in a later PR.

derekcollison (approved):
LGTM

PR #5151, opened 2024-02-29
NRG: Write peer state/term and vote files inline

neilalexander:
This should avoid some potential race conditions from scheduling persistence of the peer state and term and vote files asynchronously. It also gets rid of another goroutine per Raft group which is a nice side effect.


server/raft.go
```diff
@@ -3680,11 +3674,11 @@ func (n *raft) writePeerState(ps *peerState) {
 	if bytes.Equal(n.wps, pse) {
 		return
 	}
-	// Stamp latest and kick writer.
+	// Stamp latest and write the peer state file.
 	n.wps = pse
```
derekcollison, line 3678: If we write inline do we need to store []byte anymore?

server/raft.go
```diff
@@ -3719,6 +3713,23 @@ func readPeerState(sd string) (ps *peerState, err error) {
...
+func writeTermVote(sd string, wtv []byte) error {
+	psf := filepath.Join(sd, termVoteFile)
+	if _, err := os.Stat(psf); err != nil && !os.IsNotExist(err) {
+		return err
+	}
+
+	<-dios
+	err := os.WriteFile(psf, wtv, defaultFilePerms)
+	dios <- struct{}{}
+
```
derekcollison, line 3726: Just `return err`

server/raft.go
```diff
@@ -3680,11 +3674,11 @@ func (n *raft) writePeerState(ps *peerState) {
...
 		return
 	}
-	// Stamp latest and kick writer.
+	// Stamp latest and write the peer state file.
 	n.wps = pse
-	select {
-	case n.wpsch <- struct{}{}:
-	default:
+	if err := writePeerState(n.sd, ps); err != nil {
+		n.setWriteErr(err)
```
derekcollison, line 3680: Should we check that we are not closed like we do in the async version or is that handled above?

server/raft.go
```diff
@@ -3840,11 +3809,11 @@ func (n *raft) writeTermVote() {
 	if bytes.Equal(n.wtv, b) {
 		return
 	}
-	// Stamp latest and kick writer.
+	// Stamp latest and write the term & vote file.
 	n.wtv = b
```
derekcollison, line 3813: Same here, do not think we need to hold onto these anymore.

server/raft.go
```diff
@@ -3840,11 +3809,11 @@ func (n *raft) writeTermVote() {
...
-	select {
-	case n.wtvch <- struct{}{}:
-	default:
+	if err := writeTermVote(n.sd, n.wtv); err != nil {
+		n.setWriteErr(err)
```
derekcollison, line 3815: Same question on closed state.

server/raft.go
neilalexander: Only reason I can think of is so we don't rewrite it to disk if it hasn't changed, otherwise we could remove it.

derekcollison: I don't think we compare it anymore do we?

neilalexander: I think we do with the `bytes.Equal` checks above.

derekcollison: Yes you are correct. Apologies.

server/raft.go
neilalexander: Ah yes, have added it back in.

server/raft.go
neilalexander: Yep, done — was a copy-pasta job from what was there before.

derekcollison (approved):
LGTM

PR #5152, opened 2024-02-29
NRG: Wait for upper layer to update applied on inline CE updates

neilalexander:
In particular peer state entries don't produce committed entries, but bumping `applied` inline means that we can potentially move the Raft state ahead of the real applied state. Instead just send up a CE with no entries and wait for the upper layer to update `applied` instead, so that there is better consistency between the layers.


derekcollison (approved):
LGTM

PR #5158, opened 2024-03-01
Add support for AsOfTime for get/direct msg requests

bruth:
The motivation for this feature is to enable stream message gets as of some starting time. This is to match the same behavior with setting a start time when creating a consumer.

What this enables, combined with batches, is to read messages from a stream without the need to use a consumer. This is not intended for high-throughput use, but for querying and stream inspection.

The main driving need for this is the ability to inspect workqueue streams that cannot otherwise be done do due to the overlapping subject constraint.

### DIFF
--- a/server/jetstream_api.go
+++ b/server/jetstream_api.go
@@ -656,6 +656,8 @@ type JSApiMsgGetRequest struct {
 	UpToSeq uint64 `json:"up_to_seq,omitempty"`
 	// Only return messages up to this time.
 	UpToTime *time.Time `json:"up_to_time,omitempty"`
+	// Return messages as of this time.
+	AsOfTime *time.Time `json:"as_of_time,omitempty"`
 }
 
 type JSApiMsgGetResponse struct {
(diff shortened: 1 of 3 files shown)


server/jetstream_api.go
```diff
@@ -656,6 +656,8 @@ type JSApiMsgGetRequest struct {
 	UpToSeq uint64 `json:"up_to_seq,omitempty"`
 	// Only return messages up to this time.
 	UpToTime *time.Time `json:"up_to_time,omitempty"`
+	// Return messages as of this time.
```
derekcollison, line 659: We should move this to the batch ones v s multi. @ripienaar will redo some comments etc for go doc in a separate PR.

Also, mybe StartTime to match what we have on consumer config?

server/jetstream_api.go
```diff
@@ -656,8 +656,8 @@ type JSApiMsgGetRequest struct {
 	UpToSeq uint64 `json:"up_to_seq,omitempty"`
 	// Only return messages up to this time.
 	UpToTime *time.Time `json:"up_to_time,omitempty"`
-	// Return messages as of this time.
-	AsOfTime *time.Time `json:"as_of_time,omitempty"`
+	// Return messages as of this start time.
+	StartTime *time.Time `json:"start_time,omitempty"`
```
derekcollison, line 660: Move this up to where we have fields for batch.

derekcollison: This is the multi section.

bruth: Done.

derekcollison (approved):
LGTM

Issue #5167, opened 2024-03-03
panic: runtime error: index out of range [-1]


jeremylowery:
### Observed behavior

I encountered this going through the tutorial. When have two subscribers publishing a message causes nats server to crash with the following panic.
<pre>
$ nats-server
[28028] 2024/03/03 14:46:09.770143 [INF] Starting nats-server
[28028] 2024/03/03 14:46:09.770173 [INF]   Version:  2.10.11
[28028] 2024/03/03 14:46:09.770175 [INF]   Git:      [5fc6051]
[28028] 2024/03/03 14:46:09.770177 [INF]   Name:     NAEV2I4BY4R24YLQNE5VVCAK4IEUVBRWYENFTQPE3RZI6QF6JGQ5AT34
[28028] 2024/03/03 14:46:09.770178 [INF]   ID:       NAEV2I4BY4R24YLQNE5VVCAK4IEUVBRWYENFTQPE3RZI6QF6JGQ5AT34
[28028] 2024/03/03 14:46:09.770571 [INF] Listening for client connections on 0.0.0.0:4222
[28028] 2024/03/03 14:46:09.770739 [INF] Server is ready

panic: runtime error: index out of range [-1]

goroutine 82 [running]:
github.com/nats-io/nats-server/v2/server.(*client).processMsgResults(0xa6a2000, 0xa52c000, 0xa5aa240, {0xa6ac26b, 0x16, 0x195}, {0x0, 0x0, 0x0}, {0xa6ac233, ...}, ...)
        github.com/nats-io/nats-server/v2/server/client.go:4569 +0x2052
github.com/nats-io/nats-server/v2/server.(*client).processInboundClientMsg(0xa6a2000, {0xa6ac26b, 0x16, 0x195})
        github.com/nats-io/nats-server/v2/server/client.go:3906 +0xb23
github.com/nats-io/nats-server/v2/server.(*client).processInboundMsg(0xa6a2000, {0xa6ac26b, 0x16, 0x195})
        github.com/nats-io/nats-server/v2/server/client.go:3744 +0x74
github.com/nats-io/nats-server/v2/server.(*client).parse(0xa6a2000, {0xa6ac200, 0x81, 0x200})
        github.com/nats-io/nats-server/v2/server/parser.go:497 +0x1e4b
github.com/nats-io/nats-server/v2/server.(*client).readLoop(0xa6a2000, {0x0, 0x0, 0x0})
[...]

PR #5169, opened 2024-03-04
Fix select random qsub on 32-bit

wallyqs:
Take modulus before casting fastrand result to int as it can end up negative.

Related: https://github.com/nats-io/nats-server/issues/5167


neilalexander (approved):
LGTM

derekcollison (approved):
LGTM

PR #5173, opened 2024-03-05
Add logical name to accstatz endpoint

bruth:
In operator mode, the `acc` is the public nkey which can be difficult to correlate in monitoring dashboards. The `acc_name` as been added which includes the logical name associated with that account claims.

### DIFF
--- a/server/accounts.go
+++ b/server/accounts.go
@@ -52,6 +52,7 @@ type Account struct {
 	stats
 	gwReplyMapping
 	Name         string
+	LogicalName  string
 	Nkey         string
 	Issuer       string
 	claimJWT     string
@@ -251,9 +252,10 @@ type importMap struct {
 // NewAccount creates a new unlimited account with the given name.
 func NewAccount(name string) *Account {
 	a := &Account{
-		Name:     name,
-		limits:   limits{-1, -1, -1, -1, false},
-		eventIds: nuid.New(),
+		Name:        name,
+		LogicalName: name,
+		limits:      limits{-1, -1, -1, -1, false},
+		eventIds:    nuid.New(),
 	}
 	return a
 }
@@ -3752,6 +3754,8 @@ func (s *Server) updateAccountClaimsWithRefresh(a *Account, ac *jwt.AccountClaim
 func (s *Server) buildInternalAccount(ac *jwt.AccountClaims) *Account {
 	acc := NewAccount(ac.Subject)
 	acc.Issuer = ac.Issuer
+	// Override subject with logical name.
+	acc.LogicalName = ac.Name
 	// Set this here since we are placing in s.tmpAccounts below and may be
 	// referenced by an route RS+, etc.
 	s.setAccountSublist(acc)
--- a/server/events.go
+++ b/server/events.go
@@ -206,7 +206,7 @@ type AccountNumConns struct {
 // AccountStat contains the data common between AccountNumConns and AccountStatz
 type AccountStat struct {
 	Account       string    `json:"acc"`
-	AccountName   string    `json:"acc_name"`
+	Name          string    `json:"name"`
 	Conns         int       `json:"conns"`
 	LeafNodes     int       `json:"leafnodes"`
 	TotalConns    int       `json:"total_conns"`
@@ -2105,12 +2105,12 @@ func (a *Account) statz() *AccountStat {
 	localConns := a.numLocalConnections()
 	leafConns := a.numLocalLeafNodes()
 	return &AccountStat{
-		Account:     a.Name,
-		AccountName: a.LogicalName,
-		Conns:       localConns,
-		LeafNodes:   leafConns,
-		TotalConns:  localConns + leafConns,
-		NumSubs:     a.sl.Count(),
+		Account:    a.Name,
+		Name:       a.LogicalName,
+		Conns:      localConns,
+		LeafNodes:  leafConns,
+		TotalConns: localConns + leafConns,
+		NumSubs:    a.sl.Count(),
 		Received: DataStats{
 			Msgs:  atomic.LoadInt64(&a.inMsgs),
 			Bytes: atomic.LoadInt64(&a.inBytes),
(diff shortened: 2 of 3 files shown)


wallyqs (approved):
LGTM

server/events.go
```diff
@@ -206,6 +206,7 @@ type AccountNumConns struct {
 // AccountStat contains the data common between AccountNumConns and AccountStatz
 type AccountStat struct {
 	Account       string    `json:"acc"`
+	AccountName   string    `json:"acc_name"`
```
derekcollison, line 209: Should this be just name? Should it be omit empty?

server/accounts.go
```diff
@@ -251,9 +252,10 @@ type importMap struct {
 // NewAccount creates a new unlimited account with the given name.
 func NewAccount(name string) *Account {
 	a := &Account{
-		Name:     name,
-		limits:   limits{-1, -1, -1, -1, false},
-		eventIds: nuid.New(),
+		Name:        name,
+		LogicalName: name,
```
derekcollison, line 256: Should this not be set here?

server/accounts.go
```diff
@@ -3752,6 +3754,8 @@ func (s *Server) updateAccountClaimsWithRefresh(a *Account, ac *jwt.AccountClaim
 func (s *Server) buildInternalAccount(ac *jwt.AccountClaims) *Account {
 	acc := NewAccount(ac.Subject)
 	acc.Issuer = ac.Issuer
+	// Override subject with logical name.
```
derekcollison, line 3757: Should this be conditional?

bruth:
The reason for always setting it for consistency of the output and use within the prometheus exporter which expects a consistent set of labels for emitting metrics. See [here](https://github.com/nats-io/prometheus-nats-exporter/blob/main/collector/accstatz.go#L156) how it includes the server ID, and account (ID). We would add the account name there as well (unless @wallyqs has a different idea).

The output for non-operator mode looks like this:
```json
{
  "server_id": "NBYAQHBRMYSPO6JOZXYXWZT7KGBD2LYML7VZDPUIIHQRNBOIN5SIK6LK",
  "now": "2024-03-06T11:00:13.201452Z",
  "account_statz": [
    {
      "acc": "$G",
      "acc_name": "$G",
      "conns": 2,
      "leafnodes": 0,
      "total_conns": 2,
      "num_subscriptions": 5,
      "sent": {
        "msgs": 24,
        "bytes": 3072
      },
      "received": {
        "msgs": 36,
        "bytes": 3168
      },
      "slow_consumers": 0
    }
  ]
}
```
But operator mode would look like this:
```json
{
  "server_id": "NDXNEXLCF7V6V2IZ5IPW5DT5HERJDJNJOA55OSUUAY53FUW4ZJCFA7S3",
  "now": "2024-03-06T11:11:00.506395Z",
  "account_statz": [
    {
      "acc": "ADY46PUVYEC4QPHXTFLQRAMOXVY3TOGEQCH2K2CQFGZUAOTJDEQIPY7V",
      "acc_name": "APP",
      "conns": 2,
      "leafnodes": 0,
      "total_conns": 2,
      "num_subscriptions": 5,
      "sent": {
        "msgs": 72,
        "bytes": 9216
      },
      "received": {
        "msgs": 72,
        "bytes": 9216
      },
      "slow_consumers": 0
    }
  ]
}
```

server/events.go
bruth: Changed to "name". I checked that within the other embedded fields, there isn't a conflict and it should be contextualized enough to not be ambiguous.

derekcollison: Still seems to be AccountName

server/monitor_test.go
```diff
@@ -4050,7 +4050,9 @@ func TestMonitorAccountz(t *testing.T) {
 
 	body = string(readBody(t, fmt.Sprintf("http://127.0.0.1:%d%s?unused=1", s.MonitorAddr().Port, AccountStatzPath)))
 	require_Contains(t, body, `"acc": "$G"`)
+	require_Contains(t, body, `"acc_name": "$G"`)
```
derekcollison, line 4053: Be good to have a test with JWTs that show they are different.

bruth: Will add.

server/monitor_test.go
```diff
@@ -4050,7 +4050,78 @@ func TestMonitorAccountz(t *testing.T) {
...
+	body = string(readBody(t, fmt.Sprintf("http://127.0.0.1:%d%s?acc=%s", s.MonitorAddr().Port, AccountzPath, sysPub)))
+	require_Contains(t, body, `"account_detail": {`)
+	require_Contains(t, body, fmt.Sprintf(`"account_name": "%s",`, sysPub))
+	require_Contains(t, body, `"subscriptions": 50,`)
+	require_Contains(t, body, `"is_system": true,`)
+	require_Contains(t, body, fmt.Sprintf(`"system_account": "%s"`, sysPub))
+
+	// TODO: understand why the APP account did not show up in the accountz detail
+	// even though unused is set. It required a connection to be made to show up.
+	body = string(readBody(t, fmt.Sprintf("http://127.0.0.1:%d%s?unused=1", s.MonitorAddr().Port, AccountStatzPath)))
```
bruth, line 4120: Just noting this comment. When `unused` is set, it loops over all loaded account on the server, so the preloaded `APP` account to be present, but for some reason it does not show up in the list unless there is an active connection.

bruth:
Confused why only the MQTT tests ran and not the server tests.. Travis offline?

derekcollison (approved):
LGTM

server/events.go
```diff
@@ -2105,6 +2106,7 @@ func (a *Account) statz() *AccountStat {
 	leafConns := a.numLocalLeafNodes()
 	return &AccountStat{
 		Account:    a.Name,
+		Name:       a.LogicalName,
```
ripienaar, line 2109: Any objection to adding this to a few other places? connz mainly

bruth: Yes, will add to the rest of the endpoints.

PR #5179, opened 2024-03-06
Fix correcting `clseq` drift on high `clfs`

neilalexander:
In a scenario where the `clfs` is high (i.e. with competing KV writes or a high number of duplicate messages) then the gap on the `clseq` fix-up condition was growing, probably to the point that it wasn't running when it should. This would cause a drift.


derekcollison (changes requested):
Can we have a test that shows the issue?

derekcollison (approved):
LGTM

PR #5183, opened 2024-03-07
Lock when bumping clfs. Make sure to unlock mset lock before bumpCLFS.

derekcollison:
Locking order is clMu then mset lock.


wallyqs (approved):
LGTM

PR #5192, opened 2024-03-08
Fix sending empty reply on gateway `RMSG`

### LINKED ISSUE #5187: Using queue-groups when subscribing to $SYS.ACCOUNT.> across routes
### Observed behavior

I am currently using two superclusters of nats, one in test and one in production.

When subscribing to $SYS.ACCOUNT.> without any queue i can receive account CONNECT/DISCONNECT events from all clusters across routes. 

When subscribing with a queue group i can only receive events from the cluster i am currently connected to. 

Very simillar to https://github.com/nats-io/nats-server/issues/3177

### Expected behavior

Client within a queue group should be able to receive account events across a super cluster.

### Server and client version

nats-server: v2.10.11 (Both test & production)

~ # nats --version
v0.1.1

### Host environment

Official docker image in kubernetes using official helm charts. 

### Steps to reproduce

``nats -s <redacted>--creds=<SYS_PATH>/sys.creds sub '$SYS.ACCOUNT.*.*'``

``nats -s <redacted>--creds=<SYS_PATH>/sys.creds sub '$SYS.ACCOUNT.*.*' --queue=foo``

This behaviour has been verfied both in test and production.

neilalexander:
This should fix #5187. The `nil` check was insufficient because the reply could still be non-nil but empty.

This meant that we sent `RMSG $SYS $SYS.ACCOUNT.FOO.CONNECT +  myqueue 502` instead of `RMSG $SYS $SYS.ACCOUNT.FOO.CONNECT | myqueue 502`.

### DIFF
--- a/server/gateway.go
+++ b/server/gateway.go
@@ -2575,7 +2575,7 @@ func (c *client) sendMsgToGateways(acc *Account, msg, subject, reply []byte, qgr
 		mh = append(mh, subject...)
 		mh = append(mh, ' ')
 		if len(queues) > 0 {
-			if reply != nil {
+			if len(reply) > 0 {
 				mh = append(mh, "+ "...) // Signal that there is a reply.
 				mh = append(mh, mreply...)
 				mh = append(mh, ' ')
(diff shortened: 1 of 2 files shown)


wallyqs (approved):
LGTM

derekcollison (approved):
LGTM

PR #5195, opened 2024-03-09
test: Fix TestJetStreamBasicNilConfig when GOMEMLIMIT is set

wallyqs:
When running tests using GOMEMLIMIT this one would fail.


neilalexander (approved):
LGTM

PR #5196, opened 2024-03-11
[FIXED] Stream interest leaving messages

derekcollison:
Several fixes to consumers and interest based streams.

1. Make sure to not commit a replicated ack if we bail due to a closed consumer.
2. Check the stream state once recovery is complete. This checks NRG, we already were checking on upper state catchups.


neilalexander (approved):
LGTM

PR #5200, opened 2024-03-11
Revert #4990

neilalexander:
This reverts #4990 for now.

From a Raft correctness perspective I think removing the step-down channel is the right thing to do, but it has a negative impact on the amount of time taken to complete stream moves currently (seemingly they wait for an election timeout) so I want to better understand what's going on there and to make sure that the cooperative leader handover is doing the right thing before we reintroduce those changes.

`TestJetStreamSuperClusterMovingStreamAndMoveBack/R3` shows the issue for future reference.


derekcollison (approved):
LGTM

PR #5201, opened 2024-03-11
Add `NATS_LOGGING` test helper

neilalexander:
Setting this environment variable forces NATS Server instances running in the unit tests to log to stdout. Possible values are:

* `NATS_LOGGING=true go test ...` (or any other non-empty value)
* `NATS_LOGGING=debug go test ...`
* `NATS_LOGGING=trace go test ...`

### DIFF
--- a/server/server_test.go
+++ b/server/server_test.go
@@ -35,6 +35,7 @@ import (
 	"testing"
 	"time"
 
+	srvlog "github.com/nats-io/nats-server/v2/logger"
 	"github.com/nats-io/nats.go"
 )
 
@@ -84,6 +85,13 @@ func RunServer(opts *Options) *Server {
 
 	if !opts.NoLog {
 		s.ConfigureLogger()
+	}
+
+	if ll := os.Getenv("NATS_LOGGING"); ll != "" {
+		log := srvlog.NewTestLogger(fmt.Sprintf("[%s] | ", s), true)
+		debug := ll == "debug" || ll == "trace"
+		trace := ll == "trace"
+		s.SetLoggerV2(log, debug, trace, false)
 	}
 
 	// Run server in Go routine.
--- a/test/test.go
+++ b/test/test.go
@@ -28,6 +28,7 @@ import (
 	"testing"
 	"time"
 
+	srvlog "github.com/nats-io/nats-server/v2/logger"
 	"github.com/nats-io/nats-server/v2/server"
 )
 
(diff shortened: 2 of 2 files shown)


server/server_test.go
```diff
@@ -86,6 +87,13 @@ func RunServer(opts *Options) *Server {
 		s.ConfigureLogger()
 	}
 
+	if ll := os.Getenv("NATS_LOGGING"); ll != "" {
+		log := srvlog.NewTestLogger(fmt.Sprintf("[%-5s] | ", opts.ServerName), true)
```
kozlovic, line 91: Not sure of the need of the `%-5s`. It looks weird for small names like `A` or `B`, but I get that if you have small names (but less than 5 characters), at least the output is "aligned".
```
[B    ] | 2024/03/11 09:30:14.319038 [INF] Starting nats-server
```
I would have gone with a simple `fmt.Sprintf("[%s] ",`.

But I would recommend replacing `opts.ServerName` with `s` otherwise, when the server name is not set in the options, you would get something like:
```
=== RUN   TestGatewayBasic
[     ] | 2024/03/11 09:26:47.580013 [INF] Starting nats-server
[     ] | 2024/03/11 09:26:47.580029 [INF]   Version:  2.11.0-dev
[     ] | 2024/03/11 09:26:47.580037 [INF]   Git:      [not set]
...
```
When the name is not set, `s` (which is Server.String()) would return the server ID, which will at least allow differentiating between servers:
```
=== RUN   TestGatewayBasic
[NBZZHUIG3KYM6D434GN3PDVSKDM3UC6NGJF7DG4S4NKVN6AGSD2J3HHI] | 2024/03/11 09:29:13.391026 [INF] Starting nats-server
[NBZZHUIG3KYM6D434GN3PDVSKDM3UC6NGJF7DG4S4NKVN6AGSD2J3HHI] | 2024/03/11 09:29:13.391041 [INF]   Version:  2.11.0-dev
[NBZZHUIG3KYM6D434GN3PDVSKDM3UC6NGJF7DG4S4NKVN6AGSD2J3HHI] | 2024/03/11 09:29:13.391052 [INF]   Git:      [not set]
```

kozlovic (changes requested):

server/server_test.go
neilalexander: Have made these changes!

wallyqs (approved):
LGTM

derekcollison:
See stree for what I mean.

kozlovic (approved):
LGTM

neilalexander:
I did consider arguments but thought that using env vars for a global switch made more sense and that way we also keep flags clean for test-specific options.

PR #5207, opened 2024-03-12
Add reload lock to ensure new connections await completion of config reload

neilalexander:
This may help with some problems during configuration reload with e.g. OCSP stapling.

### DIFF
--- a/server/reload.go
+++ b/server/reload.go
@@ -999,6 +999,9 @@ func (s *Server) Reload() error {
 // type. This returns an error if an option which doesn't support
 // hot-swapping was changed.
 func (s *Server) ReloadOptions(newOpts *Options) error {
+	s.reloadMu.Lock()
+	defer s.reloadMu.Unlock()
+
 	s.mu.Lock()
 
 	curOpts := s.getOpts()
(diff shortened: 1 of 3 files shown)


server/reload.go
```diff
@@ -999,6 +999,9 @@ func (s *Server) Reload() error {
 // type. This returns an error if an option which doesn't support
 // hot-swapping was changed.
 func (s *Server) ReloadOptions(newOpts *Options) error {
+	s.reloadMu.Lock()
```
derekcollison, line 1002: We should add a note to locking order.

server/server.go
```diff
@@ -2727,7 +2728,9 @@ func (s *Server) acceptConnections(l net.Listener, acceptName string, createFunc
 		}
 		tmpDelay = ACCEPT_MIN_SLEEP
 		if !s.startGoRoutine(func() {
+			s.reloadMu.RLock()
```
derekcollison, line 2731: This for all connections?

neilalexander: Yes, as we can have OCSP configured on any connection type. We don't know that the problem was unique to gateways.

server/reload.go
neilalexander: Done!

derekcollison (approved):
LGTM

PR #5208, opened 2024-03-12
ocsp: Fixes for gateway reconnecting and reloading

wallyqs:
Fixes some issues with gateways reconnections that had OCSP stapling enabled:
- Gateways that started without explicit remotes would not always have OCSP callbacks registered so had no staples ready for the remote connections, when this condition is detected they are now applied on the fly.
- Access to GetClientCertificate could be concurrent under some conditions and could also race with OCSP monitor updates.
- Reloading races with gateway reconnections also addressed.

Tests also cover conditions from #5207, this also moves the OCSP testing helpers to an internal package.

### DIFF
--- a/internal/ocsp/ocsp.go
+++ b/internal/ocsp/ocsp.go
@@ -1,0 +1,274 @@
+// Copyright 2019-2021 The NATS Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package testhelper
+
+import (
+	"crypto"
+	"crypto/tls"
+	"crypto/x509"
+	"encoding/base64"
+	"encoding/pem"
+	"fmt"
+	"io"
+	"net/http"
+	"os"
+	"strconv"
+	"strings"
+	"sync"
+	"testing"
+	"time"
+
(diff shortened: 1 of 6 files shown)


internal/ocsp/ocsp.go
```diff
@@ -0,0 +1,274 @@
+// Copyright 2019-2021 The NATS Authors
```
derekcollison, line 1: Should this be 2024?

wallyqs:
Ok this one is ready to for review again, test running time is currently 15s.

internal/ocsp/ocsp.go
```diff
@@ -0,0 +1,274 @@
...
+				http.Error(rw, "", http.StatusBadRequest)
+				return
+			}
+			reqData, err = base64.StdEncoding.DecodeString(r.URL.Path[1:])
+		case r.Method == "POST":
+			if method != "" && r.Method != method {
+				http.Error(rw, "", http.StatusBadRequest)
+				return
+			}
+			reqData, err = io.ReadAll(r.Body)
```
derekcollison, line 133: Should we also be calling `r.Body.Close()`?

server/gateway_test.go
```diff
@@ -6950,3 +6953,403 @@ func TestGatewayConnectEvents(t *testing.T) {
...
+	if err := srvA.Reload(); err != nil {
+		t.Fatal(err)
+	}
+	waitForOutboundGateways(t, srvA, 2, 5*time.Second)
+	waitForOutboundGateways(t, srvB, 2, 5*time.Second)
+	waitForOutboundGateways(t, srvC, 2, 5*time.Second)
+
+	// Now clients connect to C can communicate with B and A.
+	_, err = cC.Request("foo", nil, 2*time.Second)
+	if err != nil {
```
derekcollison, line 7233: Require_NoError(t, err)?

internal/ocsp/ocsp.go
wallyqs: add `r.Body.Close()` here

derekcollison (approved):
LGTM

PR #5227, opened 2024-03-19
[FIXED] Improved edge situations that could leave orphan msgs in workqueues.

derekcollison:
1. Big fix on check ack floor was off by one when setting via pending.
2. Remove from redilver queue <= stream floor not just <.
3. Update skipped only when increasing o.sseq, saves lots on network.


Jarema (approved):
LGTM

PR #5228, opened 2024-03-19
[IMPROVED] Use match versus iter since more performant and less stressful on gc.

derekcollison:
We do not care about order so using Match() vs Iter() will be much more friendly to the GC. Especially for streams with lots of subjects, e.g. > 1M.


wallyqs (approved):
LGTM

Issue #5226, opened 2024-03-19
HSTS Missing From HTTPS for WebSockets


mcosta74:
### Observed behavior

One customer ran a security scanner and we got this report

```
HSTS Missing From HTTPS Medium 6.5 Configure the remote web server to use HSTS. Server (RFC 6797)
1 Affected Host(s): xxxxxxx
Initial Detection: 2024-03-11 21:08 UTC
Latest Detection: 2024-03-16 00:03 UTC
Description: The remote web server is not enforcing HSTS, as defined by RFC 6797. HSTS is an optional response header that can be configured on the server to instruct the browser to only communicate via HTTPS. The lack of HSTS allows downgrade attacks, SSL-stripping man-in-the-middle attacks, and weakens cookie-hijacking protections.
```

Is there a way to add this missing header ?

### Expected behavior

N/A

### Server and client version

`nats-server`: `v2.10.2`

bruth:
Here is the [set of examples](https://datatracker.ietf.org/doc/html/rfc6797#section-6.2) and implications. In terms of implementation, setting the header if WebSocket TLS is enabled seems obvious. The question is whether additional server options are required to tune the age and subdomain option.

PR #5230, opened 2024-03-19
[ADD] WebSocket custom response headers

ramonberrutti:
Allow to configure extra headers for the WebSocket Upgrade Response:

For Example:
```
websocket {
    port: 8080
    no_tls: true
    headers {
        Strict-Transport-Security: "max-age=15768000 ; includeSubDomains"
        X-Frame-Options: "DENY"
    }
}
```

Related Issue: https://github.com/nats-io/nats-server/issues/5226

### DIFF
--- a/server/opts.go
+++ b/server/opts.go
@@ -498,6 +498,10 @@ type WebsocketOpts struct {
 	// time needed for the TLS Handshake.
 	HandshakeTimeout time.Duration
 
+	// Headers to be added to the upgrade response.
+	// Useful for adding custom headers like Strict-Transport-Security.
+	Headers map[string]string
+
 	// Snapshot of configured TLS options.
 	tlsConfigOpts *TLSConfigOpts
 }
@@ -4810,6 +4814,23 @@ func parseWebsocket(v interface{}, o *Options, errors *[]error, warnings *[]erro
 			o.Websocket.TokenCookie = mv.(string)
 		case "no_auth_user":
 			o.Websocket.NoAuthUser = mv.(string)
+		case "headers":
+			m, ok := mv.(map[string]interface{})
+			if !ok {
+				err := &configErr{tk, fmt.Sprintf("error parsing headers: unsupported type %T", mv)}
+				*errors = append(*errors, err)
+				continue
+			}
+			o.Websocket.Headers = make(map[string]string)
+			for key, val := range m {
+				tk, val = unwrapValue(val, &lt)
+				if headerValue, ok := val.(string); !ok {
+					*errors = append(*errors, &configErr{tk, fmt.Sprintf("error parsing header key %s: unsupported type %T", key, val)})
+					continue
+				} else {
+					o.Websocket.Headers[key] = headerValue
+				}
+			}
 		default:
 			if !tk.IsUsedVariable() {
 				err := &unknownConfigFieldErr{
--- a/server/websocket.go
+++ b/server/websocket.go
@@ -133,7 +133,8 @@ type srvWebsocket struct {
 	sameOrigin     bool
 	connectURLs    []string
 	connectURLsMap refCountedUrlSet
-	authOverride   bool // indicate if there is auth override in websocket config
+	authOverride   bool   // indicate if there is auth override in websocket config
+	rawHeaders     string // raw headers to be used in the upgrade response.
 }
 
 type allowedOrigin struct {
@@ -777,6 +778,9 @@ func (s *Server) wsUpgrade(w http.ResponseWriter, r *http.Request) (*wsUpgradeRe
 	if kind == MQTT {
 		p = append(p, wsMQTTSecProto...)
 	}
+	if s.websocket.rawHeaders != _EMPTY_ {
+		p = append(p, s.websocket.rawHeaders...)
+	}
 	p = append(p, _CRLF_...)
 
 	if _, err = conn.Write(p); err != nil {
@@ -1019,6 +1023,24 @@ func validateWebsocketOptions(o *Options) error {
 	if err := validatePinnedCerts(wo.TLSPinnedCerts); err != nil {
 		return fmt.Errorf("websocket: %v", err)
 	}
+
+	// Check for invalid headers here.
+	for key := range wo.Headers {
+		k := strings.ToLower(key)
+		switch k {
+		case "host",
+			"content-length",
+			"connection",
+			"upgrade",
+			"nats-no-masking":
+			return fmt.Errorf("websocket: invalid header %q not allowed", key)
+		}
+
+		if strings.HasPrefix(k, "sec-websocket-") {
+			return fmt.Errorf("websocket: invalid header %q, \"Sec-WebSocket-\" prefix no allowed", key)
+		}
+	}
+
 	return nil
 }
 
@@ -1050,6 +1072,21 @@ func (s *Server) wsSetOriginOptions(o *WebsocketOpts) {
 	}
 }
 
+// Calculate the raw headers for websocket upgrade response.
+func (s *Server) wsSetHeadersOptions(o *WebsocketOpts) {
+	var sb strings.Builder
+	for k, v := range o.Headers {
+		sb.WriteString(k)
+		sb.WriteString(": ")
+		sb.WriteString(v)
+		sb.WriteString(_CRLF_)
+	}
+	ws := &s.websocket
+	ws.mu.Lock()
+	defer ws.mu.Unlock()
+	ws.rawHeaders = sb.String()
+}
+
 // Given the websocket options, we check if any auth configuration
 // has been provided. If so, possibly create users/nkey users and
 // store them in s.websocket.users/nkeys.
@@ -1071,6 +1108,7 @@ func (s *Server) startWebsocketServer() {
 	o := &sopts.Websocket
 
 	s.wsSetOriginOptions(o)
+	s.wsSetHeadersOptions(o)
 
 	var hl net.Listener
 	var proto string
--- a/server/websocket.go
+++ b/server/websocket.go
@@ -1037,7 +1037,7 @@ func validateWebsocketOptions(o *Options) error {
 		}
 
 		if strings.HasPrefix(k, "sec-websocket-") {
-			return fmt.Errorf("websocket: invalid header %q, \"Sec-WebSocket-\" prefix no allowed", key)
+			return fmt.Errorf("websocket: invalid header %q, \"Sec-WebSocket-\" prefix not allowed", key)
 		}
 	}
 
--- a/server/websocket_test.go
+++ b/server/websocket_test.go
@@ -1442,6 +1442,61 @@ func TestWSCompressNegotiation(t *testing.T) {
 	}
 }
 
+func TestWSSetHeader(t *testing.T) {
+	opts := testWSOptions()
+	opts.Websocket.Headers = map[string]string{
+		"X-Header":         "some-value",
+		"X-Another-Header": "another-value",
+	}
+	s := &Server{opts: opts}
+	s.wsSetHeadersOptions(&opts.Websocket)
+	rw := &testResponseWriter{}
+	req := testWSCreateValidReq()
+	res, err := s.wsUpgrade(rw, req)
+	if res == nil || err != nil {
+		t.Fatalf("Unexpected error: %v", err)
+	}
+
+	buf := bufio.NewReader(&rw.conn.wbuf)
+	resp, err := http.ReadResponse(buf, req)
+	if err != nil {
+		t.Fatalf("Error reading request: %v", err)
+	}
+	defer resp.Body.Close()
+
+	// Check that the response is a 101
+	if resp.StatusCode != http.StatusSwitchingProtocols {
+		t.Fatalf("Expected 101, got %v", resp.StatusCode)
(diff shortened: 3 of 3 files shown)


server/websocket.go
```diff
@@ -777,6 +777,9 @@ func (s *Server) wsUpgrade(w http.ResponseWriter, r *http.Request) (*wsUpgradeRe
 	if kind == MQTT {
 		p = append(p, wsMQTTSecProto...)
 	}
+	if opts.Websocket.Headers != "" {
```
ramonberrutti, line 780:
```suggestion
	if opts.Websocket.Headers != _EMPTY_ {
```

server/opts.go
```diff
@@ -4810,6 +4814,42 @@ func parseWebsocket(v interface{}, o *Options, errors *[]error, warnings *[]erro
...
+			tk, mv = unwrapValue(mv, &lt)
+			mvv, ok := mv.(map[string]interface{})
+			if !ok {
+				err := &configErr{tk, fmt.Sprintf("error parsing headers: unsupported type %T", mv)}
+				*errors = append(*errors, err)
+			}
+			m := make(map[string]string)
+			for kk, kv := range mvv {
+				_, v = unwrapValue(kv, &tk)
+				m[kk] = v.(string)
```
kozlovic, line 4827: Would make sure that we have key/value string pairs here. I also added a missing `continue` when processing handshake_timeout (harmless since we initialize the handshake to 0, so the missing continue has no effect, but just for correctness). Here is a diff of opts.go:
```
diff --git a/server/opts.go b/server/opts.go
index bfae98f5..0ba2a360 100644
--- a/server/opts.go
+++ b/server/opts.go
@@ -4794,6 +4794,7 @@ func parseWebsocket(v interface{}, o *Options, errors *[]error, warnings *[]erro
                        default:
                                err := &configErr{tk, fmt.Sprintf("error parsing handshake timeout: unsupported type %T", mv)}
+                               continue
                        }
                        o.Websocket.HandshakeTimeout = ht
                case "compress", "compression":
@@ -4820,11 +4821,21 @@ func parseWebsocket(v interface{}, o *Options, errors *[]error, warnings *[]erro
                        if !ok {
                                err := &configErr{tk, fmt.Sprintf("error parsing headers: unsupported type %T", mv)}
+                               continue
                        }
                        m := make(map[string]string)
                        for kk, kv := range mvv {
                                _, v = unwrapValue(kv, &tk)
-                               m[kk] = v.(string)
+                               if vv, ok := v.(string); ok {
+                                       m[kk] = vv
+                               } else {
+                                       err := &configErr{tk, fmt.Sprintf("headers content should be key/value strings, got %T", v)}
+                                       *errors = append(*errors, err)
+                                       break
+                       }
+                       if len(*errors) > 0 {
+                               continue
                        }
 
                        var sb strings.Builder
```

server/websocket_test.go
```diff
@@ -1609,6 +1627,42 @@ func TestWSParseOptions(t *testing.T) {
 				}
 				return nil
 			}, ""},
+		{"headers",
```
kozlovic, line 1630: I would add negative tests in the list above and for this one, you have to keep in mind that the fact that we iterate through a map (key/value), the order is not guaranteed. So if you were to run the test multiple times (`go test -race -v -run=TestWSParseOptions ./server -count=100 -vet=off -failfast`), you would get a failure for this test because the resulting string would sometimes be with the `X-Frame-Options: "DENY"` first.

Here is a diff of this file with some corrections:
```
diff --git a/server/websocket_test.go b/server/websocket_test.go
index 0af2d652..6bccf8b7 100644
--- a/server/websocket_test.go
+++ b/server/websocket_test.go
@@ -1480,6 +1480,8 @@ func TestWSParseOptions(t *testing.T) {
                {"bad handshake timeout type", `websocket: { handshake_timeout: [] }`, nil, "unsupported type"},
                {"bad handshake timeout duration", `websocket: { handshake_timeout: "abc" }`, nil, "invalid duration"},
                {"unknown field", `websocket: { this_does_not_exist: 123 }`, nil, "unknown"},
+               {"headers wrong type", `websocket: { headers: 123 }`, nil, "unsupported"},
+               {"headers content wrong type", `websocket: { headers: { 123: 456 } }`, nil, "headers content should be key/value strings"},
                // Positive tests
                {"listen port only", `websocket { listen: 1234 }`, func(wo *WebsocketOpts) error {
                        if wo.Port != 1234 {
@@ -1636,7 +1638,8 @@ func TestWSParseOptions(t *testing.T) {
                                }
                        }
                        `, func(wo *WebsocketOpts) error {
-                               if wo.Headers != "Strict-Transport-Security: max-age=15768000 ; includeSubDomains\r\nX-Frame-Options: DENY\r\n" {
+                               if !strings.Contains(wo.Headers, "Strict-Transport-Security: max-age=15768000 ; includeSubDomains\r\n") &&
+                                       !strings.Contains(wo.Headers, "X-Frame-Options: DENY\r\n") {
                                        return fmt.Errorf("Invalid headers: %v", wo.Headers)
                                }
                                return nil
```

kozlovic (changes requested):

ramonberrutti:
@kozlovic I reimplement to support equaly between the config and the server options. As we don't support reloads, I precalculate the headers before starting the WebSocket server. Let me know what you think. In a bit I also include the missing tests.

server/websocket_test.go
```diff
@@ -1442,6 +1442,27 @@ func TestWSCompressNegotiation(t *testing.T) {
...
+	s := &Server{opts: opts}
+	rw := &testResponseWriter{}
+	req := testWSCreateValidReq()
+	res, err := s.wsUpgrade(rw, req)
+	if res == nil || err != nil {
+		t.Fatalf("Unexpected error: %v", err)
+	}
+
+	output := rw.conn.wbuf.String()
+	_ = output
```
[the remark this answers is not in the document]
ramonberrutti: reimplemented some logic to have simple and better test

server/websocket_test.go
```diff
@@ -1442,6 +1442,60 @@ func TestWSCompressNegotiation(t *testing.T) {
...
+		"Connection":           "Upgrade",
+		"Sec-Websocket-Accept": wsAcceptKey(req.Header.Get("Sec-Websocket-Key")),
+	} {
+		if got := headers.Get(k); got != v {
+			t.Fatalf("Expected %q for header %q, got %q", v, k, got)
+		}
+		headers.Del(k)
+	}
+
+	for range headers {
```
kozlovic, line 1494: Since it is a `t.Fatalf()` it would stop at the first key/value found, but a `if len(headers)>0{` would be better I think.

server/websocket_test.go
```diff
@@ -1661,6 +1741,21 @@ func TestWSValidateOptions(t *testing.T) {
 			o.Websocket.Token = "mytoken"
 			return o
 		}, "websocket authentication token not compatible with presence of users/nkeys"},
+		{"headers with sec-websocket- not allowed", func() *Options {
+			o := wso.Clone()
+			o.Websocket.Headers = map[string]string{"Sec-WebSocket-Key": "123"}
+			return o
+		}, `invalid header "Sec-WebSocket-Key", "Sec-WebSocket-" prefix no allowed`},
+		{"header with ", func() *Options {
```
kozlovic, line 1749: something missing here? Maybe `header with host`?

server/websocket_test.go
```diff
@@ -1661,6 +1741,21 @@ func TestWSValidateOptions(t *testing.T) {
...
+			o := wso.Clone()
+			o.Websocket.Headers = map[string]string{"Host": "http://localhost:8080"}
+			return o
+		}, `websocket: invalid header "Host" not allowed`},
+		{"header with ", func() *Options {
```
kozlovic, line 1754: And here, `header with no masking`?

kozlovic (changes requested):
Almost there!

server/websocket_test.go
ramonberrutti: Because headers is a http.Header is not possible to do a len(leaders) without casting. But if headers have something, it will print all the headers.
Adding a description and also changing the error message to be generic for multiple headers.

server/websocket_test.go
```diff
@@ -3778,6 +3895,24 @@ func TestWSNkeyAuth(t *testing.T) {
...
+	s := RunServer(o)
+	defer s.Shutdown()
+
+	opts := testWSClientOptions{
+		host:                 o.Websocket.Host,
+		port:                 o.Websocket.Port,
+		extraResponseHeaders: o.Websocket.Headers,
+	}
+
+	testNewWSClient(t, opts)
```
kozlovic, line 3913: Change to:
```
	c, _, _ := testNewWSClient(t, opts)
	c.Close()
```

kozlovic (changes requested):

aricart:
Just wondering how `Strict-Transport-Security` is actually used here. In the case of the config, it must have `no_tls`  (otherwise only wss is supported) configuration. This option is simply a guard for the http-server to inform the client that it should be using https for any other request (again here we are only serving either ws or wss). So the only time that this would be useful would be if `no_tls` is set, which effectively means that wss is not possible - is this some workaround for a proxy setup?

kozlovic:
@ramonberrutti I was asked to review this PR, but I agree with @aricart that, although this PR is implemented correctly, I have to wonder the need for it because if you configure the NATS Server for TLS, I am not sure how a client could connect without TLS. The original issue states:
```
Description: The remote web server is not enforcing HSTS, as defined by RFC 6797.
HSTS is an optional response header that can be configured on the server to instruct
the browser to only communicate via HTTPS. The lack of HSTS allows downgrade attacks,
SSL-stripping man-in-the-middle attacks, and weakens cookie-hijacking protections.
```
First, it says that it is an optional response header, so the fact that it says it is missing does not mean that it is incorrect, otherwise it would not be optional. Second, as I stated, I am not sure how a client can really connect without TLS if the NATS Server is configured to accept TLS only.

Nevertheless, I will continue my review on a pure technical sense, but me giving LGTM does not mean that the PR/issue has merit, just that your implementation of adding the response header is valid from a code perspective. I will let @bruth and @philpennock decide if this should be accepted or not.

kozlovic (approved):
LGTM. Again, the PR itself looks good, following our code style and adding proper tests, but I will let others decide if we merge this or not.

On a separate note: for your next PR, please refrain from doing a force push when making updates to the PR once someone has started reviewing it. It forces the reviewer to go over the whole PR to see what was really changed. Add new commits until the review is complete, after which it is ok to squash everything into a single commit. Thanks!

ramonberrutti:
I focused on a generic option to add extra response headers and not specifically implement an HSTS option.
You can achieve a similar solution by having a reverse proxy before nats and adding the response headers, also terminating TLS.

For example, Nginx config:
```
        location /ws {
            proxy_pass http://nats;
            proxy_http_version 1.1;
            proxy_set_header Upgrade $http_upgrade;
            proxy_set_header Connection $connection_upgrade;
            proxy_set_header Host $host;
            add_header Strict-Transport-Security "max-age=15768000 ; includeSubDomains" always;
        }
```
The always key is very important.

aricart:
I think the ability to set custom headers is fine. Just would like a good example of the ones that would be useful or relevant. I see the `X-Frame-Options` as one of them...

ramonberrutti:
I was reading about the important headers for WebSocket and I found another one always applied by Cloudflare:

```
X-Content-Type-Options: "nosniff"
```

If you enable HSTS will also inject Strict-Transport-Security

```
Strict-Transport-Security: "max-age=63072000; includeSubDomains; preload"
```

Discord WebSocket headers (including the CloudFlare headers)

```
HTTP/1.1 101 Switching Protocols
Date: Sun, 24 Mar 2024 22:23:30 GMT
Connection: upgrade
sec-websocket-accept: /Xiz07/yeCXXXXV1wXXXTgZ8gM0=
upgrade: websocket
CF-Cache-Status: DYNAMIC
Report-To: {"endpoints":[{"url":"https:\/\/a.nel.cloudflare.com\/report\/v4?s=xxxxxxxxx"}],"group":"cf-nel","max_age":604800}
NEL: {"success_fraction":0,"report_to":"cf-nel","max_age":604800}
Strict-Transport-Security: max-age=31536000; includeSubDomains; preload
X-Content-Type-Options: nosniff
Server: cloudflare
CF-RAY: X69a10e658cdxxxx-ZZZ
```

Reference: https://developers.cloudflare.com/workers/examples/security-headers/

bruth (approved):
I support these changes. @derekcollison will need to do the final approval. Thanks @ramonberrutti!

server/websocket.go
```diff
@@ -1019,6 +1023,24 @@ func validateWebsocketOptions(o *Options) error {
...
+		case "host",
+			"content-length",
+			"connection",
+			"upgrade",
+			"nats-no-masking":
+			return fmt.Errorf("websocket: invalid header %q not allowed", key)
+		}
+
+		if strings.HasPrefix(k, "sec-websocket-") {
+			return fmt.Errorf("websocket: invalid header %q, \"Sec-WebSocket-\" prefix no allowed", key)
```
derekcollison, line 1040: prefix _not_ allowed?

derekcollison (approved):
LGTM

PR #5233, opened 2024-03-21
[IMPROVED] Waiting queue for pull consumers

derekcollison:
Moved waiting queue for pull requests to linked-list. We had issues with very large MaxWaiting configured and causing memory and GC issues, especially if compaction kicked in.

Also when new requests for msgs are received and we are full, only check waiting queue at most once per second to avoid CPU pressure when application is continually sending new requests when the queue is full.


wallyqs (approved):
LGTM

Issue #5236, opened 2024-03-22
JetStream: panic: runtime error: makeslice: cap out of range


kylemcc:
### Observed behavior

NATS server (running only JetStream) is panicking when attempting to allocate a buffer [here](https://github.com/nats-io/nats-server/blob/v2.10.12/server/filestore.go#L5242), in `indexCacheBuf` during compaction.

Stacktrace:
```
panic: runtime error: makeslice: cap out of range
goroutine 202019 [running]:
github.com/nats-io/nats-server/v2/server.(*msgBlock).indexCacheBuf(0xc00616b040, {0xc01006e000, 0x5a5ca, 0x5a5ca})
github.com/nats-io/nats-server/v2/server/filestore.go:5242 +0x1d4
github.com/nats-io/nats-server/v2/server.(*msgBlock).loadMsgsWithLock(0xc00616b040)
github.com/nats-io/nats-server/v2/server/filestore.go:5685 +0x489
github.com/nats-io/nats-server/v2/server.(*msgBlock).compact(0xc00616b040)
github.com/nats-io/nats-server/v2/server/filestore.go:3865 +0x99
github.com/nats-io/nats-server/v2/server.(*fileStore).syncBlocks(0xc0036fd500)
github.com/nats-io/nats-server/v2/server/filestore.go:5103 +0x35c
created by time.goFunc
time/sleep.go:176 +0x2d
```

### Expected behavior

No panic :)

### Server and client version

$ nats-server --version
nats-server: v2.10.12

## Relevant NATS Config

Cluster size: 5

`GOMEMLIMIT` is set to 70% of  the available RAM on the machine

This doesn't appear to be environment-specific, but sharing anyway:

## Environment

Cloud: AWS
Instance type: `m7i-flex.xlarge`
vCPUs: 4
RAM: 16GiB
Disk:  1 TiB EBS gp3
Container runtime: none

## OS

Amazon Linux 2

```
# uname -a
Linux ip-172-16-30-97.us-west-2.compute.internal 4.14.334-252.552.amzn2.x86_64 #1 SMP Tue Jan 2 17:47:37 UTC 2024 x86_64 x86_64 x86_64 GNU/Linux
```

```
#  cat /etc/os-release
NAME="Amazon Linux"
VERSION="2"
ID="amzn"
ID_LIKE="centos rhel fedora"
VERSION_ID="2"
PRETTY_NAME="Amazon Linux 2"
ANSI_COLOR="0;33"
[...]

PR #5238, opened 2024-03-22
[FIXED] Protect against corrupt msgBlock first or last sequence when doing indexing

derekcollison:
Resolves: #5236 

Although unclear how the sequences got corrupted in the first place.


wallyqs (approved):
LGTM

PR #5246, opened 2024-03-26
Tweak `clone` behaviour of `User` and `NkeyUser`

neilalexander:
In this case, an empty-but-non-nil map would not be copied, which could result in the clone referring to the original map.

Also implement a deep copy of `AllowedConnectionTypes` for `NkeyUser`, like `User` already has.


derekcollison (approved):
LGTM

PR #5247, opened 2024-03-26
Constant-time evaluation of non-bcrypt passwords

neilalexander:
This ensures that comparison of non-bcrypted passwords is done in constant time, whereas otherwise timing information may be apparent on a network with next to no jitter.

### DIFF
--- a/server/auth.go
+++ b/server/auth.go
@@ -15,6 +15,7 @@ package server
 
 import (
 	"crypto/sha256"
+	"crypto/subtle"
 	"crypto/tls"
 	"crypto/x509/pkix"
 	"encoding/asn1"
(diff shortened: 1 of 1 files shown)


server/auth.go
```diff
@@ -1418,8 +1419,14 @@ func comparePasswords(serverPassword, clientPassword string) bool {
...
 			return false
 		}
-	} else if serverPassword != clientPassword {
-		return false
+	} else {
+		// stringToBytes should be constant-time near enough compared to
+		// turning a string into []byte normally.
+		spass := stringToBytes(serverPassword)
+		cpass := stringToBytes(clientPassword)
+		if subtle.ConstantTimeCompare(spass, cpass) == 0 {
```
derekcollison, line 1427: So this will slow down reconnect yes?

neilalexander: Should be marginal, standard string comparison will give up on the first mismatching character, whereas `ConstantTimeCompare` iterates over all characters before returning.

derekcollison (approved):
LGTM

server/auth.go
philpennock: Should be about no difference in time for valid credentials. Should only make invalid credential rejection slower.

PR #5250, opened 2024-03-26
(2.12) [ADD] no responders return the original subject

ramonberrutti:
Using [Request-Reply](https://docs.nats.io/nats-concepts/core-nats/reqreply) is very useful.

When using Request with multiple replays, it is difficult to know when the requester loses interest in the replays.
Today, Nats returns an error 503 to the replay subject without any extra information.

### DIFF
--- a/server/client.go
+++ b/server/client.go
@@ -4211,7 +4211,8 @@ func (c *client) processInboundClientMsg(msg []byte) (bool, bool) {
 		c.mu.Lock()
 		if c.opts.NoResponders {
 			if sub := c.subForReply(c.pa.reply); sub != nil {
-				proto := fmt.Sprintf("HMSG %s %s 16 16\r\nNATS/1.0 503\r\n\r\n\r\n", c.pa.reply, sub.sid)
+				hdrLen := 32 /* header without the subject */ + len(c.pa.subject)
+				proto := fmt.Sprintf("HMSG %s %s %d %d\r\nNATS/1.0 503\r\nNats-Subject: %s\r\n\r\n\r\n", c.pa.reply, sub.sid, hdrLen, hdrLen, c.pa.subject)
 				c.queueOutbound([]byte(proto))
 				c.addToPCD(c)
 			}
--- a/server/client_test.go
+++ b/server/client_test.go
@@ -251,7 +251,7 @@ func TestClientNoResponderSupport(t *testing.T) {
 	if len(am) == 0 {
 		t.Fatalf("Did not get a match for %q", l)
 	}
-	checkPayload(cr, []byte("NATS/1.0 503\r\n\r\n"), t)
+	checkPayload(cr, []byte("NATS/1.0 503\r\nNats-Subject: foo\r\n\r\n\r\n"), t)
 }
 
 func TestServerHeaderSupport(t *testing.T) {


server/client_test.go
```diff
@@ -252,7 +252,7 @@ func TestClientNoResponderSupport(t *testing.T) {
 	if len(am) == 0 {
 		t.Fatalf("Did not get a match for %q", l)
 	}
-	checkPayload(cr, []byte("NATS/1.0 503\r\n\r\n"), t)
+	checkPayload(cr, []byte("NATS/1.0 503\r\nNats-Subject: foo\r\n\r\n\r\n"), t)
```
ramonberrutti, line 255: I'm also checking that the body is empty.

server/client.go
```diff
@@ -4051,7 +4051,8 @@ func (c *client) processInboundClientMsg(msg []byte) (bool, bool) {
 		c.mu.Lock()
 		if c.opts.NoResponders {
 			if sub := c.subForReply(c.pa.reply); sub != nil {
-				proto := fmt.Sprintf("HMSG %s %s 16 16\r\nNATS/1.0 503\r\n\r\n\r\n", c.pa.reply, sub.sid)
+				hdrLen := 32 /* header without the subject */ + len(c.pa.subject)
+				proto := fmt.Sprintf("HMSG %s %s %d %d\r\nNATS/1.0 503\r\nNats-Subject: %s\r\n\r\n\r\n", c.pa.reply, sub.sid, hdrLen, hdrLen, c.pa.subject)
```
ramonberrutti, line 4055: I picked Nats-Subject as the header name because it is used elsewhere.

aricart: In that case it means the original subject the current message was published as - not sure we want to use that.

aricart: Also the only way you would get this is if you are responding and adding a reply subject - which means that when the server answers with no responders, you know what the reply subject that you published a response to is.

ramonberrutti: Why I need the feature:

- Service A: pub reply to  MyService.Endpoint  with a _INBOX.<hash>
- Service B: sub to MyService.Endpoint  and publish multiple times (grpc server stream) to the client _INBOX.hash with a reply option to MyService.Endpoint. 
- The server tracks each inbox in a hashmap and deletes it when it receives a 503.

To resolve the issue, I'm doing something similar but my service subscribe to MyService.> and setting the reply as: MyService.Endpoint.<hash>, I also could have two subscriptions, one for MyService.Endpoint and another for MyService.Endpoint.>

ramonberrutti: Hi, @aricart. Can we review and consider this change?

neilalexander:
@ripienaar Seems reasonable for me for a 2.12 change, WDYT?

ripienaar:
Yeah I thinking does make sense

aricart:
at some point I remember Matthias had said there was a mechanism to figure out if there was interest on a subject, is this still true or was this retracted?

neilalexander:
Hey @ramonberrutti, would you mind rebasing on top of latest `main`?

ramonberrutti:
> Hey @ramonberrutti, would you mind rebasing on top of latest `main`?

Sorry for the delay, done.

Added the Signed-off too.

neilalexander (approved):
LGTM, thanks!

PR #5251, opened 2024-03-27
Added [ND]DEATH timestamp rewrite

levb:
- [ ] Add benchmark comparison to `main`

### DIFF
--- a/server/mqtt.go
+++ b/server/mqtt.go
@@ -5023,8 +5023,8 @@ func (c *client) mqttEnqueuePublishMsgTo(cc *client, sub *subscription, pi uint1
 
 	retain := false
 	isBirth, isDeath, isCertificate := sparkbParseBirthDeathTopic(topic)
-	if isBirth && isCertificate {
-		retain = true
+	if isBirth && qos == 0 {
+		retain = isCertificate
 	}
 	if isDeath && !isCertificate {
 		msg = sparkbReplaceDeathTimestamp(msg)
--- a/server/mqtt.go
+++ b/server/mqtt.go
@@ -4268,6 +4268,7 @@ func (c *client) mqttHandlePubRetain() {
 	pp := c.mqtt.pp
 	retainMQTT := mqttIsRetained(pp.flags)
 	isBirth, _, isCertificate := sparkbParseBirthDeathTopic(pp.topic)
+	retainSparkbBirth := isBirth && !isCertificate
 
 	// [tck-id-topics-nbirth-mqtt] NBIRTH messages MUST be published with MQTT
 	// QoS equal to 0 and retain equal to false.
@@ -4276,8 +4277,6 @@ func (c *client) mqttHandlePubRetain() {
 	// Server MUST make NBIRTH messages available on the topic:
 	// $sparkplug/certificates/namespace/group_id/NBIRTH/edge_node_id with the
 	// MQTT retain flag set to true.
-
-	retainSparkbBirth := isBirth && !isCertificate
 	if retainMQTT == retainSparkbBirth {
 		// (retainSparkbBirth && retainMQTT) : not valid, so ignore altogether.
 		// (!retainSparkbBirth && !retainMQTT) : nothing to do.
--- a/server/mqtt.go
+++ b/server/mqtt.go
@@ -191,15 +191,15 @@ const (
 )
 
 const (
-	sparkbNamespace    = "spBv1.0"
-	sparkbCertificates = "$sparkplug/certificates/spBv1.0"
-	sparkbNBIRTH       = "NBIRTH"
-	sparkbDBIRTH       = "DBIRTH"
+	sparkbNBIRTH = "NBIRTH"
+	sparkbDBIRTH = "DBIRTH"
+	sparkbNDEATH = "NDEATH"
+	sparkbDDEATH = "DDEATH"
 )
 
 var (
-	sparkbNamespacePrefix    = []byte("spBv1//0.")
-	sparkbCertificatesPrefix = []byte("$sparkplug.certificates.spBv1//0.")
+	sparkbNamespaceTopicPrefix    = []byte("spBv1.0/")
+	sparkbCertificatesTopicPrefix = []byte("$sparkplug/certificates/")
 )
 
 var (
@@ -4267,8 +4267,9 @@ func (s *Server) mqttProcessPubRel(c *client, pi uint16, trace bool) error {
 func (c *client) mqttHandlePubRetain() {
 	pp := c.mqtt.pp
 	isRetained := mqttIsRetained(pp.flags)
-	isSparkbBirth, typ, groupID, nodeID, deviceID := sparkbIsBirth(pp.subject, sparkbNamespacePrefix)
-	if !isRetained && !isSparkbBirth {
+	isBirth, _, isCertificate := sparkbParseBirthDeathTopic(pp.topic)
+	retainSparkbBirth := isBirth && !isCertificate
+	if !isRetained && !retainSparkbBirth {
 		return
 	}
 	asm := c.mqtt.asm
@@ -4296,7 +4297,7 @@ func (c *client) mqttHandlePubRetain() {
 		Source: c.opts.Username,
 	}
 
-	if isSparkbBirth {
+	if retainSparkbBirth {
 		// [tck-id-conformance-mqtt-aware-store] A Sparkplug Aware MQTT Server
 		// MUST store NBIRTH and DBIRTH messages as they pass through the MQTT
 		// Server.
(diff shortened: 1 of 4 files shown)


levb:
@kozlovic if it makes it easier to review, feel free to merge this PR into `lev-spB` branch

server/mqtt.go
```diff
@@ -4267,8 +4267,9 @@ func (s *Server) mqttProcessPubRel(c *client, pi uint16, trace bool) error {
 func (c *client) mqttHandlePubRetain() {
 	pp := c.mqtt.pp
 	isRetained := mqttIsRetained(pp.flags)
-	isSparkbBirth, typ, groupID, nodeID, deviceID := sparkbIsBirth(pp.subject, sparkbNamespacePrefix)
-	if !isRetained && !isSparkbBirth {
+	isBirth, _, isCertificate := sparkbParseBirthDeathTopic(pp.topic)
```
kozlovic, line 4270: My original question was, are those spark messages meant to have the retain flag when they are received by the server. If they MUST not, then we should probably check that here and fail if that is the case?

server/mqtt.go
```diff
@@ -4309,14 +4310,13 @@ func (c *client) mqttHandlePubRetain() {
...
+		subject, _ := mqttTopicToNATSPubSubject(topic)
+		rm.Topic = string(topic)
+		rm.Subject = string(subject)
 
 		// will use to save the retained message.
-		key = string(byteSubject)
+		key = string(subject)
 
 		// Store the retained message with the RETAIN flag set.
 		rm.Flags |= mqttPubFlagRetain
```
kozlovic, line 4322: This seem to indicate that they may not have the flag on inbound, so again, don't we need to make sure that they don't have it?

server/mqtt.go
```diff
@@ -4618,29 +4618,30 @@ func mqttIsRetained(flags byte) bool {
...
+	if !bytes.HasPrefix(topic, sparkbNamespaceTopicPrefix) {
+		return false, false, false
 	}
+	topic = topic[len(sparkbNamespaceTopicPrefix):]
 
-	groupID = string(natsSubjectToMQTTTopic(parts[0]))
-	nodeID = string(natsSubjectToMQTTTopic(parts[2]))
-	if isDevice && len(parts) > 3 {
-		deviceID = string(natsSubjectToMQTTTopic(parts[3]))
+	parts := bytes.Split(topic, []byte("/"))
```
kozlovic, line 4631: Maybe `[]byte{'/'}` instead?

server/mqtt.go
```diff
@@ -4618,29 +4618,30 @@ func mqttIsRetained(flags byte) bool {
...
+	if len(parts) < 3 || len(parts) > 4 {
+		return false, false, false
+	}
+	typ := string(parts[1])
```
kozlovic, line 4635: Using `typ := bytesToString(parts[1])` here I think would be safe since it is a conversion just for checking content but we don't reference `typ` outside of here.

server/mqtt.go
```diff
@@ -4958,9 +4947,75 @@ func isMQTTReservedSubscription(subject string) bool {
...
+	buf := bytes.NewBuffer(make([]byte, 0, len(msg)+16)) // 16 bytes should be enough if we need to add a timestamp
+	writeDeathTimestamp := func() {
+		// [tck-id-conformance-mqtt-aware-ndeath-timestamp] A Sparkplug Aware
+		// MQTT Server MAY replace the timestamp of NDEATH messages. If it does,
+		// it MUST set the timestamp to the UTC time at which it attempts to
+		// deliver the NDEATH to subscribed clients
+		//
+		// sparkB spec: 6.4.1. Google Protocol Buffer Schema
+		//      optional uint64 timestamp = 1; // Timestamp at message sending time
+		ts := uint64(time.Now().UnixMilli())
```
kozlovic, line 4964: A comment here saying that using time in millisecond is required by the spec would help.

server/mqtt.go
```diff
@@ -4958,9 +4947,75 @@ func isMQTTReservedSubscription(subject string) bool {
...
+		buf.Write(protoEncodeVarint(TIMESTAMP<<3 | VARINT))
+		buf.Write(protoEncodeVarint(ts))
+	}
+
+	for len(msg) > 0 {
+		numField, typ, size, err := protoScanField(msg)
+		if err != nil {
+			return orig
+		}
+		if typ != VARINT || numField != TIMESTAMP {
```
kozlovic, line 4974: Is `numField` here an id for the timestamp field or the position it should be in the message? In other words, if timestamp is present, MUST it be the first or not?

server/mqtt.go
```diff
@@ -4958,9 +4947,75 @@ func isMQTTReservedSubscription(subject string) bool {
...
+	// [tck-id-conformance-mqtt-aware-dbirth-mqtt-retain] A Sparkplug Aware
+	// MQTT Server MUST make DBIRTH messages available on the topic:
+	// $sparkplug/certificates/namespace/group_id/DBIRTH/edge_node_id/device_id
+	// with the MQTT retain flag set to true
+	//
+	// $sparkplug/certificates messages are sent as NATS messages, so we
+	// need to add the retain flag when sending them to MQTT ciients.
+
+	retain := false
+	isBirth, isDeath, isCertificate := sparkbParseBirthDeathTopic(topic)
```
kozlovic, line 5011: I hope that the check (we start with a bytes.HasPrefix()) is fast enough to return false for non sparkb birt/death messages to reduce perf impact.

server/mqtt.go
```diff
@@ -4958,9 +4947,75 @@ func isMQTTReservedSubscription(subject string) bool {
...
+	if isBirth && isCertificate {
+		retain = true
+	}
+	if isDeath && !isCertificate {
```
kozlovic, line 5015: Can a message be a birth and death? In other words, should we have a `else` instead of separate `if`?

server/mqtt.go
```diff
@@ -4958,9 +4947,75 @@ func isMQTTReservedSubscription(subject string) bool {
...
 
+func sparkbReplaceDeathTimestamp(msg []byte) []byte {
+	const VARINT = 0
+	const TIMESTAMP = 1
+
+	orig := msg
+	buf := bytes.NewBuffer(make([]byte, 0, len(msg)+16)) // 16 bytes should be enough if we need to add a timestamp
+	writeDeathTimestamp := func() {
+		// [tck-id-conformance-mqtt-aware-ndeath-timestamp] A Sparkplug Aware
+		// MQTT Server MAY replace the timestamp of NDEATH messages. If it does,
```
kozlovic, line 4958: Since all this seem optional, have we decided that we need this feature? I see that you have to "borrow" protobuf implementation (to reduce deps).

kozlovic (changes requested):

server/mqtt.go
levb: It's the field's numeric ID. The field order is indeterministic, according to the protobuf spec. I'll rename to be more intention-revealing.

server/mqtt.go
levb: Right. Like you said in the main PR, we can put this on an `if config option`. Otherwise I don't see how we can avoid prefix-matching all incoming MQTT messages for DEATH (even if we could somehow short circuit the BIRTH handling).

I will compare and paste the before/after benchmark results.

server/mqtt.go
levb: Is the concern here the performance or intention revealing? There are 3 "interesting" states here - `do-not-change`, `DEATH` (not certificate), or `BIRTH certificate`. Since DEATH messages are not supposed to appear in the certificates subject space, 0/5 I could express this as:
```go
switch {
case isBirth:
    retain = isCertificate
case isDeath && !isCertificate:
    msg = sparkbReplaceDeathTimestamp(msg)
default:
    // no change
}
```

server/mqtt.go
levb: It is optional in the spec, but the official test that is required to get listed requires it.

2/5 we should support it when SparkplugB-aware. Whether or not configurable altogether with a `sparkplugb: true` - a separate decision?

levb: Let's keep the config conversation in the "main" PR, add there if we decide to?

server/mqtt.go
levb: _[tck-id-topics-nbirth-mqtt] NBIRTH messages MUST be published with MQTT QoS equal to 0
and retain equal to false._

I'll make the change to treat it as an error, even though I am 0/5 on whether we should be enforcing these NODE and DEVICE aspects of the spec - we don't fail for an invalid the QoS for example.

server/mqtt.go
levb: Resolving this thread in favor of ^^

levb:
@kozlovic I'll run the benchmarks and paste the results, but otherwise please review again? We will address the config option issue in the main PR?

server/mqtt.go
kozlovic: It was just to say that if a message cannot be both `isBirth` and `isDeath`, then logically speaking it is one or the other, so more like `if .. else ..`. That being said, it is true that in Golang, there is extensive use of `switch` instead of `if`, but I guess it is a matter of clarity/conciseness. 
```
if isBirth && isCertificate {
    retain = true
} else if isDeath && !isCertificate {
    msg = sparkbReplaceDeathTimestamp(msg)
}
```

server/mqtt.go
kozlovic: But also, in the other PR, you did this equivalent check only in QoS0 callback, where putting the check here applies to all QoS's. Again, not that it matters much performance-wise for QoS > 0 anyway (since they are comparatively slow).

kozlovic (commented):
More questions before LGTM.

server/mqtt.go
levb: Right. I moved it here because we have to parse for DEATH anyway. Will add the QOS check, too.

server/mqtt.go
levb: k, then really,
```go
if isBirth {
   retain = isCertificate
} else ...same
```

will change.

server/mqtt.go
levb: • [tck-id-message-flow-edge-node-birth-publish-will-message-qos] The Edge Node’s MQTT Will
Message’s MQTT QoS MUST be 1.

I changed the test but checking the DEATH qos proved more difficult since we may deliver it as a QoS0 message.

Now checking BIRTH to be qos0

kozlovic (approved):
LGTM

levb:
@kozlovic fixed the `else` - sorry was rushing it and missed.

PR #5254, opened 2024-03-28
Update users.conf

antoniasouza:


philpennock:
This PR appears to be trying to add hard-coded back door static credentials to an example config file.  Not only do we not want to merge this, I believe that it would now be illegal to merge this in many jurisdictions, which have finally awoken to the dangers of shipping with static default credentials.

rapiz1:
It's hard not to relate to the recent xz incident

PR #5265, opened 2024-04-02
[FIXED] Fix for a race panic on macosx/darwin

derekcollison:
Was getting a panic of checkptr: converted pointer straddles multiple allocations in go 1.22.1 on darwin.


bruth (commented):
Confirmed this worked locally for me. @neilalexander provided a fix for this as well, so I defer to him on this one.

PR #5268, opened 2024-04-03
There is a useless memory allocation

chenjpu:
When trace flags is closed and trace header is set, it leads to useless string memory allocation

Signed-off-by:  chenjpu <chenjpu@gmail.com>

### DIFF
--- a/server/parser.go
+++ b/server/parser.go
@@ -495,8 +495,10 @@ func (c *client) parse(buf []byte) error {
 			// Check for mappings.
 			if (c.kind == CLIENT || c.kind == LEAF) && c.in.flags.isSet(hasMappings) {
 				changed := c.selectMappedSubject()
-				if (trace || mt != nil) && changed {
-					c.traceInOp("MAPPING", []byte(fmt.Sprintf("%s -> %s", c.pa.mapped, c.pa.subject)))
+				if changed {
+					if trace {
+						c.traceInOp("MAPPING", []byte(fmt.Sprintf("%s -> %s", c.pa.mapped, c.pa.subject)))
+					}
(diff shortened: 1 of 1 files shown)


server/parser.go
```diff
@@ -495,8 +495,10 @@ func (c *client) parse(buf []byte) error {
 			// Check for mappings.
 			if (c.kind == CLIENT || c.kind == LEAF) && c.in.flags.isSet(hasMappings) {
 				changed := c.selectMappedSubject()
-				if (trace || mt != nil) && changed {
-					c.traceInOp("MAPPING", []byte(fmt.Sprintf("%s -> %s", c.pa.mapped, c.pa.subject)))
+				if changed {
```
kozlovic, line 498: I guess we could even do without the `changed` boolean, although it adds clarity, so I don't mind keeping it. But yes, it could be:
```
if c.selectMappedSubject() {
    if trace {
         c.traceInOp("MAPPING", []byte(fmt.Sprintf("%s -> %s", c.pa.mapped, c.pa.subject)))
    }
    // c.pa.subject is the subject the original is now mapped to.
    mt.addSubjectMappingEvent(c.pa.subject)
}
```
We don't need to check for `mt != nil` before invoking `mt.addSubjectMappingEvent()` because this function checks for it.

kozlovic (approved):
LGTM, but could even get rid of the `changed` boolean.

server/parser.go
chenjpu: At the beginning, I didn't think ```changed``` was needed, but it did increase clarity, so I kept it

derekcollison:
So we are good as is to merge yes?

PR #5270, opened 2024-04-03
[IMPROVED] Improvements to interest policy and workqueue clustered streams.

derekcollison:
Specifically around interest and limited streams that are clustered. Since we operate consumers and streams independently, when a stream is removing msgs when acked and acks are coming into different servers at different times, etc. Messages can get stranded.

We introduce a better check of the interest state based on a collective low ack floor that will purge our copy of the stream to catch up. We run this periodically in monitorStream and during certain conditions.

Thanks to Wally for introducing the test that stresses this part of the system.

### DIFF
--- a/server/consumer.go
+++ b/server/consumer.go
@@ -1392,6 +1392,10 @@ func (o *consumer) setLeader(isLeader bool) {
 		} else if o.srv.gateway.enabled {
 			stopAndClearTimer(&o.gwdtmr)
 		}
+		// If we were the leader make sure to drain queued up acks.
+		if wasLeader {
+			o.ackMsgs.drain()
+		}
 		o.mu.Unlock()
 
 		// Unregister as a leader with our parent stream.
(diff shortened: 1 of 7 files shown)


server/stream.go
```diff
@@ -5425,24 +5439,87 @@ func (mset *stream) getPublicConsumers() []*consumer {
...
+				np, _ = o.calculateNumPending()
+			}
+			o.mu.RUnlock()
+			// This means we have pending and can not remove anything at this time.
+			if np > 0 {
+				return
+			}
+		}
+		// Purge the stream to lowest ack floor + 1
+		mset.store.PurgeEx(_EMPTY_, lowAckFloor+1, 0)
```
neilalexander, line 5520: Is it safe to ignore the return error here?

derekcollison: The thought here is that we are trying to clean up any orphaned messages below the low ack floor. If the purge fails the condition would simply remain the same. Also not sure what, outside of logging, we would do with an error at this point in the code.

neilalexander (approved):
LGTM

server/stream.go
neilalexander: Fair enough, this was the only thing that stood out to me in particular

server/consumer.go
```diff
@@ -2534,19 +2538,18 @@ func (o *consumer) applyState(state *ConsumerState) {
 		return
 	}
 
-	// If o.sseq is greater don't update. Don't go backwards on o.sseq.
-	if o.sseq <= state.Delivered.Stream {
+	// If o.sseq is greater don't update. Don't go backwards on o.sseq if leader.
+	if !o.isLeader() || o.sseq <= state.Delivered.Stream {
```
yordis, line 2542:
```suggestion
	if o.isFollower() || o.sseq <= state.Delivered.Stream {
```

Is there a value in adding more Domain-specific language?

neilalexander: Should be noted that there's no such function as `(*consumer).isFollower()` right now. IMO "not leader" is clear enough.

yordis: Yeah, the intent was to suggest to continue expanding the Domain-Specific code, instead of relying on Symbols.

PR #5272, opened 2024-04-03
systemd hardened example: add more guidance

philpennock:
For the example hardened systemd file, which is derived from config files which Synadia uses in production, add more examples:

 * Show setting GOMEMLIMIT env var
   + but also nudge towards using the external environment file for it
 * Use an external environment config file if it exists
 * Show mount-point dependency for JetStream
 * Configure automatic restarts
 * Show using InaccessiblePaths to lock down FS access even further
 * Setup a service alias of simply 'nats'.

### DIFF
--- a/util/nats-server-hardened.service
+++ b/util/nats-server-hardened.service
@@ -2,21 +2,35 @@
 Description=NATS Server
 After=network-online.target ntp.service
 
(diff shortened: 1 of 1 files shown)


derekcollison (approved):
LGTM - Thanks.

PR #5273, opened 2024-04-03
JetStream TPM driven encryption

ColinSullivan1:
## JetStream TPM based Encryption

This PR applies to Windows Platforms. uses the seal/unseal functionality of the TPM to encrypt and decrypt a JetStream encryption key, similar to how disk encryption works on windows platforms.

How it works:
Create an instance of the Storage Root Key (SRK) in the TPM.
If existing JS Encryption keys do not exist on disk:
 1. Create a JetStream encryption key (js key) and seal it to the SRK using a provided js encryption key password.
 2. Save the public and private blobs to a file on disk.
 3. Write the new js encryption keys to disk.

Otherwise (keys exist on disk)
 1. Read the keys blobs from disk
 2. Load them into the TPM
 3. Unseal the js key using the TPM and the provided js encryption keys password.

Configuration (full):

```text
jetstream {
    tpm {
        keys_file: "jskeys.json"
        encryption_password: $JSEKTPM_PASSWORD
        srk_password: $TPM_SRK_PASSWORD
        pcr: 19
        cipher: AES
    }
```

Configuration (Usual/Minimal):

```text
jetstream {
    tpm {
        keys_file: "jskeys.json"
        encryption_password: $JSEKTPM_PASSWORD
    }
```

The configured defaults should work on factory settings of windows TPMs.

- pcr (Platform Configuration Register): 22
- cipher (used to encrypt JS Blocks): CHACAPOLY
- srk_password (: (empty)

The created JS Key is simply an NKEY, but can be anything truly random serialized to a string.

Applies to windows builds although could be extended to linux in the future.

Note: Before merging another maintainer should test this on a windows 10+ laptop with a TPM version 2.0.

Hat tip to @philpennock for assistance in reviewing the approach and initial test code.

Signed-off-by: Colin Sullivan <colin@luxantsolutions.com>


server/tpm/js_ek_tpm_windows.go
```diff
@@ -0,0 +1,283 @@
...
+	Version    int    // json: "version"
+	PrivateKey []byte // json: "privatekey"
+	PublicKey  []byte // json: "publickey"
+}
+
+// Writes the private and public blobs to disk in a single file. If the directory does
+// not exist, it will be created. If the file already exists it will be overwritten.
+func writeTPMKeysToFile(filename string, privateBlob []byte, publicBlob []byte) error {
+	keyDir := filepath.Dir(filename)
+	if err := os.MkdirAll(keyDir, 0755); err != nil {
```
neilalexander, line 87: Wondering if this needs to be `0755` or if we could get away with `0750`.

server/tpm/js_ek_tpm_windows.go
```diff
@@ -0,0 +1,283 @@
...
+		PrivateKey: []byte(base64.StdEncoding.EncodeToString(privateBlob)),
+		PublicKey:  []byte(base64.StdEncoding.EncodeToString(publicBlob)),
+	}
+	// Convert to JSON
+	keysJSON, err := json.Marshal(tpmKeys)
+	if err != nil {
+		return fmt.Errorf("unable to marshal keys to JSON: %v", err)
+	}
+	// Write the JSON to a file
+	if err := os.WriteFile(filename, keysJSON, 0644); err != nil {
```
neilalexander, line 105: Another world-readable as above.

server/tpm/js_ek_tpm_windows.go
```diff
@@ -0,0 +1,283 @@
...
+	if err := os.MkdirAll(keyDir, 0755); err != nil {
+		return fmt.Errorf("unable to create/access directory %q: %v", keyDir, err)
+	}
+
+	// Create a new set of persisted keys. Note that the private key doesn't necessary
+	// need to be protected as the TPM password is required to use unseal, although it's
+	// a good idea to put this in a secure location accessible to the server.
+	tpmKeys := natsTPMPersistedKeys{
+		Version:    JsKeyTPMVersion,
+		PrivateKey: []byte(base64.StdEncoding.EncodeToString(privateBlob)),
```
neilalexander, line 96: It's a nit but we can avoid the unnecessary round-tripping via `string` by doing something like:
```
tpmKeys := natsTPMPersistedKeys{
    Version: JsKeyTPMVersion,
}
tpmKeys.PrivateKey = base64.StdEncoding.EncodeAppend(tpmKeys.PrivateKey, privateBlob)
tpmKeys.PublicKey = base64.StdEncoding.EncodeAppend(tpmKeys.PublicKey, publicBlob)
```

server/tpm/js_ek_tpm_windows.go
```diff
@@ -0,0 +1,283 @@
...
+	if err := json.Unmarshal(keysJSON, &tpmKeys); err != nil {
+		return nil, nil, fmt.Errorf("unable to unmarshal TPM file keys JSON from %s: %v", filename, err)
+	}
+
+	// Placeholder for future-proofing. Here is where we would
+	// check the current version against tpmKeys.Version and
+	// handle any changes.
+
+	// Base64 decode the private and public blobs.
+	publicBlob, err := base64.StdEncoding.DecodeString(string(tpmKeys.PublicKey))
```
neilalexander, line 133: Same as above but with `DecodeAppend`.

server/tpm/js_ek_tpm_windows.go
```diff
@@ -0,0 +1,283 @@
...
+			KeyBits:    2048,
+			ModulusRaw: make([]byte, 256),
+		},
+	}
+	// Create the parent key against which to seal the data
+	srkHandle, _, err := tpm2.CreatePrimary(rwc, tpm2.HandleOwner, tpm2.PCRSelection{}, "", srkPassword, srkTemplate)
+	return srkHandle, err
+}
+
+type natsTPMPersistedKeys struct {
```
neilalexander, line 77: These comments are the wrong format and are therefore ignored, struct tags should be like this:
```
type natsTPMPersistedKeys struct {
 	Version    int    `json:"version"`
 	PrivateKey []byte `json:"private_key"`
 	PublicKey  []byte `json:"public_key"`
 }
```

server/tpm/js_ek_tpm_windows.go
```diff
@@ -0,0 +1,283 @@
...
+	if err := os.WriteFile(filename, keysJSON, 0644); err != nil {
+		return fmt.Errorf("unable to write keys file to %q: %v", filename, err)
+	}
+	return nil
+}
+
+// Reads the private and public blobs from a single file. If the file does not exist,
+// or the file cannot be read and the keys decoded, an error is returned.
+func readTPMKeysFromFile(filename string) ([]byte, []byte, error) {
+	if _, err := os.Stat(filename); os.IsNotExist(err) {
```
neilalexander, line 114: Wondering if this is necessary or whether `os.ReadFile(...)` would return an `os.IsNotExist(err)` too.

server/tpm/js_ek_tpm_windows.go
```diff
@@ -0,0 +1,283 @@
...
+// Unseals the JetStream encryption key from the TPM with the provided keys.
+// The key is returned as a string.
+func unsealJsEncrpytionKey(rwc io.ReadWriteCloser, pcr int, srkHandle tpmutil.Handle, srkPassword, objectPassword string, publicBlob, privateBlob []byte) (string, error) {
+	// Load the public/private blobs into the TPM for decryption.
+	objectHandle, _, err := tpm2.Load(rwc, srkHandle, srkPassword, publicBlob, privateBlob)
+	if err != nil {
+		return "", fmt.Errorf("unable to load data: %v", err)
+	}
+	defer func() {
+		tpm2.FlushContext(rwc, objectHandle)
```
neilalexander, line 185: Another nit but these can be single-lined: `defer tpm2.FlushContext(rwc, objectHandle)`

server/tpm/js_ek_tpm_windows.go
```diff
@@ -0,0 +1,283 @@
...
+	}
+	return sessHandle, policy, nil
+}
+
+// LoadJetStreamEncryptionKeyFromTPM loads the JetStream encryption key from the TPM.
+// If the keyfile does not exist, a key will be created and sealed. Public and private blobs
+// used to decrypt the key in future sessions will be saved to disk in the file provided.
+// The key will be unsealed and returned only with the correct password and PCR value.
+func LoadJetStreamEncryptionKeyFromTPM(srkPassword, jsKeyFile, jsKeyPassword string, pcr int) (string, error) {
+	var err error
```
neilalexander, line 247: Probably unnecessary as it is overshadowed by the following line.

server/jetstream_tpm_test.go
```diff
@@ -0,0 +1,260 @@
...
+		store_dir: %q
+		encryption_key: "foo"
+		tpm {
+			keys_file: "bar.json"
+			encryption_password: "s3cr3t!"
+		}
+	}`
+
+func getTPMFileName(t *testing.T) string {
+	return os.TempDir() + "/" + t.Name() + "/jskeys.json"
```
neilalexander, line 89: Avoid `os.TempDir()` in tests and use `t.TempDir()` instead, that way they are cleaned up automatically when the test ends.

This will probably mean you also don't need the `clearAndGetStoreDir` and `os.Remove()` defers.

neilalexander (commented):
I haven't ran any tests on a real machine yet but a few first-pass comments.

ColinSullivan1:
Thank you for the review @neilalexander, much appreciated! I aim to have these addressed by the end of the day tomorrow.

server/tpm/js_ek_tpm_windows.go
ColinSullivan1: Bitten by copilot!  Thanks.

ColinSullivan1:
I believe I've addressed the requested changes - thanks for the suggestions.

Certainly let me know if there are additional changes. It'd be great if someone could independently run the tests.

server/jetstream_tpm_test.go
ColinSullivan1: Will fix...

derekcollison:
Where are we on this in terms of getting it merged?

neilalexander (approved):
LGTM, although I don't have appropriate hardware to test it with.

mtmk (approved):
using hyper-v Windows 11 dev environment image:
* Ran with above config successfully; data was encrypted, password and keys json worked as expected.
* Also ran the TPM tests successfully.

LGTM

ecki:
> Is there a way to disable the PCR part of the TPM policy with this feature?

It seems to always pass exactly one PCR to the list, the value can be configured and defaults to 22. It is claimed that’s not a boot register, so maybe it’s not hardware bound? What problem do you try to solve? (But I think the config should allow lists instead)

friedrichsenm:
> > Is there a way to disable the PCR part of the TPM policy with this feature?
> 
> It seems to always pass exactly one PCR to the list, the value can be configured and defaults to 22. It is claimed that’s not a boot register, so maybe it’s not hardware bound? What problem do you try to solve? (But I think the config should allow lists instead)

I just wouldn't want to tie my TPM key to a particular PCR state if I'm not confident it will stay the same (barring more complex TPM policies that can let you update allowed PCR values). The scenario I'm imagining is that you have some long lived Windows server you're running which you have NATS on and after you apply your monthly updates, the PCR value changes and you no longer have access to the TPM key and then can't encrypt/decrypt the jetstream files.

ColinSullivan1:
You'll definitely want to be aware of PCR updates/changes/extensions in your monthly updates. However, I would guess this scenario to be an outlier - likely depending on the register you choose and other software installed.

Bitlocker uses a similar approach and lockouts seem to be rare (although it does generate a recovery key for this scenario).

PR #5274, opened 2024-04-03
[IMPROVED] Multi-Filtered Consumers

derekcollison:
This change introduces a new LoadNextMsgMulti into the store layer. It is passed a sublist that is maintained by the consumer. The store layer matches potential messages across any positive match in the sublist.

Resolves: #4888 


neilalexander (approved):
LGTM but wonder if we should consider #5275 for testing new functionality like this.

Jarema (commented):
Looks good!

Though I have one doubt which I commented about.

PR #5276, opened 2024-04-04
Fix some missing mutex unlocks in filestore and streams

neilalexander:
These error conditions didn't release held locks.


derekcollison (approved):
LGTM

PR #5281, opened 2024-04-07
[ADDED] Publish header "Nats-Expected-Last-Subject-Sequence-Subject"

### LINKED ISSUE #5280: Allow publishers to specify a different subject for "Nats-Expected-Last-Subject-Sequence"
### Proposed change

I'd like to request a change to allow publishers to provide a specific subject for use when nats-server enforced "Nats-Expected-Last-Subject-Sequence". Adding an additional header field that allows publishers to optionally change the subject used for "Nats-Expected-Last-Subject-Sequence" from the subject of the message being published to to a subject provided in the header.

The proposal is for a "Nats-Expected-Last-Subject-Sequence-Subject" header that when set with a string value becomes the subject used during enforcement of the last sequence. The header does not necessarily need to be that string exactly, the functionality is the actual proposal.

### Use case

The specific use case I am trying to solve here is to allow publishing to complex and dynamic subjects while still enabling optimistic concurrency control.

For example in an event sourcing application an aggregate might be defined against the subject:

- **orders.[order_id]** 
  - orders.1
  - orders.2
  - orders.3
  - ...

Today we can use "Nats-Expected-Last-Subject-Sequence" to ensure that when we publish to `orders.2` we are publishing against the known last sequence for messages on the `orders.2` subject.

However, it is often desirable to have more complex subject patterns. For example image a pattern that looks like this:

- **orders.[order_id].[event_type].[visibility]**
  - orders.1.created.public
  - orders.1.approved.private
  - orders.2.created.public
  - ...

Today we cannot use "Nats-Expected-Last-Subject-Sequence" with this pattern and still have OCC for the aggregate `orders.2` when publishing. If a publisher attempts to set the expected sequence and then publish to `order.2.rejected.private` the sequence for that subject is likely 0 as opposed to sequence for the last event published to `order.2.>`

By allowing publishers to override and customize the subject used in the expect last subject sequence check OCC can be enforced for the aggregate itself versus the subject being published to. For example by setting the proposed "Nats-Expected-Last-Subject-Sequence-Subject" to "order.2.>"

Relevant Slack Conversations:

- https://natsio.slack.com/archives/C055AB5M1KM/p1710872807434689
- https://natsio.slack.com/archives/CM3T6T7JQ/p1687313617160579

### Contribution

Possibly

cchamplin:
This change adds a new header
"Nats-Expected-Last-Subject-Sequence-Subject" when when paired with "Nats-Expected-Last-Subject-Sequence" allows publishers to customize the subject used when the server enforces
"Nats-Expected-Last-Subject-Sequence". Publishers can specify a alternative subject to be used that includes wildcards.

Resolves #5280

Signed-off-by: Caleb Champlin <caleb.champlin@gmail.com>

### DIFF
--- a/server/jetstream_test.go
+++ b/server/jetstream_test.go
@@ -11270,6 +11270,100 @@ func TestJetStreamLastSequenceBySubject(t *testing.T) {
 	}
 }
 
+func TestJetStreamLastSequenceBySubjectWithSubject(t *testing.T) {
+	for _, st := range []StorageType{FileStorage, MemoryStorage} {
+		t.Run(st.String(), func(t *testing.T) {
+			c := createJetStreamClusterExplicit(t, "JSC", 3)
+			defer c.shutdown()
+
+			nc, js := jsClientConnect(t, c.randomServer())
+			defer nc.Close()
+
+			cfg := StreamConfig{
+				Name:       "KV",
+				Subjects:   []string{"kv.>"},
+				Storage:    st,
+				Replicas:   3,
+				MaxMsgsPer: 1,
+			}
+
+			req, err := json.Marshal(cfg)
+			if err != nil {
+				t.Fatalf("Unexpected error: %v", err)
+			}
+			// Do manually for now.
+			m, err := nc.Request(fmt.Sprintf(JSApiStreamCreateT, cfg.Name), req, time.Second)
+			require_NoError(t, err)
+			si, err := js.StreamInfo("KV")
+			if err != nil {
+				t.Fatalf("Unexpected error: %v, respmsg: %q", err, string(m.Data))
+			}
+			if si == nil || si.Config.Name != "KV" {
+				t.Fatalf("StreamInfo is not correct %+v", si)
+			}
+
+			js.PublishAsync("kv.1.foo", []byte("1:1")) // Last is 1 for kv.1.foo; 1 for kv.1.*;
+			js.PublishAsync("kv.1.bar", []byte("1:2")) // Last is 2 for kv.1.bar; 2 for kv.1.*;
+			js.PublishAsync("kv.2.foo", []byte("2:1")) // Last is 3 for kv.2.foo; 3 for kv.2.*;
+			js.PublishAsync("kv.3.bar", []byte("3:1")) // Last is 4 for kv.3.bar; 4 for kv.3.*;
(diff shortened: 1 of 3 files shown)


server/jetstream_test.go
```diff
@@ -11269,6 +11269,100 @@ func TestJetStreamLastSequenceBySubject(t *testing.T) {
...
+				t.Helper()
+				m := nats.NewMsg(subj)
+				m.Data = []byte("HELLO")
+
+				// Expect last to be seq.
+				m.Header.Set(JSExpectedLastSubjSeq, seq)
+
+				// Constrain the sequence restriction to a specific subject
+				// e.g. "kv.1.*" for kv.1.foo, kv.1.bar, kv.1.baz; kv.2.* for kv.2.foo, kv.2.baz; kv.3.* for kv.3.bar
+				filterSubject := fmt.Sprintf("%s.*", subj[:strings.LastIndex(subj, ".")])
```
bruth, line 11329: The test would be a bit more clear if you lifted the matching subject as a parameter. This would also enable expansion for more nested subjects if/when is relevant.

bruth:
Use case aside, the one concern I have is that `LoadLastMsg` is not nearly as optimized with a pattern as it is with a concrete subject. This is already known as a less optimized access pattern and it will be another use case to drive that improvement forward.

The question is whether this lands and the latency is not to folks expectations. @cchamplin Have you created a realistic app/test case against this to assess if the writes/second are sufficient for your needs?

I did some raw benchmarking against the filestore API to assess, but will want to try it out in a more practical test case.

cchamplin:
@bruth We have not created the type of test case you described. We aren't quite mature enough yet to know if this will negatively impact our performance in a way that matters, we suspect not, we don't have that many subjects that would be matched by our sequence subject pattern. Ultimately though we'd be willing to take a minor performance hit on publishing to unlock this functionality, especially if it's something that could be optimized in the future.

bruth:
@neilalexander @ripienaar I advocate for this coming into 2.11. There is no new risk. For those who use it, the performance profile would be equivalent to any other code path that hits `LoadLastMsg` with a wildcard which is generally known to need improvement in the extreme cases.

All that said, would like to hear your thoughts.

ripienaar:
Yeah seems like a useful feature, for sure 2.11 at the earliest

derekcollison (approved):
LGTM

PR #5287, opened 2024-04-09
[IMPROVED] Interest based streams and consumers

derekcollison:
Updates and improvements to consistency of stream peers with interest retention and limits.
We moved msgId de-dupe processing to upper layer if clustered. 
Fixed interest state checks for streams and consumers.

For both filestore and memstore, if we are discard new and interest based retention we should not enforce at the store level.
This could lead to replica drift and sequences being assigned differently during server restarts, network hiccups.

Fixed bug in filestore that would not leave a tombstone during PurgeEx or Compact leading to recovery of a zero state.

### DIFF
--- a/server/consumer.go
+++ b/server/consumer.go
@@ -2830,10 +2830,10 @@ func (o *consumer) processAckMsg(sseq, dseq, dc uint64, doSample bool) {
 						break
 					}
 				}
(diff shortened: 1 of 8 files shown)


wallyqs (approved):
LGTM

PR #5290, opened 2024-04-09
Fix race condition in `checkInterestState` by taking write lock

neilalexander:
The call to `o.applyState()` assumes that a write lock is held, whereas we were only taking a read lock.


derekcollison (approved):
LGTM - Thanks.

derekcollison:
Tests seem to be consistently failing..

derekcollison:
image

neilalexander:
Yep, just noticed that... pushed another commit that should fix it.

PR #5293, opened 2024-04-09
[FIXED] Do normal purge if ack floor is higher then our last. Do not skip.

derekcollison:
Saw a flapping test that was not properly clearing out messages and this was why.


wallyqs (approved):
LGTM

PR #5295, opened 2024-04-10
[FIXED]  Creating colliding workqueue consumers could sometimes succeed.

derekcollison:
When creating multiple workqueue consumers that collided, if the consumer landed on the meta-leader it would be created.

This was due to the fact that the meta-leader had the consumer assignment already and incorrectly passed wasExisting for isRecovering. On recovery we would allow from a previous bug in the 2.9 series and print a warning to the server log.

The fix is to not mark waExisting if the consumer assignment is not marked as recovering, and to pass the state of isRecovering from the js state itself.

Also added upper layer check to avoid thrashing the meta layer.

If the parent stream was an R1 the consumer would be created, on R>1 it would be created and then a non meta-leader server would fail and clean it up.

Resolves: #4957 

### DIFF
--- a/server/consumer.go
+++ b/server/consumer.go
@@ -604,7 +604,7 @@ func checkConsumerCfg(
 	}
 	subjectFilters := gatherSubjectFilters(config.FilterSubject, config.FilterSubjects)
 
-	// Check subject filters overlap.
+	// Check subject filters do not overlap.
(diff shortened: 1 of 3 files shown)


wallyqs (approved):
LGTM

LautaroJayat:
Thanks @derekcollison and @wallyqs

PR #5296, opened 2024-04-10
Change js-server-only healthcheck to not check JS state

wallyqs:
Currently the js-server-only check which is often used for readiness probe in k8s (so used to determines whether a service should be available in the load balancer or a kubernetes DNS entry) also checks some of the state from JS so during meta leader, this means that during restarts it could make it appear that temporarily nodes are unhealthy.
This changes the js-server-only check to report once ports are ready instead.


derekcollison (approved):
LGTM

PR #5298, opened 2024-04-10
[IMPROVED] Filestore logging on exceptional conditions.

derekcollison:
Also make sure when trying to match the first msg we do not skip on exceptional errors, even though we do not expect these.


wallyqs (approved):
LGTM

PR #5301, opened 2024-04-11
[IMPROVED] Hold the lock for a msg block while calling sync.

derekcollison:
We detected issues with sync on certain storage subsystems. So will hold lock for msg block while calling sync.


neilalexander (approved):
LGTM

PR #5303, opened 2024-04-11
[IMPROVED[ Also hold the fs lock while sync call to index.db

derekcollison:


neilalexander (approved):
LGTM

PR #5305, opened 2024-04-11
[FIXED] Multiple deliveries of messages with delivery count going backwards.

derekcollison:
When we fail to deliver a message, we were not checking if this was a redelivery already and would decrement o.sseq, meaning we would pick up the same message after the next redelivery and would have a delivered count of 1.

This could lead to a message being delivered from the redelivery queue, but that could fail, then you would see same message twice, first with dc of 2, then 1.

Now app only gets one copy with delivery count of 2.


wallyqs (approved):
LGTM

Jarema (approved):
LGTM!

PR #5312, opened 2024-04-15
Fix-up consumer `sseq` in `checkNumPending` if out of bounds

neilalexander:
Otherwise we can repeatedly hit the filestore `NumPending` with the JS account lock held, which on a stream with high subject cardinality can be expensive. 

### DIFF
--- a/server/consumer.go
+++ b/server/consumer.go
@@ -4193,6 +4193,12 @@ func (o *consumer) checkNumPending() uint64 {
 		if o.sseq > state.LastSeq && o.npc != 0 || o.npc > int64(state.Msgs) {
 			// Re-calculate.
 			o.streamNumPending()
+			// streamNumPending would have corrected the o.npc condition but
+			// not o.sseq. Fix that here if so, otherwise we may keep calling
+			// through to this again and again.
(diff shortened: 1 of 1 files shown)


derekcollison (changes requested):
o.sseq should be set to last+1 since it's next we look at..

server/consumer.go
```diff
@@ -4193,6 +4193,12 @@ func (o *consumer) checkNumPending() uint64 {
 		if o.sseq > state.LastSeq && o.npc != 0 || o.npc > int64(state.Msgs) {
 			// Re-calculate.
 			o.streamNumPending()
+			// streamNumPending would have corrected the o.npc condition but
+			// not o.sseq. Fix that here if so, otherwise we may keep calling
+			// through to this again and again.
+			if o.sseq > state.LastSeq+1 {
```
derekcollison, line 4199: This should just be LastSeq.. Assignment should be +1.

derekcollison (changes requested):

derekcollison:
Let's close this with the data we had from yesterday about this being a bad stream state.

PR #5314, opened 2024-04-15
NRG: Don't step down for old terms in AEs

neilalexander:
We shouldn't step down if the append entry contains old terms, it may just be still in-flight from a distant node that is still behind. We also shouldn't clear the vote when leader unless we've definitely seen a newer term and not just the existing term, otherwise we could cast multiple votes.

### DIFF
--- a/server/raft.go
+++ b/server/raft.go
@@ -3121,9 +3121,13 @@ func (n *raft) processAppendEntry(ae *appendEntry, sub *subscription) {
 	if n.State() == Leader {
 		// If we are the same we should step down to break the tie.
 		if ae.term >= n.term {
-			n.term = ae.term
-			n.vote = noVote
-			n.writeTermVote()
+			// If the append entry term is newer than the current term, erase our
+			// vote.
+			if ae.term > n.term {
+				n.term = ae.term
+				n.vote = noVote
+				n.writeTermVote()
+			}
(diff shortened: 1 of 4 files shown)


server/raft.go
```diff
@@ -3121,9 +3121,13 @@ func (n *raft) processAppendEntry(ae *appendEntry, sub *subscription) {
 	if n.State() == Leader {
 		// If we are the same we should step down to break the tie.
 		if ae.term >= n.term {
-			n.term = ae.term
-			n.vote = noVote
-			n.writeTermVote()
+			// If it's the same term, don't clear our vote, otherwise we could
```
mprimi, line 3124: Comment references old behavior. Could be clear to just talk about new behavior, eg:
 > If term is higher than previously seen, clear vote

The implicit `else` case is also interesting here.
 - This node is leader
 - Some other node is sending an append entry with the same term

This should never ever happen. Each term can have at most one leader (if everyone is complying to the rules)

mprimi (approved):
LGTM

server/raft.go
neilalexander: Fixed the comment at least.

derekcollison (changes requested):
Again need tests. This types of systemic changes like this could have a lerge impact on real systems and we need to have a sense of what behavior we are correcting in a test form.

derekcollison (approved):
LGTM

PR #5315, opened 2024-04-15
NRG: Only reset election timeout on vote request if granting

neilalexander:
With the current behaviour, a node that is behind but running in the candidate state could cause nodes that are current to delay switching to the candidate state themselves because they keep resetting the election timer each time they receive a vote request.

Changing this to reset the election timeout only when granting means that we will allow other possibly more up-to-date nodes to transition to the candidate state faster.

This might explain why we've seen some unexpected delays in starting a new election after a leader has gone away.

### DIFF
--- a/server/raft.go
+++ b/server/raft.go
@@ -3884,7 +3884,6 @@ func (n *raft) processVoteRequest(vr *voteRequest) error {
 	}
 
 	n.Lock()
-	n.resetElectionTimeout()
 
 	vresp := &voteResponse{n.term, n.id, false}
 	defer n.debug("Sending a voteResponse %+v -> %q", vresp, vr.reply)
@@ -3916,6 +3915,7 @@ func (n *raft) processVoteRequest(vr *voteRequest) error {
 		n.term = vr.term
 		n.vote = vr.candidate
 		n.writeTermVote()
+		n.resetElectionTimeout()
 	} else {
 		if vr.term >= n.term && n.vote == noVote {
 			n.term = vr.term
--- a/server/raft.go
+++ b/server/raft.go
@@ -150,6 +150,7 @@ type raft struct {
 	pae     map[uint64]*appendEntry        // Pending append entries
 
 	elect  *time.Timer // Election timer, normally accessed via electTimer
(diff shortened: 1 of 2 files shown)


derekcollison (changes requested):
I think we would need a test to show the behavior change.
This will be complex I know, but changes at this foundational layer I think need tests.

server/raft_test.go
```diff
@@ -291,3 +291,52 @@ func TestNRGSimpleElection(t *testing.T) {
 		require_Equal(t, rn.vote, vr.candidate)
 	}
 }
+
+func TestNRGUnsuccessfulVoteRequestDoesntResetElectionTimer(t *testing.T) {
```
mprimi, line 295: Would be nice to also see a "timer resets when it should" on top of this test "timer does not reset when it shouldn't". 
But I know writing these RAFT tests is a giant pain and may not even be possible.

server/raft.go
```diff
@@ -150,6 +150,7 @@ type raft struct {
 	pae     map[uint64]*appendEntry        // Pending append entries
 
 	elect  *time.Timer // Election timer, normally accessed via electTimer
+	etlr   time.Time   // Election timer last reset time, for unit tests only
```
derekcollison, line 153: I prefer not to have things in production code that take up space that are only used for tests.

derekcollison (commented):
Is it not possible to write a test that observes bad behavior without the timer? Meaning by watching who becomes leader in the bad case and make sure the test looks that someone else was elected?

derekcollison (changes requested):
Is there no way to prove the behavior without adding eltr? Again not a fan of test only items in production code.

neilalexander:
I'm not in love with it either but we need some state that the test can observe and I don't know what the alternative is. We have no way to introspect what the reset time or deadline of the `n.elect` timer is either before or after the vote request because Go doesn't expose that information, nor are there any other side-effects from resetting the election timer that are visible in a healthy cluster. I also tried forcing a node into an unhealthy state and checking the time taken to switch to candidate but it ended up being flakey as the time precision wasn't good enough.

derekcollison (approved):
LGTM

PR #5316, opened 2024-04-15
Reuse `SublistResult` for `LoadNextMsgMulti`

neilalexander:
This PR adds `(*Sublist).MatchInline(..., *SublistResult)` so that we can reuse the same `SublistResult` multiple times when we know that we're going to be matching multiple times in rapid succession.

On the `JetStreamConsumeWithFilters/.*` benchmark, this reduces the number of `SublistResult` allocations in `LoadNextMsgMulti()` from 10GB down to 4GB:

Screenshot 2024-04-15 at 17 47 01


server/filestore.go
```diff
@@ -2273,7 +2274,7 @@ func (mb *msgBlock) firstMatchingMulti(sl *Sublist, start uint64, sm *StoreMsg)
 		}
 		expireOk := seq == lseq && mb.llseq == seq
 
-		if r := sl.Match(fsm.subj); len(r.psubs) > 0 {
+		if r := sl.MatchInline(fsm.subj, &result); len(r.psubs) > 0 {
```
derekcollison, line 2277: I think MatchWithResult would be better here as a name.

derekcollison (commented):
This is a good addition. 

Should we add it to the client hotpath in ths PR as well?

server/filestore.go
neilalexander: That's a much better name, updated.

neilalexander:
For now I haven't touched the client code as it's quite a bit more involved with the caching and probably deserves its own PR.

derekcollison (approved):
LGTM

neilalexander:
Good spot, for now I've raised #5324 just to stop us from taking `MatchWithResult()` entries and putting them in the cache but there might be more intelligent things we can do.

PR #5329, opened 2024-04-22
Implement `(*Sublist).HasInterest()`

neilalexander:
This may end up being more efficient than `MatchWithResult()` in some cases as we will stop processing on the first match and not spend time building up a `*SublistResult`. May want to give more thought on the caching here — right now it checks the cache, but `HasInterest()` won't populate it, only `Match()` will.

/cc @kozlovic 

### DIFF
--- a/server/filestore.go
+++ b/server/filestore.go
@@ -2257,7 +2257,6 @@ func (mb *msgBlock) firstMatchingMulti(sl *Sublist, start uint64, sm *StoreMsg) 
 		sm = new(StoreMsg)
 	}
 
-	var result SublistResult
 	for seq := start; seq <= lseq; seq++ {
 		llseq := mb.llseq
 		fsm, err := mb.cacheLookup(seq, sm)
@@ -2266,7 +2265,7 @@ func (mb *msgBlock) firstMatchingMulti(sl *Sublist, start uint64, sm *StoreMsg) 
 		}
 		expireOk := seq == lseq && mb.llseq == seq
 
-		if r := sl.MatchWithResult(fsm.subj, &result); len(r.psubs) > 0 {
+		if sl.HasInterest(fsm.subj) {
 			return fsm, expireOk, nil
 		}
 		// If we are here we did not match, so put the llseq back.
--- a/server/memstore.go
+++ b/server/memstore.go
@@ -1055,13 +1055,12 @@ func (ms *memStore) LoadNextMsgMulti(sl *Sublist, start uint64, smp *StoreMsg) (
 	// Initial setup.
 	fseq, lseq := start, ms.state.LastSeq
 
-	var result SublistResult
 	for nseq := fseq; nseq <= lseq; nseq++ {
 		sm, ok := ms.msgs[nseq]
 		if !ok {
 			continue
 		}
-		if r := sl.MatchWithResult(sm.subj, &result); len(r.psubs) > 0 {
+		if sl.HasInterest(sm.subj) {
 			if smp == nil {
 				smp = new(StoreMsg)
 			}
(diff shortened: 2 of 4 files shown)


server/filestore.go
```diff
@@ -2274,7 +2273,7 @@ func (mb *msgBlock) firstMatchingMulti(sl *Sublist, start uint64, sm *StoreMsg)
 		}
 		expireOk := seq == lseq && mb.llseq == seq
 
-		if r := sl.MatchWithResult(fsm.subj, &result); len(r.psubs) > 0 {
+		if sl.HasInterest(sm.subj) {
```
kozlovic, line 2276: I think you still need to distinguish between plain subs and queue subs. The original code here is just checking for plain sub interest (`len (r.psubs) > 0`), which means that if there was only queue interest, the "if" statement would have evaluate to false. You need to check with @derekcollison if it was an omission or not. If not, then you need to have booleans/int to indicate which interest type you want to check.

server/memstore.go
```diff
@@ -1055,13 +1055,12 @@ func (ms *memStore) LoadNextMsgMulti(sl *Sublist, start uint64, smp *StoreMsg) (
...
 	fseq, lseq := start, ms.state.LastSeq
 
-	var result SublistResult
 	for nseq := fseq; nseq <= lseq; nseq++ {
 		sm, ok := ms.msgs[nseq]
 		if !ok {
 			continue
 		}
-		if r := sl.MatchWithResult(sm.subj, &result); len(r.psubs) > 0 {
+		if sl.HasInterest(sm.subj) {
```
kozlovic, line 1063: Same.

server/sublist.go
```diff
@@ -621,6 +615,49 @@ func (s *Sublist) match(subject string, doLock bool, result *SublistResult) *Sub
...
 
+func (s *Sublist) hasInterest(subject string, doLock bool) bool {
+	// Check cache first.
+	if doLock {
+		s.RLock()
+	}
+	var matched bool
+	if s.cache != nil {
+		if r, ok := s.cache[subject]; ok {
+			matched = len(r.psubs)+len(r.qsubs) > 0
```
kozlovic, line 626: If there was no distinction of psubs/qsubs, the mere presence of an entry is enough, that is, the `ok` already means that there is interest. But I am of the opinion that we would need to check for the user check intention.

kozlovic (changes requested):
If there is no need to distinguish type of subs that we need to check (@derekcollison will comment if in the current use case it is important to know that it is plain subs), then LGTM except that you don't have to check the result's len when getting a hit match, otherwise, we need to add what type of sub we need to have to consider that there is interest.

server/filestore.go
derekcollison: I think any interest should short circuit. For where this would be used heavily its only plain subs. We could use this for all permissions checks as well.

server/filestore.go
```diff
@@ -2266,7 +2265,7 @@ func (mb *msgBlock) firstMatchingMulti(sl *Sublist, start uint64, sm *StoreMsg)
...
```
kozlovic, line 2268: @neilalexander Note that it used to be `fsm.subj` and you replaced with `sm.subj`. Is it intentional? Are they the same?

server/sublist.go
kozlovic: Since @derekcollison seem to say that we don't need to have a specialized HasInterest() call, then I would have replaced this all "if" statement with:
```
if s.cache != nil {
    _, matched = s.cache[subject]
}
```

kozlovic (changes requested):

server/sublist.go
neilalexander: Makes sense, done.

server/filestore.go
neilalexander: Good spot, have switched back to `fsm.subj`.

kozlovic (approved):
LGTM

derekcollison:
Let me know when no longer draft, etc.

PR #5330, opened 2024-04-22
NRG: Don't truncate WAL during catch-up until after peerstate/snapshot check

neilalexander:
Otherwise we might blow away a part of the log and still have to abandon the catchup anyway, which doesn't feel ideal.


derekcollison (approved):
LGTM

derekcollison:
If you unmark as draft we can merge.

PR #5332, opened 2024-04-22
Split JetStream cluster tests again

kozlovic:
Try to keep jetstream_cluster_3_test.go to just under 8 minutes and move the trail end to new jetstream_cluster_4_test.go.

Notice that this file contains a test that itself takes more than 6 minutes to run. I checked without `-race` and it takes about as much. The test is:
```
TestJetStreamClusterStreamOrphanMsgsAndReplicasDrifting
```
I understand it does a lot of things, but wonder if it could be made a bit faster? @wallyqs I believe you have worked on that one?


kozlovic:
@derekcollison I did what you asked, but the new "4" test file is already at 8+ minutes. This is because it contains the test aforementioned that takes 6+ minutes to run. If this test cannot be made way shorter, it is almost like it deserves its own matrix run, or in other words, we need already to start a "5" test file.

derekcollison (approved):
LGTM - Thanks Ivan

wallyqs (approved):
LGTM

kozlovic:
@derekcollison Just to be clear, we leave as-is and @wallyqs will see if it can make the test shorter, or do I - part of this PR - make the file "4" contain only the long running test and add a file "5" with some of the tail tests that used to be in "3"?

derekcollison:
Leave as is for now..

PR #5333, opened 2024-04-23
[IMPROVED] Do not write out meta state for filestore if it would block normal operations for too long.

derekcollison:
Since our meta state is technically optional, do not adversely effect normal operations if writing the state out would take too much time. We will force a write on a stop, e.g. server shutdown etc.

Also includes fix to recovery of partial last block to not force full recovery.


wallyqs (approved):
LGTM

PR #5338, opened 2024-04-24
[FIXED] Delivery counts on failed deliveries.

derekcollison:
When we failed to deliver a message, meaning the server knows it was not delivered, we did not properly send the message with deliver count of 1 when we redelivered.

Also fixed a bug where if a message was pulled from rdq but did not have a valid waiting request we would not redeliver until ackWait expired even though was never delivered to the client.

Updated test to make sure NumDelivered was always 1.

Big thanks to @piotrpio for finding the issue and writing first test to show the issue.


wallyqs (approved):
LGTM

PR #5339, opened 2024-04-24
[FIXED]  Stree Find() would fail if delete caused a new prefix that was too large.

derekcollison:
With fixed prefixes this would cause the tree to be broken and find to not work and match would work but with shortened subjects. Could not fix up with fixed prefix arrays when delete caused a non-leaf node to merge a larger prefix.

Changed to normal []byte slice for meta node prefixes.


wallyqs (approved):
LGTM

PR #5340, opened 2024-04-24
Add logging to measure writeFullState and enforceMsgPerSubjectLimit

wallyqs:
Adds some extra logging to writeFullState and enforceMsgPerSubjectLimit to detect when they take took long.

### DIFF
--- a/server/filestore.go
+++ b/server/filestore.go
@@ -545,6 +545,13 @@ func (fs *fileStore) unlockAllMsgBlocks() {
 }
 
 func (fs *fileStore) UpdateConfig(cfg *StreamConfig) error {
+	start := time.Now()
+	defer func() {
+		if took := time.Since(start); took > time.Minute {
+			fs.warn("UpdateConfig took %v", took.Round(time.Millisecond))
+		}
+	}()
+
 	if fs.isClosed() {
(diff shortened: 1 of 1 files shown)


derekcollison (approved):
LGTM

server/filestore.go
```diff
@@ -545,6 +545,13 @@ func (fs *fileStore) unlockAllMsgBlocks() {
 }
 
 func (fs *fileStore) UpdateConfig(cfg *StreamConfig) error {
+	start := time.Now()
+	defer func() {
+		if took := time.Since(start); took > time.Minute {
+			fs.warn("UpdateConfig took %v", took.Round(time.Millisecond))
```
ripienaar, line 551: We get asked multiple times a week when what the warnigns around time taken for internal subs mean and how it can be fixed.  Could we avoid such questions by improving these warnings?

wallyqs: We maybe need a page in the docs with some of these log lines and they mean, when these appear some of the side effects would be latency doing operations on the stream so would explain timeouts during the timespan reported.

derekcollison:
For these might also be good to report on number of subjects and interior deletes, the two major contributors for now.

PR #5341, opened 2024-04-24
Fix race condition in `TestNRGUnsuccessfulVoteRequestDoesntResetElectionTimer`

neilalexander:
Fixes a race that was picked up by the race detector.


derekcollison (approved):
LGTM

PR #5342, opened 2024-04-24
Deduplicate some `stree` functions

neilalexander:
The `isLeaf`, `base`, `setPrefix`, `numChildren`, `path` and `matchParts` functions all have the same implementation for each node type, so this cuts them down to a single copy.

The leaf type will continue to override these as normal with its own implementations.


derekcollison (approved):
LGTM

PR #5347, opened 2024-04-24
NRG: Tidy up `applyCommit`

neilalexander:
This PR doesn't change any logical behaviour, it just removes unnecessary `fpae` and `original` variables/steps.

### DIFF
--- a/server/raft.go
+++ b/server/raft.go
@@ -2684,14 +2684,10 @@ func (n *raft) applyCommit(index uint64) error {
 		n.debug("Ignoring apply commit for %d, already processed", index)
 		return nil
 	}
-	original := n.commit
-	n.commit = index
 
 	if n.State() == Leader {
 		delete(n.acks, index)
 	}
-
-	var fpae bool
 
 	ae := n.pae[index]
 	if ae == nil {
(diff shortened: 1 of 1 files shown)


derekcollison (approved):
LGTM

server/raft.go
```diff
@@ -2684,15 +2684,11 @@ func (n *raft) applyCommit(index uint64) error {
 		n.debug("Ignoring apply commit for %d, already processed", index)
 		return nil
 	}
-	original := n.commit
-	n.commit = index
```
kozlovic, line 2688: It may have been a bug, but to say that this does not change the behavior, I am not too sure about that. For instance, in original code, n.commit would be set to index and if `ae == nil` would have stayed with that value in several conditions where we return (line 2697 for instance) without restoring it to its original value. With this change, it will always be to the original value until after the `if ae == nil` test.

derekcollison: @neilalexander be good for you to double check on this and make sure we are happy with the change.

kozlovic: Ok. Like I said, it may have been an earlier bug where n.commit was incorrectly set.

PR #5353, opened 2024-04-25
Fixed `(*Sublist).HasInterest()`

kozlovic:
I made an incorrect comment previously about not needing to check the cache's result to know if there is a match. I added a test for that, but also discovered that there was an issue where `HasInterest()` would incorrectly return true if checking for `foo` when there is only a subscription on `foo.*`. Added more tests.

### DIFF
--- a/server/sublist.go
+++ b/server/sublist.go
@@ -621,7 +621,9 @@ func (s *Sublist) hasInterest(subject string, doLock bool) bool {
 	}
 	var matched bool
 	if s.cache != nil {
-		_, matched = s.cache[subject]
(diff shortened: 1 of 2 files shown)


neilalexander (approved):
Good catch, LGTM!

kozlovic:
I think it is same fix for `if n != nil` at the end of the loop. Adding more tests and fix.

kozlovic:
@neilalexander I have added the fix and more tests again.

neilalexander (approved):

derekcollison (approved):
LGTM - Thanks Ivan

PR #5356, opened 2024-04-25
[FIXED] Gateway: Outbound may fail to detect stale connection

### LINKED ISSUE #5349: Outbound gateway connections stuck waiting on first INFO response
### Observed behavior

I'm observing an issue with outbound gateway connections dropping during high cross-cluster packet loss, and remaining disconnected even after the network fully recovers. This leaves the super-cluster in a degraded state where messages aren't delivered across clusters because of the missing gateway connection.

Each server has explicitly-configured gateways with `reject_unknown_cluster` set in its config.

Inspecting the server logs, I see a `Creating outbound gateway connection to "[remote]"` entry but no corresponding `Outbound gateway connection to "[remote]" ([name]) registered` log entry afterwards.

Looking at the source code, it appears that for outbound gateway connections, the `PING` timers usually used to detect stale connections aren't initiated until _after_ the first `INFO` is received from the remote gateway connection. This means that if the initial `CONNECT` or the `INFO` response messages are lost, it does seem possible for the connection to get stuck indefinitely.

### Expected behavior

All explicitly-configured outbound gateway connections should continue retrying during periods of intermittent connectivity and eventually recover when network conditions resolve.

### Server and client version

We're running a somewhat old server version in production (2.7.4), but it appears this bug is still present in the latest commit as far as I can tell from inspecting the source code.

### Host environment

_No response_

### Steps to reproduce

_No response_

kozlovic:
If the first INFO never makes it back and for some reasons the remote also does not close the connection, or connection is half closed, the outbound would not have detected that the connection is stale.

The outbound connection cannot send PING before it receives the INFO from the remote because that is only when the outbound sends the CONNECT. But the remote will fail if the first protocol from an outbound is not CONNECT.

Using the same principle that we used for routes with compression enabled, where PINGs cannot be sent until a certain time in the handshake, we will use the ping timer to simply close the connection as stale if the ping timer has not been reset before the time needed to send the max number of pings.

Resolves #5349

### DIFF
--- a/server/gateway_test.go
+++ b/server/gateway_test.go
@@ -6490,15 +6490,15 @@ func TestGatewayAuthDiscovered(t *testing.T) {
 	waitForOutboundGateways(t, srvB, 1, time.Second)
 }
 
-func TestTLSGatewaysCertificateImplicitAllowPass(t *testing.T) {
-	testTLSGatewaysCertificateImplicitAllow(t, true)
+func TestGatewayTLSCertificateImplicitAllowPass(t *testing.T) {
+	testGatewayTLSCertificateImplicitAllow(t, true)
 }
 
-func TestTLSGatewaysCertificateImplicitAllowFail(t *testing.T) {
-	testTLSGatewaysCertificateImplicitAllow(t, false)
+func TestGatewayTLSCertificateImplicitAllowFail(t *testing.T) {
+	testGatewayTLSCertificateImplicitAllow(t, false)
 }
 
-func testTLSGatewaysCertificateImplicitAllow(t *testing.T, pass bool) {
+func testGatewayTLSCertificateImplicitAllow(t *testing.T, pass bool) {
 	// Base config for the servers
 	cfg := createTempFile(t, "cfg")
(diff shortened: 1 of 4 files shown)


server/gateway_test.go
```diff
@@ -7349,3 +7349,63 @@ func TestOCSPGatewayMissingPeerStapleIssue(t *testing.T) {
 	waitForOutboundGateways(t, srvC, 2, 5*time.Second)
 	wg.Wait()
 }
+
+func TestGatewayOutboundDetectsStaleConnectionIfNoInfo(t *testing.T) {
```
wallyqs, line 7353: we've had some reports of idle websocket connections not pruned by the ping interval but peer actually disconnected already, so I wonder this could be a similar symptom or related for those type of connections

wallyqs: I think it might be different issue then, just that in websocket connections it is possible for a client connection to look connected and ping interval not happening so it would still show in connz but tcp connection is closed already.

derekcollison (approved):
LGTM

PR #5358, opened 2024-04-26
Protect delivery count from becoming negative

wallyqs:


yordis (approved):

derekcollison (commented):
This breaks the fix I just did. Will explain offline @wallyqs

PR #5362, opened 2024-04-27
[FIXED] Stream catchup would not sync after server crash and restart.

derekcollison:
We had a bug that would overwrite the sync subject during parallel stream creation which would cause upper layer stream catchups to fail on server crash and subsequent restarts.
    
We also were reporting first sequence mismatch when we hit max retries to force a reset but this was misleading, so added in proper error for max retires limit.

Also on some extreme kill and restart cases our internal checks were being called before complete state was achieved, so delay the initial check and added in periodic checks to ensure replica consistency.


wallyqs (approved):
LGTM

PR #5363, opened 2024-04-28
[IMPROVED] Do not hold filestore lock on msg block loads for firstSeqForSubj.

derekcollison:
In cases where max per subject is utilized, once that condition is hit storing a new message involves writing the new one, flushing it, then determining the current first seq for the subject so we can remove it. Determining the first sequence for the subject could involve walking multiple blocks and loading them in as well, all while the filestore lock is being held causing delays for other access to the filestore, even for things like FastState().


Jarema (approved):
LGTM

PR #5366, opened 2024-04-29
Improve stream sourcing consumers stall handling

jnmoyne:
- Resets the stalling detection timestamp upon re-creating the consumer
- Adjust the idle heart beat for source/mirror consumers from 1s to 5s


derekcollison (approved):
LGTM

PR #5370, opened 2024-05-01
[IMPROVED] Put a maximum on the idle flush time for the mb.fss.

derekcollison:
For certain streams that have a large subject space and max msgs per subject with moderate to high ingress rates and very large subjects, holding onto a message blocks can cause undue memory pressure and GC issues.


wallyqs (approved):
LGTM

PR #5371, opened 2024-05-01
[IMPROVED] Up the compression threshold

derekcollison:
We were seeing degraded performance and possible memory bloat on small messages being compressed for replication in NRGs. This updates the threshold to 8k vs 256.


wallyqs (approved):
LGTM

PR #5372, opened 2024-05-01
[IMPROVED] Handling of sourcing into a WQ/Interest streamwith a limit and discard new

jnmoyne:
Small improvement to the handling of sourcing into a WQ/Interest stream with a limit and discard new.
 
 Avoids re-scanning the stream to find the consumer's starting sequence number for that source every time. Do not log a warning every time this happens as it becomes constant log spam for something that happens temporarily all the time.

Strengthen test TestJetStreamClusterSourceWorkingQueueWithLimit to also check the payload (to catch for example the same message being sourced more than once, or messages not getting sourced).

### DIFF
--- a/server/jetstream_cluster_4_test.go
+++ b/server/jetstream_cluster_4_test.go
@@ -24,6 +24,7 @@ import (
 	"math/rand"
 	"os"
 	"path/filepath"
+	"strconv"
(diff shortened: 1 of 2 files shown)


derekcollison (approved):
LGTM

PR #5374, opened 2024-05-01
Add a exprvar monitoring endpoint

ripienaar:

### DIFF
--- a/server/events_test.go
+++ b/server/events_test.go
@@ -1674,8 +1674,8 @@ func TestSystemAccountWithGateways(t *testing.T) {
 	nca.Flush()
 
 	// If this tests fails with wrong number after 10 seconds we may have
-	// added a new inititial subscription for the eventing system.
-	checkExpectedSubs(t, 56, sa)
+	// added a new initial subscription for the eventing system.
+	checkExpectedSubs(t, 58, sa)
 
 	// Create a client on B and see if we receive the event
 	urlb := fmt.Sprintf("nats://%s:%d", ob.Host, ob.Port)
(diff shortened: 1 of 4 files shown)


server/events_test.go
```diff
@@ -2990,6 +2990,7 @@ func TestServerEventsPingMonitorz(t *testing.T) {
 		{"HEALTHZ", nil, &JSzOptions{}, []string{"status"}},
 		{"HEALTHZ", &HealthzOptions{JSEnabledOnly: true}, &JSzOptions{}, []string{"status"}},
 		{"HEALTHZ", &HealthzOptions{JSServerOnly: true}, &JSzOptions{}, []string{"status"}},
+		{"EXPVARZ", nil, &ExpvarzStatus{}, []string{"memstats", "cmdline"}},
```
derekcollison, line 2993: Should that be GC stats?

derekcollison: Meaning the label?

server/monitor.go
```diff
@@ -2231,6 +2232,11 @@ func (s *Server) Leafz(opts *LeafzOptions) (*Leafz, error) {
 			leafnodes = append(leafnodes, lni)
 		}
 	}
+
+	if len(leafnodes) == 0 {
```
derekcollison, line 2236: Why do we need this?

server/events_test.go
ripienaar: Could be but it’s a good hint about what this format is for people who know go

server/monitor.go
ripienaar: Because it was putting nulls into the json data and the test was poor and didn’t detect this. 

Checking for nil would probably be better but from this code it’s equivalent

derekcollison: Could we do omitempty?

ripienaar: Yes but since the tests specifically checked this key is present I figured just correcting the data type would be safest and smallest impact.

derekcollison: How did test work before?

ripienaar: It only checked if the key was in the result. 

I made a mistake at first and one of mine was also null but test was green after I added mine to the cases. 

Checked the test and noticed it was a bit weak so I made the test check for nil which then found mine and this one. 

So test was improved which found it.

derekcollison, line 2236: Probably should be `if leafnodes == nil`

ripienaar: nods - I moved the make out of the if though so now there's no need to check it, its always right this is better.

derekcollison: Thanks..

derekcollison (approved):
LGTM

PR #5378, opened 2024-05-02
[IMPROVED] Simplify writing the full state to index.db and protect with dios.

derekcollison:


wallyqs (approved):
LGTM

PR #5379, opened 2024-05-02
[IMPROVED] Raise the timeout for consumer creation requests for sources and mirrors.

derekcollison:
More often we are seeing filtered sources and mirrors, and on server restart they could take time to calculate num pending.


wallyqs (approved):
LGTM

PR #5380, opened 2024-05-02
[IMPROVED] Added in separate last subject timestamp to track access and use of mb.fss

derekcollison:
We were expiring the fss through forced expire and we were not properly tracking the actual access activity. When we call ensurePerSubjectInfoLoaded() it will mark activity. If we load manually to check for forced expires we mark manually.


wallyqs (approved):
LGTM

PR #5389, opened 2024-05-06
[IMPROVED] Streams ingest performance with many stream sources.

derekcollison:
We were struggling with inbound performance from a large number of streams, and were not properly getting all messages when our stream had a leader change.

The general issue with performance was that we had a go routine per source pulling from individual internal message queues. If we were behind and catching up, or if all sources were receiving messages in a quick fashion, we would be very slow. This was due to us lighting up all the go routines and hence kernel threads only to contend on locks and then IO. So were were thrashing.

We now mux all sources into a single processing go routine with a shared internal message queue.

During testing there were other optimizations that were realized as well.

@jnmoyne also did some restructuring of the sourcing code to make it more understandable and easier to approach and improve.

Thanks for @jnmoyne and @wallyqs for the testing!

### DIFF
--- a/server/stream.go
+++ b/server/stream.go
@@ -255,7 +255,7 @@ type stream struct {
 
 	// Sources
 	sources              map[string]*sourceInfo
-	sourceRetries        map[string]*time.Timer
+	sourceSetupSchedules map[string]*time.Timer
 	sourcesConsumerSetup *time.Timer
 
 	// Indicates we have direct consumers.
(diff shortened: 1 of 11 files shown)


server/stream.go
```diff
@@ -3061,32 +3059,10 @@ func (mset *stream) setupSourceConsumer(iname string, seq uint64, startTime time
...
-							"consumer": si.cname,
-						},
-					) {
-						ready.Done()
-					}
 				}
 			}
 			mset.mu.Unlock()
-			ready.Wait()
+			//ready.Wait()
```
neilalexander, line 3065: Leftover comment

neilalexander (approved):
LGTM

wallyqs (approved):
LGTM

server/stream.go
derekcollison: Thanks will fix.

PR #4255, opened 2023-06-20
WebSocket-specific `INFO`

neilalexander:
This fixes #4252 by ensuring that `tls_available`, `tls_required`, `host` and `port` are populated based on the WebSocket listener rather than standard listeners.

Signed-off-by: Neil Twigg <neil@nats.io>

Issue #4252, opened 2023-06-19
No `tls_required` in INFO for TLS-only WebSocket connections


Envek:
It is a pretty minor counter-intuitive behavior that bugged me while I was working on adding websocket support to the nats-pure.rb client.

It seems that `INFO` doesn't really account on how client is connected as it is always return info about standard protocol.

#### Versions of `nats-server` and affected client libraries used:

nats-server 2.9.18

#### Steps or code to reproduce the issue:

 1. Set up NATS server that accepts only plaintext connections over standard protocol, but accepts only secure connections over WebSocket protocol.

    ```
    # nats.conf
    net:  "127.0.0.1"
    port: 4222
    websocket {
      port: 8443
      tls {
        cert_file:  "./spec/configs/certs/server.pem"
        key_file:   "./spec/configs/certs/key.pem"
      }
    }
    ```

    E.g. demo.nats.io seems to be configured like this.

 2. Connect over websocket to this server, e.g. using [`websocat`](https://github.com/vi/websocat)

    ```
    $ websocat --binary --insecure wss://127.0.0.1:8443
    INFO {"server_id":"…","server_name":"…","version":"2.9.16","proto":1,"git_commit":"f84ca24","go":"go1.19.8","host":"127.0.0.1","port":4222,"headers":true,"max_payload":1048576,"client_id":6,"client_ip":"127.0.0.1"}
    ```

#### Expected result:

`INFO` contains `tls_required` equals to `true`.

#### Actual result:

`INFO` doesn't contain `tls_required`.

PR #4122, opened 2023-05-02
[ADDED] Custom Muxer for websocket httpserver

jordan-rash:
Closes #3994 

 - [X] Link to issue, e.g. `Resolves #3994`
 - https://github.com/nats-io/nats-server/issues/3994
 - [x] Documentation added (if applicable)
 - [x] Tests added
 - [X] Branch rebased on top of current main (`git pull --rebase origin main`)
 - [ ] Changes squashed to a single commit (described [here](http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html))
 - [x] Build is green in Travis CI
 - [X] You have certified that the contribution is your original work and that you license the work to the project under the [Apache 2 license](https://github.com/nats-io/nats-server/blob/main/LICENSE)

### Changes proposed in this pull request

This PR adds the ability to provide the HTTP Server powering the Websocket with a custom muxer.  The reason for this change is that it allows people to add a NATs powered websocket to an existing webapp without having multiple ports.  Sample code below

### Example usage

```go
package main

import (
	"fmt"
	"net/http"

	"github.com/nats-io/nats-server/v2/server"
)

func main() {
	mux := http.NewServeMux()

	mux.HandleFunc("/", func(w http.ResponseWriter, r *http.Request) {
		w.WriteHeader(200)
		_, _ = w.Write([]byte("hello"))
	})

	wsOpts := server.WebsocketOpts{
		Port:     8000,
		Muxer:    mux,
		Endpoint: "/ws",
		NoTLS:    true,
	}

	s, err := server.NewServer(
		&server.Options{
			Port:      0,
			Websocket: wsOpts,
			Debug:     true,
		},
	)
	if err != nil {
		server.PrintAndDie(fmt.Sprintf("%s: %s", "nats-server", err))
	}

	s.ConfigureLogger()

	if err := server.Run(s); err != nil {
		server.PrintAndDie(err.Error())
	}

	s.WaitForShutdown()
}

```

#### Custom Endpoint usage
```bash
nats-server on  feat/websocket-mux via 🐹 v1.20.2 
❯ http localhost:8000   
HTTP/1.1 200 OK
Content-Length: 5
Content-Type: text/plain; charset=utf-8
Date: Tue, 02 May 2023 15:40:08 GMT

hello
```

#### Websocket usage
```bash
nats-server on  feat/websocket-mux via 🐹 v1.20.2 
❯ websocat ws://localhost:8000/ws
[...]

PR #5392, opened 2024-05-06
Exposing a WebSocket Handler for embeded NATS servers

BlizzTom:
When embedding a NATS server with a UI, in order to take advantage of the Web Socket connection on the same server, it was required that you made a reverse proxy to the web socket port.

This exposes a small handler without MQTT/LEAF node support for embedding scenarios.

Signed-off-by: Tom Anderson <tanderson@blizzard.com>

### DIFF
--- a/server/websocket.go
+++ b/server/websocket.go
@@ -1099,6 +1099,29 @@ func (s *Server) wsConfigAuth(opts *WebsocketOpts) {
 	ws.authOverride = opts.Username != _EMPTY_ || opts.Token != _EMPTY_ || opts.NoAuthUser != _EMPTY_
 }
 
+// HandleWebSocket allows for an embedded server to expose a NATS websocket Handler
+// preventing the need to reverse proxy the NATS connection back into a separate
+// port on the same process. By using this handler, the embedded server is responsible
+// for handling the tls configurations as well as the requested origins, as the internal
+// code in used by this handler will allow all origins.
+// This endpoint does come with some limitations, such as not supporting MQTT or Leaf Nodes,
+// as well as if the HTTP handler is not on the root path (/), most NATS clients will not be
+// able to connect to the endpoint. The intention of this handler is for in browser
+// websocket connections.
+func (s *Server) HandleWebSocket(w http.ResponseWriter, r *http.Request) {
+	res, err := s.wsUpgrade(w, r)
+	if err != nil {
+		s.Errorf(err.Error())
+		return
+	}
+	switch res.kind {
+	case CLIENT:
+		s.createWSClient(res.conn, res.ws)
+	default:
+		// all other cases are not supported on the direct WebSocket endpoint
+	}
+}
+
 func (s *Server) startWebsocketServer() {
 	if s.isShuttingDown() {
 		return


kozlovic (commented):
@BlizzTom How would you use it? It looks to me that there is already a PR that would maybe cover this but on a more complete? Can you have a look at https://github.com/nats-io/nats-server/pull/4122 and see if that one would work? If so, maybe we need to revive it and go from there?

jordan-rash:
> @BlizzTom How would you use it? It looks to me that there is already a PR that would maybe cover this but on a more complete? Can you have a look at https://github.com/nats-io/nats-server/pull/4122 and see if that one would work? If so, maybe we need to revive it and go from there?

Happy to get #4122 rebased here in the next few hours so @BlizzTom can see if it satisfies his requirements

jordan-rash:
#4122 has been rebased.  Looked forward to any feedback!

BlizzTom:
The only issues I see of that other PR is that the http server is still controlled by the NATS server instance, rather than my own, and I am forced to use the `http.ServeMux` instead of any other `http.Handler`.

I could be talked into being ok with that, it's more of a limitation of that PR, and I didn't even realize it existed.

A use case that I have is that my application has pretty strict security access through a single port (443), and NATS sits behind that network boundary. The application uses NATS embedded to cluster some logic between nodes. The website uses websockets to communicate with the application API. Currently, I have to reverse proxy the websocket port back into `localhost:<port>` on the same process to expose that websocket.

While the work around hasn't been terrible, it's an extra step of complexity and a network hop. With the web server mux being quite complicated and not using `http.ServeMux` as the mux, the other PR would not suit the needs.

BlizzTom:
I should probably clarify the statements above about the "complexity" of the application.

There are a few things we set on the `http.Server` instance, and how we customize our listener:

* `server.BaseContext`: this is set across the entire space so that every request context is rooted in the application context
* `server.ReadTimeout`: configured outside of default values (can pass into NATS websocket options as `HandshakeTimeout`)
* `server.ReadHeaderTimeout`: configured outside of default values
* `server.WriteTimeout`: configured outside of default values
* `server.IdleTimeout`: configured outside of default values
* `server.MaxHeaderBytes`: configured outside of default values
* `server.Shutdown`: use configurable timeout to finish up any pending requests after closing listener
* `net.Listener`: tls configuration is integrated with an ACME solution to provide latest up to date certificates on every connect
* `net.Listener`: support for PROXY protocols for being hosted behind an L4 load balancer for H2 compatibility
* `net.Listener`: configured tcp keep alive and keep alive period with configurable timeouts

The mux is really the least of the issues, as we could just route a specific path in our mux to use the `http.ServeMux` `ServeHTTP` function.

kozlovic:
@BlizzTom Ok, I think you have a compelling reason why you would not be able to use the other PR and need your own http server and handler. I have few comments/questions:
- `http.Server` needs a `Handler` which must implement the `ServerHTTP` interface, so I assume that the new function you propose would be invoked from your own handler?
- I understand that you would not even configure the NATS Server websocket options, which means that the websocket internal structure will not be "initialized", which means that your NATS websocket clients will not benefit from the server features such as checking origin (that you document in the function doc), but also receiving INFO protocols with information regarding the addresses where clients could reconnect should the connection be disconnected. I guess this is fine since in your context, but something to keep in mind.
- Because of the point above, the INFO protocol received by the server will have possibly erroneous information regarding TLS since the NATS Server itself will not have websocket's TLS configuration specified while your http server may. It was reported in https://github.com/nats-io/nats-server/issues/4252 and addressed in https://github.com/nats-io/nats-server/pull/4255 (which I was planning on reworking by lifting that to the initialization of the NATS websocket server since these info - as far as I know - can't be config reloaded and will be immutable).
- I would have liked to see a test added to demonstrate how this would be used and have a bit of code coverage for this new function.

BlizzTom:
Thanks for that feedback @kozlovic, I will work on a test case for it, as well as seeing if I can figure out the `INFO` point brought up , as that could potentially be an issue, since in my use case, the `host` and `port` would always be the same. 

I would need to check the `nats.js` implementation to see if it actually uses the `INFO` block or not for reconnects. From my own experiences, it hasn't been using it, and just reconnects to the original endpoint.

On the testing front, the intended usage is from the `nats.js` package, and I am not really sure how to invoke that properly from the go tests, do you have any recommendations on that?

kozlovic:
@BlizzTom The `nats.js` uses the `INFO` protocol to get the list of possible "connect URLs", which is a list of URLs that a client can reconnect to (augmented to the list of NATS Server URLs in a given cluster). In your approach, this list would not be specified, so the client library would indeed always attempt to reconnect to the initial given URL, which again, in your context makes sense.

As for a test, I wanted to at least see something like starting a NATS Server and create an http.Server and run it, passing the new function you have added, then a simple test that checks that a NATS client can connect:
```
// port will be the port of the net listener that you have created and passed to http.Server
nc := natsConnect(t, fmt.Sprintf("ws://127.0.0.1:%d", port))
defer nc.Close()
```
If the connection succeeds, and knowing that the port is not a NATS Server port (neither normal client nor websocket), it will good enough for a basic test.

Otherwise, using the low level test helpers (as in TestWSTLSVerifyAndMap) would be able to inspect an incoming INFO protocol. But there is no need since you won't have a way to influence this with the new function (HandleWebsocket) you want to add. But yes, say that if in the test the NATS Server does not have websocket{} defined and your http.Server uses TLS, and you connect the NATS client with `wss://` instead of `ws://`, it would work, but the INFO would likely not show that TLS is available/required/etc.. as described in the previous comment.

Issue #5387, opened 2024-05-05
user JWT Src limits can be used for in-process connection


chezgi:
### Proposed change

in user JWT we can limit src of connection to a list of CIDR addresses. like: 127.0.0.1
but when we want to use an user JWT only for pipe in-process connection, it doesn't works.

### Use case

when using in-process connections and we want to use user credentials for it,
i can be secured to don't allow this credential can be used for out of process connections.
something like `pipe` in CIDR list gives error:
```
User JWT no longer valid: &{Issues:[invalid cidr "pipe" in user src limits]}
```

chezgi:
or maybe we can add another connection type: `Pipe`
and we can limit connection type of user to it

kozlovic:
@chezgi Adding a new connection type may not be enough because the server would still do network source validation and fail (unless we change that code anyway).

Since one can technically run a server from a process but that server be configured to have listen ports opened and still make InProcess connections (see https://github.com/nats-io/nats-server/blob/a0fac40cd839a4e8903bcf2f347dcb647e1a47db/server/server.go#L2698), if one wants to use a JWT that contains CIDR list, I see only 2 options:

- disable source validation if we detect that the client is in-process
- do source validation but add support for "inprocess" (or "pipe") in the JWT library's CIDR list

One could argue that the JWT configuration/limits should be respected and therefore, adding the "inprocess" CIDR would make more sense? @derekcollison what do you think?

@chezgi By the way, how are you running the server? Are you setting the `DontListen` option or somehow connect in-process clients without the `DontListen` option?

kozlovic:
@chezgi Or are you saying that you would then add a "PIPE" connection type to that JWT but then remove the CIDR? In other words, you want to use a JWT but want to make sure that only an in-process client can use that JWT?

If you were to configure the server with the `DontListen` option, you would necessarily limit clients that are in-process (since they would not have any way to connect to this server) and that proposal would be moot. Any reason why you are not running this way?

chezgi:
@kozlovic 
my usage: i want to use my authorization callout user to be in process.
therefore my jwt must limit this user for only in process connection, 
or  127.0.0.1 must interpreted to `in process` connection type .

server is servicing others from network, therefore `DontListen` is disabled.
and this user can't have source connection limit. and it is very insecure in this situation.

kozlovic:
[...]

PR #5395, opened 2024-05-07
[ADDED] Allowed connection type "in process"

kozlovic:
With the JWT library update, there is now a `jwt.ConnectionTypeInProcess` connection type that can be added to the list of allowed connection types when defining an user. If the list is not empty and does not have this type, an "in process" connection will not be allowed.

Conversely, it is now possible to limit a user to be used only for in process connections (or several types based on the content of the list).

Possible solution for #5387

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -6,7 +6,7 @@ require (
 	github.com/google/go-tpm v0.9.0
 	github.com/klauspost/compress v1.17.8
 	github.com/minio/highwayhash v1.0.2
-	github.com/nats-io/jwt/v2 v2.5.6
+	github.com/nats-io/jwt/v2 v2.5.7-0.20240507172914-6b33489b89ba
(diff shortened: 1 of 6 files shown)


kozlovic:
@derekcollison This would be a possible solution to issue #5387. If you think this is reasonable, I will make the JWT PR, have it merged, and then update this PR with the new JWT lib dep and open this for formal review.

derekcollison:
Will take a look.

derekcollison (commented):
LGTM - Let's proceed.

kozlovic:
@derekcollison I have updated the PR with the go.mod change to point to newly released JWT library that adds the "in process" connection type and updated the test to cover JWT test. It is ready for review now.

derekcollison (approved):
LGTM

PR #5399, opened 2024-05-08
[IMPROVED] Improve mux'd routes with large subject space.

derekcollison:
Moved account expired check to atomic.
Slightly increased cache size and floor threshold for route acc result cache. Make sure to prune first, then add new entry to avoid immediate recycling. Also moved account lock to read lock for selectMappedSubject.


wallyqs (approved):
LGTM

PR #5401, opened 2024-05-09
[IMPROVED] LoadNextMsg performance.

derekcollison:
1. If we think out filter matches our stream do linear since first should match.
2. If literal, make sure to look up directly from fss and uses ss.First as fseq, bypass other logic.
3. Only do fss scan if len(fss) is 25% of scan range.


neilalexander (approved):
LGTM

PR #5405, opened 2024-05-09
[IMPROVED] Websocket: generating INFO to send to clients

kozlovic:
PR #4255 added code in generateClientInfoJSON to set the proper info Host/Port/TLSAvailable/TLSRequired fields to send to clients.

However, this was requiring a lock but more importantly was computing the listener's host/port everytime, which is not necessary since this is immutable because we don't support the change during a config reload.

Also, the TLSRequired field was set based on the server TLSConfig's InsecureSkipVerify value, which is irrelevant for a server. The mere presence of a TLSConfig (c.srv.websocket.tls being true) is enough.

I have modified the TestWSReloadTLSConfig test to verify that the tls block cannot be removed and no_tls set to true, which means that tls value can't change. I also added check for the info's Host/Port/TLSAvailable/TLSRequired values.

### DIFF
--- a/server/client.go
+++ b/server/client.go
@@ -2383,24 +2383,11 @@ func (c *client) generateClientInfoJSON(info Info) []byte {
 	info.MaxPayload = c.mpay
 	if c.isWebsocket() {
 		info.ClientConnectURLs = info.WSConnectURLs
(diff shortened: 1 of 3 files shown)


derekcollison (approved):
LGTM

Issue #5396, opened 2024-05-07
Comment at end of config file is a parse error


philpennock:
### Observed behavior

With a `nats.conf` file which has a `#` line-comment at the end of the file, 

Error from `nats-server -t`:
```
nats-server: Parse error on line 153: 'Expected a top-level value to end with a new line, comment or EOF, but got 'T' instead.'
```

The comment was:

```
# This is a dummy comment added to force a config deploy and server restart.
# $RANDOM number: 23566
```

and it is the `T` of `This` which is being reported.

### Expected behavior

A line-comment should be valid on any line, even after the last actual directive

### Server and client version

Server: v2.10.15-RC.6

derekcollison:
Need more info, just test this and works fine.

```go
// https://github.com/nats-io/nats-server/issues/5396
func TestServerConfigLastLineComments(t *testing.T) {
	conf := createConfFile(t, []byte(`
		listen: 127.0.0.1:-1
		server_name: s-test

		# This is a dummy comment added to force a config deploy and server restart.
		# $RANDOM number: 23566
	`))

	s, _ := RunServerWithConfig(conf)
	defer s.Shutdown()

	// This should work of course.
	nc, err := nats.Connect(s.ClientURL(), nats.UserInfo("u", "pass"))
	require_NoError(t, err)
	defer nc.Close()
}
```

Is RANDOM defined?

derekcollison:
And I tried both norma string and a quoted number..

kozlovic:
Since it works for @derekcollison, @philpennock, could it be that you have an error prior to that line? Something like you started the definition of an object but don't have the closing `]` or `}`, something like that?

philpennock:
Additional pre-condition for failure: using braces around a bunch of config.


### Valid

```
{
  "listen":  "0.0.0.0:4222"
}
```

```
{
# wibble
  "listen":  "0.0.0.0:4222"
}
```

```
# wibble
{
  "listen":  "0.0.0.0:4222"
}
```

### Invalid

```
{
  "listen":  "0.0.0.0:4222"
}
# wibble
```

derekcollison:
Thanks will adapt the test.

PR #5406, opened 2024-05-10
[FIXED] Fix to properly deal with block scopes in lexer.

derekcollison:
We had a bug that was reported that was not handling a block scope where the terminating '}' was on a new line.

Resolves: #5396 

### DIFF
--- a/conf/parse_test.go
+++ b/conf/parse_test.go
@@ -811,6 +811,8 @@ func TestBlocks(t *testing.T) {
 				}
 			} else if !strings.Contains(err.Error(), test.err) || !strings.Contains(err.Error(), test.linepos) {
 				t.Errorf("expected invalid conf error, got: %v", err)
+			} else if err != nil {
+				t.Error(err)
 			}
 		})
 	}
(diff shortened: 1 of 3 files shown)


wallyqs (approved):
LGTM

conf/parse_test.go
```diff
@@ -740,3 +740,78 @@ func TestJSONParseCompat(t *testing.T) {
...
+			if err := os.WriteFile(f.Name(), []byte(test.input), 066); err != nil {
+				t.Error(err)
+			}
+			if m, err := ParseFile(f.Name()); err == nil {
+				if !reflect.DeepEqual(m, test.expected) {
+					t.Fatalf("Not Equal:\nReceived: '%+v'\nExpected: '%+v'\n", m, test.expected)
+				}
+			} else if !strings.Contains(err.Error(), test.err) || !strings.Contains(err.Error(), test.linepos) {
+				t.Errorf("expected invalid conf error, got: %v", err)
+			}
```
wallyqs, line 814: Maybe should add here an extra `else` here to check in case there was a unexpected error:
```suggestion
			} else if err != nil {
				t.Error(err)
			}
```

PR #5420, opened 2024-05-15
[FIXED] Fix for data races in stream.

derekcollison:
Most of these were introduced by consideration of an atomic for leader status. This approach was abandoned.


wallyqs (approved):
LGTM

PR #5424, opened 2024-05-15
[FIXED] LeafNode: connection may fail on slow link

### LINKED ISSUE #5417: Connection between Leafnode and Core NATS over satellite link fails to get established
### Observed behavior

Siden is running into an odd issue with NATS Leafnode not being able to connect to our Core NATS server.
After TCP 3-way handshake Core NATS sends INFO message and then NATS Leafnode responds with INFO and 400+ms later sends closes connection (sends TCP FIN).
The latency of connection is 500ms.

Here are packets exchanged. Packets with P flag are INFO messages, followed by FIN packet
```
22:00:32.440826 IP 10.99.0.99.34786 > 3.136.10.169.7422: Flags [S], seq 1607668652, win 65535, options [mss 1460,sackOK,TS val 2932985413 ecr 0,nop,wscale 8], length 0
22:00:32.980351 IP 3.136.10.169.7422 > 10.99.0.99.34786: Flags [S.], seq 2383013068, ack 1607668653, win 65535, options [mss 1460,sackOK,TS val 1779976004 ecr 2932985413,nop,wscale 8], length 0
22:00:32.980410 IP 10.99.0.99.34786 > 3.136.10.169.7422: Flags [.], ack 1, win 256, options [nop,nop,TS val 2932985953 ecr 1779976004], length 0
22:00:33.520223 IP 3.136.10.169.7422 > 10.99.0.99.34786: Flags [P.], seq 1:468, ack 1, win 256, options [nop,nop,TS val 1779976544 ecr 2932985953], length 467
22:00:33.520276 IP 10.99.0.99.34786 > 3.136.10.169.7422: Flags [.], ack 468, win 261, options [nop,nop,TS val 2932986493 ecr 1779976544], length 0
22:00:33.520719 IP 10.99.0.99.34786 > 3.136.10.169.7422: Flags [P.], seq 1:164, ack 468, win 261, options [nop,nop,TS val 2932986493 ecr 1779976544], length 163
22:00:33.981008 IP 10.99.0.99.34786 > 3.136.10.169.7422: Flags [F.], seq 164, ack 468, win 261, options [nop,nop,TS val 2932986954 ecr 1779976544], length 0
```
in the console the NATS Leafnode outputs the following:
[7] [INF] 3.13.54.170:7422 - lid:53 - Leafnode connection closed: Read Error - Account: $G
[7] [INF] 3.13.54.170:7422 - lid:54 - Leafnode connection created for account: $G
[7] [INF] 3.13.54.170:7422 - lid:54 - Leafnode connection closed: Read Error - Account: $G

After investigation the issue is caused by too short timeout DEFAULT_LEAFNODE_INFO_WAIT, which is hardcoded to 1s.
Unfortunately it's a constant and not configurable,
We changed the timeout value in the vendor directory to 3s and that resolved the issue.

Please make this parameter configurable.

### Expected behavior

NATS connection can be established successfully

### Server and client version

v2.10.10

### Host environment

_No response_

### Steps to reproduce

Test NATS Leafnode - Server connection over high latency link.

kozlovic:
Added the leafnode remote configuration parameter `first_info_timeout`
which is the amount of time that a server creating a leafnode
connection will wait for the initial INFO from the remote server.

Resolves #5417

### DIFF
--- a/server/config_check_test.go
+++ b/server/config_check_test.go
@@ -2009,6 +2009,23 @@ func TestConfigCheck(t *testing.T) {
 			errorPos:  9,
 		},
 		{
+			name: "invalid duration for remote leafnode first info timeout",
+			config: `
+				leafnodes {
+					port: -1
+					remotes [
+						{
+							url: "nats://127.0.0.1:123"
+							first_info_timeout: abc
+						}
+					]
+				}
+			`,
+			err:       fmt.Errorf("error parsing first_info_timeout: time: invalid duration %q", "abc"),
+			errorLine: 7,
+			errorPos:  8,
+		},
+		{
 			name:       "show warnings on empty configs without values",
 			config:     ``,
 			warningErr: errors.New(`config has no values or is empty`),
--- a/server/leafnode.go
+++ b/server/leafnode.go
@@ -982,6 +982,7 @@ func (s *Server) createLeafNode(conn net.Conn, rURL *url.URL, remote *leafNodeCf
 	c.Noticef("Leafnode connection created%s %s", remoteSuffix, c.opts.Name)
 
 	var tlsFirst bool
+	var infoTimeout time.Duration
 	if remote != nil {
 		solicited = true
 		remote.Lock()
@@ -991,6 +992,7 @@ func (s *Server) createLeafNode(conn net.Conn, rURL *url.URL, remote *leafNodeCf
 			c.leaf.isSpoke = true
 		}
 		tlsFirst = remote.TLSHandshakeFirst
+		infoTimeout = remote.FirstInfoTimeout
 		remote.Unlock()
 		c.acc = acc
 	} else {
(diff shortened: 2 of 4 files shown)


derekcollison (commented):
If no one sets anything, how long can a leafnode exists before being declared stalled?

kozlovic:
> If no one sets anything, how long can a leafnode exists before being declared stalled?

`pingInterval*time.Duration(pingMax+1)`, which since default ping interval is 2min, and max pings 2, it would be 2min*3=6min. Now for route and gateways, we cap the ping interval to 15 seconds I believe, so it is much smaller.

If you think this is too much, I could pass different values, but then it means adding a new configuration setting? (which was what the user wanted: make the wait configurable).

derekcollison:
I think we want to make the wait configurable for leafnodes vs using the ping mechanism.

kozlovic:
> I think we want to make the wait configurable for leafnodes vs using the ping mechanism.

Ok, so a new value in a leaf node's remote configuration block and I assume that if not set, default to that constant that we were using.

derekcollison:
Yes I believe so..

kozlovic:
@derekcollison Sorry, did a force push with the amend because I wanted the commit log to be in phase with the use of a new config param instead of the original approach.

derekcollison:
Just ping when ready for review.

kozlovic:
It was/is ready for review. I was just saying that I did a push force (which I usually don't do once a code review has started).

derekcollison (approved):
LGTM

PR #5427, opened 2024-05-16
[IMPROVED] Several improvements to raft layer with limited or no state.

derekcollison:
1. Pull term from tav.idx even with no voted peer on recovery.
2. If we have no pterm or pindex after recovery inherit from empty WAL and tav.idx.
3. Clear wtv if we fail to write the tav.idx file.

Also fixes flapper test with removing snapshots and block state.


wallyqs (approved):
LGTM

Jarema (approved):
LGTM

PR #5428, opened 2024-05-16
Fixes for rescaling streams with sources

wallyqs:
Includes improvements for streams rescaling:

- When scaling up and down we keep same leader so no reason to tear everything down and then back up.
   We do make sure when we scale down to one to tear down cluster specific subs etc.


neilalexander (approved):
LGTM

PR #5432, opened 2024-05-16
[FIXED] Avoid overflow of rand.Intn on 32bit platforms

derekcollison:


wallyqs (approved):
LGTM

PR #5438, opened 2024-05-17
Use Go v1.22 in v2.10 releases

wallyqs:
This will make the next release use Go v1.22 instead when the next version is tagged.


derekcollison (approved):
LGTM

PR #5439, opened 2024-05-17
Update go.mod to use Go v1.21 toolchain

wallyqs:
Binary would be released with Go v1.22 but need to allow building with Go v1.21 as well, so updating to that version first.


derekcollison (approved):
LGTM

wallyqs:
closing in favor of https://github.com/nats-io/nats-server/pull/5538

PR #5446, opened 2024-05-17
[IMPROVED] Consumer with AckAll performance of first ack with large first stream sequence.

derekcollison:
When the first ack was called on an AckAll consumer attached to a stream with a very high first sequence, the server would become unresponsive for that consumer and potentially the stream due to looping down to the existing ackfloor which in this case would be zero.


wallyqs (approved):
LGTM

PR #5448, opened 2024-05-17
Fixed TestMsgTraceJetStreamWithSuperCluster test

kozlovic:
This test was flapping because the order of events are not guaranteed. Made a fix to ensure we have all the expected events (regardless the order they are received).


derekcollison (approved):
LGTM

kozlovic:
@derekcollison I assume you will merge, correct?

PR #5449, opened 2024-05-18
[IMPROVED] First ack with large first sequence and not acking highest pending.

derekcollison:
Make sure to not run away when acking from the middle of pending.
Also optimize loop for interest based streams to ack msgs based on actual floor of pending.


wallyqs (approved):
LGTM

PR #5454, opened 2024-05-20
[IMPROVED] Various stream catchup improvements

derekcollison:
Various improvements to catchup logistics.

1. Make sure to register our sync sub earlier to avoid missing an early request for catchup on a scaleup.
2. Make sure we send trailing delete range blocks to avound getting stuck during catchup.
3. When parallel catchups running make sure to periodically check to see if we can send next batch since we could be blocked on others and not ourselves.
4. When catching up a stream that has a large number of interior deletes use LoadNextMsg() to more efficiently skip over large gaps.
5. When resuming apply queues in raft layer do not unpause our state until after processing to avoid possible elections under heavy load.
6. when syncing deleted blocks from snapshots only consider those applicable to our state.

### DIFF
--- a/server/filestore.go
+++ b/server/filestore.go
@@ -3543,7 +3543,6 @@ func (fs *fileStore) SkipMsgs(seq uint64, num uint64) error {
 	lseq := seq + num - 1
 
 	mb.mu.Lock()
-	var needsRecord bool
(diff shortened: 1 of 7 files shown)


wallyqs (approved):
LGTM

PR #5460, opened 2024-05-21
Add goroutine annotation for stream internal loop

neilalexander:
This will improve observability on the `mset.internalLoop()` in profiles.


derekcollison (commented):
This means that shutdown will wait on this now, we need to make sure that this does not adversely effect shutdown.

neilalexander:
I think it would have been safe as that goroutine listens for `qch` and bails, but have instead switched it around to set the labels inline instead.

derekcollison (approved):
LGTM

PR #5464, opened 2024-05-22
NRG: Avoid panic on corrupted TAV file

neilalexander:
If the file ends up truncated somehow then we should not try to call `Uint64`, otherwise it will result in an out-of-bounds panic.

### DIFF
--- a/server/raft.go
+++ b/server/raft.go
@@ -3735,7 +3735,8 @@ func readPeerState(sd string) (ps *peerState, err error) {
 }
 
 const termVoteFile = "tav.idx"
-const termVoteLen = idLen + 8
+const termLen = 8 // uint64
+const termVoteLen = idLen + termLen
 
 // Writes out our term & vote outside of a specific raft context.
 func writeTermVote(sd string, wtv []byte) error {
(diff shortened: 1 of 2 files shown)


server/raft.go
```diff
@@ -3761,6 +3761,10 @@ func (n *raft) readTermVote() (term uint64, voted string, err error) {
 	if err != nil {
 		return 0, noVote, err
 	}
+	if len(buf) < 8 {
```
Jarema, line 3764: I wonder, should we return an error here?

Though I see that we don't handle errors from `readTermVote()` anyway, beyond checking if its nil or not.

neilalexander: An error isn't really actionable in that scenario anyway as we'd still end up resetting back to term 0, in which case we'd might as well just return term 0.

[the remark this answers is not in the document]
derekcollison: Also be good to understand how this could have ever happened.

Maybe run full test suite and add in check when raft group goes away to check for this condition.

derekcollison (approved):
LGTM - Thanks.

PR #5466, opened 2024-05-23
NRG (2.11): Add ability to move cluster traffic into asset accounts

neilalexander:
This PR adds the ability to move NRG traffic out of the system account and into the asset accounts. Particularly in heavily multi-tenanted systems, this can help cases where head-of-line blocking may happen in the system account due to large amounts of replication traffic.

This needs to be enabled on a per-account basis on each participating server. Servers advertise their capability to support account NRG to each other. If a Raft group detects a peer coming up without support, they will revert back to using the system account automatically.

Requires https://github.com/nats-io/jwt/pull/223 and then `go.mod` updating.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -6,7 +6,7 @@ require (
 	github.com/google/go-tpm v0.9.0
 	github.com/klauspost/compress v1.17.9
 	github.com/minio/highwayhash v1.0.3
-	github.com/nats-io/jwt/v2 v2.5.8
+	github.com/nats-io/jwt/v2 v2.6.0
 	github.com/nats-io/nats.go v1.36.0
 	github.com/nats-io/nkeys v0.4.7
 	github.com/nats-io/nuid v1.0.1
--- a/server/accounts.go
+++ b/server/accounts.go
@@ -62,6 +62,7 @@ type Account struct {
 	sqmu         sync.Mutex
 	sl           *Sublist
 	ic           *client
+	sq           *sendq
 	isid         uint64
 	etmr         *time.Timer
 	ctmr         *time.Timer
@@ -3655,6 +3656,7 @@ func (s *Server) updateAccountClaimsWithRefresh(a *Account, ac *jwt.AccountClaim
 
 	a.updated = time.Now()
 	clients := a.getClientsLocked()
+	ajs := a.js
 	a.mu.Unlock()
 
 	// Sort if we are over the limit.
@@ -3677,6 +3679,26 @@ func (s *Server) updateAccountClaimsWithRefresh(a *Account, ac *jwt.AccountClaim
 		// our imports properly. This allows this server to proxy JS traffic correctly.
 		s.checkJetStreamExports()
 		a.enableAllJetStreamServiceImportsAndMappings()
+	}
+
+	if ajs != nil {
+		// Check whether the account NRG status changed. If it has then we need to notify the
+		// Raft groups running on the system so that they can move their subs if needed.
+		a.mu.Lock()
+		previous := ajs.nrgAccount
+		switch tokens := strings.SplitN(ac.ClusterTraffic, ":", 2); tokens[0] {
+		case "system":
+			a.js.nrgAccount = _EMPTY_
+		case "owner":
+			a.js.nrgAccount = a.Name
+		default:
+			s.Errorf("Account claim for %q has invalid value %q for cluster traffic account", a.Name, ac.ClusterTraffic)
+		}
+		changed := ajs.nrgAccount != previous
+		a.mu.Unlock()
+		if changed {
+			s.updateNRGAccountStatus()
+		}
 	}
 
 	for i, c := range clients {
--- a/server/events.go
+++ b/server/events.go
@@ -264,6 +264,7 @@ type ServerInfo struct {
 const (
 	JetStreamEnabled     ServerCapability = 1 << iota // Server had JetStream enabled.
 	BinaryStreamSnapshot                              // New stream snapshot capability.
+	AccountNRG                                        // Move NRG traffic out of system account.
 )
 
 // Set JetStream capability.
@@ -287,6 +288,17 @@ func (si *ServerInfo) SetBinaryStreamSnapshot() {
 // JetStreamEnabled indicates whether or not we have binary stream snapshot capbilities.
 func (si *ServerInfo) BinaryStreamSnapshot() bool {
 	return si.Flags&BinaryStreamSnapshot != 0
+}
+
+// Set account NRG capability.
+func (si *ServerInfo) SetAccountNRG() {
+	si.Flags |= AccountNRG
+}
+
+// AccountNRG indicates whether or not we support moving the NRG traffic out of the
+// system account and into the asset account.
+func (si *ServerInfo) AccountNRG() bool {
+	return si.Flags&AccountNRG != 0
 }
 
 // ClientInfo is detailed information about the client forming a connection.
@@ -475,10 +487,14 @@ RESET:
 					si.Version = VERSION
 					si.Time = time.Now().UTC()
 					si.Tags = tags
+					si.Flags = 0
 					if js {
 						// New capability based flags.
 						si.SetJetStreamEnabled()
 						si.SetBinaryStreamSnapshot()
+						if s.accountNRGAllowed.Load() {
+							si.SetAccountNRG()
+						}
 					}
 				}
 				var b []byte
@@ -1616,7 +1632,8 @@ func (s *Server) remoteServerUpdate(sub *subscription, c *client, _ *Account, su
 	}
 
 	node := getHash(si.Name)
-	s.nodeToInfo.Store(node, nodeInfo{
+	accountNRG := si.AccountNRG()
+	oldInfo, _ := s.nodeToInfo.Swap(node, nodeInfo{
 		si.Name,
 		si.Version,
 		si.Cluster,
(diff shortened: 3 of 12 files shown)


server/opts.go
```diff
@@ -2310,6 +2311,8 @@ func parseJetStream(v any, opts *Options, errors *[]error, warnings *[]error) er
 					return &configErr{tk, fmt.Sprintf("%s %s", strings.ToLower(mk), err)}
 				}
 				opts.JetStreamMaxCatchup = s
+			case "account_nrg":
```
ripienaar, line 2314: ngr means something to us, but maybe not to users, can we make something a bit more descriptive here?

ripienaar (dismissed):
Can we also expose this in `$SYS.REQ.SERVER.PING.VARZ`? Be nice if `nats server info` could easily show this.

Also in JSZ.

Might also be worth to ask @mprimi to get into audit a check that all machines have this enabled and alert if not symetrical

server/opts.go
neilalexander: I'm not at all attached to the name, suggestions welcome! 😄

mprimi:
@ripienaar what's a good way to track requested/suggested checks and other improvements for `audit`?
Should i open an issue in that project?

@neilalexander I'm not familiar with this feature or the subsystems around it, however just from your description i have a comment/concern:
> enabled on all servers participating in the cluster.
You cannot know if you know all members of the cluster, there may be some unknown unknowns.

How about instead having the the cluster meta-leader is turning this on or off based on the same (ephemeral, possibly outdated) view?

ripienaar:
@mprimi you are welcome to open them here if you wish, add a label (or ask me) so you can find them easily

ripienaar:
IF a new server joins a cluster that is in this mode and actively all happy but that new node is not configured in this manner what happens?

Can we make that node disable JetStream with a error in the logs? 

Is there any concerns with Leafnodes?

derekcollison:
I will try to take a look at this next week when I am back.. Will need to take my time.

Do we have a plan for visibility into $SYS via `nats traffic` when they is opted in?

derekcollison:
@neilalexander let's sync this week on this one and get it merged. Agree we should be able to place in any account.

go.mod
```diff
@@ -2,6 +2,8 @@ module github.com/nats-io/nats-server/v2
 
 go 1.22.0
 
+replace github.com/nats-io/jwt/v2 => github.com/nats-io/jwt/v2 v2.5.9-0.20240910093232-2e1251e51094
```
derekcollison, line 5: Won't merge to main with this here. So need to do the JWT lib release first.

server/accounts.go
```diff
@@ -3679,6 +3680,28 @@ func (s *Server) updateAccountClaimsWithRefresh(a *Account, ac *jwt.AccountClaim
...
 	}
 
+	if a.js != nil {
+		// Check whether the account NRG status changed. If it has then we need to notify the
+		// Raft groups running on the system so that they can move their subs if needed.
+		a.mu.Lock()
+		previous := a.js.nrgAccount
+		switch tokens := strings.SplitN(ac.ClusterTraffic, ":", 2); tokens[0] {
+		case "system":
+			a.js.nrgAccount = ""
```
derekcollison, line 3690: In server we use `_EMPTY_ `vs ""

server/opts.go
```diff
@@ -2141,6 +2141,21 @@ func parseJetStreamForAccount(v any, acc *Account, errors *[]error) error {
...
 				}
 				jsLimits.MaxAckPending = int(vv)
+			case "cluster_traffic":
+				vv, ok := mv.(string)
+				if !ok {
+					return &configErr{tk, fmt.Sprintf("Expected either 'system' or 'account' string value for %q, got %v", mk, mv)}
+				}
+				switch tokens := strings.SplitN(vv, ":", 2); strings.ToLower(tokens[0]) {
+				case "system":
+					acc.js.nrgAccount = ""
```
derekcollison, line 2151: Same here with ""

server/server.go
```diff
@@ -360,6 +360,9 @@ type Server struct {
 
 	// Queue to process JS API requests that come from routes (or gateways)
 	jsAPIRoutedReqs *ipQueue[*jsAPIRoutedReq]
+
+	// Whether moving NRG traffic into accounts is permitted on this server.
+	accountNRGAllowed atomic.Bool
```
derekcollison, line 365: This will always be true yes for 2.11 servers?

derekcollison (changes requested):

go.mod
neilalexander: Yep, as mentioned in the PR description.

server/accounts.go
neilalexander: Ah yes, will fix.

server/server.go
neilalexander: Yes, it's mostly there to test mixed compatibility in the unit tests.

derekcollison: ok, still not a fan of things in released server that are only for tests..

go.mod
derekcollison: ok but to merge we need the jwt lib released and the go.mod properly updated etc..

neilalexander: JWT library released & `go.mod` updated.

wallyqs:
got a panic in one of the runs:
```
=== RUN   TestJetStreamClusterSnapshotBeforePurgeAndCatchup
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0xe90e12]
goroutine 67601 [running]:
github.com/nats-io/nats-server/v2/server.(*raft).recreateInternalSubsLocked(0xc00bfff808)
	/home/travis/build/nats-io/nats-server/server/raft.go:553 +0x92
github.com/nats-io/nats-server/v2/server.(*raft).RecreateInternalSubs(0xc00bfff808)
	/home/travis/build/nats-io/nats-server/server/raft.go:548 +0x89
github.com/nats-io/nats-server/v2/server.(*Server).updateNRGAccountStatus(0xc001944008)
	/home/travis/build/nats-io/nats-server/server/events.go:1725 +0x2f5
github.com/nats-io/nats-server/v2/server.(*Server).remoteServerUpdate(0xc001944008, 0x0?, 0xc00d5a4138?, 0xc001b1c000?, {0xc00082de40?, 0x446685?}, {0xc009036868?, 0xc009036830?}, {0x0, 0x0, ...}, ...)
	/home/travis/build/nats-io/nats-server/server/events.go:1654 +0x736
github.com/nats-io/nats-server/v2/server.(*Server).internalReceiveLoop(0xc001944008, 0xc009036800)
	/home/travis/build/nats-io/nats-server/server/events.go:436 +0x464
created by github.com/nats-io/nats-server/v2/server.(*Server).setSystemAccount in goroutine 67583
	/home/travis/build/nats-io/nats-server/server/server.go:1785 +0x945
FAIL	github.com/nats-io/nats-server/v2/server	373.110s
```

neilalexander:
Panic fixed BTW.

server/raft.go
```diff
@@ -76,6 +76,7 @@ type RaftNode interface {
 	Stop()
 	Delete()
 	Wipe()
+	RecreateInternalSubs() error
```
derekcollison, line 79: Should we throw away old one and create a new one with same name vs trying to patch on the fly?

derekcollison (commented):
We are very close. Just a general question when we see a change, do we want to redo internal state of existing node or should we delete old node and create a new one with same name?

server/raft.go
neilalexander: The advantage of doing it this way is append entries etc that have been received and are being processed will continue to be without interruption, whereas if we have to restart the whole node, we need to flush state to disk for a potentially large number of nodes and pause operations.

derekcollison: The new node will do catchup properly as designed.. It might yield a better result, but I am not adamant about this approach vs current one.

neilalexander: I did go back and forth on this a little to start with. In the end what I came up with was that you can push an account claim update that changes the account at pretty much any time and the system will honour it.

If you have a busy account with hundreds of assets running, restarting all the Raft groups can create a run on the disk (and dios) by reloading all the peer states, TAVs, WALs etc, whereas actually the only things we really need to update are the sendqs and subs.

neilalexander:
For now have disabled being able to select a third account, so now limited to `system` and `owner`.

server/events.go
```diff
@@ -1675,14 +1699,35 @@ func (s *Server) processNewServer(si *ServerInfo) {
...
 	// Announce ourselves..
 	// Do this in a separate Go routine.
 	go s.sendStatszUpdate()
 }
 
+// Works out whether all nodes support moving the NRG traffic into
+// the account and moves it appropriately.
+// Server lock MUST NOT be held on entry.
+func (s *Server) updateNRGAccountStatus() {
+	s.rnMu.Lock()
```
derekcollison, line 1716: nit can be Rlock.

derekcollison (approved):
LGTM - one nit but not a blocker.

derekcollison (approved):
LGTM

server/accounts.go
```diff
@@ -3679,6 +3681,26 @@ func (s *Server) updateAccountClaimsWithRefresh(a *Account, ac *jwt.AccountClaim
...
+		// Raft groups running on the system so that they can move their subs if needed.
+		a.mu.Lock()
+		previous := ajs.nrgAccount
+		switch tokens := strings.SplitN(ac.ClusterTraffic, ":", 2); tokens[0] {
+		case "system":
+			a.js.nrgAccount = _EMPTY_
+		case "owner":
+			a.js.nrgAccount = a.Name
+		default:
+			s.Errorf("Account claim for %q has invalid value %q for cluster traffic account", a.Name, ac.ClusterTraffic)
```
aricart, line 3695: I have a PR for the validation of the JWT, if this is the case, the JWT should be valid when decoded. See https://github.com/nats-io/jwt/pull/224

PR #5470, opened 2024-05-23
Limit statsz updates

### LINKED ISSUE #5469: Flood of STATSZ system events starting nodes in a large cluster
### Observed behavior

The number of `STATSZ` system event messages sent by a newly-started (or restarted) server appears to increase quickly with the total size of the (super)cluster. When the number of nodes grows large, restarting too many at once can trigger a massive flood of system events that can overwhelm capacity and cause an outage across the entire system.

On a production supercluster of ~900 total nodes and a single-cluster size of 100 nodes, restarting a single server generates ~3 GB of total data and ~150k messages (as measured by `out_bytes` and `out_msgs` fields on the `/varz` endpoint) on startup.

Based on my reading of the code, `STATSZ` events are published periodically by a timer (waiting 250ms for the first event, then 500ms, etc until a max interval of 30s)

`STATSZ` events are _also_ published:
- When establishing an outbound gateway connection (`createGateway`)
- When we receive an update from any new server in the supercluster (`processNewServer`)

It's the latter that causes the total number of `STATSZ` messages to increase quickly with the number of nodes, since each update is sent to every other connected client.

### Expected behavior

A much smaller number of `STATSZ` updates when starting/restarting a server that does not increase as quickly with the total number of nodes in the cluster. A rate-limit of 1 update per second for `STATSZ` messages sent to each client subscription would prevent system-event floods while still allowing timely updates to be published across the cluster.

### Server and client version

I observed this issue in production with v2.7.4, but I can also reproduce this issue in an isolated test using the latest nats-server version.

### Host environment

_No response_

### Steps to reproduce

The following test case demonstrates the issue. In a cluster of `totalServers` servers, we would expect the total number of messages to be close to `totalServers`<sup>2</sup>, or a statsz update sent from each server to every other server. However, the result is closer to `totalServers`<sup>3</sup>.

```go
package server

import (
	"testing"
	"time"
)

func TestClusterSetupMsgs(t *testing.T) {
	numServers := 50
	c := createClusterEx(t, true, 100*time.Millisecond, true, "cluster", numServers)
	var totalOut int
	for _, server := range c.servers {
		totalOut += int(server.outMsgs)
	}
	totalExpected := numServers * numServers
	if totalOut >= totalExpected {
		t.Fatalf("Total outMsgs is %d, expected < %d\n", totalOut, totalExpected)
	}
}
```

Sample output, with `numServers` set to 10, 50, 70:
```
Total outMsgs is 759, expected < 100
Total outMsgs is 84573, expected < 2500
Total outMsgs is 226932, expected < 4900
```

wjordan:
This PR adds a filter to `sendStatsz` that limits statsz updates to the current heartbeat interval (max once per second), adding a `time.Time` field to track the time the last statsz update was sent. This limit should reduce overall `STATSZ` system event load in large clusters while still allowing initial statsz update to quickly reach newly-discovered nodes.

Fixes #5469.

Signed-off-by: Will Jordan <will.jordan@gmail.com>

### DIFF
--- a/server/events.go
+++ b/server/events.go
@@ -1002,6 +1002,12 @@ func (s *Server) heartbeatStatsz() {
 	go s.sendStatszUpdate()
 }
 
+// Reset statsz rate limit for the next broadcast.
+// This should be wrapChk() to setup common locking.
+func (s *Server) resetLastStatsz() {
+	s.sys.lastStatsz = time.Time{}
+}
+
 func (s *Server) sendStatszUpdate() {
 	s.sendStatsz(fmt.Sprintf(serverStatsSubj, s.ID()))
 }
@@ -1905,10 +1911,7 @@ func (s *Server) statszReq(sub *subscription, c *client, _ *Account, subject, re
 	// No reply is a signal that we should use our normal broadcast subject.
 	if reply == _EMPTY_ {
 		reply = fmt.Sprintf(serverStatsSubj, s.info.ID)
-		// Reset rate limit for the next broadcast.
-		s.mu.Lock()
-		s.sys.lastStatsz = time.Time{}
-		s.mu.Unlock()
+		s.wrapChk(s.resetLastStatsz)
 	}
 
 	opts := StatszEventOptions{}
--- a/server/events.go
+++ b/server/events.go
@@ -134,6 +134,7 @@ type internal struct {
 	shash          string
 	inboxPre       string
 	remoteStatsSub *subscription
+	lastStatsz     time.Time
 }
 
 // ServerStatsMsg is sent periodically with stats updates.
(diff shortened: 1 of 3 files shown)


server/events.go
```diff
@@ -949,6 +954,35 @@ func (s *Server) sendStatsz(subj string) {
...
+	if s.sys == nil {
+		return true
+	}
+
+	// Only limit the normal broadcast subject.
+	if subj != fmt.Sprintf(serverStatsSubj, s.ID()) {
+		return false
+	}
+
+	interval := time.Second
```
derekcollison, line 971: Make constant?

server/events.go
```diff
@@ -1869,6 +1903,10 @@ func (s *Server) statszReq(sub *subscription, c *client, _ *Account, subject, re
 	// No reply is a signal that we should use our normal broadcast subject.
 	if reply == _EMPTY_ {
 		reply = fmt.Sprintf(serverStatsSubj, s.info.ID)
+		// Reset rate limit for the next broadcast.
+		s.mu.Lock()
```
derekcollison, line 1907: Maybe make small helper function like `limitStatsz`

derekcollison (commented):
Looks pretty good, few small comments.

derekcollison (approved):
LGTM

derekcollison:
This change seems to be causing some tests to flap since the meta layer for jetstream does not have up to date information in order to complete the test.

e.g. TestJetStreamSuperClusterOverflowPlacement

derekcollison:
Might need to fix these flappers before merging. 

We could set the `cstatsz` to a lower value possibly for those tests.

derekcollison:
Also seems we now see a race..

image

derekcollison:
`TestRoutePings` has the datarace now.

wjordan:
the race seems unrelated, neither (`TestRoutePings`) nor `adjustPingInterval` behavior is touched by this PR.

derekcollison (approved):
LGTM

derekcollison:
The race was because your new test `TestClusterSetupMsgs` was not shutting down the cluster when done, so it kept running and it was accessing that `routeMaxPingInterval`. Will clean up.

Also need to make the tweak to `statszRateLimit` global since other tests are flapping too.

Will take care of it.

PR #5475, opened 2024-06-02
[FIXED] Performance regression in LoadNextMsg with very sparse or no msgs.

derekcollison:


neilalexander (approved):
LGTM

PR #5476, opened 2024-06-02
[IMPROVED] Memstore NumPending() and checkInterestState. 

derekcollison:
Optimize NumPending for memstore when large number of interior deletes are present.
Also as we periodically check interest state for interest policy streams and consumers, if we have zero ack consumers do not hold stream lock while checking.


wallyqs (approved):
LGTM

PR #5481, opened 2024-06-04
NRG: Stepdown candidate when AE ahead of last log term

neilalexander:
A node that has been isolated in the candidate state for a considerable amount of time might have bumped its vote term up considerably due to a number of failed election attempts, so when the network recovers, new AEs would not trigger a stepdown. Instead, check the `pterm`, which is the term of the last entry in our log, as this will not have ratcheted up during the period of isolation.


derekcollison (approved):
LGTM - Thanks

PR #5482, opened 2024-06-04
Fix message being redelivered after successful ack on rollout restarts

wallyqs:
Sometimes during rollout restarts, retrying quickly a previously failed ack could succeed but still the consumer would get the message being redelivered.  To prevent this it is now checked if an inbound ack on takeover is ahead of the consumer stream sequence.


derekcollison (approved):
LGTM

PR #5484, opened 2024-06-04
Prevent duplicate routes

### LINKED ISSUE #5483: 'Duplicate Route' connection floods on implicit routes
### Observed behavior

When a large cluster establishes connections to an implicit route, all of the nodes can flood a host with a large number of redundant 'Duplicate Route' connections.

In production with a large cluster of 100 nodes, I've observed intermittent floods of `Router connection closed: Duplicate Route` logged, up to 10k connections/sec to a single server, when a server experiences a broken/unreliable network connection to the rest of the cluster.

### Expected behavior

I would expect 'Duplicate Route' connections to be relatively rare and the rate of connection attempts for any given route to be limited to the retry rate (one per second), instead of an unbounded flood of duplicate connections being sent all at once.

### Server and client version

nats-server: v2.10.16

### Host environment

_No response_

### Steps to reproduce

My proposed fix in ba05173c2878edc763ea5090b528399b2b626f75 includes a test demonstrating the issue, counting the total number of 'Duplicate Route' log entries when establishing a 10-server cluster:
https://github.com/nats-io/nats-server/blob/ba05173c2878edc763ea5090b528399b2b626f75/server/routes_test.go#L4324-L4336

Without any fix, the number of duplicate routes:

| n=10  | n=30 | n=50 | n=70 |
| - | - | - | - |
| 150 | 7000 | 38000 | 80000 |

My best understanding is that there's a feedback loop on implicit routes:
1. When a route is registered (`addRoute`), the server sends an `INFO` broadcast to all servers in the cluster (`forwardNewRouteInfoToKnownServers`).
2. When each server handles this INFO message (`processImplicitRoute`), it creates an implicit-route connection (`connectToRoute`).
3. When the implicit-route connection is registered (`addRoute`), see 1.

Although `processImplicitRoute` skips the connection if it's explicitly-configured or if the route already exists, this only checks against registered routes (not unregistered routes or unestablished connections), so a flood of connection attempts can pile up before a successful route registration prevents future connections.

With a fix that checks the route-connection count before dialing a new connection, the number of duplicate routes isn't completely eliminated for larger clusters, but is significantly less:

| n=10  | n=30 | n=50 | n=70 |
| - | - | - | - |
| 0 | 0 | 500 | 5000 |

wjordan:
This PR attempts to prevent duplicate routes by checking the existing route-connection count before attempting to create a new connection.

Fixes #5483.

Before attempting a new connection, `connectToRoute` checks `s.numRouteConns` (the total sum of registered routes, unregistered routes, and pending route-connections), and skips the connection attempt if the server is already at its limit. This check prevents a pileup of pending connections (which will be eventually closed as duplicate routes) described in the linked issue #5483.

In order to track pending route-connections, this PR adds a counter map (`s.pendingRouteConns`) that is incremented on each route-connection attempt and decremented after the connection is established/aborted.

Signed-off-by: Your Name <will.jordan@gmail.com>

### DIFF
--- a/server/route.go
+++ b/server/route.go
@@ -3016,24 +3016,11 @@ func (s *Server) forEachRemote(f func(r *client)) {
 	}
 }
 
-func (s *Server) forEachRouteAndTempRoute(f func(r *client)) {
-	s.forEachRoute(f)
-	s.grMu.Lock()
-	defer s.grMu.Unlock()
-	for _, c := range s.grTmpClients {
-		if c != nil && c.kind == ROUTER {
-			f(c)
-		}
-	}
-}
-
 // Returns number of route connections for the host,
-// including not-yet-registered and unestablished connections.
+// including pending (not-yet-registered and unestablished) connections.
 func (s *Server) numRouteConns(host string, accName string) int {
 	nr := s.pendingRouteConns.Value(fmt.Sprintf("%s/%s", host, accName))
-	s.forEachRouteAndTempRoute(func(c *client) {
-		c.mu.Lock()
-		defer c.mu.Unlock()
+	s.forEachRoute(func(c *client) {
 		if c.route.url != nil &&
 			c.route.url.Host == host &&
 			accName == bytesToString(c.route.accName) {
--- a/server/route.go
+++ b/server/route.go
@@ -2603,9 +2603,20 @@ func (s *Server) connectToRoute(rURL *url.URL, tryForEver, firstConnect bool, ac
 
 	const connErrFmt = "Error trying to connect to route (attempt %v): %v"
 
+	if rURL == nil {
+		return
+	}
 	s.mu.Lock()
 	resolver := s.routeResolver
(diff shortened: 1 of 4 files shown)


derekcollison:
@kozlovic could you possibly take a look?

kozlovic:
@derekcollison @wjordan Sorry for the delay. I did look at this PR but with PR #5602 test, I still could see duplicate routes with @wjordan approach. The reason this PR did not see after the changes is likely the way the cluster was formed, starting one server at a time. The ref count added in this PR is done "too late", in connectToRoute(), which is running from a go routine. When servers start concurrently, there are way more chances for the duplicate routes situation to occur, even with this PR.

That made me realize that the issue was really the excess of gossip protocol, which is what I tackled in PR #5602. @wjordan Let me know if you think the other approach is solving the issue in your environment. Thanks!

derekcollison:
Closing in favor of @kozlovic's PR.

PR #5485, opened 2024-06-05
Fixes for tests that depend on fast statsz

derekcollison:
Also minor fix for new test to cleanup cluster when done.

### DIFF
--- a/server/events.go
+++ b/server/events.go
@@ -98,7 +98,11 @@ const (
 // FIXME(dlc) - make configurable.
 var eventsHBInterval = 30 * time.Second
 
-var statszRateLimit = 1 * time.Second
+// Default minimum wait time for sending statsz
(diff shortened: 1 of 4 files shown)


server/events.go
```diff
@@ -98,7 +98,11 @@ const (
 // FIXME(dlc) - make configurable.
 var eventsHBInterval = 30 * time.Second
 
-var statszRateLimit = 1 * time.Second
+// Default minimum wate time for sending statsz
```
bruth, line 101:
```suggestion
// Default minimum wait time for sending statsz
```

bruth (approved):
LGTM

server/events.go
derekcollison: Thanks will fix..

derekcollison:
Thanks..

wallyqs (approved):

PR #5489, opened 2024-06-05
Fix returning `maximum consumers limit reached` on some consumer updates

### LINKED ISSUE #5488: JetStream - Unable to update consumer config when max consumers is 1
### Observed behavior

Created stream with max consumers = 1, created a consumer, tried to update the consumer config - and got "maximum consumers limit reached" error.

### Expected behavior

Consumer config updated successfully.

### Server and client version

Server versions tested: 2.10.7, 2.10.16
Client version: 0.1.4

### Host environment

_No response_

### Steps to reproduce

Run the following commands (pressing enter on all prompts, for default configuration):
```
nats stream add temp_stream --max-consumers=1 --retention=interest --subjects=temp_stream
nats consumer add temp_stream temp_consumer --max-pending=1000
nats consumer edit temp_stream temp_consumer --max-pending=1001 -f

```
Last command printed the following error:
`nats: error: maximum consumers limit reached (10026)`

(It also happens when creating using nats req "$JS.API.CONSUMER.CREATE")

neilalexander:
This will fix #5488 by doing two things:

1. Only enforcing the max consumer count on consumer create or create-or-update requests;
2. Not incorrectly counting a potentially existing matching consumer on create-or-update requests.

Note this only happens in clustered mode, on a single node system this does not happen.

### DIFF
--- a/server/jetstream_cluster.go
+++ b/server/jetstream_cluster.go
@@ -7159,25 +7159,45 @@ func (s *Server) jsClusteredConsumerRequest(ci *ClientInfo, acc *Account, subjec
 		return
 	}
 
+	// Was a consumer name provided?
+	var oname string
+	if isDurableConsumer(cfg) || cfg.Name != _EMPTY_ {
+		if cfg.Name != _EMPTY_ {
+			oname = cfg.Name
+		} else {
+			oname = cfg.Durable
+		}
+	}
+
 	// Check for max consumers here to short circuit if possible.
 	// Start with limit on a stream, but if one is defined at the level of the account
 	// and is lower, use that limit.
-	maxc := sa.Config.MaxConsumers
-	if maxc <= 0 || (selectedLimits.MaxConsumers > 0 && selectedLimits.MaxConsumers < maxc) {
-		maxc = selectedLimits.MaxConsumers
-	}
-	if maxc > 0 {
-		// Don't count DIRECTS.
-		total := 0
-		for _, ca := range sa.consumers {
-			if ca.Config != nil && !ca.Config.Direct {
-				total++
-			}
+	if action == ActionCreate || action == ActionCreateOrUpdate {
+		maxc := sa.Config.MaxConsumers
+		if maxc <= 0 || (selectedLimits.MaxConsumers > 0 && selectedLimits.MaxConsumers < maxc) {
+			maxc = selectedLimits.MaxConsumers
 		}
-		if total >= maxc {
-			resp.Error = NewJSMaximumConsumersLimitError()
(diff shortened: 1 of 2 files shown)


server/jetstream_cluster.go
```diff
@@ -7204,16 +7224,10 @@ func (s *Server) jsClusteredConsumerRequest(ci *ClientInfo, acc *Account, subjec
 	}
 
 	var ca *consumerAssignment
-	var oname string
 
 	// See if we have an existing one already under same durable name or
 	// if name was set by the user.
 	if isDurableConsumer(cfg) || cfg.Name != _EMPTY_ {
```
derekcollison, line 7230: Just check `oname != _EMPTY_`?

server/jetstream_cluster.go
```diff
@@ -7159,25 +7159,45 @@ func (s *Server) jsClusteredConsumerRequest(ci *ClientInfo, acc *Account, subjec
...
-	if maxc <= 0 || (selectedLimits.MaxConsumers > 0 && selectedLimits.MaxConsumers < maxc) {
-		maxc = selectedLimits.MaxConsumers
-	}
-	if maxc > 0 {
-		// Don't count DIRECTS.
-		total := 0
-		for _, ca := range sa.consumers {
-			if ca.Config != nil && !ca.Config.Direct {
-				total++
+	if action == ActionCreate || action == ActionCreateOrUpdate {
```
derekcollison, line 7175: For ` ActionCreateOrUpdate`, what if we determine it is an update?

neilalexander: The `if oname != "" && cn == oname && sa.consumers[oname] != nil` is checking if we think it's an update, if it is then we skip counting the matching consumer assignment so that we don't hit the `>=` case that returns an error. 

If it turns out not to be a genuine update (i.e. the consumer doesn't exist) then it doesn't matter as there are checks below that catch that case and error out.

server/jetstream_cluster.go
neilalexander: Good point, will update.

server/jetstream_cluster.go
```diff
@@ -7227,7 +7227,7 @@ func (s *Server) jsClusteredConsumerRequest(ci *ClientInfo, acc *Account, subjec
 
 	// See if we have an existing one already under same durable name or
 	// if name was set by the user.
-	if isDurableConsumer(cfg) || cfg.Name != _EMPTY_ {
+	if oname != "" {
```
derekcollison, line 7230: _EMPTY_

derekcollison (approved):
LGTM

PR #5496, opened 2024-06-05
Improve management of creating/deleting internal subscriptions and source/mirror consumers on leadership changes

jnmoyne:
Because leadership status can sometimes change very quickly (depending on what causes the change, e.g. false-positive on idle HB timeout due to high message load, multiple causes have been observed running the test many time) and raft.updateLeadChange() can 'drop' some lead change signals which means that sometimes stream.setLeader(isLeader bool) gets called twice in a row with the same isLeader value.

This could lead to in some cases the sourcing to 'stall' (not all messages gets sourced) or end up with 'duplicate messages' (more messages than expected get sourced).

These changes improve clean up and setup of all the internal subscriptions and source/mirror consumers associated with becoming the stream leader (or stepping down). Including doing a proper clean up beforet re-establishing those subscriptions and direct consumers in case some of the leadership state change notification are discarded in the channel.

The effects of the change can be demonstrated by running the TestStreamSourcingScalingSourcingManyBenchmark load test a large number of times (I try to do at least 100, which takes a long time): without these changes the test flaps and fails between 10 and 20% of the time (most of the time 'stalling' but also sometimes with 'too many').

With the changes the test didn't fail so far (note that I am still runing more test iterations) with enough iterations to be confident enough that at the very least these changes very much improve the behavior from the current code and are worthy of creating this PR.

### DIFF
--- a/server/jetstream_sourcing_scaling_test.go
+++ b/server/jetstream_sourcing_scaling_test.go
@@ -16,11 +16,94 @@ package server
 import (
 	"encoding/json"
 	"fmt"
+	"github.com/nats-io/nats.go"
+	"strconv"
 	"testing"
 	"time"
-
-	"github.com/nats-io/nats.go"
 )
+
+var serverConfig1 = `
+server_name: server1
+listen: 127.0.0.1:4222
+http: 8222
+
+prof_port = 18222
+
+jetstream: {max_mem_store: 256MB, max_file_store: 2GB, store_dir: '%s'}
+
+cluster {
+  name: my_cluster
+  listen: 127.0.0.1:4248
+  routes: [nats://127.0.0.1:4249,nats://127.0.0.1:4250]
+}
+
+accounts { $SYS { users = [ { user: "admin", pass: "s3cr3t!" } ] }}
+	`
+
+var serverConfig2 = `
+server_name: server2
+listen: 127.0.0.1:5222
+http: 8223
+
+prof_port = 18223
+
+jetstream: {max_mem_store: 256MB, max_file_store: 2GB, store_dir: '%s'}
+
+cluster {
+  name: my_cluster
+  listen: 127.0.0.1:4249
+  routes: [nats://127.0.0.1:4248,nats://127.0.0.1:4250]
+}
+
+accounts { $SYS { users = [ { user: "admin", pass: "s3cr3t!" } ] }}
+`
+
+var serverConfig3 = `
+server_name: server3
+listen: 127.0.0.1:6222
+http: 8224
+
+prof_port = 18224
+
+jetstream: {max_mem_store: 256MB, max_file_store: 2GB, store_dir: '%s'}
+
+cluster {
+  name: my_cluster
+  listen: 127.0.0.1:4250
+  routes: [nats://127.0.0.1:4248,nats://127.0.0.1:4249]
+}
+
+accounts { $SYS { users = [ { user: "admin", pass: "s3cr3t!" } ] }}
+`
+
+var connectURL = "nats://127.0.0.1:4222,nats://127.0.0.1:4223,nats://127.0.0.1:4224"
+
+func createMyLocalCluster(t *testing.T) *cluster {
+	c := &cluster{servers: make([]*Server, 0, 3), opts: make([]*Options, 0, 3), name: "C3"}
+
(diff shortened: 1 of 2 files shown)


server/jetstream_sourcing_scaling_test.go
```diff
@@ -345,45 +433,83 @@ func TestStreamSourcingScalingSourcingManyBenchmark(t *testing.T) {
...
+	c.waitOnStreamLeader(globalAccountName, "sourcing")
 	sl := c.streamLeader(globalAccountName, "sourcing")
-	mset, err := sl.GlobalAccount().lookupStream("sourcing")
-	require_NoError(t, err)
+	fmt.Printf("Sourcing stream created leader is *** %s ***\n", sl.Name())
 
-	start := time.Now()
+	expectedSeq := make([]int, numSourced)
+
+	//var steppedDown bool
```
wallyqs, line 442: leftover  comment

server/jetstream_sourcing_scaling_test.go
```diff
@@ -345,45 +433,83 @@ func TestStreamSourcingScalingSourcingManyBenchmark(t *testing.T) {
...
-				sl = c.streamLeader(globalAccountName, "sourcing")
-				mset, err = sl.GlobalAccount().lookupStream("sourcing")
-				require_NoError(t, err)
-			}
-			return fmt.Errorf("Expected %d messages, got %d", numMsgPerSource*uint64(numSourced), state.Msgs)
+			//if !steppedDown && state.Msgs >= 200_000 {
+			//	fmt.Printf("\nStepping Down\n")
+			//	mset.node.StepDown()
+			//	steppedDown = true
+			//	// Need to repull our data.
```
wallyqs, line 467: commented section here

server/jetstream_sourcing_scaling_test.go
```diff
@@ -16,12 +16,95 @@ package server
...
+  listen: 127.0.0.1:4250
+  routes: [nats://127.0.0.1:4248,nats://127.0.0.1:4249]
+}
+
+accounts { $SYS { users = [ { user: "admin", pass: "s3cr3t!" } ] }}
+`
+
+var connectURL = "nats://127.0.0.1:4222,nats://127.0.0.1:4223,nats://127.0.0.1:4224"
+
+func createMyLocalCluster(t *testing.T) *cluster {
```
wallyqs, line 81: you can use this helper instead: 
https://github.com/nats-io/nats-server/blob/main/server/jetstream_cluster_4_test.go#L1435-L1457

server/jetstream_sourcing_scaling_test.go
jnmoyne: Cleaned up. Those were part of a previous change to the test that would artificially trigger a leader change towards the start of the test, which actually would hide some of the possible failure scenarios.

server/jetstream_sourcing_scaling_test.go
jnmoyne: Cleaned up. Those were part of a previous change to the test that would artificially trigger a leader change towards the start of the test, which actually would hide some of the possible failure scenarios.

server/jetstream_sourcing_scaling_test.go
jnmoyne: I did it this way rather than using the helper because with the helper I do not control which ports the servers will actually use, and since this is a long running test where I often will monitor things while the test is running (outside of the test itself) it is convenient if the servers' ports are the same each time (so that I have a context setup that I can use to interact directly with the servers from outside the test while the test is running). 

Doesn't affect the test, just a convenience way to easily look at or interact with the servers while the test is running.

jnmoyne:
Also note (if you are trying for yourself to run the test many times to verify it did flap and doesn't flap anymore) that the test can sometimes fail by hanging in the publish phase of the test (and the larger the batch size, the more likely it is to happen). While this is an issue worthy to investigate on it's own, it's not at all the intent of the changes in this PR to address this publish issue, purely the sourcing stalling or generating duplicated messages, so when the test fails in the publish phase (sympthoms are publish are 'frozen' and CPU on the servers is very low) I just discard that particular run.

jnmoyne:
Closed in favor of #5504 which is simpler as it only increases the channel's size yet seems to yield similar improvements.

PR #5497, opened 2024-06-06
[FIXED] Last sequence stream reset on server restarts

derekcollison:
When we had skip msgs on the last msg block we were not writing proper tombstone markers if the block was empty.
 Also no checksum should match no recorded block checksum to pull in metadata from index.db.

Also we will now wait for all async processes to complete on shutdown from signal handlers and lameduck mode.
 
### DIFF
--- a/server/filestore.go
+++ b/server/filestore.go
@@ -1869,7 +1869,7 @@ func (mb *msgBlock) lastChecksum() []byte {
 		mb.rbytes = uint64(fi.Size())
 	}
 	if mb.rbytes < checksumSize {
-		return nil
+		return lchk[:]
 	}
 	// Encrypted?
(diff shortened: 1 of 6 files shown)


wallyqs (approved):
LGTM

server/filestore.go
```diff
@@ -3453,6 +3453,17 @@ func (mb *msgBlock) skipMsg(seq uint64, now time.Time) {
 		mb.last.ts = nowts
 		atomic.StoreUint64(&mb.first.seq, seq+1)
 		mb.first.ts = nowts
+		needsRecord = mb == mb.fs.lmb
+		if needsRecord && mb.rbytes > 0 {
+			// We wany to make sure since we have no messages
```
jarretlavallee, line 3458:
```suggestion
			// We want to make sure since we have no messages
```

PR #5499, opened 2024-06-06
systemd example: reduce risk of naive breakage

philpennock:
In commit c7887427 I made various updates to one of the example systemd service
scripts, but one change causes breakage if and only if the service file is
itself named `nats.service` when installed.

We now know that some folks are doing just that, automatically; it's bad to
have a config file which only works if it has certain names (and a little
surprising that systemd doesn't detect that the alias matches its current name
and filters).

So turn that directive into a commented-out suggestion.


derekcollison (approved):
LGTM

PR #5501, opened 2024-06-07
Fix data race between `processClusterCreateConsumer` and `resetClusteredState`

neilalexander:
Antithesis found this.


server/jetstream_cluster.go
```diff
@@ -4450,7 +4450,10 @@ func (js *jetStream) processClusterCreateConsumer(ca *consumerAssignment, state
 			}
 		}
 
-		if rg.node == nil {
+		js.mu.RLock()
```
derekcollison, line 4453: We have a check above that it did not check. Best to capture node right as we enter the upper else clause.

derekcollison (changes requested):

derekcollison (approved):
LGTM

PR #5503, opened 2024-06-07
Prevent interleaving of setting/unsetting observer states with `jetstream_cluster_migrate`

neilalexander:
There is a possibility this might happen if `reConnectToRemoteLeafNode` from a connection drop and `connectToRemoteLeafNode` from solicitation are running in separate goroutines for the same remotes.

### DIFF
--- a/server/leafnode.go
+++ b/server/leafnode.go
@@ -584,6 +584,9 @@ func (s *Server) clearObserverState(remote *leafNodeCfg) {
 		return
 	}
 
+	acc.jscmMu.Lock()
+	defer acc.jscmMu.Unlock()
+
 	// Walk all streams looking for any clustered stream, skip otherwise.
 	for _, mset := range acc.streams() {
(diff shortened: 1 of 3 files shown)


derekcollison (commented):
What prompted this? Should we have a test?

Also any introduction of a lock should update locking order doc.

server/leafnode.go
```diff
@@ -584,6 +584,9 @@ func (s *Server) clearObserverState(remote *leafNodeCfg) {
 		return
 	}
 
+	acc.jscmMu.Lock()
+	defer acc.jscmMu.Unlock()
+
 	// Walk all streams looking for any clustered stream, skip otherwise.
 	for _, mset := range acc.streams() {
```
derekcollison, line 591: if acc.streams takes lock we should make sure this is called out in ordering.

derekcollison (approved):
LGTM

PR #5504, opened 2024-06-07
Improve leadership change signaling

jnmoyne:
Because leadership status can sometimes change very quickly and raft.updateLeadChange() can 'drop' some lead change signals, especially with the leadc channel having a capacity of one, which means that sometimes stream.setLeader(isLeader bool) gets called twice in a row with the same isLeader value.

This could lead to in some cases the sourcing to 'stall' (not all messages gets sourced) or 'duplicate messages' (more messages than expected get sourced).

Changes are increasing the leadc channel's size to 32.

The effects of the change can be demonstrated by running the TestStreamSourcingScalingSourcingManyBenchmark load test a large number of times (e.g. 100 times): without these changes the test flaps and fails between 10 and 20% of the time (most of the time 'stalling' but also sometimes with 'too many'). With the changes over around a hundred runs not once did the test fail with 'stalling' and only once did it fail with 'too many' messages (so there's still a bug/corner case that could result in 'more than expected' messages to be sourced, although it seems to be a rare occurence (i.e. 1% of the time)).

This is a replacement for #5496

### DIFF
--- a/server/jetstream_sourcing_scaling_test.go
+++ b/server/jetstream_sourcing_scaling_test.go
@@ -22,57 +22,57 @@ import (
 )
 
 var serverConfig1 = `
-server_name: server1
-listen: 127.0.0.1:4222
-http: 8222
+	server_name: server1
+	listen: 127.0.0.1:4222
+	http: 8222
 
-prof_port = 18222
+	prof_port = 18222
 
-jetstream: {max_mem_store: 256MB, max_file_store: 2GB, store_dir: '%s'}
+	jetstream: {max_mem_store: 256MB, max_file_store: 2GB, store_dir: '%s'}
 
-cluster {
-  name: my_cluster
-  listen: 127.0.0.1:4248
-  routes: [nats://127.0.0.1:4249,nats://127.0.0.1:4250]
-}
+	cluster {
+  		name: my_cluster
+  		listen: 127.0.0.1:4248
+  		routes: [nats://127.0.0.1:4249,nats://127.0.0.1:4250]
+	}
 
-accounts { $SYS { users = [ { user: "admin", pass: "s3cr3t!" } ] }}
-	`
+	accounts { $SYS { users = [ { user: "admin", pass: "s3cr3t!" } ] }}
+`
 
 var serverConfig2 = `
-server_name: server2
-listen: 127.0.0.1:5222
-http: 8223
+	server_name: server2
+	listen: 127.0.0.1:5222
+	http: 8223
 
-prof_port = 18223
+	prof_port = 18223
 
-jetstream: {max_mem_store: 256MB, max_file_store: 2GB, store_dir: '%s'}
+	jetstream: {max_mem_store: 256MB, max_file_store: 2GB, store_dir: '%s'}
 
-cluster {
-  name: my_cluster
-  listen: 127.0.0.1:4249
-  routes: [nats://127.0.0.1:4248,nats://127.0.0.1:4250]
-}
+	cluster {
+  		name: my_cluster
+  		listen: 127.0.0.1:4249
+  		routes: [nats://127.0.0.1:4248,nats://127.0.0.1:4250]
+	}
 
-accounts { $SYS { users = [ { user: "admin", pass: "s3cr3t!" } ] }}
+	accounts { $SYS { users = [ { user: "admin", pass: "s3cr3t!" } ] }}
 `
 
 var serverConfig3 = `
-server_name: server3
-listen: 127.0.0.1:6222
-http: 8224
+	server_name: server3
+	listen: 127.0.0.1:6222
+	http: 8224
 
-prof_port = 18224
+	prof_port = 18224
 
-jetstream: {max_mem_store: 256MB, max_file_store: 2GB, store_dir: '%s'}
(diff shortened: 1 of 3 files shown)


server/raft.go
```diff
@@ -387,7 +387,7 @@ func (s *Server) startRaftNode(accName string, cfg *RaftConfig, labels pprofLabe
 		apply:    newIPQueue[*CommittedEntry](s, qpfx+"committedEntry"),
 		stepdown: newIPQueue[string](s, qpfx+"stepdown"),
 		accName:  accName,
-		leadc:    make(chan bool, 1),
+		leadc:    make(chan bool, 32),
```
neilalexander, line 390: Wonder if it is better for this to be an `*ipQueue[bool]` instead? I'd like to hope that we don't switch ever leadership state that quickly 32 times, but if it's important that we never lose that transition, IP queues don't have backpressure like plain channels do.

server/jetstream_sourcing_scaling_test.go
```diff
@@ -14,259 +14,126 @@
 package server
 
 import (
-	"encoding/json"
 	"fmt"
+	"github.com/nats-io/nats.go"
```
derekcollison, line 18: I know a PITA but let's put proper tab indentation on these configs.

server/stream.go
```diff
@@ -839,6 +839,28 @@ func (mset *stream) isLeader() bool {
 	return true
 }
 
+// Lock should be held
+func (mset *stream) stepDown() {
+	// cancel timer to create the source consumers if not fired yet
```
derekcollison: Start comments with caps.

server/stream.go
```diff
@@ -839,6 +839,28 @@ func (mset *stream) isLeader() bool {
...
+	if mset.sourcesConsumerSetup != nil {
+		mset.sourcesConsumerSetup.Stop()
+		mset.sourcesConsumerSetup = nil
+	} else { // otherwise cancel the source consumers
```
derekcollison: Should we qualify if needed here?

server/stream.go
```diff
@@ -876,14 +854,23 @@ func (mset *stream) setLeader(isLeader bool) error {
...
 			mset.mu.Unlock()
 			return err
 		}
 	} else {
-		mset.stepDown()
+		// cancel timer to create the source consumers if not fired yet
+		if mset.sourcesConsumerSetup != nil {
+			mset.sourcesConsumerSetup.Stop()
+			mset.sourcesConsumerSetup = nil
+		}
```
derekcollison, line 867: This else clause got dropped from above `mset.stopSourceConsumers()`

server/raft.go
derekcollison, line 390: Also did we want to change the second conditional in placing values on this like we discussed?

derekcollison (changes requested):

server/stream.go
jnmoyne: Technically the stopSourceConsumers() could be always executed rather than only if mset.sourcesConsumerSetpup != nil.

But it is now removed along with all the other cleanup I had added initially.

server/stream.go
jnmoyne: The else clause was part of the cleanup that I had initially added and that you asked me to remove

server/stream.go
jnmoyne: Now removed along with all the other cleanup stuff I had added initially

server/jetstream_sourcing_scaling_test.go
jnmoyne: ? it is with a tab indentation for me. 
(I use gofmt to reformat)

server/stream.go
derekcollison: Let's add that one in, but put conditional so code reads a bit cleaner, so `if ... {`

server/raft.go
jnmoyne: In my experimentation with the `TestStreamSourcingScalingSourcingManyBenchmark` test it only flipped quickly once in a row. But it could certainly be possible that in other scenarios it could flip quickly more than once, but 32 is a pretty high number so (at least in my testing) it doesn't ever get that high and the limit (where signals start getting dropped) is never reached.

Note that in the previous version (#5496) I had not touched the size of the channel and instead made `setLeader()` remember what state it is in and force a cleanup and re-setup in case it was already the leader and it gets a setLeader(true) (which would happen when a signal of false was followed a fraction of a second later with a signal of true - and therefore the signal of false was dropped by the channel because of capacity of 1). But this is not there anymore in this version of the PR.

server/jetstream_sourcing_scaling_test.go
derekcollison: Manually put tabs in front of lines on configs, like we do in other tests etc. So reads easier, parser will handle.

jnmoyne: Sorry mis-understood which part of the code the comment was for, added tabs to the configs.

server/stream.go
jnmoyne: Added back

server/stream.go
jnmoyne: ok done.

derekcollison (approved):
LGTM

PR #5505, opened 2024-06-08
Optimize #5504

jnmoyne:
Optimizes #5504 by moving the tests for queued appendEntryResponses to the start of `lostQuorumLocked()` rather than at the end.


derekcollison (approved):
LGTM

PR #5506, opened 2024-06-10
[IMPROVED] Memory based streams and NRG behavior during server restarts

derekcollison:
Improvements to catchups and health checks.

Improvements to handling snapshots for memory based wals.
With memory based wals we can not use snapshots on restarts, but we do use them while they are running.
However if a server becomes a leader with no snapshot it will be forced to stepdown when asked to catchup a follower. So we now inherit a leaders snapshot.

Also when we tried to truncate on a mismatch, we needed to truncate the previous index, not current.
When we fail due to the previous entry being compacted away, we would reset. We now reset the wal to the prior index and use the truncate term and index.

Lastly if we receive a heartbeat with correct index but newer term just inherit.
For stream health checks for replicated streams make sure that the monitor routine is running.
When waiting on consumer assignments at the beginning of the stream monitor, make sure the consumer monitor is running as well if replicated.

On a consumer snapshot, register pre-acks as needed.
On stream checkInterestState reset an empty stream to the low ack floor from all consumers.

Last fix consistency bug with memstore when skipping msgs on empty stream to ensure first == last + 1.

### DIFF
--- a/server/jetstream_cluster.go
+++ b/server/jetstream_cluster.go
@@ -2384,7 +2384,7 @@ func (js *jetStream) monitorStream(mset *stream, sa *streamAssignment, sendSnaps
 		mset.waitOnConsumerAssignments()
 		// Setup a periodic check here.
 		// We will fire in 5s the first time then back off to 30s
-		cist = time.NewTicker(10 * time.Second)
+		cist = time.NewTicker(5 * time.Second)
 		cistc = cist.C
 	}
 
--- a/server/jetstream_cluster.go
+++ b/server/jetstream_cluster.go
@@ -534,12 +534,18 @@ func (js *jetStream) isStreamHealthy(acc *Account, sa *streamAssignment) bool {
 		return false
 	}
 
-	// If we are catching up return false.
-	if mset.isCatchingUp() {
+	// If R1 we are good.
+	if node == nil {
+		return true
+	}
+
+	// Here we are a replicated stream.
+	// First make sure our monitor routine is running.
+	if !mset.isMonitorRunning() {
 		return false
 	}
 
-	if node == nil || node.Healthy() {
+	if node.Healthy() {
 		// Check if we are processing a snapshot and are catching up.
 		if !mset.isCatchingUp() {
 			return true
@@ -553,7 +559,6 @@ func (js *jetStream) isStreamHealthy(acc *Account, sa *streamAssignment) bool {
 			js.restartStream(acc, sa)
 		}
 	}
-
 	return false
 }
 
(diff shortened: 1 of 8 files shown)


server/jetstream_cluster.go
```diff
@@ -2373,7 +2383,8 @@ func (js *jetStream) monitorStream(mset *stream, sa *streamAssignment, sendSnaps
 		// since we process streams first then consumers as an asset class.
 		mset.waitOnConsumerAssignments()
 		// Setup a periodic check here.
-		cist = time.NewTicker(30 * time.Second)
+		// We will fire in 5s the first time then back off to 30s
+		cist = time.NewTicker(10 * time.Second)
```
philpennock, line 2387: This is not 5s.

server/raft.go
```diff
@@ -3247,6 +3255,12 @@ func (n *raft) processAppendEntry(ae *appendEntry, sub *subscription) {
 		n.updateLeadChange(false)
 	}
 
+	// Check if this a new heartbeat with a newer term but same index.
+	// If so we take on the new term and proceed.
+	if isNew && ae.pterm > n.pterm && ae.pindex == n.pindex && len(ae.entries) == 0 {
+		n.term, n.pterm = ae.pterm, ae.term
```
neilalexander, line 3261: I'm not sure that updating our own `pterm` in this way is safe. Shouldn't it represent what we believe the term was at our last local log entry?

server/jetstream_cluster.go
derekcollison: Thanks will fix.

server/raft.go
derekcollison: Yes however we aggressively send HBs and they can come before peer state in my testing over weekend.

derekcollison: So without this the heartbeat causes us to get rid of last entry, enter catchuo, get it back exactly the same, then the peerstate from the takeover which moves pterm, etc..

neilalexander: I think resetting the vote `term` is safe, but for `pterm` it seems like notching that up from the heartbeats means we lose our ability to tell if the leader's log is consistent with ours, i.e. if you have one node that has index 10 as belonging to term 5 and another node that has index 10 belonging to index 6, those are divergent logs and we'd rely on `pterm` to highlight that I think.

derekcollison: I can remove it, the system will work just in the case of leader takeover from restart the path will be a bit slower with a catchup loop that is a no-op.

derekcollison: Will remove for now and track down the HB trigger condition later.

neilalexander (approved):
LGTM

PR #5513, opened 2024-06-10
Fix js-server-only health check on v2.10.17

wallyqs:
Got removed while applying last cherry pick so adding it back with the expected behavior for v2.10.X.


derekcollison (approved):
LGTM

PR #5516, opened 2024-06-11
NRG: Reset election timer when leaving observer mode

neilalexander:
Previously if `runAsFollower` or `PauseApply` had ran while in observer state, i.e. due to `jetstream_cluster_migrate` being set when leafnode connections dropped, then the election timer would have been reset to a very high value. Then leaving observer state wouldn't have noticed the lack of a leader for some time. This fix allows a leadership election to take place very quickly after leaving observer mode.

### DIFF
--- a/server/jetstream_leafnode_test.go
+++ b/server/jetstream_leafnode_test.go
@@ -1242,3 +1242,85 @@ func TestJetStreamLeafNodeSvcImportExportCycle(t *testing.T) {
 	_, err = js.Publish("foo", []byte("msg"))
 	require_NoError(t, err)
 }
+
+func TestJetStreamLeafNodeJSClusterMigrateRecovery(t *testing.T) {
+	tmpl := strings.Replace(jsClusterAccountsTempl, "store_dir:", "domain: hub, store_dir:", 1)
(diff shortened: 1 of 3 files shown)


server/jetstream_leafnode_test.go
```diff
@@ -1242,3 +1242,85 @@ func TestJetStreamLeafNodeSvcImportExportCycle(t *testing.T) {
...
+
+	// Create an asset in the leafnode cluster.
+	si, err := ljs.AddStream(&nats.StreamConfig{
+		Name:     "TEST",
+		Subjects: []string{"foo"},
+		Replicas: 3,
+	})
+	require_NoError(t, err)
+	require_Equal(t, si.Cluster.Name, "leaf")
+	require_NotEqual(t, si.Cluster.Leader, "")
```
derekcollison, line 1274: _EMPTY_,or actually noLeader is more correct.

derekcollison (approved):
LGTM - One minor nit.

server/jetstream_leafnode_test.go
neilalexander: Fixed!

PR #5517, opened 2024-06-11
Re-order struct fields in `stree` for improved memory alignment

neilalexander:
Saves:
* 16 bytes per `leaf`
* 24 bytes per `node4`, `node16` and `node256`


derekcollison (approved):
LGTM! - Nice.

Issue #5473, opened 2024-05-29
Message duplication with leafnode connection


pcsegal:
### Observed behavior

This issue may be related to #4367.

The observed behavior is that message duplication occurs in a specific situation. I've attached an example reproducing it, below under "Steps to reproduce".

Here is a summary of the setup:

* There is one NATS cluster, **cluster A**, and a standalone NATS server, **server B**.
* In cluster A, we have multiple customer accounts; here, to simplify, we only have one customer account, `Tenant1`.
* Also in cluster A, we have an account `system-a` that imports subjects from all customer accounts, adding a prefix to distinguish each tenant. In this case, subjects from `Tenant1` are imported with the `tenant1` prefix.
* Account `system-a` in Cluster A connects to the same-named `system-a` account in Server B via a leafnode connection.
* In Server B, account `system-b` can then import the `tenant1`-prefixed subjects from the `system-a` account.

The problem happens when events are sent from `Tenant1` in cluster A to `system-b` in Cluster B, but only in a specific situation: if, in cluster B's `system-b` account, I have more than one subscriber, with one subscriber using a queue name, and other subscribers not using a queue name, then the subscriber that does **not** use a queue name receives a duplicate, while the subscriber without a queue name doesn't see any duplicates.

If I change this example so that all the subscribers use a queue name, or none of the subscribers use a queue name, the duplication doesn't happen.

### Expected behavior

No message duplication should occur.

### Server and client version

Client: 0.1.1.
Server: v2.10.14.

### Host environment

Ubuntu 20.04.

### Steps to reproduce

The ZIP file below contains a reproducible example:

[nats-ln-issue.zip](https://github.com/nats-io/nats-server/files/15487343/nats-ln-issue.zip)

The `test.sh` script simulates the behavior that causes the problem. Note that it runs two subscribers in Cluster B, both subscribing to `tenant1`-prefixed subjects, but one uses a queue name and the other does not.

The output shows the duplication:

```
[#1] Received on "tenant1.system-a.events.objectnotfound"
eventA


[#2] Received on "tenant1.system-a.events.objectnotfound"
eventA
```

derekcollison:
I would not recommend export both a stream and a service with the same prefix as you have done above.

pcsegal:
I tested removing the service import and export and leaving only the stream, but, unfortunately, I still see duplicates.

dcruzin:
[...]

PR #5519, opened 2024-06-11
[FIXED] Only forward shadow subs if we are not a leafnode or a hub leafnode client

derekcollison:
Resolves: #5473 


neilalexander (approved):
LGTM

PR #5521, opened 2024-06-11
[FIXED] Proper tiered accounting for consumers with a different tier than parent stream.

derekcollison:
We were not taking into account that consumers could be a different tier for limits or tracking usage.


wallyqs (approved):
LGTM

PR #5522, opened 2024-06-11
[FIXED] Fix for segfault if mset.mirror was nil when calculating last

derekcollison:


wallyqs (approved):
LGTM

PR #5528, opened 2024-06-13
[FIXED] Make sure jsz reporting of filtered consumers delivered stream sequence is consistent.

derekcollison:
For a leader, it will skip msgs at the end to eof and remember o.sseq for the next getNextMsg call.
But this will report stream delivered different for the leader vs the followers.

Also if a consumer leader processed an ack ahead of delivered it would sync, but followers would not.
This makes the behavior consistent between leaders and followers.


neilalexander (approved):
LGTM

PR #5532, opened 2024-06-13
Fix linear scan in `LoadNextMsg` with partial per-subject info state

neilalexander:
If the per-subject info didn't have a correct first block (which is possible as it is populated lazily), then `LoadNextMsg` could end up making a very expensive call through to `numFilteredPending` which would do a linear scan of potentially many blocks.

### DIFF
--- a/server/filestore.go
+++ b/server/filestore.go
@@ -6378,17 +6378,29 @@ func (fs *fileStore) LoadNextMsg(filter string, wc bool, start uint64, sm *Store
 				// Similar to above if start <= first seq.
 				// TODO(dlc) - For v2 track these by filter subject since they will represent filtered consumers.
 				if i == bi {
-					var ss SimpleState
-					fs.numFilteredPending(filter, &ss)
-					// Nothing available.
-					if ss.Msgs == 0 {
-						return nil, fs.state.LastSeq, ErrStoreEOF
+					var fblk uint32
+					if subjectHasWildcard(filter) {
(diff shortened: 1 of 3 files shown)


server/filestore.go
```diff
@@ -6378,17 +6378,25 @@ func (fs *fileStore) LoadNextMsg(filter string, wc bool, start uint64, sm *Store
...
-					if ss.Msgs == 0 {
-						return nil, fs.state.LastSeq, ErrStoreEOF
-					}
-					// See if we can jump ahead here.
-					// Right now we can only spin on first, so if we have interior sparseness need to favor checking per block fss if loaded.
-					// For v2 will track all blocks that have matches for psim.
-					if nbi, _ := fs.selectMsgBlockWithIndex(ss.First); nbi > i {
-						i = nbi - 1 // For the iterator condition i++
+					var fblk uint32
+					fs.psim.Match(stringToBytes(filter), func(_ []byte, psi *psi) {
```
derekcollison, line 6382: This is ok but we only have to do this if a wildcard, if filter is a literal we can do direct lookup.

Also is 0 a potential valid block index?

neilalexander: Pushed an update. Also file block indices seem to start at 1:
```
func (fs *fileStore) newMsgBlockForWrite() (*msgBlock, error) {
	index := uint32(1)
```

derekcollison (approved):
LGTM

PR #5533, opened 2024-06-13
[FIXED] Align ack gap fill from leaders to followers.

derekcollison:
Make these consistent from leaders and followers regardless of store type. Fixed a bug that would have followers stores inherit redelivered consumer delivered sequence which would break ack gap fill.


wallyqs (approved):
LGTM

PR #5535, opened 2024-06-14
Fix using goreleaser v1, fetch without installer script

wallyqs:
The new installer depends on a recent version of `cosign` being installed so just fetching the v1 release manually for now.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -55,7 +55,7 @@ script: ./scripts/runTestsOnTravis.sh $TEST_SUITE
 deploy:
   provider: script
   cleanup: true
-  script: curl -sfL https://goreleaser.com/static/run | VERSION=v1.26.2 bash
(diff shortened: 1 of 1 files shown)


neilalexander (approved):
LGTM

derekcollison:
Is their risk for us in terms of updating cosign?

philpennock:
Risk minimal: in the old installer, we curled shell script and piped it straight to bash.

Without cosign, if the artifact to be fetched has been tampered with, we won't have audit-trails of the mismatch.  But, the place where the artifact could be tampered with will be the same place that the bash script came from, so the "way which protects against supply chain attacks" in this scenario is the way where the attacker would just remove the calls to cosign and replace them with echo statements to emit a false audit trail.

For cosign to give benefit, we'd need to be downloading the artifact-to-be-verified (goreleaser) and performing the verification from code under our control, so that we know cosign is being invoked, and we can specify the properties we're looking for to accept a signature as authentic.  Without that, it's all theater.

So I think this change is fine.

PR #5537, opened 2024-06-14
Revert #5532

neilalexander:
This PR reverts #5532 as it did not solve the problem I expected it to solve and has also surfaced another performance regression visible in `Benchmark_FileStoreLoadNextMsgNoMsgsNotFirstSeq`.


derekcollison (approved):
LGTM

PR #5538, opened 2024-06-14
[IMPROVED] Update dependencies

derekcollison:
Note that upgrading is fickle at the moment since go1.22 toolchain is not backward compatible. So if you upgrade using it the CI/CD tests for go1.21 would fail. So I did a manual backdrop to go1.21 in the go.mod and did not run go mod tidy.


wallyqs (approved):
LGTM

neilalexander:
I think you can do `go mod tidy -go 1.21` if you want to pin it to an earlier Go version but still tidy.

derekcollison:
Ah cool! Thanks @neilalexander

PR #5543, opened 2024-06-15
[FIXED] ProposeDirect would bypass internal prop ipq which could cause interleaving of state.

derekcollison:
With skipMsgs calling ProposeDirect, if the prop queue was not empty this would place the skip entry ahead of the other entries. This would result in sequence mismatch and stalled processing.

Bug was we should not bypass the internal prop ipq for the nrg. This corrects it and renames to ProposeMulti.


wallyqs (approved):
LGTM

Issue #5320, opened 2024-04-17
Allow variables to be used as part of a value in configuration


cintolas:
### Proposed change

Currently the following configuration will produce authorization errors when trying to distribute this configuration to multiple systems

```
cluster {
  # set the variable token
  TOKEN: abc

  authorization {
      user: user
      password: $TOKEN
  }

  # use the token variable in a field. The password is evaluated as $TOKEN instead of abc
  nats://user:$TOKEN@server.example.com:6222,
  nats://user:$TOKEN@server2.example.com:6222
}
```
### ### Proposal
Allow a way for variables to be interpolated in right hand side of strings 

### Use case

When distributing configuration across multiple systems via an NFS mount, it makes it easier to map configuration across variables.  

Since authentication in a cluster is done in the URL, you must explicitly define the password in the URL.  This also disables the ability to define authentication tokens in environment variables

TomasVojacek:
https://docs.nats.io/running-a-nats-service/configuration#variables

PR #5544, opened 2024-06-15
Allow variables to be used as part of a value in configuration.

derekcollison:
Resolves: #5320

### DIFF
--- a/conf/parse_test.go
+++ b/conf/parse_test.go
@@ -143,6 +143,32 @@ func TestEnvVariableStringStartingWithNumberUsingQuotes(t *testing.T) {
 	test(t, fmt.Sprintf("foo = $%s", evar), ex)
 }
 
+func TestEnvVariableURL(t *testing.T) {
+	cluster := `
+		cluster {
+			# set the variable token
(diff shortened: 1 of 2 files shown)


conf/parse_test.go
```diff
@@ -143,6 +143,32 @@ func TestEnvVariableStringStartingWithNumberUsingQuotes(t *testing.T) {
...
 }
 
+func TestEnvVariableURL(t *testing.T) {
+	cluster := `
+		cluster {
+			# set the variable token
+			TOKEN: abc
+			authorization {
+				user: user
+				password: $TOKEN
```
wallyqs, line 153: one side effect from this change is that now env vars within quotes would also work, so these would be the same now:

```
password: $TOKEN
password: "$TOKEN"
```

thinking that there might be some edge cases which could be introduced potentially where an env var is replaced by accident on a string, for example if `SYS` for some reason is defined this could be replaced in permissions:

```
allow = ["$SYS"]
```

derekcollison: Let me think about some of those, good points.

PR #5546, opened 2024-06-15
[IMPROVED] Make sure to use psim on loadLast with a wildcarded subject.

derekcollison:


wallyqs (approved):
LGTM

PR #5548, opened 2024-06-16
Formalize behavior for creating audit streams for reserved subjects.

derekcollison:
For subjects `$JS`, `$JS.API`, `$JSC` and `$SYS` subjects, If NoAck is true they are now allowed, otherwise they will not be allowed.

This allows proper setup of audit streams for production use cases.


wallyqs (approved):
LGTM

PR #5549, opened 2024-06-17
[FIXED] On failed snapshot restore make sure healthz is clean.

derekcollison:
On a bad snapshot restore, e.g. R3 on single server, make sure to remove the restored stream since otherwise it would cause bad healthz reports.


wallyqs (approved):
LGTM

PR #5555, opened 2024-06-17
Allow callout users to be revoked

### LINKED ISSUE #5477: Decentralized auth callout: Connections of revoked users are not closed immediately
### Observed behavior

We are using a decentralized auth callout handler [similar to the nats-by-example service](https://github.com/ConnectEverything/nats-by-example/blob/main/examples/auth/callout-decentralized/cli/service/main.go).

Following relevant accounts & users are in the system:

- `service-account` (auth-callout encryption target)
    - `sentinel-user`
- `customer-account`
    - `customer-user`

Customer accounts like `customer-account` are created in an automated manner. The signing keys for these accounts are stored in a KV for the auth callout handling service.

NATS clients are connecting to the NATS server using the `sentinel-user` and token info including the account ID like `customer-account`. The NATS server is requesting an auth callout to the handling service. The auth callout handling service is creating a `customer-user` and allowing the initial connection request.

Now `customer-user` is being added to the revocations list of `customer-account`. The `customer-account` JWT having `customer-user` in it's revocation list is then pushed to the NATS server. The initial connection from the NATS client remains open.

### Expected behavior

As with statically created users, it is expected that once the updated account JWT is pushed to the NATS server, the open connection from a user in the accounts revocation list is immediately closed ([as documented here for example](https://docs.nats.io/using-nats/nats-tools/nsc/revocation#pushing-the-changes-to-the-nats-servers)).

### Server and client version

Server v2.10.16
CLI Client v0.1.4 and nats.go v1.35.0

### Host environment

_No response_

### Steps to reproduce

_No response_

aricart:
[FIX] added logic to enable users generated by a callout to be revoked - the callout is responsible to map the user ids for revocation purposes.

Fix #5477

### DIFF
--- a/server/accounts.go
+++ b/server/accounts.go
@@ -3688,8 +3688,8 @@ func (s *Server) updateAccountClaimsWithRefresh(a *Account, ac *jwt.AccountClaim
 		c.applyAccountLimits()
 		// if we have an nkey user we are a callout user - save
 		// the issuedAt, and nkey user id to honor revocations
-		nkeyUserID := ""
-		issuedAt := int64(0)
+		var nkeyUserID string
+		var issuedAt int64
 		if c.user != nil {
 			issuedAt = c.user.Issued
 			nkeyUserID = c.user.Nkey
@@ -3713,13 +3713,13 @@ func (s *Server) updateAccountClaimsWithRefresh(a *Account, ac *jwt.AccountClaim
 			}
 		}
 
-		// if we extracted nkeyUserID and issuedAt we are a callout user
-		// calloutIAT should only be set if we are in callout type scenario
+		// if we extracted nkeyUserID and issuedAt we are a callout type
+		// calloutIAT should only be set if we are in callout scenario as
 		// the user JWT is _NOT_ associated with the client for callouts,
 		// so we rely on the calloutIAT to know when the JWT was issued
 		// revocations simply state that JWT issued before or by that date
 		// are not valid
-		if ac.Revocations != nil && nkeyUserID != "" && issuedAt > 0 {
+		if ac.Revocations != nil && nkeyUserID != _EMPTY_ && issuedAt > 0 {
 			seconds, ok := ac.Revocations[jwt.All]
 			if ok && seconds >= issuedAt {
 				c.sendErrAndDebug("User Authentication Revoked")
--- a/server/accounts.go
+++ b/server/accounts.go
@@ -3686,6 +3686,14 @@ func (s *Server) updateAccountClaimsWithRefresh(a *Account, ac *jwt.AccountClaim
 		}
 		c.mu.Lock()
 		c.applyAccountLimits()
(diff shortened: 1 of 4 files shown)


server/accounts.go
```diff
@@ -3686,6 +3686,14 @@ func (s *Server) updateAccountClaimsWithRefresh(a *Account, ac *jwt.AccountClaim
 		}
 		c.mu.Lock()
 		c.applyAccountLimits()
+		// if we have an nkey user we are a callout user - save
+		// the issuedAt, and nkey user id to honor revocations
+		nkeyUserID := ""
```
derekcollison, line 3691: var nkeyUserID string

server/accounts.go
```diff
@@ -3686,6 +3686,14 @@ func (s *Server) updateAccountClaimsWithRefresh(a *Account, ac *jwt.AccountClaim
...
+		issuedAt := int64(0)
```
derekcollison, line 3692: vat issuedAt uint64

server/accounts.go
```diff
@@ -3704,6 +3712,27 @@ func (s *Server) updateAccountClaimsWithRefresh(a *Account, ac *jwt.AccountClaim
 				continue
 			}
 		}
+
+		// if we extracted nkeyUserID and issuedAt we are a callout user
+		// calloutIAT should only be set if we are in callout type scenario
```
derekcollison, line 3717: in a callout type..

server/accounts.go
```diff
@@ -3704,6 +3712,27 @@ func (s *Server) updateAccountClaimsWithRefresh(a *Account, ac *jwt.AccountClaim
...
+		// the user JWT is _NOT_ associated with the client for callouts,
+		// so we rely on the calloutIAT to know when the JWT was issued
+		// revocations simply state that JWT issued before or by that date
+		// are not valid
+		if ac.Revocations != nil && nkeyUserID != "" && issuedAt > 0 {
```
derekcollison, line 3722: nkeyUserID != _EMPTY_

server/auth.go
```diff
@@ -59,7 +59,9 @@ type ClientAuthentication interface {
 
 // NkeyUser is for multiple nkey based users
 type NkeyUser struct {
-	Nkey                   string              `json:"user"`
+	Nkey string `json:"user"`
+	// this is a copy of the issued at (iat) field in the jwt
```
derekcollison, line 63: PLace comment at end to keep spacing same.

derekcollison (changes requested):
LGTM - a few small nits.

server/accounts.go
aricart: Fixed

server/accounts.go
aricart: Fixed

server/auth.go
aricart: Fixed

derekcollison (approved):
LGTM

PR #5558, opened 2024-06-18
When installing a new snapshot remove last inline if present.

derekcollison:
Previously (but not in released code) we were cleaning up all old ones in a go routine. However this could cause multiple go routines to race and delete the wrong snapshot leaving none available.


neilalexander (approved):
LGTM

derekcollison:
Had a small bug, but tracked it down. If green will merge after this run.

Issue #5471, opened 2024-05-27
MQTT Memory Leaks on consumer connections


slice-srinidhis:
### Observed behavior

Seeing consistent memory leak when using NATS while MQTT consumers just makes connections.  The pattern is observed when we are scaling the system to have 5K connections .Though there not many messages produced or consumed .  The `consumer_inactive_threshold: 0.2s ` is also set but the memory is not getting released. The only way to cleanup is to restart the pods . 
![Screenshot 2024-04-22 at 4 45 58 PM](https://github.com/nats-io/nats-server/assets/160851861/62153849-b832-4725-9e7c-2a3d4f30d1af)
![Screenshot 2024-04-24 at 1 01 46 PM](https://github.com/nats-io/nats-server/assets/160851861/f4328765-acfa-4a9d-9ba0-f0526e332c36)
On getting the pprof , its observed createInternalClient seem to take a lot of heap memory (attaching the pprof for ref) . [
[heap_pprof_24-04.pdf](https://github.com/nats-io/nats-server/files/15458528/heap_pprof_24-04.pdf)
](url)
Kindly look into issue and let know if there are any parameter tweaking that will help resolving the issue . 

### Expected behavior

Optimal memory management without memory leaks for MQTT .

### Server and client version

Nats Server version 2.10.14 

### Host environment

Kubernetes v1.25

### Steps to reproduce

Setup NATS with MQTT , on making client connections the leak is reproducible at scale .

neilalexander:
Can you please attach the `/debug/pprof/allocs?debug=0` file itself instead of the PDF extract?

slice-srinidhis:
@neilalexander Please find the pprofs associated . 28-05-05-19 taken during the first iteration and 28-05-07-30 during the second one where the older memory is accumulated , Have also attached the memory graph for reference 
[28-05-07-30.pb.gz](https://github.com/nats-io/nats-server/files/15472759/28-05-07-30.pb.gz)

[28-05-05-19.pb.gz](https://github.com/nats-io/nats-server/files/15472757/28-05-05-19.pb.gz)

<img width="1428" alt="Screenshot 2024-05-28 at 9 31 07 PM" src="https://github.com/nats-io/nats-server/assets/160851861/e50511f0-26bd-4ab4-be94-aea17c6dda00">

levb:
[...]

PR #5566, opened 2024-06-19
[FIXED] Make sure to always remove internal clients from the account regardless of kind.

derekcollison:
Internal clients for NRG (Raft Groups) were of type SYSTEM and were not being removed from the system's account when they were closed, causing a leak of closed client connections and memory buildup.

Resolves: #5471 


neilalexander (approved):
LGTM

PR #5568, opened 2024-06-19
[IMPROVED] Increase cache invalidation times for mb.cache.buf and mb.fss.

derekcollison:
This is to avoid thrashing on sparse streams and streams that hit max msgs limits where we have to remove the first that nay be in a totally different block.


neilalexander (approved):
LGTM

Issue #5570, opened 2024-06-20
nats-server panic when create a stream with a wrong mapping destination


AetheWu:
### Observed behavior

We use nats-go sdk create a stream with below options: 
```go
jetstream.StreamConfig {
  Sources: []*jetstream.StreamSource{
			{
				Name: "mapping",
				SubjectTransforms: []jetstream.SubjectTransformConfig{
					{
						Source:     "events.*",
						Destination: "events.{{wildcard(1)}}{{split(3,1)}}",
					},
				},
			},
		}
}
```

The nats-server panic with below errors and is no longer running :
```
[94] 2024/06/20 07:57:37.684644 [ERR] Unable to get subject transform for source: invalid mapping destination: too many arguments passed to the function in {{wildcard(1)}}{{split(3,1)}}
fatal error: sync: Unlock of unlocked RWMutex
```

### Expected behavior

do not panic

### Server and client version

nats-server version: 2.10.16
nats-client: nats.go-1.36.0

### Host environment

Deploy with official helm charts

AetheWu:
In addition, how to use nats.go to create stream partitions

derekcollison:
I quickly created a test under `server/jetstream_test.go` and added it and it does not fail, but does provide an error. I tested on `main`, `release/v2.10.17` and `release/v2.10.16` branches.

```go
// https://github.com/nats-io/nats-server/issues/5570
func TestJetStreamBadSubjectMappingStream(t *testing.T) {
	s := RunBasicJetStreamServer(t)
	defer s.Shutdown()

	// Client for API requests.
	nc, js := jsClientConnect(t, s)
	defer nc.Close()

	_, err := js.AddStream(&nats.StreamConfig{
		Name: "test",
		Sources: []*nats.StreamSource{
			{
				Name: "mapping",
				SubjectTransforms: []nats.SubjectTransformConfig{
					{
						Source:      "events.*",
						Destination: "events.{{wildcard(1)}}{{split(3,1)}}",
					},
				},
			},
		},
	})
	require_NoError(t, err)
}
```

derekcollison:
Although I do see a logic error @jnmoyne that you unlock the stream but do not return, so I can see where this would fail but the example above does not recreate it that error.

derekcollison:
ok key is it has to be an update of an existing stream, not new one. I see it.

jnmoyne:
[...]

PR #5571, opened 2024-06-20
[FIXED] Do not unlock here, will panic server.

derekcollison:
Resolves: #5570 


neilalexander (approved):
LGTM

PR #5573, opened 2024-06-20
Reduce allocations in `isMatch` in filestore/memstore

neilalexander:
Otherwise long subjects could cause us to make the same reallocations on each call to `tokenizeSubjectIntoSlice` without reusing the newly sized underlying array for the next call.


derekcollison (approved):
LGTM

PR #5574, opened 2024-06-20
Return error when trying to update a stream with a source with a bad subject transform

jnmoyne:
In some cases a bad subject transform could still pass the validation test that is part of the config check when updating a stream config, and no error would be returned. Now return an error when trying to update a stream config with a source that has a bad subject transform that is not caught by the subject transform destination check in the stream config check.

Improves #5571

### DIFF
--- a/server/stream.go
+++ b/server/stream.go
@@ -1846,7 +1846,7 @@ func (mset *stream) updateWithAdvisory(config *StreamConfig, sendAdvisory bool) 
 							var err error
 							si.trs[i], err = NewSubjectTransform(s.SubjectTransforms[i].Source, s.SubjectTransforms[i].Destination)
 							if err != nil {
-								mset.srv.Errorf("Unable to get subject transform for source: %v", err)
+								return fmt.Errorf("unable to get subject transform for source: %v", err)
 							}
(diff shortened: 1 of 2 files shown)


derekcollison:
Need to get tests passing cleanly first.

server/stream.go
```diff
@@ -1846,7 +1846,7 @@ func (mset *stream) updateWithAdvisory(config *StreamConfig, sendAdvisory bool)
 							var err error
 							si.trs[i], err = NewSubjectTransform(s.SubjectTransforms[i].Source, s.SubjectTransforms[i].Destination)
 							if err != nil {
-								mset.srv.Errorf("Unable to get subject transform for source: %v", err)
+								return fmt.Errorf("unable to get subject transform for source: %v", err)
```
derekcollison, line 1849: You will need to unlock here before returning I would imagine.

jnmoyne: Yes, that's the reason the test was hanging on the server shutdown. Added it (back, since it was removed in #5571, we're getting there in the end :)).

derekcollison (approved):
LGTM

PR #5575, opened 2024-06-20
[CHANGED] MQTT s.clear() do not wait for JS responses when disconnecting the session

levb:
MQTT s.clear(): do not wait for JS responses when disconnecting the session

Related to https://github.com/nats-io/nats-server/issues/5471

Previously we were making `jsa.NewRequest` as it is needed when connecting a clean session. On disconnect, there is no reason to wait for the response (and tie up the MQTT read loop of the client).

This should specifically help situations when a client app with many MQTT connections and QOS subscriptions disconnects suddenly, causing a flood of JSAPI deleteConsumer requests.

Test: n/a, not sure how to instrument for it.

### DIFF
--- a/server/mqtt.go
+++ b/server/mqtt.go
@@ -1563,6 +1563,10 @@ func (jsa *mqttJSA) newRequestEx(kind, subject, cidHash string, hdr int, msg []b
 	if err != nil {
 		return nil, err
 	}
+	if timeout == 0{
+		return nil, err
+	}
+	
 	if len(responses) != 1 {
 		return nil, fmt.Errorf("unreachable: invalid number of responses (%d)", len(responses))
 	}
@@ -1718,6 +1722,9 @@ func (jsa *mqttJSA) deleteConsumer(streamName, consName string, noWait bool) (*J
 	cdri, err := jsa.newRequestEx(mqttJSAConsumerDel, subj, _EMPTY_, -1, nil, timeout)
 	if err != nil {
 		return nil, err
+	}
+	if noWait {
+		return nil, nil
 	}
 	cdr := cdri.(*JSApiConsumerDeleteResponse)
 	return cdr, cdr.ToError()
--- a/server/mqtt.go
+++ b/server/mqtt.go
@@ -1713,9 +1713,6 @@ func (jsa *mqttJSA) deleteConsumer(streamName, consName string, noWait bool) (*J
 	if err != nil {
 		return nil, err
 	}
-	if noWait {
-		return nil, nil
-	}
 	cdr := cdri.(*JSApiConsumerDeleteResponse)
 	return cdr, cdr.ToError()
 }
--- a/server/mqtt.go
+++ b/server/mqtt.go
@@ -1563,10 +1563,6 @@ func (jsa *mqttJSA) newRequestEx(kind, subject, cidHash string, hdr int, msg []b
 	if err != nil {
 		return nil, err
 	}
-	if timeout == 0{
-		return nil, err
-	}
-	
 	if len(responses) != 1 {
 		return nil, fmt.Errorf("unreachable: invalid number of responses (%d)", len(responses))
 	}
@@ -1624,10 +1620,6 @@ func (jsa *mqttJSA) newRequestExMulti(kind, subject, cidHash string, hdrs []int,
 			msg:   msg,
 		})
 		r2i[reply] = i
-	}
-
-	if timeout == 0 {
-		return nil, nil
 	}
 
 	// Wait for all responses to come back, or for the timeout to expire. We
@@ -1712,14 +1704,12 @@ func (jsa *mqttJSA) createDurableConsumer(cfg *CreateConsumerRequest) (*JSApiCon
 // if noWait is specified, does not wait for the JS response, returns nil
 func (jsa *mqttJSA) deleteConsumer(streamName, consName string, noWait bool) (*JSApiConsumerDeleteResponse, error) {
 	subj := fmt.Sprintf(JSApiConsumerDeleteT, streamName, consName)
-
-	timeout := mqttJSAPITimeout
 	if noWait {
-		timeout = 0
+		jsa.sendMsg(subj, nil)
+		return nil, nil
 	}
 
-	// timeout == 0 signals that we don't want to wait for the response.
-	cdri, err := jsa.newRequestEx(mqttJSAConsumerDel, subj, _EMPTY_, -1, nil, timeout)
+	cdri, err := jsa.newRequestEx(mqttJSAConsumerDel, subj, _EMPTY_, -1, nil, mqttJSAPITimeout)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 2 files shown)


server/mqtt.go
```diff
@@ -1000,7 +1000,7 @@ func (s *Server) mqttHandleClosedClient(c *client) {
 
 	// This needs to be done outside of any lock.
 	if doClean {
-		if err := sess.clear(); err != nil {
+		if err := sess.clear(true); err != nil {
```
kozlovic, line 1003: This will be invoked for "clean" sessions, which means sessions which state should be discarded and not reused in any subsequent session. Could there be a situation where a create/close/create causes an issue since some assets may not have been completely removed from the JS server?

server/mqtt.go
```diff
@@ -3823,7 +3829,7 @@ CHECK:
 			// This Session lasts as long as the Network Connection. State data
 			// associated with this Session MUST NOT be reused in any subsequent
 			// Session.
-			if err := es.clear(); err != nil {
+			if err := es.clear(false); err != nil {
```
kozlovic, line 3832: At least here we make sure that the delete is complete, which I think is the right thing to do.

kozlovic (changes requested):
The test `TestMQTTSessionNotDeletedOnDeleteConsumerError` fails because it was expecting an error that is no longer reported due to this change.

server/mqtt.go
levb: We don't do anything with the error other than logging it anyway, so the logic doesn't change. Also, if the session is re-created as clean, we'd clean out its state on CONNECT. If it is re-created as persistent... 0/5 we could wipe out and re-create the consumers _if_ there was a session message; but if we failed to delete the session message on the "clean" DISCONNECT... well, it still would not change the existing logic, would it?

server/mqtt.go
levb: Right. I understand why checking the success is important here, thus preserved as before.

server/mqtt.go
```diff
@@ -3484,7 +3496,7 @@ func (sess *mqttSession) untrackPubRel(pi uint16) (jsAckSubject string) {
 func (sess *mqttSession) deleteConsumer(cc *ConsumerConfig) {
 	sess.mu.Lock()
 	sess.tmaxack -= cc.MaxAckPending
-	sess.jsa.sendq.push(&mqttJSPubMsg{subj: sess.jsa.prefixDomain(fmt.Sprintf(JSApiConsumerDeleteT, mqttStreamName, cc.Durable))})
+	sess.jsa.deleteConsumer(mqttStreamName, cc.Durable, true)
```
levb, line 3499: Added for code re-use. Also, when I first implemented the noWait-deleteConsumer similarly to this, it started failing `TestMQTTSessionNotDeletedOnDeleteConsumerError` because the expected error (can't delete the consumer) was missing in the server log. Using the idiomatic `NewRequest` restored the error, so I figured it's a better way of sending the request.

server/mqtt.go
```diff
@@ -1701,9 +1705,17 @@ func (jsa *mqttJSA) createDurableConsumer(cfg *CreateConsumerRequest) (*JSApiCon
 	return ccr, ccr.ToError()
 }
 
-func (jsa *mqttJSA) deleteConsumer(streamName, consName string) (*JSApiConsumerDeleteResponse, error) {
+// if noWait is specified, does not wait for the JS response, returns nil
+func (jsa *mqttJSA) deleteConsumer(streamName, consName string, noWait bool) (*JSApiConsumerDeleteResponse, error) {
 	subj := fmt.Sprintf(JSApiConsumerDeleteT, streamName, consName)
-	cdri, err := jsa.newRequest(mqttJSAConsumerDel, subj, 0, nil)
+
+	timeout := mqttJSAPITimeout
```
levb, line 1712: I checked that in all other places we use an explicit value. 1/5 this is intention-revealing enough, no need for another flag.

server/mqtt.go
```diff
@@ -1622,6 +1622,10 @@ func (jsa *mqttJSA) newRequestExMulti(kind, subject, cidHash string, hdrs []int,
 		r2i[reply] = i
 	}
 
+	if timeout == 0 {
```
levb, line 1625: i.e. "no wait"

kozlovic: Humm... not really. That would cause `newRequestEx` to return `fmt.Errorf("unreachable: invalid number of responses (%d)", len(responses))`, which is not what we want here...

server/mqtt.go
```diff
@@ -1701,9 +1705,17 @@ func (jsa *mqttJSA) createDurableConsumer(cfg *CreateConsumerRequest) (*JSApiCon
...
 	subj := fmt.Sprintf(JSApiConsumerDeleteT, streamName, consName)
-	cdri, err := jsa.newRequest(mqttJSAConsumerDel, subj, 0, nil)
+
+	timeout := mqttJSAPITimeout
+	if noWait {
+		timeout = 0
+	}
+
+	// timeout == 0 signals that we don't want to wait for the response.
+	cdri, err := jsa.newRequestEx(mqttJSAConsumerDel, subj, _EMPTY_, -1, nil, timeout)
```
kozlovic, line 1718: This would return "invalid number of responses", which is not quite good.

kozlovic (changes requested):

server/mqtt.go
kozlovic: In other words, the original approach may have been ok, you would have simply needed to remove the test that was no longer relevant...

levb:
@kozlovic Sorry, I misunderstood the failing test. I thought the error in the log was supposed to come from JetStream itself, where it is actually the very MQTT error I was skipped that the test was expecting.

I need a little time to re-think how to do this, there can still be errors sending the request to delete the consumer, so the test is still valid.

levb:
@kozlovic here's my thinking.

0/5 for this PR we can keep the disconnect cleanup as opportunistic, and remove the failing `TestMQTTSessionNotDeletedOnDeleteConsumerError`. There really aren't any plausible errors in the "no wait" case, we just push a message onto the jsapi queue. So we will always attempt to delete the session message, just like before this PR.

Longer term:
1. For clean sessions, can we make subscription consumers ephemeral, with a short inactive threshold, so that they get auto-collected if we fail to delete?
2. For all new sessions, even clean, try to load the existing session message first. If the previous persisted session was clean, then synchronously delete its subs' consumers, we know the names. Ditto the pubrel consumer.
3. PubRel consumers can be consolidated, like the retained messages into one, account-specific consumer. There’s no need to match the messages to subs, just to the sessions, by client ID.

What do you think?

server/mqtt.go
levb: Right. Apologies for the confusion.

kozlovic (approved):
LGTM

kozlovic:
> Longer term:
> For clean sessions, can we make subscription consumers ephemeral, with a short inactive threshold, so that they get auto-collected if we fail to delete?

But is not an ephemeral consumer supposed to be told that the "client" side is still active through pings? Meaning that would that mean that the MQTT code would then need to be more involved in maintaining this consumer? And if so, it means that if there was a disconnect between where the MQTT sub lives and the JS consumer leader, we could end-up with the consumer being deleted while the MQTT code does not know that this is the case and end-up with a stop in message delivery?

levb:
@derekcollison This can be merged now.

PR #5578, opened 2024-06-21
[FIXED] Fixed bug that would return no msg found for loadLast.

derekcollison:
Bug was that new wildcard matching code would not reverse start and stop, which is needed since we walk backwards. Since they were not flipped as soon as start != stop, it would not enter the loop condition, e.g. start:1, stop:2 fails start >= stop.

Only existed in 2.10.17-RC4 and beyond, not in previous releases.


neilalexander (approved):
LGTM

wallyqs (approved):
LGTM

PR #5584, opened 2024-06-22
[IMPROVED] Avoid loading last blks on LoadNextMsg() miss.

derekcollison:
Make sure on a miss from a starting sequence that if no other msgs exists we avoid loading and blocks.
Match logic for interior skips.

### DIFF
--- a/server/filestore.go
+++ b/server/filestore.go
@@ -2596,28 +2596,38 @@ func (fs *fileStore) FilteredState(sseq uint64, subj string) SimpleState {
 
 // This is used to see if we can selectively jump start blocks based on filter subject and a floor block index.
 // Will return -1 if no matches at all.
-func (fs *fileStore) checkSkipFirstBlock(filter string, wc bool) int {
-	start := uint32(math.MaxUint32)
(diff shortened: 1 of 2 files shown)


server/filestore.go
```diff
@@ -2596,28 +2596,38 @@ func (fs *fileStore) FilteredState(sseq uint64, subj string) SimpleState {
...
-	// Here we need to translate this to index into fs.blks.
+	// Here we need to translate this to index into fs.blks properly.
 	mb := fs.bim[start]
 	if mb == nil {
-		return -1
+		return -1, -1
 	}
-	bi, _ := fs.selectMsgBlockWithIndex(atomic.LoadUint64(&mb.last.seq))
-	return bi
+	fi, _ := fs.selectMsgBlockWithIndex(atomic.LoadUint64(&mb.last.seq))
```
neilalexander, line 2622: Should this be `first.seq`?

derekcollison: Block selection is inclusive of first and last, so either would work.

wallyqs (approved):
LGTM

neilalexander (approved):
LGTM

PR #5585, opened 2024-06-24
Add `node48` to stree

neilalexander:
A `node256` is nearly 4KB in memory whereas a `node48` is closer to 1KB, so there is a potential for anywhere up to 75% memory savings in some subject spaces.

Implemented as per the ART paper, although with the caveat that our pointers are actually 16 bytes (as `interface{}`s) rather than 8 bytes as the paper assumes:

> Node48: As the number of entries in a node increases, searching the key array becomes expensive. Therefore, nodes with more than 16 pointers do not store the keys explicitly. Instead, a 256-element array is used, which can be indexed with key bytes directly. If a node has between 17 and 48 child pointers, this array stores indexes into a second array which contains up to 48 pointers. This indirection saves space in comparison to 256 pointers of 8 bytes, because the indexes only require 6 bits (we use 1 byte for simplicity).

### DIFF
--- a/server/stree/node48.go
+++ b/server/stree/node48.go
@@ -1,0 +1,110 @@
+// Copyright 2023-2024 The NATS Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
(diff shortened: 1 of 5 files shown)


server/stree/node48.go
```diff
@@ -0,0 +1,110 @@
...
+
+// Node with 48 children
+// Memory saving vs node256 comes from the fact that the child array is 16 bytes
+// per `node` entry, so node256's 256*16=4096 vs node48's 256+(48*16)=1024
+// Note that `key` is effectively 1-indexed, as 0 means no entry, so offset by 1
+// Order of struct fields for best memory alignment (as per govet/fieldalignment)
+type node48 struct {
+	child [48]node
+	meta
+	key [256]byte
```
derekcollison, line 24: Interesting approach, most of the complexity for a node48 comes from the key handling, this avoids that for a slight increase per node size but with still good savings from child array.

derekcollison (approved):
LGTM

PR #5586, opened 2024-06-24
NRG: Fix leaving observer state when applies are paused

neilalexander:
Previously, if the applies were paused (i.e. due to an on-going catchup) then calls to `SetObserver(...)` from other places, i.e. `jetstream_cluster_migrate`, would not leave observer mode after applies were resumed. This could result in assets getting stuck in observer mode.

### DIFF
--- a/server/raft_test.go
+++ b/server/raft_test.go
@@ -549,3 +549,42 @@ func TestNRGSystemClientCleanupFromAccount(t *testing.T) {
 	finish := numClients()
 	require_Equal(t, start, finish)
 }
+
+func TestNRGLeavesObserverAfterPause(t *testing.T) {
+	c := createJetStreamClusterExplicit(t, "R3S", 3)
+	defer c.shutdown()
+
+	rg := c.createMemRaftGroup("TEST", 3, newStateAdder)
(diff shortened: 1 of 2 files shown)


server/raft_test.go
```diff
@@ -549,3 +549,41 @@ func TestNRGSystemClientCleanupFromAccount(t *testing.T) {
...
+
+	rg := c.createMemRaftGroup("TEST", 3, newStateAdder)
+	rg.waitOnLeader()
+
+	n := rg[0].node().(*raft)
+
+	checkState := func(observer, pobserver bool) {
+		n.RLock()
+		defer n.RUnlock()
+		require_Equal(t, n.observer, observer)
```
derekcollison, line 565: These fail will block tests since lock still held.

derekcollison (approved):
LGTM

server/raft_test.go
neilalexander: It does block if you do:
```
	n.RLock()
	require_Equal(t, n.observer, observer)
	require_Equal(t, n.pobserver, pobserver)
	n.RUnlock()
```
... but not if you use `defer` as the PR does.

PR #5589, opened 2024-06-25
Fix imports not being available for a client sometimes after a server restart

wallyqs:
When a client would reconnect to a server that was still setting up the imports and exports a client that reconnected too soon to a server that had just been restarted might be missing some of the imports that were defined in its JWT.


derekcollison (approved):
LGTM

PR #5597, opened 2024-06-26
Fix logging when exceeding JetStream account limits

### LINKED ISSUE #5594: Improve WRN error message resource limits exceeded for account (10002) text
### Proposed change

Make WRN error message resource limits exceeded for account (10002) showing in which account there is an issue with resources.

### Use case

`2024-06-26 10:30:45.565 | [1100] 2024/06/26 08:30:45.465496 [WRN] resource limits exceeded for account (10002)`

There should a way to have account id or name here. For big number of accounts  (like 50+) it can be challenging to  loop trough every single account and check all limits to know what is actually happening.

### Contribution

_No response_

neilalexander:
Fixes #5594. Otherwise it prints a very unhelpful error that doesn't tell you which account is at fault.

### DIFF
--- a/server/jetstream_cluster.go
+++ b/server/jetstream_cluster.go
@@ -7793,7 +7793,7 @@ func (mset *stream) processClusteredInboundMsg(subject, reply string, hdr, msg [
 		if err == nil {
 			err = NewJSAccountResourcesExceededError()
 		}
-		s.RateLimitWarnf(err.Error())
+		s.RateLimitWarnf("JetStream account limits exceeded for '%s': %s", jsa.acc().GetName(), err.Error())
 		if canRespond {
 			var resp = &JSPubAckResponse{PubAck: &PubAck{Stream: name}}
 			resp.Error = err


wallyqs (approved):
LGTM

server/jetstream_cluster.go
```diff
@@ -7793,7 +7793,7 @@ func (mset *stream) processClusteredInboundMsg(subject, reply string, hdr, msg [
 		if err == nil {
 			err = NewJSAccountResourcesExceededError()
 		}
-		s.RateLimitWarnf(err.Error())
+		s.RateLimitWarnf("JetStream account limits exceeded for '%s': %s", jsa.acc().Name, err.Error())
```
derekcollison, line 7796: Would use GetName(), Does locking and also handles nil properly.

derekcollison (changes requested):

server/jetstream_cluster.go
neilalexander: Done.

derekcollison (approved):
LGTM

PR #5599, opened 2024-06-26
Use the tag name for the version on release

LaurentGoderre:
Use the tag name for the version on release. This also makes the version easier to detect in scanners such as Syft. Otherwise the version that gets detected is listed as `(devel)`.

Signed-off-by: Laurent Goderre <laurent.goderre@docker.com>

### DIFF
--- a/.goreleaser-nightly.yml
+++ b/.goreleaser-nightly.yml
@@ -5,7 +5,7 @@ builds:
     id: nats-server
     binary: nats-server
     ldflags:
-      - -w -X github.com/nats-io/nats-server/v2/server.gitCommit={{.ShortCommit}}
+      - -w -X github.com/nats-io/nats-server/v2/version.GitCommit={{.ShortCommit}} -X 'github.com/nats-io/nats-server/v2/version.Version={{.Summary}}'
     env:
       - GO111MODULE=on
       - CGO_ENABLED=0
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -16,7 +16,7 @@ builds:
   flags:
   - -trimpath
   ldflags:
-    - -w -X github.com/nats-io/nats-server/v2/server.gitCommit={{.ShortCommit}}
+    - -w -X 'github.com/nats-io/nats-server/v2/server.gitCommit={{.ShortCommit}}' -X 'github.com/nats-io/nats-server/v2/server.serverVersion={{.CurrentTag}}' -X 'github.com/nats-io/nats-server/v2/version.Version={{.CurrentTag}}'
   env:
     - GO111MODULE=on
     - CGO_ENABLED=0
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -16,7 +16,7 @@ builds:
   flags:
   - -trimpath
   ldflags:
-    - -w -X 'github.com/nats-io/nats-server/v2/server.gitCommit={{.ShortCommit}}' -X 'github.com/nats-io/nats-server/v2/server.serverVersion={{.CurrentTag}}' -X 'github.com/nats-io/nats-server/v2/version.Version={{.CurrentTag}}'
+    - -w -X 'github.com/nats-io/nats-server/v2/version.GitCommit={{.ShortCommit}}' -X 'github.com/nats-io/nats-server/v2/version.Version={{.CurrentTag}}'
   env:
     - GO111MODULE=on
     - CGO_ENABLED=0
--- a/server/server.go
+++ b/server/server.go
@@ -48,6 +48,7 @@ import (
 	"github.com/nats-io/nuid"
 
 	"github.com/nats-io/nats-server/v2/logger"
+	serverVersion "github.com/nats-io/nats-server/v2/version"
 )
 
 const (
@@ -648,10 +649,6 @@ func New(opts *Options) *Server {
 func NewServer(opts *Options) (*Server, error) {
 	setBaselineOptions(opts)
 
-	if serverVersion == _EMPTY_ {
-		serverVersion = VERSION
-	}
-
 	// Process TLS options, including whether we require client certificates.
 	tlsReq := opts.TLSConfig != nil
 	verify := (tlsReq && opts.TLSConfig.ClientAuth == tls.RequireAndVerifyClientCert)
@@ -682,9 +679,9 @@ func NewServer(opts *Options) (*Server, error) {
 	info := Info{
 		ID:           pub,
 		XKey:         xpub,
-		Version:      serverVersion,
+		Version:      serverVersion.Version,
 		Proto:        PROTO,
-		GitCommit:    gitCommit,
+		GitCommit:    serverVersion.GitCommit,
 		GoVersion:    runtime.Version(),
 		Name:         serverName,
 		Host:         opts.Host,
@@ -761,7 +758,7 @@ func NewServer(opts *Options) (*Server, error) {
 		ourNode := getHash(serverName)
 		s.nodeToInfo.Store(ourNode, nodeInfo{
 			serverName,
-			VERSION,
+			serverVersion.Version,
 			opts.Cluster.Name,
 			opts.JetStreamDomain,
 			info.ID,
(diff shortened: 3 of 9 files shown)


wallyqs:
Thanks for the contribution, I think this would make syft recognize the version but we use `--trimpath` during the release and that removes the ldflags in Go at the moment due to this issue: https://github.com/golang/go/issues/63432
I think it would still show as `(devel)` after including the changes on a release (unless we stop using the `--trimpath` option)

LaurentGoderre:
The way it is done we can do `cat nats-server| grep -aPo '\x00[0-9]+\.[0-9]+\.[0-9]+\x00' to extract the version

wallyqs:
Thanks I see, ok looks like more recent syft has been able to workaround the issue of ldflags.  Confirmed locally:

```
$ go build --trimpath -ldflags="-X github.com/nats-io/nats-server/v2/server.serverVersion=2.10.17" .
$ syft nats-server
github.com/klauspost/compress      v1.17.9   go-module    
github.com/minio/highwayhash       v1.0.2    go-module    
github.com/nats-io/jwt/v2          v2.5.7    go-module    
github.com/nats-io/nats-server/v2  2.10.17   go-module    
github.com/nats-io/nkeys           v0.4.7    go-module    
go.uber.org/automaxprocs           v1.5.3    go-module    
golang.org/x/crypto                v0.24.0   go-module    
golang.org/x/time                  v0.5.0    go-module    
stdlib                             go1.22.4  go-module  
```

although  `go version` still shows as `(devel)` but `syft` found it in an alternative way now:

```
go version -m nats-server
nats-server: go1.22.4
	path	github.com/nats-io/nats-server/v2
	mod	github.com/nats-io/nats-server/v2	(devel)	
	dep	github.com/klauspost/compress	v1.17.9	h1:6KIumPrER1LHsvBVuDa0r5xaG0Es51mhhB9BQB2qeMA=
	dep	github.com/minio/highwayhash	v1.0.2	h1:Aak5U0nElisjDCfPSG79Tgzkn2gl66NxOMspRrKnA/g=
	dep	github.com/nats-io/jwt/v2	v2.5.7	h1:j5lH1fUXCnJnY8SsQeB/a/z9Azgu2bYIDvtPVNdxe2c=
	dep	github.com/nats-io/nkeys	v0.4.7	h1:RwNJbbIdYCoClSDNY7QVKZlyb/wfT6ugvFCiKy6vDvI=
	dep	github.com/nats-io/nuid	v1.0.1	h1:5iA8DT8V7q8WK2EScv2padNa/rTESc1KdnPw4TC2paw=
	dep	go.uber.org/automaxprocs	v1.5.3	h1:kWazyxZUrS3Gs4qUpbwo5kEIMGe/DAvi5Z4tl2NW4j8=
	dep	golang.org/x/crypto	v0.24.0	h1:mnl8DM0o513X8fdIkmyFE/5hTYxbwYOjDS/+rK6qpRI=
	dep	golang.org/x/time	v0.5.0	h1:o7cqy6amK/52YcAKIPlM3a+Fpj35zvRj2TP+e1xFSfk=
	build	-buildmode=exe
	build	-compiler=gc
	build	-trimpath=true
	build	DefaultGODEBUG=httplaxcontentlength=1,httpmuxgo121=1,tls10server=1,tlsrsakex=1,tlsunsafeekm=1
	build	CGO_ENABLED=1
	build	GOARCH=arm64
	build	GOOS=darwin
	build	vcs=git
	build	vcs.revision=e99bb7010424bd7446b34c307ff9b0b03e918343
	build	vcs.time=2024-06-26T20:19:13Z
	build	vcs.modified=true
```

wallyqs (approved):
LGTM, thanks!

server/server.go
```diff
@@ -648,6 +648,10 @@ func New(opts *Options) *Server {
 func NewServer(opts *Options) (*Server, error) {
 	setBaselineOptions(opts)
 
+	if serverVersion == "" {
```
derekcollison, line 651: Use _EMPTY_ here. Convention in the server.

derekcollison (changes requested):

server/server.go
```diff
@@ -1546,7 +1550,7 @@ func PrintAndDie(msg string) {
 
 // PrintServerAndExit will print our version and exit.
 func PrintServerAndExit() {
-	fmt.Printf("nats-server: v%s\n", VERSION)
+	fmt.Printf("nats-server: v%s\n", serverVersion)
```
thaJeztah, line 1553: What does serverVersion / VERSION contain for builds that are not a tag? Would a `git describe -tags` be an option, or is that's already what's used?

server/server.go
wallyqs, line 651:
```suggestion
	if serverVersion == _EMPTY_ {
```

server/server.go
LaurentGoderre: `VERSION` is in the code: https://github.com/nats-io/nats-server/blob/main/server/const.go#L47

LaurentGoderre: So the way this would work now is that the dev version would show unless the build overwrites it with a tag

LaurentGoderre:
@derekcollison can I get an updated review?

server/server.go
```diff
@@ -648,6 +648,10 @@ func New(opts *Options) *Server {
 func NewServer(opts *Options) (*Server, error) {
 	setBaselineOptions(opts)
 
+	if serverVersion == _EMPTY_ {
+		serverVersion = VERSION
```
wallyqs, line 652: seems that syft only requires the `serverVersion` / `Version` being included in the ldflags which is addressed in the goreleaser yaml above. 
I wonder whether we should be doing this here after all? We already have a test `TestVersionMatchesTag` that it is called previous to releasing so the released version should already be a tag.

server/server.go
wallyqs: I think for now maybe we could just include the changes to goreleaser yaml so that the proper metadata is included for the scanners and leave out changing VERSION dynamically.

server/server.go
LaurentGoderre: yeah. I can probably simplify it.

LaurentGoderre: @wallyqs the only issue is that now, two different versions could be reported if the code doesn't get updated in the tagged release.

wallyqs: agree but we have check in place to prevent that so I think we could skip for now

.goreleaser.yml
```diff
@@ -16,7 +16,7 @@ builds:
   flags:
   - -trimpath
   ldflags:
-    - -w -X github.com/nats-io/nats-server/v2/server.gitCommit={{.ShortCommit}}
+    - -w -X 'github.com/nats-io/nats-server/v2/server.gitCommit={{.ShortCommit}}' -X 'github.com/nats-io/nats-server/v2/server.serverVersion={{.CurrentTag}}' -X 'github.com/nats-io/nats-server/v2/version.Version={{.CurrentTag}}'
```
wallyqs, line 19: I see that `serverVersion` makes syft set the version correctly, but do we also need `github.com/nats-io/nats-server/v2/version.Version`?  Doest that also improve other scanners detecting the version?

LaurentGoderre:
@wallyqs I created what I believe is a more comprehensive solution. The first pass made the version searchable but still required some manual intervention in Syft. The new implementation (using `version.Version`) gets it picked up automatically.

.goreleaser-nightly.yml
```diff
@@ -5,7 +5,7 @@ builds:
     id: nats-server
     binary: nats-server
     ldflags:
-      - -w -X github.com/nats-io/nats-server/v2/server.gitCommit={{.ShortCommit}}
+      - -w -X github.com/nats-io/nats-server/v2/version.GitCommit={{.ShortCommit}} -X 'github.com/nats-io/nats-server/v2/version.Version={{.Summary}}'
```
LaurentGoderre, line 8: In this case `{{.Summary}}` behaves like `git describe` and uses the previous tag to generate a version

https://goreleaser.com/customization/templates/

wallyqs:
@LaurentGoderre thanks again for the PR and raising the issue, I would rather not remove the `VERSION` constant and instead introduce less breaking changes if possible to help `syft` detect the version from the metadata.  I have opened the following PR based on yours which I think will help add the version for `syft` for now relying on it extracting the info from the binary: https://github.com/nats-io/nats-server/pull/5650 
In the future I think the right fix would be that the ldflags metadata is not removed during the go build so that it is present via `go version -m` (meaning this issue is addressed: https://github.com/golang/go/issues/63432)

LaurentGoderre:
Closing in favor of https://github.com/nats-io/nats-server/pull/5650

PR #5600, opened 2024-06-27
[FIXED] Memory growth from NRG create and delete operations 

derekcollison:
Fix for situations where GC struggled to clean up things.

This mostly stemmed from subscriptions being held in per connection caches (routers and gateways). These held references to both clients and any internal callbacks, which usually were closures so would also hold reference to the struct target as well.
 

wallyqs (approved):
LGTM

PR #5602, opened 2024-06-27
[IMPROVED] Routing: reduce chances of duplicate implicit routes

### LINKED ISSUE #5483: 'Duplicate Route' connection floods on implicit routes

kozlovic:
This is an alternate approach to the PR #5484 from @wjordan.

Using the code in that PR with the test added in this PR, I could still see duplicate routes (up to 125 in one of the matrix), and still had a data race (that could have easily be fixed). The main issue is that the increment happens in connectToRoute, which is running from a go routine, so there were still chances for duplicates.

Instead, I took the approach that those duplicates were the result of way too many gossip protocols. Suppose that you have servers A and B already connected. C connects to A. A gossips to B that it should connect to C. When that happened, B would gossip to A the server C and C would gossip to A the server B, which all that was unnecessary. It would grow quite fast with the size of the cluster (that is, several thousands for a cluster size of 15 or so).

Resolves #5483

### DIFF
--- a/server/route.go
+++ b/server/route.go
@@ -82,6 +82,9 @@ type route struct {
 	// Selected compression mode, which may be different from the
 	// server configured mode.
 	compression string
+	// Transient value used to set the Info.NoGossip when initiating
+	// an implicit route and sending to the remote.
+	noGossip bool
 }
 
 type connectInfo struct {
@@ -684,12 +687,15 @@ func (c *client) processRouteInfo(info *Info) {
 		return
 	}
 
+	var sendDelayedInfo bool
+
 	// First INFO, check if this server is configured for compression because
 	// if that is the case, we need to negotiate it with the remote server.
 	if needsCompression(opts.Cluster.Compression.Mode) {
 		accName := bytesToString(c.route.accName)
 		// If we did not yet negotiate...
-		if !c.flags.isSet(compressionNegotiated) {
+		compNeg := c.flags.isSet(compressionNegotiated)
+		if !compNeg {
 			// Prevent from getting back here.
 			c.flags.set(compressionNegotiated)
 			// Release client lock since following function will need server lock.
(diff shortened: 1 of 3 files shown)


neilalexander:
@kozlovic We noticed that `TestStressChainedSolicitWorks` fails pretty reliably on our local machines now and have bisected it back to this change, would you mind taking a look?

PR #5603, opened 2024-06-27
[TEST only] A test for memory leak when creating/deleting consumers

levb:
Verifies the fix in https://github.com/nats-io/nats-server/pull/5600

When ran locally, it was using ~180MB extra for the test before, and under 40Mb after.


levb:
@derekcollison @neilalexander https://app.travis-ci.com/github/nats-io/nats-server/jobs/623464530 is causing a race in raft, do you mind taking a look?

This line here is not being executed under a lock.

levb:
Apologies, but either I messed something up when merging/rebasing or GitHub is having issues, I will close this PR and submit a new one since there's little history here anyway.

image

PR #5607, opened 2024-06-28
[TEST only] A test for memory leak when creating/deleting consumers

levb:
Verifies the fix in https://github.com/nats-io/nats-server/pull/5600

When ran locally, it was using ~180MB extra for the test before, and under 40Mb after.

(this is a replacement for #5603 that ran into some git issues.


derekcollison (approved):
LGTM

PR #5559, opened 2024-06-18
Use stree for message block `fss`

neilalexander:
Should provide some deduplication of long subjects in memory when message blocks are loaded in.

Signed-off-by: Neil Twigg <neil@nats.io>

PR #5616, opened 2024-07-04
Fixes for filestore FSS state

neilalexander:
This PR contains some bug-fixes for the filestore FSS state.

First, if the filter provided is `_EMPTY_` then we need to overwrite it to `>` for the stree functions to work, but we failed to set `wc` when doing so. This could affect the decision on whether to linear-scan or not. We will now set `wc` too in this case.

Second, if the FSS only contains a single subject then we may not correctly set `isAll` if the filter contained a wildcard, as `Find` does not work with wildcards and therefore wouldn't match the single subject. Updated to use `Match` instead, so that it correctly uses wildcards.

Third, replaces `Match(>)` with `Iter` in `enforceMsgPerSubjectLimit`, not because it is behaviourally different, but because we need to walk the entire stree anyway and `Iter` saves CPU cycles by skipping the match step.

Fourth, updates the bounds checking on various functions that use `Find` so that the logic matches the `Match` equivalents. This also matters in particular in `SubjectsState` which could fail altogether when the filter was a subject literal rather than a wildcard and the `psim` was out-of-date/needing update.

Fifth, in `NumPending` and in `filteredPendingLocked` when matching the FSS state, if we find a partial, don't process any further matches after that. This bug was introduced in #5559 as there used to be a `break` there, whereas now the `Match` that replaces it can't be interrupted in the same way.

### DIFF
--- a/server/filestore.go
+++ b/server/filestore.go
@@ -3084,6 +3084,10 @@ func (fs *fileStore) NumPending(sseq uint64, filter string, lastPerSubject bool)
 
 			var havePartial bool
 			mb.fss.Match(stringToBytes(filter), func(bsubj []byte, ss *SimpleState) {
+				if havePartial {
+					// If we already found a partial then don't do anything else.
+					return
+				}
 				subj := bytesToString(bsubj)
 				if ss.firstNeedsUpdate {
 					mb.recalculateFirstForSubj(subj, ss.First, ss)
--- a/server/filestore.go
+++ b/server/filestore.go
@@ -2323,7 +2323,10 @@ func (mb *msgBlock) firstMatching(filter string, wc bool, start uint64, sm *Stor
 
 	// If we only have 1 subject currently and it matches our filter we can also set isAll.
 	if !isAll && mb.fss.Size() == 1 {
-		_, isAll = mb.fss.Find(stringToBytes(filter))
+		// Since mb.fss.Find won't work if filter is a wildcard, need to use Match instead.
+		mb.fss.Match(stringToBytes(filter), func(subject []byte, _ *SimpleState) {
+			isAll = true
+		})
 	}
 	// Make sure to start at mb.first.seq if fseq < mb.first.seq
 	if seq := atomic.LoadUint64(&mb.first.seq); seq > fseq {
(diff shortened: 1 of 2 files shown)


server/filestore.go
```diff
@@ -2613,7 +2613,12 @@ func (fs *fileStore) checkSkipFirstBlock(filter string, wc bool) (int, int) {
 			}
 		})
 	} else if psi, ok := fs.psim.Find(stringToBytes(filter)); ok {
-		start, stop = psi.fblk, psi.lblk
+		if psi.fblk < start {
```
derekcollison, line 2620: I think the original was ok

server/filestore.go
```diff
@@ -2684,7 +2689,12 @@ func (fs *fileStore) numFilteredPendingWithLast(filter string, last bool, ss *Si
 		})
 	} else if psi, ok := fs.psim.Find(stringToBytes(filter)); ok {
 		ss.Msgs += psi.total
-		start, stop = psi.fblk, psi.lblk
+		if psi.fblk < start {
```
derekcollison, line 2696: Same here.

server/filestore.go
```diff
@@ -3715,7 +3730,12 @@ func (fs *fileStore) firstSeqForSubj(subj string) (uint64, error) {
 	// See if we can optimize where we start.
 	start, stop := fs.blks[0].index, fs.lmb.index
 	if info, ok := fs.psim.Find(stringToBytes(subj)); ok {
-		start, stop = info.fblk, info.lblk
+		if info.fblk > start {
```
derekcollison, line 3741: Same here.

derekcollison (approved):
LGTM - Small comments but really good changes.

server/filestore.go
neilalexander: You may be right, but just figured having equal handling of `Find` and `Match` paths eliminates one possible way that we could hit surprising behaviour in the future. 

BTW please feel free to un-mark draft if you want to merge or make changes!

derekcollison: Let's revert those then will merge.

Issue #5621, opened 2024-07-07
server deadlocks on /jsz access


arthur-cylus:
### Observed behavior

We have a server running inside docker, multiple consumers publishing and consuming messages, and prometheus-nats-exporter exporting metrics of the server. During normal state, we get to ~350% CPU load, but somewhere around once per day, suddenly the nats-server gets stuck, uses exactly 1 core (100% CPU load), and doesn't respond to publish or consumers, as can be seen in the following CPU usage graph:

![cpu usage](https://github.com/nats-io/nats-server/assets/133011827/b80a0111-7172-4592-bfd3-07d8d9b8d9f3)

We managed to profile using pprof the server, and we have the following recordings (the CPU is the interesting one):

[pprof.nats-server.samples.cpu.001.pb.gz](https://github.com/user-attachments/files/16118970/pprof.nats-server.samples.cpu.001.pb.gz)
[pprof.nats-server.samples.cpu.002.pb.gz](https://github.com/user-attachments/files/16118971/pprof.nats-server.samples.cpu.002.pb.gz)
[pprof.nats-server.alloc_objects.alloc_space.inuse_objects.inuse_space.001.pb.gz](https://github.com/user-attachments/files/16118973/pprof.nats-server.alloc_objects.alloc_space.inuse_objects.inuse_space.001.pb.gz)
[pprof.nats-server.alloc_objects.alloc_space.inuse_objects.inuse_space.002.pb.gz](https://github.com/user-attachments/files/16118974/pprof.nats-server.alloc_objects.alloc_space.inuse_objects.inuse_space.002.pb.gz)
[staging_nats_goroutine_full_stack_dump.txt](https://github.com/user-attachments/files/16119031/staging_nats_goroutine_full_stack_dump.txt)
[staging_nats_trace.out.zip](https://github.com/user-attachments/files/16119033/staging_nats_trace.out.zip)

It looks like there is a stuck lock in the handling of the response to `/jsz` HTTP route (this is how prometheus-nats-exporter queries the nats-server), halting the server. After a period it fixes itself (30 minutes to 2 hours).

We are checking if disabling prometheus-nats-exporter fixes the issue, but it will take time to verify (since it occurs once daily).

It started to happen between 2.10.12 and 2.10.16, because we were using 2.10.12 and it never occurred to us before, but started occurring after upgrading to 2.10.16 (we also attempted 2.10.17, but also occurred in it).

### Expected behavior

[...]

PR #5629, opened 2024-07-08
[IMPROVED] CPU spike during recalc of first msg in memstore.

derekcollison:
When recalculating for the first, make sure not to trip accidentally with a negative npc, and make sure to floor against first sequence as needed.

Resolves: #5621 5621


wallyqs (approved):
LGTM

PR #5630, opened 2024-07-09
[FIXED] checkSkipFirstBlock could return negative index if psim.fblk outdated.

derekcollison:
This would cause LoadNextMsg() to return EOF incorrectly.


wallyqs (approved):
LGTM

PR #5632, opened 2024-07-09
[FIXED] Had a perf regression with some fss fixes.

derekcollison:
We could still use find if we know the status of wildcard.


wallyqs (approved):
LGTM

PR #5635, opened 2024-07-09
Update Go to v1.22.5

wallyqs:


derekcollison (approved):
LGTM

derekcollison:
Do we have to change Go releaser stuff?

wallyqs:
We don't have to make Go releaser changes to update the Go version, I want to change to use goreleaser 2 on main but that would be a separate change.

PR #5636, opened 2024-07-09
Fix go directive version in go.mod

mauri870:
According to [https://go.dev/issues/62278](https://github.com/golang/go/issues/62278) a concrete go version has to include
the patch number. The project declares the version as 1.21 which is invalid and
will result in an error:

	GOTOOLCHAIN="go1.20+auto" go1.21.5 mod tidy
	go: downloading go1.21 (linux/amd64)
	go: download go1.21 for linux/amd64: toolchain not available

Changed to 1.21.0 as this is the first actual release of go 1.21.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/nats-io/nats-server/v2
 
-go 1.21
+go 1.21.0
 
 require (
 	github.com/google/go-tpm v0.9.1


wallyqs:
Thanks for bringing this into our attention, this currently fails the building on Go 1.20 which we allow compiling for at the moment but would remove soon:
```
go: errors parsing go.mod:
./src/github.com/nats-io/nats-server/go.mod:3: invalid go version '1.21.0': must match format 1.23
```
One question, this would not mean that we are forcing the Go toolchain to be v1.21.0 correct?

neilalexander:
> this currently fails the building on Go 1.20 which we allow compiling for at the moment

AFAIK we only support Go versions N and N-1, so that would be Go 1.22 and Go 1.21?

derekcollison:
What is the status of this PR?

mauri870:
> Thanks for bringing this into our attention, this currently fails the building on Go 1.20 which we allow compiling for at the moment but would remove soon:
> 
> ```
> go: errors parsing go.mod:
> ./src/github.com/nats-io/nats-server/go.mod:3: invalid go version '1.21.0': must match format 1.23
> ```
> 
> One question, this would not mean that we are forcing the Go toolchain to be v1.21.0 correct?

Sorry for the delay in replying. No, this simply means that you need Go >= 1.21.0 to build. The 1.21.0 is just a placeholder; 1.21 is not a valid toolchain version. The first valid Go version for any release is vx.xx.0. You are correct that we would be dropping Go 1.20 with this change. In order to pin a specific toolchain version you have to use the go directive plus the toolchain directive.

mauri870:
> What is the status of this PR?

I guess the main issue here is dropping support for Go 1.20, which is not maintained anymore since the release of 1.22 back in February.

derekcollison:
@wallyqs should we merge this?

wallyqs:
Needs another PR first to remove Go 1.20 otherwise build won't be green, will send PR

Issue #5612, opened 2024-07-02
work queue stream losing messages


aubergenepizza:
### Observed behavior

In an R3 cluster, when creating a work queue policy stream with explicit acks and a push consumers, it's possible to lose messages. The way to replicate it is to create and delete ephemeral consumers that listen to different subjects (I've attached a script that can quite reliably reproduce it). Based on the docs I'd expect the messages to not get deleted, as there's never an ack from the consumers, hence the work queue stream (with acks) has to retain the messages.

### Expected behavior

All messages sent should be present in the stream.

### Server and client version

nats-server: v2.10.16, v2.10.17
nats --version: 0.1.1

### Host environment

linux, either cloud k8s or a local cluster ([kind](https://kind.sigs.k8s.io/)) with 3 replicas for the servers

### Steps to reproduce

The main idea is to send some messages to different subjects in the stream, subscribe to the delivery subjects, then repeatedly create and delete ephemeral consumers that target the delivery subjects. On my system, with a kind cluster in an R3 setup, this script reliably causes message loss:
```bash
#!/bin/bash

nats stream create test --subjects 1 --subjects 2 --subjects 3 --subjects 4 --subjects 5 --subjects 6 --subjects 7 --subjects 8 --subjects 9 --subjects 10 --storage file --replicas 3 --retention work --deny-delete --no-deny-purge  --defaults

nats publish 2 --count 10 "test alskdjalksdjalskdjaksdjlaksjdlkajsdlakjsdlakjsdlakjdlakjsdlaksjdlj"
nats publish 5 --count 10 "test alskdjalksdjalskdjaksdjlaksjdlkajsdlakjsdlakjsdlakjdlakjsdlaksjdlj"
nats publish 7 --count 10 "test alskdjalksdjalskdjaksdjlaksjdlkajsdlakjsdlakjsdlakjdlakjsdlaksjdlj"
nats publish 9 --count 10 "test alskdjalksdjalskdjaksdjlaksjdlkajsdlakjsdlakjsdlakjdlakjsdlaksjdlj"


nats sub bob &
nats sub bob2 &

for i in {0..5}; do
        nats consumer rm test test --force; nats consumer add test test --trace --replay instant --deliver=all --ephemeral --ack explicit --max-deliver=-1 --max-pending=1 --target bob --defaults --filter 1 --filter 2 --filter 3 --filter 4  --filter 5
        sleep 7
[...]

Issue #5638, opened 2024-07-10
JetStream is discarding messages it likely shouldn't


aksdb:
### Observed behavior

We have a work queue stream that is being consumed by about 250 pull consumers and which requires ACKs. Recently (past weeks) we noticed, that a small fraction of messages get lost. Watching the ADVISORY subjects, it seems like JetStream drops the messages itself. Example:

```
13:20:39 {"type":"io.nats.jetstream.advisory.v1.terminated","id":"8pZH7RdhGdWrEJbkdfdy2D","timestamp":"2024-07-09T13:20:39.504816844Z","stream":"corejobs","consumer":"posts-PostSaved-worker","consumer_seq":311629,"stream_seq":49245214,"deliveries":1,"reason":"Unacknowledged message was deleted"}
```

The stream doesn't have a `MaxDeliver` set, and as you can see, it was just delivered once. So there should be no reason to drop it if an ACK was still pending. (We also saw messages being dropped with that reason that had `deliveries:0`.) From what I can tell, also limits were not exceeded - since the work queue only has very small messages (basically just headers without payload), the stream limit of 500 MB was certainly not reached. The message in question was also far younger than the maximum age (which is 7d; the message was barely a few minutes old).

Even more suspicious: our workers log processing of messages and we were not seeing this message ever hitting our worker. Also we sniffed the $JS-API and the _INBOX, and from what we can tell, the messages in question also never were delivered. We basically only see them being ACK'ed by JetStream as being received and soon after we see them being terminated in the advisory subject.

Here is the stream config of the stream in question:
```
Information for Stream corejobs created 2022-07-22 08:16:16

              Subjects: core.jobs.>
              Replicas: 3
              Storage: File

Options:

             Retention: WorkQueue
       Acknowledgments: true
        Discard Policy: Old
      Duplicate Window: 2m0s
     Allows Msg Delete: true
          Allows Purge: true
        Allows Rollups: false

Limits:

      Maximum Messages: unlimited
   Maximum Per Subject: unlimited
         Maximum Bytes: 500 MiB
           Maximum Age: 7d0h0m0s
  Maximum Message Size: unlimited
     Maximum Consumers: unlimited

Cluster Information:

                  Name: nats
                Leader: nats-cluster-4_v1
               Replica: nats-cluster-0_v1, current, seen 1.85s ago
               Replica: nats-cluster-3_v1, current, seen 1.85s ago

State:

              Messages: 0
                 Bytes: 0 B
[...]

PR #5639, opened 2024-07-10
[FIXED] Don't ack msgs if we are filtered and they are not applicable.

derekcollison:
When checking interest state, make sure to take into account filtered consumers when acking msgs.
We were checking interest state on a leader change but we would ack msgs that may not be applicable to our consumer.

Resolves: #5612 


kozlovic (approved):
LGTM. I think it will also resolve the issue recently reported that sounds really similar: https://github.com/nats-io/nats-server/issues/5638

PR #5649, opened 2024-07-14
[FIXED] Make sure to send consumer create advisory regardless of node presence.

derekcollison:
For R1 consumers we used to have the leader set as we called add so was handled there similar to standalone server mode. Now process in same place as R>1 for clustered usage.


wallyqs (approved):
LGTM

PR #5650, opened 2024-07-15
Add serverVersion var set via ldflags

wallyqs:
Setting a `serverVersion` variable which is set via ldflags to help security scanners like `syft` detect the version when inspecting the binary.  Goreleaser is modified to set this ldflag and a test is added to make sure that the `VERSION` constant matches the tag from the release.

Result:

```sh
$ export TRAVIS_TAG=v2.10.18

$ go test -race -v -run=TestVersionMatchesTag ./server --trimpath -ldflags="-X=github.com/nats-io/nats-server/v2/server.serverVersion=$TRAVIS_TAG" -count=1 -vet=off

# Example failure
=== RUN   TestVersionMatchesTag
    server_test.go:144: Version (2.11.0-dev) does not match tag (2.10.18)
--- FAIL: TestVersionMatchesTag (0.00s)

$ go build --trimpath -ldflags="-X 'github.com/nats-io/nats-server/v2/server.serverVersion=$TRAVIS_TAG'" .

$ syft ./nats-server
...   
github.com/nats-io/nats-server/v2  v2.10.18  go-module    
...
```

### DIFF
--- a/server/server_test.go
+++ b/server/server_test.go
@@ -143,6 +143,17 @@ func TestVersionMatchesTag(t *testing.T) {
 	if VERSION != tag[1:] {
 		t.Fatalf("Version (%s) does not match tag (%s)", VERSION, tag[1:])
 	}
+	// Check that the version dynamically set via ldflags matches the version
+	// from the server previous to releasing.
+	if serverVersion == _EMPTY_ {
+		t.Fatal("Version missing in ldflags")
+	}
+	// Unlike VERSION constant, serverVersion is prefixed with a 'v'
+	// since it should be the same as the git tag.
+	expected := "v" + VERSION
+	if serverVersion != _EMPTY_ && expected != serverVersion {
+		t.Fatalf("Version (%s) does not match ldflags version (%s)", expected, serverVersion)
+	}
 }
 
 func TestStartProfiler(t *testing.T) {
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -16,7 +16,7 @@ builds:
   flags:
   - -trimpath
   ldflags:
-    - -w -X github.com/nats-io/nats-server/v2/server.gitCommit={{.ShortCommit}}
+    - -w -X 'github.com/nats-io/nats-server/v2/server.gitCommit={{.ShortCommit}}' -X 'github.com/nats-io/nats-server/v2/server.serverVersion={{.Tag}}'
   env:
     - GO111MODULE=on
     - CGO_ENABLED=0
--- a/scripts/runTestsOnTravis.sh
+++ b/scripts/runTestsOnTravis.sh
@@ -10,7 +10,7 @@ if [ "$1" = "compile" ]; then
     go install github.com/golangci/golangci-lint/cmd/golangci-lint@v1.56.1;
     golangci-lint run;
     if [ "$TRAVIS_TAG" != "" ]; then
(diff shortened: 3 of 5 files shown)


server/server_test.go
```diff
@@ -143,6 +143,16 @@ func TestVersionMatchesTag(t *testing.T) {
...
 		t.Fatalf("Version (%s) does not match tag (%s)", VERSION, tag[1:])
 	}
+	// Check that the version dynamically set via ldflags matches the version
+	// from the server previous to releasing.
+	if serverVersion == _EMPTY_ {
+		t.Fatal("Version missing in ldflags")
+	}
+	// Unlike VERSION constant, this one is prefixed with a 'v'
+	// since it is the same as the git tag.
+	if serverVersion != _EMPTY_ && VERSION != serverVersion[1:] {
```
bruth, line 153: A slight improvement to confirm there is a `v` prefix would be to do: `"v" + VERSION != serverVersion`

wallyqs: good idea, should check that `serverVersion` does include the `v` prefix

bruth (approved):

LaurentGoderre:
@wallyqs `serverVersion` isn't one of the detected one. The one that are supported are `BuildVersion`, `GitVersion`, `ReleaseVersion` or `Tag`.  With that being said the flag can be there and not pointing to an actual variable in the code and Syft and it would work.

LaurentGoderre (approved):

LaurentGoderre:
Funny it kind of went full circle to the original PR. Thank you so much for being open to this change!

LaurentGoderre:
@wallyqs is this going to be backported for 2.9.x?

wallyqs:
@LaurentGoderre no, there are no releases planned for v2.9 series

LaurentGoderre:
@wallyqs the `ldflag` section is absent from the binary in the release because of the `-trimpath` flag. Is that something that is required?

wallyqs:
@LaurentGoderre yes we need the `--trimpath` flag to have reproducible builds (https://go.dev/blog/rebuild). Latest release though seems to be ok for syft?

```
syft ./nats-server-v2.10.18-darwin-arm64/nats-server 
github.com/klauspost/compress      v1.17.9   go-module    
github.com/minio/highwayhash       v1.0.3    go-module    
github.com/nats-io/jwt/v2          v2.5.8    go-module    
github.com/nats-io/nats-server/v2  v2.10.18  go-module    
github.com/nats-io/nkeys           v0.4.7    go-module    
go.uber.org/automaxprocs           v1.5.3    go-module    
golang.org/x/crypto                v0.25.0   go-module    
golang.org/x/sys                   v0.22.0   go-module    
golang.org/x/time                  v0.5.0    go-module    
stdlib                             go1.22.5  go-module  
```

LaurentGoderre:
With the default Syft config yes but it gets them by doing a regex for anything that looks like a version in the binary. With Docker Scout we use a stricter rule and rely on the ldflag in the binary.

LaurentGoderre:
Ha!! seems like you ran into this too!! https://github.com/golang/go/issues/63432

wallyqs:
@LaurentGoderre yes... that is what I was afraid we would stumble upon eventually (https://github.com/nats-io/nats-server/pull/5599#issuecomment-2192619470), but at least now situation has improved since `syft` does manage to get the version right.

PR #5651, opened 2024-07-15
Use ReadBuildInfo if available to detect gitCommit

wallyqs:
Currently we use an `ldflag` to add the git commit from a server version, but this information is already present in many cases such as when building the package or using `go install`.

```
$ go build .
$  ./nats-server 
[90016] 2024/07/15 17:01:31.280879 [INF] Starting nats-server
[90016] 2024/07/15 17:01:31.281050 [INF]   Version:  2.11.0-dev
[90016] 2024/07/15 17:01:31.281054 [INF]   Git:      [dcd20c0]
```

When calling `go run` or `go build main.go` it would be still left unset:

```
go run main.go 
[90162] 2024/07/15 17:02:13.081669 [INF] Starting nats-server
[90162] 2024/07/15 17:02:13.081845 [INF]   Version:  2.11.0-dev
[90162] 2024/07/15 17:02:13.081849 [INF]   Git:      [not set]
```

### DIFF
--- a/server/const.go
+++ b/server/const.go
@@ -15,6 +15,7 @@ package server
 
 import (
 	"regexp"
+	"runtime/debug"
 	"time"
 )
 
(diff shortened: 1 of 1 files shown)


bruth (approved):

derekcollison (approved):
LGTM

PR #5652, opened 2024-07-15
Report the js limits in jsz output

ripienaar:
Reports some often requested information about JS config in jsz

### DIFF
--- a/server/opts.go
+++ b/server/opts.go
@@ -233,10 +233,10 @@ type RemoteLeafOpts struct {
 }
 
 type JSLimitOpts struct {
-	MaxRequestBatch int
-	MaxAckPending   int
-	MaxHAAssets     int
-	Duplicates      time.Duration
+	MaxRequestBatch int           `json:"max_request_batch,omitempty"`
+	MaxAckPending   int           `json:"max_ack_pending,omitempty"`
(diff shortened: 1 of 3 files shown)


server/opts.go
```diff
@@ -233,10 +233,10 @@ type RemoteLeafOpts struct {
...
 
 type JSLimitOpts struct {
-	MaxRequestBatch int
-	MaxAckPending   int
-	MaxHAAssets     int
-	Duplicates      time.Duration
+	MaxRequestBatch int           `json:"max_request_batch,omitempty"`
+	MaxAckPending   int           `json:"max_ack_pending,omitempty"`
+	MaxHAAssets     int           `json:"max_ha_assets,omitempty"`
+	Duplicates      time.Duration `json:"duplicates,omitempty"`
```
ripienaar, line 239: I didnt do it here but propose we rename this to something more descriptive MaxDuplicateWindow maybe

derekcollison: You can change the json filed name and have clients match that..

derekcollison (approved):
LGTM - let me know if you want changes before merge.

ripienaar:
Ok I will update that json key in a few hours

ripienaar:
I changed it, so good to go I think @derekcollison

PR #5655, opened 2024-07-15
[FIXED] Consumer NumPending higher than total stream messages.

derekcollison:
When we matched all and had no interior deletes we would short circuit num pending calculations but were not accounting for sseq < state.FirstSeq.


wallyqs (approved):
LGTM

PR #5660, opened 2024-07-16
Export `LameDuckShutdown` so that it is accessible when embedding NATS

neilalexander:
Otherwise this function isn't available when embedding the NATS Server, where using `SIGUSR2` likely isn't appropriate.

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -4150,6 +4150,16 @@ func (s *Server) isLameDuckMode() bool {
 	return s.ldm
 }
 
+// LameDuckShutdown will perform a lame duck shutdown of NATS, whereby
+// the client listener is closed, existing client connections are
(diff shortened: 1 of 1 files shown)


derekcollison (changes requested):
To minimize changes just make public one call private one.

server/server.go
```diff
@@ -4150,6 +4150,16 @@ func (s *Server) isLameDuckMode() bool {
 	return s.ldm
 }
 
+// This function will perform a lame duck shutdown of NATS, whereby
```
wallyqs, line 4153:
```suggestion
// LameDuckShutdown will perform a lame duck shutdown of NATS, whereby
```

derekcollison (approved):
LGTM

Issue #5656, opened 2024-07-15
Panic nil pointer dereference


TroyKomodo:
### Observed behavior

Whilst consuming multiple streams concurrently nats panics.
```
seventv-nats-1  | panic: runtime error: invalid memory address or nil pointer dereference
seventv-nats-1  | [signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0xa50dde]
seventv-nats-1  | 
seventv-nats-1  | goroutine 1 [running]:
seventv-nats-1  | github.com/nats-io/nats-server/v2/server/stree.(*SubjectTree[...]).Match(0xcc9180, {0xc065630860, 0x20, 0x20}, 0xc0003c6050)
seventv-nats-1  |       github.com/nats-io/nats-server/v2/server/stree/stree.go:107 +0xbe
seventv-nats-1  | github.com/nats-io/nats-server/v2/server.(*fileStore).NumPending(0xc000276e08, 0x173be30, {0xc065630860, 0x20}, 0x0)
seventv-nats-1  |       github.com/nats-io/nats-server/v2/server/filestore.go:3051 +0xcf5
seventv-nats-1  | github.com/nats-io/nats-server/v2/server.(*consumer).calculateNumPending(0xc0001e6f08)
seventv-nats-1  |       github.com/nats-io/nats-server/v2/server/consumer.go:4241 +0x135
seventv-nats-1  | github.com/nats-io/nats-server/v2/server.(*consumer).streamNumPending(0xc0001e6f08)
seventv-nats-1  |       github.com/nats-io/nats-server/v2/server/consumer.go:4203 +0x2c
seventv-nats-1  | github.com/nats-io/nats-server/v2/server.(*consumer).setLeader(0xc0001e6f08, 0x1)
seventv-nats-1  |       github.com/nats-io/nats-server/v2/server/consumer.go:1207 +0x3db
seventv-nats-1  | github.com/nats-io/nats-server/v2/server.(*stream).addConsumerWithAssignment(0xc000276a88, 0xc0a06e8028, {0x0, 0x0}, 0x0, 0x1, 0x0)
seventv-nats-1  |       github.com/nats-io/nats-server/v2/server/consumer.go:1037 +0x24e5
seventv-nats-1  | github.com/nats-io/nats-server/v2/server.(*Account).EnableJetStream(0xc00029c288, 0xc000025560)
[...]

PR #5662, opened 2024-07-16
[FIXED] More protection against stree being nil

derekcollison:
There are cases in filestore where errors can leave an stree nil. We report those errors but do not terminate all executions so this protects us generally from the stree being nil.

Resolves: #5656 


neilalexander (approved):
LGTM

wallyqs (approved):
LGTM

PR #5663, opened 2024-07-16
[FIXED] Prevent a server from panicking when shutdown immediately after starting.

derekcollison:


wallyqs (approved):
LGTM

PR #5666, opened 2024-07-17
NRG (2.11): Ensure proposal and AE response queues drain after stepdown

neilalexander:
This ensures that when a Raft node steps down, any items that remain in the proposal or append entry response queues are correctly dropped. Otherwise the group might end up in an inconsistent state if that node becomes leader again.


derekcollison (approved):
LGTM

PR #5671, opened 2024-07-18
NRG (2.11): Fix term handling in candidate state and use higher term from vote request

neilalexander:
A candidate could incorrectly revert to an older term without resetting if an old AE arrived with a term that is at least newer than the pterm but not necessarily newer than the term.

Additionally, we also weren't handling the case that the rest of the cluster should assume the higher term number from the vote requests once an isolated candidate node rejoins the cluster.

This PR also rewrites `TestNRGCandidateStepsDownAfterAE` and replaces it with `TestNRGAssumeHighTermAfterCandidateIsolation` as the old test was checking for flawed behaviour: it was relying on the term being able to go backwards after a period of isolation, which is not correct. Instead we should prove that the rest of the cluster assumes the higher term from the isolated node.

Co-authored-by: Reuben Ninan <reuben@nats.io>

### DIFF
--- a/server/raft.go
+++ b/server/raft.go
@@ -3201,11 +3201,11 @@ func (n *raft) processAppendEntry(ae *appendEntry, sub *subscription) {
 			// If the append entry term is newer than the current term, erase our
 			// vote.
 			if ae.term > n.term {
+				n.term = ae.term
 				n.vote = noVote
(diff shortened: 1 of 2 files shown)


derekcollison (approved):
LGTM

PR #5672, opened 2024-07-18
Fix panic on `/raftz` when shutting down

neilalexander:


wallyqs (approved):
LGTM

server/monitor.go
```diff
@@ -3788,7 +3788,13 @@ func (s *Server) HandleRaftz(w http.ResponseWriter, r *http.Request) {
 	gfilter := r.URL.Query().Get("group")
 	afilter := r.URL.Query().Get("acc")
 	if afilter == "" {
```
derekcollison, line 3790: _EMPTY_

derekcollison (approved):
LGTM - one nit

server/monitor.go
neilalexander: fixed

PR #5686, opened 2024-07-23
Make jsz limits optional

ripienaar:
To create tooling and to be able to detect
when ran against old NATS Servers it would
be better if this was a pointer so the absense
of it would indicate unsupported servers rather
than ones that support this feature but have
no limits set


ripienaar:
Sorry @derekcollison tried to add support to CLI for the PR https://github.com/nats-io/nats-server/pull/5652 and I realised I screwed up so small follow up here

derekcollison (approved):
LGTM

PR #5687, opened 2024-07-23
NRG: Don't delete on-disk state if failing to create internal subs

neilalexander:
If we failed to start the Raft group subscriptions then we were calling `shutdown` with the `shouldDelete` flag set, which would nuke the state on disk, blowing away the WAL, the term and vote etc.

However, this could happen if a Raft group tried to be started while the server was shutting down. When this happened, we would see a log entry saying `Error creating raft group: system account not setup` and then the Raft state would get deleted, so after a restart, all state was lost.

This PR changes `shouldDelete` to false so that we preserve the state on disk for the next startup.


derekcollison (approved):
LGTM

PR #5689, opened 2024-07-23
NRG (2.11): Don't reset WAL on append entry response

neilalexander:
Log consistency should only be enforced when handling append entries, not in other types of RPC. In this case, the higher term could cause us to blow away our entire log if a node with a higher term but a more out-of-date log comes along as leader without doing any other checks, and that may force us to vote for another node that's behind as a result.


derekcollison (approved):
LGTM

PR #5693, opened 2024-07-24
Export js limits in varz

ripienaar:
Same as previous limits related PR but also adding it to varz, noticed varz didnt have JetStream tests so reworked a few things there to also cover that part of varz


derekcollison (approved):
LGTM

PR #5694, opened 2024-07-24
(2.11) Revert to `go-tpm` 0.9.0 due to supported Go versions

neilalexander:
The 0.9.1 version of this dependency requires Go 1.22 whereas 0.9.0 requires Go 1.20.

We'll bump this back up before the 2.11 release as Go 1.23 will be out by then, so we'll only need to support Go 1.22 and Go 1.23 from that point forward.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,7 +3,7 @@ module github.com/nats-io/nats-server/v2
 go 1.21.0
 
 require (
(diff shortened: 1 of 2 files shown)


wallyqs (approved):
LGTM

PR #5695, opened 2024-07-24
Send the correct struct to the restore API

ripienaar:
While validating the ideas in ADR-44 the proposed improvements caught the fact that a snapshot request was being sent to a restore API call. Tests passed because there was enough overlap in the structs but strictly should have been a failure due to the invalid request


derekcollison (approved):
LGTM

PR #5697, opened 2024-07-24
Fix `WorkQueuePolicy` retention issue

neilalexander:
The `isInterestRetention` function, which gates whether or not we run `checkInterestState`, was matching for both `InterestPolicy` and `WorkQueuePolicy`.

This meant that a work queue stream could throw away messages for which there was no consumer interest as the consumer ack floors for other subjects moved up.

This PR changes `isInterestRetention` to match only `InterestPolicy` which fixes the issue. Notably this doesn't change the `ackMsg` behaviour which will continue to remove messages once they have been acked. 

### DIFF
--- a/server/consumer.go
+++ b/server/consumer.go
@@ -2990,6 +2990,28 @@ func (o *consumer) isFiltered() bool {
 	return false
 }
 
+// Check if we would have matched and needed an ack for this store seq.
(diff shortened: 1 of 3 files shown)


derekcollison (approved):
LGTM

derekcollison (approved):
LGTM

PR #5699, opened 2024-07-25
When checking consumer state for interest policy streams, make sure match filters

derekcollison:


neilalexander (approved):
LGTM

PR #5700, opened 2024-07-25
(2.11) Don't send meta snapshot when becoming metaleader

neilalexander:
Antithesis testing has found that late or out-of-order delivery of these snapshots, likely due to latency or thread pauses, can cause stream assignments to be reverted which results in assets being deleted and recreated. There may also be a race condition where the metalayer comes up before network connectivity to all other nodes is fully established so we may end up generating snapshots that don't include assets we don't know about yet.

We will want to audit all uses of `SendSnapshot` as it somewhat breaks the consistency model, especially now that we have fixed a significant number of Raft bugs that `SendSnapshot` usage may have been papering over.

Further Antithesis runs without this code run fine and have eliminated a number of unexpected calls to `processStreamRemoval`.

We've also added a new unit test `TestJetStreamClusterHardKillAfterStreamAdd` for a long-known issue, as well as a couple tweaks to the ghost consumer tests to make them reliable.

### DIFF
--- a/server/consumer.go
+++ b/server/consumer.go
@@ -1637,6 +1637,16 @@ func (o *consumer) updateDeliveryInterest(localInterest bool) bool {
 	return false
 }
 
+const (
+	defaultConsumerNotActiveStartInterval = 30 * time.Second
+	defaultConsumerNotActiveMaxInterval   = 5 * time.Minute
+)
+
+var (
+	consumerNotActiveStartInterval = defaultConsumerNotActiveStartInterval
+	consumerNotActiveMaxInterval   = defaultConsumerNotActiveMaxInterval
+)
+
 func (o *consumer) deleteNotActive() {
 	o.mu.Lock()
 	if o.mset == nil {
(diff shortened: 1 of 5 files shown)


server/consumer.go
```diff
@@ -1703,7 +1703,7 @@ func (o *consumer) deleteNotActive() {
 			// Don't think this needs to be a monitored go routine.
 			go func() {
 				const (
-					startInterval = 30 * time.Second
+					startInterval = 5 * time.Second
```
derekcollison, line 1706: Why is this being changed in a PR about meta snapshots?

neilalexander: Sending the snapshot was forcing the ghost consumers tests to pass for the wrong reason.

When we removed it, the test started to flake unless it was given over a minute to run. Reducing the initial time to clean up those ghost consumers de-flaked those tests without having to extend the test to such a long runtime.

derekcollison: Yes but that can dramatically impact production systems with 10s or 100s of thousands of these. Hence the 30s vs shorter.

If need be make a var off of a const and set var to what you need in test and reset at the end of the test.

neilalexander: OK will do tomorrow.

neilalexander: In fact have just done now, as it was a quick change.

neilalexander: Just removed the accidental `const ()` that was left over too.

derekcollison (approved):
LGTM

PR #5704, opened 2024-07-26
[jsz] Add StreamLeaderOnly filter option

### LINKED ISSUE #5698: Add JSZ filter to return stream/consumer details only if the server is the leader
### Proposed change

Add a `stream_leader` (other names welcome) filter option to the `JszOptions` to reduce the data of `stream_detail` to only include data where the server being pinged is the leader of streams.

### Use case

For systems having a lot of JetStream assets, this will reduce the overall resources use on the server as well as the data sent to the client.

### Contribution

Yes

bruth:
The stream state and replica info is only guaranteed to be accurate when returned by the leader of a given stream. This new option returns stream details only for the stream in which the server is the leader for that stream. For systems with many streams this can significantly reduce the amount of data returned when scraping across all servers since non-leader details will likely be ignored.

Fix #5698

### DIFF
--- a/server/monitor.go
+++ b/server/monitor.go
@@ -2727,15 +2727,16 @@ func (s *Server) accountInfo(accName string) (*AccountInfo, error) {
 
 // JSzOptions are options passed to Jsz
 type JSzOptions struct {
-	Account    string `json:"account,omitempty"`
-	Accounts   bool   `json:"accounts,omitempty"`
-	Streams    bool   `json:"streams,omitempty"`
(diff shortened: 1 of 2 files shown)


bruth:
Thinking about this a bit more, for the cases that a leader is not elected/available for some reason, is the output of JSZ to not set the cluster leader field? If so, then having this option set would never return those cases by any of the servers.

derekcollison (approved):
LGTM

PR #5710, opened 2024-07-28
[IMPROVED] LoadNextMsg for wildcard consumers with millions of subjects

derekcollison:
When LoadNextMsg misses, make sure to consult psim but conservatively.

We had a use case with millions of subjects and the last sequence checked being in the next to the last block. The consumer had a wildcard that matched lots of entries that were behind where we were. This would burn alot of cpu and when a stream had lots of consumers and they shift leadership this would introduce some instability due to all the cpu cycles.


neilalexander (approved):
LGTM

PR #5714, opened 2024-07-29
NRG (2.11): Send AE response when term is lower than ours

neilalexander:
Otherwise we may end up in a situation where we are dropping append entries from a leader that is obviously behind, but we cannot help to bring the cluster forward to make progress.

This resolves the flakiness in `TestJetStreamLeafNodeClusterExtensionWithSystemAccount` in local testing.

### DIFF
--- a/server/raft.go
+++ b/server/raft.go
@@ -3265,8 +3265,11 @@ func (n *raft) processAppendEntry(ae *appendEntry, sub *subscription) {
 			n.stepdownLocked(ae.leader)
 		}
 	} else if ae.term < n.term && !catchingUp && isNew {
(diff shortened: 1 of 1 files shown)


server/raft.go
```diff
@@ -3266,7 +3266,10 @@ func (n *raft) processAppendEntry(ae *appendEntry, sub *subscription) {
 		}
 	} else if ae.term < n.term && !catchingUp && isNew {
 		n.debug("Ignoring AppendEntry from a leader (%s) with term %d which is less than ours", ae.leader, ae.term)
```
ReubenMathew, line 3268: we should change this log message since we're no longer ignoring

derekcollison (approved):
LGTM

PR #5724, opened 2024-07-30
[auth-callout] Support opt-in list of accounts to be delegated in config-mode

bruth:
This introduces a new `auth_callout` field `allowed_accounts` which enables explicitly listing which accounts in the server config should be delegated to the auth callout service.

This improves the current behavior which is an all-or-nothing switch over from config auth to auth callout.

The primary motivation for this change is to enable the system account to authenticate against the server rather than being dependent on the availability of the auth callout service.

(For reference the name `allowed_accounts` [matches what exists in nsc](https://github.com/nats-io/nsc/blob/main/cmd/editauthorization.go#L33))

### DIFF
--- a/server/auth.go
+++ b/server/auth.go
@@ -606,12 +606,38 @@ func (s *Server) processClientOrLeafAuthentication(c *client, opts *Options) (au
 			}
 			return
 		}
-		// We have a juc defined here, check account.
+		// We have a juc, check if externally managed, i.e. should be delegated
+		// to the auth callout service.
 		if juc != nil && !acc.hasExternalAuth() {
 			if !authorized {
 				s.sendAccountAuthErrorEvent(c, c.acc, reason)
 			}
 			return
+		}
+		// Check config-mode. The global account is a condition since users that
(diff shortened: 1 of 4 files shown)


aricart (approved):
LGTM

derekcollison (approved):
LGTM

bondehagen:
@derekcollison Using an auth_callout service without a fallback admin account available (in case the auth service goes down) is preventing us from moving to production with callout. I strongly hope this fix can be included in the next release.

It's a necessary adjustment to make auth_callout a fully usable feature.

derekcollison:
@bondehagen meaning you want this PR released or this does not fit the bill?

IIRC you could always use the auth callout user itself to connect to the system..

bondehagen:
> @bondehagen meaning you want this PR released or this does not fit the bill?
> 
> IIRC you could always use the auth callout user itself to connect to the system..

I want the PR released, it's exactly what I need. You are right I can use the auth user to connect for the case I mentioned. But we have multiple accounts that should not use callout. Just wanted to let you know that someone is waiting eagerly for it.

PR #5729, opened 2024-07-31
Use `O_SYNC` for filestore meta files when `SyncAlways` is enabled

neilalexander:
This will use `O_SYNC` when writing `meta.inf`, `meta.sum` and `meta.key` files for both streams and consumers when the `SyncAlways` filestore option is enabled. It will also do the same when writing the consumer state file, or rewriting it if we are converting between ciphers.

### DIFF
--- a/server/filestore.go
+++ b/server/filestore.go
@@ -27,6 +27,7 @@ import (
 	"fmt"
 	"hash"
 	"io"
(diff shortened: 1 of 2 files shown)


derekcollison (approved):
LGTM

PR #5730, opened 2024-07-31
Watch for file changes using fsnotify

akhenakh:
Add an option:
```
        --watch-conf                 Watch for config files changes and reload
```
As well as `watch_conf` syntax in configuration file.
Supports reload on and off.

This could open the door to get rid of the `nats-server-config-reloader` container in Kubernetes configurations.

If you are interested in this PR I'll continue the work and add tests.

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -42,12 +42,14 @@ import (
 	"sync/atomic"
 	"time"
 
+	"github.com/fsnotify/fsnotify"
 	"github.com/klauspost/compress/s2"
 	"github.com/nats-io/jwt/v2"
 	"github.com/nats-io/nkeys"
 	"github.com/nats-io/nuid"
 
 	"github.com/nats-io/nats-server/v2/logger"
+	"github.com/nats-io/nats-server/v2/server/debounce"
 )
 
 const (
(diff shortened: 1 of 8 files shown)


server/server.go
```diff
@@ -1403,6 +1405,82 @@ func (s *Server) configureResolver() error {
...
+					// we watch the parent directory then filter
+					var found bool
+					if opts.ConfigFile == filepath.Base(event.Name) {
+						found = true
+					}
+					if !found {
+						continue
+					}
+
+					if event.Has(fsnotify.Write) || event.Has(fsnotify.Rename) {
```
wallyqs, line 1448: There are some extra things to do to handle the fsnotify events properly in some environments, this issue and PR to fix can offer some background:
- https://github.com/nats-io/nack/issues/130
- https://github.com/nats-io/nack/pull/131

wallyqs:
I think this is going to be cumbersome to integrate with the server as it is as we need to consider multiple things that the reloader process handles, like monitoring TLS certificates, multiple files, handling some K8S filesystem quirks, etc... and then have those be exposed as options somehow to the nats-server.  A first step I think would be to try to use the reloader as a library rather than reimplementing: https://github.com/nats-io/nack/blob/main/pkg/natsreloader/natsreloader.go

akhenakh:
Your recommendations make sense, closing this since it's not the right direction.

PR #5731, opened 2024-08-01
Improve config parsing API: file vs. string

johnweldon:
The publicly exposed API for parsing a string containing configuration and a string denoting a config file were unequal.

When parsing a config file, additional checks are automatically done in the (*Options).ProcessConfigFile method, however if parsing a string containing config then those additional checks have to be recreated in consumers of the public API.

This adds:

- conf.ParseWithChecks (pedantic mode equivalent to conf.Parse)
- (*Options).ProcessConfigString, which shares the same mechanism as (*Options).ProcesConfigFile.

### DIFF
--- a/conf/parse_test.go
+++ b/conf/parse_test.go
@@ -1,6 +1,8 @@
 package conf
 
 import (
+	"bytes"
+	"encoding/json"
 	"fmt"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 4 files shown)


conf/parse_test.go
```diff
@@ -25,6 +25,17 @@ func test(t *testing.T, data string, ex map[string]any) {
...
+	// Also test ParseWithChecks
+	n, err := ParseWithChecks(data)
+	if err != nil {
+		t.Fatalf("Received err: %v\n", err)
+	}
+	if n == nil {
+		t.Fatal("Received nil map")
+	}
+
+	// DeepEqual will fail due to pedantic mode creating a different map.
```
wallyqs, line 38: I think we can use JSON here to confirm that the maps are equivalent, can you cherry pick this commit into your branch?
https://github.com/nats-io/nats-server/commit/066cf38d40d87757a718279ad029397bbae204b1

wallyqs (approved):
LGTM

server/opts.go
```diff
@@ -874,6 +874,22 @@ func (o *Options) ProcessConfigFile(configFile string) error {
...
+
+// ProcessConfigString is the same as ProcessConfigFile, but expects the
+// contents of the config file to be passed in rather than the file name.
+func (o *Options) ProcessConfigString(data string) error {
+	m, err := conf.ParseWithChecks(data)
+	if err != nil {
+		return err
+	}
+
+	return o.processConfigFile("", m)
```
derekcollison, line 889: In server tree use `_EMPTY_` vs `""`.

derekcollison (commented):
This one ready?

wallyqs (approved):
LGTM

PR #5732, opened 2024-08-01
Reject leafnode cluster names with spaces

wallyqs:
This is a relaxed version of v2.11s change (https://github.com/nats-io/nats-server/pull/5676/files) which still allows cluster names with spaces in gateways and clusters.

### DIFF
--- a/server/errors.go
+++ b/server/errors.go
@@ -180,6 +180,9 @@ var (
 	// ErrClusterNameRemoteConflict signals that a remote server has a different cluster name.
 	ErrClusterNameRemoteConflict = errors.New("cluster name from remote server conflicts")
 
+	// ErrClusterNameHasSpaces signals that the cluster name contains spaces, which is not allowed.
+	ErrClusterNameHasSpaces = errors.New("cluster name cannot contain spaces or new lines")
+
 	// ErrMalformedSubject is returned when a subscription is made with a subject that does not conform to subject rules.
 	ErrMalformedSubject = errors.New("malformed subject")
 
--- a/server/leafnode.go
+++ b/server/leafnode.go
@@ -34,6 +34,7 @@ import (
 	"sync"
 	"sync/atomic"
 	"time"
+	"unicode"
 
 	"github.com/klauspost/compress/s2"
 	"github.com/nats-io/jwt/v2"
@@ -1762,6 +1763,13 @@ func (c *client) processLeafNodeConnect(s *Server, arg []byte, lang string) erro
 	proto := &leafConnectInfo{}
 	if err := json.Unmarshal(arg, proto); err != nil {
 		return err
+	}
+
+	// Reject a cluster that contains spaces or line breaks.
+	if proto.Cluster != _EMPTY_ && strings.ContainsFunc(proto.Cluster, unicode.IsSpace) {
+		c.sendErrAndErr(ErrClusterNameHasSpaces.Error())
+		c.closeConnection(ProtocolViolation)
+		return ErrClusterNameHasSpaces
 	}
 
 	// Check for cluster name collisions.
--- a/server/jetstream_cluster_4_test.go
+++ b/server/jetstream_cluster_4_test.go
@@ -2003,3 +2003,339 @@ func TestJetStreamClusterWQRoundRobinSubjectRetention(t *testing.T) {
 	require_Equal(t, si.State.NumDeleted, 20)
 	require_Equal(t, si.State.NumSubjects, 4)
 }
+
+type captureLeafClusterSpacesLogger struct {
+	DummyLogger
+	ch chan string
+}
+
+func (l *captureLeafClusterSpacesLogger) Errorf(format string, args ...any) {
+	msg := fmt.Sprintf(format, args...)
+	if strings.Contains(msg, `Leafnode Error 'cluster name cannot contain spaces or new lines'`) {
+		select {
+		case l.ch <- msg:
+		default:
+		}
+	}
+}
+
+func TestJetStreamClusterAndNamesWithSpaces(t *testing.T) {
+	gwConf := `
+		listen: 127.0.0.1:-1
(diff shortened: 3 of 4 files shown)


ripienaar:
Should also check for new lines ?

server/leafnode.go
```diff
@@ -1764,6 +1765,13 @@ func (c *client) processLeafNodeConnect(s *Server, arg []byte, lang string) erro
 		return err
 	}
 
+	// Reject a cluster that contains spaces or line breaks.
+	if proto.Cluster != _EMPTY_ && (strings.Contains(proto.Cluster, " ") || strings.Contains(proto.Cluster, "\n")) {
```
wallyqs, line 1769: @ripienaar yes actually, updated here

ripienaar: Would strings.ContainsAny() be faster/better?

wallyqs: yeah like that better, updated

server/errors.go
```diff
@@ -180,6 +180,9 @@ var (
 	// ErrClusterNameRemoteConflict signals that a remote server has a different cluster name.
 	ErrClusterNameRemoteConflict = errors.New("cluster name from remote server conflicts")
 
+	// ErrClusterNameHasSpaces signals that the cluster name contains spaces, which is not allowed.
+	ErrClusterNameHasSpaces = errors.New("cluster name cannot contain spaces")
```
ripienaar, line 184: `....or new lines`?

wallyqs: would have to update main anyway with this so sounds good

server/leafnode.go
```diff
@@ -1764,6 +1764,13 @@ func (c *client) processLeafNodeConnect(s *Server, arg []byte, lang string) erro
 		return err
 	}
 
+	// Reject a cluster that contains spaces or line breaks.
+	if proto.Cluster != _EMPTY_ && strings.ContainsAny(proto.Cluster, " \n") {
```
neilalexander, line 1768: `CONNECT_ARG` in the parser is also sensitive to `\r` and `OP_CONNECT` is sensitive to `\t` so let's include those too.

wallyqs: thanks, now added

server/leafnode.go
```diff
@@ -1764,6 +1764,13 @@ func (c *client) processLeafNodeConnect(s *Server, arg []byte, lang string) erro
 		return err
 	}
 
+	// Reject a cluster that contains spaces or line breaks.
+	if proto.Cluster != _EMPTY_ && strings.ContainsAny(proto.Cluster, " \n\r\t") {
```
mprimi, line 1768: Consider using `strings.ContainsFunc(proto.Cluster, unicode.IsSpace)`

wallyqs: that works, updated

derekcollison:
This ready for review? @bruth you and @kozlovic ok with approach here?

server/leafnode.go
```diff
@@ -1764,6 +1765,13 @@ func (c *client) processLeafNodeConnect(s *Server, arg []byte, lang string) erro
 		return err
 	}
 
+	// Reject a cluster that contains spaces or line breaks.
+	if proto.Cluster != _EMPTY_ && strings.ContainsFunc(proto.Cluster, unicode.IsSpace) {
```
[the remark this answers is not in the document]
wallyqs: Thanks Ivan, we eventually went back into that direction to avoid breaking changes on the patch release.
Adding the check on INFO would make leafnodes prevent being able to connect to clusters/gateways names that have spaces which although we would be disallowing in v2.11, does not seem to cause issues so leaving that for v2.10 for now.
About the server name being reused for the cluster name, added that as a test here but I think we could add a warning during the config check?
https://github.com/nats-io/nats-server/blob/3833649378d36a407b045adeb288c534465a0208/server/jetstream_cluster_4_test.go#L2104-L2109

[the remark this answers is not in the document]
kozlovic: I was just saying that users may get failure to connect but may not understand that it is due to the server name, since they did not specify a cluster name. Maybe [here](https://github.com/nats-io/nats-server/blob/92a857907237f1d35ab1ebc7c42e3bd41cd76008/server/server.go#L744) when we use the server name, you could add the check and indeed issue a warning if it contains spaces/new_lines?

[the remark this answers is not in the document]
wallyqs: Updated for it to show a warning:

```
[61124] 2024/08/06 06:04:13.218801 [INF] Starting nats-server
[61124] 2024/08/06 06:04:13.219424 [INF]   Version:  2.10.19-RC.2
[61124] 2024/08/06 06:04:13.219428 [INF]   Git:      [not set]
[61124] 2024/08/06 06:04:13.219431 [INF]   Cluster:  L E A F 2
[61124] 2024/08/06 06:04:13.219433 [INF]   Name:     L E A F 2
[61124] 2024/08/06 06:04:13.219435 [INF]   ID:       NCYPTW4H3TEKDHA2H46FCM6WTX7GOSSTQ3LALGKBUTUPGZ4P7XH37GSD
[61124] 2024/08/06 06:04:13.219466 [INF] Using configuration file: leaf.conf
[61124] 2024/08/06 06:04:13.220748 [INF] Starting http monitor on 127.0.0.1:56401
[61124] 2024/08/06 06:04:13.220925 [INF] Listening for leafnode connections on 127.0.0.1:56402
[61124] 2024/08/06 06:04:13.221059 [INF] Advertise address for leafnode is set to 127.0.0.1:56402
[61124] 2024/08/06 06:04:13.221403 [WRN] Server name has spaces and used as the cluster name, leaf remotes may not connect properly
[61124] 2024/08/06 06:04:13.221486 [INF] Listening for client connections on 127.0.0.1:56403
```

kozlovic (approved):
LGTM

neilalexander (approved):
LGTM

PR #5734, opened 2024-08-01
stree: Make `Iter` faster by using generic sorting function

neilalexander:
This is roughly 4x faster since there are no interface conversions and reduces allocations.


derekcollison (approved):
LGTM

PR #5735, opened 2024-08-01
Use `HasInterest` in place of `Match` where `SublistResult` not needed

neilalexander:
This reduces allocations and potentially speeds up some hot paths, such as in consumer waiting and gateway reply mapping.


derekcollison (approved):
LGTM

PR #5736, opened 2024-08-01
Optimise sublist matching for gateways, client replies

neilalexander:
This should reduce the number of string allocations on some gateway hot paths quite considerably.

Note that `MatchBytes()` passes in a new option to the internal `match()` function to ensure a copy is taken when writing to the cache to ensure we don't create race conditions from memory reuse. This doesn't affect `Match()` so the behaviour there is unchanged.


derekcollison (approved):
LGTM

PR #5743, opened 2024-08-02
Add support for collecting CPU profiles via `profilez`

neilalexander:
This makes it possible to collect a CPU profile via `profilez`. The CLI would need to be updated to add support for this since it requires sending a `"duration"`.


derekcollison (approved):
LGTM

PR #5744, opened 2024-08-02
Optimise `getAccAndResultFromCache` by reducing allocations

neilalexander:
We'll still copy the string when asking the resolver, as we can't be sure whether a resolver will want to hold onto the string or not, but this should lead to improvements on the hot paths for routes and gateways when the cache doesn't contain the entries we want.


derekcollison (approved):
LGTM

PR #5746, opened 2024-08-03
(2.11) [IMPROVED] Routing: reduce chances of duplicate implicit routes

kozlovic:
The PR#5602 solved the issue for typical seed setups, but it was found that the test `TestStressChainedSolicitWorks` would sometimes fail. This is a situation where we have S4->S3->S2->S1 and all servers start at the same type. This is not a typical setup, but still the regular gossip would allow the creation of the full mesh, so this PR helps in those situations.

Related to #5602


derekcollison (approved):
LGTM

PR #5750, opened 2024-08-05
Limit vector count in client outbound flushes

neilalexander:
This is a simpler PR than #5749 which just limits at the point of writing rather than at the point of buffering/framing. It allows us to set the deadline for each batch.

### DIFF
--- a/server/client_test.go
+++ b/server/client_test.go
@@ -2137,7 +2137,7 @@ type testConnWritePartial struct {
 func (c *testConnWritePartial) Write(p []byte) (int, error) {
 	n := len(p)
 	if c.partial {
-		n = 15
+		n = n/2 + 1
 	}
 	return c.buf.Write(p[:n])
 }
--- a/server/client_test.go
+++ b/server/client_test.go
@@ -3055,3 +3055,71 @@ func TestInProcessAllowedConnectionType(t *testing.T) {
 		})
 	}
 }
+
+func TestClientFlushOutboundNoSlowConsumer(t *testing.T) {
+	opts := DefaultOptions()
+	opts.MaxPending = 1024 * 1024 * 140 // 140MB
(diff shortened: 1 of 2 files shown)


derekcollison (commented):
LGTM but let's add a test around queueing up more than 64MB and make sure we do not fail protocol after.

server/client_test.go
```diff
@@ -3055,3 +3055,70 @@ func TestInProcessAllowedConnectionType(t *testing.T) {
...
+	_, err = nca.Subscribe("test", func(msg *nats.Msg) {
+		wait <- nil
+	})
+	require_NoError(t, err)
+
+	// Publish 128MB of data onto the test subject. This will
+	// mean that the outbound queue for nca has more than 64MB,
+	// which is the max we will send into a single writev call.
+	for i := 0; i < 8; i++ {
+		require_NoError(t, ncb.Publish("test", make([]byte, 1024*1024*16))) // 16MB
```
derekcollison, line 3094: Can probably do the make once, or use bytes.Repeat but outside loop.

server/client_test.go
```diff
@@ -3055,3 +3055,70 @@ func TestInProcessAllowedConnectionType(t *testing.T) {
...
+	ca.mu.Lock()
+	pba := ca.out.pb
+	ca.mu.Unlock()
+	require_True(t, pba > 1024*1024*64)
+
+	// Wait for our messages to be delivered. This will take
+	// a few seconds as the client is limited to 8MB/s, so it
+	// can't deliver messages to us as quickly as the other
+	// client can publish them.
+	msgs := 0
```
derekcollison, line 3115: var msgs int

derekcollison (commented):
Small comments, but otherwise LGTM.

derekcollison (approved):
LGTM

PR #5753, opened 2024-08-05
Allow $G to be listed in `accounts` option

### LINKED ISSUE #4138: Upgrade path for implicit global account to multi-account configuration
#### Use Case:

When a server is configured with the `authorization` block (implicitly using the global account `$G`), it is currently not possible to declare a user for the system account. This requires the `accounts` block to be defined either specifying a user for `$SYS` (implicit system account name) or a new account declared via the `system_account` option.

However, migrating from the `authorization` block to `accounts` currently not possible (whether the system account is in need or not). Declaring `$G` as an account in `accounts` currently errors indicating `$G` is a reserved account. As a result, a new account name would need to be used when moving to `accounts`.

This is problematic specifically for JetStream since account names are used as the folders on disk to store stream data. The workaround would be to backup and restore the data from global account to the new account, however this requires downtime (and not a nice user experience).

#### Proposed Change:

Either support `$G` in the `accounts` block to support this migration path or introduce a `global_account` config option analogous to `system_account` to explicitly map a new name in `accounts` to the underlying implicit account, e.g. `$G` and `$SYS`.

Anyone who starts out in single account mode and wants to expand to multiple accounts or forgets/is unaware of needing to explicitly enable the system account with users.

#### Alternative Approaches

Backup and restore JetStream data to a new account.

bruth:
Fix #4138

### DIFF
--- a/server/accounts_test.go
+++ b/server/accounts_test.go
@@ -388,7 +388,7 @@ func TestAccountFromOptions(t *testing.T) {
 	s := New(&opts)
 	defer s.Shutdown()
 
-	ta := s.numReservedAccounts() + 2
+	ta := 2
 	if la := s.numAccounts(); la != ta {
 		t.Fatalf("Expected to have a server with %d active accounts, got %v", ta, la)
 	}
--- a/server/opts.go
+++ b/server/opts.go
@@ -3108,12 +3108,6 @@ func parseAccounts(v any, opts *Options, errors *[]error, warnings *[]error) err
 		for _, n := range v.([]any) {
 			tk, name := unwrapValue(n, &lt)
 			ns := name.(string)
-			// Check for reserved names.
-			if isReservedAccount(ns) {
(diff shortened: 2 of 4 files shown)


derekcollison (commented):
Prefer to suppress $G in multi-account setups. What prompted this?

bruth:
I referenced the issue above, #4138. I stumbled upon it again, so I took a few minutes to investigate what it would take to support it (hence the draft PR) and the checks that I removed did not appear to guard against anything meaningful.

The justification in the code for the `globalAccountName` is clear: https://github.com/nats-io/nats-server/blob/main/server/accounts.go#L41-L43

Your last comment to the linked issue referenced the demo config, but that uses a different account name `default` for the `no_auth_user`. Right now, you can't use `$G` in the `accounts` block which has caused issues for people transitioning from single account to multi-account.

There is one relevant test that I need to fix in the test failure, but that simply asserts the number of accounts implicitly including the global account.

neilalexander:
Closing as old/stale.

PR #5755, opened 2024-08-05
[FIXED] False positives on cyclic import maps with subject mappings.

derekcollison:
When checking for import cycles, only need to check 'from', not the 'to' which could lead to false positives.
However we need to pre-transform the 'from' token in case it contains transform directives, e.g. $1, {{}}.


neilalexander (approved):
LGTM

PR #5757, opened 2024-08-07
Upgrade sort.Slice with slices.sort

jnmoyne:
Replace usage of the new sort package with the slices package (Go 1.21) where it is faster (i.e. when it has to do a conversion from `interface{}`).
Use the new (Go 1.21) cmp.Compare in the sort functions.
Also fixes a variable shadowing that somehow wasn't being detected when using `sort.Slices` on that shadowing variable.

### DIFF
--- a/server/accounts.go
+++ b/server/accounts.go
@@ -387,7 +387,7 @@ func (a *Account) updateRemoteServer(m *AccountNumConns) []*client {
 	var clients []*client
 	if mtce {
 		clients = a.getClientsLocked()
-		slices.SortFunc(clients, func(i, j *client) int { return -i.start.Compare(j.start) }) // reserve order
+		slices.SortFunc(clients, func(i, j *client) int { return -i.start.Compare(j.start) }) // reserve
 		over := (len(a.clients) - int(a.sysclients) + int(a.nrclients)) - int(a.mconns)
 		if over < len(clients) {
 			clients = clients[:over]
--- a/server/accounts.go
+++ b/server/accounts.go
@@ -15,6 +15,7 @@ package server
 
 import (
 	"bytes"
+	"cmp"
 	"encoding/hex"
 	"errors"
 	"fmt"
@@ -25,7 +26,7 @@ import (
 	"net/http"
 	"net/textproto"
 	"reflect"
-	"sort"
+	"slices"
 	"strconv"
 	"strings"
 	"sync"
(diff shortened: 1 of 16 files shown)


server/accounts.go
```diff
@@ -385,10 +386,8 @@ func (a *Account) updateRemoteServer(m *AccountNumConns) []*client {
 	// conservative and bit harsh here. Clients will reconnect if we over compensate.
 	var clients []*client
 	if mtce {
-		clients := a.getClientsLocked()
-		sort.Slice(clients, func(i, j int) bool {
-			return clients[i].start.After(clients[j].start)
-		})
+		clients = a.getClientsLocked()
+		slices.SortFunc(clients, func(i, j *client) int { return -i.start.Compare(j.start) }) // reserve order
```
neilalexander, line 390: "reverse"

server/filestore.go
```diff
@@ -1079,17 +1080,14 @@ func (fs *fileStore) addLostData(ld *LostStreamData) {
...
-		}
-		if tseq > seq {
-			return +1
-		}
-		return 0
+	i := slices.IndexFunc(ld.Msgs, func(i uint64) bool {
+		return i == seq
 	})
-	return i, found
+	if i == -1 {
```
neilalexander, line 1086: Simplify to `return i, i > -1`?

server/jetstream_cluster.go
```diff
@@ -5908,15 +5909,15 @@ func (cc *jetStreamCluster) selectPeerGroup(r int, cluster string, cfg *StreamCo
...
 	// Sort based on available from most to least, breaking ties by number of total streams assigned to the peer.
-	sort.Slice(nodes, func(i, j int) bool {
-		if nodes[i].avail == nodes[j].avail {
-			return nodes[i].ns < nodes[j].ns
+	slices.SortFunc(nodes, func(i, j wn) int {
+		if i.avail == j.avail {
+			return cmp.Compare(i.ns, j.ns)
 		}
-		return nodes[i].avail > nodes[j].avail
+		return cmp.Compare(i.avail, j.avail)
```
neilalexander, line 5916: Should be reverse sort order on the second `cmp.Compare` here.

neilalexander (changes requested):
Just one or two minor tweaks needed.

server/accounts.go
jnmoyne: You are  just asking to update the comment, right?

neilalexander: Yep, should have been clearer, sorry!

derekcollison:
Where are we on this one?

neilalexander (approved):
LGTM

PR #5760, opened 2024-08-07
In client and leafnode results cache, populate new entry after pruning

neilalexander:
Otherwise we might add the entry into the cache and then immediately prune it afterwards, depending on luck/map iteration order.


derekcollison (approved):
LGTM

PR #5761, opened 2024-08-07
[FIXED] JetStream: sending reply to Ack before processed (standalone)

### LINKED ISSUE #5702: NATS JetStream >= 2.10.17 slows down under load in a specific situation
### Observed behavior

In recent NATS JetStream versions (2.10.17, 2.10.18 and 2.11.0-dev), when I publish messages at a relatively high rate into a stream with WorkQueue retention policy, the stream operations start to slow down and time out after a while. When I test the same example against NATS 2.10.12, it runs to completion with no errors.

### Expected behavior

I would expect the performance to be the same that I observe on NATS 2.10.12.

### Server and client version

I tested the following server versions of NATS: 2.10.12, 2.10.17, 2.10.18 and the development version (2.11.0-dev) in the main branch.

In version 2.10.12, I observed normal performance. In all the other versions, I observed the performance degradation.

Go client (github.com/nats-io/nats.go): v1.36.0.

### Host environment

OS: Ubuntu 22.04.1 LTS on WSL2 (Windows 10)
CPU: Intel(R) Core(TM) i3-3220 CPU @ 3.30GHz
RAM: 8.00 GB

### Steps to reproduce

1. Run `nats-server -js` with NATS server version 2.10.18.
2. Execute the script whose listing is given below. You should see several lines of the type `error fetching from workqueue nats: timeout` and `error acking msg from workqueue nats: timeout`.
3. Run `nats-server -js` with NATS server version 2.10.12.
4. Execute the script again against the older NATS version. You should see the following output when the script is done executing:

    ```
    Finished populating work queue 100000
    Final stream size 100000
    ```

The script below creates a WorkQueue stream with 100000 items and a durable consumer that runs for a limited number of iterations where it continually 1) fetches the next item from the work queue, 2) acks that item and then 3) publishes that item back to the work queue. When all the processing is done, it prints the number of items in the stream, which should be 100000.

```go
package main

import (
	"fmt"
	"strconv"
	"sync"
	"time"

	"github.com/nats-io/nats.go"
)

func main() {
	var nc *nats.Conn
	var js nats.JetStreamContext
	nc, err := nats.Connect("127.0.0.1")
	if err != nil {
		panic(err)
	}
	defer nc.Close()
	if js, err = nc.JetStream(); err != nil {
		panic(err)
	}

	// Delete and recreate test streams / buckets.

	_ = js.DeleteStream("TEST_WORKQUEUE")

		Name:                 "TEST_WORKQUEUE",
		Discard:              nats.DiscardNew,
		MaxMsgsPerSubject:    1,
		DiscardNewPerSubject: true,
		Retention:            nats.WorkQueuePolicy,
		Subjects:             []string{"TEST_WORKQUEUE.>"},
	})
	if err != nil {
		panic(err)
	}

	// Populate work queue

	NUM_MSGS := 100000

	var wg sync.WaitGroup

	for i := 0; i < NUM_MSGS; i++ {
		wg.Add(1)
		i := i
		go func() {
			defer wg.Done()
			itemID := fmt.Sprintf("%d", i)
			js.Publish(fmt.Sprintf("TEST_WORKQUEUE.%s", itemID), []byte(itemID))
		}()
	}
	wg.Wait()

	streamInfo, err := js.StreamInfo("TEST_WORKQUEUE")
	if err != nil {
		panic(err)
	}

	fmt.Println("Finished populating work queue", streamInfo.State.Msgs)

	// Create consumer that runs a loop where it
	// 1) fetches the next item from the work queue,
	// 2) acks that item,
	// 3) and then publishes that item back to the work queue.
	// The outer loop of the consumer runs for 7 iterations, each one processing a batch of 20000 items.

	if _, err = js.AddConsumer("TEST_WORKQUEUE", &nats.ConsumerConfig{
		Durable:       "testworkqueueconsumer",
		AckPolicy:     nats.AckExplicitPolicy,
		MaxWaiting:    20000,
		MaxAckPending: -1,
	}); err != nil {
		panic(err)
	}

	workqueueSub, err := js.PullSubscribe("TEST_WORKQUEUE.>", "testworkqueueconsumer", nats.BindStream("TEST_WORKQUEUE"))
	if err != nil {
		panic(err)
	}

	for j := 0; j < 7; j++ {
		NUM_ITEMS := 20000
		for i := 0; i < NUM_ITEMS; i++ {
			wg.Add(1)
			i := i
			go func() {
				defer wg.Done()
				msgs, err := workqueueSub.Fetch(1)
				if err != nil {
					fmt.Println("error fetching from workqueue", err)
					return
				}
				for {
					err = msgs[0].AckSync()
					if err != nil {
						fmt.Println("error acking msg from workqueue", err)
						continue
					}
					break
				}
				itemID, err := strconv.Atoi(string(msgs[0].Data))
				if err != nil {
					panic(err)
				}
				for {
					_, err = js.Publish(fmt.Sprintf("TEST_WORKQUEUE.%d", itemID), msgs[0].Data)
					if err != nil {
						fmt.Println("error publishing msg back to workqueue", err)
						continue
					}
					return
				}
			}()
			if i%1000 == 0 {
				time.Sleep(100 * time.Millisecond)
			}
		}
		wg.Wait()
	}
	wg.Wait()

	fmt.Println("Final stream size", streamInfo.State.Msgs)

}
```

kozlovic:
The server would send the reply back to the client before removing message from the stream, which could cause some issues if the client expected the message to be removed after the AckSync() call returns.
For instance, a WorkQueue that would have a max message of 1 per subject and discard new message would fail a client that got a message, ack it, and re-publish this message. Depending on timing, the client may get an error `maximum messages per subject exceeded`.

Used @pcsegal description of the problem and example code to write a test that demonstrated the issue before the fix.

Resolves #5702

### DIFF
--- a/server/consumer.go
+++ b/server/consumer.go
@@ -2113,9 +2113,10 @@ func (o *consumer) processAck(subject, reply string, hdr int, rmsg []byte) {
 
 	switch {
 	case len(msg) == 0, bytes.Equal(msg, AckAck), bytes.Equal(msg, AckOK):
-		o.processAckMsg(sseq, dseq, dc, reply, true)
-		// We handle replies for acks in updateAcks
-		skipAckReply = true
+		if !o.processAckMsg(sseq, dseq, dc, reply, true) {
+			// We handle replies for acks in updateAcks
+			skipAckReply = true
+		}
 	case bytes.HasPrefix(msg, AckNext):
 		o.processAckMsg(sseq, dseq, dc, _EMPTY_, true)
 		o.processNextMsgRequest(reply, msg[len(AckNext):])
@@ -2129,9 +2130,10 @@ func (o *consumer) processAck(subject, reply string, hdr int, rmsg []byte) {
 		if buf := msg[len(AckTerm):]; len(buf) > 0 {
 			reason = string(bytes.TrimSpace(buf))
 		}
-		o.processTerm(sseq, dseq, dc, reason, reply)
-		// We handle replies for acks in updateAcks
-		skipAckReply = true
+		if !o.processTerm(sseq, dseq, dc, reason, reply) {
+			// We handle replies for acks in updateAcks
+			skipAckReply = true
+		}
 	}
 
 	// Ack the ack if requested.
@@ -2481,9 +2483,12 @@ func (o *consumer) processNak(sseq, dseq, dc uint64, nak []byte) {
 }
 
 // Process a TERM
-func (o *consumer) processTerm(sseq, dseq, dc uint64, reason, reply string) {
+// Returns `true` if the ack was processed in place and the sender can now respond
+// to the client, or `false` if there was an error or the ack is replicated (in which
+// case the reply will be sent later).
+func (o *consumer) processTerm(sseq, dseq, dc uint64, reason, reply string) bool {
 	// Treat like an ack to suppress redelivery.
-	o.processAckMsg(sseq, dseq, dc, reply, false)
+	ackedInPlace := o.processAckMsg(sseq, dseq, dc, reply, false)
 
 	o.mu.Lock()
 	defer o.mu.Unlock()
@@ -2506,11 +2511,14 @@ func (o *consumer) processTerm(sseq, dseq, dc uint64, reason, reply string) {
 
 	j, err := json.Marshal(e)
 	if err != nil {
-		return
+		// We had an error during the marshal, so we can't send the advisory,
+		// but we still need to tell the caller that the ack was processed.
+		return ackedInPlace
 	}
 
 	subj := JSAdvisoryConsumerMsgTerminatedPre + "." + o.stream + "." + o.name
 	o.sendAdvisory(subj, j)
+	return ackedInPlace
 }
 
 // Introduce a small delay in when timer fires to check pending.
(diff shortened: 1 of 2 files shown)


server/consumer.go
```diff
@@ -2914,22 +2931,19 @@ func (o *consumer) processAckMsg(sseq, dseq, dc uint64, reply string, doSample b
...
-	// using the consumer lock to avoid a race.
-	retention := o.retention
-	clustered := o.node != nil
 	o.mu.Unlock()
 
-	// Let the owning stream know if we are interest or workqueue retention based.
-	// If this consumer is clustered this will be handled by processReplicatedAck
-	// after the ack has propagated.
-	if !clustered && mset != nil && retention != LimitsPolicy {
+	if ackInPlace {
```
wallyqs, line 2946: not needed to add `mset != nil` here?

wallyqs (approved):
LGTM

server/consumer.go
kozlovic: No, I had a warning (in VSCode) because the check is made at the top of the function that `mset` would never be `nil` (otherwise the function would have returned).

neilalexander (approved):
LGTM

PR #5763, opened 2024-08-08
Remove `ordered` constraint now that we can use `cmp.Ordered`

neilalexander:


wallyqs (approved):
LGTM

PR #5765, opened 2024-08-08
De-flake tests

neilalexander:
This PR tries to deflake some unit tests.


server/tav.idx
wallyqs, line 1: seems this file got it accidentally

neilalexander: Ay, good catch, thanks. Fixed.

derekcollison (approved):
LGTM

PR #5767, opened 2024-08-08
[FIXED] Bad meta state on restart could cause deletion of assets.

derekcollison:
When checking for orphaned assets in a cluster, make sure we are properly synched with the leader before proceeding.
When R>1 the server would rebuild them after getting the proper meta state, but in cases of R1 this could lead to dataloss.


wallyqs (approved):
LGTM

PR #5770, opened 2024-08-09
Test to demonstrate bug with Account limits being ignored

johnweldon:
introduced in 94fe77d6e8abe2620009a36af8675076f5e99178

Signed-off-by: John Weldon <jweldon@synadia.com>

### DIFF
--- a/server/jetstream_limits_test.go
+++ b/server/jetstream_limits_test.go
@@ -1,0 +1,67 @@
+package server
+
+import (
+	"crypto/rand"
+	"testing"
+)
+
+func TestJetstreamAccountFileStoreLimits(t *testing.T) {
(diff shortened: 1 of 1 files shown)


server/jetstream_limits_test.go
```diff
@@ -0,0 +1,67 @@
+package server
+
+import (
+	"crypto/rand"
+	"testing"
+)
+
+func TestJetstreamAccountFileStoreLimits(t *testing.T) {
```
wallyqs, line 8:
```suggestion
func TestJetStreamAccountFileStoreLimits(t *testing.T) {
```
we should move this one to jetstream_cluster_3_test.go below the TestJetStreamClusterLimitsBasedStreamFileStoreDesync test which is related

PR #5771, opened 2024-08-09
Fix R1 streams exceeding quota limits

wallyqs:
Fixes #5770 


neilalexander (approved):
LGTM

server/jetstream_cluster_3_test.go
```diff
@@ -6433,3 +6433,98 @@ Consume3:
 		t.Errorf("Consumers to same stream are at different sequences: %d vs %d", a, b)
 	}
 }
+
+func TestJetStreamAccountFileStoreLimits(t *testing.T) {
```
derekcollison, line 6437: Naming convention here may mean that Travis will not run this test. We need to check that.

server/jetstream_cluster_3_test.go
```diff
@@ -6433,3 +6433,98 @@ Consume3:
...
+		"R3": {
+			MaxMemory:    1 << 10,
+			MaxStore:     1 << 10,
+			MaxStreams:   -1,
+			MaxConsumers: -1,
+		},
+	}
+
+	// Update the limits in all servers.
+	for _, s := range c.servers {
```
derekcollison, line 6483: If you do this by hand just run a normal cluster, no need for separate conf.

derekcollison (changes requested):
Great find, a few comments.

wallyqs:
`TestJWTJetStreamTiers` is failing now so investigating...

wallyqs:
Fixed the test and now green ✅

PR #5775, opened 2024-08-09
(2.11) Add ability for IP queues to track total size

neilalexander:
You can optionally pass in an `ipQueue_SizeCalculation` function to an IP queue that fires each time an item is pushed or popped so that the total size can be tracked.

There is a slight performance impact but mostly it will depend on the complexity of the calculation function:
```
go test -v ./server -run=XXX -bench=IPQueueSizeCalculation -benchtime=10s
goos: darwin
goarch: arm64
pkg: github.com/nats-io/nats-server/v2/server
BenchmarkIPQueueSizeCalculation
BenchmarkIPQueueSizeCalculation/WithoutCalc
BenchmarkIPQueueSizeCalculation/WithoutCalc-24          396342910               30.75 ns/op      520.35 MB/s
BenchmarkIPQueueSizeCalculation/WithEmptyCalc
BenchmarkIPQueueSizeCalculation/WithEmptyCalc-24        377004524               33.91 ns/op      471.80 MB/s
BenchmarkIPQueueSizeCalculation/WithLenCalc
BenchmarkIPQueueSizeCalculation/WithLenCalc-24          375148896               33.76 ns/op      473.99 MB/s
PASS
ok      github.com/nats-io/nats-server/v2/server        48.150s
```

### DIFF
--- a/server/ipqueue.go
+++ b/server/ipqueue.go
@@ -14,6 +14,7 @@
 package server
 
 import (
+	"errors"
 	"sync"
 	"sync/atomic"
 )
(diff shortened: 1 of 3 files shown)


derekcollison:
What is delta in terms of performance hit?

neilalexander:
It depends on how complex the calc function is, but in the unit tests where it does a simple `len()`, the difference is roughly 10%.

server/ipqueue.go
```diff
@@ -28,36 +28,69 @@ type ipQueue[T any] struct {
...
 	m    *sync.Map
+	ipQueueOpts[T]
 }
 
-type ipQueueOpts struct {
-	maxRecycleSize int
+type ipQueueOpts[T any] struct {
+	mrs   int              // Max recycle size
+	calc  func(e T) uint64 // Calc function for tracking size
+	lsz   uint64           // Limit by total calculated size
```
derekcollison, line 40: Maybe msz and mlen might be better names?

server/ipqueue.go
```diff
@@ -66,10 +99,14 @@ func newIPQueue[T any](s *Server, name string, opts ...ipQueueOpt) *ipQueue[T] {
...
 // entry is the first to be added, and returns the length of the queue after
 // this element is added.
-func (q *ipQueue[T]) push(e T) int {
+func (q *ipQueue[T]) push(e T) (int, bool) {
 	var signal bool
 	q.Lock()
 	l := len(q.elts) - q.pos
+	if q.lelts > 0 && l == q.lelts {
+		q.Unlock()
+		return l, false
```
derekcollison, line 108: I think just returning an error here might be better so we can distinguish between which limit we hit. WDYT?

server/ipqueue.go
```diff
@@ -189,6 +241,12 @@ func (q *ipQueue[T]) len() int {
 	return l
 }
 
+// Returns the calculated size of the queue (if ipQueue_SizeCalculation has been
+// passed in), otherwise returns zero.
+func (q *ipQueue[T]) size() uint64 {
+	return q.sz.Load()
```
derekcollison, line 247: This the only reason we made it an atomic? Do we call this function alot?

server/ipqueue_test.go
```diff
@@ -42,7 +42,7 @@ func TestIPQueueBasic(t *testing.T) {
 	}
 
 	// Try to change the max recycle size
-	q2 := newIPQueue[int](s, "test2", ipQueue_MaxRecycleSize(10))
+	q2 := newIPQueue[int](s, "test2", ipQueue_MaxRecycleSize[int](10))
```
derekcollison, line 45: I like how this works, but the name of the options might need tweaking.

ipqMaxRecycleSize() - nix the long name and the '_' I think here.

server/ipqueue_test.go
```diff
@@ -290,15 +290,15 @@ func TestIPQueueRecycle(t *testing.T) {
 	for iter := 0; iter < 5; iter++ {
 		var sz int
 		for i := 0; i < total; i++ {
-			sz = q.push(i)
+			sz, _ = q.push(i)
```
derekcollison, line 293: If we only use sz in tests I would make the case that push just returns an err.

server/jetstream_api.go
```diff
@@ -863,7 +863,7 @@ func (js *jetStream) apiDispatch(sub *subscription, c *client, acc *Account, sub
 	// header from the msg body. No other references are needed.
 	// Check pending and warn if getting backed up.
 	const warnThresh = 128
-	pending := s.jsAPIRoutedReqs.push(&jsAPIRoutedReq{jsub, sub, acc, subject, reply, copyBytes(rmsg), c.pa})
+	pending, _ := s.jsAPIRoutedReqs.push(&jsAPIRoutedReq{jsub, sub, acc, subject, reply, copyBytes(rmsg), c.pa})
```
derekcollison, line 866: Ah I see we do use it here and for warning, so limit would not help us here.

derekcollison (changes requested):

server/ipqueue.go
neilalexander: I did consider it but it doesn't really matter to any of the call sites, the entry doesn't get queued either way. Can add it if you prefer but may not be needed.

server/ipqueue.go
neilalexander: Probably not anymore now that limits are set in the IPQ itself, I can put it back to non-atomic.

server/jetstream_api.go
neilalexander: Yep, this is the only place where the size is returned & used.

server/ipqueue.go
derekcollison: We need to know which limit for the log warning IMO. We will want to rate limit them of course but we should be clear which limit we hit, bytes or number of elements.

server/jetstream_api.go
derekcollison: Well still could be ok, if err then grab len() for the log warning since should not happen too often etc.. WDYT?

neilalexander: Left the return value here, since otherwise we need to take the lock again to check the length. Think it's better to avoid lock contention for now than optimise against a return val at this point.

derekcollison (approved):
LGTM

PR #5783, opened 2024-08-13
(2.11) [ADDED] LeafNode: Support for TLS handshake_first duration

### LINKED ISSUE #5772: Time duration for `handshake_first` option is not respected for leafnodes
### Observed behavior

According to this [ADR](https://github.com/nats-io/nats-architecture-and-design/blob/main/adr/ADR-40.md) passing a time duration for the `handshake_first` option, makes the server backward compatible because after the timeout the server is supposed to use the standard TLS flow.

This works for standard connections (from clients) but not for leafnodes.
If the leafnode doesn't declare the `handshake_first` field in the `remotes` section, its connection will be never accepted by the server

I'm able to connect to my server using `nats-cli` with or without the `--tlsfirst` flag (because of the timeout) but the leafnode can connect to the server only if it use the field in the remote section.

### Expected behavior

A leafnode connection without `handshake_first` is accepted (standard TLS flow) after the initial timeout

### Server and client version

nats-server: v2.10.18 (both server and leafnode)
nats-cli: 0.1.5

### Host environment

_No response_

### Steps to reproduce

Main server has this configuration
```
TLS = {
  cert_file: $CERT_FILE
  key_file: $KEY_FILE
  ca_file: $CA_FILE
  verify_and_map: true
  timeout: 2
  handshake_first: 500ms
}

tls : $TLS

leafnodes {
    port: 20023
    tls: $TLS
}
```

The leafnode has this one
```
TLS = {
  cert_file: $CERT_FILE
  key_file: $KEY_FILE
  ca_file: $CA_FILE
  verify_and_map: true
  handshake_first: 500ms
}

tls : $TLS

leafnodes {
  remotes: [
    {
      url: $SERVER
      tls {
         cert_file: $CERT_FILE
         key_file: $KEY_FILE
         ca_file: $CA_FILE
         handshake_first: true
      }
      account: $MY_ACCOUNT
    }
  ] 
}
```

In this case works, but if I remove the `handshake_first` from `remotes` section it doesn't

```
TLS = {
  cert_file: $CERT_FILE
  key_file: $KEY_FILE
  ca_file: $CA_FILE
  verify_and_map: true
  handshake_first: 500ms
}

tls : $TLS

leafnodes {
  remotes: [
    {
      url: $SERVER
      tls {
         cert_file: $CERT_FILE
         key_file: $KEY_FILE
         ca_file: $CA_FILE
      }
      account: $FILEWAVE_ACCOUNT
    }
  ] 
}
```

kozlovic:
It will now be possible to configure the hub's `handshake_first` field in the `tls{}` block as `auto` or a duration (e.g. `"250ms"`).

When that is the case, if the server does not receive data within this time frame, it will revert to send the `INFO` protocol in plaintext because it expects the remote LeafNode to perform the TLS handshake only after receiving it.

Resolves #5772.

### DIFF
--- a/server/leafnode.go
+++ b/server/leafnode.go
@@ -20,6 +20,7 @@ import (
 	"encoding/base64"
 	"encoding/json"
 	"fmt"
+	"io"
 	"math/rand"
 	"net"
 	"net/http"
@@ -987,8 +988,11 @@ func (s *Server) createLeafNode(conn net.Conn, rURL *url.URL, remote *leafNodeCf
 	c.initClient()
 	c.Noticef("Leafnode connection created%s %s", remoteSuffix, c.opts.Name)
 
-	var tlsFirst bool
-	var infoTimeout time.Duration
+	var (
+		tlsFirst         bool
+		tlsFirstFallback time.Duration
+		infoTimeout      time.Duration
+	)
 	if remote != nil {
 		solicited = true
 		remote.Lock()
@@ -1005,6 +1009,10 @@ func (s *Server) createLeafNode(conn net.Conn, rURL *url.URL, remote *leafNodeCf
 		c.flags.set(expectConnect)
 		if ws != nil {
 			c.Debugf("Leafnode compression=%v", c.ws.compress)
+		}
+		tlsFirst = opts.LeafNode.TLSHandshakeFirst
+		if f := opts.LeafNode.TLSHandshakeFirstFallback; f > 0 {
+			tlsFirstFallback = f
 		}
 	}
 	c.mu.Unlock()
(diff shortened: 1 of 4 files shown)


kozlovic:
@bruth Not sure if this would be a 2.11 or not. This is now supporting a "TLS first fallback delay" which means that remotes that are not updated to be configured to "handshake_first" could still connect if the "hub" is configured with a delay, instead of the boolean.
If you think it should be only 2.11, then I can force-push to update the PR title (I know we can edit in GitHub, but once merged, the list of commit shows the original commit title, not the PR as far as I know).

bruth:
Thanks @kozlovic, yes this would be for 2.11

derekcollison (approved):
LGTM

mcosta74:
He'll @kozlovic , @derekcollison . Any ETA for v2.11 ?

derekcollison:
Ite merged into main, and will be in nightly builds tonight. 2.11 is loosely targeted for Sep.

mcosta74:
Ok thanks, September should be OK for our next release.

alexbozhenko:
I do not see this option mentioned here https://docs.nats.io/running-a-nats-service/configuration/leafnodes/leafnode_conf.
Or is this niche/rare scenario when someone would need to use this option, so this one doesn't need to be mentioned in the docs?

PR #5785, opened 2024-08-13
Fix consumer start sequence when sequence not yet in stream

neilalexander:
The consumer `sseq` was being clipped back to the range between the stream `FirstSeq` and `LastSeq` even when the consumer start sequence was provided, so it wasn't possible to provide a start sequence that didn't exist in the stream yet. This PR changes that to now obey the start sequence even in that case.

Note that this problem will still exist with the consumer start timestamp, as that's predicated on being able to match the timestamp to a sequence number in the stream, which it won't be able to do if the stream doesn't contain any sequences for that time.

### DIFF
--- a/server/consumer.go
+++ b/server/consumer.go
@@ -4983,12 +4983,16 @@ func (o *consumer) selectStartingSeqNo() {
 			o.sseq = o.cfg.OptStartSeq
 		}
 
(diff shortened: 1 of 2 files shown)


derekcollison (approved):
LGTM - Thanks

PR #5792, opened 2024-08-15
NRG (2.11): Don't revert `pterm` to beginning of log when installing snapshots

neilalexander:
Otherwise on a restart with a restore from snapshot, we can revert `pterm` back to an earlier position than it should be, which can create consistency issues and failed catchups.


derekcollison (approved):
LGTM

PR #5796, opened 2024-08-16
Stream ingest rate limiting

neilalexander:
Built on top of #5575. Tries to prevent the stream `msgs` queue from becoming overwhelmed by core NATS publishes. In this case, where a reply subject is known, the sender will receive a 429 "Too Many Requests". Otherwise it's rate-logged.

Two new configuration options are added to the JetStream block: `max_buffered_size` and `max_buffered_msgs`. If not configured, defaults are used.

### DIFF
--- a/server/opts.go
+++ b/server/opts.go
@@ -331,6 +331,8 @@ type Options struct {
 	JetStreamLimits       JSLimitOpts
 	JetStreamTpm          JSTpmOpts
 	JetStreamMaxCatchup   int64
+	StreamMaxBufferedMsgs int               `json:"-"`
+	StreamMaxBufferedSize int64             `json:"-"`
 	StoreDir              string            `json:"-"`
 	SyncInterval          time.Duration     `json:"-"`
 	SyncAlways            bool              `json:"-"`
@@ -2373,6 +2375,18 @@ func parseJetStream(v any, opts *Options, errors *[]error, warnings *[]error) er
 					return &configErr{tk, fmt.Sprintf("%s %s", strings.ToLower(mk), err)}
 				}
 				opts.JetStreamMaxCatchup = s
+			case "max_buffered_size":
+				s, err := getStorageSize(mv)
+				if err != nil {
+					return &configErr{tk, fmt.Sprintf("%s %s", strings.ToLower(mk), err)}
+				}
+				opts.StreamMaxBufferedSize = s
+			case "max_buffered_msgs":
+				mlen, ok := mv.(int64)
+				if !ok {
+					return &configErr{tk, fmt.Sprintf("Expected a parseable size for %q, got %v", mk, mv)}
+				}
+				opts.StreamMaxBufferedMsgs = int(mlen)
 			default:
 				if !tk.IsUsedVariable() {
 					err := &unknownConfigFieldErr{
--- a/server/stream.go
+++ b/server/stream.go
@@ -218,6 +218,12 @@ type ExternalStream struct {
 	DeliverPrefix string `json:"deliver"`
 }
 
+// For managing stream ingest.
+const (
+	streamDefaultMaxQueueMsgs  = 10_000
+	streamDefaultMaxQueueBytes = 1024 * 1024 * 128
+)
+
 // Stream is a jetstream stream of messages. When we receive a message internally destined
 // for a Stream we will direct link from the client to this structure.
 type stream struct {
@@ -576,6 +582,16 @@ func (a *Account) addStreamWithAssignment(config *StreamConfig, fsConfig *FileSt
 	c := s.createInternalJetStreamClient()
 	ic := s.createInternalJetStreamClient()
 
+	// Work out the stream ingest limits.
+	mlen := s.opts.StreamMaxBufferedMsgs
+	msz := uint64(s.opts.StreamMaxBufferedSize)
+	if mlen == 0 {
+		mlen = streamDefaultMaxQueueMsgs
+	}
+	if msz == 0 {
+		msz = streamDefaultMaxQueueBytes
+	}
+
 	qpfx := fmt.Sprintf("[ACC:%s] stream '%s' ", a.Name, config.Name)
 	mset := &stream{
 		acc:       a,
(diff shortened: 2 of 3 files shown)


server/stream.go
```diff
@@ -576,6 +582,16 @@ func (a *Account) addStreamWithAssignment(config *StreamConfig, fsConfig *FileSt
 	c := s.createInternalJetStreamClient()
 	ic := s.createInternalJetStreamClient()
 
+	// Work out the stream ingest limits.
+	mlen := s.opts.JetStreamMaxBufferMsgs
+	msz := uint64(s.opts.JetStreamMaxBufferSize)
```
wallyqs, line 587: buffer size int but uint64 here will overflow?

neilalexander: `int64` to `uint64`, overflow I suppose is possible but only over 9 million terabytes.

wallyqs: ah got it confused with `JetStreamMaxBufferMsgs` above which is an int

wallyqs: no wait, still if someone sets -1 on the int64 it will overflow here. Need to be explicit whether negative means unlimited or protect against that

neilalexander: Yes, at the moment nothing stops you from setting an arbitrarily large limit normally, possibly even more than the system has available, so the negative underflow is really no worse than doing that.

server/stream.go
```diff
@@ -576,6 +582,16 @@ func (a *Account) addStreamWithAssignment(config *StreamConfig, fsConfig *FileSt
...
+	if mlen == 0 {
+		mlen = streamDefaultMaxQueueMsgs
+	}
+	if msz == 0 {
+		mlen = streamDefaultMaxQueueBytes
```
wallyqs, line 592: should be this maybe?
```suggestion
		msz = streamDefaultMaxQueueBytes
```

server/stream.go
wallyqs: yeah good point

server/stream.go
neilalexander: Good catch, fixed.

server/stream.go
wallyqs: so in that sense -1 underflow as a feature means that it is kind of infinite and disables the default

server/stream.go
wallyqs: Different test needed maybe to cover `JetStreamMaxBufferSize` taking effect?

neilalexander: Done.

wallyqs (approved):
LGTM

server/opts.go
```diff
@@ -277,87 +277,89 @@ type AuthCallout struct {
...
+	JetStreamDomain        string        `json:"-"`
+	JetStreamExtHint       string        `json:"-"`
+	JetStreamKey           string        `json:"-"`
+	JetStreamOldKey        string        `json:"-"`
+	JetStreamCipher        StoreCipher   `json:"-"`
+	JetStreamUniqueTag     string
+	JetStreamLimits        JSLimitOpts
+	JetStreamTpm           JSTpmOpts
+	JetStreamMaxCatchup    int64
+	JetStreamMaxBufferMsgs int
```
derekcollison, line 334: Should it be StreamMaxBufferedMsgs etc?

server/opts.go
```diff
@@ -277,87 +277,89 @@ type AuthCallout struct {
...
+	JetStreamMaxBufferSize int64
```
derekcollison, line 335: Should we have explicit json annotation tag?

server/opts.go
```diff
@@ -2373,6 +2375,14 @@ func parseJetStream(v any, opts *Options, errors *[]error, warnings *[]error) er
...
 				}
 				opts.JetStreamMaxCatchup = s
+			case "max_buffer_size":
+				s, err := getStorageSize(mv)
+				if err != nil {
+					return &configErr{tk, fmt.Sprintf("%s %s", strings.ToLower(mk), err)}
+				}
+				opts.JetStreamMaxBufferSize = s
+			case "max_buffer_msgs":
+				opts.JetStreamMaxBufferMsgs = int(mv.(int64))
```
derekcollison, line 2385: Will panic is presented as a string "22"

derekcollison (commented):
in general LGTM, a few small comments / questions.

server/opts.go
neilalexander: There's a comment further up on that struct saying the JSON tags were deprecated, so wasn't sure if it was needed.

server/opts.go
neilalexander: Can be yes, happy with any naming suggestions.

server/opts.go
neilalexander: I can fix that, but wondering if that's typical?

server/opts.go
```diff
@@ -331,6 +331,8 @@ type Options struct {
 	JetStreamLimits       JSLimitOpts
 	JetStreamTpm          JSTpmOpts
 	JetStreamMaxCatchup   int64
+	StreamMaxBufferedMsgs int
```
derekcollison, line 334: Do we need explicit json annotations here? What is the default name? Does it move camel case breaks to `_`?

server/opts.go
```diff
@@ -2373,6 +2375,14 @@ func parseJetStream(v any, opts *Options, errors *[]error, warnings *[]error) er
...
 				}
 				opts.JetStreamMaxCatchup = s
+			case "max_buffered_size":
+				s, err := getStorageSize(mv)
+				if err != nil {
+					return &configErr{tk, fmt.Sprintf("%s %s", strings.ToLower(mk), err)}
+				}
+				opts.StreamMaxBufferedSize = s
+			case "max_buffered_msgs":
+				opts.StreamMaxBufferedMsgs = int(mv.(int64))
```
derekcollison, line 2385: This will panic is passed a string.. We should trap that if possible to not panic the server.

server/opts.go
neilalexander: The default JSON name would be the variable name exactly, but as I mentioned above there's a comment on that srtuct saying the JSON tags are deprecated as that struct is no longer used for monitoring, so I think it's irrelevant?

derekcollison: We should be consistent, either mark it as non exported or pick a name in case it is IMO. I prefer camel case -> snake case.

derekcollison (approved):
LGTM

PR #5812, opened 2024-08-20
[IMPROVED] When we detect a bump in clfs try to snapshot (if prior was zero force).

derekcollison:
Test from @MauriceVanVeen 


MauriceVanVeen (approved):
LGTM

PR #5815, opened 2024-08-21
Fix log backup file extension

zakk616:
When `logfile_size_limit` is used the log rotation creates new log files and renames old ones by adding a suffix:

.yyyy.mm.dd.hh.mm.ss.micros

the file extension becomes difficult to configure for compressing and archiving based on file extension using tools such as `logrotate ` 

Logs will now be backed up with an extension of `.log`. so these files can be compressed or moved to archiveing paths based on file extensions.

**Before:**
```
nats-server.log
nats-server.log.2024.08.21.14.37.57.558600300
nats-server.log.2024.08.21.14.40.28.743600300
```

**After:**
```
nats-server.log
nats-server.2024.08.21.14.37.57.558600300.log
nats-server.2024.08.21.14.40.28.743600300.log
```

### DIFF
--- a/logger/log.go
+++ b/logger/log.go
@@ -254,10 +254,12 @@ func (l *fileLogger) Write(b []byte) (int, error) {
 				return n, err
 			}
 			fname := l.f.Name()
+			ext := filepath.Ext(fname)
+			base := strings.TrimSuffix(fname, ext)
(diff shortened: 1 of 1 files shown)


neilalexander:
@ripienaar Any thoughts on this?

PR #5816, opened 2024-08-21
[IMPROVED] Handling of StatsZ/Ping and Profilez system events

kozlovic:
Use a dedicated internal receive queue to dispatch the StatsZ/Ping events, and have Profilez run in its own go routine (not even using internal receive queues) because execution of profilez() can take several seconds.

### DIFF
--- a/server/events.go
+++ b/server/events.go
@@ -130,6 +130,7 @@ type internal struct {
 	replies        map[string]msgHandler
 	sendq          *ipQueue[*pubMsg]
 	recvq          *ipQueue[*inSysMsg]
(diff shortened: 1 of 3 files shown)


kozlovic:
@derekcollison @wallyqs I usually prefer to have the run from Travis on the branch's push to run before submitting the PR, but something must be wrong with GitHub/Travis, since that did not happen. So I opened the PR to see if it would trigger maybe just on "pull" and no longer on "push", but it seems that there is still an issue. So apologies if we find some failing tests due to this change since I was not able to get a run before submitting.

derekcollison (approved):
LGTM - Thanks!

PR #5819, opened 2024-08-22
[IMPROVED] Do not load blocks unnecessarily when checking if we can skip ahead in LoadNextMsg().

derekcollison:
When we would miss a LoadNextMsg() from the block associated with the starting sequence, we would check if we could skip ahead. The logic for this could load old blocks, behind where we were, if the psim fblk was outdated, causing memory bloat for cache expiration time intervals.

We reworked this to load no blocks in the skip check since this is hot path for LoadNextMsg() and the linear scan, if applicable, properly will expire anything loaded that did not contain a match.

I tested that benchmarks were not affected.

`go test -v -run=XXX --bench=Benchmark_FileStoreLoadNext`

### DIFF
--- a/server/filestore.go
+++ b/server/filestore.go
@@ -1236,6 +1236,13 @@ func (mb *msgBlock) convertToEncrypted() error {
 	return nil
 }
 
+// Return the mb's index.
+func (mb *msgBlock) getIndex() uint32 {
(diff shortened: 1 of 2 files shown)


wallyqs (approved):
LGTM

PR #5820, opened 2024-08-23
[FIXED] JetStream: data race between consumer and stream when updated

kozlovic:
Some tests were showing often a race between updateWithAdvisory() and isFiltered().

There may be other data races in other files that read the mset's configuration without proper locking, but this PR addresses specifically issues between consumer (consumer.go) and stream.

### DIFF
--- a/server/consumer.go
+++ b/server/consumer.go
@@ -717,8 +717,7 @@ func (mset *stream) addConsumerWithAssignment(config *ConsumerConfig, oname stri
 	}
 
 	mset.mu.RLock()
-	s, jsa, cfg, acc := mset.srv, mset.jsa, mset.cfg, mset.acc
-	retention := cfg.Retention
+	s, jsa, cfg, acc := mset.srv, mset.jsa, mset.getCfg(), mset.acc
 	mset.mu.RUnlock()
 
 	// If we do not have the consumer currently assigned to us in cluster mode we will proceed but warn.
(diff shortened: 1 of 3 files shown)


server/consumer.go
```diff
@@ -717,8 +717,7 @@ func (mset *stream) addConsumerWithAssignment(config *ConsumerConfig, oname stri
 	}
 
 	mset.mu.RLock()
-	s, jsa, cfg, acc := mset.srv, mset.jsa, mset.cfg, mset.acc
-	retention := cfg.Retention
+	s, jsa, cfg, acc := mset.srv, mset.jsa, mset.getCfg(), mset.acc
```
kozlovic, line 720: `meet.getCfg()` may not even be needed here, since meet.setCfg() is still called under mset.mu.Lock().

server/stream.go
```diff
@@ -2029,6 +2030,39 @@ func (mset *stream) updateWithAdvisory(config *StreamConfig, sendAdvisory bool)
...
+	mset.cfgMu.Unlock()
+}
+
+// The stream configuration is a structure. Normally, code that would do
+// something like cfg := mset.cfg would get a copy of that structure. However,
+// this would race with code that assigns a new congiguration to the stream
+// (think stream update). The update will be done using setCfg(), and therefore,
+// to avoid races, callers that need the stream's congiguration should call
+// this function, or protect using mset.cfgMu read lock.
+func (mset *stream) getCfg() StreamConfig {
```
kozlovic, line 2051: If it is agreed that call to `getCfg()` in `addConsumerWithAssignment()` is not needed, then this method may not be needed at all (for now at least).

kozlovic (commented):
@derekcollison I think I don't even need to call to mset.getCfg() since we are protected by the stream's lock there. If you agree, I will update the PR tomorrow.

server/consumer.go
kozlovic: I have reverted that part in the new commit.

server/stream.go
kozlovic: I have removed it, and setCfg() too (I did inline it since there was only one place we were using it).

derekcollison: I don't think it is needed. With it not there can you see the race?

derekcollison (approved):
LGTM

derekcollison:
I will merge after all green.

kozlovic:
@derekcollison I got a data race on a test for completely different thing (non JS). Can I add it here or do you prefer new PR for that?

kozlovic:
Sorry, I did not wait for your answer. It was a quick fix, so added to this PR.

kozlovic:
@derekcollison Ok, got all green, ready to merge!

PR #5826, opened 2024-08-25
[FIXED] New staticcheck fixes

derekcollison:
@neilalexander I do believe this uncovered two off by one conditions in node48 under stree. Could you take a look, kept that change separate.


neilalexander (approved):
Confirmed the fix with an update to the node48 unit test.

derekcollison (commented):
LGTM - Thanks.

derekcollison:
@neilalexander would need official LGTM from you or @wallyqs

neilalexander (approved):
LGTM

PR #5830, opened 2024-08-26
Update Go version for v2.11

wallyqs:
Bumps Go version used in CI and nightlies to v1.23


wallyqs:
need to fix go vet, will reopen again when ready

Issue #5595, opened 2024-06-26
JetStream: high disk IO, CPU usage, timeouts (Internal subscription took too long)


tierpod:
### Observed behavior

We experience strange nats jetstream behavior: it periodically produces huge amount of disk IO (reads and writes) even without any clients connected (0 producers, 0 consumers connected). As a result - high CPU usage during reads and writes and timeouts in log:

```
Internal subscription on "$JS.API.CONSUMER.CREATE.XXX" took too long: 26.464577809
```

Stream configuration:

```
     Acknowledgements: true
            Retention: File - WorkQueue
             Replicas: 1
       Discard Policy: Old
     Duplicate Window: 2m0s
    Allows Msg Delete: true
         Allows Purge: true
       Allows Rollups: false
     Maximum Messages: unlimited
        Maximum Bytes: 30 GiB
          Maximum Age: 30d0h0m0s
 Maximum Message Size: unlimited
    Maximum Consumers: unlimited

State:

             Messages: 37,886,565
                Bytes: 19 GiB
             FirstSeq: 19,941,809 @ 2024-06-25T07:57:13 UTC
              LastSeq: 65,677,737 @ 2024-06-25T10:28:36 UTC
     Deleted Messages: 7849364
     Active Consumers: 10
```

One service produced all these messages, another service should consume them in batches (slower than producer).

I'm not sure how to reproduce it, but I have a backup of this "broken" stream locally and did some experiments with previous versions.

![1](https://github.com/nats-io/nats-server/assets/7151579/48ebfb30-4f3a-4419-bc76-ee6a9781e544)

Nats 2.10.14 (1), 2.10.12 (2): after restoring this stream from backup (first spike), I see only 1 such IO/CPU spike after 3-5 minutes and later nats works fine.

Nats 2.10.16 (3) : after restoring this stream from backup (first spike), I see repeating peaks every N minutes. N changes if I change `jetstream.sync_interval` configuration:

![2](https://github.com/nats-io/nats-server/assets/7151579/d5475857-32be-4af1-a7e2-a2f0d7c7dea8)

I created pprof dump and see these functions in top: syncBlocks, compact, loadMsgsWithLock:

![3](https://github.com/nats-io/nats-server/assets/7151579/a2ffba7f-d014-4402-bd84-e4403a1ec6e5)

After consuming most messages these spikes becomes less visible, so it depends on stream size (data size or number of messages).

### Expected behavior

[...]

PR #5831, opened 2024-08-26
[IMPROVED] When a block can not be compressed, mark it as such to avoid syncBlocks trying continously

derekcollison:
We had reports of disk spikes and memory pops queued off of syncBlocks, and looks like it was coming from blocks that were no longer compressible, but would load each time and go through the process.

Resolves: #5595 


wallyqs (approved):
LGTM

neilalexander (approved):
LGTM

PR #5832, opened 2024-08-26
NRG: Wait for goroutines to shutdown when recreating group

neilalexander:
This should fix some logical races where multiple sets of goroutines can fight over the same store directory, i.e. when shutting down and recreating a group rapidly.

### DIFF
--- a/server/jetstream_cluster.go
+++ b/server/jetstream_cluster.go
@@ -2099,7 +2099,15 @@ func (js *jetStream) createRaftGroup(accName string, rg *raftGroup, storage Stor
 	}
 
 	// Check if we already have this assigned.
+retry:
 	if node := s.lookupRaftNode(rg.Name); node != nil {
+		if node.State() == Closed {
+			// We're waiting for this node to finish shutting down before we replace it.
+			js.mu.Unlock()
+			node.WaitForStop()
+			js.mu.Lock()
+			goto retry
+		}
 		s.Debugf("JetStream cluster already has raft group %q assigned", rg.Name)
 		rg.node = node
 		js.mu.Unlock()
--- a/server/jetstream_cluster.go
+++ b/server/jetstream_cluster.go
@@ -2109,6 +2109,22 @@ retry:
 			goto retry
 		}
 		s.Debugf("JetStream cluster already has raft group %q assigned", rg.Name)
+		// Check and see if the group has the same peers. If not then we
+		// will update the known peers, which will send a peerstate if leader.
+		groupPeerIDs := append([]string{}, rg.Peers...)
+		var samePeers bool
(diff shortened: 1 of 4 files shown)


neilalexander:
Have rebased, this one is ready for review.

server/jetstream_cluster.go
```diff
@@ -2050,7 +2050,15 @@ func (js *jetStream) createRaftGroup(accName string, rg *raftGroup, storage Stor
 	}
 
 	// Check if we already have this assigned.
+retry:
 	if node := s.lookupRaftNode(rg.Name); node != nil {
+		if node.State() == Closed {
```
derekcollison, line 2055: Feels like there may be a better approach here.

Also I think we need to check that the peer lists are the same if we are going to re-use.

server/raft.go
```diff
@@ -1758,6 +1766,13 @@ func (n *raft) shutdown(shouldDelete bool) {
 		n.c = nil
 	}
 
+	// Wait for goroutines to shut down before trying to clean up
+	// the log below, otherwise we might end up deleting the store
+	// from underneath running goroutines.
+	n.Unlock()
```
derekcollison, line 1772: We have these patterns for sure elsewhere but I try hard to avoid them if at all possible.

server/jetstream_cluster.go
derekcollison: Do we want to return a closed node here?

server/raft.go
```diff
@@ -1728,12 +1731,17 @@ func (n *raft) Stop() {
 	n.shutdown(false)
 }
 
+func (n *raft) WaitForStop() {
+	n.wg.Wait()
```
derekcollison, line 1735: Should we check that the state is set to stopped to avoid calling in here and never returning?

server/raft.go
```diff
@@ -1744,12 +1752,12 @@ func (n *raft) shutdown(shouldDelete bool) {
 	if n.state.Swap(int32(Closed)) == int32(Closed) {
 		// If we get called again with shouldDelete, in case we were called first with Stop() cleanup
 		if shouldDelete {
+			n.wg.Wait()
```
derekcollison, line 1755: Since we hold the lock here this may never return.

neilalexander:
I have shuffled things around a bit:
* The `run` goroutine is now responsible for cleaning up once the loop ends, which is a server-monitored goroutine;
* The `shutdown` function now just signals the shutdown and doesn't block the caller;
* The `WaitForStop` function only blocks now if the Raft group is really in a closed state;
* We otherwise only block if we're looking up a Raft node and find that one already exists.

derekcollison (approved):
LGTM

PR #5836, opened 2024-08-28
Windows: Search for DLLs in system paths only

neilalexander:
Otherwise DLLs can potentially be loaded in from the current working directory.

Tested on a Windows 11 ARM64 virtual machine.


derekcollison (approved):
LGTM

PR #5842, opened 2024-08-28
Update default file and directory permissions

neilalexander:
Filestore-owned items don't need to be group-accessible and log files don't need to be group-writable.


derekcollison (approved):
LGTM

I wonder if somehow this might break folks though.

PR #5862, opened 2024-09-06
Test to verify removing and adding back in a source stream works as expected.

derekcollison:


wallyqs (approved):
LGTM

PR #5878, opened 2024-09-11
Fixed nightly code coverage [ci skip]

kozlovic:
The code coverage GA was referencing 1.21.x which seem to prevent the install of gocovmerge.
Also added an early check to see if we can run gocovmerge before starting the test suite.

### DIFF
--- a/.github/workflows/cov.yaml
+++ b/.github/workflows/cov.yaml
@@ -22,7 +22,8 @@ jobs:
       - name: Set up Go
         uses: actions/setup-go@v5
         with:
-          go-version: "1.21.x"
+          go-version-file: src/github.com/nats-io/nats-server/go.mod
+          cache-dependency-path: src/github.com/nats-io/nats-server/go.sum
 
       - name: Run code coverage
         shell: bash --noprofile --norc -x -eo pipefail {0}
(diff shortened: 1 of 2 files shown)


wallyqs (approved):
LGTM

.github/workflows/cov.yaml
```diff
@@ -22,7 +22,8 @@ jobs:
       - name: Set up Go
         uses: actions/setup-go@v5
         with:
-          go-version: "1.21.x"
+          go-version-file: src/github.com/nats-io/nats-server/go.mod
```
kozlovic, line 25: @wallyqs Although this is right now using go 1.22 (in go.mod), but if the go.mod does get updated for a while, this may not be the good approach (took that from the mqtt yml file). What do you think, should I be explicitly using "1.23.x" here? (and update as we update our go version requirements?)

neilalexander: We're generally keeping `go.mod` up-to-date with support for Go N and N-1 so I think this is probably fine.

derekcollison:
This ready for merge?

kozlovic:
@derekcollison Yes, since Neil confirmed that the go.mod is usually up-to-date with the go versions.

PR #5885, opened 2024-09-13
Tests: Wait for shutdown on `stopAll`, parallelise LDM shutdowns

neilalexander:


derekcollison (approved):
LGTM

PR #5893, opened 2024-09-16
[FIXED] Improvements to dealing with old or non-existant index.db

derekcollison:
We had a condition where an old index.db was not able to properly restore a stream due to max msgs per subject being set and certain blocks being compacted away and removing subject info for those sequences. In addition we fixed recovery after Truncate and PurgeEx by subject when the index.db was corrupt or not available.

This change also moves generating the index.db file to after writing the blocks during a snapshot and we do a force call to make sure it is written even when complex.


neilalexander (approved):
LGTM

PR #5897, opened 2024-09-17
Empty `cluster_traffic` value should default to `"system"` in JWT

neilalexander:
Avoids unnecessary `has invalid value "" for cluster traffic account` error.


ripienaar (approved):
LGTM

PR #5900, opened 2024-09-17
Global JetStream API request queue limiting

neilalexander:
This should provide some protection against the server being completely overwhelmed by a build-up of JS API requests.

### DIFF
--- a/server/jetstream.go
+++ b/server/jetstream.go
@@ -107,6 +107,7 @@ type jetStream struct {
 	storeReserved int64
 	memUsed       int64
 	storeUsed     int64
+	queueLimit    int64
 	clustered     int32
 	mu            sync.RWMutex
 	srv           *Server
@@ -415,6 +416,9 @@ func (s *Server) enableJetStream(cfg JetStreamConfig) error {
 		s.gcbOutMax = defaultMaxTotalCatchupOutBytes
 	}
 	s.gcbMu.Unlock()
+
+	// TODO: Not currently reloadable.
+	atomic.StoreInt64(&js.queueLimit, s.getOpts().JetStreamRequestQueueLimit)
 
 	s.js.Store(js)
 
--- a/server/jetstream_api.go
+++ b/server/jetstream_api.go
@@ -307,6 +307,9 @@ const (
 	// JSAdvisoryServerRemoved notification that a server has been removed from the system.
 	JSAdvisoryServerRemoved = "$JS.EVENT.ADVISORY.SERVER.REMOVED"
 
+	// JSAdvisoryAPILimitReached notification that a server has reached the JS API hard limit.
+	JSAdvisoryAPILimitReached = "$JS.EVENT.ADVISORY.API.LIMIT_REACHED"
+
 	// JSAuditAdvisory is a notification about JetStream API access.
 	// FIXME - Add in details about who..
 	JSAuditAdvisory = "$JS.EVENT.ADVISORY.API"
@@ -353,6 +356,10 @@ const JSMaxMetadataLen = 128 * 1024
 // JSMaxNameLen is the maximum name lengths for streams, consumers and templates.
 // Picked 255 as it seems to be a widely used file name limit
 const JSMaxNameLen = 255
+
+// JSDefaultRequestQueueLimit is the default number of entries that we will
+// put on the global request queue before we react.
+const JSDefaultRequestQueueLimit = 10_000
 
 // Responses for API calls.
 
@@ -862,10 +869,22 @@ func (js *jetStream) apiDispatch(sub *subscription, c *client, acc *Account, sub
 	// Copy the state. Note the JSAPI only uses the hdr index to piece apart the
 	// header from the msg body. No other references are needed.
 	// Check pending and warn if getting backed up.
-	const warnThresh = 128
 	pending, _ := s.jsAPIRoutedReqs.push(&jsAPIRoutedReq{jsub, sub, acc, subject, reply, copyBytes(rmsg), c.pa})
-	if pending >= warnThresh {
-		s.rateLimitFormatWarnf("JetStream request queue has high pending count: %d", pending)
+	limit := atomic.LoadInt64(&js.queueLimit)
+	if pending >= int(limit) {
+		s.rateLimitFormatWarnf("JetStream API queue limit reached, dropping %d requests", pending)
+		s.jsAPIRoutedReqs.drain()
+
+		s.publishAdvisory(nil, JSAdvisoryAPILimitReached, JSAPILimitReachedAdvisory{
+			TypedEvent: TypedEvent{
+				Type: JSAPILimitReachedAdvisoryType,
+				ID:   nuid.Next(),
+				Time: time.Now().UTC(),
+			},
+			Server:  s.Name(),
+			Domain:  js.config.Domain,
+			Dropped: int64(pending),
+		})
 	}
 }
 
--- a/server/jetstream_events.go
+++ b/server/jetstream_events.go
@@ -295,3 +295,14 @@ type JSServerRemovedAdvisory struct {
 	Cluster  string `json:"cluster"`
 	Domain   string `json:"domain,omitempty"`
 }
+
+// JSAPILimitReachedAdvisoryType is sent when the JS API request queue limit is reached.
+const JSAPILimitReachedAdvisoryType = "io.nats.jetstream.advisory.v1.api_limit_reached"
+
+// JSAPILimitReachedAdvisory is a advisory published when JetStream hits the queue length limit.
+type JSAPILimitReachedAdvisory struct {
+	TypedEvent
+	Server  string `json:"server"`           // Server that created the event, name or ID
+	Domain  string `json:"domain,omitempty"` // Domain the server belongs to
+	Dropped int64  `json:"dropped"`          // How many messages did we drop from the queue
+}
--- a/server/opts.go
+++ b/server/opts.go
@@ -277,89 +277,90 @@ type AuthCallout struct {
 // NOTE: This structure is no longer used for monitoring endpoints
 // and json tags are deprecated and may be removed in the future.
 type Options struct {
-	ConfigFile            string        `json:"-"`
-	ServerName            string        `json:"server_name"`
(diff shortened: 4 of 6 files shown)


server/jetstream_api.go
```diff
@@ -307,6 +307,9 @@ const (
 	// JSAdvisoryServerRemoved notification that a server has been removed from the system.
 	JSAdvisoryServerRemoved = "$JS.EVENT.ADVISORY.SERVER.REMOVED"
 
+	// JSAdvisoryAPIRateLimit notification that a server has reached the JS API rate limit.
+	JSAdvisoryAPIRateLimit = "$JS.EVENT.ADVISORY.API.RATE_LIMIT"
```
derekcollison, line 311: I wonder if this should fall under API AUDIT that already exists?

server/jetstream_api.go
```diff
@@ -862,10 +869,25 @@ func (js *jetStream) apiDispatch(sub *subscription, c *client, acc *Account, sub
 	// Copy the state. Note the JSAPI only uses the hdr index to piece apart the
 	// header from the msg body. No other references are needed.
 	// Check pending and warn if getting backed up.
-	const warnThresh = 128
+	s.optsMu.RLock()
+	limit := s.opts.JetStreamRequestQueueLimit
```
derekcollison, line 873: For now make a global var that is set to constant but we can override on config parse?

server/jetstream_api.go
```diff
@@ -862,10 +869,25 @@ func (js *jetStream) apiDispatch(sub *subscription, c *client, acc *Account, sub
...
+		s.publishAdvisory(nil, JSAdvisoryAPIRateLimit, JSAPIRateLimitAdvisory{
+			TypedEvent: TypedEvent{
+				Type: JSAPIRateLimitAdvisoryType,
+				ID:   nuid.Next(),
+				Time: time.Now().UTC(),
+			},
+			Server:      s.Name(),
+			Domain:      js.config.Domain,
+			Account:     _EMPTY_, // as it's systemwide
+			QueueLength: int64(pending),
```
derekcollison, line 889: Just Pending?

server/jetstream_events.go
```diff
@@ -295,3 +295,15 @@ type JSServerRemovedAdvisory struct {
...
+
+// JSAPIRateLimitAdvisoryType is sent when the JS API request queue limit is reached.
+const JSAPIRateLimitAdvisoryType = "io.nats.jetstream.advisory.v1.api_rate_limit"
+
+// JSAPIRateLimitAdvisory is a advisory published when JetStream limits API access due to rate limits
+type JSAPIRateLimitAdvisory struct {
+	TypedEvent
+	Server      string `json:"server"`            // Server that created the event, name or ID
+	Domain      string `json:"domain,omitempty"`  // Domain the server belongs to
+	Account     string `json:"account,omitempty"` // Optionally account that hit the limit, or empty for system
```
derekcollison, line 307: Will be a hard limit in 2.11 even with flow control, so can remove.

server/opts.go
```diff
@@ -2417,6 +2418,12 @@ func parseJetStream(v any, opts *Options, errors *[]error, warnings *[]error) er
 					return &configErr{tk, fmt.Sprintf("Expected a parseable size for %q, got %v", mk, mv)}
 				}
 				opts.StreamMaxBufferedMsgs = int(mlen)
+			case "request_queue_limit":
+				lim, ok := mv.(int64)
+				if !ok {
+					return &configErr{tk, fmt.Sprintf("Expected a parseable size for %q, got %v", mk, mv)}
+				}
+				opts.JetStreamRequestQueueLimit = lim
```
derekcollison, line 2426: Maybe set global var here as mentioned above.

server/jetstream_api.go
ripienaar: not really an audit trail but otoh might be worth knowing when things are dropped in auditing, but introducing a new type of message on that subject my trip up some things people have written

server/jetstream_events.go
ripienaar: We discussed that we want to introduce a limit for like a free account can have 10 req/sec or something, so there will be a concept of an account limit and that's why I added this here now so that if we release CLI before 2.11 when that comes it'll already know how to handle these messages and report the account

derekcollison: ok but them maybe this is the wrong advisory here. This one is hard limit, and we drop everything. RateLimit we will most likely drop last request in hand and probably respond to the app tier (with possible delay etc).

So I think this one should be different for its purpose.

server/jetstream_api.go
derekcollison: I was suggesting use same format, nil request and response that essentially says overloaded, etc..

server/jetstream_events.go
ripienaar: yeah that's fair.

So we can drop `account` here and later introduce a new advisory on new subjects etc

derekcollison: I would name advisory different as well. Not really a rate limit here, LimitReached?

server/jetstream_api.go
ripienaar: Ah no I would rather have a specific advisory on specific subject. The api audit tends to spam a lot especially when busy and then you miss these. 

On a separate subject and format you will be able to just listen on those for example.

server/jetstream_events.go
ripienaar: Yeah

server/jetstream_api.go
```diff
@@ -862,10 +869,22 @@ func (js *jetStream) apiDispatch(sub *subscription, c *client, acc *Account, sub
 	// Copy the state. Note the JSAPI only uses the hdr index to piece apart the
 	// header from the msg body. No other references are needed.
 	// Check pending and warn if getting backed up.
-	const warnThresh = 128
 	pending, _ := s.jsAPIRoutedReqs.push(&jsAPIRoutedReq{jsub, sub, acc, subject, reply, copyBytes(rmsg), c.pa})
-	if pending >= warnThresh {
-		s.rateLimitFormatWarnf("JetStream request queue has high pending count: %d", pending)
+	limit := atomic.LoadInt64(&js.queueLimit)
+	if pending >= int(limit) {
+		s.rateLimitFormatWarnf("Draining JetStream request queue due to high pending count: %d", pending)
```
derekcollison, line 875: "JetStream API queue limit reached, dropping %d requests"

server/jetstream_api.go
```diff
@@ -862,10 +869,22 @@ func (js *jetStream) apiDispatch(sub *subscription, c *client, acc *Account, sub
...
+
+		s.publishAdvisory(nil, JSAdvisoryAPILimitReached, JSAPILimitReachedAdvisory{
+			TypedEvent: TypedEvent{
+				Type: JSAPILimitReachedAdvisoryType,
+				ID:   nuid.Next(),
+				Time: time.Now().UTC(),
+			},
+			Server:      s.Name(),
+			Domain:      js.config.Domain,
+			QueueLength: limit,
```
derekcollison, line 886: Dropped

server/jetstream_api.go
```diff
@@ -862,10 +869,22 @@ func (js *jetStream) apiDispatch(sub *subscription, c *client, acc *Account, sub
 	// Copy the state. Note the JSAPI only uses the hdr index to piece apart the
 	// header from the msg body. No other references are needed.
 	// Check pending and warn if getting backed up.
-	const warnThresh = 128
 	pending, _ := s.jsAPIRoutedReqs.push(&jsAPIRoutedReq{jsub, sub, acc, subject, reply, copyBytes(rmsg), c.pa})
-	if pending >= warnThresh {
-		s.rateLimitFormatWarnf("JetStream request queue has high pending count: %d", pending)
+	limit := atomic.LoadInt64(&js.queueLimit)
```
derekcollison, line 873: Not sure its worth protecting, especially if not reloadable, but should be cheap enough..

derekcollison (changes requested):

server/jetstream_api.go
neilalexander: Yeah, mostly just did this in case we wanted to make it reloadable in the future.

server/jetstream.go
```diff
@@ -416,6 +417,9 @@ func (s *Server) enableJetStream(cfg JetStreamConfig) error {
 	}
 	s.gcbMu.Unlock()
 
+	// TODO: Not currently reloadable.
+	atomic.StoreInt64(&js.queueLimit, s.getOpts().JetStreamRequestQueueLimit)
```
derekcollison, line 421: getOpts grabs a lock. So this is now doing lock and atomic.. I do not think either are needed.

Let's consider a global var that is set to the constant 10k to start and opts.go parsing will update both opts (meaning it still can be found in opts) but also updates the global, without any locks..

server/jetstream_api.go
derekcollison: See above. We are now doing a lock and an atomic, neither are needed IMO.

derekcollison (changes requested):

server/jetstream.go
neilalexander: Main problem with a global var is if we have multiple instances in one process, and we'll end up with a race condition if the options code updates the global while one of the JS goroutines is reading it.

In this case, the `getOpts()` call only happens once which is when JetStream is enabled. The atomic isn't really needed right now since even if the opts are reloaded, `js.queueLimit` isn't.

derekcollison: Not if we set global var to constant like 

```go
var apiPendingLimit = 10_000
```

Then only if we override it we might see it..

derekcollison: ok you are right not in hot path. My bad..

server/jetstream_api.go
derekcollison, line 873: One trick here we could do is that we know hard limit is greater them warn limit, so only if we hit warn limit do we look up limit with atomic and check hard limit.

derekcollison: We can do as a followup once we figure out test approach.

derekcollison (approved):
LGTM

PR #5903, opened 2024-09-18
[ADDED] Delay option to js_cluster_migrate

piotrpio:
This adds a `delay` option to `js_cluster_migrate` as an option in nested object. The assets assets will only be migrated after the specified delay (rather than immediately, which is the default).

### DIFF
--- a/server/leafnode.go
+++ b/server/leafnode.go
@@ -494,7 +494,7 @@ func (s *Server) connectToRemoteLeafNode(remote *leafNodeCfg, firstConnect bool)
 
 	opts := s.getOpts()
 	reconnectDelay := opts.LeafNode.ReconnectInterval
-	s.mu.Lock()
+	s.mu.RLock()
 	dialTimeout := s.leafNodeOpts.dialTimeout
 	resolver := s.leafNodeOpts.resolver
 	var isSysAcc bool
@@ -502,7 +502,7 @@ func (s *Server) connectToRemoteLeafNode(remote *leafNodeCfg, firstConnect bool)
 		isSysAcc = remote.LocalAccount == s.sys.account.Name
 	}
 	jetstreamMigrateDelay := remote.JetStreamClusterMigrateDelay
-	s.mu.Unlock()
+	s.mu.RUnlock()
 
 	// If we are sharing a system account and we are not standalone delay to gather some info prior.
 	if firstConnect && isSysAcc && !s.standAloneMode() {
--- a/server/leafnode.go
+++ b/server/leafnode.go
@@ -591,10 +591,7 @@ func (s *Server) connectToRemoteLeafNode(remote *leafNodeCfg, firstConnect bool)
 
 func (cfg *leafNodeCfg) cancelMigrateTimer() {
 	cfg.Lock()
-	if cfg.jsMigrateTimer != nil {
-		cfg.jsMigrateTimer.Stop()
-		cfg.jsMigrateTimer = nil
-	}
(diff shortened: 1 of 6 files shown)


derekcollison:
This ready for review @piotrpio ?

server/opts.go
```diff
@@ -2746,6 +2750,16 @@ func parseRemoteLeafNodes(v any, errors *[]error, warnings *[]error) ([]*RemoteL
 				remote.Websocket.NoMasking = v.(bool)
 			case "jetstream_cluster_migrate", "js_cluster_migrate":
 				remote.JetStreamClusterMigrate = true
+				migrateConfig, ok := v.(map[string]any)
```
neilalexander, line 2753: Although it was this way before, currently setting `jetstream_cluster_migrate: false` in the config looks like it will do the wrong thing and set to `true` anyway. Do we want to fix this?

Might also be better just to check for bool vs string, parse the string as duration, rather than having to open up a new map. That way you could do any of:
```
jetstream_cluster_migrate: true
jetstream_cluster_migrate: 30s
jetstream_cluster_migrate: false
```
WDYT?

neilalexander: Maybe @derekcollison has an opinion on this too.

derekcollison: I would prefer `leader_migrate_delay`, although that is pretty long winded for me but is more descriptive IMO. It is a value that defaults to 0, meaning instant, and you can set to a value that means after Nsecs of being disconnected from leafnode we will migrate leaders.

piotrpio:
@derekcollison Yes, it's ready for review, I was waiting for the tests to pass and did not change the state.

server/opts.go
piotrpio: @neilalexander I agree but I wonder how much of a breaking change that would be `js_cluster_enabled: false` seems like a bug which can be fixed, but right now e.g. `js_cluster_migrate: enabled` also works (or any other value really).

@derekcollison with `leader_migrate_delay` you would see this as a separate option in `remote` or an option in object on `js_cluster_migrate` (as I did it here with `delay`)?

server/leafnode.go
```diff
@@ -573,6 +589,15 @@ func (s *Server) connectToRemoteLeafNode(remote *leafNodeCfg, firstConnect bool)
 	}
 }
 
+func (cfg *leafNodeCfg) cancelMigrateTimer() {
+	cfg.Lock()
+	if cfg.jsMigrateTimer != nil {
```
neilalexander, line 594: We have a helper function for this:
```
stopAndClearTimer(&cfg.jsMigrateTimer)
```

derekcollison:
Let me know when current feedback has been addressed and I will take a look.

neilalexander (approved):
LGTM

server/leafnode.go
```diff
@@ -500,6 +501,7 @@ func (s *Server) connectToRemoteLeafNode(remote *leafNodeCfg, firstConnect bool)
 	if s.eventsEnabled() {
 		isSysAcc = remote.LocalAccount == s.sys.account.Name
 	}
+	jetstreamMigrateDelay := remote.JetStreamClusterMigrateDelay
```
derekcollison, line 504: This server lock could be read lock.

server/opts.go
derekcollison: option on js_cluster_migrate I think.

derekcollison: Let's make sure `js_cluster_migrate: false` works correctly.

derekcollison (approved):
LGTM

PR #5904, opened 2024-09-18
Add support for setting TLS min version number in tls block

wallyqs:
For example, to set minimum version to TLS v1.3:

```
  tls { min_version = "1.3" }
```


derekcollison (approved):
LGTM

PR #5906, opened 2024-09-19
OCSP: support using https port when ocsp peer enabled

jarretlavallee:
This allows client certificates on the https monitoring port to be validated when OCSP peering is enabled.

Signed-off-by: Jarret Lavallee <jarret@synadia.com>


derekcollison (approved):
LGTM

PR #5912, opened 2024-09-22
[FIXED] Deadlock when removing a peer that happened to be the leader.

derekcollison:


neilalexander (approved):
LGTM

PR #5917, opened 2024-09-23
Log only durable orphaned consumers

ripienaar:
But log the rest at debug for dev purposes


neilalexander (approved):
LGTM

PR #5918, opened 2024-09-23
Add `NumInterest` to optimise searching for number of sublist entries

neilalexander:
There are some cases where we care about the number of subscriptions on the sublist but don't need the `SublistResult`s. We can save quite a few allocations on these paths by reusing the match-any logic from `HasInterest` to count the number of subs.


derekcollison (approved):
LGTM

PR #5925, opened 2024-09-25
Publish STATSZ messages every 10 seconds rather than 30

ripienaar:


neilalexander (approved):
LGTM if tests happy.

derekcollison (approved):
LGTM

PR #5927, opened 2024-09-25
[FIXED] When scaling down a stream make sure consumers are adjusted properly.

derekcollison:
When scaling down a stream make sure replica count is correct if adjusted and also make sure we do not have orphan consumers.

When we scale down a replicated stream, say R5, if it has consumers that are a lower replica count, say R1, they could be placed on the peers that may go away. We need to make sure we properly assign peers and transfer state as needed.

Note the consumer state transfer expects the state to be stable, so should be paused.

### DIFF
--- a/server/jetstream_cluster.go
+++ b/server/jetstream_cluster.go
@@ -6357,8 +6357,9 @@ func (s *Server) jsClusteredStreamUpdateRequest(ci *ClientInfo, acc *Account, su
 	}
 
 	if isReplicaChange {
+		isScaleUp := newCfg.Replicas > len(rg.Peers)
 		// We are adding new peers here.
-		if newCfg.Replicas > len(rg.Peers) {
+		if isScaleUp {
 			// Check that we have the allocation available.
 			if err := js.jsClusteredStreamLimitsCheck(acc, newCfg); err != nil {
 				resp.Error = err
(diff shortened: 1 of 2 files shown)


server/jetstream_cluster.go
```diff
@@ -6434,22 +6435,82 @@ func (s *Server) jsClusteredStreamUpdateRequest(ci *ClientInfo, acc *Account, su
...
+						// We have to transfer state to new peers.
+						// we will grab our state and attach to the new assignment.
+						// TODO(dlc) - In practice we would want to make sure the consumer is paused.
+						// Need to release js lock.
+						js.mu.Unlock()
+						if ci, err := sysRequest[ConsumerInfo](s, clusterConsumerInfoT, acc, osa.Config.Name, ca.Name); err != nil {
+							s.Warnf("Did not receive consumer info results for '%s > %s > %s' due to: %s", acc, osa.Config.Name, ca.Name, err)
+						} else if ci != nil {
+							cca.State = &ConsumerState{
+								Delivered: SequencePair{
```
neilalexander, line 6495: Do we need to include the `Redelivered` state here too?

neilalexander (approved):
Just one question but otherwise LGTM.

server/jetstream_cluster.go
derekcollison: We do not have access to that state here.

derekcollison: So not perfect, and we should revisit for 2.11, but better then what we had for sure.

PR #5940, opened 2024-09-27
(2.11) Internal: Small udpates to ipQueue

kozlovic:
Small updates to ipQueue:
- Document that caller must not reuse the slice after call to recycle().
- Reset size to 0 on pop() or last popOne() without need to call q.caclc().
- Added missing recycle() calls in some places.
- Added benchmarks to check perf impact on future changes.

### DIFF
--- a/server/ipqueue.go
+++ b/server/ipqueue.go
@@ -85,8 +85,15 @@ var errIPQSizeLimitReached = errors.New("IPQ size limit reached")
 
 func newIPQueue[T any](s *Server, name string, opts ...ipQueueOpt[T]) *ipQueue[T] {
 	q := &ipQueue[T]{
-		ch:   make(chan struct{}, 1),
-		pool: &sync.Pool{},
+		ch: make(chan struct{}, 1),
+		pool: &sync.Pool{
+			New: func() any {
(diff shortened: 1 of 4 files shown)


kozlovic:
@neilalexander As discussed in the other PR, I am splitting changes to ipQueue starting with this one, that contains small updates. I ended-up keeping `[]T` for `pop()` (which also minimize changes) and simply document that one must not reuse the slice after recycle(), which, after thinking about it makes sense. Anything that is put back into a pool must not be reused by the caller, since the pool owns it now, and is then used by other code after a Get().

neilalexander (approved):
LGTM

kozlovic:
@neilalexander Thanks. In the next PR, I would like to use `*[]T` all the way and have pop() be given an iteration function instead, so pop() will be responsible to update the len/size per element for queues that have limit and take care of the recycling itself. I believe that the callback/iter approach is crucial since with current approach but with the new way of taking into account pop() elements for the len/size limits calculations, forgetting a recycle() (which by the way was doing more than just putting slice back into pool) would have a critical impact possibly leading to a queue not accepting new elements (because len/size would not be properly reduced). Are you ok with the approach I am suggesting?

PR #5947, opened 2024-10-01
Reject publishes on `$NRG.>` subjects from non-system clients

neilalexander:

### DIFF
--- a/server/client.go
+++ b/server/client.go
@@ -3905,6 +3905,10 @@ func (c *client) selectMappedSubject() bool {
 	return changed
 }
 
+// clientNRGPrefix is used in processInboundClientMsg to detect if publishes
+// are being made from normal clients to NRG subjects.
+var clientNRGPrefix = []byte("$NRG.")
+
 // processInboundClientMsg is called to process an inbound msg from a client.
 // Return if the message was delivered, and if the message was not delivered
 // due to a permission issue.
@@ -3936,6 +3940,13 @@ func (c *client) processInboundClientMsg(msg []byte) (bool, bool) {
 		return false, true
 	}
 	c.mu.Unlock()
+
+	// Check if the client is trying to publish to reserved NRG subjects.
+	// Doesn't apply to NRGs themselves as they use SYSTEM-kind clients instead.
+	if c.kind == CLIENT && bytes.HasPrefix(c.pa.subject, clientNRGPrefix) && acc != c.srv.SystemAccount() {
+		c.pubPermissionViolation(c.pa.subject)
+		return false, true
+	}
 
 	// Now check for reserved replies. These are used for service imports.
 	if c.kind == CLIENT && len(c.pa.reply) > 0 && isReservedReply(c.pa.reply) {
--- a/server/client_test.go
+++ b/server/client_test.go
@@ -3123,3 +3123,50 @@ func TestClientFlushOutboundNoSlowConsumer(t *testing.T) {
 	}
 	require_Equal(t, msgs, 8)
 }
+
+func TestClientRejectsNRGSubjects(t *testing.T) {
+	conf := createConfFile(t, []byte(`
+		listen: 127.0.0.1:-1
+		accounts: {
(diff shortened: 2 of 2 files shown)


server/client.go
```diff
@@ -3935,6 +3935,13 @@ func (c *client) processInboundClientMsg(msg []byte) (bool, bool) {
 		c.pubPermissionViolation(c.pa.subject)
 		return false, true
 	}
+	// Check if the client is trying to publish to reserved NRG subjects.
+	// Doesn't apply to NRGs themselves as they use SYSTEM-kind clients instead.
+	if c.kind == CLIENT && strings.HasPrefix(bytesToString(c.pa.subject), "$NRG.") && c.acc != c.srv.SystemAccount() {
```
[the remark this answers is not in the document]
neilalexander: Done.

server/client.go
```diff
@@ -3935,6 +3939,13 @@ func (c *client) processInboundClientMsg(msg []byte) (bool, bool) {
 		c.pubPermissionViolation(c.pa.subject)
 		return false, true
 	}
+	// Check if the client is trying to publish to reserved NRG subjects.
+	// Doesn't apply to NRGs themselves as they use SYSTEM-kind clients instead.
+	if c.kind == CLIENT && bytes.HasPrefix(c.pa.subject, clientNRGPrefix) && c.acc != c.srv.SystemAccount() {
```
derekcollison, line 3944: Doe we grab server lock for SystemAccount? If so are we violating locking order?

neilalexander: Good point, it does take the lock, but `acc` is already taken under the client lock earlier too, so moved this down to after the client unlock.

server/client.go
```diff
@@ -3937,6 +3941,13 @@ func (c *client) processInboundClientMsg(msg []byte) (bool, bool) {
 	}
 	c.mu.Unlock()
 
+	// Check if the client is trying to publish to reserved NRG subjects.
+	// Doesn't apply to NRGs themselves as they use SYSTEM-kind clients instead.
+	if c.kind == CLIENT && bytes.HasPrefix(c.pa.subject, clientNRGPrefix) && acc != c.srv.SystemAccount() {
```
derekcollison, line 3946: Still have some things here we are accessing from c.

Maybe add a new lock for looking up the system account?

kozlovic: Client lock was released just above, so there is no lock ordering issue. If the worry is about accessing `c.kind` and `c.srv`, these could be captured above under the lock. `c.pa.subject` should not be an issue since this is "thread local" to the processing of an inbound message.

If the concern is accessing server lock here just to see if the account is system account, maybe adding an atomic boolean in the `Account` struct that is set when setting the system account and could be checked here with atomic opeartion?

neilalexander: `c.kind` doesn't change after the creation of the client connection, `c.pa.subject` is indeed thread-local and `c.srv` also doesn't change. I think this is safe.

Note there are existing accesses to `c.kind` and `c.pa.subject` further up outside of the client lock.

neilalexander:
Rebased on latest `main`.

server/client.go
derekcollison: Still think we may want separate server mutex around looking up system account. Or make it an atomic.Pointer.

derekcollison:
Would like to see atomic.Pointer for system account access from server struct.

server/client.go
derekcollison: Or just compare against s.sys with no locks.

neilalexander:
I don't think an atomic pointer is going to solve the problem here, because not only do we need to worry about `s.sys`, but also the contents of `s.sys` which are seemingly implicitly covered by the server lock.

It's possible to hit data races via the config reload and/or eventing shutdown paths, so I think this is probably the least bad option for now, given that we'd only attempt to take the read lock if it was a client trying to publish to an NRG subject, which would be extremely unusual anyway, and there should be next-to-no read contention on the server lock in normal circumstances even if they did.

derekcollison (approved):
LGTM

PR #5948, opened 2024-10-01
Refactor delayed API responses

neilalexander:
If we are going to dispatch delayed API errors as a part of the rate limiting work, then we need machinery that can send out delayed responses without leaking an endless amount of goroutines that are all just blocked on `time.Sleep`.

This refactors things so that the delayed API responses are queued into an IPQ and will be sent out by a single goroutine at the right time, which will be much friendlier on the scheduler.

Additionally we also add new sort functionality to the IPQs so that on each push, we can re-sort the pending items so that those that are due to be processed first will be popped first.


server/jetstream_api.go
```diff
@@ -1012,30 +1012,48 @@ func (s *Server) sendAPIErrResponse(ci *ClientInfo, acc *Account, subject, reply
...
+	rg       *raftGroup
+	after    time.Time
+}
 
-	s.startGoRoutine(func() {
-		defer s.grWG.Done()
+func (s *Server) delayedAPIResponder() {
+	defer s.grWG.Done()
+	for {
+		resp, ok := s.delayedAPIResponses.popOne()
```
kozlovic, line 1029: This is wrong. You need to wait on the queue's `ch` to wait for a notification that there is something to pop.

server/jetstream_api.go
```diff
@@ -1012,30 +1012,48 @@ func (s *Server) sendAPIErrResponse(ci *ClientInfo, acc *Account, subject, reply
...
+		if !ok {
+			return // IPQueue was shut down.
```
kozlovic, line 1031: No, this means that the queue is empty, which if you respond to a notification that there is something should not happen, expect in cases with drain(). Since I don't see the select on the queue's `ch`, this function should have returned right away...

server/ipqueue.go
```diff
@@ -80,6 +82,16 @@ func ipqLimitByLen[T any](max int) ipQueueOpt[T] {
 	}
 }
 
+// This option allows providing a function which is called to re-sort the
+// queued values each time something is pushed to the IPQ.
+// WARNING: This has the potential to be slow on very heavily utilized IPQs.
+// WARNING: It is only safe to use this with popOne() instead of pop().
```
kozlovic, line 88: Not sure about this feature. In the use-case you have, there is still no guarantee that a `1sec` delay is not stuck behind a `3sec`, and by that I mean the currently popOne() may wait for 3sec. So if respecting the delay is really important, then the IPQueue approach, even with adding this sort feature, may not be great.

Also, you would move the "least delayed" at the front, but then that makes the others in the queue be delayed by that much. Say you have 3sec, 4sec in the queue, then add 1sec, 2sec, you now have 1sec, 2sec, 3sec, 4sec. You understand that each popOne() will sleep by that much and so the first that was going to be pop'ed and wait for 3sec is now itself delayed by 3sec before being pop'ed, etc...

kozlovic (changes requested):

PR #5950, opened 2024-10-01
canary: changed uses of jwt.TagList.Contains() to jwt.TagList.ContainsEqualFold(), this relaxes matching to be case-insensitive

aricart:
This changes uses of jwt.Contains() to jwt.ContainsEqualsFold(), allowing case-insensitive matching in the jwt.TagList
Signed-off-by: Your Name <alberto@synadia.com>

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -6,7 +6,7 @@ require (
 	github.com/google/go-tpm v0.9.0
 	github.com/klauspost/compress v1.17.10
 	github.com/minio/highwayhash v1.0.3
-	github.com/nats-io/jwt/v2 v2.6.0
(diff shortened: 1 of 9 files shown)


aricart:
This requires the pointed-at branch of JWT, which we need to merge before updating the go.mod to a released version.

aricart:
Added an utility to jwt.TagList that can return value portion of a tag - where the tag format is `tagname:tagvalue`, but at the very least tag name has a single character, and tag value is not empty. In those cases, tagname is case-insensitive, the tag value utility returns the value portion of the entry (could be multiple) with the `tagname:` stripped.

PR #5951, opened 2024-10-01
[DRAFT] Raft Improvements

ReubenMathew:
DRAFT

### DIFF
--- a/server/raft.go
+++ b/server/raft.go
@@ -4082,15 +4082,9 @@ func (n *raft) processVoteRequest(vr *voteRequest) error {
 	voteOk := n.vote == noVote || n.vote == vr.candidate
 	if voteOk && (vr.lastTerm > n.pterm || vr.lastTerm == n.pterm && vr.lastIndex >= n.pindex) {
 		vresp.granted = true
-		n.term = vr.term
 		n.vote = vr.candidate
 		n.writeTermVote()
 		n.resetElectionTimeout()
-	} else {
-		if vr.term >= n.term && n.vote == noVote {
-			n.term = vr.term
-			n.resetElect(randCampaignTimeout())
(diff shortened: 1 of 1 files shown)


server/raft.go
```diff
@@ -3619,6 +3613,17 @@ func (n *raft) processAppendEntryResponse(ar *appendEntryResponse) {
 		n.stepdownLocked(noLeader)
 		n.Unlock()
 		arPool.Put(ar)
+		return
+	} else if ar.term < n.term {
```
derekcollison, line 3617: Don't need this else, just normal if here since you return from the previous block.

bruth:
@ReubenMathew it would be helpful to share a short summary of what the change addresses.

ReubenMathew:
> @ReubenMathew it would be helpful to share a short summary of what the change addresses.

I am still chewing on one of the changes; the one related to AppendEntries response handling is theoretically correct, but it's breaking a lot of unit tests. There's some behavior I'm not accounting for...

I will update the PR and take it out of draft when I figure it out.

server/raft.go
```diff
@@ -3619,6 +3613,20 @@ func (n *raft) processAppendEntryResponse(ar *appendEntryResponse) {
 		n.stepdownLocked(noLeader)
 		n.Unlock()
 		arPool.Put(ar)
+		return
```
neilalexander, line 3616: Is it right to `return` in this case? Looks like we will step down but won't actually process the AE we were sent.

ReubenMathew: I don't think we should process the response of a node that is ahead of us. They have another leader (since we're behind), so the request we sent should be ignored.

neilalexander:
Is this PR still relevant?

PR #5953, opened 2024-10-02
Added test on direct get with subject filter

aricart:
Signed-off-by: Your Name <alberto@nats.io>

### DIFF
--- a/server/jetstream_test.go
+++ b/server/jetstream_test.go
@@ -22486,6 +22486,52 @@ func TestJetStreamDirectGetBatch(t *testing.T) {
 	checkResponses(sub, 3, "foo.foo", "foo.bar", "foo.baz", _EMPTY_)
 }
 
+func TestJetStreamDirectGetBatchFiltered(t *testing.T) {
+	s := RunBasicJetStreamServer(t)
+	defer s.Shutdown()
+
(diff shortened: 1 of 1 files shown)


server/jetstream_test.go
```diff
@@ -22486,6 +22486,52 @@ func TestJetStreamDirectGetBatch(t *testing.T) {
...
+		require_NoError(t, err)
+		return sub
+	}
+
+	// Run some simple tests.
+	sub := sendRequest(&JSApiMsgGetRequest{Seq: 1, Batch: 1, MultiLastFor: []string{"a", "b"}})
+	defer sub.Unsubscribe()
+	msg, err := sub.NextMsg(10 * time.Millisecond)
+	require_NoError(t, err)
+	require_Equal(t, "1", msg.Header.Get(JSSequence))
```
aricart, line 22530: Expected the first message to be sequence `1` - tried with the request being `0` as well. But first message here is always sequence `2` - `b`

aricart: I think the main issue is that currently the filter for the subjects triggers as a last for subject - but need an option that is simply a subject filter without last semantics. Perhaps this could be aligned with consumer semantics where `last_per_subject` is a boolean, and we just have a `filter_subjects` that specifies the filter.

PR #5955, opened 2024-10-02
Refactor delayed API responses

kozlovic:
This is similar to @neilalexander PR (#5948) but has a slightly different approach. We use the struct to have a `next` so that we can use a list since the list will be maintained ordered. The for-loop is in one level, as opposed to two, which means that pushing to the IPQ will be "immediately" popped, even if we have a timer initialized and waiting to fire.
The timer is reset (not recreated everytime) only if the added response ends-up being at the head of the list.

Added a test that verifies that it works properly.

### DIFF
--- a/server/jetstream_api.go
+++ b/server/jetstream_api.go
@@ -1056,13 +1056,13 @@ func addDelayedResponse(head, tail **delayedAPIResponse, r *delayedAPIResponse) 
 func (s *Server) delayedAPIResponder() {
 	defer s.grWG.Done()
 	var (
-		rgQuitCh   <-chan struct{}
-		head, tail *delayedAPIResponse
-		r          *delayedAPIResponse
+		head, tail *delayedAPIResponse // Linked list.
+		r          *delayedAPIResponse // Updated by calling next().
+		rqch       <-chan struct{}     // Quit channel of the Raft group (if present).
 		tm         = time.NewTimer(time.Hour)
 	)
-	setup := func() {
-		r, rgQuitCh = nil, nil
+	next := func() {
+		r, rqch = nil, nil
 		// Check that JetStream is still on. Do not exit the go routine
 		// since JS can be enabled/disabled. The go routine will exit
 		// only if server is shutdown.
@@ -1073,14 +1073,21 @@ func (s *Server) delayedAPIResponder() {
 			s.delayedAPIResponses.drain()
 			// Fall back into next "if" that resets timer.
 		}
+		// If there are no delayed messages then delay the timer for
+		// a while.
 		if head == nil {
 			tm.Reset(time.Hour)
 			return
 		}
+		// Get the first expected message and then reset the timer.
 		r = head
 		js.mu.RLock()
 		if r.rg != nil && r.rg.node != nil {
-			rgQuitCh = r.rg.node.QuitC()
+			// If there's an attached Raft group to the delayed response
+			// then pull out the quit channel, so that we don't bother
+			// sending responses for entities which are now no longer
+			// running.
+			rqch = r.rg.node.QuitC()
 		}
 		js.mu.RUnlock()
 		tm.Reset(time.Until(r.deadline))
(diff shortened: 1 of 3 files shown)


server/jetstream_api.go
```diff
@@ -1887,12 +1978,12 @@ func (s *Server) jsStreamInfoRequest(sub *subscription, c *client, a *Account, s
 			if js.isLeaderless() {
 				resp.Error = NewJSClusterNotAvailError()
 				// Delaying an error response gives the leader a chance to respond before us
-				s.sendDelayedAPIErrResponse(ci, acc, subject, reply, string(msg), s.jsonResponse(&resp), nil)
+				s.sendDelayedAPIErrResponse(ci, acc, subject, reply, string(msg), s.jsonResponse(&resp), nil, errRespDelay)
```
kozlovic, line 1981: @neilalexander Since this was based on your PR, I kept that, but do you have plans in the future to have a different deadline? It seems that we always pass the constant as of now.

If we don't have plans to have different values, then this (your) PR could be simplified. We could set the deadline after the pop() (in the go routine handler - granted, that would disregard the time from push to pop, but do we really care since the goal is to delay the responses?), and there would not be a need for a list/sort since they would naturally be ordered.

neilalexander: The plan was to be able to use increasing deadlines for the rate limiting responses, yep. I just used `errRespDelay` in the current cases because that's the behaviour they have today.

kozlovic: @neilalexander Ok. How do you want to proceed? Using this PR or yours? If yours, please use the test from this PR and you will likely need to fix some of the issues in yours (I can comment on there). If you want to use this one, let me know and I will remove the Draft status.

neilalexander: I am planning to properly comb through this but overall I think your approach is good and certainly should be lighter on allocations/copies, so I'm very happy to close my PR and we can focus on this one.

server/jetstream_api.go
```diff
@@ -1012,30 +1012,121 @@ func (s *Server) sendAPIErrResponse(ci *ClientInfo, acc *Account, subject, reply
...
+		head, tail *delayedAPIResponse
+		r          *delayedAPIResponse
+		tm         = time.NewTimer(time.Hour)
+	)
+	setup := func() {
+		r, rgQuitCh = nil, nil
+		// Check that JetStream is still on. Do not exit the go routine
+		// since JS can be enabled/disabled. The go routine will exit
+		// only if server is shutdown.
+		js := s.getJetStream()
```
kozlovic, line 1069: The check for JS and the raft node quit channel, although working and the test checks that, may be a bit of an overkill. Does it really matter if a delayed response is sent or not after a raft group is shutdown or JS turned off? Again, code is there and test verifies it works, but we may have done without I think.

neilalexander:
Rebased on top of `main`.

server/jetstream_api.go
neilalexander: I think it makes sense that we don't send anything back after the Raft group is stopped, otherwise it might be a signal for the client/application to "try again" rather than just timing out in the rate-limited scenario.

server/jetstream_api.go
```diff
@@ -1012,30 +1012,121 @@ func (s *Server) sendAPIErrResponse(ci *ClientInfo, acc *Account, subject, reply
...
-		defer s.grWG.Done()
+func (s *Server) delayedAPIResponder() {
+	defer s.grWG.Done()
+	var (
+		rgQuitCh   <-chan struct{}
+		head, tail *delayedAPIResponse
+		r          *delayedAPIResponse
+		tm         = time.NewTimer(time.Hour)
+	)
+	setup := func() {
```
neilalexander, line 1064: Minor nit but I suggest we call this function `next` instead of `setup`, and perhaps add a comment that `r` is the next response we're expecting the timer to fire for, updated only by calling `next` again.

neilalexander (commented):
Overall looks good!

server/jetstream_api.go
neilalexander: In fact I've just done this myself in a new commit.

kozlovic: In one of my attempts, I had named it `next()`, so I would agree with that ;-).

derekcollison:
This one ready?

derekcollison (approved):
LGTM

PR #5971, opened 2024-10-07
[UPDATED] Updates for x/crypto, x/sys and x/time dependencies

derekcollison:


go.mod
```diff
@@ -1,6 +1,8 @@
 module github.com/nats-io/nats-server/v2
 
-go 1.22.0
+go 1.22
+
+toolchain go1.22.8
```
neilalexander, line 5: I feel like there was a reason the `go` line was 1.22.0 and not 1.22, @wallyqs can you remember?

wallyqs: At the time it was to avoid this issue: https://github.com/golang/go/issues/62278 but it seems it may have been fixed in Go v1.22.4: https://github.com/golang/go/issues/67236

wallyqs: Since we dropped Go v1.21 already, I think it should work as it is in the PR right now: https://github.com/golang/go/issues/62278#issuecomment-1693538776

wallyqs (approved):
LGTM

Issue #5929, opened 2024-09-25
Unrecoverable OOMKill during stream/consumer recovery on startup


jzhn:
### Observed behavior

- NATs pod enters CrashLoopBackoff state as it gets OOMKilled immediately after startup.

- Debug log shows no useful information. the last line of log mentions that NATs is `starting restore for stream`.
```
2024-09-25 06:01:53.177 | [3187] 2024/09/25 13:01:53.177737 [INF] Starting nats-server |  
-- | -- | --
  |   | 2024-09-25 06:01:53.177 | [3187] 2024/09/25 13:01:53.177828 [INF]   Version:  2.10.20 |  
  |   | 2024-09-25 06:01:53.177 | [3187] 2024/09/25 13:01:53.177830 [INF]   Git:      [7140387] |  
  |   | 2024-09-25 06:01:53.177 | [3187] 2024/09/25 13:01:53.177832 [DBG]   Go build: go1.22.6 |  
  |   | 2024-09-25 06:01:53.177 | [3187] 2024/09/25 13:01:53.177833 [INF]   Cluster:  kubernetes-nats |  
  |   | 2024-09-25 06:01:53.177 | [3187] 2024/09/25 13:01:53.177836 [INF]   Name:     kubernetes-nats-0 |  
  |   | 2024-09-25 06:01:53.177 | [3187] 2024/09/25 13:01:53.177840 [INF]   Node:     0EXN8Tyd |  
  |   | 2024-09-25 06:01:53.177 | [3187] 2024/09/25 13:01:53.177842 [INF]   ID:       NA7DSDLL34XFRE3M3NVTZCLIBZSVR7AB6KOP3IP5EVRGFVPJQ4IGGC7G |  
  |   | 2024-09-25 06:01:53.179 | [3187] 2024/09/25 13:01:53.179886 [INF] profiling port: 65432 |  
  |   | 2024-09-25 06:01:53.179 | [3187] 2024/09/25 13:01:53.179902 [INF] Using configuration file: /etc/nats-config/nats.conf |  
  |   | 2024-09-25 06:01:53.180 | [3187] 2024/09/25 13:01:53.179997 [DBG] Created system account: "$SYS" |  
  |   | 2024-09-25 06:01:53.180 | [3187] 2024/09/25 13:01:53.180343 [INF] Starting http monitor on 0.0.0.0:8222 |  
[...]

PR #5973, opened 2024-10-08
[FIXED] When writing tombstones we could violate max block size.

derekcollison:
This fixes that condition and also adds in compaction logic once we know the tombstones are no longer relevant.

Resolves: #5929  (Why we have large block exceeding maximum block size)


neilalexander (approved):
LGTM

PR #5978, opened 2024-10-08
[FIXED] Server panic on bad subject and infinite recursion.

derekcollison:
When a subject with embedded nulls was inserted into an stree followed by another one with more nulls we could recurse infinitely and panic the server.

We now changed the no pivot to 127(DEL) and enforce that you can not insert a subject with that byte. Also make sure we do not recursively call into insert with no cpi movement.

This condition was from an old server that had corrupted the PSIM data (known issue fixed), but we were not detecting it, so added test for this when checking runes under isValidSubject.


wallyqs (approved):
LGTM

PR #5980, opened 2024-10-09
[BUMP] jwt to v2.7.2

aricart:
[BUMP] jwt to v2.7.2 - which retracts case-sensitive jwt.TagList
[FIX] jwt cluster test reference ClusterTraffic as a string


derekcollison (approved):
LGTM

PR #5981, opened 2024-10-09
[FEAT] enable scoped users to have templates that are not limited to a subject token

aricart:
Relaxed template logic to allow a template in a scoped user to be in line with the subject, enabling uses like `KV_{{tag(foo)}}`. This allows more flexibility for some arcane cases even if the tag requirement is that it will still be lower-cased.

### DIFF
--- a/server/auth.go
+++ b/server/auth.go
@@ -521,7 +521,7 @@ func processUserPermissionsTemplate(lim jwt.UserPermissionLimits, ujwt *jwt.User
 					if len(valueList) != 0 {
 						values[tokenNum] = valueList
 					} else if failOnBadSubject {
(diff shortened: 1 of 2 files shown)


derekcollison (approved):
LGTM

derekcollison:
Need to fix this..

image

ripienaar:
nice, I'd be interested in exporting `processUserPermissionsTemplate` to provide a UI in `nats auth` that can show ACLs expanded during JWT info expanded just as the server would

aricart:
> nice, I'd be interested in exporting `processUserPermissionsTemplate` to provide a UI in `nats auth` that can show ACLs expanded during JWT info expanded just as the server would

I got that and better: https://github.com/aricart/aclgenerator

PR #5982, opened 2024-10-10
[FIXED] LeafNode's queue group load balancing and Sublist.NumInterest

### LINKED ISSUE #5972: Queue Groups on leaf clusters not balancing correctly when messages are routed in from hub cluster [v2.10.21]
### Observed behavior

From two subscribers in a queue group connected to two different nodes in a leaf cluster only one will receive messages IFF those messages arrive from a hub cluster to which the leaf cluster is connected. The second subscribers (initially idle) will receive messages once the primary is terminated. This is reversible (restart primary connected to the original node and will take over again).

### Expected behavior

All subcribers in the same queue group a leaf cluster are treated equally.

### Server and client version

Server 2.10.21
Nats - compiled from main

### Host environment

Windows/Mac/Linux - reproduced by customer, Borja and me

### Steps to reproduce

To reproduce deterministically, all leaf node connections and nats-cli connections will connect to specific nodes.

1. Create HUB cluster with nodes HUB1, HUB2, HUB3
2. Create LEAF cluster with  nodes LEAF1, LEAF2, LEAF3
3. Connect leaf nodes such that LEAF1-->HUB1, LEAF2-->HUB2 LEAF3-->HUB3
4. Start queue group listeners
6. `nats --server LEAF1 sub --queue=q1 foo `
7. `nats --server LEAF2 sub --queue=q1 foo `
8. Publish messages to HUB1
9. `nats --server HUB1 pub foo Hello`   //Only the subscriber on LEAF2 will receive messages
10. Publish messages to HUB3
11. `nats --server HUB3 pub foo Hello`   //Both subscribers receive messages

Tentative explanation: The "prefer local cluster listeners in work queues" logic is not leaf node aware. When the message is published such that both listeners have the same "distance" (routing hops) to the publisher the work queue LB works. When the message is published such that both listeners have the different "distance" to the publisher the work queue LB fails.

kozlovic:
While writing the test, I needed to make sure that each server in
the hub has registered interest for 2 queue subscribers from the
same group. I noticed that `Sublist.NumInterest()` (that I was
invoking from `Account.Interest()` was returning 1, even after
I knew that the propagation should have happened. It turns out
that `NumInterest()` was returning the number of queue groups, not
the number of queue subs in all those queue groups.
    
For the leafnode queue balancing issue, the code was favoring
local/routed queue subscriptions, so in the described issue,
the message would always go from HUB1->HUB2->LEAF2->QSub instead
of HUB1->LEAF1->QSub.
    
Since we had another test that was a bit reversed where we had
a HUB and LEAF1<->LEAF2 connecting to HUB and a qsub on both
HUB and LEAF1 and requests originated from LEAF2, and we were
expecting all responses to come from LEAF1 (instead of the
responder on HUB), I went with the following approach:
    
If the message originates from a client that connects to a server
that has a connection from a remote LEAF, then we pick that LEAF the
same as if it was a local client or routed server.
However, if the client connects to a server that has a leaf
connection to another server, then we keep track of the sub
but do not sent to that one if we have local or routed qsubs.
    
This makes the 2 tests pass, solving the new test and maintaining
the behavior for the old test.

Resolves #5972 

### DIFF
--- a/server/auth_callout_test.go
+++ b/server/auth_callout_test.go
@@ -250,6 +250,7 @@ func TestAuthCalloutBasics(t *testing.T) {
 
 	// This one will use callout since not defined in server config.
 	nc := at.Connect(nats.UserInfo("dlc", "zzz"))
+	defer nc.Close()
 
 	resp, err := nc.Request(userDirectInfoSubj, nil, time.Second)
 	require_NoError(t, err)
@@ -319,6 +320,7 @@ func TestAuthCalloutMultiAccounts(t *testing.T) {
 
 	// This one will use callout since not defined in server config.
 	nc := at.Connect(nats.UserInfo("dlc", "zzz"))
+	defer nc.Close()
 
 	resp, err := nc.Request(userDirectInfoSubj, nil, time.Second)
 	require_NoError(t, err)
@@ -373,13 +375,13 @@ func TestAuthCalloutAllowedAccounts(t *testing.T) {
 		t.Helper()
 
 		var nc *nats.Conn
-		defer nc.Close()
 		// Assume no auth user.
 		if password == "" {
 			nc = at.Connect()
 		} else {
 			nc = at.Connect(nats.UserInfo(user, password))
 		}
+		defer nc.Close()
 
 		resp, err := nc.Request(userDirectInfoSubj, nil, time.Second)
 		require_NoError(t, err)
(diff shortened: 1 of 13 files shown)


kozlovic:
@neilalexander I believe there was an issue with Sublist.NumInterest for queue subs since it looked like it was simply counting the number of groups, not the total number of queue subscriptions. Let me know if I misunderstood the intent.

@derekcollison Please review the PR description and see if the choice I made is ok.

kozlovic:
You can review the first commit for the leafnode/sublist issues. The second is simply a bunch of missing "defer nc.Close()" and the likes.

neilalexander:
Something else that's just occurred to me is that `NumInterest()` was never back ported into 2.10.x, so if there's a problem on those versions too (as opposed to just on `main`), it's probably because of the `Account.Interest()` doing `len(res.psubs) + len(res.qsubs)`.

@derekcollison Don't know whether we want to cherry-pick in `NumInterest()` into 2.10.x and apply this on top, or if we want to raise a separate PR against the `release/v2.10.22` branch to just fix `Account.Interest()`?

derekcollison:
@neilalexander let's pull those into 2.10.22 from main once this lands.

PR #5983, opened 2024-10-10
Update Go versions

wallyqs:
Signed-off-by: Waldemar Quevedo [wally@nats.io](mailto:wally@nats.io)

PR #5984, opened 2024-10-10
Cherry-picks for 2.10.22-RC.2

neilalexander:
Includes the following:

- #5918
- #5982
- #5983 (although only the 1.22.8 upgrade, since 1.21.x is no longer receiving updates)


derekcollison (approved):
LGTM

PR #5992, opened 2024-10-11
De-flake `TestNRGCandidateDontStepdownDueToLeaderOfPreviousTerm`

neilalexander:


neilalexander:
Looking into a race detector problem first...

neilalexander:
Think this has already been addressed in more recent PRs.

PR #5995, opened 2024-10-12
[FIXED] Max delivered stuck consumers and mis-reported redelivered stats.

derekcollison:
Process max delivery boundaries when expiring vs always putting back on the redelivered queue.

Previously we would always re-queue and do this on getNextMsg() which worked well for push consumers, but with pull based consumers would require a new pull request to be present to process any redelivered and this could report redelivered status incorrectly on max deliver of 1.


neilalexander (approved):
LGTM

PR #6009, opened 2024-10-15
Fixes to test JetStreamClusterStreamOrphanMsgsAndReplicasDrifting

mprimi:
- Wait for all routines (producers, consumers, chaos) to complete before checking stream state
 - Avoid leaking routines
 - Minor comment enhancements

### DIFF
--- a/server/jetstream_cluster_4_test.go
+++ b/server/jetstream_cluster_4_test.go
@@ -785,17 +785,83 @@ func TestJetStreamClusterConsumerPauseSurvivesRestart(t *testing.T) {
 }
 
 func TestJetStreamClusterStreamOrphanMsgsAndReplicasDrifting(t *testing.T) {
-	type testParams struct {
-		restartAny       bool
-		restartLeader    bool
-		rolloutRestart   bool
-		ldmRestart       bool
-		restarts         int
-		checkHealthz     bool
-		reconnectRoutes  bool
-		reconnectClients bool
+
+	const PublishDuration = 30 * time.Second
+	const MaxPublishedPerProducer = 200_000
+	const MessagesSize = 1024
+	const NumPublishers = 50
+	const NumConsumersPerSubject = 5
+	const ConsumersStartDelay = 15 * time.Second
+	const ChecksDelay = 5 * time.Second
+	const ConsumeDuration = 45 * time.Second
+	const ConsumersFirstLifecycleDuration = 5 * time.Second
+	const FetchSize = 10
+	const DisconnectRoutesInterval = 8 * time.Second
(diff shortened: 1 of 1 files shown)


wallyqs (approved):
LGTM

server/jetstream_cluster_4_test.go
```diff
@@ -1129,6 +1132,8 @@ func TestJetStreamClusterStreamOrphanMsgsAndReplicasDrifting(t *testing.T) {
 							for range time.NewTicker(2 * time.Second).C {
 								select {
 								case <-hctx.Done():
+									t.Fatalf("Timeout checking for health of %s", s.Name())
```
wallyqs, line 1135: think this line above needs fixing: `s = c.restartServer(s)`

mprimi: Thank you 👍

server/jetstream_cluster_4_test.go
```diff
@@ -1371,19 +1537,19 @@ func TestJetStreamClusterStreamOrphanMsgsAndReplicasDrifting(t *testing.T) {
...
+			checkHealthz:      true,
+			disconnectRoutes:  true,
+			disconnectClients: true,
 		}
 		test(t, params, &nats.StreamConfig{
 			Name:       "OWQTEST_R3F_DO_NOLIMIT",
 			Subjects:   []string{"MSGS.>"},
 			Replicas:   3,
 			Duplicates: 30 * time.Second,
+			Retention:  nats.WorkQueuePolicy,
```
wallyqs, line 1552: this one not meant to be a workqueue

derekcollison: What should it be @wallyqs ?

wallyqs, line 1552:
```suggestion
			Retention:  nats.LimitsPolicy,
```

wallyqs: should be LimitsPolicy, added to the PR

mprimi: Should the name be updated? it contains `WQ` and `NO_LIMITS`.
The test logic also need to change, we can't expect `stream.messages` to match `SUM(consumer.pending)`

wallyqs: it is a limits based stream without limits, I would remove the WQ from the stream name which I left in... but it is covering a particular failure mode, you can rename it into something like `TEST_R3F_DO_LIMITS` (just limits based with discard old)

mprimi: Ok, moving this PR back to draft until i address it.

PR #6014, opened 2024-10-17
Restore consumer `OptStartSeq` clipping behaviour

neilalexander:
This PR backs out #5785, which was released in v2.10.19, as it has a negative impact on sourcing and mirroring. See discussion at #6005.

Two new unit tests, `TestJetStreamSourcingClipStartSeq` and `TestJetStreamMirroringClipStartSeq`, are added to show this behaviour in action.

We will probably want to revisit this for 2.11 so that we can come up with something that works for both use-cases.


derekcollison (approved):
LGTM

PR #6019, opened 2024-10-18
Windows: Set `CERT_STORE_READONLY_FLAG` when opening cert stores

### LINKED ISSUE #6018: Server won't start when TLS configuration uses "WindowsLocalMachine" store
### Observed behavior

I am using a certificate from the Windows Certificate Store ([see also](https://github.com/nats-io/nats-architecture-and-design/blob/main/adr/ADR-39.md)) in my TLS configuration. The certificate is stored in the Windows Local Machine certificate store.

```
port: 3333
http_port: 4444
tls: {
  cert_store: "WindowsLocalMachine"
  cert_match_by: "Subject"
  cert_match: "TestCertificate"
  verify: true
  timeout: 2
}
jetstream: enabled
jetstream: {
  max_memory_store: 1G
  max_file_store: 100G
}
```

The server refuses to start, with the following error:

```
nats-server: nats-streaming.conf:3:1: unable to open certificate store or store not available
```

### Expected behavior

The server should load the specified certificate from the Local Machine store, and start successfully.

### Server and client version

nats-server: v2.10.22

### Host environment

Windows 10, AMD64

### Steps to reproduce

This problem goes away if I use the `WindowsCurrentUser` store, or if I run `nats-server.exe` in **admin** mode. Unfortunately corporate policy does not allow either of these workarounds. 

Note: This does **not** seem to be a permissions issue, as there are other applications running in the same (non-administrative) context which _can_ access the Local Machine certificate store. It seems to be an issue with the way NATS attempts access.

neilalexander:
This should allow access to `CERT_SYSTEM_STORE_LOCAL_MACHINE` from non-admin users.

Fixes #6018.

### DIFF
--- a/server/certstore/certstore_windows.go
+++ b/server/certstore/certstore_windows.go
@@ -41,20 +41,21 @@ import (
 
 const (
 	// wincrypt.h constants
-	winAcquireCached           = 0x1                                                   // CRYPT_ACQUIRE_CACHE_FLAG
-	winAcquireSilent           = 0x40                                                  // CRYPT_ACQUIRE_SILENT_FLAG
-	winAcquireOnlyNCryptKey    = 0x40000                                               // CRYPT_ACQUIRE_ONLY_NCRYPT_KEY_FLAG
-	winEncodingX509ASN         = 1                                                     // X509_ASN_ENCODING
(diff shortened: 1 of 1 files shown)


derekcollison (approved):
LGTM

PR #6031, opened 2024-10-22
New test group and pipeline for long-running tests

mprimi:
This pull request introduces a new group of tests for server: long-running test.
Anything that takes more than a few seconds can be relocated here to speedup the regular CI flow.

 - Add a new test group
 - Add an example test (KV Put for 3 minutes with server restarts)
 - Set up [GH workflow](https://github.com/nats-io/nats-server/actions/workflows/long-tests.yaml) (scheduled daily) 

### DIFF
--- a/.github/workflows/long-tests.yaml
+++ b/.github/workflows/long-tests.yaml
@@ -1,7 +1,11 @@
-name: NATS Server Long-Running Tests
+name: NATS Server Long Tests
 
 on:
+  # Allow manual trigger (any branch)
   workflow_dispatch:
+  # Run daily at 12:30 on default branch
+  schedule:
+    - cron: "30 12 * * *"
 
 concurrency:
   # At most one of these workflow per ref running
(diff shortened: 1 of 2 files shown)


scripts/runTestsOnTravis.sh
```diff
@@ -10,7 +10,7 @@ if [ "$1" = "compile" ]; then
     go install github.com/golangci/golangci-lint/cmd/golangci-lint@v1.60.3;
     golangci-lint run;
     if [ "$TRAVIS_TAG" != "" ]; then
-        go test -race -v -run=TestVersionMatchesTag ./server -ldflags="-X=github.com/nats-io/nats-server/v2/server.serverVersion=$TRAVIS_TAG" -count=1 -vet=off
+        go test -race -v -run=TestVersionMatchesTag ./server -ldflags="-X=github.com/nats-io/nats-server/v2/server.serverVersion=$TRAVIS_TAG" -count=1 -vet=off -tags=include_js_long_tests
```
wallyqs, line 13: `-tags=include_js_long_tests` needed here?

mprimi: Whoops, good catch, thank you.

.github/workflows/long-tests.yaml
```diff
@@ -10,9 +14,20 @@ concurrency:
...
-      - name: Dummy step
-        run: echo "Hello World"
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Install Go
+        uses: actions/setup-go@v5
+        with:
+          # Use go version from go.mod
+          go-version-file: "go.mod"
```
alexbozhenko, line 28: @mprimi 
In go.mod we specify the lowest supported go version (1.22.0) at the moment, but we actually build releases with the version specified in travis:
https://github.com/nats-io/nats-server/blob/5e6017135b4b1d333b435ff55b720275a39bb7af/.travis.yml#L64-L67
(1.23.5 at the moment)
So this test runs with go toolchain version that is not used for the release
https://github.com/actions/setup-go?tab=readme-ov-file#getting-go-version-from-the-gomod-file

Was this an intention here?

PR #6039, opened 2024-10-24
Reduce allocations in consumers

neilalexander:
This PR makes the following changes that should reduce allocations in consumers:

* `JSApiConsumerGetNextRequest` structs are now pooled, as they were unavoidably escaping to the heap;
* `waitingRequest` structs are now returned to the pool correctly if hitting `MaxWaiting`;
* Reuse the `hdr` buffer from pooled `jsPubMsg` when returning consumer errors to the client.

### DIFF
--- a/server/consumer.go
+++ b/server/consumer.go
@@ -3547,6 +3547,7 @@ func (o *consumer) processNextMsgRequest(reply string, msg []byte) {
 
 	if err := o.waiting.add(wr); err != nil {
 		sendErr(409, "Exceeded MaxWaiting")
+		wr.recycle()
 		return
 	}
 	o.signalNewMessages()
--- a/server/consumer.go
+++ b/server/consumer.go
@@ -3126,6 +3126,14 @@ func (o *consumer) needAck(sseq uint64, subj string) bool {
 	return needAck
 }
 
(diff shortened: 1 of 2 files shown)


server/consumer.go
```diff
@@ -3135,7 +3143,11 @@ func nextReqFromMsg(msg []byte) (time.Time, int, int, bool, time.Duration, time.
 		return time.Time{}, 1, 0, false, 0, time.Time{}, nil
 
 	case req[0] == '{':
-		var cr JSApiConsumerGetNextRequest
+		cr := jsGetNextPool.Get().(*JSApiConsumerGetNextRequest)
+		defer func() {
+			*cr = JSApiConsumerGetNextRequest{}
+			jsGetNextPool.Put(cr)
```
kozlovic, line 3149: The pool returns normally a pointer to `JSApiConsumerGetNextRequest` struct, and cr is a struct, not a pointer to it. So is this correct or should it be `&cr`?

neilalexander: This is correct, yes, as `cr` is `*JSApiConsumerGetNextRequest`. `&cr` would be a pointer to a pointer.

kozlovic: Oh right, `cr` used to be a struct but is now a pointer to the struct. Sorry for the noise.

PR #6041, opened 2024-10-25
NRG: Use configured sync intervals for group stores

neilalexander:

### DIFF
--- a/server/filestore.go
+++ b/server/filestore.go
@@ -29,6 +29,7 @@ import (
 	"io"
 	"io/fs"
 	"math"
+	mrand "math/rand/v2"
 	"net"
 	"os"
 	"path/filepath"
@@ -7946,7 +7947,11 @@ func (fs *fileStore) setSyncTimer() {
 	if fs.syncTmr != nil {
 		fs.syncTmr.Reset(fs.fcfg.SyncInterval)
 	} else {
-		fs.syncTmr = time.AfterFunc(fs.fcfg.SyncInterval, fs.syncBlocks)
+		// First time this fires will be any time up to the fs.fcfg.SyncInterval,
+		// so that different stores are spread out, rather than having many of
+		// them trying to all sync at once, causing blips and contending dios.
+		start := time.Duration(mrand.Int64N(int64(fs.fcfg.SyncInterval)))
+		fs.syncTmr = time.AfterFunc(min(start, time.Second), fs.syncBlocks)
 	}
 }
 
--- a/server/jetstream_cluster.go
+++ b/server/jetstream_cluster.go
@@ -782,8 +782,12 @@ func (js *jetStream) setupMetaGroup() error {
 	sysAcc := s.SystemAccount()
 	storeDir := filepath.Join(js.config.StoreDir, sysAcc.Name, defaultStoreDirName, defaultMetaGroupName)
 
+	js.srv.optsMu.RLock()
+	syncAlways := js.srv.opts.SyncAlways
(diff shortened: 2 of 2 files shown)


server/jetstream_cluster.go
```diff
@@ -2078,9 +2087,19 @@ func (js *jetStream) createRaftGroup(accName string, rg *raftGroup, storage Stor
 	storeDir := filepath.Join(js.config.StoreDir, sysAcc.Name, defaultStoreDirName, rg.Name)
 	var store StreamStore
 	if storage == FileStorage {
+		// If the server is set to sync always, do the same for the Raft log.
+		js.srv.optsMu.RLock()
+		syncAlways := js.srv.opts.SyncAlways
+		syncInterval := js.srv.opts.SyncInterval
```
derekcollison, line 2093: Does this match 5 mins? Noticed this was slightly more delayed then default 2m.

server/jetstream_cluster.go
```diff
@@ -782,9 +782,18 @@ func (js *jetStream) setupMetaGroup() error {
...
+	js.srv.optsMu.RUnlock()
 	fs, err := newFileStoreWithCreated(
-		FileStoreConfig{StoreDir: storeDir, BlockSize: defaultMetaFSBlkSize, AsyncFlush: false, srv: s},
-		StreamConfig{Name: defaultMetaGroupName, Storage: FileStorage},
+		FileStoreConfig{
+			StoreDir: storeDir, BlockSize: defaultMetaFSBlkSize, AsyncFlush: false,
+			SyncAlways: syncAlways, SyncInterval: syncInterval, srv: s,
+		},
+		StreamConfig{
+			Name: defaultMetaGroupName, Storage: FileStorage,
```
derekcollison, line 795: nit can we go back to 1 line for these?

server/jetstream_cluster.go
neilalexander: They will now match the default 2 minutes. There's nothing in the git history that explains why it was 5 minutes.

server/jetstream_cluster.go
neilalexander: Done.

server/jetstream_cluster.go
derekcollison: We probably would like to not have them firing at the same time.

neilalexander: I've pushed another commit with an idea, curious on your thoughts: what if the stores used a random interval between 0 and `fs.fcfg.SyncInterval` for the first timer fire, and then tick along on `fs.fcfg.SyncInterval` after that? That way stores syncing should be spread out more.

server/filestore.go
```diff
@@ -7946,7 +7947,11 @@ func (fs *fileStore) setSyncTimer() {
 	if fs.syncTmr != nil {
 		fs.syncTmr.Reset(fs.fcfg.SyncInterval)
 	} else {
-		fs.syncTmr = time.AfterFunc(fs.fcfg.SyncInterval, fs.syncBlocks)
+		// First time this fires will be any time up to the fs.fcfg.SyncInterval,
+		// so that different stores are spread out, rather than having many of
+		// them trying to all sync at once, causing blips and contending dios.
+		start := time.Duration(mrand.Int64N(int64(fs.fcfg.SyncInterval)))
```
derekcollison, line 7953: This could return 0, should we add a fixed floor?

neilalexander: Sure, will do.

derekcollison (approved):
LGTM

PR #6042, opened 2024-10-25
Windows certificate store improvements

### LINKED ISSUE #4383: Handling of expired certificates in Windows Certificate Store

### LINKED ISSUE #6024: Server loads incorrect certificate from Windows Certificate Store
### Observed behavior

I start `nats-server.exe` with the following configuration:

```
tls: {
  cert_store: "WindowsLocalMachine"
  cert_match_by: "Subject"
  cert_match: "MyCertificate"
  timeout: 2
}
jetstream: enabled
host: my.local.machine
```

My Windows Certificate Store has _multiple_ test certificates defined, with similar subject names. One is called `MyCertificate`, another `MyCertificateForClient`. The former is fully trusted (as it is also registered as a Trusted Root CA), and is the certificate I expect the server to load. 

### Expected behavior

The server loads certificate `MyCertificateForClient`. I am unable to connect to the server, with the following error:
```
nats: error: tls: failed to verify certificate: x509: certificate signed by unknown authority
```

### Server and client version

nats-server: v2.10.22

### Host environment

Windows 10, AMD64

### Steps to reproduce

It looks like this is due to use of `CERT_FIND_ISSUER_STR` or `CERT_FIND_SUBJECT_STR`, more details [here](https://learn.microsoft.com/en-us/windows/win32/api/wincrypt/nf-wincrypt-certfindcertificateinstore). The lookup is for a certificate that "contains the specified string", which both of my test certificates satisfy. 

Furthermore, certificates within the store are unordered, and I also suspect that the server could retrieve _expired_, _revoked_ or _untrusted_ certificates in favour of one that can actually be used. 

I found a comment with similar issue [here](https://github.com/nats-io/nats-server/issues/2130#issuecomment-1655164371). The [follow up here](https://github.com/nats-io/nats-server/issues/2130#issuecomment-1656235997) makes sense to me - if I could specify a certificate by _thumbprint_ I could guarantee that the correct certificate is loaded. 

### See also
https://github.com/nats-io/nats-server/issues/2130
https://github.com/nats-io/nats-architecture-and-design/blob/main/adr/ADR-39.md

neilalexander:
This PR:

* Removes some magic numbers from the certstore code in favour of constants already defined in `x/sys/windows`
* Adds the `thumbprint` option to `cert_match_by` by allowing matching a specific certificate my SHA1 thumbprint rather than possibly matching multiple certificates by name
* Adds the `cert_match_skip_invalid` option by integrating & rebasing a community PR along with some fix-ups

Fixes #6024 
Fixes #4383
Closes #4384


derekcollison (approved):
LGTM

PR #6047, opened 2024-10-28
Fix matching by `thumbprint` in the Windows certificate store

neilalexander:
Fixes a bug in #6042 where the `thumbprint` would never match, as we were passing in a hex-decoded form instead of the actual string value.


derekcollison (commented):
Should we have a test?

derekcollison (approved):
LGTM

PR #6052, opened 2024-10-29
Refactor encoding replicated stream messages to properly enforce framing

neilalexander:
This PR fixes two bugs in `encodeStreamMsgAllowCompress`:

* The boundary lengths could have overflowed when casting potentially large values down to uint16 or uint32 and we wouldn't have noticed;
* We would then copy in the subject/reply/header/msg with the original length regardless, not the overflowed length, meaning the decoding side would have the wrong boundaries.

Now we will correctly calculate the bounds, avoiding overflows, and only append as many bytes as we clipped down to.

### DIFF
--- a/server/jetstream_cluster.go
+++ b/server/jetstream_cluster.go
@@ -7691,54 +7691,46 @@ const compressThreshold = 8192 // 8k
 
 // If allowed and contents over the threshold we will compress.
 func encodeStreamMsgAllowCompress(subject, reply string, hdr, msg []byte, lseq uint64, ts int64, compressOK bool) []byte {
-	shouldCompress := compressOK && len(subject)+len(reply)+len(hdr)+len(msg) > compressThreshold
+	// Clip the subject, reply, header and msgs down. Operate on
+	// uint64 lengths to avoid overflowing.
+	slen := min(uint64(len(subject)), math.MaxUint16)
+	rlen := min(uint64(len(reply)), math.MaxUint16)
+	hlen := min(uint64(len(hdr)), math.MaxUint16)
+	mlen := min(uint64(len(msg)), math.MaxUint32)
+	total := slen + rlen + hlen + mlen
 
-	elen := 1 + 8 + 8 + len(subject) + len(reply) + len(hdr) + len(msg)
+	shouldCompress := compressOK && total > compressThreshold
+	elen := int(1 + 8 + 8 + total)
 	elen += (2 + 2 + 2 + 4) // Encoded lengths, 4bytes
-	// TODO(dlc) - check sizes of subject, reply and hdr, make sure uint16 ok.
-	buf := make([]byte, elen)
+
+	buf := make([]byte, 1, elen)
 	buf[0] = byte(streamMsgOp)
+
 	var le = binary.LittleEndian
-	wi := 1
-	le.PutUint64(buf[wi:], lseq)
-	wi += 8
-	le.PutUint64(buf[wi:], uint64(ts))
-	wi += 8
-	le.PutUint16(buf[wi:], uint16(len(subject)))
-	wi += 2
-	copy(buf[wi:], subject)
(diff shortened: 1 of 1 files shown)


server/jetstream_cluster.go
```diff
@@ -7691,54 +7691,46 @@ const compressThreshold = 8192 // 8k
 
 // If allowed and contents over the threshold we will compress.
 func encodeStreamMsgAllowCompress(subject, reply string, hdr, msg []byte, lseq uint64, ts int64, compressOK bool) []byte {
-	shouldCompress := compressOK && len(subject)+len(reply)+len(hdr)+len(msg) > compressThreshold
-
-	elen := 1 + 8 + 8 + len(subject) + len(reply) + len(hdr) + len(msg)
+	// Clip the subject, reply, header and msgs down. Operate on
+	// uint64 lengths to avoid overflowing.
+	slen := min(uint64(len(subject)), math.MaxUint16)
```
derekcollison, line 7696: If we clip that is bad though and we should not try to force through?

server/jetstream_cluster.go
```diff
@@ -7691,54 +7691,46 @@ const compressThreshold = 8192 // 8k
...
-		copy(buf[wi:], hdr)
-		wi += len(hdr)
-	}
-	le.PutUint32(buf[wi:], uint32(len(msg)))
-	wi += 4
-	if len(msg) > 0 {
-		copy(buf[wi:], msg)
-		wi += len(msg)
-	}
+	buf = le.AppendUint64(buf, lseq)
```
derekcollison, line 7710: Last time I checked Append was 5x slower then Put.. Hence why I did not use it.

server/jetstream_cluster.go
```diff
@@ -7691,54 +7691,46 @@ const compressThreshold = 8192 // 8k
...
+	buf = le.AppendUint32(buf, uint32(mlen))
+	buf = append(buf, msg[:mlen]...)
 
 	// Check if we should compress.
 	if shouldCompress {
 		nbuf := make([]byte, s2.MaxEncodedLen(elen))
 		nbuf[0] = byte(compressedStreamMsgOp)
-		ebuf := s2.Encode(nbuf[1:], buf[1:wi])
+		ebuf := s2.Encode(nbuf[1:], buf[1:])
 		// Only pay cost of decode the other side if we compressed.
```
derekcollison, line 7726: "Only pay the cost of decode on the other side..."

server/jetstream_cluster.go
neilalexander: It's only slower if the backing array isn't preallocated, but in this case it is, as we already provided a recalculated capacity to `buf`.

server/jetstream_cluster.go
neilalexander: Maybe, but if we want to enforce sizes, we should be doing that when the message comes into the system, otherwise we could have different behaviour for clustered vs non-clustered setups.

server/jetstream_cluster.go
neilalexander: Fixed.

server/jetstream_cluster.go
derekcollison: Good to know.

server/jetstream_cluster.go
derekcollison: What will happen to this message if we clip? Can it be retrieved?

neilalexander: Yes, it'll go into the stream as normal and be retrievable. It'll just be truncated. For truncation to happen the subject/reply/headers would need to exceed 64KB and the message would need to exceed 4GB.

derekcollison: If we clip the subject then probably not retrievable. But understand that is exceeding 64k.

derekcollison (approved):
LGTM

PR #6060, opened 2024-10-31
NRG: Do not set `pterm`/`pindex` if no snapshots and/or log entries

neilalexander:
If we haven't recovered the `pterm`/`pindex` from a snapshot, and the WAL is empty, then we shouldn't tell other nodes that we have a log that we don't have (i.e. by vote request). Instead leave both `pterm` and `pindex` as zero to correctly signal that we don't know anything about the state of the log.


derekcollison (approved):
LGTM

PR #6064, opened 2024-10-31
[FIXED] Fix request/reply performance when using allow_responses perms

### LINKED ISSUE #6058: Severe request/reply performance hit when using allow_responses map [v2.10.20]
### Observed behavior

When using the `allow_responses` map in user permissions with `max` set, request/reply takes a significant performance hit depending on the value of `expires`. If the `max` value is set but the `expires` value is unset, performance continually degrades over time. 
Even if using `expires: "10s"`, the request/reply throughput drops by about 70% (on my machine). 

Not sure if there's any way around this - although I haven't looked into the server code, I would imagine having to cache the request subjects for the max reply count indefinitely would cause memory usage to creep up and performance to go down over time. 
That said, I think this should really be either fixed (if possible) or at least documented so that it's clear that using the `allow_responses` map will have an impact on request/reply performance. It might also be a good idea to just not allow specifying `max` without an expiry, as it seems like this will always lead to a memory leak. 

### Expected behavior

Performance should not degrade or it should be documented 

### Server and client version

v2.10.20

### Host environment

Tested with the latest docker image on an ubuntu server and separately with the server binary in a WSL ubuntu environment

### Steps to reproduce

nats.conf with no issues:
```
host: 0.0.0.0

authorization {
  users = [
    {user: test, password: test, permissions: { publish: ">", subscribe: ">" }}
  ]
}
```

nats.conf with worse performance:
```
host: 0.0.0.0

authorization {
  users = [
    {user: test, password: test, permissions: { publish: ">", subscribe: ">", allow_responses: { max: 1, expires: "10s" } }}
  ]
}
```

nats.conf with memory leak:
```
host: 0.0.0.0

authorization {
  users = [
    {user: test, password: test, permissions: { publish: ">", subscribe: ">", allow_responses: { max: 1 } }}
  ]
}
```

Using nats cli (`nats`) and nats server (`nats-server`) binaries in three terminals:

`./nats-server -c nats.conf`
`./nats bench test --sub 1 --reply --user test --password test`
`./nats bench test --pub 1 --request --user test --password test`

jack7803m:
Fixes performance issues noted in #6058. Attempts to prune reply map every `replyPermLimit` messages or if it has been more than `replyPruneTime` since the last prune.

Resolves #6058 

Signed-off-by: Jack Morris <jack@jackmorris.me>

### DIFF
--- a/server/client.go
+++ b/server/client.go
@@ -3767,7 +3767,7 @@ func (c *client) pruneReplyPerms() {
 	}
 
 	c.repliesSincePrune = 0
-	c.lastReplyPrune = time.Now()
+	c.lastReplyPrune = now
 }
 
 // pruneDenyCache will prune the deny cache via randomly
--- a/server/client.go
+++ b/server/client.go
@@ -3638,7 +3638,8 @@ func (c *client) deliverMsg(prodIsMQTT bool, sub *subscription, acc *Account, su
 
 	// If we are tracking dynamic publish permissions that track reply subjects,
 	// do that accounting here. We only look at client.replies which will be non-nil.
-	if client.replies != nil && len(reply) > 0 {
+	// Only reply subject permissions if the client is not already allowed to publish to the reply subject.
+	if client.replies != nil && len(reply) > 0 && !client.pubAllowedFullCheck(string(reply), true, true) {
 		client.replies[string(reply)] = &resp{time.Now(), 0}
 		client.repliesSincePrune++
 		if client.repliesSincePrune > replyPermLimit || time.Since(client.lastReplyPrune) > replyPruneTime {
@@ -3836,7 +3837,7 @@ func (c *client) pubAllowedFullCheck(subject string, fullCheck, hasLock bool) bo
 		allowed = np == 0
 	}
 
-	// If we are currently not allowed but we are tracking reply subjects
+	// If we are tracking reply subjects
 	// dynamically, check to see if we are allowed here but avoid pcache.
 	// We need to acquire the lock though.
 	if !allowed && fullCheck && c.perms.resp != nil {
--- a/server/client.go
+++ b/server/client.go
@@ -262,6 +262,9 @@ type client struct {
 	last       time.Time
 	lastIn     time.Time
 
+	repliesSincePrune uint16
+	lastReplyPrune    time.Time
+
 	headers bool
 
 	rtt      time.Duration
@@ -422,6 +425,7 @@ const (
 	pruneSize            = 32
 	routeTargetInit      = 8
 	replyPermLimit       = 4096
+	replyPruneTime       = time.Second
 )
 
 // Represent read cache booleans with a bitmask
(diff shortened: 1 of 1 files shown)


jack7803m:
Unsure how those failing tests could be affected by the minimal changes I made. 

Still need to add some sort of solution to the infinite expiry, though I'm not sure exactly what direction to go with that (i.e. error or set to a default), so I'll leave that decision up to the maintainers.

server/client.go
```diff
@@ -3760,6 +3765,9 @@ func (c *client) pruneReplyPerms() {
 			delete(c.replies, k)
 		}
 	}
+
+	c.repliesSincePrune = 0
+	c.lastReplyPrune = time.Now()
```
derekcollison, line 3770: Can just now from above.

server/client.go
```diff
@@ -3636,7 +3640,8 @@ func (c *client) deliverMsg(prodIsMQTT bool, sub *subscription, acc *Account, su
 	// do that accounting here. We only look at client.replies which will be non-nil.
 	if client.replies != nil && len(reply) > 0 {
 		client.replies[string(reply)] = &resp{time.Now(), 0}
-		if len(client.replies) > replyPermLimit {
+		client.repliesSincePrune++
+		if client.repliesSincePrune > replyPermLimit || time.Since(client.lastReplyPrune) > replyPruneTime {
```
derekcollison, line 3644: Do we have a sense under heaby load how much more memory this will hold onto?

server/client.go
jack7803m: Good point! Just fixed it

server/client.go
jack7803m: The original issue was it holding onto too much memory and looping through all of it every message. 

Just added some debug statements and found that if the reply subject is already allowed (eg. `"pub": ">"`), then the reply counter never actually goes up and therefore it never is able to prune that subject out until it expires by time. Just noting this because I'm going to be looking at a fix for that too but not sure if it'll have broader effects (hopefully not).

Assuming that the subjects are getting pruned as they're replied to, at most it should only be able to get the `replies` map to an extra `replyPermLimit` size than what it could've possible been before in the worst case scenario. Even for that to happen it would have to fill the map with subjects, attempt to prune, then expire them all by the next message - in that case, pruning for every message over the `replyPermLimit` would cause it to immediately prune whereas this solution would hold onto that memory for the next `replyPermLimit` messages, making the map size `replyPermLimit * 2`. 

Under normal heavy load it shouldn't make any significant difference, as the current behavior typically should only run the prune once every `replyPermLimit` messages anyway when it's configured properly.

MauriceVanVeen (approved):
LGTM

@jack7803m, the PR's title mentions it's WIP, but is this ready/good to merge?

@neilalexander, would you maybe also want to review, given your comment here? https://github.com/nats-io/nats-server/issues/6058#issuecomment-2449631175.
Think this PR would at least relieve some pressure, by not calling `client.pruneReplyPerms()` every single time just because the map is large enough.

jack7803m:
Forgot to change the title - this should be good to merge!

derekcollison:
Let's have @neilalexander take a look as well real quick, but then we can get this merged.

neilalexander (approved):
I think we probably want to address the non-expiring replies long-term but for now I'm OK with this to improve performance, LGTM.

PR #6066, opened 2024-11-01
Group metalayer consumer assignments by stream during recovery

neilalexander:
This updates the metalayer recovery so that consumer assignments are grouped by the stream recovery key, such that a stream removal can also clean up any pending consumer updates.

Previously, if the metalayer recovered the following:
1. Stream update assignment
2. Consumer update assignment
3. Stream remove assignment
4. Stream update assignment

... then 2 would survive the recovery process because 4 cancelled out 3 and 3 didn't cancel out 2.

Co-authored-by: Maurice van Veen <github@mauricevanveen.com>


MauriceVanVeen (approved):
LGTM

derekcollison (approved):
LGTM

PR #6067, opened 2024-11-01
Improve metalayer snapshot on shutdown

neilalexander:
We can't know whether `s.quitCh` or `qch` will fire first, since the `select` order is indeterminate. Try to snapshot either way.


MauriceVanVeen (approved):
LGTM

derekcollison (approved):
LGTM

PR #6086, opened 2024-11-07
Update params in `TestJetStreamClusterStreamOrphanMsgsAndReplicasDrifting`

neilalexander:
This may not fix this test entirely, but right now all of the different R=3 permutations are doing different things, so the results that it gives us are not comparatively useful at all. Instead this PR harmonises all the R=3 tests.

### DIFF
--- a/server/jetstream_cluster_4_test.go
+++ b/server/jetstream_cluster_4_test.go
@@ -1299,7 +1299,7 @@ func TestJetStreamClusterStreamOrphanMsgsAndReplicasDrifting(t *testing.T) {
 			ldmRestart:     true,
 			rolloutRestart: false,
 			restarts:       1,
-			checkHealthz:   false,
+			checkHealthz:   true,
 		}
 		test(t, params, &nats.StreamConfig{
 			Name:        "OWQTEST_R3M",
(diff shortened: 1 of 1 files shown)


derekcollison:
@wallyqs would like your feedback here as well.

server/jetstream_cluster_4_test.go
```diff
@@ -1369,13 +1371,11 @@ func TestJetStreamClusterStreamOrphanMsgsAndReplicasDrifting(t *testing.T) {
 	// Clustered file based with discard old policy and no limits.
 	t.Run("R3F_DO_NOLIMIT", func(t *testing.T) {
 		params := &testParams{
-			restartAny:       false,
-			ldmRestart:       true,
-			rolloutRestart:   true,
-			restarts:         3,
-			checkHealthz:     true,
-			reconnectRoutes:  true,
```
wallyqs, line 1377: These were to trigger some specific failure conditions and as it is this version of the test is at 99% reliability at the moment, but to track down the current issue I think sounds good to change these to being the same for now.

server/jetstream_cluster_4_test.go
```diff
@@ -1349,6 +1350,7 @@ func TestJetStreamClusterStreamOrphanMsgsAndReplicasDrifting(t *testing.T) {
 			ldmRestart:     true,
 			rolloutRestart: false,
 			restarts:       1,
+			checkHealthz:   true,
```
wallyqs, line 1353: Getting similar results with `checkHealthz` enabled here but we can give it a try to add since using healthz would be closer to what a usual rollout would look like.

wallyqs (approved):
LGTM, we can give it a try but does not seem to change the results from `TestJetStreamClusterStreamOrphanMsgsAndReplicasDrifting/R3F_DO` sub test.

PR #6089, opened 2024-11-08
Introduce a NumPendingMulti that uses a sublist for efficient selection.

derekcollison:
Consumers with many filtered subjects and large streams could suffer from slow creation times due to NumPending() being called for all filtered subjects.

This improvement is similar to the LoadNextMsgMulti for consumers with large number of filtered subjects.

Also fixed a bug in normal NumPending for memstore.

### DIFF
--- a/server/consumer.go
+++ b/server/consumer.go
@@ -1077,7 +1077,7 @@ func (mset *stream) addConsumerWithAssignment(config *ConsumerConfig, oname stri
 	// If we have multiple filter subjects, create a sublist which we will use
 	// in calling store.LoadNextMsgMulti.
 	if len(o.cfg.FilterSubjects) > 0 {
-		o.filters = NewSublistWithCache()
+		o.filters = NewSublistNoCache()
 		for _, filter := range o.cfg.FilterSubjects {
 			o.filters.Insert(&subscription{subject: []byte(filter)})
 		}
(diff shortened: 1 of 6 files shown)


neilalexander (approved):
LGTM!

server/consumer.go
```diff
@@ -4730,37 +4730,15 @@ func (o *consumer) calculateNumPending() (npc, npf uint64) {
...
-	for _, filter := range o.subjf {
-		lnpc, lnpf := o.mset.store.NumPending(o.sseq, filter.subject, false)
-		npc += lnpc
-		if lnpf > npf {
-			npf = lnpf // Always last
-		}
+	if filters != nil {
+		return o.mset.store.NumPendingMulti(o.sseq, filters, isLastPerSubject)
+	} else if subjf != nil {
+		filter := subjf[0].subject
```
bruth, line 4738: Is this guaranteed to be non-empty even if `subjf` is a non-nil slice? `else if len(subjf) > 0` would be a safer assertion. I ask because the previous code looped over `o.subjf` so an empty slice would have been a no-op.

derekcollison: Yes the code that sets things up guarantees it and we use same logic for LoadNextMsg().

Could make it safer though.

derekcollison: Fixed in both spots to check `len(subjf) > 0`

neilalexander (approved):
LGTM

PR #6091, opened 2024-11-08
WebSocket: Ensure full writes to compressor

neilalexander:
The `cp.Write()` call may have been partial and we would not have noticed.

### DIFF
--- a/server/websocket.go
+++ b/server/websocket.go
@@ -1367,7 +1367,19 @@ func (c *client) wsCollapsePtoNB() (net.Buffers, int64) {
 		}
 		var csz int
 		for _, b := range nb {
-			cp.Write(b)
+			for len(b) > 0 {
+				n, err := cp.Write(b)
+				if err != nil {
+					if err == io.EOF {
+						break
+					}
+					c.Errorf("Error during compression: %v", err)
+					c.markConnAsClosed(WriteError)
+					nbPoolPut(b)
+					return nil, 0
+				}
+				b = b[n:]
+			}
 			nbPoolPut(b) // No longer needed as contents written to compressor.
 		}
 		if err := cp.Flush(); err != nil {


server/websocket.go
```diff
@@ -1367,7 +1367,12 @@ func (c *client) wsCollapsePtoNB() (net.Buffers, int64) {
 		}
 		var csz int
 		for _, b := range nb {
-			cp.Write(b)
+			wb := bytes.NewBuffer(b)
```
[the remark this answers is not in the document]
neilalexander: It doesn't duplicate `b`, just allocates the state for the `bytes.Buffer`. Either way I've refactored it into a normal loop tracking the written bytes.

server/websocket.go
```diff
@@ -1367,7 +1367,18 @@ func (c *client) wsCollapsePtoNB() (net.Buffers, int64) {
 		}
 		var csz int
 		for _, b := range nb {
-			cp.Write(b)
+			for len(b) > 0 {
+				n, err := cp.Write(b)
+				if err != nil {
```
derekcollison, line 1372: Can this be a short write error as well?

server/websocket.go
```diff
@@ -1367,7 +1367,18 @@ func (c *client) wsCollapsePtoNB() (net.Buffers, int64) {
...
+					if err == io.EOF {
+						break
+					}
+					c.Errorf("Error during compression: %v", err)
+					c.markConnAsClosed(WriteError)
```
derekcollison, line 1377: Should we recycle b?

neilalexander: Done.

server/websocket.go
neilalexander: Don't believe so, no, doesn't look like the compressor library ever returns that error code.

derekcollison (approved):
LGTM

PR #6106, opened 2024-11-11
Add `node10` to stree

neilalexander:
Although we probably don't want too many different node sizes here, the `node10` case is particularly interesting because it perfectly fits the full 0-9 numeric range without wasting bytes. In fact it saves 96 bytes (208 bytes instead of 304) compared to using `node16` for the same purpose.

This means memory savings for tracking subjects which are either mostly numerical throughout, or have tokens that are primarily numerical.

For a subject space that is mostly numerical, this can be several GBs less at the half-billion subjects mark and the saving can grow above that.


derekcollison (approved):
LGTM

PR #6109, opened 2024-11-11
[IMPROVED] Pre-acks handling, detecting ack for out of bounds sequence.

derekcollison:
Detect if we receive an ack past our last stream sequence.

We also no longer register pre-acks when we detect this from a consumer snapshot since we properly handle this now and this could lead to excessive memory usage.

### DIFF
--- a/server/consumer.go
+++ b/server/consumer.go
@@ -3062,16 +3062,28 @@ func (o *consumer) processAckMsg(sseq, dseq, dc uint64, reply string, doSample b
 		return false
 	}
 
-	// Check if this ack is above the current pointer to our next to deliver.
-	// This could happen on a cooperative takeover with high speed deliveries.
-	if sseq >= o.sseq {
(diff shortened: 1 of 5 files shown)


derekcollison:
Question for the team, currently I do return if the ack is waiting for a reply. Currently we do not respond with any errors in 2.10.x so that will need to wait for 2.11 (@jnmoyne) but should we not return and let the ack timeout instead?

@ripienaar interested in your thoughts here.

neilalexander (approved):
LGTM

derekcollison:
ok after thinking I believe we should timeout on an AckSync and err when we hit 2.11.

I force pushed the change to the code and the test.

jnmoyne:
> ok after thinking I believe we should timeout on an AckSync and err when we hit 2.11.

I agree that it should error out in 2.11 (e.g. especially if at the same time we start erroring out also in case it's the message is already ack'd).

PR #6112, opened 2024-11-12
Use interest-based intersection for `NumPendingMulti`

neilalexander:
This optimises #6089 by ensuring that we don't over-walk either the sublist or the subject tree, instead only matching the subject tree on subjects for which the sublist expresses interest.

On my machine, in the `JetStreamConsumeWithFilters` benchmark with `-benchtime 1000x`, this reduces the CPU time taken by `NumPendingMulti` by 19.41 seconds, reducing it to only 0.06 seconds.

### DIFF
--- a/server/sublist.go
+++ b/server/sublist.go
@@ -20,6 +20,8 @@ import (
 	"sync"
 	"sync/atomic"
 	"unicode/utf8"
+
+	"github.com/nats-io/nats-server/v2/server/stree"
 )
 
 // Sublist is a routing mechanism to handle subject distribution and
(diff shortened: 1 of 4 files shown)


server/sublist.go
```diff
@@ -1735,3 +1737,44 @@ func getAllNodes(l *level, results *SublistResult) {
 		getAllNodes(n.next, results)
 	}
 }
+
+// interestBasedIntersection will match all items in the given subject tree that
+// have interest expressed in the given sublist. The callback will only be called
+// once for each subject, regardless of overlapping subscriptions in the sublist.
+func interestBasedIntersection[T any](st *stree.SubjectTree[T], sl *Sublist, cb func(subj []byte, entry *T)) {
```
derekcollison, line 1744: Should this be a function on a *sublist? Or that not possible with generics?

If not make it public? IntersectStree()?

server/sublist.go
```diff
@@ -1735,3 +1737,44 @@ func getAllNodes(l *level, results *SublistResult) {
...
+	var _tokens [255]byte
+	_interestBasedIntersection(st, sl.root, _tokens[:0], cb)
+}
+
+func _interestBasedIntersection[T any](st *stree.SubjectTree[T], r *level, subj []byte, cb func(subj []byte, entry *T)) {
```
derekcollison, line 1749: Just make intersectStree?

derekcollison (commented):
Nice work! Some comments on naming.

server/sublist.go
neilalexander: Not possible with generics, but yep, happy to rename it to that.

derekcollison (approved):
LGTM

PR #6115, opened 2024-11-12
fix TestJetStreamJWTMove flapper

aricart:
test(flapper): fix TestJetStreamJWTMove - by removing a check on stream update for the configuration, likely the server was able to accomplish the move, right after, the test will assert that C2 is the stream was placed in the correct cluster


derekcollison (approved):
LGTM

PR #6120, opened 2024-11-13
De-flake `TestFileStoreCorruptPSIMOnDisk`

neilalexander:
Using `crand` was unpredictable because it might produce bytes that are a valid subject and therefore we wouldn't notice the problem in `recoverFullState`. Dropping in a UTF-8 rune error trips the detection always and is therefore more reliable.


derekcollison (approved):
LGTM

PR #6126, opened 2024-11-13
[FIXED] LeafNode: queue distribution with daisy-chain and gateway

### LINKED ISSUE #6125: Queue sub may not receive message in leafnode daisy-chain and gateway setups
### Observed behavior

In the following setup:
```
Cluster A --leaf connection--> Cluster B --leaf connection--> Cluster C --gateway connection--> Cluster D
```
If there are queue subscribers on "foo" for queue "queue" attached to "Cluster A" and "Cluster D", and a message is published on "Cluster B", the message will be delivered to one of the queue subscriber on "Cluster A" but sometimes, none of the queue subscriber on "Cluster A" will receive it, and "Cluster D" queue subscribers never receive any message.

### Expected behavior

Since there is a queue subscription interest in both "Cluster A" and "Cluster D", message should either go to "A" or "D" but not be dropped.

### Server and client version

Latest versions.

### Host environment

n/a

### Steps to reproduce

Create the setup described above and publish message and notice that sometimes a message does not make it to the queue subs.

kozlovic:
In complex setup, a message produced from a cluster that had queue interest from leafnodes (either hub or spoke) would sometimes not deliver a message if the interest was a leafnode that had the interest on behalf of a gateway.

In the setup described in the issue this PR fixes, "Cluster B" may have picked "Cluster C", but that cluster does not have local queue interest, only the leafnode interest from "Cluster B", and would pick a LEAF connection to this cluster, but then suppress the message since it came from "B" so "C" cannot send it back there.

But picking a queue sub for "B" in "C" would then prevent the message to be delivered to the gateway "D".

Resolves #6125

### DIFF
--- a/server/client.go
+++ b/server/client.go
@@ -4752,6 +4752,21 @@ func (c *client) processMsgResults(acc *Account, r *SublistResult, msg, deliver,
 	// Declared here because of goto.
 	var queues [][]byte
 
+	var leafOrigin string
+	switch c.kind {
+	case ROUTER:
+		if len(c.pa.origin) > 0 {
+			// Picture a message sent from a leafnode to a server that then routes
+			// this message: CluserA -leaf-> HUB1 -route-> HUB2
+			// Here we are in HUB2, so c.kind is a ROUTER, but the message will
+			// contain a c.pa.origin set to "ClusterA" to indicate that this message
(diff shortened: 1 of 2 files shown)


derekcollison (approved):
LGTM

PR #6127, opened 2024-11-14
[IMPROVED] Header construction during republish with existing headers.…

derekcollison:
When republishing use a more optimal version to avoid multiple copying of the header for each additional field addition.


neilalexander (approved):
LGTM

PR #6128, opened 2024-11-14
Tweak initial filestore sync timer

neilalexander:
Previously this was always firing at the exact interval. Then #6041 changed this so that the first firing was anytime between 1s and the interval to spread out different stores. This tweaks it once more to anytime between interval/2 and interval. It will still spread things out a bit but hit the benchmarks a bit less badly.

Also go back to `math/rand` instead of `math/rand/v2` due to availability in Go versions that we still care about for 2.10.x.

### DIFF
--- a/server/filestore.go
+++ b/server/filestore.go
@@ -29,7 +29,7 @@ import (
 	"io"
 	"io/fs"
 	"math"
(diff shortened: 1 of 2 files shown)


derekcollison (approved):
LGTM

PR #6132, opened 2024-11-14
NRG: Reduce sendq allocations

neilalexander:
This reduces allocations in the NRG `sendRPC` and `sendReply` functions by promoting buffer reuse.

### DIFF
--- a/server/sendq.go
+++ b/server/sendq.go
@@ -78,12 +78,14 @@ func (sq *sendq) internalLoop() {
 				if len(pm.hdr) > 0 {
 					c.pa.hdr = len(pm.hdr)
 					c.pa.hdb = append(hdb[:0], strconv.Itoa(c.pa.hdr)...)
-					msg = append(pm.hdr, pm.msg...)
+					msg = append(msg, pm.hdr...)
+					msg = append(msg, pm.msg...)
 					msg = append(msg, _CRLF_...)
 				} else {
 					c.pa.hdr = -1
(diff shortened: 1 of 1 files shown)


server/sendq.go
```diff
@@ -74,16 +76,18 @@ func (sq *sendq) internalLoop() {
 				} else {
 					c.pa.reply = nil
 				}
-				var msg []byte
+				msg = msg[:0]
```
derekcollison, line 79: Since we know how much we will need here, if len(pm.hdr) + len(pm.msg) > cap(_msg) we should allocate and avoid second copy?

server/sendq.go
```diff
@@ -57,6 +57,8 @@ func (sq *sendq) internalLoop() {
 		rply [256]byte
 		szb  [10]byte
 		hdb  [10]byte
+		_msg [1024]byte
```
derekcollison, line 60: Can we increase and decrease more allocations? Say 4k?

server/sendq.go
derekcollison: The variable msg is new each signal, so it would only last for a single pop, and we would still have to unnecessarily copy twice. Also think of a case where each msg is slightly bigger.. Then every message has 2 copies to expand.

derekcollison: ok so I see you changed its scope. Good deal.

derekcollison (approved):
LGTM

PR #6135, opened 2024-11-15
Update to Go 1.23.3/1.22.9, use Go 1.23 for goreleaser

neilalexander:
This updates the patch versions. It also updates the version that we use for goreleaser — all releases are currently built using Go 1.22, but there isn't really a good reason why we shouldn't build them with Go 1.23 instead, since there will be new compiler optimisations we can take advantage of.


wallyqs (approved):
LGTM

derekcollison (approved):
LGTM

PR #6152, opened 2024-11-20
Tests: Fix `require_NotNil`

neilalexander:
Previous behaviour of checking interface is `nil` didn't do the right thing as a `nil` parameter could still be wrapped in an interface descriptor that itself would be non-`nil`, so use reflection instead.

### DIFF
--- a/server/norace_test.go
+++ b/server/norace_test.go
@@ -8517,13 +8517,16 @@ func TestNoRaceJetStreamClusterDifferentRTTInterestBasedStreamPreAck(t *testing.
 	c := &cluster{t: t, servers: make([]*Server, 3), opts: make([]*Options, 3), name: "F3"}
 
 	// S1
-	conf := fmt.Sprintf(tmpl, "S1", t.TempDir(), 14622, "route://127.0.0.1:15622, route://127.0.0.1:16622")
+	// The route connection to S2 must be through a slow proxy
+	np12 := createNetProxy(10*time.Millisecond, 1024*1024*1024, 1024*1024*1024, "route://127.0.0.1:15622", true)
+	routes := fmt.Sprintf("%s, route://127.0.0.1:16622", np12.routeURL())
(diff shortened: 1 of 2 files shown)


server/norace_test.go
```diff
@@ -8601,12 +8601,6 @@ func TestNoRaceJetStreamClusterDifferentRTTInterestBasedStreamPreAck(t *testing.
 	mset, err := slow.GlobalAccount().lookupStream("EVENTS")
 	require_NoError(t, err)
 
-	// Make sure preAck is non-nil, so we know the logic has kicked in.
```
derekcollison, line 8604: Why did we remove this check?

neilalexander: It has never done the right thing, even going back to the commit where the test was added a year ago. It's just that the bug in `require_NotNil` hid the fact it was `nil` at this point all along.

derekcollison: So prechecks is always nil here? In the test?

neilalexander: Yep. Notably nowhere else in the codebase looks at whether it's `nil`, they just do `len(...)` on it, which is what the test then goes on to do in `checkFor`.

derekcollison: Then we need to slow down the server more since this is supposed to be testing that preacks work etc. We should not simply remove the check that it is not nil.

neilalexander: Turns out this test was wrong in all sorts of ways:

1. The routes between S1 and S2 were only slow in one direction, not both;
2. The stream was listening on one subject but the publisher was publishing on a totally different subject, therefore the consumer never received any messages to ack;
3. The final `checkFor` loop didn't wait for all of the messages to have even arrived before judging whether preacks had happened or not.

I've now fixed it I think.

derekcollison (approved):
LGTM

PR #6154, opened 2024-11-20
[FIXED] check of the backoff and max delivery configs

Luisgustavom1:
Match the check of the backoff and maxDeliver configs with the official documentation -> https://docs.nats.io/nats-concepts/jetstream/consumers#:~:text=The%20sequence%20length,MaxDelivery

Documentation, tell that `Backoff ` config **must** be **less than or equal** to `MaxDelivery`. But we were checking the opposite 

Signed-off-by: Luisao <luisgustavomacedo13@gmail.com>

### DIFF
--- a/server/consumer.go
+++ b/server/consumer.go
@@ -623,7 +623,7 @@ func checkConsumerCfg(
 	}
 
 	// Check if we have a BackOff defined that MaxDeliver is within range etc.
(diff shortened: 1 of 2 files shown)


neilalexander (approved):
Thanks for the contribution, LGTM!

neilalexander:
Is `[WIP]` because you are planning future changes to this PR?

PR #6160, opened 2024-11-21
De-flake `TestJetStreamSuperClusterTagInducedMoveCancel`

mtmk:
Request to cancel the move sometimes lands as the stream is being moved
and at that point there is not leader hence no response is sent back.
With this change we're retrying the request and falling back to the case
later in the test where we skip as stream move is not in progress.

Signed-off-by: Ziya Suzen <ziya@synadia.com>

### DIFF
--- a/server/jetstream_super_cluster_test.go
+++ b/server/jetstream_super_cluster_test.go
@@ -2706,10 +2706,17 @@ func TestJetStreamSuperClusterTagInducedMoveCancel(t *testing.T) {
 	_, err = js.UpdateStream(cfg)
 	require_NoError(t, err)
 
-	rmsg, err := ncsys.Request(fmt.Sprintf(JSApiServerStreamCancelMoveT, "$G", "TEST"), nil, 5*time.Second)
-	require_NoError(t, err)
+	// Retry in case stream is still being created and without a leader we won't receive a response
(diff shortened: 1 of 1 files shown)


server/jetstream_super_cluster_test.go
```diff
@@ -2706,10 +2706,17 @@ func TestJetStreamSuperClusterTagInducedMoveCancel(t *testing.T) {
 	_, err = js.UpdateStream(cfg)
 	require_NoError(t, err)
 
-	rmsg, err := ncsys.Request(fmt.Sprintf(JSApiServerStreamCancelMoveT, "$G", "TEST"), nil, 5*time.Second)
-	require_NoError(t, err)
+	// Retry in case stream is still being created and without a leader we won't receive a response
 	var cancelResp JSApiStreamUpdateResponse
-	require_NoError(t, json.Unmarshal(rmsg.Data, &cancelResp))
+	checkFor(t, 5*time.Second, 0, func() error {
```
neilalexander, line 2711: Is the test reliable if you have a short interval (i.e. `100*time.Millisecond`) instead of `0`?

mtmk: it should be. let me try

PR #6161, opened 2024-11-22
[FIXED] LeafNode: propagation of (no)interest issues

### LINKED ISSUE #5972: Queue Groups on leaf clusters not balancing correctly when messages are routed in from hub cluster [v2.10.21]

### LINKED ISSUE #6148: LeafNode: (no)interest propagation issues
### Observed behavior

Probably more cases than described here, but one would be that in a cluster hub/spoke setup, it was possible for interest registered in the leaf cluster to not be fully unregistered when the subscriptions were going away (depending if subscription was removed one by one or if a connection having multiple was closed), and more over, the hub may have incorrectly sent to the leaf cluster an interest on the subject, making publish from the leaf go to the hub where no interest was really present. This would happen if one of the hub server was restarted while subscription was still active, but then the subscriptions were removed.

### Expected behavior

Subscription interest properly handled when subscriptions go away, either individually or as a whole when their connection is closed, or when servers are restarted.

### Server and client version

Latest

### Host environment

N/A

### Steps to reproduce

```
	// B1 <--- route ---> B2
	//  |                 |
	// Leaf              Leaf
	//  |                 |
	// A1 <--- route ---> A2
```
Have queue subs connecting to A2, restart B1, unsubscribe one of the qsub from A2. When that happens, B1 may have sent a subscription interest to A1 (A1's Leafz would show a subscription interest). Removing the remaining subscriptions would not remove the subscription interest from A1->B1.
Test can be repeated where a single connection has all 3 subs and connection is closed.

kozlovic:
There were multiple issues, but basically the fact that we would not store the routed subscriptions with the origin of the LEAF they came from made the server unable to differentiate those compared to "local" routed subscriptions, which in some cases (like a server restart and the resend of subscriptions) could lead to servers sending incorrectly subscription interest to leaf connections.

We are now storing the subscriptions with a sub type indicator and the origin (for leaf subscriptions) as part of the key. This allows to differentiate "regular" routed subs versus the ones on behalf of a leafnode.
An INFO boolean is added `LNOCU` to indicate support for origin in the `LS-` protocol, which is required to properly handle the removal. Therefore, if a route does not have `LNOCU`, the server will behave like an old server, and store with the key that does not contain the origin, so that it can be removed when getting an LS- without the origin. Note that in the case of a mix of servers in the same cluster, some of the issues this PR is trying to fix will be present (since the server will basically behave like a server without the fix).

Having a different routed subs for leaf connections allow to revisit the fix #5982 that was done for issue #5972, which was about a more fair queue distribution to a cluster of leaf connections. That fix actually introduced a change in that we always wanted to favor queue subscriptions of the cluster where the message is produced, which that fix possibly changed. With this current PR, the server can now know if a remote queue sub is for a "local" queue sub there or on behalf of a leaf and therefore will not favor that route compared to a leaf subscription that it may have directly attached.

Resolves #5972
Resolves #6148

### DIFF
--- a/server/leafnode.go
+++ b/server/leafnode.go
@@ -2363,10 +2363,10 @@ const (
 // Helper function to build the key that prevents collisions between normal
 // routed subscriptions and routed subscriptions on behalf of a leafnode.
 // Keys will look like this:
-// "R foo" 			-> plain routed sub on "foo"
-// "R foo bar"		-> queue routed sub on "foo", queue "bar"
-// "L foo bar"		-> plain routed leaf sub on "foo", leaf "bar"
-// "L foo bar baz"	-> queue routed sub on "foo", queue "bar", leaf "baz"
+// "R foo"          -> plain routed sub on "foo"
+// "R foo bar"      -> queue routed sub on "foo", queue "bar"
+// "L foo bar"      -> plain routed leaf sub on "foo", leaf "bar"
+// "L foo bar baz"  -> queue routed sub on "foo", queue "bar", leaf "baz"
 func keyFromSubWithOrigin(sub *subscription) string {
 	var sb strings.Builder
 	sb.Grow(2 + len(sub.origin) + 1 + len(sub.subject) + 1 + len(sub.queue))
--- a/server/leafnode.go
+++ b/server/leafnode.go
@@ -2353,6 +2353,42 @@ func keyFromSub(sub *subscription) string {
 	return sb.String()
 }
 
+const (
+	keyRoutedSub         = "R"
+	keyRoutedSubByte     = 'R'
+	keyRoutedLeafSub     = "L"
+	keyRoutedLeafSubByte = 'L'
+)
+
+// Helper function to build the key that prevents collisions between normal
+// routed subscriptions and routed subscriptions on behalf of a leafnode.
+// Keys will look like this:
+// "R foo" 			-> plain routed sub on "foo"
+// "R foo bar"		-> queue routed sub on "foo", queue "bar"
+// "L foo bar"		-> plain routed leaf sub on "foo", leaf "bar"
+// "L foo bar baz"	-> queue routed sub on "foo", queue "bar", leaf "baz"
+func keyFromSubWithOrigin(sub *subscription) string {
+	var sb strings.Builder
+	sb.Grow(2 + len(sub.origin) + 1 + len(sub.subject) + 1 + len(sub.queue))
+	leaf := len(sub.origin) > 0
+	if leaf {
+		sb.WriteByte(keyRoutedLeafSubByte)
+	} else {
+		sb.WriteByte(keyRoutedSubByte)
+	}
+	sb.WriteByte(' ')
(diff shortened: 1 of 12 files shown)


server/route.go
```diff
@@ -2333,8 +2479,10 @@ func (s *Server) updateRouteSubscriptionMap(acc *Account, sub *subscription, del
 		return
 	}
 
-	// Create the fast key which will use the subject or 'subject<spc>queue' for queue subscribers.
-	key := keyFromSub(sub)
+	// Create the fast key which will use the subject or '[origin]<spc>subject<spc>queue'
```
kozlovic, line 2482: This comment is wrong. I did change the way this key is constructed (see keyFromSubWithOrigin). Will update tomorrow.

server/leafnode.go
```diff
@@ -2353,6 +2353,42 @@ func keyFromSub(sub *subscription) string {
...
+	keyRoutedSub         = "R"
+	keyRoutedSubByte     = 'R'
+	keyRoutedLeafSub     = "L"
+	keyRoutedLeafSubByte = 'L'
+)
+
+// Helper function to build the key that prevents collisions between normal
+// routed subscriptions and routed subscriptions on behalf of a leafnode.
+// Keys will look like this:
+// "R foo" 			-> plain routed sub on "foo"
```
kozlovic, line 2366: Will fix that tomorrow.

derekcollison (approved):
LGTM - Nice work!

PR #6164, opened 2024-11-22
Facilitate flake tests reproduction with Antithesis

mprimi:
1. Introduce dependency on antithesis-sdk-go (default NOOP variant)
2. Introduce assertions utilities suitable for use in tests
3. Instrument frequently used test utilities with the new assertions

N.B. unless the `enable_antithesis_sdk` tag is specified at build time, both the SDK and the assertions wrappers are NOOP, and should have no effect.

This will also facilitate other Antithesis one off experiments since developers no longer have to manually add the SDK to their branch.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -5,6 +5,7 @@ go 1.22
 toolchain go1.22.8
 
 require (
+	github.com/antithesishq/antithesis-sdk-go v0.4.3-default-no-op
 	github.com/google/go-tpm v0.9.3
 	github.com/klauspost/compress v1.17.11
 	github.com/minio/highwayhash v1.0.3
--- a/server/jetstream_helpers_test.go
+++ b/server/jetstream_helpers_test.go
@@ -35,6 +35,8 @@ import (
 
 	"github.com/nats-io/nats.go"
 	"golang.org/x/time/rate"
+
+	"github.com/nats-io/nats-server/v2/internal/antithesis"
 )
 
 // Support functions
@@ -104,6 +106,10 @@ func (sc *supercluster) waitOnStreamLeader(account, stream string) {
 		}
 		time.Sleep(100 * time.Millisecond)
 	}
+	antithesis.AssertUnreachable(sc.t, "Timeout in supercluster.waitOnStreamLeader", map[string]any{
+		"account": account,
+		"stream":  stream,
+	})
 	sc.t.Fatalf("Expected a stream leader for %q %q, got none", account, stream)
 }
 
@@ -556,6 +562,7 @@ func (sc *supercluster) waitOnLeader() {
 		}
 		time.Sleep(25 * time.Millisecond)
 	}
+	antithesis.AssertUnreachable(sc.t, "Timeout in supercluster.waitOnStreamLeader", nil)
 	sc.t.Fatalf("Expected a cluster leader, got none")
 }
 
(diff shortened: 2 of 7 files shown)


neilalexander (commented):
I'm a bit confused about `internal/antithesis/fallback_sdk.go` here. Antithesis [already produce `default-no-op` builds](https://github.com/antithesishq/antithesis-sdk-go/tags) of their SDK, which contain all of the functionality, but only when built with the `enable_antithesis_sdk` tag. Wouldn't that be what we want?

mprimi:
> Antithesis [already produce `default-no-op` builds](https://github.com/antithesishq/antithesis-sdk-go/tags) of their SDK

I'll be honest, i didn't notice this release (it's not documented and wasn't brought up in discussion).
I also assumed (wrongly?) you'd not be comfortable bringing in the full SDK.

But if you are, then I agree using the official client is a great option.

I just launched a side-by-side comparison for the test flakes from last week, we should see the official client find the same violated assertions.

Assuming the two are equivalent in terms of results for that experiment, there's still a few minor differences and tradeoffs worth considering.

This `fallback` implementation gives us more control:
 - Add additional behavior (e.g. stack trace on assert false is useful, even outside Antithesis), and more
 - Control grouping and de-duplication by test (i.e. 2 distinct tests failing on the same `require_*` or `waitFor*` will produce 2 example traces. vs. the `official` client will produce a single trace/example per assertion violation)
 - Control over panic or continue execution on assert violation
 - No surprise breakage when upgrading dependencies

Of course there are upsides of using the `official` library:
 - Less code to own/maintain, simpler upgrades
 - Devs no longer need to `go get antithesis` in their branch when running one-off experiments

One more important distinction:

The `fallback` impl. is specifically for tests, it only exposes `Assert(condition)` and `AssertUnreachable`. Both take a `*testing.T` argument which means we won't end up using it by mistake in server code.

The `official` client is generic and has a much richer API suitable for the entire project, not just tests.
It exposes things like `Reachable`, `Sometimes`, which do not work properly inside a test context**. Devs may use these API count on the results without realizing they do not actually work.

So the richer API is a double-edged sword.


** Antithesis instrumentor skips test code => Assertions in tests code are not catalogued => Antithesis does not know of these assertion ahead of time =>  Antithesis cannot know it failed to reach a `Reachable` statement => There will be no failed `Reachable` property in the report. This can be misleading, as a failed property ins not reported.

server/jetstream_helpers_test.go
```diff
@@ -1307,6 +1322,9 @@ func (c *cluster) waitOnPeerCount(n int) {
 			leader = c.leader()
 		}
 	}
+	antithesis.AssertUnreachable(c.t, "Timeout in cluster::waitOnPeerCount", map[string]any{
```
neilalexander, line 1325: It's a minor nit but `::` isn't Go-standard notation, let's just stick with `.`.

neilalexander (approved):
Otherwise LGTM, thanks!

neilalexander (approved):
LGTM

go.mod
```diff
@@ -5,6 +5,7 @@ go 1.22
 toolchain go1.22.8
 
 require (
+	github.com/antithesishq/antithesis-sdk-go v0.4.2-default-no-op
```
alexbozhenko, line 8: btw, they just published a newer version: https://pkg.go.dev/github.com/antithesishq/antithesis-sdk-go@v0.4.3-default-no-op

mprimi: Thank you. Updated.

go.mod
```diff
@@ -5,6 +5,7 @@ go 1.22
 toolchain go1.22.8
 
 require (
+	github.com/antithesishq/antithesis-sdk-go v0.4.3-default-no-op
```
derekcollison, line 8: So anyone who tries to install the server this will also be required?

neilalexander: It's a no-op version, so the entire SDK is stubbed out with empty functions which get inlined away unless the `enable_antithesis_sdk` build tag is passed in, which it only is under Antithesis.

PR #6165, opened 2024-11-24
[IMPROVED] Encoding for large meta snapshots

derekcollison:
Improve very large meta snapshots by moving to goccy json pkg and utilize s2.Encode vs EncodeBetter which is faster and compresses better on on our meta snapshots.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -5,6 +5,7 @@ go 1.22
 toolchain go1.22.8
 
 require (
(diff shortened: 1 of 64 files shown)


derekcollison:
FYI plan to add test to remove us snapshotting so much on consumer deletes as well here so feel free to hold off, but looks very promising and sans the one test that I had to fix passed the first time all others.

PR #6174, opened 2024-11-25
Refactor duplicate `Nats-Msg-Id` tracking

neilalexander:
The current duplicate tracking has a couple of problems, namely that the `ddmap` capacity is potentially never reclaimed if it grows suddenly, and secondly that we have to constantly monitor `ddarr`'s capacity and copy it to stop the array from growing infinitely too.

This PR makes the following changes:

- `ddmap` is now using the stree, which means there is efficient in-place deduplication of message IDs and no infinitely growing backing capacity;
- `ddarr` and `ddindex` are now gone, replaced with linked-list references and `ddnext` and `ddlast` tracking, so we also don't need to watch the capacity and reallocate this either;
- `ddtmr` is now only scheduled when there is actually something to do.

Also added a unit test to prove the behaviour as well as the linked-list stitching.

### DIFF
--- a/server/stream.go
+++ b/server/stream.go
@@ -33,6 +33,7 @@ import (
 
 	"github.com/goccy/go-json"
 	"github.com/klauspost/compress/s2"
+	"github.com/nats-io/nats-server/v2/server/stree"
 	"github.com/nats-io/nuid"
 )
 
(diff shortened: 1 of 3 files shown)


derekcollison (commented):
LGTM

server/stream.go
```diff
@@ -4121,15 +4126,48 @@ func (mset *stream) storeMsgId(dde *ddentry) {
...
-		mset.ddmap = make(map[string]*ddentry)
+		mset.ddmap = stree.NewSubjectTree[*ddentry]()
 	}
-	mset.ddmap[dde.id] = dde
-	mset.ddarr = append(mset.ddarr, dde)
-	if mset.ddtmr == nil {
-		mset.ddtmr = time.AfterFunc(mset.cfg.Duplicates, mset.purgeMsgIds)
+
+	// Clean up the previous entry we just replaced. Stitch the gap.
+	if o, updated := mset.ddmap.Insert(stringToBytes(dde.id), dde); updated && o != nil {
```
derekcollison, line 4136: If updated is true that means we had a collision yes? Should this not happen?

derekcollison: Meaning we should only put new items in here and the msgID should have been checked prior?

neilalexander: I've added a comment to the code to this effect. Ordinarily yes, `checkMsgId()` has been called by now, but when we have to call `rebuildDedupe()`, we will replay all of the stream messages in the deduplication window and they may have duplicate message IDs. Therefore we only want to keep the last `ddentry` for each ID.

neilalexander: I think the only way it should normally happen is if the duplicate window was changed before calling `rebuildDedupe()`.

derekcollison:
Want me to take a look?

neilalexander:
Yes please. I did take a look at `container/heap` as an option but it still requires a backing array and has the same problem that the current `ddarr` has in terms of backing capacity. There's also `container/list` but it isn't generic (although we could copy it in from the stdlib and _make_ it generic if you think that's a better path).

derekcollison (approved):
LGTM

PR #6176, opened 2024-11-25
[IMPROVED] Bypass meta-layer's API queue for non-existent consumers.

derekcollison:
Processing a missing consumer was a heavier operation on the meta-layer and could lead to API queue buildup.
This processes that condition at point of entry to the system vs at the meta-layer.
If the consumer assignment does exist we process as normal for now.

Also moved jetstream clustered state and system account to atomics to avoid lock contention.
Add in lookup for both stream and consumer assignments at the same time.

### DIFF
--- a/server/jetstream_cluster.go
+++ b/server/jetstream_cluster.go
@@ -4728,6 +4728,7 @@ func (js *jetStream) consumerAssignment(account, stream, consumer string) *consu
 }
 
 // Return both the stream and consumer assignments.
+// Lock should be held.
 func (js *jetStream) assignments(account, stream, consumer string) (*streamAssignment, *consumerAssignment) {
 	if sa := js.streamAssignment(account, stream); sa != nil {
 		return sa, sa.consumers[consumer]
--- a/server/jetstream_cluster.go
+++ b/server/jetstream_cluster.go
@@ -224,11 +224,7 @@ func (s *Server) getJetStreamCluster() (*jetStream, *jetStreamCluster) {
 }
 
 func (s *Server) JetStreamIsClustered() bool {
-	js := s.getJetStream()
-	if js == nil {
-		return false
-	}
-	return js.isClustered()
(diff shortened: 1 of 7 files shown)


wallyqs (approved):
LGTM

server/jetstream_cluster.go
```diff
@@ -4731,6 +4727,14 @@ func (js *jetStream) consumerAssignment(account, stream, consumer string) *consu
 	return nil
 }
 
+// Return both the stream and consumer assignments.
```
neilalexander, line 4730: Maybe add to the comment that the JS lock must be held.

server/norace_test.go
```diff
@@ -11252,3 +11252,59 @@ func TestNoRaceJetStreamClusterLargeMetaSnapshotTiming(t *testing.T) {
...
+
+	// Check to make sure we never have any pending on the API queue.
+	go func() {
+		ml := c.leader()
+		for {
+			select {
+			case <-done:
+				return
+			case <-time.After(100 * time.Millisecond):
+				if qlen := ml.jsAPIRoutedReqs.len(); qlen > 0 {
```
neilalexander, line 11282: May not be an issue but might be a bit more robust if it's `ml.jsAPIRoutedReqs.len() + ml.jsAPIRoutedReqs.inProgress()` as then it'll capture if there's a consumer info pulled off the queue and in progress. (The API worker doesn't `recycle()` until the ICB has finished.)

neilalexander (approved):
Otherwise LGTM!

server/norace_test.go
derekcollison: ok will do. Noticed this though which felty weird.

`./norace_test.go:11282:13: invalid operation: ml.jsAPIRoutedReqs.len() + ml.jsAPIRoutedReqs.inProgress() (mismatched types int and int64)`

PR #6178, opened 2024-11-26
Add logging for large metalayer snapshots duration

wallyqs:
This adds extra WRN level logging to detect cases when there is a slowness in the metalayer snapshot duration, example line:
```
[WRN] Metalayer snapshot took 1.113s (streams: 326, consumers: 29382, marshal: 1.103s, uncompressed: 33268722, compressed: 2768064)
```


derekcollison (approved):
LGTM

PR #6183, opened 2024-11-27
Add test covering handling of delayed acks from recreated streams

wallyqs:
Adds explicit test showing behavior of consumer state when ack is received after a stream has been recreated.


derekcollison (commented):
This run for a long time or use lots of memory? Should it be in norace?

wallyqs:
This one takes around 6 - 7 seconds right now.

PR #6189, opened 2024-11-29
Reduce size of meta snapshots further

neilalexander:
We would not expect to respond to a consumer assignment after, e.g. a restart, and we don't store these for stream assignments already, therefore the `subject` and `reply` fields do not need to be encoded for consumer assignments either.

Similarly, the `stream` name doesn't need to be repeated for each consumer assignment either, as we can just pull that out of the parent stream assignment when decoding, so don't encode that either.

This should further reduce the size of meta snapshots and, with it, reduce their encoding/decoding time.


derekcollison (approved):
LGTM

I think would be good to periodically check the assignments, maybe during snapshot and if we know we have responded can drop it to the bare essentials in memory.

PR #6202, opened 2024-12-03
Re-add `"stream"` to consumer assignments

neilalexander:
This was changed in #6189 but we need it to be in the snapshot after all in order to be able to safely roll back to earlier versions.


derekcollison (approved):
LGTM

PR #6210, opened 2024-12-04
[IMPROVED] Move calculating source consumer sequences back to inline.

derekcollison:
We had added jitter to this however this happens when a stream becomes a leader, which will not be on startup and will already be spread out. However if the stream is an R1, this delays this processing to after the client port is open whereas inline processes it before.

The reason this could be important is that the stream read lock is held while we reverse scan the stream looking for sequences, if the client port is open this can stall apps.

This does not optimize the lookup process when there are a vast amount of interior deletes vs total msgs.

### DIFF
--- a/server/stream.go
+++ b/server/stream.go
@@ -307,7 +307,6 @@ type stream struct {
 	// Sources
 	sources              map[string]*sourceInfo
 	sourceSetupSchedules map[string]*time.Timer
-	sourcesConsumerSetup *time.Timer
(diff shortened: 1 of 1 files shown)


neilalexander (approved):
LGTM

derekcollison:
@wallyqs will take a look..

derekcollison:
Going to try different approach where inline is only for R1s.. Will close this one and open new.

PR #6218, opened 2024-12-04
[TEST ONLY] skip TestMQTTClusterConnectDisconnect...

levb:
The purpose of this test was to illustrate and verify https://github.com/nats-io/nats-server/pull/4734. Due to its timing sensitivity it has never been 100% reliable, prone to flaking on an occasional MQTT Connect failure (API timeout?). Skip for now, to reduce the flaky noise.

Signed-off-by: Lev Brouk <lev@synadia.com>


wallyqs (approved):
LGTM

PR #6219, opened 2024-12-05
[IMPROVED] Process sourced streams consumer sequences inline iff R1.

derekcollison:
If a stream that is sourcing is R1, setup the consumers inline on setLeader which is at server start on recovery of the stream.


wallyqs (approved):
LGTM

PR #6221, opened 2024-12-05
Increase disk I/O semaphore on systems with more CPU cores

neilalexander:
This should provide less contention on systems where we have plenty of headroom for more I/O-blocked threads.

### DIFF
--- a/server/filestore.go
+++ b/server/filestore.go
@@ -33,6 +33,7 @@ import (
 	"net"
 	"os"
 	"path/filepath"
+	"runtime"
 	"slices"
 	"sort"
 	"strings"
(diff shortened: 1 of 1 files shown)


derekcollison (commented):
I think we up base/minimum level to 8, unless 2x num cores is higher.

This is just a thread thing, not as much concurrency thing based on the way Go does blocking IO.

Also should make configurable IMO, even if we only really document it for 2.11.

neilalexander:
For now, given it's for a patch release, it may be best to stay at minimum 4 so we don't unexpectedly change the performance characteristics of lower-end/single-core/embedded platforms.

I'm all for increasing it further in 2.11 & making it configurable there.

derekcollison (approved):
LGTM

server/filestore.go
```diff
@@ -9504,8 +9505,9 @@ var dios chan struct{}
 // Used to setup our simplistic counting semaphore using buffered channels.
 // golang.org's semaphore seemed a bit heavy.
 func init() {
-	// Limit ourselves to a max of 4 blocking IO calls.
-	const nIO = 4
+	// Limit ourselves to a sensible number of blocking I/O calls. Range between
+	// 4-16 concurrent disk I/Os based on the number of CPU cores available.
+	nIO := min(16, max(4, runtime.GOMAXPROCS(-1)))
```
wallyqs, line 9510: I think capping at 16 sounds good, only thing is that gomaxprocs check happens here before auto maxprocs runs, so in container setups for this to apply it is needed to setup GOMAXPROCS explicitly instead

derekcollison: Excellent point @wallyqs

We need to make sure we have adjusted for running in a container. @neilalexander could you do followup PR?

PR #6223, opened 2024-12-05
Improve JSConsumerInvalidGroupNameErr's description

jnmoyne:
As the error is returned from a call to create/update a consumer "Valid name" was potentially confusing as the user may think the problem is with the consumer's name, not the priority group.


Jarema (approved):
LGTM!

derekcollison (approved):
LGTM

Issue #6211, opened 2024-12-04
Stream state resets if the Jetstream store directory goes into read only mode [v2.10.23]


sudojha:
### Observed behavior

Upon jetstream storage directory going into read only mode, the in memory stream state gets reset.

<img width="492" alt="Screenshot 2024-12-04 at 11 36 17 AM" src="https://github.com/user-attachments/assets/6b237cf4-4daa-42bb-88ec-adbb7f01117d">

It takes a restart to restore the stream state. 

Code inspection revealed the following:

For a stream with ttl enabled, [expireMsgs](https://github.com/nats-io/nats-server/blob/835840bb347bd0e846ffdb0af7d75152b60ad7df/server/filestore.go#L4538) is called periodically.

This internally calls [removeMsgBlock](https://github.com/nats-io/nats-server/blob/835840bb347bd0e846ffdb0af7d75152b60ad7df/server/filestore.go#L7002). Once the last message block is removed, a new tombstone block is created and needs to be assigned to lmb. Since the store directory is in read only mode the [error is returned before](https://github.com/nats-io/nats-server/blob/835840bb347bd0e846ffdb0af7d75152b60ad7df/server/filestore.go#L3008) the[ block gets assigned to lmb](https://github.com/nats-io/nats-server/blob/835840bb347bd0e846ffdb0af7d75152b60ad7df/server/filestore.go#L3027). 

This causes future publish to fail until file system recovers and the server is restarted.
Why does subsequent publish fail? 
Once a message block is to be removed from the file system, [dirtyCloseWithRemove](https://github.com/nats-io/nats-server/blob/835840bb347bd0e846ffdb0af7d75152b60ad7df/server/filestore.go#L7019) is called. This method clears the required state of the message block (particularly mfn which stores the qualified file name of the message block on disk).
The in memory representation of message block is stored in filestore under [blks](https://github.com/nats-io/nats-server/blob/835840bb347bd0e846ffdb0af7d75152b60ad7df/server/filestore.go#L178). The [lmb](https://github.com/nats-io/nats-server/blob/835840bb347bd0e846ffdb0af7d75152b60ad7df/server/filestore.go#L177) in filestore shares the same reference. Once the last block which is the lmb is removed, the state of lmb is cleared because file system is in read-only mode. Once this happens, future publish fails even if file system recovers with these errors:

[...]

PR #6237, opened 2024-12-10
Stop stream when stream state inconsistency is detected during flush

pranavmehta94:
This PR is intended to handle the stream state reset observed when filesystem enters read-only mode.(for reference: https://github.com/nats-io/nats-server/issues/6211)

This PR introduces the following:

1. A new server configuration `stream_stop_on_corruption` for jetstream to stop stream when filestore corruption is observed.
        a. When filesystem enters read-only mode, filestore detects stream state corruption which resets the in-memory state of stream to zero and future publish fails even if filesystem recovers until nats server restarts and restores stream state on server startup
        b. We have also seen scenarios when steam state was reset and publish resumed without needing a restart. This caused the consumers to stop consuming new msgs from stream since consumer stream seq was way ahead of stream seq
        c. Hence, to avoid these situations, we introduce a new server configuration to stop stream as soon as stream state corruption is detected during writing stream state onto disk.
2. Mark stream’s filestore is corrupt when the server is configured to stop stream on state corruption. 
        a. This change allows to ensure that inconsistent state is not flushed onto disk before stream is stopped

### DIFF
--- a/server/filestore.go
+++ b/server/filestore.go
@@ -67,6 +67,8 @@ type FileStoreConfig struct {
 	Cipher StoreCipher
 	// Compression is the algorithm to use when compressing.
 	Compression StoreCompression
+	// HandleStreamCorruption calls the stream corruption handler when internal stream state becomes inconsistent
+	HandleStreamCorruption bool
 
 	// Internal reference to our server.
 	srv *Server
@@ -172,6 +174,7 @@ type fileStore struct {
 	tombs       []uint64
 	ld          *LostStreamData
 	scb         StorageUpdateHandler
+	sscb        StreamStateCorruptionHandler
 	ageChk      *time.Timer
 	syncTmr     *time.Timer
 	cfg         FileStreamInfo
@@ -197,6 +200,7 @@ type fileStore struct {
 	fip         bool
 	receivedAny bool
 	firstMoved  bool
+	corrupt     bool
 }
 
 // Represents a message store block and its data.
@@ -3500,6 +3504,13 @@ func (fs *fileStore) RegisterStorageUpdates(cb StorageUpdateHandler) {
 	if cb != nil && bsz > 0 {
 		cb(0, int64(bsz), 0, _EMPTY_)
 	}
+}
+
+// RegisterStreamStateCorruptionCB registers a callback for stream state corruption
+func (fs *fileStore) RegisterStreamStateCorruptionCB(cb StreamStateCorruptionHandler) {
+	fs.mu.Lock()
+	fs.sscb = cb
+	fs.mu.Unlock()
 }
 
 // Helper to get hash key for specific message block.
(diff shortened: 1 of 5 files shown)


neilalexander (commented):
Understand that the current behaviour probably isn't ideal but we need more context on why the filesystem would suddenly become read-only before proceeding here. This seems like a serious environmental misconfiguration if the store directory suddenly changes underneath the server.

pranavmehta94:
@neilalexander We are running NATS in K8s cluster and CSI driver attaches a storage volume to the K8s pod over network(iSCSI). We see filesystem can go to read-only in our environment due to various issues. Hence, we want NATS to stop itself cleanly for now when such a situation arises.

skyrocknroll:
@neilalexander when a network outage heppens in any of network attached  storage we experience this kind of issue.

derekcollison:
We could potentially ensure that the server disables jetstream on a write error, but it should do that already IIRC..

pranav-mehta:
@derekcollison I think server disabling jetstream would impact streams which are configured with other stream stores as well(e.g. memstore). For filesystem issues, only streams configured with filestore are impacted.Also, based on our observation, stream enters an inconsistent state if the TTL expires all the messages in the stream before filesystem recovers. Hence, the implementation in this PR is trying limit the corruption handling only to the stream which detects state corruption. 
Also, when we hit the filesystem read-only issue, we did not observe server disabling jetstream.

pranavmehta94:
@derekcollison I think server disabling jetstream would impact streams which are configured with other stream stores as well(e.g. memstore). For filesystem issues, only streams configured with filestore are impacted.Also, based on our observation, stream enters an inconsistent state if the TTL expires all the messages in the stream before filesystem recovers. Hence, the implementation in this PR is trying limit the corruption handling only to the stream which detects state corruption.
Also, when we hit the filesystem read-only issue, we did not observe server disabling jetstream.

derekcollison:
There is only one store directory per server, so that error would have to disable all of jetstream.

shiv4289:
Hi @derekcollison Just want to be doubly sure on this. 

Stopping jetsream will also stop memory based streams. Is that fine?

Implementation-wise this will get tricky in how one can find file based without iterating all the streams which will be inefficient. So it might be okay to stop jetstream.

derekcollison:
Yes disabling jetstream is the right thing to do here. But we should do that when we encounter a disk error.

souravagrawal:
> Yes disabling jetstream is the right thing to do here. But we should do that when we encounter a disk error.

Hello @derekcollison, thank you for your feedback on the PR. @pranavmehta94 and I work in same team. 
Based on your suggestion, I have raised a new PR #6292.
Looking forward to your feedback.

neilalexander:
Replaced by #6292.

PR #6249, opened 2024-12-13
Adds a number of helper structs for z responses

ripienaar:
These are used by the cli and other tools to decode the responses from the z endpoints

### DIFF
--- a/server/events.go
+++ b/server/events.go
@@ -2004,13 +2004,78 @@ type ServerAPIResponse struct {
 	compress compressionType
 }
 
-// Specialized response types for unmarshalling.
+// Specialized response types for unmarshalling. These structures are not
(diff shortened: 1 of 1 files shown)


ripienaar:
These types doesn't exist anywhere its constructed from a `any` wrapped into all in the same structure

So to use them everyone either have to double json decode (This is how MH designed it and insisted its used) or write these structs themselves, so this just helps out.

derekcollison:
ok let's comment why they are here and not used in any of the server code base just for clairity.

ripienaar:
> ok let's comment why they are here and not used in any of the server code base just for clairity.

Have clarified the comment that was there already a bit further

derekcollison (approved):
LGTM

PR #6260, opened 2024-12-14
[FIXED] Websocket: handling of close message frame without status

### LINKED ISSUE #6259: Websocket: Server incorrectly sends close message status 1005 or 1006
### Observed behavior

(This issue was reported internally by @aricart)

When a client does not send any status/payload in the close message, the server detects correctly that situation but is sending back status 1005 (no status specified), which is wrong. It should simply not send any status/payload in its close message frame.
Also, server may send status 1006 (abnormal closure), which according to the specification, this status (like 1005) must not be set as a status code in a close control frame.

### Expected behavior

If a client sends a close message frame without status/payload, the server should echo the close message frame without status/payload, and the server should never send status 1006 (even if the server detects a client going away).

### Server and client version

Latest

### Host environment

N/A

### Steps to reproduce

Have a client sends a close message frame without status and observe that server is sending 1005. Also, in some situations, server would send 1006 (when it would detect a write or read error or when considering the connection stale).

kozlovic:
The server was incorrectly sending back a close message frame with status 1005 when not receiving a status in the close message frame from the client. This was wrong. This status was to be used internally to keep track that no status was received and therefore the server should not send a status back.

Also, the server was sending status 1006 in some error conditions, while the spec also states that this status must not be used to set the status, instead it was to internally keep track that there was an error condition.

Resolves #6259

### DIFF
--- a/server/websocket.go
+++ b/server/websocket.go
@@ -67,7 +67,6 @@ const (
 	wsCloseStatusProtocolError      = 1002
 	wsCloseStatusUnsupportedData    = 1003
 	wsCloseStatusNoStatusReceived   = 1005
-	wsCloseStatusAbnormalClosure    = 1006
 	wsCloseStatusInvalidPayloadData = 1007
 	wsCloseStatusPolicyViolation    = 1008
 	wsCloseStatusMessageTooBig      = 1009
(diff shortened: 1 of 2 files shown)


derekcollison (approved):
LGTM

aricart:
Thank you @kozlovic

PR #6272, opened 2024-12-17
Initial support for per-message TTLs

neilalexander:
This is an incomplete but somewhat-working implementations of per-message TTLs.

Notes:
- Adds `AllowMsgTTL` to stream config;
- Parses the `Nats-TTL` message header as a parsable duration, or if not, second precision;
- Bumps the `index.db` magic version so that we can add a new field for tracking how many TTL'd messages are in each message block (means that `index.db` needs to be rebuilt if downgrading);
- Adds a new `thw.db` to store the timed hash wheel state, and tries to rebuild it from a linear scan if it is missing or out-of-date.

Future work:
- Memory store support;
- Expires "never" including not aging out with MaxAge;
- Ability to update a TTL for a message;
- Decide what we want to do for enabling/disabling TTL support;
- Decide how mirroring/sourcing streams will handle TTL headers.

### DIFF
--- a/server/filestore.go
+++ b/server/filestore.go
@@ -45,6 +45,7 @@ import (
 	"github.com/minio/highwayhash"
 	"github.com/nats-io/nats-server/v2/server/avl"
 	"github.com/nats-io/nats-server/v2/server/stree"
+	"github.com/nats-io/nats-server/v2/server/thw"
 	"golang.org/x/crypto/chacha20"
 	"golang.org/x/crypto/chacha20poly1305"
 )
@@ -68,6 +69,8 @@ type FileStoreConfig struct {
 	Cipher StoreCipher
 	// Compression is the algorithm to use when compressing.
 	Compression StoreCompression
+	// EnforceTTLs decides whether or not to enforce per-message TTLs.
+	EnforceTTLs bool
 
 	// Internal reference to our server.
 	srv *Server
@@ -198,6 +201,8 @@ type fileStore struct {
 	fip         bool
 	receivedAny bool
 	firstMoved  bool
+	ttls        *thw.HashWheel
+	ttlseq      uint64 // How up-to-date is the `ttls` THW?
 }
 
 // Represents a message store block and its data.
@@ -244,6 +249,7 @@ type msgBlock struct {
 	syncAlways bool
 	noCompact  bool
 	closed     bool
+	ttls       uint64 // How many msgs have TTLs?
 
 	// Used to mock write failures.
 	mockWriteErr bool
@@ -320,6 +326,9 @@ const (
 
 	// This is the full snapshotted state for the stream.
 	streamStreamStateFile = "index.db"
+
+	// This is the encoded time hash wheel for TTLs.
+	ttlStreamStateFile = "thw.db"
 
 	// AEK key sizes
 	minMetaKeySize = 64
@@ -408,6 +417,11 @@ func newFileStoreWithCreated(fcfg FileStoreConfig, cfg StreamConfig, created tim
 		srv:    fcfg.srv,
 	}
 
+	// Only create a THW if we're going to allow TTLs.
+	if fs.fcfg.EnforceTTLs {
+		fs.ttls = thw.NewHashWheel()
+	}
+
 	// Set flush in place to AsyncFlush which by default is false.
 	fs.fip = !fcfg.AsyncFlush
 
@@ -469,6 +483,13 @@ func newFileStoreWithCreated(fcfg FileStoreConfig, cfg StreamConfig, created tim
 		}
 		// Since we recovered here, make sure to kick ourselves to write out our stream state.
 		fs.dirty++
+	}
+
+	// See if we can bring back our TTL timed hash wheel state from disk.
+	if fcfg.EnforceTTLs {
+		if err = fs.recoverTTLState(); err != nil && !os.IsNotExist(err) {
+			fs.warn("Recovering TTL state from index errored: %v", err)
+		}
 	}
 
 	// Also make sure we get rid of old idx and fss files on return.
@@ -1380,6 +1401,10 @@ func (mb *msgBlock) rebuildStateLocked() (*LostStreamData, []uint64, error) {
 		rl, slen := le.Uint32(hdr[0:]), int(le.Uint16(hdr[20:]))
 
 		hasHeaders := rl&hbit != 0
+		var ttl int64
+		if mb.fs.ttls != nil && len(hdr) > 0 {
+			ttl, _ = getMessageTTL(hdr)
+		}
 		// Clear any headers bit that could be set.
 		rl &^= hbit
 		dlen := int(rl) - msgHdrSize
@@ -1455,6 +1480,11 @@ func (mb *msgBlock) rebuildStateLocked() (*LostStreamData, []uint64, error) {
 		if !mb.dmap.Exists(seq) {
 			mb.msgs++
 			mb.bytes += uint64(rl)
+			if mb.fs.ttls != nil && ttl > 0 {
+				expires := time.Duration(ts) + (time.Second * time.Duration(ttl))
+				mb.fs.ttls.Add(seq, int64(expires))
+				mb.ttls++
+			}
 		}
 
 		// Check for any gaps from compaction, meaning no ebit entry.
@@ -1593,7 +1623,8 @@ func (fs *fileStore) recoverFullState() (rerr error) {
 		}
 	}
 
-	if buf[0] != fullStateMagic || buf[1] != fullStateVersion {
+	version := buf[1]
+	if buf[0] != fullStateMagic || version < fullStateMinVersion || version > fullStateVersion {
 		os.Remove(fn)
 		fs.warn("Stream state magic and version mismatch")
 		return errCorruptState
@@ -1685,6 +1716,10 @@ func (fs *fileStore) recoverFullState() (rerr error) {
 		fs.blks = make([]*msgBlock, 0, numBlocks)
 		for i := 0; i < int(numBlocks); i++ {
 			index, nbytes, fseq, fts, lseq, lts, numDeleted := uint32(readU64()), readU64(), readU64(), readI64(), readU64(), readI64(), readU64()
+			var ttls uint64
+			if version >= 2 {
+				ttls = readU64()
+			}
 			if bi < 0 {
 				os.Remove(fn)
 				return errCorruptState
@@ -1694,6 +1729,7 @@ func (fs *fileStore) recoverFullState() (rerr error) {
 			atomic.StoreUint64(&mb.last.seq, lseq)
 			mb.msgs, mb.bytes = lseq-fseq+1, nbytes
 			mb.first.ts, mb.last.ts = fts+baseTime, lts+baseTime
+			mb.ttls = ttls
 			if numDeleted > 0 {
 				dmap, n, err := avl.Decode(buf[bi:])
 				if err != nil {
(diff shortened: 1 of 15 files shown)


MauriceVanVeen (approved):
LGTM

Found and fixed two bugs; if there was no `index.db` the TTLs would not be recovered (with or without `thw.db`), and if a single message needed to be recovered (when no `thw.db`) it would not due to an off-by-one.

Still work to be done in future PRs, but think this one is good. :+1:

server/filestore.go
```diff
@@ -1455,6 +1473,11 @@ func (mb *msgBlock) rebuildStateLocked() (*LostStreamData, []uint64, error) {
 		if !mb.dmap.Exists(seq) {
 			mb.msgs++
 			mb.bytes += uint64(rl)
+			if ttl > 0 {
+				expires := time.Duration(ts) + (time.Second * time.Duration(ttl))
+				mb.fs.ttls.Add(seq, int64(expires))
```
derekcollison, line 1478: Is fs locked?

Should check ttls not nil here.

derekcollison: Should we also check here if this has already expired?

server/filestore.go
```diff
@@ -3744,18 +3850,31 @@ func (fs *fileStore) storeRawMsg(subj string, hdr, msg []byte, seq uint64, ts in
 	fs.enforceMsgLimit()
 	fs.enforceBytesLimit()
 
+	// Per-message TTL.
+	if fs.ttls != nil && ttl > 0 {
```
derekcollison, line 3854: Should we create fs.ttls here if nil?

server/filestore.go
```diff
@@ -5152,19 +5299,30 @@ func (fs *fileStore) expireMsgs() {
...
 	fs.mu.Lock()
 	defer fs.mu.Unlock()
 
+	// TODO: Not great that we're holding the lock here, but the timed hash wheel isn't thread-safe.
+	var nextTTL int64
+	if fs.ttls != nil {
+		fs.ttls.ExpireTasks(func(seq uint64, ts int64) {
+			fs.removeMsgViaLimits(seq)
+		})
+		nextTTL = fs.ttls.GetNextExpiration(math.MaxInt64)
```
derekcollison, line 5321: This should only pass max int if we are not doing max age, if max age then we should capture the next fireIn from the loop above and pass that here.

server/filestore.go
```diff
@@ -6595,6 +6770,7 @@ func (mb *msgBlock) msgFromBuf(buf []byte, sm *StoreMsg, hh hash.Hash64) (*Store
 	rl := le.Uint32(hdr[0:])
 	hasHeaders := rl&hbit != 0
 	rl &^= hbit // clear header bit
+	rl &^= xbit // clear ttl bit
```
derekcollison, line 6773: I think this can be one line.

`rl &^= (hbit || xbit)`

server/stream.go
```diff
@@ -100,6 +100,13 @@ type StreamConfig struct {
 	// TODO(nat): Can/should we name these better?
 	ConsumerLimits StreamConsumerLimits `json:"consumer_limits"`
 
+	// AllowMsgTTL allows header initiated per-message TTLs. If disabled,
+	// then the `NATS-TTL` header will be ignored.
+	AllowMsgTTL bool `json:"allow_msg_ttl"`
+
+	// LimitsTTL activates writing of messages when limits are applied with a specific TTL.
```
derekcollison, line 107: Not following what this is?

server/stream.go
```diff
@@ -4189,6 +4198,12 @@ func getExpectedLastSeqPerSubjectForSubject(hdr []byte) string {
 	return string(getHeader(JSExpectedLastSubjSeqSubj, hdr))
 }
 
+// Fast lookup of the message TTL.
+func getMessageTTL(hdr []byte) int64 {
```
derekcollison, line 4202: I think for this since it is a header it should be a parseable duration vs an integer.

E.g. `30s, 4h, 7d`

This will resolve the nanos issue.

We should reject if < 1s. So no `20ms` etc.

Also we should reject messages with per ttl set and the stream is not configured to opt-in. I don't think we need to do this at the lower store level but here we should.

server/stream.go
```diff
@@ -4996,17 +5011,20 @@ func (mset *stream) processJetStreamMsg(subject, reply string, hdr, msg []byte,
 		}
 	}
 
+	// Find the message TTL if any.
+	ttl := getMessageTTL(hdr)
```
derekcollison, line 5015: Reject if ttl present and stream is not opted in.

derekcollison (changes requested):
In general looks good.

Let's move hdr to a parseable duration (see inline comments). Also we should reject messages when the stream has not opted in to per msg TTLs.

server/filestore.go
neilalexander: `fs.ttls` is only non-nil when the `AllowMsgTTL` stream option is enabled.

server/filestore.go
neilalexander: Yes I believe so re. lock.

Also we can check if already expired, although I guess if they get put into the THW, they should clean up on the next call to `expireMsgs` normally.

server/stream.go
neilalexander: I don't think we can do this, otherwise we can run into a problem when we source/mirror messages with a TTL into another stream.

server/stream.go
neilalexander: Updated so that it parses as a duration if it can, otherwise it reads the value as whole seconds. So `30s`, `1m` etc work, as does just `30` for 30 seconds.

server/filestore.go
derekcollison: Famous last words ;)

server/stream.go
derekcollison: We need special processing for streams and mirrors anyway that ignores (for ingest criteria) certain headers. This would be one added to the list.

server/filestore.go
derekcollison: Did we get this changed as discussed?

server/filestore.go
```diff
@@ -5152,19 +5305,30 @@ func (fs *fileStore) expireMsgs() {
...
+			fs.mu.Unlock()
+			// Recalculate in case we are expiring a bunch.
+			minAge = time.Now().UnixNano() - maxAge
+		}
 	}
 
 	fs.mu.Lock()
 	defer fs.mu.Unlock()
 
+	// TODO: Not great that we're holding the lock here, but the timed hash wheel isn't thread-safe.
```
derekcollison, line 5321: Could make it fine grained like above.

server/filestore.go
```diff
@@ -5941,6 +6107,7 @@ func (mb *msgBlock) indexCacheBuf(buf []byte) error {
 		seq = le.Uint64(hdr[4:])
 
 		// Clear any headers bit that could be set.
+		hasHeaders := rl&hbit != 0
```
derekcollison, line 6110: This does not mark headers, it marks if we have checked the hash checksum or not.

server/filestore.go
```diff
@@ -6002,6 +6169,16 @@ func (mb *msgBlock) indexCacheBuf(buf []byte) error {
 					})
 				}
 			}
+
+			// Count how many TTLs we think are in this message block.
+			// TODO(nat): Not terribly optimal...
+			if hasHeaders {
```
derekcollison, line 6175: This is not correct.

server/filestore_test.go
```diff
@@ -8296,3 +8296,190 @@ func TestFileStoreNumPendingMulti(t *testing.T) {
 	}
 	require_Equal(t, total, checkTotal)
 }
+
+func TestFileStoreMessageTTL(t *testing.T) {
+	fs, err := newFileStore(
+		FileStoreConfig{StoreDir: t.TempDir(), EnforceTTLs: true},
```
derekcollison, line 8302: Can we not inherit enforeTTLS from the stream config?

server/stream.go
neilalexander: Will add that to the list of follow-up changes to make.

server/filestore_test.go
```diff
@@ -8296,3 +8296,190 @@ func TestFileStoreNumPendingMulti(t *testing.T) {
...
+		require_Equal(t, ss.Msgs, 1)
+
+		time.Sleep(time.Second * 2)
+
+		fs.FastState(&ss)
+		require_Equal(t, ss.FirstSeq, 2)
+		require_Equal(t, ss.LastSeq, 1)
+		require_Equal(t, ss.Msgs, 0)
+	})
+}
```
derekcollison, line 8485: Need to add in test with max age but with msgs with never expire and make sure they do not get expired.

server/stream.go
```diff
@@ -100,6 +100,10 @@ type StreamConfig struct {
 	// TODO(nat): Can/should we name these better?
 	ConsumerLimits StreamConsumerLimits `json:"consumer_limits"`
 
+	// AllowMsgTTL allows header initiated per-message TTLs. If disabled,
+	// then the `NATS-TTL` header will be ignored.
```
derekcollison, line 104: Update comment if we reject these vs ignore.

server/stream.go
```diff
@@ -4194,6 +4206,30 @@ func getExpectedLastSeqPerSubjectForSubject(hdr []byte) string {
...
+// - Positive return value: duration in seconds.
+// - Zero return value: no TTL or parse error.
+// - Negative return value: don't expire.
+func getMessageTTL(hdr []byte) (int64, error) {
+	ttl := getHeader(JSMessageTTL, hdr)
+	if len(ttl) == 0 {
+		return 0, nil
+	}
+	sttl := bytesToString(ttl)
+	dur, err := time.ParseDuration(sttl)
```
derekcollison, line 4219: I thought we were going to use the one we have for configs since it is expanded past what std lib can parse and may be applicable.

derekcollison (changes requested):

server/filestore.go
neilalexander: I looked into this after we spoke yesterday and found it actually has two purposes. When applied to the `mb.cache.idx` it's for the hash checksum, when applied to the `rl` it signifies the presence of a header. 

The `hbit` definition also says about this, although it's a bit vague about the difference:
```
	// Used for marking messages that have had their checksums checked.
	// Used to signal a message record with headers.
	hbit = 1 << 31
```

server/filestore_test.go
neilalexander: In `stream.go` when creating the store for streams, we do indeed set it from the `AllowMsgTTLs` value from the stream config. In the filestore-specific tests, like here, we set it to true directly.

server/stream.go
neilalexander: The `parseDuration` for the config has a function signature that only makes sense for the config parser and returning errors/warnings etc. It also turns out that the `parseDuration` helper is just falling `time.ParseDuration` too.

server/filestore.go
neilalexander: Not straight-forward to do because of how `ExpireTasks` mutates the internal `fs.ttls` state internally in between callbacks, in order to do that we'd either need the THW to have its own lock or for it to be thread-safe to begin with (which it isn't right now).

server/stream.go
neilalexander: At the moment the comment is correct, as of right now the header is passed through and ignored. We probably want to figure out the sourcing/mirroring behaviour at the same time as changing this.

server/filestore.go
derekcollison: My apologies. I wonder if we should have a new constant (same value) that allows us to distinguish?

server/filestore.go
neilalexander: Not yet, I'm still fiddling with this and want to write a proper test for it. Happy to do in either this PR or in a follow-up one.

server/filestore_test.go
derekcollison: Since we pass in a stream config in the tests we could do it there, or you feel we should keep the redundancy and have it in two places?

server/filestore_test.go
derekcollison: ok maybe remove the header then from this PR. Confusing as just a placeholder.

server/stream.go
derekcollison: I saw the thread between you and R.I. and Byron. From an app perspective if I set a header and get no feedback it is not honored that is not good IMO.

Yes we need to fixup the sourcing and mirroring to ignore certain headers. It does so already on some but we need a better solution here.

server/stream.go
derekcollison: ok I thought it was wider.

server/filestore.go
derekcollison: Let's update this one.

server/filestore.go
derekcollison: Gotcha..

server/filestore_test.go
neilalexander: Not attached to the duplication at all, will remove it and use the passed-in `StreamConfig` one.

server/filestore.go
neilalexander: This was done.

neilalexander:
Have rebased the PR, I think it's best to deal with rejecting the TTL'd messages in a separate PR at the same time as stripping them from sourcing/mirroring.

derekcollison (approved):
LGTM

neilalexander:
Fixed merge conflicts.

PR #6282, opened 2024-12-19
Add support for placement tags and `preferred` to metaleader stepdown request

neilalexander:
This makes it possible to nominate a JetStream metaleader transfer to servers matching specific placement tags in addition to cluster names, or optionally, to a specific named server.

* If `preferred_server` is specified then placement will select that server only.
* Otherwise, either `cluster` or `tags` can be supplied for placement. If both the `cluster` and `tags` are supplied then the placement will select a server that matches both.
* If multiple `tags` are specified, the placement will select servers only if they match all supplied tags.

Also factors out some of the logic for making the preferred leader decision so that it could be reused for stream & consumer stepdowns.

### DIFF
--- a/server/jetstream_api.go
+++ b/server/jetstream_api.go
@@ -20,7 +20,6 @@ import (
 	"errors"
 	"fmt"
 	"io"
-	"math/rand"
 	"os"
 	"path/filepath"
 	"runtime"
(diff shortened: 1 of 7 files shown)


server/jetstream_api.go
```diff
@@ -3066,6 +3041,81 @@ func isEmptyRequest(req []byte) bool {
...
+			if si.(nodeInfo).name == preferredServer {
+				preferredLeader = p.ID
+				break
+			}
+		}
+		if preferredLeader == group.ID() {
+			return _EMPTY_, NewJSClusterNoPeersError(fmt.Errorf("preferred server %q is already leader", preferredServer))
+		}
+		if preferredLeader == _EMPTY_ {
+			return _EMPTY_, NewJSClusterNoPeersError(fmt.Errorf("preferred server %q not known", preferredServer))
```
derekcollison, line 3065: Could this be a way for folks to poke and discover all the server names?

neilalexander: I don't think this is any worse than, say, creating and deleting lots of replicated streams and looking at the stream infos to see where the replicas are.

derekcollison: ok

server/jetstream_api.go
```diff
@@ -617,7 +616,8 @@ const JSApiConsumerLeaderStepDownResponseType = "io.nats.jetstream.api.v1.consum
 
 // JSApiLeaderStepdownRequest allows placement control over the meta leader placement.
 type JSApiLeaderStepdownRequest struct {
-	Placement *Placement `json:"placement,omitempty"`
+	PreferredServer string     `json:"preferred_server,omitempty"`
```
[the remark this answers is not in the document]
neilalexander: We can't modify the original `Placement` struct without changing the API for stream configs etc, but I've added a new `PlacementForStepdown` struct that embeds it, so that all keys come under the JSON `placement` key.

server/jetstream_api.go
```diff
@@ -2987,7 +2986,7 @@ func (s *Server) jsLeaderStepDownRequest(sub *subscription, c *client, _ *Accoun
 			s.sendAPIErrResponse(ci, acc, subject, reply, string(msg), s.jsonResponse(&resp))
 			return
 		}
-		if preferredLeader, resp.Error = s.getStepDownPreferredPlacement(cc.meta, req.PreferredServer, req.Placement); resp.Error != nil {
+		if preferredLeader, resp.Error = s.getStepDownPreferredPlacement(cc.meta, req.Placement); resp.Error != nil {
```
derekcollison, line 2989: I think elsewhere in SA and CA groups its just preferred.

server/jetstream_cluster.go
```diff
@@ -83,6 +83,14 @@ type Placement struct {
 	Tags    []string `json:"tags,omitempty"`
 }
 
+// Used for stepdown requests to guide placement in clustered JetStream. Separate
+// struct because we don't want PreferredServer field to be part of the API in
+// stream configs etc.
```
derekcollison, line 88: We could allow this in normal Placement. And we could honor it if selected when creating a stream IMO.

neilalexander: Yep, this PR just deals with the metalayer but could do for streams/consumers for sure.

server/jetstream_api.go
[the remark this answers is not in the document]
neilalexander: Have put it into `Placement` for now and guarded against it being set in stream/consumer configs.

derekcollison (approved):
LGTM

PR #6284, opened 2024-12-19
Add support for placement tags and `preferred` to stream and consumer stepdown requests

neilalexander:
Like #6282, this adds preferred placement support to stream and consumer stepdowns too. 


derekcollison (approved):
LGTM

Issue #6241, opened 2024-12-11
Can not download/install latest nats-server on ARMv7


bligneri:
### Observed behavior

I follow the install process:

curl -sf https://binaries.nats.dev/nats-io/nats-server/v2@v2.10.20 | sh

  ==> Downloading github.com/nats-io/nats-server/v2@v2.10.20
  ==> Resolved version v2.10.20 to 7140387
  ==> Downloading binary for linux armv7
  ==>   URL: https://binaries.nats.dev/binary/github.com/nats-io/nats-server/v2?os=linux&arch=armv7&version=7140387

  Error downloading, got 500 response from server

### Expected behavior

The binary is found and downloaded

### Server and client version

Trying to install `v2.10.20`

### Host environment

This is on  dietpi system (Debian GNU/Linux 11) and this kernel
`Linux hub0xB3D6E65289 4.14.0 #1 SMP Mon Dec 18 11:43:22 CST 2017 armv7l GNU/Linux`

### Steps to reproduce

I tried on three similar machines with different debian version, same architecture and same problem. For instance:
`Linux hub_0xDB84BEEFAE 4.14.0 #1 SMP Mon Dec 18 11:43:22 CST 2017 armv7l GNU/Linux` and Debian Linux 9 and 10.

Same problem: 500 error from the nats-binary server.

wallyqs:
Thanks @bligneri for the report, there is an issue with this downloader script when dealing with armv7 versions which will need to solve, but you can find the latest version of the `nats-server` here: 
https://github.com/nats-io/nats-server/releases/download/v2.10.23/nats-server-v2.10.23-linux-arm7.tar.gz

PR #6286, opened 2024-12-19
Add buildinfo flag

alexbozhenko:
Use https://pkg.go.dev/runtime/debug#BuildInfo to see _all_ the knobs that were used to build the binary, including the Go toolchain version, https://go.dev/doc/godebug values, etc...

This information is already available through 
`go version -m path/to/binary`
But it will be more convenient to have it available without Go installed(which is most likely the case in production)
 
## Test plan:
`go build .`
`./nats-server -buildinfo`
`./nats-server --buildinfo`

Return:
```
# ./nats-server --buildinfo
go      go1.23.4
path    github.com/nats-io/nats-server/v2
mod     github.com/nats-io/nats-server/v2       (devel)
dep     github.com/klauspost/compress   v1.17.11        h1:In6xLpyWOi1+C7tXUUWv2ot1QvBjxevKAaI6IXrJmUc=
dep     github.com/minio/highwayhash    v1.0.3  h1:kbnuUMoHYyVl7szWjSxJnxw11k2U709jqFPPmIUyD6Q=
dep     github.com/nats-io/jwt/v2       v2.7.3  h1:6bNPK+FXgBeAqdj4cYQ0F8ViHRbi7woQLq4W29nUAzE=
dep     github.com/nats-io/nkeys        v0.4.9  h1:qe9Faq2Gxwi6RZnZMXfmGMZkg3afLLOtrU+gDZJ35b0=
dep     github.com/nats-io/nuid v1.0.1  h1:5iA8DT8V7q8WK2EScv2padNa/rTESc1KdnPw4TC2paw=
dep     go.uber.org/automaxprocs        v1.6.0  h1:O3y2/QNTOdbF+e/dpXNNW7Rx2hZ4sTIPyybbxyNqTUs=
dep     golang.org/x/crypto     v0.31.0 h1:ihbySMvVjLAeSH1IbfcRTkD/iNscyz8rGzjF/E5hV6U=
dep     golang.org/x/sys        v0.28.0 h1:Fksou7UEQUWlKvIdsqzJmUmCX3cZuD2+P3XyyzwMhlA=
dep     golang.org/x/time       v0.8.0  h1:9i3RxcPv3PZnitoVGMPDKZSq1xW1gK1Xy3ArNOGZfEg=
build   -buildmode=exe
build   -compiler=gc
build   DefaultGODEBUG=asynctimerchan=1,gotypesalias=0,httpservecontentkeepheaders=1,tls3des=1,tlskyber=0,x509keypairleaf=0,x509negativeserial=1
build   CGO_ENABLED=1
build   CGO_CFLAGS=
build   CGO_CPPFLAGS=
build   CGO_CXXFLAGS=
build   CGO_LDFLAGS=
build   GOARCH=amd64
build   GOOS=linux
build   GOAMD64=v1
build   vcs=git
build   vcs.revision=fedf08be59ab96a74906dd2a3f3c67fa3b286795
build   vcs.time=2024-12-26T22:27:24Z
build   vcs.modified=true
```

### DIFF
--- a/main.go
+++ b/main.go
@@ -107,7 +107,6 @@ func main() {
 	opts, err := server.ConfigureOptions(fs, os.Args[1:],
 		server.PrintServerAndExit,
 		fs.Usage,
-		server.PrintBuildinfoAndExit,
 		server.PrintTLSHelpAndDie)
 	if err != nil {
 		server.PrintAndDie(fmt.Sprintf("%s: %s", exe, err))
--- a/main.go
+++ b/main.go
@@ -86,7 +86,6 @@ Profiling Options:
 Common Options:
     -h, --help                       Show this message
     -v, --version                    Show version
-        --buildinfo                  Show buildinfo
         --help_tls                   TLS help
 `
 
--- a/main.go
+++ b/main.go
@@ -86,6 +86,7 @@ Profiling Options:
 Common Options:
     -h, --help                       Show this message
     -v, --version                    Show version
+        --buildinfo                  Show buildinfo
         --help_tls                   TLS help
 `
 
@@ -106,6 +107,7 @@ func main() {
 	opts, err := server.ConfigureOptions(fs, os.Args[1:],
 		server.PrintServerAndExit,
 		fs.Usage,
+		server.PrintBuildinfoAndExit,
 		server.PrintTLSHelpAndDie)
 	if err != nil {
 		server.PrintAndDie(fmt.Sprintf("%s: %s", exe, err))
--- a/server/opts.go
+++ b/server/opts.go
@@ -5676,7 +5676,7 @@ func getDefaultAuthTimeout(tls *tls.Config, tlsTimeout float64) float64 {
 // specific flags. On success, an options structure is returned configured
 // based on the selected flags and/or configuration file.
 // The command line options take precedence to the ones in the configuration file.
-func ConfigureOptions(fs *flag.FlagSet, args []string, printVersion, printHelp, printBuildinfo, printTLSHelp func()) (*Options, error) {
+func ConfigureOptions(fs *flag.FlagSet, args []string, printVersion, printHelp, printTLSHelp func()) (*Options, error) {
 	opts := &Options{}
 	var (
 		showVersion            bool
@@ -5775,7 +5775,7 @@ func ConfigureOptions(fs *flag.FlagSet, args []string, printVersion, printHelp, 
 	}
 
 	if showBuildInfo {
-		printBuildinfo()
+		PrintBuildinfoAndExit()
 		return nil, nil
 	}
 
@@ -5790,8 +5790,8 @@ func ConfigureOptions(fs *flag.FlagSet, args []string, printVersion, printHelp, 
 	}
 
 	// Process args looking for non-flag options,
-	// 'version', 'help' and 'buildinfo' only for now
-	showVersion, showHelp, showBuildInfo, err = ProcessCommandLineArgs(fs)
+	// 'version' and 'help' only for now
+	showVersion, showHelp, err = ProcessCommandLineArgs(fs)
 	if err != nil {
 		return nil, err
 	} else if showVersion {
@@ -5799,9 +5799,6 @@ func ConfigureOptions(fs *flag.FlagSet, args []string, printVersion, printHelp, 
 		return nil, nil
 	} else if showHelp {
 		printHelp()
-		return nil, nil
-	} else if showBuildInfo {
-		printBuildinfo()
 		return nil, nil
 	}
 
--- a/server/opts.go
+++ b/server/opts.go
@@ -5775,7 +5775,7 @@ func ConfigureOptions(fs *flag.FlagSet, args []string, printVersion, printHelp, 
 	}
 
 	if showBuildInfo {
-		printBuildinfoAndExit()
+		printBuildInfoAndExit()
 		return nil, nil
 	}
 
--- a/server/opts.go
+++ b/server/opts.go
@@ -5775,7 +5775,7 @@ func ConfigureOptions(fs *flag.FlagSet, args []string, printVersion, printHelp, 
 	}
 
 	if showBuildInfo {
-		PrintBuildinfoAndExit()
+		printBuildinfoAndExit()
 		return nil, nil
 	}
 
--- a/server/opts.go
+++ b/server/opts.go
@@ -5676,10 +5676,11 @@ func getDefaultAuthTimeout(tls *tls.Config, tlsTimeout float64) float64 {
 // specific flags. On success, an options structure is returned configured
 // based on the selected flags and/or configuration file.
 // The command line options take precedence to the ones in the configuration file.
-func ConfigureOptions(fs *flag.FlagSet, args []string, printVersion, printHelp, printTLSHelp func()) (*Options, error) {
+func ConfigureOptions(fs *flag.FlagSet, args []string, printVersion, printHelp, printBuildinfo, printTLSHelp func()) (*Options, error) {
 	opts := &Options{}
 	var (
 		showVersion            bool
+		showBuildInfo          bool
 		showHelp               bool
 		showTLSHelp            bool
 		signal                 string
@@ -5735,6 +5736,7 @@ func ConfigureOptions(fs *flag.FlagSet, args []string, printVersion, printHelp, 
 	fs.StringVar(&opts.RemoteSyslog, "remote_syslog", _EMPTY_, "Syslog server addr (udp://127.0.0.1:514).")
 	fs.BoolVar(&showVersion, "version", false, "Print version information.")
 	fs.BoolVar(&showVersion, "v", false, "Print version information.")
+	fs.BoolVar(&showBuildInfo, "buildinfo", false, "Print buildinfo.")
 	fs.IntVar(&opts.ProfPort, "profile", 0, "Profiling HTTP port.")
 	fs.StringVar(&opts.RoutesStr, "routes", _EMPTY_, "Routes to actively solicit a connection.")
 	fs.StringVar(&opts.Cluster.ListenStr, "cluster", _EMPTY_, "Cluster url from which members can solicit routes.")
@@ -5772,6 +5774,11 @@ func ConfigureOptions(fs *flag.FlagSet, args []string, printVersion, printHelp, 
 		return nil, nil
 	}
 
+	if showBuildInfo {
+		printBuildinfo()
+		return nil, nil
+	}
+
 	if showHelp {
 		printHelp()
 		return nil, nil
@@ -5783,8 +5790,8 @@ func ConfigureOptions(fs *flag.FlagSet, args []string, printVersion, printHelp, 
 	}
 
 	// Process args looking for non-flag options,
-	// 'version' and 'help' only for now
-	showVersion, showHelp, err = ProcessCommandLineArgs(fs)
+	// 'version', 'help' and 'buildinfo' only for now
+	showVersion, showHelp, showBuildInfo, err = ProcessCommandLineArgs(fs)
 	if err != nil {
 		return nil, err
 	} else if showVersion {
@@ -5792,6 +5799,9 @@ func ConfigureOptions(fs *flag.FlagSet, args []string, printVersion, printHelp, 
 		return nil, nil
 	} else if showHelp {
 		printHelp()
+		return nil, nil
+	} else if showBuildInfo {
+		printBuildinfo()
 		return nil, nil
 	}
 
--- a/server/server.go
+++ b/server/server.go
@@ -1599,22 +1599,20 @@ func PrintBuildinfoAndExit() {
 // ProcessCommandLineArgs takes the command line arguments
 // validating and setting flags for handling in case any
 // sub command was present.
-func ProcessCommandLineArgs(cmd *flag.FlagSet) (showVersion bool, showHelp bool, showBuildinfo bool, err error) {
+func ProcessCommandLineArgs(cmd *flag.FlagSet) (showVersion bool, showHelp bool, err error) {
 	if len(cmd.Args()) > 0 {
 		arg := cmd.Args()[0]
 		switch strings.ToLower(arg) {
 		case "version":
-			return true, false, false, nil
+			return true, false, nil
 		case "help":
-			return false, true, false, nil
-		case "buildinfo":
-			return false, false, true, nil
+			return false, true, nil
 		default:
-			return false, false, false, fmt.Errorf("unrecognized command: %q", arg)
+			return false, false, fmt.Errorf("unrecognized command: %q", arg)
 		}
 	}
 
-	return false, false, false, nil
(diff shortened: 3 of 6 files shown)


server/server.go
```diff
@@ -1583,23 +1584,37 @@ func PrintServerAndExit() {
 	os.Exit(0)
 }
 
+// PrintBuildinfoAndExit will print out https://pkg.go.dev/runtime/debug#BuildInfo and exit.
+func PrintBuildinfoAndExit() {
+	bi, ok := debug.ReadBuildInfo()
+	if !ok || bi == nil {
+		//nolint:forbidigo
```
alexbozhenko, line 1591: @neilalexander ok to ignore like this here?

server/opts.go
```diff
@@ -5676,10 +5676,11 @@ func getDefaultAuthTimeout(tls *tls.Config, tlsTimeout float64) float64 {
 // specific flags. On success, an options structure is returned configured
 // based on the selected flags and/or configuration file.
 // The command line options take precedence to the ones in the configuration file.
-func ConfigureOptions(fs *flag.FlagSet, args []string, printVersion, printHelp, printTLSHelp func()) (*Options, error) {
+func ConfigureOptions(fs *flag.FlagSet, args []string, printVersion, printHelp, printBuildinfo, printTLSHelp func()) (*Options, error) {
```
bruth, line 5679: Embedding the NATS server is quite common and since this is a public function, this is a breaking change we don't want to do.

server/server.go
```diff
@@ -1583,23 +1584,37 @@ func PrintServerAndExit() {
...
+	//nolint:forbidigo
+	fmt.Println(bi)
+	os.Exit(0)
+}
+
 // ProcessCommandLineArgs takes the command line arguments
 // validating and setting flags for handling in case any
 // sub command was present.
-func ProcessCommandLineArgs(cmd *flag.FlagSet) (showVersion bool, showHelp bool, err error) {
+func ProcessCommandLineArgs(cmd *flag.FlagSet) (showVersion bool, showHelp bool, showBuildinfo bool, err error) {
```
bruth, line 1602: Same as above, this is a breaking change.

bruth (changes requested):
IMO it is a pretty niche desire for wanting to read the build info if you do not have Go installed. That said, I suppose there is no harm in adding a CLI option as long as it does not introducing breaking changes.

server/server.go
bruth: What is the purpose of introducing this directive? After reading about the `forbidigo` tool, it flags `fmt.Print*` statements by default, however we use them elsewhere without issue.

alexbozhenko: I added it to disable the lint warning, which was added here:
https://github.com/nats-io/nats-server/pull/3812/files

All the usages that we have are excluded in the .golangci.yml config:
```
# rg 'fmt\.Print(f|ln)?' --glob '!*_test.go'
main.go
95:     fmt.Printf("%s\n", usageStr)

server/server.go
1583:   fmt.Printf("nats-server: v%s\n", VERSION)
1592:           fmt.Println("failed to read Buildinfo")
1595:   fmt.Println(bi)

server/opts.go
4578:   fmt.Printf("%s", tlsUsage)
4580:           fmt.Printf("    %s\n", k)
4582:   fmt.Printf("\nAvailable curve preferences include:\n")
4584:           fmt.Printf("    %s\n", k)
4587:           fmt.Printf("%s\n", certstore.Usage)
4589:   fmt.Printf("%s", certidp.OCSPPeerUsage)
4590:   fmt.Printf("%s", OCSPResponseCacheUsage)
```
All other usages are in test files.

alexbozhenko: And this
https://github.com/nats-io/nats-server/blob/cee44bef8de38b6c8a11ca6e332f998831a8e4c2/scripts/runTestsOnTravis.sh#L9-L11
will exit non-0 if there are any linter warnings, so I had to do something about it.

server/server.go
```diff
@@ -1583,6 +1584,18 @@ func PrintServerAndExit() {
 	os.Exit(0)
 }
 
+// PrintBuildinfoAndExit will print out https://pkg.go.dev/runtime/debug#BuildInfo and exit.
+func PrintBuildinfoAndExit() {
```
ripienaar, line 1588: should this be exported? Doesn't seem like something that would be useful when embedded or have use as a callable API really

server/server.go
bruth: Thanks for the context.

server/server.go
bruth: I was going to make the same comment, grant it the `PrintServerAndExit` function is also exported. I am generally in favor of exporting less..

bruth (approved):
LGTM

ripienaar:
Like @bruth not really sure of the use case here.  We'd get better milage from publishing a SBOM in the industry standard format.

alexbozhenko:
@ripienaar: the use case is to find out what kind of binary I am dealing with.
Even though I know "version" and "Git" from the output that is printed when the server starts, I do not have visibility into [all the other knobs](https://github.com/nats-io/nats-server/blob/cee44bef8de38b6c8a11ca6e332f998831a8e4c2/.goreleaser.yml#L14-L39) that may be tuned, and knowing just git commit sha is not enough to, e.g. reproduce the binary.
I am not sure about SBOM... Looks like it is concerned with dependency info?
Related: https://github.com/ConnectEverything/nats-binaries/issues/32

ripienaar:
I understand why you would want it @alexbozhenko - but this is not something anyone else have asked for.  So as a developer I can see the use case, but then I'd just run `go version -m nats-server` and get my answer.

What I wouldn't do is expose developer specific information as flags to end users who would only find it adds more UI clutter and surprising or confusing output.  So my question really is: As a end-user, how is this useful?

server/opts.go
```diff
@@ -5772,6 +5774,11 @@ func ConfigureOptions(fs *flag.FlagSet, args []string, printVersion, printHelp,
 		return nil, nil
 	}
 
+	if showBuildInfo {
+		printBuildinfoAndExit()
```
wallyqs, line 5778:
```suggestion
		printBuildInfoAndExit()
```

server/server.go
```diff
@@ -1583,6 +1584,18 @@ func PrintServerAndExit() {
 	os.Exit(0)
 }
 
+// printBuildinfoAndExit will print out https://pkg.go.dev/runtime/debug#BuildInfo and exit.
+func printBuildinfoAndExit() {
```
wallyqs, line 1588:
```suggestion
func printBuildInfoAndExit() {
```

server/server.go
```diff
@@ -1583,6 +1584,18 @@ func PrintServerAndExit() {
...
+	bi, ok := debug.ReadBuildInfo()
+	if !ok || bi == nil {
+		//nolint:forbidigo
+		fmt.Println("failed to read Buildinfo")
```
wallyqs, line 1592: maybe instead something like this:
```suggestion
		fmt.Fprintln(os.Stderr, "failed to read BuildInfo")
```

alexbozhenko:
@ripienaar 
Here is an example of an issue the user faced:
https://github.com/nats-io/nats-server/issues/6241
It is easy to imagine scenarios where having easy access to the specifics of the build would be useful.

`go version -m` doesn't work without jumping through hoops, e.g. with a container. But this works and it is nice:
```
# docker run -it nats:latest --version
nats-server: v2.10.24
```

Other tools expose additional info about how binary was built too:
image
image

main.go
```diff
@@ -86,6 +86,7 @@ Profiling Options:
 Common Options:
     -h, --help                       Show this message
     -v, --version                    Show version
+        --buildinfo                  Show buildinfo
```
wallyqs, line 89: I think we can keep this one hidden since mostly for troubleshooting.

ripienaar:
The linked user issue is a case of the binary not being downloaded at all - how would the flag have helped there?

```
curl -sf https://binaries.nats.dev/nats-io/nats-server/v2@v2.10.20 | sh
...
Error downloading, got 500 response from server
```

Again, I understand this makes the information easy to find - I am wondering what the end user utility of this information is, the CLI flags to ants-server is end user interface, not developer user interface so we need to frame these kinds of additions in the light of end users

alexbozhenko:
Here is how I thought the  flag can be beneficial for the user:
1. (once it is fixed) user downloads `curl -sf https://binaries.nats.dev/nats-io/nats-server/v2@v2.10.20 | sh`
2. The user runs `nats-server --buildinfo` to check that it is indeed an ARMv7 binary

Or, let's imagine we decided to tune some other flag(existing or some future one). Let's say we start to build with GOAMD64=v4. And let's say such binary crashes for some users. If we make it easier to figure out the reason - that would be great. That's why I would like to have buildinfo exposed for the users.

Answering the question "What kind of binary I am dealing with" is a valid user concern. And our users are developers, so they won't consider this info clutter...
In addition to being useful for the users, it also would be useful for us - the support team, supporting paying customers.

wallyqs (commented):
Added some comments, also think might be too niche but could be convenient when trying to get on the same page with users when reproducing issues to get the build info from the Go version used to build it just in case instead of the Go version that might be available locally or when do not have one at hand.

ripienaar:
> Here is how I thought the flag can be beneficial for the user in this case:
> 
> 1. (once it is fixed) user downloads `curl -sf https://binaries.nats.dev/nats-io/nats-server/v2@v2.10.20 | sh`
> 2. The user runs `nats-server --buildinfo` to check that it is indeed an ARMv7 binary
> 
> Or, let's imagine we decided to tune some other flag(existing or some future one). Let's say we start to build with GOAMD64=v4. And let's say such binary crashes for some users. If we make it easier to figure out the reason - that would be great. That's why I would like to have buildinfo exposed for the users.
> 
> Answering the question "What kind of binary I am dealing with" is a valid user concern. And our users are developers, so they won't consider this info clutter... In addition to being useful for the users, it also would be useful for us - the support team, supporting paying customers.

It's either going to fail to run if its the wrong architecture or in some cases like some ARM combos you can just check it in the case that it would run:

```
~ $ file /usr/bin/nats-server
/usr/bin/nats-server: ELF 64-bit LSB executable, ARM aarch64, version 1 (SYSV), statically linked, Go BuildID=fOeCzTD4TNmbYbIIzz2c/oD-ysM25TgDdDGwD0JwT/uRlUS4nmuvKtXW5cMzA4/pSUltOa7AgJKAdmrnUrK, stripped
```

It seems very contrived to have to add this imo, super niche need, minimal utility to end users.  As we only support binaries released officially, none of these buildinfo checks would tell you if this is a supported version / build or not - verifying signatures, checksums against GitHub  etc, those are the only ways.

alexbozhenko:
@ripienaar I updated to hide the helpfrom the users, is it a good compromise? It may be useful for us in support, so could we plase have it there? Users won't see it, and won't be confused...

ripienaar:
Hiding the flag is better. 

I think with 3 out of 3 reviewers voicing concern about the validity of the feature there may be an opportunity to take step back and see if this can be done in a more  Idiomatic way. We don’t tend to expose useful info only on the command line

wallyqs:
maybe this would be better suited to report via varz? there is some buildinfo in that struct too

wallyqs (approved):
LGTM

wallyqs:
Another option would be to have something like a `nats-server --debug version` instead of adding a new flag.

ripienaar:
varz would be better than command line but I think doing that could be very undesirable in some embedded uses, people might be embedding sensitive information into their binaries using linker flags that will then become exposed

we should hear from the support team if this is really needed for their purposes @jarretlavallee

alexbozhenko:
> It seems very contrived to have to add this imo, super niche need, minimal utility to end users. 

I just gave you a very concrete example. Users may want to see how the binary was built.
File utility does not tell that this is an armv7 binary:
```
# file nats-server-v2.10.23-linux-arm7/nats-server
nats-server-v2.10.23-linux-arm7/nats-server: ELF 32-bit LSB executable, ARM, EABI5 version 1 (SYSV), statically linked, Go BuildID=7wTsSUPSvxOlzSxeWQae/KeLmkr5FXBN0b70VaGDv/RpLkgf7Zu61EX0gl147Y/5waZ3bhqH5ySTYMbjNcr, not stripped
```
But buildinfo does.

> varz would be better than the command line 
Why? With `varz` I can not do e.g. this: `docker run -it nats:latest --buildinfo`.

We removed it from the usage text, so the flag will be hidden. So your concern about "cluttering the UI" is addressed. Now it seems like you just do not want me to add any cli flag...
We can do `nats-server --debug version` if you think this is better. In that case, it will be very explicit, what kind of output user may expect there. We can also make it output JSON, making it more idiomatic(other tools do that)

> we should hear from the support team

I am on the support team.

ripienaar:
When adding information to the NATS Server we have worked hard to do so in a way that supports our eco system of tools - this means adding it to the network queryable endpoints.  This is there to support tools like `nats`, support efforts like `nats audit` so we can gather this in bulk and later build reporting tools that can proactively alert users of issues, tools like Control Plane and our efforts around things like fleet management.  

This is not something we take lightly, we've invested many months of development time in these efforts and its the central enabler for many tools and features.  It's how we do things, it's what is idiomatic in our eco system.

Every bit of new code is code that requires maintenance and considerations around security exposure and more. So it's important that we add things that are really needed and that we consider them from all angles like their utility and security.

So to your questions:

 * You are right I do not want to add this to the CLI, simply because that is just now how we do things. We put metadata on the network where it's accessible. We already expose some build info adding more in the same way would be the correct approach if doing that satisfies the criteria of it being needed and being secure. We want to support the efforts of various teams in building tooling so doing this in an idiomatic way is best. Why would I support a way that is specifically counter to our general established approach?
 * You said this is so utilities in docker isnt needed for this and so the cli flag make sense. However the basic way that docker works just makes this unnecessary- you will automatically run the right architecture for your particular setup unless the containers are corrupt.  And if you are using docker you can use standard docker utilities to view the architecture and OS of your container.  And the basic reality is in most cases you simply can't run the wrong architecture binary.
 * As before multiple reviewers here have said this feature is probably not useful. I have asked other via DM and they too do not think this feature is useful, so its worth asking the question of Why?
 * Yes you are on the support team, however the insight of team members who have been on the team for years and have extensive experience in the issues WE run into are valuable. Especially given that there seems little support for this feature. Asking for input from others is not a problem.

It's valuable to bring up these kinds of feature requests in meetings like the Field Ops discussions where background information can be obtained and methodologies shared so that implementations are done in line with our expectations, a bit of discussing ahead of time will yield much better outcomes for the time you invest in PRs

As you say its 20 lines of code, it would be much more valuable if implemented in a way in line with how the project approach these kinds of features and thus supporting our wider eco system at the same time.

Jarema:
Just to ad my two cents - I feel that if that went through ADR proposal followed up by a short discussion we would reach the consensus quicker. Let's chat about it.

alexbozhenko:
Thanks, @ripienaar. Ok, closing.
Will try to bring up adding it to `/varz` in the oss meeting

PR #6287, opened 2024-12-20
Filestore: Disambiguate `hbit` usage with new `cbit` flag

neilalexander:
The `hbit` had two purposes depending on whether it was applied to the `mb.cache.idx` or the `rl` but it could be confusing. This PR adds a new `cbit` for clarity, although it has the same computed value. 


derekcollison (approved):
LGTM

PR #6292, opened 2024-12-22
Disable JetStream on disk errors

souravagrawal:
Currently, there are scenarios where NATS JetStream may encounter permission errors when file system goes into read only mode, which can lead to an inconsistent state. In such cases, the system continues to allow publishing messages by resetting stream state, leading to a misaligned consumer stream sequence.

This PR introduces changes to gracefully handle these permission errors and prevent NATS from continuing in an inconsistent state when:

- Flushing stream state to disk
- Deleting expired messages on startup
- Creating new block for messages
After this PR, If NATS is running in non-clustered mode, the user will be unable to issue write requests until the issue is resolved. In clustered mode, only the affected node will stop accepting requests, while the system will continue to function as long as the required quorum remains healthy.

PR potentially fixes : #6211 which leads to consumer sequence reaching higher than stream sequence.
Signed-off-by: Sourabh Agrawal souravagrawal1111@gmail.com


neilalexander (commented):
We should probably have better detection of a wider array of filesystem errors like permissions etc, agreed, but I'm not entirely sure why this would be a configurable option and I don't particularly like it either.

The disk being pulled out from underneath JetStream is effectively a catastrophic operational issue and for JS to continue to try to operate in those circumstances feels problematic, i.e. with the potential for data loss that no one may even notice. Even more so in a clustered scenario where the metalayer and Raft logs are being written out to disk.

souravagrawal:
> We should probably have better detection of a wider array of filesystem errors like permissions etc, agreed, but I'm not entirely sure why this would be a configurable option and I don't particularly like it either.
> 
> The disk being pulled out from underneath JetStream is effectively a catastrophic operational issue and for JS to continue to try to operate in those circumstances feels problematic, i.e. with the potential for data loss that no one may even notice. Even more so in a clustered scenario where the metalayer and Raft logs are being written out to disk.

Thank you for taking the time to review this. 
I will remove the flag. As this is my first patch in the server, I wanted to err on the side of caution. 
Are there any other cases you think I should address as part of this?

souravagrawal:
Reopening PR, as it was closed mistakenly

souravagrawal:
> > We should probably have better detection of a wider array of filesystem errors like permissions etc, agreed, but I'm not entirely sure why this would be a configurable option and I don't particularly like it either.
> > The disk being pulled out from underneath JetStream is effectively a catastrophic operational issue and for JS to continue to try to operate in those circumstances feels problematic, i.e. with the potential for data loss that no one may even notice. Even more so in a clustered scenario where the metalayer and Raft logs are being written out to disk.
> 
> Thank you for taking the time to review this. I will remove the flag. As this is my first patch in the server, I wanted to err on the side of caution. Are there any other cases you think I should address as part of this?

Thank you for the suggestion @neilalexander. I have implemented the recommended change and removed the flag.

souravagrawal:
Hello @neilalexander, Happy New Year. 
Just following up to check if you’ve had a chance to review the changes.

server/filestore.go
```diff
@@ -516,7 +516,14 @@ func newFileStoreWithCreated(fcfg FileStoreConfig, cfg StreamConfig, created tim
 
 	// Do age checks too, make sure to call in place.
 	if fs.cfg.MaxAge != 0 {
-		fs.expireMsgsOnRecover()
+		err := fs.expireMsgsOnRecover()
+		if err != nil && err == errFileSystemPermissionDenied {
+			fs.srv.Warnf("file system permission denied while expiring msgs, disabling jetstream: %v", err)
```
neilalexander, line 521: Please make sure any log lines are capitalised/cased correctly.

server/filestore.go
```diff
@@ -8111,13 +8131,20 @@ func (mb *msgBlock) dirtyCloseWithRemove(remove bool) {
 		// Clear any tracking by subject if we are removing.
 		mb.fss = nil
 		if mb.mfn != _EMPTY_ {
-			os.Remove(mb.mfn)
+			err := os.Remove(mb.mfn)
+			if err != nil && os.IsPermission(err){
+				return errFileSystemPermissionDenied
```
neilalexander, line 8136: Is there a reason to do this rather than just check `os.IsPermission` further up the callstack?

neilalexander (changes requested):
I think having the `filestore.go` call `DisableJetStream` directly may not be the best approach here. In a number of places we already surface errors up to `stream.go`, i.e. for `isOutOfSpaceErr`.

It feels like the more natural thing to do here would be to extend that same pattern, such that the stream examines surfaced errors and reacts, similar to how we do with the disk out-of-space warnings. Otherwise we end up with different types of disk errors being handled differently and that makes the code difficult to trace.

server/filestore.go
neilalexander: It may even make sense to define a new error-detect function for this case, similar to how we already have this in `store.go` for out-of-space:
```
func isOutOfSpaceErr(err error) bool {
	return err != nil && (strings.Contains(err.Error(), "no space left"))
}
```

souravagrawal:
> I think having the `filestore.go` call `DisableJetStream` directly may not be the best approach here. In a number of places we already surface errors up to `stream.go`, i.e. for `isOutOfSpaceErr`.
> 
> It feels like the more natural thing to do here would be to extend that same pattern, such that the stream examines surfaced errors and reacts, similar to how we do with the disk out-of-space warnings. Otherwise we end up with different types of disk errors being handled differently and that makes the code difficult to trace.

I agree; it makes sense to handle this within the same pattern. I have implemented the change as you suggested. 
However, there are a few other cases where stream-related operations, such as expireMsgs and writeFullState, are executed within a Go routine and need to be addressed in filestore.go, I believe.

server/filestore.go
```diff
@@ -8534,7 +8555,15 @@ func (fs *fileStore) flushStreamStateLoop(qch, done chan struct{}) {
 	for {
 		select {
 		case <-t.C:
-			fs.writeFullState()
+			err := fs.writeFullState()
+			if isPermissionError(err) {
+				fs.warn("File system permission denied when flushing stream state, disabling jetstream %v", err)
+				// messages in block cache could be lost in the worst case.
+				// In the clustered mode it is very highly unlikely as a result of replication.
+				fs.srv.DisableJetStream()
```
neilalexander, line 8563: I don't _think_ `fs.srv` can be `nil` here but let's add a `fs.srv != nil` guard just in case.

server/filestore.go
```diff
@@ -8743,6 +8772,10 @@ func (fs *fileStore) _writeFullState(force bool) error {
 	// Protect with dios.
 	<-dios
 	err := os.WriteFile(fn, buf, defaultFilePerms)
+	// if file system is not writable isPermissionError is set to true
+	if isPermissionError(err) {
```
neilalexander, line 8776: This **must** be done after the `dios <- struct{}{}` line, otherwise the semaphore becomes unbalanced in a failure scenario.

server/stream.go
```diff
@@ -5098,6 +5098,14 @@ func (mset *stream) processJetStreamMsg(subject, reply string, hdr, msg []byte,
 	}
 
 	if err != nil {
+		if isPermissionError(err) {
+			mset.mu.Unlock()
+			// messages in block cache could be lost in the worst case.
+			// In the clustered mode it is very highly unlikely as a result of replication.
+			mset.srv.DisableJetStream()
+			mset.srv.Warnf("File system permission denied while writing msg, disabling jetstream: %v", err)
```
neilalexander, line 5106: Same nit here with capitalisation etc. 

`Filesystem permission denied while writing message, disabling JetStream: %v`

souravagrawal:
Thank you, @neilalexander. All the comments have been addressed.

neilalexander:
Overall looks OK, but CI is reporting `server/opts.go:6089: File is not `goimports`-ed (goimports)`. 

It doesn't actually look like anything meaningful in that file has changed, so should just be able to do `git checkout main -- server/opts.go` and re-add.

souravagrawal:
> Overall looks OK, but CI is reporting `server/opts.go:6089: File is not goimports-ed (goimports)`.
> 
> It doesn't actually look like anything meaningful in that file has changed, so should just be able to do `git checkout main -- server/opts.go` and re-add.

Hello, @neilalexander. I have made the suggested change, and CI checks are now passing.

neilalexander (approved):
LGTM

PR #6298, opened 2024-12-23
Stream message proposal flags, allow sourcing/mirroring to preserve TTL headers

neilalexander:
Now that sourcing/mirroring strips the TTL header, it is safe to reject messages with the `Nats-TTL` header if `AllowMsgTTL` is not set on the stream.

Sourcing/mirroring tests are both clustered and non-clustered as there are different codepaths for sourcing/mirroring when clustered.


server/stream.go
```diff
@@ -2418,6 +2418,12 @@ func (mset *stream) processInboundMirrorMsg(m *inMsg) bool {
 		return false
 	}
 
+	// If the TTL header is set, remove it.
+	hdr := m.hdr
+	if len(hdr) > 0 {
```
derekcollison, line 2423: I think instead of removing we flag that the messages is sourced to this function and if so we ignore this plus a bunch of others..

server/stream.go
```diff
@@ -3447,8 +3453,10 @@ func (mset *stream) processInboundSourceMsg(si *sourceInfo, m *inMsg) bool {
 	hdr, msg := m.hdr, m.msg
 
 	// If we are daisy chained here make sure to remove the original one.
+	// If the TTL header is set, remove it.
```
derekcollison, line 3456: Let's rework this logic to not remove but ignore.

server/stream.go
```diff
@@ -4780,6 +4789,19 @@ func (mset *stream) processJetStreamMsg(subject, reply string, hdr, msg []byte,
 			}
 		}
 
+		// TTL'd messages are rejected entirely if TTLs are not enabled on the stream.
+		if ttl, _ := getMessageTTL(hdr); ttl != 0 && !mset.cfg.AllowMsgTTL {
+			mset.mu.Unlock()
```
derekcollison, line 4794: We can do this before proposing to the NRG so no need to bump CLFS etc.

derekcollison (changes requested):

server/stream.go
neilalexander: Is this wise? If we have to rebuild the mirroring/sourcing stream, how do we know whether or not to recover the TTLs? Do we just never recover the TTLs if there are mirrors/sources present?

server/stream.go
neilalexander: Have added code to do the check before proposal in clustered mode too as well as a unit test to check CLFS wasn't bumped. We still need it in `processJetStreamMsg` for single server setups though.

server/stream.go
neilalexander: Until we have a good story for how TTLs should work across sources/mirrors, it seems like a better approach to remove the headers. That way there is no possible way during, i.e. a recovery scenario, that we can apply TTLs in the wrong places.

derekcollison: I don't prefer removing things from the original messages though myself.

PR #6306, opened 2024-12-28
Enable SBOM config for GoReleaser

bruth:
This will generate a new set of SPDX-formatted SBOM files on a release using the the GoReleaser + Syft toolchain.

Travis currenly invocates GoReleaser so the Syft dependency needs to be installed ahead of the `deploy` action.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -52,6 +52,9 @@ jobs:
 
 script: ./scripts/runTestsOnTravis.sh $TEST_SUITE
 
+before_deploy:
+  - curl -sSfL https://raw.githubusercontent.com/anchore/syft/main/install.sh | sh -s -- -b /usr/local/bin
+
 deploy:
   provider: script
   cleanup: true
--- a/.travis.yml
+++ b/.travis.yml
@@ -52,8 +52,11 @@ jobs:
 
 script: ./scripts/runTestsOnTravis.sh $TEST_SUITE
 
+# Install Syft which is used by GoReleaser in the deploy step.
+# Cosign is an optional dependency to verify the Syft binary.
 before_deploy:
-  - curl -sSfL https://raw.githubusercontent.com/anchore/syft/main/install.sh | sh -s -- -b /usr/local/bin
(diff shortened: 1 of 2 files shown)


ripienaar (approved):
nice addition

.travis.yml
```diff
@@ -52,6 +52,9 @@ jobs:
 
 script: ./scripts/runTestsOnTravis.sh $TEST_SUITE
 
+before_deploy:
+  - curl -sSfL https://raw.githubusercontent.com/anchore/syft/main/install.sh | sh -s -- -b /usr/local/bin
```
wallyqs, line 56: I think we need to also fetch a recent `cosign` version since the default one that is installed will be picked up by the script might be too old to be compatible, then override `COSIGN_BINARY` to use that one when calling it

bruth: Cosign is used for verifying the checksum of Syft itself and only used when the `-v` option is set when running the script. Ironically, it is not in the default one-liner (for which I copied from the website). However, I will add since it is extra validation.

wallyqs (approved):
LGTM

.travis.yml
wallyqs: Thanks for clarifying 👍

.travis.yml
```diff
@@ -52,6 +52,12 @@ jobs:
 
 script: ./scripts/runTestsOnTravis.sh $TEST_SUITE
 
+# Install Syft which is used by GoReleaser in the deploy step.
+# Cosign is an optional dependency to verify the Syft binary.
+before_deploy:
+  - curl -o /usr/local/bin/cosign -L https://github.com/sigstore/cosign/releases/latest/download/cosign-linux-amd64 && chmod +x /usr/local/bin/cosign
```
alexbozhenko, line 58: @bruth should versions of these tools be pinned to specific version?
https://reproducible-builds.org/docs/commandments/

bruth: Cosign I am not worried about, but I will add a tag to the install script for Syft.

bruth: As a followup, Syft is not used for the build itself, only for the SBOM files.

PR #6319, opened 2025-01-03
Add support for "never expires" TTL

neilalexander:
A message with `Nats-TTL: never` will never expire, even if the `MaxAge` policy is set.

### DIFF
--- a/server/filestore.go
+++ b/server/filestore.go
@@ -5304,7 +5304,14 @@ func (fs *fileStore) expireMsgs() {
 	fs.mu.RUnlock()
 
 	if maxAge > 0 {
-		for sm, _ = fs.msgForSeq(0, &smv); sm != nil && sm.ts <= minAge; sm, _ = fs.msgForSeq(0, &smv) {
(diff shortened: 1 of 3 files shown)


server/filestore.go
```diff
@@ -5299,23 +5300,38 @@ func (fs *fileStore) expireMsgs() {
...
+					eflr = sm.seq + 1
+					continue
+				}
+			}
 			fs.mu.Lock()
 			fs.removeMsgViaLimits(sm.seq)
 			fs.mu.Unlock()
 			// Recalculate in case we are expiring a bunch.
 			minAge = time.Now().UnixNano() - maxAge
+			eflr = sm.seq + 1
```
neilalexander, line 5323: Note that when we do this, `sm.seq + 1` may not actually exist and we could stumble across a gap, as it could have been deleted. Hence the need to add the `next` flag into `msgForSeq` to scan for the next sequence that _does_ exist. 

Without this, it's not possible to have messages A, B and C where A has `Nats-TTL: never`, B has maybe been deleted and C has any other TTL.

MauriceVanVeen (approved):
LGTM

PR #6326, opened 2025-01-06
Strip unnecessary `ClientInfo` fields from stream & consumer assignment proposals

neilalexander:
This avoids JWTs etc from the client info from being encoded into stream and consumer assignments proposals when they are not needed.

This will reduce the network bandwidth required to send such proposals in some deployments, as well as reducing the memory footprint of proposals queued in memory and the disk footprint required for the metalayer WAL.

### DIFF
--- a/server/events.go
+++ b/server/events.go
@@ -336,6 +336,17 @@ func (ci *ClientInfo) forAssignmentSnap() *ClientInfo {
 	}
 }
 
+// forProposal returns the minimum amount of ClientInfo we need for assignment proposals.
+func (ci *ClientInfo) forProposal() *ClientInfo {
+	if ci == nil {
+		return nil
+	}
+	cci := *ci
+	cci.Jwt = _EMPTY_
+	cci.IssuerKey = _EMPTY_
+	return &cci
+}
+
 // ServerStats hold various statistics that we will periodically send out.
 type ServerStats struct {
 	Start              time.Time           `json:"start"`
--- a/server/jetstream_cluster.go
+++ b/server/jetstream_cluster.go
@@ -7297,23 +7297,29 @@ func (s *Server) jsClusteredMsgDeleteRequest(ci *ClientInfo, acc *Account, mset 
 }
 
 func encodeAddStreamAssignment(sa *streamAssignment) []byte {
+	csa := *sa
+	csa.Client = csa.Client.forProposal()
 	var bb bytes.Buffer
 	bb.WriteByte(byte(assignStreamOp))
-	json.NewEncoder(&bb).Encode(sa)
+	json.NewEncoder(&bb).Encode(csa)
 	return bb.Bytes()
 }
 
 func encodeUpdateStreamAssignment(sa *streamAssignment) []byte {
+	csa := *sa
+	csa.Client = csa.Client.forProposal()
 	var bb bytes.Buffer
 	bb.WriteByte(byte(updateStreamOp))
-	json.NewEncoder(&bb).Encode(sa)
+	json.NewEncoder(&bb).Encode(csa)
 	return bb.Bytes()
 }
 
 func encodeDeleteStreamAssignment(sa *streamAssignment) []byte {
+	csa := *sa
+	csa.Client = csa.Client.forProposal()
 	var bb bytes.Buffer
 	bb.WriteByte(byte(removeStreamOp))
-	json.NewEncoder(&bb).Encode(sa)
+	json.NewEncoder(&bb).Encode(csa)
 	return bb.Bytes()
 }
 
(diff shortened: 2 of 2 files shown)


server/events.go
```diff
@@ -336,6 +336,17 @@ func (ci *ClientInfo) forAssignmentSnap() *ClientInfo {
 	}
 }
 
+// forProposal returns the minimum amount of ClientInfo we need for assignment proposals.
+func (ci *ClientInfo) forProposal() *ClientInfo {
+	if ci == nil {
+		return nil
+	}
+	cci := *ci
```
derekcollison, line 344: If we are considering these at scale, should we only copy what we need vs copying everything and removing which makes the allocation and GC system work a bit harder? Should we also consider a pool for these?

server/jetstream_cluster.go
```diff
@@ -7295,23 +7295,29 @@ func (s *Server) jsClusteredMsgDeleteRequest(ci *ClientInfo, acc *Account, mset
 }
 
 func encodeAddStreamAssignment(sa *streamAssignment) []byte {
+	csa := *sa
```
derekcollison, line 7298: We are copying the fields here we will throw away, can we be a bit smarter here and avoid unecessary allocations and copies?

server/jetstream_cluster.go
```diff
@@ -7741,25 +7751,21 @@ func decodeConsumerAssignment(buf []byte) (*consumerAssignment, error) {
 }
 
 func encodeAddConsumerAssignmentCompressed(ca *consumerAssignment) []byte {
-	b, err := json.Marshal(ca)
-	if err != nil {
-		return nil
-	}
-	// TODO(dlc) - Streaming better approach here probably.
+	cca := *ca
```
derekcollison, line 7754: Saem here.

derekcollison (changes requested):

server/jetstream_cluster.go
neilalexander: In the case of `ca` and `sa`, all the fields are needed, the only one that gets overwritten is `Client`. We could pool them, since annoyingly every use of `json.Marshal(...)` or `json.NewEncoder(...).Encode(...)` will result in a heap escape...

derekcollison: ok let's attempt that change.

If we hold the js lock so that the sa or ca can not be changed, we could capture ci, nil it out of original before the csa := *sa and then put it back. Can also use the temp vairable in the call ro function to return stripped down version for proposals.

server/jetstream_cluster.go
```diff
@@ -7296,24 +7297,49 @@ func (s *Server) jsClusteredMsgDeleteRequest(ci *ClientInfo, acc *Account, mset
...
+// saPool pools stream assignment structs for encoding.
+var saPool = sync.Pool{
+	New: func() any {
+		return &streamAssignment{}
+	},
+}
+
+func saPoolPut(sa *streamAssignment) {
+	ciPool.Put(sa.Client)
+	sa.Client = nil
```
derekcollison, line 7309: Do we know js lock is held here?

server/jetstream_cluster.go
```diff
@@ -7296,24 +7297,49 @@ func (s *Server) jsClusteredMsgDeleteRequest(ci *ClientInfo, acc *Account, mset
...
 	bb.WriteByte(byte(assignStreamOp))
-	json.NewEncoder(&bb).Encode(sa)
+	json.NewEncoder(&bb).Encode(&csa)
 	return bb.Bytes()
 }
 
 func encodeUpdateStreamAssignment(sa *streamAssignment) []byte {
+	csa := saPool.Get().(*streamAssignment)
+	defer saPoolPut(csa)
+	*csa = *sa
```
derekcollison, line 7327: Maybe capture client info here in local variable from sa, and nil out before this line and put it back after?

server/jetstream_cluster.go
```diff
@@ -7722,17 +7748,38 @@ func (s *Server) jsClusteredConsumerRequest(ci *ClientInfo, acc *Account, subjec
...
+// caPool pools consumer assignment structs for encoding.
+var caPool = sync.Pool{
+	New: func() any {
+		return &consumerAssignment{}
+	},
+}
+
+func caPoolPut(ca *consumerAssignment) {
+	ciPool.Put(ca.Client)
+	ca.Client = nil
```
derekcollison, line 7760: Same here, do we know js write lock is held? Maybe add comment that the lock must be held?

server/jetstream_cluster.go
neilalexander: I don't think we need do, `csa` is a copy and therefore the `csa.Client = ...` line is operating on the copy. We don't modify the original SA.

server/jetstream_cluster.go
neilalexander: The only places `saPoolPut()` are called from are places where we do `encodeAddStream()` etc, so the JS lock is held. That said, only copies of the stream assignment for encoding are returned to the pool.

neilalexander:
For now I've pulled the pooling out, going to look at that separately.

derekcollison (approved):
LGTM

PR #6327, opened 2025-01-06
Refactor consumer `deleteNotActive`

neilalexander:
This PR makes two changes:

1. Instead of `deleteNotActive()` spawning more goroutines to clean up if the assignment is not present, we stay within the current goroutine to repeat the cleanup operation — this is safe because `deleteNotActive()` is only ever started via `time.AfterFunc()` and therefore callers are not blocked;
2. Once we receive confirmation that the consumer delete proposal has been accepted by the metaleader, stop generating new ones. This will reduce the amount of duplicated proposals in the metalayer WAL.

### DIFF
--- a/server/consumer.go
+++ b/server/consumer.go
@@ -1821,6 +1821,8 @@ var (
 	consumerNotActiveMaxInterval   = defaultConsumerNotActiveMaxInterval
 )
 
+// deleteNotActive should only be called via time.AfterFunc or in a goroutine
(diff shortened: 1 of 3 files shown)


neilalexander:
Going to ponder this one a bit more. Not sure if this is the right approach, as if a leader transfer happens then the proposal (even though it was accepted) may not be applied and then the cleanup may not happen as a result.

PR #6335, opened 2025-01-07
Optimise JetStream and Raft locking

neilalexander:
This PR optimises some uses of the JetStream lock by reducing interactions between that and Raft group locks, which should mean that we don't end up in situations where we hold the JS lock while waiting to take a Raft lock etc.

Also updates a couple of the Raft group helper functions to not take the lock as they are returning values that are never mutated, creating unnecessary contention. 


neilalexander:
(Please regular merge rather than squash merge, I'd like these changes to stay in the commit log for individual blames.)

derekcollison (approved):
LGTM

PR #6341, opened 2025-01-07
Only encode & send advisories when there is interest

neilalexander:
This stops us from spending time & CPU cycles encoding advisory JSONs when there is no one listening for them.


server/consumer.go
```diff
@@ -1588,8 +1588,26 @@ func (o *consumer) unsubscribe(sub *subscription) {
...
+	if o.acc == nil {
+		return
+	}
+
+	// If there is no one listening for this advisory then save ourselves the effort
+	// and don't bother encoding the JSON or sending it.
+	o.acc.mu.RLock()
+	sl := o.acc.sl
+	o.acc.mu.RUnlock()
+	if !sl.HasInterest(subject) && !o.srv.hasGatewayInterest(o.acc.Name, subject) {
```
derekcollison, line 1601: Prefer us to hook updates for this somehow vs checking each time. Also if we know o.acc is not nil, can it go back to nil? Meaning safe to reference without a lock?

derekcollison: Maybe not too bad check each time..

server/consumer.go
```diff
@@ -1605,13 +1623,8 @@ func (o *consumer) sendDeleteAdvisoryLocked() {
 		Domain:   o.srv.getOpts().JetStreamDomain,
 	}
 
-	j, err := json.Marshal(e)
```
derekcollison, line 1608: I think I did this here on purpose..

server/consumer.go
neilalexander: I've checked and don't think `o.acc` can go back to being nil, so I've removed the lock. I don't think the check for interest is expensive enough to worry about for now given that we now have `HasInterest` which is non-allocating.

server/consumer.go
neilalexander: I'm not sure it matters where the marshaling takes place, but if we can avoid doing it until after the interest check, that saves us some CPU time & allocs.

derekcollison: We should do it after the check for sure. But I need to recall why I put it there? Maybe use more cores?

derekcollison (approved):
LGTM

PR #6344, opened 2025-01-08
Do not start additional goroutines from consumer `deleteNotActive`

neilalexander:
Since `deleteNotActive` already runs in its own goroutine via `time.AfterFunc`, we just create more work for the scheduler by having extra goroutines running on top of that.


derekcollison (approved):
LGTM

PR #6347, opened 2025-01-08
Fix panic in `isGroupLeaderless`

neilalexander:
Need to capture `rg.node` while the JetStream lock is held, otherwise stopping an asset could race and it could be set back to `nil`.


wallyqs (approved):
LGTM

PR #6348, opened 2025-01-09
Updates to TestJetStreamClusterStreamOrphanMsgsAndReplicasDrifting

wallyqs:
- Warn when stream subjects and num pending counts in consumers differs but there is no stream drift
- Log the subjects and consumers maps when there is drift


derekcollison (approved):
LGTM

PR #6351, opened 2025-01-09
Ensure `deleteNotActive` stops on JetStream or server shutdown

neilalexander:
This ensures that we stop these goroutines more quickly when either the server or the JetStream system shuts down.

This also avoids a race in `TestJetStreamClusterGhostEphemeralsAfterRestart`, as the unit test was reverting a modified constant before these goroutines would have finished in some cases.

### DIFF
--- a/server/consumer.go
+++ b/server/consumer.go
@@ -1853,6 +1853,13 @@ func (o *consumer) deleteNotActive() {
 
 	s, js := o.mset.srv, o.srv.js.Load()
 	acc, stream, name, isDirect := o.acc.Name, o.stream, o.name, o.cfg.Direct
+	var qch, cqch chan struct{}
+	if o.srv != nil {
+		qch = o.srv.quitCh
+	}
+	if o.js != nil {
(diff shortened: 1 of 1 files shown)


server/consumer.go
```diff
@@ -1891,7 +1891,14 @@ func (o *consumer) deleteNotActive() {
 			interval := consumerNotActiveStartInterval + jitter
 			ticker := time.NewTicker(interval)
 			defer ticker.Stop()
-			for range ticker.C {
+			for {
+				select {
+				case <-ticker.C:
+				case <-o.js.clusterQuitC():
+					return
+				case <-o.srv.quitCh:
```
wallyqs, line 1899: could o.srv or o.js be nil here while attempting the shutdown?

neilalexander: As best as I can tell, neither `o.srv` or `o.js` can go back to being `nil`, but I can add extra guards just to be sure.

derekcollison (approved):
LGTM

PR #6354, opened 2025-01-09
Don't run TTL state recovery if no messages

neilalexander:
This avoids an unnecessary `TTL state is outdated` message in the log when there is nothing to do.


derekcollison (approved):
LGTM

PR #6355, opened 2025-01-09
Use `popOne` in JS API routed request workers

neilalexander:
This updates the JS API workers to use `popOne`, so that if the queue stacks up at all, the recovery is fairer across workers, rather than single workers stealing the entire queue.

This was extracted from #6342.

### DIFF
--- a/server/jetstream_api.go
+++ b/server/jetstream_api.go
@@ -922,8 +922,10 @@ func (s *Server) processJSAPIRoutedRequests() {
 	for {
 		select {
 		case <-queue.ch:
-			reqs := queue.pop()
-			for _, r := range reqs {
+			// Only pop one item at a time here, otherwise if the system is recovering
+			// from queue buildup, then one worker will pull off all the tasks and the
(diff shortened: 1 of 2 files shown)


server/jetstream_api.go
```diff
@@ -932,7 +934,6 @@ func (s *Server) processJSAPIRoutedRequests() {
 				}
 				atomic.AddInt64(&js.apiInflight, -1)
 			}
-			queue.recycle(&reqs)
```
derekcollison, line 935: Do we lose recycling these with PopOne all together?

neilalexander: Nope, the `popOne` function manages it when the slice is emptied.

derekcollison (approved):
LGTM

neilalexander:
Rebased to include a race condition fix.

derekcollison:
I am just waiting on green run to merge.

derekcollison (approved):
LGTM

PR #6356, opened 2025-01-09
Fix race condition between server shutdown and consumer info

neilalexander:


wallyqs (approved):
LGTM

wallyqs (approved):
LGTM

PR #6361, opened 2025-01-10
Error handling in `metaSnapshot`

neilalexander:
Since we use `MarshalJSON()` on some types, we must be careful that a JSON marshalling error cannot be dropped.


MauriceVanVeen (approved):
LGTM

wallyqs (approved):
LGTM

PR #6363, opened 2025-01-10
Per-message TTLs set metadata API level to `1`

neilalexander:
As per the ADR:

> When either these settings are set the Stream should require API level `1`.


ripienaar (approved):
LGTM

PR #6370, opened 2025-01-13
Reject TTL messages when not enabled on stream

neilalexander:
Factored out the rejection behaviour from #6298 so that it can be reviewed/merged separately from the sourcing/mirroring.


server/jetstream_cluster.go
```diff
@@ -8048,6 +8048,18 @@ func (mset *stream) processClusteredInboundMsg(subject, reply string, hdr, msg [
 			mset.storeMsgIdLocked(&ddentry{msgId, 0, time.Now().UnixNano()})
 			mset.mu.Unlock()
 		}
+
+		// TTL'd messages are rejected entirely if TTLs are not enabled on the stream.
+		if ttl, _ := getMessageTTL(hdr); ttl != 0 && !allowTTL {
+			if canRespond {
+				var resp = &JSPubAckResponse{PubAck: &PubAck{Stream: name}}
+				resp.PubAck = &PubAck{Stream: name}
```
derekcollison, line 8056: You set this above already. I think you can remove this line.

derekcollison (changes requested):

server/jetstream_cluster.go
neilalexander: Oops, yes. Fixed.

derekcollison (approved):
LGTM

PR #6373, opened 2025-01-14
Fix JS API in-flight metric

neilalexander:
After a drain this would have been misreporting, as we did not remove drained entries from the `apiInflight` count.


derekcollison (approved):
LGTM

PR #6377, opened 2025-01-16
[FIXED] LeafNodes: Queue interest may be lost in super cluster

kozlovic:
If a cluster has leafnode connections and each have the same queue group, the loss of a leafnode connection could cause the server in the "hub" cluster to drop interest across a gateway for this queue group.

The issue is fixed by properly accounting for all queue sub and unsub in the server gateway interest map.


derekcollison (approved):
LGTM

PR #6378, opened 2025-01-16
Add support for subject delete markers by `MaxAge`

neilalexander:
This PR adds the `LimitsTTL` functionality for system tombstones when a subject is removed via `MaxAge`. The tombstone will have the header `Nats-Applied-Limit` and the specified `LimitsTTL` value as the TTL.

Further limits and/or administrative actions to follow separately.

This also requires API version 1, cannot be set if TTLs are disabled and cannot be combined with a mirror configuration.

### DIFF
--- a/server/filestore.go
+++ b/server/filestore.go
@@ -174,6 +174,7 @@ type fileStore struct {
 	tombs       []uint64
 	ld          *LostStreamData
 	scb         StorageUpdateHandler
+	sdmcb       SubjectDeleteMarkerUpdateHandler
 	ageChk      *time.Timer
 	syncTmr     *time.Timer
 	cfg         FileStreamInfo
@@ -2131,7 +2132,11 @@ func (fs *fileStore) expireMsgsOnRecover() {
 			break
 		}
 		// Can we remove whole block here?
-		if mb.last.ts <= minAge {
+		// TODO(nat): We can't do this with LimitsTTL as we have no way to know
+		// if we're throwing away real messages or other tombstones without
+		// loading them, so in this case we'll fall through to the "slow path".
+		// There might be a better way of handling this though.
+		if !mb.fs.cfg.SubjectDeleteMarkers && mb.last.ts <= minAge {
 			purged += mb.msgs
 			bytes += mb.bytes
 			deleteEmptyBlock(mb)
@@ -2198,7 +2203,17 @@ func (fs *fileStore) expireMsgsOnRecover() {
 			// Update fss
 			// Make sure we have fss loaded.
 			mb.removeSeqPerSubject(sm.subj, seq)
-			fs.removePerSubject(sm.subj)
+			if fs.removePerSubject(sm.subj) && fs.cfg.SubjectDeleteMarkers {
+				// Need to release the mb lock here in case we need to write a new
+				// tombstone into the same mb in subjectDeleteMarkerIfNeeded. However
+				// at this point fs.mu is held, so nothing else should happen here.
+				// No need to process the callbacks from subjectDeleteMarkerIfNeeded
+				// here as none will have been registered yet (we haven't yet returned
+				// from newFileStore*).
+				mb.mu.Unlock()
+				fs.subjectDeleteMarkerIfNeeded(sm, JSAppliedLimitMaxAge)
+				mb.mu.Lock()
+			}
 		}
 		// Make sure we have a proper next first sequence.
 		if needNextFirst {
@@ -3612,6 +3627,13 @@ func (fs *fileStore) RegisterStorageUpdates(cb StorageUpdateHandler) {
 	}
 }
 
+// RegisterSubjectDeleteMarkerUpdates registers a callback for updates to new tombstones.
+func (fs *fileStore) RegisterSubjectDeleteMarkerUpdates(cb SubjectDeleteMarkerUpdateHandler) {
+	fs.mu.Lock()
+	fs.sdmcb = cb
+	fs.mu.Unlock()
+}
+
 // Helper to get hash key for specific message block.
 // Lock should be held
 func (fs *fileStore) hashKeyForBlock(index uint32) []byte {
@@ -4325,10 +4347,10 @@ func (fs *fileStore) EraseMsg(seq uint64) (bool, error) {
 }
 
 // Convenience function to remove per subject tracking at the filestore level.
-// Lock should be held.
-func (fs *fileStore) removePerSubject(subj string) {
+// Lock should be held. Returns if we deleted the last message on the subject.
+func (fs *fileStore) removePerSubject(subj string) bool {
 	if len(subj) == 0 || fs.psim == nil {
-		return
+		return false
 	}
 	// We do not update sense of fblk here but will do so when we resolve during lookup.
 	bsubj := stringToBytes(subj)
@@ -4339,9 +4361,11 @@ func (fs *fileStore) removePerSubject(subj string) {
 		} else if info.total == 0 {
 			if _, ok = fs.psim.Delete(bsubj); ok {
 				fs.tsl -= len(subj)
+				return true
 			}
 		}
 	}
+	return false
 }
 
 // Remove a message, optionally rewriting the mb file.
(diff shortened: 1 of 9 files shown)


server/filestore.go
```diff
@@ -2131,7 +2132,11 @@ func (fs *fileStore) expireMsgsOnRecover() {
 			break
 		}
 		// Can we remove whole block here?
-		if mb.last.ts <= minAge {
+		// TODO(nat): We can't do this with LimitsTTL as we have no way to know
+		// if we're throwing away real messages or other tombstones without
+		// loading them, so in this case we'll fall through to the "slow path".
+		// There might be a better way of handling this though.
+		if mb.fs.cfg.LimitsTTL <= 0 && mb.last.ts <= minAge {
```
derekcollison, line 2139: LimitsTTL too limiting IMO. Especially if we expand to purge and purgeEx and maybe normal user msg deletes.

server/filestore.go
```diff
@@ -2198,7 +2203,11 @@ func (fs *fileStore) expireMsgsOnRecover() {
 			// Update fss
 			// Make sure we have fss loaded.
 			mb.removeSeqPerSubject(sm.subj, seq)
-			fs.removePerSubject(sm.subj)
+			if fs.removePerSubject(sm.subj) {
+				mb.mu.Unlock()
```
derekcollison, line 2207: A bit nervous here releasing and re-acquiring the mb lock. Also we do not need to do this unless the stream is configured to want tombstones yes? Might be good to avoid this all together if not configured for it.

server/filestore.go
```diff
@@ -5295,6 +5313,45 @@ func (fs *fileStore) cancelAgeChk() {
...
+// system tombstone. If the tombstone is written successfully then
+// this function returns a callback func to call scb and tcb after the
+// lock has been released.
+func (fs *fileStore) systemTombstoneIfNeeded(sm *StoreMsg, reason string) func() {
+	// If the deleted message was itself a tombstone then don't
+	// write out more tombstones or we'll churn endlessly.
+	if len(getHeader(JSAppliedLimit, sm.hdr)) != 0 {
+		return nil
+	}
+	if !fs.cfg.AllowMsgTTL || fs.cfg.LimitsTTL <= 0 {
```
derekcollison, line 5327: Be better to do this check in upper sections and not do the unlock / lock in removeMsg.

server/filestore.go
```diff
@@ -5295,6 +5313,45 @@ func (fs *fileStore) cancelAgeChk() {
...
+		return nil
+	}
+	if _, ok := fs.psim.Find(stringToBytes(sm.subj)); ok {
+		// There are still messages left with this subject,
+		// therefore it wasn't the last message deleted.
+		return nil
+	}
+	// Build the system tombstone.
+	var _hdr [128]byte
+	hdr := fmt.Appendf(_hdr[:0], "NATS/1.0\r\n%s: %s\r\n%s: %s\r\n\r\n", JSAppliedLimit, reason, JSMessageTTL, fs.cfg.LimitsTTL)
```
derekcollison, line 5337: LimitsTTL seems like wrong name IMO.

server/filestore.go
```diff
@@ -2198,7 +2203,11 @@ func (fs *fileStore) expireMsgsOnRecover() {
 			// Update fss
 			// Make sure we have fss loaded.
 			mb.removeSeqPerSubject(sm.subj, seq)
-			fs.removePerSubject(sm.subj)
+			if fs.removePerSubject(sm.subj) {
+				mb.mu.Unlock()
+				fs.systemTombstoneIfNeeded(sm, "MaxAge")
```
derekcollison, line 2208: Should we make MaxAge and others const?

server/filestore.go
```diff
@@ -5295,6 +5313,45 @@ func (fs *fileStore) cancelAgeChk() {
...
+	var _hdr [128]byte
+	hdr := fmt.Appendf(_hdr[:0], "NATS/1.0\r\n%s: %s\r\n%s: %s\r\n\r\n", JSAppliedLimit, reason, JSMessageTTL, fs.cfg.LimitsTTL)
+	seq, ts, ttl := fs.state.LastSeq+1, time.Now().UnixNano(), int64(fs.cfg.LimitsTTL.Seconds())
+	// Store it in the stream and then prepare the callbacks
+	// to return to the caller.
+	if err := fs.storeRawMsg(sm.subj, hdr, nil, seq, ts, ttl); err != nil {
+		return nil
+	}
+	cb, tcb := fs.scb, fs.tcb
+	return func() {
```
derekcollison, line 5345: Why are we returning a func here vs just calling inline?

server/filestore.go
derekcollison, line 2208: Why do we ignore the func return param here?

derekcollison (changes requested):

server/filestore.go
neilalexander: Best that @ripieneer has input here as this is the name from the ADR.

server/filestore.go
neilalexander: Yep, absolutely can.

server/filestore.go
neilalexander: If we don't release the lock then `storeRawMsg` deadlocks if it is writing the tombstone into the same block, so not sure what the best way around that is at the moment. At this point though the overall filestore lock is held so not sure it matters.

server/filestore.go
neilalexander: The filestore generally seems to avoid holding the FS lock while calling the storage update handler, I assumed there was a reason for it, so I've just tried to recreate that pattern by passing back a function that the caller can call when it has released the lock.

If we don't care about holding the lock while the callbacks are called then I can remove this and make it inline.

server/filestore.go
neilalexander: Best that @ripienaar has input here as this is the name from the ADR.

ripienaar: When limits are applied and tombstones are written, this is the TTL of those messages.

We didnt consider purges would be involved until later, do you have a name suggestion @derekcollison ?

derekcollison: Naming is hard, so not at the moment but will give it some thought.

server/filestore.go
derekcollison: ok got it..

server/filestore.go
ripienaar: something like `ActionTTL` or `NotificationTTL`? Other ideas? Failing to find a better name :)

neilalexander: Is it weird that the functionality is only turned on with a TTL? Do we want something like `LimitsTombstones bool` and `LimitsTombstonesTTL time.Duration`? What if you want tombstones but you want them to "live forever"?

ripienaar: could set LimitsTTL to `never`? Just seemed excessive to add 2 fields

neilalexander: It feels weird still, if it's `LimitsTombstones` then `never` doesn't make sense for something that *does* leave TTLs, and if it's `LimitsTTL` or `LimitsTombstonesTTL` then it's difficult to explain why the option is named after TTLs when it has a secondary larger effect.

ripienaar: We can add 2 fields, I am not super stuck to the idea or anything, just conscious of how many config options we have, though I agree making it awkward is also not good

derekcollison: I think the turning on should be a bool that is clear we will generate markers when a subject disappears, and that should have a default TTL. We can have a second that allows them to override or change the TTL for he markers, including a "never" if needed.

server/filestore.go
derekcollison: At least we should place a comment if we do mean to ignore returned func and not call it.

server/filestore.go
jnmoyne: How about `tombstoneTTL`?

neilalexander: We've gone with `SubjectDeleteMarkers bool` and `SubjectDeleteMarkerTTL string`.

The default if not specified will be 15 minutes if not specified, RI and I think this is a good value as it's well outside the recovery time for fetch/ordered consumer/client/route/gateway etc failures as well as even application restarts.

derekcollison:
This ready for a second review?

derekcollison (approved):
LGTM

PR #6385, opened 2025-01-20
Correctly handle TTLs from upper-layer stream catchups

neilalexander:
This ensures that messages received from an upper-layer catchup also have their TTLs added to the timed hash wheel correctly on the follower.

(Also updates `require_NotNil` to be variadic because it looks nicer in the test.)


derekcollison (approved):
LGTM

PR #6386, opened 2025-01-20
fixed unclosed tmp file handle in filestore.go

deem0n:
Static analyzer found unclosed filehandle tmpFD in case of errors. It is not strictly necessary as GC will eventually close tmpFD, but explicit call of Close is always better than magic finalizer.

Based on the [discussion on SO](https://stackoverflow.com/questions/58351084/does-go-automatically-close-resources-if-not-explicitly-closed)

Signed-off-by: Dmitry Dorofeev <dima@yasp.com>

### DIFF
--- a/server/filestore.go
+++ b/server/filestore.go
@@ -5838,8 +5838,15 @@ func (mb *msgBlock) recompressOnDiskIfNeeded() error {
 	<-dios
 	tmpFD, err := os.OpenFile(tmpFN, os.O_CREATE|os.O_TRUNC|os.O_WRONLY, defaultFilePerms)
 	dios <- struct{}{}
+
 	if err != nil {
 		return fmt.Errorf("failed to create temporary file: %w", err)
+	}
+
+	errorCleanup := func(err error) error {
+		tmpFD.Close()
+		os.Remove(tmpFN)
+		return err
 	}
 
 	// The original buffer at this point is uncompressed, so we will now compress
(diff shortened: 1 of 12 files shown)


neilalexander (approved):
LGTM

neilalexander:
FYI CI complaining about `server/filestore.go:5841: File is not goimports-ed (goimports)`.

deem0n:
> FYI CI complaining about `server/filestore.go:5841: File is not goimports-ed (goimports)`.

checked locally with linter, it seems that I touched empty line:

```
diff --git a/server/filestore.go b/server/filestore.go
index 7b6df2ee..19ddd278 100644
--- a/server/filestore.go
+++ b/server/filestore.go
@@ -5762,7 +5762,7 @@ func (mb *msgBlock) recompressOnDiskIfNeeded() error {
        <-dios
        tmpFD, err := os.OpenFile(tmpFN, os.O_CREATE|os.O_TRUNC|os.O_WRONLY, defaultFilePerms)
        dios <- struct{}{}
-
+
        if err != nil {
                return fmt.Errorf("failed to create temporary file: %w", err)
        }
```

Should I fix it with new commit to this PR?

neilalexander:
Please, although if you can squash down to a single commit, that'd be even better.

neilalexander:
I think the reason you have `main` commits showing (and GH showing the branch out of date) is because you need to rebase against the latest `main` branch.

deem0n:
> I think the reason you have `main` commits showing (and GH showing the branch out of date) is because you need to rebase against the latest `main` branch.

I guess it will be easier to make completely new PR with just one commit, my git-fu is not that good :-)

deem0n:
Reopened as #6390

PR #6389, opened 2025-01-21
Fix subject delete marker stream config validation

neilalexander:
This fixes an issue where the stream config would fail to validate empty stream delete marker TTLs.


derekcollison (approved):
LGTM

PR #6390, opened 2025-01-21
fixed unclosed tmp file handle in filestore.go

deem0n:
Static analyzer found unclosed filehandle tmpFD in case of errors. It is not strictly necessary as GC will eventually close tmpFD, but explicit call of Close is always better than magic finalizer.

Based on the [SO discussion](https://stackoverflow.com/questions/58351084/does-go-automatically-close-resources-if-not-explicitly-closed)

Signed-off-by: Dmitry Dorofeev <dima@yasp.com>


neilalexander (approved):
LGTM, thanks!

PR #6393, opened 2025-01-21
Place default subject delete marker TTL into stream config

neilalexander:
This avoids problems where we have "hidden" defaults. This also replaces a test from the filestore with one in JetStream, and ensures pedantic mode also does the right thing.


derekcollison (approved):
LGTM

PR #6400, opened 2025-01-23
Add support for TTLs and subject delete markers to memory store

neilalexander:
This adds support for TTLs and subject delete markers into the memory store. Much of this is just replicating and aligning the logic from the filestore. None of the recovery/rebuilding/re-indexing code is needed as the memory store does not recover from disk.

Some of the relevant JetStream unit tests have been updated to test both `MemoryStorage` and `FileStorage`. The relevant filestore tests have also been replicated into the memstore tests. 


derekcollison (approved):
LGTM

PR #6404, opened 2025-01-24
Change `SubjectDeleteMarkerTTL` stream config field to duration

neilalexander:
After experimenting with this, it feels strange to have `MaxAge` etc typed as `time.Duration` but then have `SubjectDeleteMarkerTTL` typed as a maybe-string-integer maybe-string-duration. This PR instead standardises on `time.Duration`.

This does not change the behaviour of the `Nats-TTL` header.


derekcollison (approved):
LGTM

PR #6408, opened 2025-01-26
Improve error message when attempting to change consumer type

piotrpio:
Before this change, an attempt to change consumer type (by either setting or removing deliver subject), resulted in `max waiting can not be updated` error as `MaxWaiting` has a default value for pull consumers. This ensures that changing consumer type is checked before max waiting to return a more useful error message.

### DIFF
--- a/server/jetstream_test.go
+++ b/server/jetstream_test.go
@@ -9279,6 +9279,47 @@ func TestJetStreamPullConsumerMaxWaiting(t *testing.T) {
 	}
 }
 
+func TestJetStreamChangeConsumerType(t *testing.T) {
+	s := RunBasicJetStreamServer(t)
+	defer s.Shutdown()
+
+	nc, js := jsClientConnect(t, s)
+	defer nc.Close()
+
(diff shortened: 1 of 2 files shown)


server/jetstream_test.go
```diff
@@ -9279,6 +9279,51 @@ func TestJetStreamPullConsumerMaxWaiting(t *testing.T) {
...
+	})
+	require_NoError(t, err)
+
+	// cannot update pull -> push
+	_, err = js.UpdateConsumer("TEST", &nats.ConsumerConfig{
+		Name:           "pull",
+		AckPolicy:      nats.AckExplicitPolicy,
+		DeliverSubject: "foo",
+	})
+	if !strings.Contains(err.Error(), "can not update pull consumer to push based") {
```
neilalexander, line 9305: Minor nit but assertions like `require_Equals(err.Error(), "can not update pull consumer to push based")` are better not just because they are cleaner but because they properly integrate with the Antithesis SDK assertions.

piotrpio: done, thanks 👍

neilalexander (approved):
LGTM

PR #6412, opened 2025-01-27
Remove the comment that durable_name is deprecated

ripienaar:
We intended to deprecate this but did not do so in the end and it is critical in current API clients


neilalexander (approved):
LGTM

PR #6414, opened 2025-01-27
Descriptive stream & consumer health errors

neilalexander:
This should make the output of `healthz` less opaque when there are stream or consumer healthcheck errors. Also more closely aligns the checks between streams and consumers.


derekcollison:
Safe to ignore?

PR #6423, opened 2025-01-28
Add router and gateway connections to /connz

alexbozhenko:


wallyqs:
there is `/gatewayz` and `/routez` which get this info instead

alexbozhenko:
@wallyqs I had this in draft for a reason: I wanted to check if it would work, and if it will give additional information.
Is there a way to find info on slow gateway and route consumers
https://github.com/nats-io/nats-server/pull/6422
?

PR #6428, opened 2025-01-29
Subject delete markers on purges and removes for filestore

neilalexander:
This PR adds subject delete markers when doing purge/compact/remove operations for the filestore.

Memstore will be a separate PR but expect it to look a lot like this one.

Remaining questions before I mark for review:

- [ ] Are we happy with the approach?
- [ ] Does the name `Nats-Applied-Limit` make sense or do we want a more generic name like `Nats-Marker-Reason`?
- [ ] ... or do we want a separate header for administrative reasons like `Purge` and `Remove`?

### DIFF
--- a/server/filestore.go
+++ b/server/filestore.go
@@ -202,6 +202,7 @@ type fileStore struct {
 	firstMoved  bool
 	ttls        *thw.HashWheel
 	ttlseq      uint64 // How up-to-date is the `ttls` THW?
+	markers     []string
 }
 
 // Represents a message store block and its data.
@@ -2119,7 +2120,7 @@ func (fs *fileStore) expireMsgsOnRecover() error {
 		mb.fss.Iter(func(bsubj []byte, ss *SimpleState) bool {
 			subj := bytesToString(bsubj)
 			for i := uint64(0); i < ss.Msgs; i++ {
-				fs.removePerSubject(subj)
+				fs.removePerSubject(subj, false)
 			}
 			return true
 		})
@@ -2213,17 +2214,7 @@ func (fs *fileStore) expireMsgsOnRecover() error {
 			// Update fss
 			// Make sure we have fss loaded.
 			mb.removeSeqPerSubject(sm.subj, seq)
-			if fs.removePerSubject(sm.subj) && fs.cfg.SubjectDeleteMarkers {
-				// Need to release the mb lock here in case we need to write a new
-				// tombstone into the same mb in subjectDeleteMarkerIfNeeded. However
-				// at this point fs.mu is held, so nothing else should happen here.
-				// No need to process the callbacks from subjectDeleteMarkerIfNeeded
-				// here as none will have been registered yet (we haven't yet returned
-				// from newFileStore*).
-				mb.mu.Unlock()
-				fs.subjectDeleteMarkerIfNeeded(sm, JSAppliedLimitMaxAge)
-				mb.mu.Lock()
-			}
+			fs.removePerSubject(sm.subj, fs.cfg.SubjectDeleteMarkers && len(getHeader(JSMarkerReason, sm.hdr)) == 0)
 		}
 		// Make sure we have a proper next first sequence.
 		if needNextFirst {
@@ -2279,6 +2270,9 @@ func (fs *fileStore) expireMsgsOnRecover() error {
 		// Clear any global subject state.
 		fs.psim, fs.tsl = fs.psim.Empty(), 0
 	}
+
+	// If we have pending markers, then create them.
+	fs.subjectDeleteMarkersAfterOperation(JSMarkerReasonMaxAge)
 
 	// If we purged anything, make sure we kick flush state loop.
 	if purged > 0 {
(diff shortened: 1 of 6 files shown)


server/filestore.go
```diff
@@ -202,6 +202,7 @@ type fileStore struct {
 	firstMoved  bool
 	ttls        *thw.HashWheel
 	ttlseq      uint64 // How up-to-date is the `ttls` THW?
+	markers     *stree.SubjectTree[struct{}]
```
derekcollison, line 205: This could be heavy for lighter use cases, but understand why this might be more applicable the `[]string` but I would start with that tbh.

server/filestore.go
```diff
@@ -406,15 +407,16 @@ func newFileStoreWithCreated(fcfg FileStoreConfig, cfg StreamConfig, created tim
...
+		fcfg:    fcfg,
+		psim:    stree.NewSubjectTree[psi](),
+		bim:     make(map[uint32]*msgBlock),
+		cfg:     FileStreamInfo{Created: created, StreamConfig: cfg},
+		prf:     prf,
+		oldprf:  oldprf,
+		qch:     make(chan struct{}),
+		fsld:    make(chan struct{}),
+		srv:     fcfg.srv,
+		markers: stree.NewSubjectTree[struct{}](),
```
derekcollison, line 419: If []string can leave without allocating and just use append.

server/filestore.go
```diff
@@ -4491,7 +4495,7 @@ func (fs *fileStore) removeMsg(seq uint64, secure, viaLimits, needFSLock bool) (
 
 	// If we are tracking multiple subjects here make sure we update that accounting.
 	mb.removeSeqPerSubject(sm.subj, seq)
-	fs.removePerSubject(sm.subj)
+	wasLast := fs.removePerSubject(sm.subj)
```
derekcollison, line 4498: Since removePerSubject is fs scoped can just do this inside that function IMO..

server/filestore.go
```diff
@@ -5365,16 +5380,38 @@ func (fs *fileStore) subjectDeleteMarkerIfNeeded(sm *StoreMsg, reason string) fu
...
+}
+
+// Filestore lock must be held but message block locks must not be.
+// The caller should call the callback, if non-nil, after releasing
+// the filestore lock.
+func (fs *fileStore) subjectDeleteMarkersAfterPurge() func() {
+	if !fs.cfg.SubjectDeleteMarkers || fs.markers.Size() == 0 {
+		return nil
+	}
+	cbs := make([]func(), 0, fs.markers.Size())
```
derekcollison, line 5404: Why the cbs versus just doing inline? Locking things?

server/filestore.go
```diff
@@ -7542,7 +7575,9 @@ func (fs *fileStore) PurgeEx(subject string, sequence, keep uint64) (purged uint
 				}
 				// PSIM and FSS updates.
 				mb.removeSeqPerSubject(sm.subj, seq)
-				fs.removePerSubject(sm.subj)
+				if fs.removePerSubject(sm.subj) && fs.cfg.SubjectDeleteMarkers {
```
derekcollison, line 7578: I still might consider burying this inside removePerSubject.

server/filestore.go
neilalexander: I did try that, but there's a couple cases where it doesn't make sense, i.e. in `expireMsgs` and `expireMsgsOnRecover`, we want to first make sure that the deleted message isn't itself a delete marker (so that we don't end up making more of them endlessly), but we still need to do `removePerSubject` regardless.

server/filestore.go
neilalexander: Yes, so we aren't holding the FS locks while doing callbacks.

server/filestore.go
neilalexander: Yes, is rather a premature optimisation I suppose. I can change it to `[]string` for now.

server/filestore.go
derekcollison: But that can all be in removePerSubject() if we pass in bool yes that indicates if this is a marker?

neilalexander: Have updated it so we pass in a bool on whether or not to queue a marker.

neilalexander:
Have renamed the `Nats-Applied-Limit` header to `Nats-Marker-Reason` with @ripienaar's agreement so that it also makes sense for the `"Purge"` and `"Remove"` reasons.

derekcollison (commented):
In general LGTM -small question on possible perf issue for non-marker configured streams.

server/filestore.go
```diff
@@ -2213,17 +2214,7 @@ func (fs *fileStore) expireMsgsOnRecover() error {
...
-				// tombstone into the same mb in subjectDeleteMarkerIfNeeded. However
-				// at this point fs.mu is held, so nothing else should happen here.
-				// No need to process the callbacks from subjectDeleteMarkerIfNeeded
-				// here as none will have been registered yet (we haven't yet returned
-				// from newFileStore*).
-				mb.mu.Unlock()
-				fs.subjectDeleteMarkerIfNeeded(sm, JSAppliedLimitMaxAge)
-				mb.mu.Lock()
-			}
+			fs.removePerSubject(sm.subj, len(getHeader(JSMarkerReason, sm.hdr)) == 0)
```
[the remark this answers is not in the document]
neilalexander: It might be narrowly, so I've uplifted the config check up as you suggested.

derekcollison (approved):
LGTM

PR #6432, opened 2025-01-30
Single configuration option for subject delete markers

neilalexander:
Remove the boolean and the default value and preserve only the single TTL configuration option. We also no longer need the pedantic checks anymore either.


derekcollison (approved):
LGTM

PR #6433, opened 2025-01-30
Subject delete markers on purges and removes for memstore

neilalexander:
Like #6248 but this time for the memstore.


derekcollison (approved):
LGTM

Issue #6434, opened 2025-01-30
Introduce a new CUSTOM_LOOKUP resolver


ghstahl:
### Proposed change

The goal is that all accounts and users are handled externally and that the nats configuration started with static accounts.  
The creation of accounts is done on demand as part of the authoriztion.auth_callout flows.

The starting point.
```
host: 0.0.0.0
port: 4222
http_port: 8222

websocket {
  port: 8080
  no_tls: true
}

leafnodes {
  port: 7422
}

server_name: $SERVER_NAME
system_account: SYS

jetstream {
	store_dir: "/nats"
}

accounts: {
    SYS: {

    }
    AUTH: {
      users: [
          {user: auth, password: auth}
      ]
    }

}

authorization {
  # Docs 
  # https://github.com/nats-io/nats.docs/blob/master/running-a-nats-service/configuration/securing_nats/auth_callout.md#auth-callout
  # Example of implementing a custom auth server
  # https://github.com/ConnectEverything/nats-by-example/blob/main/examples/auth/callout/cli/service/main.go
  auth_callout {
    # Generated by running
    # nsc generate nkey --account
    # use this private key to sign JWTs:
    # SAAEXFSYMLINXLKR2TG5FLHCJHLU62B3SK3ESZLGP4B4XGLUNXICW3LGAY
    issuer: "ABNXLUXZG427NJ4YLMDR6HYLWCBHEOMN4TQGYGS2T5XNRUK5Y6ZG6XJ3"
    auth_users: [ auth ]
    account: AUTH
    # Generated by running
    # nsc generate nkey --curve
    # use this private key to encrypt/decrypt traffic between auth service & NATS: 
    # SXAMKSXEE3LCBT4NNMKGEDFRGGO4DDIPO5JQSPW6W5MHLZDMG6N2SKB2ZI
    xkey: "XCND2ELXRACFDAD7CFHXHZE7QPSEHW5IKNLPM5Y2FVFS7PHU6NUDMHKR"
  }
}
# configuration of the nats based resolver

resolver {
    type: CUSTOM_LOOKUP

    # Directory in which account jwt will be stored
    dir: './jwt'
    # In order to support jwt deletion, set to true
    # If the resolver type is full delete will rename the jwt.
    # This is to allow manual restoration in case of inadvertent deletion.
    # To restore a jwt, remove the added suffix .delete and restart or send a reload signal.
    # To free up storage you must manually delete files with the suffix .delete.
    allow_delete: false
[...]

PR #6435, opened 2025-01-30
Introduce new CUSTOM_LOOKUP and CUSTOM_LOOKUP_FULL Account Resolvers

ghstahl:
[issue](https://github.com/nats-io/nats-server/issues/6434)

### DIFF
--- a/server/account_resolver_custom.go
+++ b/server/account_resolver_custom.go
@@ -18,6 +18,44 @@ import (
 	"time"
 )
 
+// CustomLookupCacheDirAccResolver is based on the FULL CacheDirAccResolver and will lookup the account every time by sending a message to $SYS.REQ.ACCOUNT.*.CLAIMS.LOOKUP
+type CustomLookupCacheDirAccResolver struct {
+	*CacheDirAccResolver
+}
+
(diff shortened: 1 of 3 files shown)


aricart:
I made a proof of concept of implementing dynamic account creation using the callout feature. That would be a simpler mechanism without having to modify the server. 

https://github.com/aricart/callout.go/pull/11

shaunco:
@aricart - Thanks for the example, but requiring every single IoT device in our environment to be pre-seeded with a `sentinel.creds` file, and potentially having 100k+ leaf-nodes drop offline if that file ever changes, seems extremely impractical. Why can't that just be specified in the cluster config just contain the auth_users in the config like we can with static accounts and an `authorization` block?

Maybe I'm missing something here in the NATS auth design, but if we want to own `auth_callout` and the account resolver, why do we need to get involved in the operator/NSC/creds file complexity? We have a very different definition of "simpler mechanism" when we're looking to own all auth/accounts and your version brings in 13 `nsc` calls and a creds file that must be distributed to a large number of physically inaccessible devices.

aricart:
@ghstahl can you reach out to me in slack?

neilalexander:
Closing this for now as stale unless @aricart feels it is needed.

PR #6436, opened 2025-01-30
Remove obsolete file

ReubenMathew:
Signed-off-by: Reuben Ninan <reuben@synadia.com>


ReubenMathew:
Fun fact: this file contains the oldest committed `.txt` lines in this repo.

wallyqs (approved):

wallyqs:
don't think was committed by accident though: https://github.com/nats-io/nats-server/commit/b56ca22d1bfb571fa395a35fe698b0eb7f95e706
we could also update

ReubenMathew:
@derekcollison @wallyqs, my apologies. I should've checked the context of why it got added.

Is it still worth to keep in the repo? I changed the commit message in case we want to remove the file.

derekcollison:
We should keep it, but update it I think..

PR #6438, opened 2025-01-31
NRG: Lockless `Leaderless` and `HadPreviousLeader`

neilalexander:
This removes lock contention around JS API requests figuring out if the meta group is leaderless or not by instead tracking that state atomically. 

Also fixes a race condition in inactivity thresholds/consumer pause that this seemed to uncover.

### DIFF
--- a/server/jetstream_cluster.go
+++ b/server/jetstream_cluster.go
@@ -825,7 +825,7 @@ func (js *jetStream) isLeaderless() bool {
 
 	// If we don't have a leader.
 	// Make sure we have been running for enough time.
-	if meta.GroupLeader() == _EMPTY_ && time.Since(meta.Created()) > lostQuorumIntervalDefault {
+	if meta.Leaderless() && time.Since(meta.Created()) > lostQuorumIntervalDefault {
 		return true
 	}
 	return false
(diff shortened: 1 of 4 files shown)


server/jetstream_cluster.go
```diff
@@ -4436,7 +4436,11 @@ func (js *jetStream) processClusterCreateConsumer(ca *consumerAssignment, state
 		if isConfigUpdate = !reflect.DeepEqual(&cfg, ca.Config); isConfigUpdate {
 			// Call into update, ignore consumer exists error here since this means an old deliver subject is bound
 			// which can happen on restart etc.
-			if err := o.updateConfig(ca.Config); err != nil && err != NewJSConsumerNameExistError() {
+			// JS lock needed as this can mutate the consumer assignments and race with updateInactivityThreshold.
+			js.mu.Lock()
```
derekcollison, line 4440: These suppose to be part of this PR?

derekcollison (commented):
LGTM - Just question on other changes.

server/jetstream_cluster.go
neilalexander: Something about the changing in the locking caused this race condition to hit far more often, to the point that it was taking several attempts of kicking CI.

I've kept it isolated in a second commit though, as that change won't be relevant for a 2.10.x backport.

derekcollison (approved):
LGTM

PR #6439, opened 2025-01-31
Expose `raftz` and `ipqueuesz` via system account

neilalexander:
This PR adds `$SYS.REQ.SERVER.*.RAFTZ` and `$SYS.REQ.SERVER.*.IPQUEUESZ` so that they can be queried via the system account as well as the monitoring port.

### DIFF
--- a/server/monitor.go
+++ b/server/monitor.go
@@ -1123,20 +1123,16 @@ func (s *Server) HandleStacksz(w http.ResponseWriter, r *http.Request) {
 	ResponseHandler(w, r, buf[:n])
 }
 
-type monitorIPQueue struct {
+type IpqueueszStatusIPQ struct {
 	Pending    int `json:"pending"`
 	InProgress int `json:"in_progress,omitempty"`
 }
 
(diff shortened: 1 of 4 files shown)


server/monitor.go
```diff
@@ -1128,15 +1128,9 @@ type monitorIPQueue struct {
...
 }
 
-func (s *Server) HandleIPQueuesz(w http.ResponseWriter, r *http.Request) {
-	all, err := decodeBool(w, r, "all")
-	if err != nil {
-		return
-	}
-	qfilter := r.URL.Query().Get("queues")
-
+func (s *Server) Ipqueuesz(opts *IpqueueszOptions) map[string]monitorIPQueue {
```
ripienaar, line 1131: Better if this returned an exported type.

I added a set of response types here https://github.com/nats-io/nats-server/commit/8841131b99b3a8c004d9061370071026eae54c0f to help callers use these things, be good to have those for these also

ripienaar (changes requested):

server/monitor.go
neilalexander: Done.

ripienaar (approved):
LGTM

PR #6447, opened 2025-02-04
Add filestore permutations to `BenchmarkJetStreamConsumeWithFilters`

neilalexander:
The relevant code for multi-filtered consumers differs considerably between the two stores, so we need to be watching both.


derekcollison (approved):
LGTM

PR #6448, opened 2025-02-04
Optimisations for `LoadNextMsgMulti`

neilalexander:
This PR optimises `LoadNextMsgMulti` by doing two things:

* Avoid large linear scans in `firstMatchingMulti` by using subject tree intersection when it looks like the FSS size is smaller than the scan range, this should be less expensive (although the threshold for choosing strategy may need further attention);
* Check the `dmap` before calling `mb.cacheLookup()` in either the FSS search or the linear scan, such that we can avoid calling `time.Now()` a lot when we are tight-looped scanning over a large number of interior deletes.

This PR provides a 2x improvement on the 1000-filter `LoadNextMsgMulti` unit test and provides a 10x improvement when tested against a store that has extremely sparse messages and 8 consumer filters.

### DIFF
--- a/server/filestore.go
+++ b/server/filestore.go
@@ -2329,9 +2329,6 @@ func (mb *msgBlock) firstMatchingMulti(sl *Sublist, start uint64, sm *StoreMsg) 
 	mb.mu.Lock()
 	defer mb.mu.Unlock()
 
-	// Will just do linear walk for now.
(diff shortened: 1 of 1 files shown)


derekcollison (commented):
In general LGTM, but when we hit a dmap entry, before we would update load ts. So maybe we should set a boolean when we match dmap and on exit update that just once?

derekcollison (approved):
LGTM

PR #6449, opened 2025-02-04
Increase disk I/O semaphore limit

neilalexander:
This updates `dios` to range up to 50% of the core count on a system with more than 32 CPU cores, otherwise obeying the 4 to 16 limit.

### DIFF
--- a/server/filestore.go
+++ b/server/filestore.go
@@ -10007,8 +10007,14 @@ var dios chan struct{}
 // golang.org's semaphore seemed a bit heavy.
 func init() {
 	// Limit ourselves to a sensible number of blocking I/O calls. Range between
-	// 4-16 concurrent disk I/Os based on the number of CPU cores available.
-	nIO := min(16, max(4, runtime.GOMAXPROCS(-1)))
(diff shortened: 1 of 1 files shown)


server/filestore.go
```diff
@@ -10007,8 +10007,14 @@ var dios chan struct{}
...
 func init() {
 	// Limit ourselves to a sensible number of blocking I/O calls. Range between
-	// 4-16 concurrent disk I/Os based on the number of CPU cores available.
-	nIO := min(16, max(4, runtime.GOMAXPROCS(-1)))
+	// 4-16 concurrent disk I/Os based on CPU cores, or 50% of cores if greater
+	// than 32 cores.
+	mp := runtime.GOMAXPROCS(-1)
+	nIO := min(16, max(4, mp))
+	if mp > 16 {
+		// If the system has more than 16 cores then limit dios to 50% of cores.
```
derekcollison, line 10015: Comments do not match with line 1011.

neilalexander: Fixed. Also fixed the `if mp` to 32 also, not that it really mattered because it would still have to be greater than `32` to influence `nIO`.

derekcollison (approved):
LGTM

PR #6450, opened 2025-02-04
Avoid last load timestamp update when ranging interior deletes

neilalexander:
This updates `generatePerSubjectInfo`, `NumPending` and `NumPendingMulti` to avoid updating the last load timestamp in a tight-loop while skipping over a potentially large number of interior deletes, as this becomes noticeable in the CPU profile.

Similar to one of the changes in #6448.

### DIFF
--- a/server/filestore.go
+++ b/server/filestore.go
@@ -2443,7 +2443,13 @@ func (mb *msgBlock) firstMatchingMulti(sl *Sublist, start uint64, sm *StoreMsg) 
 // fs lock should be held.
 func (mb *msgBlock) firstMatching(filter string, wc bool, start uint64, sm *StoreMsg) (*StoreMsg, bool, error) {
 	mb.mu.Lock()
-	defer mb.mu.Unlock()
+	var updateLLTS bool
+	defer func() {
+		if updateLLTS {
+			mb.llts = time.Now().UnixNano()
+		}
+		mb.mu.Unlock()
+	}()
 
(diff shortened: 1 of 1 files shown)


server/filestore.go
```diff
@@ -3193,10 +3207,16 @@ func (fs *fileStore) NumPending(sseq uint64, filter string, lastPerSubject bool)
...
+				if mb.dmap.Exists(seq) {
+					// Optimisation to avoid calling cacheLookup which hits time.Now().
+					updateLLTS = true
+					continue
+				}
 				sm, _ := mb.cacheLookup(seq, &smv)
 				if sm == nil || sm.subj == _EMPTY_ || !lbm[sm.subj] {
 					continue
 				}
+				updateLLTS = false // cacheLookup already updated it.
```
derekcollison, line 3219: Since we are looping here we probably need a llts updated to turn it off. Since with this if we have interleaved records that are there and deleted but do not match we will set it for each lookup and again for the deletes.

derekcollison (commented):
In general good. One comment etc.

server/filestore.go
derekcollison: True, should we expand cacheLookup to call cacheLookupEx that takes a bool indicating if we should update LLTS?

neilalexander: We could do, would you prefer I take that approach for this PR?

derekcollison: Not here but wanted the context here if we decide to do new PR.

PR #6451, opened 2025-02-04
WebSocket compression fixes

neilalexander:
This should ensure the compressors are more aggressively cleared and also handle a couple of error scenarios better by releasing/repooling buffers.


derekcollison (approved):
LGTM

PR #6453, opened 2025-02-04
Add `sliceHeader` for zero-copy parsing of message headers, use for client info

neilalexander:
In the various places that we are handling client info from headers, like when queuing and pulling from the JS API queue or due to the service export, we could safely refer to the underlying header slice, instead of making a copy as `getHeader()` does today.

This PR adds a new `sliceHeader()` that merely slices the input header instead, reducing copies considerably in any system where there is heavy usage of JS API requests.

### DIFF
--- a/server/client.go
+++ b/server/client.go
@@ -4248,9 +4248,20 @@ func (c *client) setHeader(key, value string, msg []byte) []byte {
 	return bb.Bytes()
 }
 
-// Will return the value for the header denoted by key or nil if it does not exists.
-// This function ignores errors and tries to achieve speed and no additional allocations.
+// Will return a copy of the value for the header denoted by key or nil if it does not exist.
+// If you know that it is safe to refer to the underlying hdr slice for the period that the
+// return value is used, then sliceHeader() will be faster.
 func getHeader(key string, hdr []byte) []byte {
+	v := sliceHeader(key, hdr)
+	if v == nil {
+		return nil
+	}
+	return append(make([]byte, 0, len(v)), v...)
+}
+
+// Will return the sliced value for the header denoted by key or nil if it does not exists.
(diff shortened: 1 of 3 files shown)


server/client.go
```diff
@@ -4275,15 +4286,14 @@ func getHeader(key string, hdr []byte) []byte {
...
+	start := index
 	for index < hdrLen {
 		if hdr[index] == '\r' && index < hdrLen-1 && hdr[index+1] == '\n' {
 			break
 		}
-		value = append(value, hdr[index])
 		index++
 	}
-	return value
+	return hdr[start:index]
```
derekcollison, line 4296: Should we also set cap here? That way if someone does try to do an append it will force a copy at that point?

neilalexander: Yep, done.

server/client.go
```diff
@@ -4248,9 +4248,20 @@ func (c *client) setHeader(key, value string, msg []byte) []byte {
...
-// This function ignores errors and tries to achieve speed and no additional allocations.
+// Will return a copy of the value for the header denoted by key or nil if it does not exist.
+// If you know that it is safe to refer to the underlying hdr slice for the period that the
+// return value is used, then sliceHeader() will be faster.
 func getHeader(key string, hdr []byte) []byte {
+	v := sliceHeader(key, hdr)
+	if v == nil {
+		return nil
+	}
+	return append(make([]byte, 0, len(v)), v...)
```
derekcollison, line 4259: Maybe do allocation and copy? Unless we know append is as efficient for larger sizes..

derekcollison (approved):
LGTM

server/client.go
neilalexander: It's just as efficient when the capacity is preallocated, so just makes for a tidier one-liner.

PR #6458, opened 2025-02-05
stree: Split `Iter` into `IterFast` and `IterOrdered`

neilalexander:
Not all of the `Iter()` call-sites required ordering, and ordering would slow down the process quite a bit, so replace with two new `IterFast()` and `IterOrdered()` functions instead.

This should speed up populating the per-subject infos, multi-filter num pending and enforcing the per-subject limits.

### DIFF
--- a/server/stree/stree.go
+++ b/server/stree/stree.go
@@ -124,13 +124,22 @@ func (t *SubjectTree[T]) Match(filter []byte, cb func(subject []byte, val *T)) {
 	t.match(t.root, parts, _pre[:0], cb)
 }
 
-// Iter will walk all entries in the SubjectTree lexographically. The callback can return false to terminate the walk.
-func (t *SubjectTree[T]) Iter(cb func(subject []byte, val *T) bool) {
+// IterOrdered will walk all entries in the SubjectTree lexographically. The callback can return false to terminate the walk.
+func (t *SubjectTree[T]) IterOrdered(cb func(subject []byte, val *T) bool) {
 	if t == nil || t.root == nil {
 		return
 	}
 	var _pre [256]byte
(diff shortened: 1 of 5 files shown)


server/stree/stree.go
```diff
@@ -378,6 +387,19 @@ func (t *SubjectTree[T]) iter(n node, pre []byte, cb func(subject []byte, val *T
...
 	// Note that this append may reallocate, but it doesn't modify "pre" at the "iter" callsite.
 	pre = append(pre, bn.prefix...)
+	// Not everything requires lexicographical sorting, so support a fast path for iterating in
+	// whatever order the stree has things stored instead.
+	if !ordered {
+		for _, cn := range n.children() {
+			if cn == nil {
+				continue
+			}
+			if !t.iter(cn, pre, ordered, cb) {
```
derekcollison, line 397: Make ordered `false` I think he to be more clear.

server/stree/stree.go
```diff
@@ -390,7 +412,7 @@ func (t *SubjectTree[T]) iter(n node, pre []byte, cb func(subject []byte, val *T
 	slices.SortStableFunc(nodes, func(a, b node) int { return bytes.Compare(a.path(), b.path()) })
 	// Now walk the nodes in order and call into next iter.
 	for i := range nodes {
-		if !t.iter(nodes[i], pre, cb) {
+		if !t.iter(nodes[i], pre, ordered, cb) {
```
derekcollison, line 415: Same here but true.

derekcollison (changes requested):
LGTM - one style question.

derekcollison (approved):
LGTM

PR #6461, opened 2025-02-05
[IMPROVED] Optimized calculating source sequences for sourced stream with lots of deletes.

derekcollison:
Added LoadPrevMsg to optimize when we walk backwards looking for source sequences and we have lots of interior deletes.

### DIFF
--- a/server/stream.go
+++ b/server/stream.go
@@ -3663,16 +3663,21 @@ func (mset *stream) setStartingSequenceForSources(iNames map[string]struct{}) {
 	}
 
 	var smv StoreMsg
-	for seq := state.LastSeq; seq >= state.FirstSeq; seq-- {
-		sm, err := mset.store.LoadMsg(seq, &smv)
-		if err != nil || len(sm.hdr) == 0 {
+	for seq := state.LastSeq; seq >= state.FirstSeq; {
+		sm, err := mset.store.LoadPrevMsg(seq, &smv)
+		if err == ErrStoreEOF || err != nil {
+			break
+		}
+		seq = sm.seq - 1
+		if len(sm.hdr) == 0 {
 			continue
 		}
+
 		ss := getHeader(JSStreamSource, sm.hdr)
 		if len(ss) == 0 {
 			continue
 		}
-		streamName, indexName, sseq := streamAndSeq(string(ss))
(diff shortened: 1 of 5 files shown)


server/stream.go
```diff
@@ -3663,16 +3663,21 @@ func (mset *stream) setStartingSequenceForSources(iNames map[string]struct{}) {
 	}
 
 	var smv StoreMsg
-	for seq := state.LastSeq; seq >= state.FirstSeq; seq-- {
-		sm, err := mset.store.LoadMsg(seq, &smv)
-		if err != nil || len(sm.hdr) == 0 {
+	for seq := state.LastSeq; seq >= state.FirstSeq; {
+		sm, err := mset.store.LoadPrevMsg(seq, &smv)
+		if err == ErrStoreEOF || err != nil {
```
neilalexander, line 3668: Seems like a tautological condition, if the err is `ErrStoreEOF` then it's also not nil. Is the EOF meant to be handled differently?

server/stream.go
```diff
@@ -3778,17 +3783,21 @@ func (mset *stream) startingSequenceForSources() {
 	}
 
 	var smv StoreMsg
-	for seq := state.LastSeq; seq >= state.FirstSeq; seq-- {
-		sm, err := mset.store.LoadMsg(seq, &smv)
-		if err != nil || sm == nil || len(sm.hdr) == 0 {
+	for seq := state.LastSeq; ; {
+		sm, err := mset.store.LoadPrevMsg(seq, &smv)
+		if err == ErrStoreEOF || err != nil {
```
neilalexander, line 3788: Same here.

neilalexander (commented):
Overall looks good, just one minor thing stood out.

server/stream.go
derekcollison: EOF is the normal exit condition, but I was being pedantic. Could move it to be just err != nil but that does not read well in terms of intent.

neilalexander (approved):
LGTM

server/stream.go
derekcollison: What do you suggest we do? Or leave as is for now?

neilalexander: I think it's fine if it is trying to be clear that it's mostly trying to trap EOFs, the function doesn't have any way to bubble up other errors anyway.

PR #6470, opened 2025-02-07
Purge requests with no subject delete markers

neilalexander:
As per [the ADR](https://github.com/nats-io/nats-architecture-and-design/blob/main/adr/ADR-43.md#purge-api-call-marker), this adds the `no_markers` flag to the JS API stream purge request, such that a stream can be purged without leaving behind any subject delete markers, if they are enabled.

This helps to avoid the "poison pill" problem if there are subjects for example that are causing problems with a consuming application.

Individual message deletes are less of a concern here because a single message delete will leave behind at most one marker, which can be deleted too (and would not be replaced in that case), but a purge can potentially leave behind lots of markers for lots of subjects, which could be difficult to sort through by hand otherwise.


derekcollison (approved):
LGTM

PR #6471, opened 2025-02-07
Apply client write deadline to batches rather than total flushes

neilalexander:
By calculating the write deadline at the beginning of the flush operation, we could be hitting it early if there is a lot of data queued up. Instead this PR applies it to each `writev` batch of at most 1024 vectors, which means that the deadline applies to no more than 64MB in one go.

This makes it simpler to tune the write deadline as it can be thought of as "the maximum time taken to allow writing at most 64MB" instead.

### DIFF
--- a/server/client.go
+++ b/server/client.go
@@ -1642,8 +1642,10 @@ func (c *client) flushOutbound() bool {
 		}
 		consumed := len(wnb)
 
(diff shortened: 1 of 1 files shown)


derekcollison (approved):
LGTM

PR #6474, opened 2025-02-07
Account for replicated msgs in user account.

derekcollison:
When replicating messages over NRGs that still use the system account, make sure to add in to account stats the sent msgs and bytes to replicate.


neilalexander (approved):
LGTM

Issue #6475, opened 2025-02-07
`pending_size` in /routez is not populated


alexbozhenko:
### Observed behavior

It looks like `pending_size` is never set in `/routez`:
https://docs.nats.io/running-a-nats-service/nats_admin/monitoring#route-information-routez
I guess it would be used here:
https://github.com/nats-io/nats-server/blob/67aa107d750eec8c1344aed2c336d8373ed7c437/server/monitor.go#L823

In https://github.com/nats-io/nats-server/pull/6422 I deleted the field from the struct, and the test passed...

### Expected behavior

The value is populated, with pending_bytes, I guess?

### Server and client version

main

PR #6476, opened 2025-02-07
Populate pending_bytes in routez

alexbozhenko:
Fixes: https://github.com/nats-io/nats-server/issues/6475

### DIFF
--- a/server/monitor.go
+++ b/server/monitor.go
@@ -784,7 +784,7 @@ type RouteInfo struct {
 	Idle         string             `json:"idle"`
 	Import       *SubjectPermission `json:"import,omitempty"`
 	Export       *SubjectPermission `json:"export,omitempty"`
-	Pending      int                `json:"pending_size"`
+	Pending      int64              `json:"pending_size"`
 	InMsgs       int64              `json:"in_msgs"`
 	OutMsgs      int64              `json:"out_msgs"`
 	InBytes      int64              `json:"in_bytes"`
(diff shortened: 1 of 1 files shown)


server/monitor.go
```diff
@@ -784,7 +784,7 @@ type RouteInfo struct {
 	Idle         string             `json:"idle"`
 	Import       *SubjectPermission `json:"import,omitempty"`
 	Export       *SubjectPermission `json:"export,omitempty"`
-	Pending      int                `json:"pending_size"`
+	Pending      int64              `json:"pending_size"`
```
alexbozhenko, line 787: Can we make it `json:"pending_bytes"`, to match other monitoring endpoints?

wallyqs: I think we leave as is already, also why changing to int64?

alexbozhenko: Changed type because `r.out.pb` on line 835 is int64.

Sorry, I did not understand you reply.
If we have the same entity referred to by two different names, it is bad. If this name was not actually used, then can we fix it here?

wallyqs: would rather not change the type and leave it as an `int` like ConnInfo, also if we leave as is then old code using the same structs would work already, like the natscli which imports these structs as is from the server.

bruth: @alexbozhenko Type changes are breaking changes for exported types in the package API, so we can't do that.

wallyqs (approved):

PR #6491, opened 2025-02-10
(v2.11) [IMPROVED] sourcing with limits: adds support for sourcing to a stream with a max bytes limit and discard new, like you can already do with max messages.

jnmoyne:
Improve sourcing to allow you to source into a stream with a max bytes limit and discard new, like you can already do with max messages.

replaces #5774


derekcollison (approved):
LGTM

PR #6492, opened 2025-02-11
[FIXED] Auth-Callout: Allow username-password/token in Leaf remote URL.

### LINKED ISSUE #6481: Add User and/or Token fields to server.LeafNodeOpts to enable auth callout for leaf node connections
### Proposed change

Right now a leaf node is able to successfully complete an auth callout flow assuming the auth callout server doesn't bother to check the presence or value of `user` or `token`. It also does not work to add User/Password to the `url.URL`s under `server.RemoteLeafOpts.URLs`.

Can we add User and Token fields to `server.RemoteLeafOpts.URLs` such that they flow out to the auth callout flow, without breaking any intended design goals or existing NATS behavior? If so I'd love to try to implement or otherwise participate in enabling this flow.

### Use case

The use case is: enable full auth callout behavior for leaf node operation, which is currently not fully possible (unless I am mistaken in my usage)

### Contribution

I would like to implement this change and tests if contribution requirements are made clear (or already available in documentation)

kozlovic:
If a username and password is provided in a leaf remote URL, as such:
```
leafnodes {
   remotes [
     {url: "nats://leaf:pwd@hub_host:7422"
   ]
```
or a authorization token, as such:
```
leafnodes {
   remotes [
     {url: "nats://myToken@hub_host:7422"
   ]
```
and if the server uses authorization callout, it will be possible to authenticate such leafnode, in either server configuration or operator mode.

Resolves #6481

### DIFF
--- a/server/auth.go
+++ b/server/auth.go
@@ -615,8 +615,8 @@ func (s *Server) processClientOrLeafAuthentication(c *client, opts *Options) (au
 		// Check config-mode. The global account is a condition since users that
 		// are not found in the config are implicitly bound to the global account.
 		// This means those users should be implicitly delegated to auth callout
-		// if configured.
-		if juc == nil && opts.AuthCallout != nil && c.acc.Name != globalAccountName {
+		// if configured. Exclude LEAF connections from this check.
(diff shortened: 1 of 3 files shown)


derekcollison (approved):
LGTM

Issue #6496, opened 2025-02-11
Inability to override MaxClosedClients via configuration file


ajax-surovskyi-y:
### Observed behavior

It looks like there no option to pass value for MaxClosedClients. Adjusting this parameter is crucial for my use case to effectively manage closed client connections. The default value 10000 for this parameter is relatively high, which can lead to increased memory usage and potential memory leaks in certain scenarios. Adjusting this parameter is crucial for my use case to effectively manage closed client connections and optimize resource utilization.
This a pprof file for server with no connections right now

![Image](https://github.com/user-attachments/assets/bf83d4dd-fb7f-4e2a-8be4-78b99b85f692)

### Expected behavior

Create a configuration file named nats.conf with the following content:

`max_closed_clients: 100`
The MaxClosedClients value should be set to 100 as specified in the configuration file.

### Server and client version

nats-server 2.10.25

PR #6497, opened 2025-02-11
Handle max_closed_clients property

ajax-surovskyi-y:
Fixes: https://github.com/nats-io/nats-server/issues/6496

Signed-off-by: Yevhen Surovskyi <surovskyi.y@ajax.systems>

### DIFF
--- a/server/opts.go
+++ b/server/opts.go
@@ -1182,8 +1182,6 @@ func (o *Options) processConfigFileLine(k string, v any, errors *[]error, warnin
 		o.MaxTracedMsgLen = int(v.(int64))
 	case "max_subscriptions", "max_subs":
 		o.MaxSubs = int(v.(int64))
-	case "max_closed_clients":
-		o.MaxClosedClients = int(v.(int64))
 	case "max_sub_tokens", "max_subscription_tokens":
 		if n := v.(int64); n > math.MaxUint8 {
 			err := &configErr{tk, fmt.Sprintf("%s value is too big", k)}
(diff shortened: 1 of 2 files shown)


server/opts_test.go
```diff
@@ -1360,6 +1360,17 @@ func TestPanic(t *testing.T) {
...
 
+func TestMaxClosedClients(t *testing.T) {
+	conf := createConfFile(t, []byte(`max_closed_clients: 5`))
+	opts := &Options{}
+	if err := opts.ProcessConfigFile(conf); err != nil {
+		t.Fatalf("expected no error")
+	}
+	if opts.MaxClosedClients != 5 {
+		t.Fatalf("expected max closed clients to be 5")
+	}
```
kozlovic, line 1371: This could be replaced with:
```
opts, err := ProcessConfigFile(conf)
require_NoError(t, err)
require_Equal(t, opts.MaxClosedClients, 5)
```

server/opts.go
```diff
@@ -1182,6 +1182,8 @@ func (o *Options) processConfigFileLine(k string, v any, errors *[]error, warnin
 		o.MaxTracedMsgLen = int(v.(int64))
 	case "max_subscriptions", "max_subs":
 		o.MaxSubs = int(v.(int64))
+	case "max_closed_clients":
```
kozlovic, line 1185: We don't have really a rule on where to place it, but it looks like it is in between 2 properties related to subscriptions, so I would move it either down by one, or like I tend to do, add the new ones at the end of the `switch` (before the `default`) to help when it is time to document (easier to find the new ones ;-)).

kozlovic (changes requested):
Test could be "simplified".

ajax-surovskyi-y:
@kozlovic, I've fixed your comments. re-review please

server/opts.go
```diff
@@ -1677,6 +1677,8 @@ func (o *Options) processConfigFileLine(k string, v any, errors *[]error, warnin
 		}
 	case "no_fast_producer_stall":
 		o.NoFastProducerStall = v.(bool)
+  case "max_closed_clients":
```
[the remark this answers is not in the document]
ajax-surovskyi-y: oh, i'm sorry. i fixed merge conflict through ui =( that the reason
fixed

kozlovic (approved):
LGTM

PR #6506, opened 2025-02-13
Updated test for PR #6500

kozlovic:
Consolidated the 2 tests into 1, reducing run time from 7-15 sec down to less than 0.4 second and memory usage from 1GB+ down to 18M or so (which is likely not even the usage for the test itself but the sublist tests).


wallyqs (approved):
LGTM

PR #6517, opened 2025-02-17
[FIXED] LeafNode: possible delivery to several DQ members across Gateway

kozlovic:
If the same queue group has members running on different leafnodes connected through a gateway, it was possible for a message to be delivered to several members running on different leaf nodes if there was an interest (either plain subscription or for other queue groups) that made the produce message travel through the gateway.


derekcollison (approved):
LGTM

PR #6534, opened 2025-02-20
Fix `TestLongClusterWorkQueueMessagesNotSkipped`

neilalexander:
We need to check for `ErrSubscriptionClosed` as the defer to close the client and shutdown the server will happen while the consumer goroutines are still running, so we need to detect that gracefully rather than accidentally failing the test.

[ci skip] as this test isn't run under Travis.


derekcollison (approved):
LGTM

PR #6537, opened 2025-02-20
Add automemlimit

alexbozhenko:
## Summary
The problem:
image

A Guide to the Go Garbage Collector [suggests](https://go.dev/doc/gc-guide#:~:text=Do%20take%20advantage,is%20unaware%20of):

> Do take advantage of the memory limit when the execution environment of your Go program is entirely within your control, and the Go program is the only program with access to some set of resources (i.e. some kind of memory reservation, like a container memory limit).
> A good example is the deployment of a web service into containers with a fixed amount of available memory.
> In this case, a good rule of thumb is to leave an additional 5-10% of headroom to account for memory sources the Go runtime is unaware of.

When running in Kubernetes, the memory limit [set using cgroups]( https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#how-pods-with-resource-limits-are-run)

Let's use the [library](https://pkg.go.dev/github.com/KimMachineGun/automemlimit@v0.7.0/memlimit) that can automatically set the [GOMEMLIMIT](https://pkg.go.dev/runtime/debug#SetMemoryLimit) to a set percentage of memory limit of the cgroup

## Reproducibe test plan:
1. When no GOMEMLIMIT is set, the effective value is the default, math.MaxInt64 
```
# go build .

# ./nats-server 
...
[81117] 2025/02/20 12:59:21.150405 [INF] Server is ready
2025/02/20 12:59:21 INFO memory is not limited, skipping package=github.com/KimMachineGun/automemlimit/memlimit
[81117] 2025/02/20 12:59:21.151533 [INF] Effective GOMEMLIMIT: 9223372036854775807 bytes
```

2. When env var is set, its value is respected:
```
# GOMEMLIMIT=1000MiB ./nats-server 
[81345] 2025/02/20 13:00:50.034879 [INF] Server is ready
2025/02/20 13:00:50 INFO GOMEMLIMIT is already set, skipping package=github.com/KimMachineGun/automemlimit/memlimit GOMEMLIMIT=1000MiB
[81345] 2025/02/20 13:00:50.035615 [INF] Effective GOMEMLIMIT: 1048576000 bytes
```

Build a docker image for test: 
```
goreleaser release --snapshot --clean -f .goreleaser-nightly.yml 
```

3. When running in docker, by default no limit is set:
```
# docker run synadia/nats-server:add_automemlimit
...
[1] 2025/02/20 21:03:10.188627 [INF] Server is ready
[1] 2025/02/20 21:03:10.188685 [INF] Cluster name is 9QPBj9CtvlZlE9b8iyUn3a
[1] 2025/02/20 21:03:10.188829 [WRN] Cluster name was dynamically generated, consider setting one
[1] 2025/02/20 21:03:10.188923 [INF] Listening for route connections on 0.0.0.0:6222
2025/02/20 21:03:10 INFO memory is not limited, skipping package=github.com/KimMachineGun/automemlimit/memlimit
[1] 2025/02/20 21:03:10.189336 [INF] Effective GOMEMLIMIT: 9223372036854775807 bytes
```

4. When running in a container with a memory limit:
```
# docker run --cpus="4.0" --memory="2000m"  --name nats_container synadia/nats-server:add_automemlimit
[1] 2025/02/20 21:03:58.776768 [INF] Server is ready
[1] 2025/02/20 21:03:58.776869 [INF] Cluster name is ffq6b2J0D4cUuImY6PcyRA
[1] 2025/02/20 21:03:58.776899 [WRN] Cluster name was dynamically generated, consider setting one
[1] 2025/02/20 21:03:58.776986 [INF] Listening for route connections on 0.0.0.0:6222
2025/02/20 21:03:58 INFO GOMEMLIMIT is updated package=github.com/KimMachineGun/automemlimit/memlimit GOMEMLIMIT=1677721600 previous=9223372036854775807
[1] 2025/02/20 21:03:58.777953 [INF] Effective GOMEMLIMIT: 1677721600 bytes
```
The GOMEMLIMIT is set to 80% of the cgroup memory limit, as expected.
```
# cat  /sys/fs/cgroup/system.slice/docker-$(docker inspect --format="{{.Id}}" nats_container).scope/memory.max
2097152000
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -5,6 +5,7 @@ go 1.23.0
 toolchain go1.23.6
 
 require (
+	github.com/KimMachineGun/automemlimit v0.7.0
 	github.com/antithesishq/antithesis-sdk-go v0.4.3-default-no-op
 	github.com/google/go-tpm v0.9.3
 	github.com/klauspost/compress v1.17.11
@@ -18,3 +19,5 @@ require (
 	golang.org/x/sys v0.30.0
 	golang.org/x/time v0.10.0
 )
+
+require github.com/pbnjay/memory v0.0.0-20210728143218-7b4eea64cf58 // indirect
--- a/main.go
+++ b/main.go
@@ -18,8 +18,11 @@ package main
 import (
 	"flag"
 	"fmt"
+	"log/slog"
 	"os"
+	"runtime/debug"
 
+	"github.com/KimMachineGun/automemlimit/memlimit"
 	"github.com/nats-io/nats-server/v2/server"
 	"go.uber.org/automaxprocs/maxprocs"
 )
@@ -135,6 +138,11 @@ func main() {
 	} else {
 		defer undo()
 	}
+
+	// If GOMEMLIMIT is not set, adjust GOMEMLIMIT if running under linux/cgroups quotas.
+	memlimit.SetGoMemLimitWithOpts(memlimit.WithRatio(0.8), memlimit.WithLogger(slog.Default()))
+	currentMemLimit := debug.SetMemoryLimit(-1)
+	s.Noticef("Effective GOMEMLIMIT: %d bytes", currentMemLimit)
 
 	s.WaitForShutdown()
 }
--- a/go.sum
+++ b/go.sum
@@ -1,3 +1,5 @@
+github.com/KimMachineGun/automemlimit v0.7.0 h1:7G06p/dMSf7G8E6oq+f2uOPuVncFyIlDI/pBWK49u88=
+github.com/KimMachineGun/automemlimit v0.7.0/go.mod h1:QZxpHaGOQoYvFhv/r4u3U0JTC2ZcOwbSr11UZF46UBM=
 github.com/antithesishq/antithesis-sdk-go v0.4.3-default-no-op h1:+OSa/t11TFhqfrX0EOSqQBDJ0YlpmK0rDSiB19dg9M0=
 github.com/antithesishq/antithesis-sdk-go v0.4.3-default-no-op/go.mod h1:IUpT2DPAKh6i/YhSbt6Gl3v2yvUZjmKncl7U91fup7E=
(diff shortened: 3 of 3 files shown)


main.go
```diff
@@ -136,5 +139,10 @@ func main() {
 		defer undo()
 	}
 
+	// If GOMEMLIMIT is not set, adjust GOMEMLIMIT if running under linux/cgroups quotas.
+	memlimit.SetGoMemLimitWithOpts(memlimit.WithRatio(0.8), memlimit.WithLogger(slog.Default()))
```
alexbozhenko, line 143: Probably do not want to use slog here, but I was not sure how to plug it to the server logger, since memlimit.WithLogger expects slog logger.

go.mod
```diff
@@ -18,3 +19,5 @@ require (
 	golang.org/x/sys v0.30.0
 	golang.org/x/time v0.10.0
 )
+
+require github.com/pbnjay/memory v0.0.0-20210728143218-7b4eea64cf58 // indirect
```
wallyqs, line 23: this dependency is not tagged nor released in 4-7 years... pinged the author on whether they could tag one

main.go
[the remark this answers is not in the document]
wallyqs: yes I think sudden default to 0.8 might be too sudden as a knob, in the config would be better. With default GOMEMLIMIT you cannot do fractions but would be interesting if this dependency supports doing 0.85 for example.

wallyqs: I think we would need to make the server logger implement the `Handler` interface from slog, then here call `slog.New(s.Logger())`

neilalexander:
I'm a little bit concerned that we're introducing more magic here that's difficult to reason about, and it may have surprising side effects, like when using both a limit and a reservation. We also often see people trip up due to how Kubernetes handles the page cache accounting and I suspect this may make things worse, so going to close this for now.

PR #6568, opened 2025-02-21
[IMPROVED] Adjusted stalled producers to not be as penalizing.

### LINKED ISSUE #5394: Performance degradation with slow consumers [v2.9.X, v2.10.X]
### Observed behavior

Performance degradation after the slow consumer connection. 
As you can see below, we are observing about 30% degradation of the incoming messages when the slow consumer connected, and about 50% after the second one.

CleanShot 2024-05-07 at 16 58 44@2x

### Expected behavior

Stop sending messages to the slow consumers until their buffers are empty without slowing down the server.

### Server and client version

Server: 2.9.20
Python library: nats-py 2.7.2

### Host environment

**Local:**
MacOS 14.4.1, arm64, Docker 26.0.0
The same behavior with the amd64 emulator (--platform=linux/amd64 flag).

**GKE**
Container-Optimized OS, amd64, containerd

### Steps to reproduce

Prepared required configs and docker-compose file that will start NATS, Prometheus, an exporter, and two consumers: [https://github.com/nenych/nats-test](https://github.com/nenych/nats-test).

#### Steps to run
1. Clone the repository.
2. Build the docker image:
```
docker build -t test/nats:latest .
```
3. Install NATS cli: https://docs.nats.io/using-nats/nats-tools/nats_cli
4. Run docker-compose (will start NATS, prometheus and 2 consumers):
```
docker-compose -f ./docker-compose.yaml up -d
```
5. Start NATS benchmark:
```
nats bench updates --pub=4 --msgs 1000000000 --size=1000
```
6. Wait a little and start the slow consumer:
```
docker run --rm -it --network=nats-test_default test/nats:latest python3 slow-consumer.py
```

#### Explore metrics
1. Open prometheus: http://localhost:9091/graph
2. Insert query:
```
sum by (job) (rate(nats_varz_in_msgs[30s]))
```

derekcollison:
Fixes #5394

### DIFF
--- a/server/client.go
+++ b/server/client.go
@@ -113,8 +113,9 @@ const (
 	maxNoRTTPingBeforeFirstPong = 2 * time.Second
 
 	// For stalling fast producers
-	stallClientMinDuration = 100 * time.Millisecond
-	stallClientMaxDuration = time.Second
+	stallClientMinDuration = 2 * time.Millisecond
+	stallClientMaxDuration = 5 * time.Millisecond
+	stallTotalAllowed      = 10 * time.Millisecond
 )
 
 var readLoopReportThreshold = readLoopReport
@@ -464,6 +465,9 @@ type readCache struct {
 
 	// Capture the time we started processing our readLoop.
 	start time.Time
+
+	// Total time stalled so far for readLoop processing.
+	tst time.Duration
 }
 
 // set the flag (would be equivalent to set the boolean to true)
@@ -1416,6 +1420,11 @@ func (c *client) readLoop(pre []byte) {
 				}
 				return
 			}
+			// Clear total stalled time here.
+			if c.in.tst > 0 {
+				c.rateLimitFormatWarnf("Producer was stalled for a total of %v", c.in.tst.Round(time.Millisecond))
+				c.in.tst = 0
+			}
 		}
 
 		// If we are a ROUTER/LEAF and have processed an INFO, it is possible that
@@ -1734,7 +1743,7 @@ func (c *client) flushOutbound() bool {
 
 	// Check if we have a stalled gate and if so and we are recovering release
 	// any stalled producers. Only kind==CLIENT will stall.
-	if c.out.stc != nil && (n == attempted || c.out.pb < c.out.mp/2) {
+	if c.out.stc != nil && (n == attempted || c.out.pb < c.out.mp/4*3) {
 		close(c.out.stc)
 		c.out.stc = nil
 	}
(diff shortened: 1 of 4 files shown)


server/client.go
```diff
@@ -1416,6 +1420,11 @@ func (c *client) readLoop(pre []byte) {
 				}
 				return
 			}
+			// Clear total stalled time here.
+			if c.in.tst > 0 {
+				c.rateLimitFormatWarnf("Producer was stalled for a total of %v", c.in.tst.Round(time.Millisecond))
```
wallyqs, line 1425: something that I noticed is that occasionally when the cluster is busy the natscli would stall for 2ms on simple queries, in case it gets noisy in the logs maybe we log when larger than 100ms?

wallyqs (approved):
LGTM

alexbozhenko:
Speaking from the perspective of someone who is not very familiar with NATS internals(majority of users?): it would be nice to have a sentence or two in the PR description on what is changing and why.

server/client_test.go
```diff
@@ -2061,32 +2061,6 @@ func TestClientNoSlowConsumerIfConnectExpected(t *testing.T) {
 	}
 }
 
-func TestClientStalledDuration(t *testing.T) {
```
neilalexander, line 2064: Is it worthwhile adding a new test by mocking up some `*client`s with custom `tst`, `pb` and `mp` values and then calling `stalledWait`?

wallyqs:
@alexbozhenko linked to the kind of issues that this would fix: #5394

server/client.go
derekcollison: Stall total is now 10ms, so will use that..

server/client_test.go
derekcollison: For what purpose?

derekcollison:
Fixed a broken test and updated warning based on @wallyqs feedback.

wallyqs (approved):
LGTM

server/client_test.go
neilalexander: Mostly thinking that we don't have any unit tests that prove that `stalledWait` actually blocks for as long as we expect or when we'd expect, or that it unblocks when the `stc` is closed.

derekcollison:
We do have some actually, and we have lots of benchmarks that test it.. I could add more, but the msgtrace ones are pretty good.

If we feel that is required, I can add it in..

neilalexander (approved):
I hadn't looked at the message tracing tests but if you think the coverage is OK then the changes look good to me.

derekcollison:
I will see if I can write a test..

derekcollison:
I added a test, which caught a minor bug, so good idea.

Should be ready for final review and I will squash once approved and tests are green.

PR #6585, opened 2025-02-26
Use GitHub Actions for releases

neilalexander:
This switches the release process to use GitHub Actions instead of Travis CI.

Tested on a forked repository:
- Successful run: https://github.com/neilalexander/nats-server/actions/runs/13546082159/job/37857914525
- Created release: https://github.com/neilalexander/nats-server/releases/tag/v2.11.0-RC.4

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -1,0 +1,44 @@
+name: NATS Server Releases
+on:
+  push:
+    tags:
+      - v*
+
+jobs:
+  run:
+    name: GitHub Release
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+          fetch-tags: true
+
+      - name: Set up Go
+        uses: actions/setup-go@v5
+        with:
+          go-version: "stable"
+
+      - name: Check version matches tag
+        env:
+          TRAVIS_TAG: ${{ github.ref_name }}
+        run: |
+          go test -race -v -run=TestVersionMatchesTag ./server -ldflags="-X=github.com/nats-io/nats-server/v2/server.serverVersion=$TRAVIS_TAG" -count=1 -vet=off
+
+      - name: Install cosign
+        uses: sigstore/cosign-installer@v3.7.0
+
+      - name: Install syft
+        uses: anchore/sbom-action/download-syft@v0.18.0
+        with:
+          syft-version: "v1.20.0"
+
+      - name: Create release
+        uses: goreleaser/goreleaser-action@v6
+        with:
+          distribution: goreleaser
+          version: "~> v2"
+          args: release --clean
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -3,7 +3,7 @@ version: 2
 
 release:
   github:
-    owner: nats-io
+    owner: '{{ envOrDefault "GITHUB_REPOSITORY_OWNER" "nats-io" }}'
     name: nats-server
   name_template: "Release {{.Tag}}"
   draft: true
@@ -71,10 +71,10 @@ archives:
   - name_template: "{{.ProjectName}}-{{.Tag}}-{{.Os}}-{{.Arch}}{{if .Arm}}{{.Arm}}{{end}}"
     id: targz-archives
     wrap_in_directory: true
-    format: tar.gz
+    formats: ["tar.gz"]
     format_overrides:
       - goos: windows
-        format: zip
+        formats: ["zip"]
     files:
       - README.md
       - LICENSE
--- a/.travis.yml
+++ b/.travis.yml
@@ -53,17 +53,3 @@ jobs:
       env: TEST_SUITE=build_only
 
 script: ./scripts/runTestsOnTravis.sh $TEST_SUITE
-
-# Install Syft which is used by GoReleaser in the deploy step.
-# Cosign is an optional dependency to verify the Syft binary.
-before_deploy:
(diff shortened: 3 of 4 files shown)


wallyqs:
We need to add the version check as well: https://github.com/nats-io/nats-server/blob/b79960e33ca48f0b0ba716328f2946d8bb5990dc/scripts/runTestsOnTravis.sh#L13

neilalexander:
OK, added the version check, this is what it looks like when it fails: https://github.com/nats-io/nats-server/actions/runs/13547337052/job/37862128062

wallyqs:
in the binary syft is detecting the pseudo version instead so something is missing atm

```
syft nats-server-v2-11-rc4 
NAME                               VERSION                             TYPE        
github.com/klauspost/compress      v1.18.0                             go-module    
github.com/minio/highwayhash       v1.0.3                              go-module    
github.com/nats-io/jwt/v2          v2.7.3                              go-module    
github.com/nats-io/nats-server/v2  v0.0.0-20250226142255-613a6b72547d  go-module    
github.com/nats-io/nkeys           v0.4.10                             go-module    
github.com/nats-io/nuid            v1.0.1                              go-module    
go.uber.org/automaxprocs           v1.6.0                              go-module    
golang.org/x/crypto                v0.34.0                             go-module    
golang.org/x/sys                   v0.30.0                             go-module    
golang.org/x/time                  v0.10.0                             go-module    
stdlib                             go1.24.0                            go-module  
```

v2.10.26

```
github.com/klauspost/compress      v1.18.0   go-module    
github.com/minio/highwayhash       v1.0.3    go-module    
github.com/nats-io/jwt/v2          v2.7.3    go-module    
github.com/nats-io/nats-server/v2  v2.10.26  go-module    
github.com/nats-io/nkeys           v0.4.10   go-module    
go.uber.org/automaxprocs           v1.6.0    go-module    
golang.org/x/crypto                v0.34.0   go-module    
golang.org/x/sys                   v0.30.0   go-module    
golang.org/x/time                  v0.10.0   go-module    
stdlib                             go1.23.6  go-module  
```

wallyqs:
Not seeing why the serverVersion would not be picked up from the goreleaser... maybe it is a Go 1.24 issue?
 https://github.com/nats-io/nats-server/blob/b79960e33ca48f0b0ba716328f2946d8bb5990dc/.goreleaser.yml#L20

.goreleaser.yml
```diff
@@ -3,7 +3,7 @@ version: 2
 
 release:
   github:
-    owner: nats-io
+    owner: '{{ envOrDefault "GITHUB_REPOSITORY_OWNER" "nats-io" }}'
```
kozlovic, line 6: Is that essentially so that you can override in your fork repo to test? (and if so, that's great and not asking to remove, just checking)

.github/workflows/release.yaml
```diff
@@ -0,0 +1,44 @@
...
+      - name: Install cosign
+        uses: sigstore/cosign-installer@v3.7.0
+
+      - name: Install syft
+        uses: anchore/sbom-action/download-syft@v0.18.0
+        with:
+          syft-version: "v1.20.0"
+
+      - name: Create release
+        uses: goreleaser/goreleaser-action@v6
```
kozlovic, line 38: Do you know if the goreleaser behind this will be "pinned" in that the version does not change? If it does, is there a way to pin to be sure we don't get a bad surprise during a release, or alternatively and like I was doing when I was doing releases, I would run the goreleaser locally (with the `--skip-publish` option) to make sure that everything is ok.

.github/workflows/release.yaml
```diff
@@ -0,0 +1,44 @@
...
+      - name: Checkout code
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+          fetch-tags: true
+
+      - name: Set up Go
+        uses: actions/setup-go@v5
+        with:
+          go-version: "stable"
```
kozlovic, line 21: Will there be a need to pin the version? Meaning, suppose that we are at Go 1.23.x at the moment when releasing NATS v2.10.26 but then the Go stable release becomes Go 1.24, and we need to release v2.10.27, I think we may want to stay in the same Go release version to minimize the changes, no?

kozlovic (commented):
Some comments before approving...

.goreleaser.yml
neilalexander: Yep, just means that forking to `neilalexander/nats-server` doesn't need more fudging to work.

.github/workflows/release.yaml
neilalexander: If we backported this to the 2.10.x branch then I'd likely pin Go 1.23 at the same time, but not sure it's really necessary for `main`.

neilalexander:
@wallyqs I've updated the syft version while investigating this, but I think it's unrelated to the Go version.

I think it's because my forked repository has the same `github.com/nats-io/nats-server/v2` module name in `go.mod` but the actual GitHub VCS path is different, so the tags that I'm creating on my fork can't be resolved by the `go` tool at `github.com/nats-io/nats-server` (as expected). We'd ideally need to see what happens when cutting RC.2 from the main repository.

kozlovic (approved):
LGTM

.github/workflows/release.yaml
alexbozhenko, line 21: Should we add a comment clarifying that:
`# Note that the version we set up here does not matter, since we specify GOTOOLCHAIN in goreleaser.yml`

neilalexander: The `toolchain` is a minimum, if `stable` surpasses `toolchain` then `stable` will be used instead.

wallyqs (approved):
LGTM

wallyqs:
syft was not working because go uses pseudo versions for RCs, but confirmed that using v2.11.0 as the tag on Go v1.24 works ok: 
```
go build --trimpath -ldflags="-X github.com/nats-io/nats-server/v2/server.serverVersion=2.11.0" .
```

PR #6607, opened 2025-03-05
[FIXED] Gateway: Possible loss of queue sub interest on restart

kozlovic:
If a server has a connection from a route or a leaf and the remote has more than one member of the same queue group, when the connection is recreated (due to network issue or server restart), then count coming from the connection would be greater than 1 but the server would store the interest count for outbound gateways as 1.

When one of the remote queue subscriber unsubscribes, the count would go down to 0 and the server would incorrectly send a protocol to the gateway indicating that there is no longer interest for this queue group, which would stop the flow.

A new queue subscriber on that group would resume the flow.

### DIFF
--- a/server/gateway.go
+++ b/server/gateway.go
@@ -2412,7 +2412,7 @@ func (s *Server) gatewayUpdateSubInterest(accName string, sub *subscription, cha
 		if change < 0 {
 			return
 		}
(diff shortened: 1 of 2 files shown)


derekcollison (approved):
LGTM

PR #6609, opened 2025-03-07
[FIXED] Config Reload: JetStream command line override ignored

### LINKED ISSUE #6606: JetStream cmdline arguments are discarded upon reload
### Observed behavior

Upon reload, Jetstream cmdline arguments are discarded.

Some args, like `--addr` and `--user`, are kept. Unsure if there are other args beyond Jetstream that are also discarded.

### Expected behavior

Upon a reload, all original cmdline arguments are still honored.

#### Notes

FWIW, we're using a mix of conf file + cmdline args in our test/CI setup. We use a static conf file, and pass different Jetstream arguments depending on whether certain tests are running Jetstream or not, and to give cluster members unique store-dir's.

This was discovered when testing the `$SYS.REQ.SERVER.server-id.RELOAD` API, although the repro below shows SIGHUP because it's simpler.

Ideally reloads would remember the cmdline args. If for some reason that's not possible, then either reloads should be blocked in a mixed cmdline + conf file situation, or at the least a warning should be emitted during reload.

### Server and client version

nats-server 2.10.26 (installed via homebrew)
nats cli 0.1.6 (installed via homebrew)

### Host environment

macOS 13.7.4

### Steps to reproduce

```
$ cat nats.conf
accounts: {
  '$SYS': {
    users: [ {user: 'sys'} ]
  },
  'TEST': {
    jetstream: true,
    users: [ {user: 'a', password: 'pass'} ]
  }
}

$ nats-server -a localhost -p 4223 -c nats.conf --js --store_dir tmp/nats/
[67210] 2025/03/05 09:59:18.550546 [INF] Starting nats-server
[67210] 2025/03/05 09:59:18.551623 [INF]   Version:  2.10.26
[67210] 2025/03/05 09:59:18.551625 [INF]   Git:      [not set]
[67210] 2025/03/05 09:59:18.551627 [INF]   Name:     NBBJUV7WS3GGCOYXKJQNX4KDXI5PTQ7NTLE3VKC4FDFGSY7FTFDLSVT2
[67210] 2025/03/05 09:59:18.551629 [INF]   Node:     ZjKmbo9f
[67210] 2025/03/05 09:59:18.551630 [INF]   ID:       NBBJUV7WS3GGCOYXKJQNX4KDXI5PTQ7NTLE3VKC4FDFGSY7FTFDLSVT2
[67210] 2025/03/05 09:59:18.551648 [WRN] Plaintext passwords detected, use nkeys or bcrypt
[67210] 2025/03/05 09:59:18.551670 [INF] Using configuration file: nats.conf
[67210] 2025/03/05 09:59:18.552985 [INF] Starting JetStream
...


$ nats -s nats://localhost:4223 account info --user a --password a
...
JetStream Account Information:
Account Usage:
                        Storage: 0 B
                         Memory: 0 B
                        Streams: 0
                      Consumers: 0
...

$ kill -HUP 67210
[67210] 2025/03/05 10:03:29.518293 [INF] Reloaded: authorization users
[67210] 2025/03/05 10:03:29.518314 [INF] Reloaded: accounts
[67210] 2025/03/05 10:03:29.518319 [INF] Reloaded: JetStream
[67210] 2025/03/05 10:03:29.519294 [INF] Initiating JetStream Shutdown...
[67210] 2025/03/05 10:03:29.519332 [INF] JetStream Shutdown
[67210] 2025/03/05 10:03:29.519376 [INF] Reloaded server configuration

$ nats -s nats://localhost:4223 account info --user a --password a
...
    No response from JetStream server
```

As seen, when JS is enabled via cmdline, reloading disables it. Reloads via the `$SYS.REQ.SERVER.server-id.RELOAD` API have the same problem.

kozlovic:
When the `--js` (and `--store_dir`) command line argument(s) are
given and a configuration reload is issued, JetStream would be
disabled.
    
This is because we were not tracking the JetStream command line
argument boolean, and also not merging the store_dir option.
    
When expanding the `TestConfigReloadBoolFlags` test, before making
the code fix, I noticed that the test would pass. This was due to
two reasons: the JS flag was pointing to the wrong object, the
reload would not do anything (since there was no digest change)
and the options post-reload were not refreshed.
    
Since the introduction of the config digest, some tests would
skip the ReloadOptions(), when tests were specifically written
to have that executed (for instance, there is a test that was
checking for data races, etc..). In other words, the skip of
ReloadOptions() (because digest would be the same without a
change of the underlying file) could mask issues that we wrote
tests for, so I decided to disable in general, and enable for
specific ones.
    
Resolves #6606
    

server/reload_test.go
```diff
@@ -3808,6 +3818,49 @@ func TestConfigReloadBoolFlags(t *testing.T) {
...
+		},
+		{
+			"jetstream_true_in_config_no_override",
+			"jetstream: true",
+			nil,
+			true,
+			func() bool { return opts.JetStream },
+		},
+		{
+			"jetstream_true_in_config_override_false",
```
kozlovic, line 3858: @wallyqs When adding this (and without the code fix), the test would pass for some reasons, including the fact that Reload() was not doing what it should be doing. This is why I had to add the disable function for the configuration digest.

Arguably, we could have all tests that call Reload() without an actual file change be checked/fixed, but I feel like it is more prudent to disable config digest in most tests but enable only for the ones that specifically check for the digest. Does that sound reasonable?

derekcollison:
Thanks @kozlovic - @wallyqs can you take a look here?

conf/parse.go
```diff
@@ -127,6 +129,10 @@ func cleanupUsedEnvVars(m map[string]any) {
 	}
 }
 
+func DisableConfigurationDigest(disable bool) {
```
neilalexander, line 132: Since this would become a part of the exported API, I think it would be best if we add comments on why you might want to (or moreover, why you probably shouldn't) use it. It's also not thread-safe if someone does decide to fiddle with things with a running embedded server.

kozlovic: You know, I don't like it either, and I had started with a private variable in the server that was set it sets, but I would have had the same problem with the tests in the `test` package.

What I am going to do, is likely revert the changes unrelated to the JS command line parameters and on a different branch, run all tests with a debug statement in place where we ignore and have a sense of how many tests need review and go from there.

I am going to switch this PR to Draft until I have a chance to update it later today.

kozlovic:
@derekcollison See my comments to Neil. I will update this PR later today and will work on the config digest "issue" separately.

kozlovic:
@derekcollison @wallyqs @neilalexander Ok, I have removed the changes for the config digest and this PR is ready for review. As I mentioned, I will evaluate separately tests that would need review due to the introduction of config digest.

derekcollison (approved):
LGTM

kozlovic:
Have we just disable Travis? It did not run for this update and yet the "All checks have passed" is green...

server/reload_test.go
wallyqs: Thanks @kozlovic, I think probably need to revisit the skip reload if no changes to config since I think it would mean that TLS certs changes without modifying the paths to the certs do not trigger a reload.

wallyqs:
(travis should still be enabled)

server/reload_test.go
kozlovic: But that was the point of the config digest, no? Meaning don't do the reload if no change is detected. Of course, and I think it may have been mentioned in the original PR, what constitute a change may be difficult when it comes to files.. you may need to do additional checks. Does that mean I should not bother evaluating the tests that call Reload() where nothing happens since you may revisit the config digest?

derekcollison:
It should kick Travis still..

server/reload_test.go
wallyqs: @kozlovic main point of config digest is that we can track in the logs when behavior in the server changed, then I got the idea to use it to skip reload if not needed to avoid redoing imports, etc... but for that I should also keep a digest of the contents of the certs used by the server to confirm whether do need a reload or not, so I think I will disable not doing reload when config is not changed for now

wallyqs (approved):
LGTM

PR #6612, opened 2025-03-07
Fix panic when subject transform has missing tokens

neilalexander:
It is likely that this is to do with a faulty service import, and we'll almost certainly be able to do better validation there too, but in the meantime we can and should avoid a panic with more rigorous bounds checking.


derekcollison (approved):
LGTM

PR #6633, opened 2025-03-10
v2.11: Disable max_buffered_msgs by default, bump max_buffered_size

wallyqs:
- Bump max_buffered_size to 256MB by default
- Disable max_buffered_msgs and make it opt-in

This will avoid running into IPQ len on simple benchmarks or use cases that the server handles fine, the max_buffered_msgs is enough to protect the memory usage in most cases.

### DIFF
--- a/server/stream.go
+++ b/server/stream.go
@@ -264,8 +264,8 @@ type ExternalStream struct {
 
 // For managing stream ingest.
 const (
-	streamDefaultMaxQueueMsgs  = 10_000
-	streamDefaultMaxQueueBytes = 1024 * 1024 * 128
+	streamDefaultMaxQueueMsgs  = -1
+	streamDefaultMaxQueueBytes = 256 * 1024 * 1024
 )
 
 // Stream is a jetstream stream of messages. When we receive a message internally destined


neilalexander (commented):
In principle I think this is OK, as we discussed last week, but I think it needs some kind of test to prove that the server copes and there aren't secondary effects.

Maybe copy `TestJetStreamRateLimitHighStreamIngest` into a new long-running test to prove it?

neilalexander:
Do we want to consider this for 2.12 or shall we close for now?

wallyqs:
Should have been for v2.11 I think... we can close for now and revisit when needed.

neilalexander:
What is the status of this?

wallyqs:
These defaults have been working better in scale tests without the servers struggling much and fixed issues in larger deployments with >100K connections. Reopened to see if we can try to change defaults for v2.12.

neilalexander:
Wondering if there's a fairer balance somewhere for defaults, i.e. increasing the default msgs to 100,000 from 10,000. That way messages over 1KB would be more likely to hit the max size limit (if left at 128MB) and messages under 1KB would be more likely to hit the max messages limit.

Also keep in mind with the bump to 256MB that this is _per stream_, not server-wide, so I'm a bit less comfortable with increasing that too much. 128MB should be more than fair.

neilalexander (approved):
LGTM

PR #6634, opened 2025-03-10
v2.11: Disable skipping reload on no config changes

wallyqs:
Enabling this would require taking the digest of TLS certs among other extra state at the moment.  We could potentially re-enable in the future so that authorization with no changes is skipped since that would be safe to do.


neilalexander (approved):
LGTM

PR #6638, opened 2025-03-11
(2.11.x) [log] Add TraceHeaders option

bruth:
When true, trace logs will no longer emit the payload of the message, only the headers if these are present.

An additional change, is not emitting the MSG_PAYLOAD log line if the payload is empty.

### DIFF
--- a/server/client.go
+++ b/server/client.go
@@ -1897,12 +1897,29 @@ func (c *client) flushSignal() {
 // Traces a message.
 // Will NOT check if tracing is enabled, does NOT need the client lock.
 func (c *client) traceMsg(msg []byte) {
-	maxTrace := c.srv.getOpts().MaxTracedMsgLen
-	if maxTrace > 0 && (len(msg)-LEN_CR_LF) > maxTrace {
+	opts := c.srv.getOpts()
+	maxTrace := opts.MaxTracedMsgLen
+	headersOnly := opts.TraceHeaders
+	suffix := LEN_CR_LF
+
+	// If TraceHeaders is enabled, extract only the header portion of the msg.
+	// If a header is present, it ends with an additional trailing CRLF.
+	if headersOnly {
+		msg, _ = c.msgParts(msg)
+		suffix += LEN_CR_LF
+	}
+
+	// Do not emit a log line for zero-length payloads.
+	l := len(msg) - suffix
+	if l <= 0 {
+		return
+	}
+
+	if maxTrace > 0 && l > maxTrace {
 		tm := fmt.Sprintf("%q", msg[:maxTrace])
-		c.Tracef("<<- MSG_PAYLOAD: [\"%s...\"]", tm[1:maxTrace+1])
+		c.Tracef("<<- MSG_PAYLOAD: [\"%s...\"]", tm[1:len(tm)-1])
 	} else {
-		c.Tracef("<<- MSG_PAYLOAD: [%q]", msg[:len(msg)-LEN_CR_LF])
+		c.Tracef("<<- MSG_PAYLOAD: [%q]", msg[:l])
 	}
 }
 
--- a/server/client_test.go
+++ b/server/client_test.go
@@ -1735,6 +1735,74 @@ func TestTraceMsg(t *testing.T) {
 	}
 }
 
+func TestTraceMsgHeadersOnly(t *testing.T) {
+	c := &client{}
+	// Enable message trace
+	c.trace = true
+
+	hdr := fmt.Sprintf(`NATS/1.0%sFoo: 1%s%s`, CR_LF, CR_LF, CR_LF)
+	hdr2 := fmt.Sprintf(`NATS/1.0%sFoo: 1%sBar: 2%s%s`, CR_LF, CR_LF, CR_LF, CR_LF)
+
+	cases := []struct {
+		Desc            string
+		Msg             []byte
+		Hdr             int
+		Wanted          string
+		MaxTracedMsgLen int
+	}{
+		{
+			Desc:            "payload only",
+			Msg:             []byte(`test\r\n`),
+			Hdr:             0,
+			Wanted:          _EMPTY_,
+			MaxTracedMsgLen: 0,
+		},
+		{
+			Desc:            "header only",
+			Msg:             []byte(hdr),
+			Hdr:             len(hdr),
+			Wanted:          ` - <<- MSG_PAYLOAD: ["NATS/1.0\r\nFoo: 1"]`,
+			MaxTracedMsgLen: 0,
+		},
+		{
+			Desc:            "with header and payload",
+			Msg:             []byte(fmt.Sprintf("%stest%s", hdr, CR_LF)),
+			Hdr:             len(hdr),
+			Wanted:          ` - <<- MSG_PAYLOAD: ["NATS/1.0\r\nFoo: 1"]`,
+			MaxTracedMsgLen: 0,
+		},
+		{
+			Desc:            "max length",
+			Msg:             []byte(hdr),
+			Hdr:             len(hdr),
+			Wanted:          ` - <<- MSG_PAYLOAD: ["NATS/1..."]`,
+			MaxTracedMsgLen: 6,
+		},
+		{
+			Desc:            "two headers max length",
+			Msg:             []byte(hdr2),
+			Hdr:             len(hdr2),
+			Wanted:          ` - <<- MSG_PAYLOAD: ["NATS/1.0\r\nFoo: 1\r\nBar..."]`,
+			MaxTracedMsgLen: 21,
+		},
+	}
+
+	for _, ut := range cases {
+		t.Run(ut.Desc, func(t *testing.T) {
+			c.srv = &Server{
+				opts: &Options{MaxTracedMsgLen: ut.MaxTracedMsgLen, TraceHeaders: true},
+			}
+			c.srv.SetLogger(&DummyLogger{}, true, true)
+			c.pa.hdr = ut.Hdr
+
+			c.traceMsg(ut.Msg)
+
+			got := c.srv.logging.logger.(*DummyLogger).Msg
+			require_Equal(t, string(ut.Wanted), got)
+		})
+	}
+}
+
 func TestClientMaxPending(t *testing.T) {
 	opts := DefaultOptions()
 	opts.MaxPending = math.MaxInt32 + 1
--- a/server/reload.go
+++ b/server/reload.go
@@ -139,7 +139,7 @@ func (t *traceOption) Apply(server *Server) {
 	server.Noticef("Reloaded: trace = %v", t.newValue)
 }
 
-// traceOption implements the option interface for the `trace` setting.
+// traceVersboseOption implements the option interface for the `trace_verbose` setting.
 type traceVerboseOption struct {
 	traceLevelOption
 	newValue bool
@@ -148,6 +148,17 @@ type traceVerboseOption struct {
 // Apply is a no-op because logging will be reloaded after options are applied.
 func (t *traceVerboseOption) Apply(server *Server) {
 	server.Noticef("Reloaded: trace_verbose = %v", t.newValue)
+}
+
+// traceHeadersOption implements the option interface for the `trace_headers` setting.
+type traceHeadersOption struct {
+	traceLevelOption
+	newValue bool
+}
+
+// Apply is a no-op because logging will be reloaded after options are applied.
+func (t *traceHeadersOption) Apply(server *Server) {
+	server.Noticef("Reloaded: trace_headers = %v", t.newValue)
 }
 
 // debugOption implements the option interface for the `debug` setting.
@@ -1231,6 +1242,8 @@ func (s *Server) diffOptions(newOpts *Options) ([]option, error) {
 		switch optName {
 		case "traceverbose":
 			diffOpts = append(diffOpts, &traceVerboseOption{newValue: newValue.(bool)})
+		case "traceheaders":
+			diffOpts = append(diffOpts, &traceHeadersOption{newValue: newValue.(bool)})
 		case "trace":
 			diffOpts = append(diffOpts, &traceOption{newValue: newValue.(bool)})
 		case "debug":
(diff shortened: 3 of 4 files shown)


ripienaar:
I like this, but defer for 2.11.x?

server/client.go
```diff
@@ -1897,12 +1897,28 @@ func (c *client) flushSignal() {
...
+
+	// If TraceHeaders is enabled, extract only the header portion of the msg.
+	// The suffix is zeroed since the header is only being included.
+	if headersOnly {
+		msg, _ = c.msgParts(msg)
+		suffix = 0
+	}
+
+	// Do not emit a log line for empty zero-length payloads.
+	if len(msg) == 0 {
```
kozlovic, line 1913: It should be `len(msg)-suffix` here.

Otherwise, if `headersOnly` is false, `msg` still contains `\r\n` for an empty message, so it will be logged as such:
```
[25208] 2025/03/11 10:05:35.092689 [TRC] 127.0.0.1:55199 - cid:9 - <<- [PUB foo 0]
[25208] 2025/03/11 10:05:35.092697 [TRC] 127.0.0.1:55199 - cid:9 - <<- MSG_PAYLOAD: [""]
```
If `headersOnly` is true, and has no headers, then it won't:
```
[25308] 2025/03/11 10:06:50.181919 [TRC] 127.0.0.1:55207 - cid:5 - <<- [PUB foo 0]
```

server/client.go
```diff
@@ -1897,12 +1897,28 @@ func (c *client) flushSignal() {
 // Traces a message.
 // Will NOT check if tracing is enabled, does NOT need the client lock.
 func (c *client) traceMsg(msg []byte) {
-	maxTrace := c.srv.getOpts().MaxTracedMsgLen
-	if maxTrace > 0 && (len(msg)-LEN_CR_LF) > maxTrace {
+	opts := c.srv.getOpts()
+	maxTrace := opts.MaxTracedMsgLen
+	headersOnly := opts.TraceHeaders
+	suffix := LEN_CR_LF
```
kozlovic, line 1903: We may not need that, see below...

server/client.go
```diff
@@ -1897,12 +1897,28 @@ func (c *client) flushSignal() {
...
+
+	// If TraceHeaders is enabled, extract only the header portion of the msg.
+	// The suffix is zeroed since the header is only being included.
+	if headersOnly {
+		msg, _ = c.msgParts(msg)
```
kozlovic, line 1908: `msg` here, will be actually the headers, but they still end with a `\r\n`.

server/mqtt.go
```diff
@@ -952,7 +952,20 @@ func (c *client) mqttParse(buf []byte) error {
 }
 
 func (c *client) mqttTraceMsg(msg []byte) {
-	maxTrace := c.srv.getOpts().MaxTracedMsgLen
+	opts := c.srv.getOpts()
+	maxTrace := opts.MaxTracedMsgLen
+	headersOnly := opts.TraceHeaders
```
kozlovic, line 957: I don't think this is relevant for MQTT messages since they don't have headers. This is invoked from the parsing of an MQTT client's pub message.

server/mqtt.go
```diff
@@ -952,7 +952,20 @@ func (c *client) mqttParse(buf []byte) error {
...
+
+	// If TraceHeaders is enabled, extract only the header portion of the msg.
+	if headersOnly {
+		msg, _ = c.msgParts(msg)
+	}
+
+	// Do not emit a log line for empty zero-length payloads.
+	if len(msg) == 0 {
```
kozlovic, line 965: You could however use the same fix here and check for `len(msg)-LEN_CR_LF` to decide that you want to not trace the payload.

kozlovic (changes requested):

server/client.go
kozlovic: wow... my update of this comment was completely ignored... let me post a new one.

kozlovic: That sucks, I lost a big edit of this comment, but let me try to summarize:

You need to take into account that `msg` will always have a trailing `\r\n`, either because `msg` points to the whole message ([hdr+]payload+\r\n) or the header section which also ends with extra `\r\n`. Example with your code:
```
[25796] 2025/03/11 10:23:42.557480 [TRC] 127.0.0.1:55280 - cid:5 - <<- [HPUB foo 23 28]
[25796] 2025/03/11 10:23:42.557495 [TRC] 127.0.0.1:55280 - cid:5 - <<- MSG_PAYLOAD: ["NATS/1.0\r\ntest: val\r\n\r\n"]
```

So if we agree to remove the trailing `\r\n` of a message payload, then when tracing headers only, we could also remove the trailing `\r\n` (there would still be the one from the last field). If you agree with that, we could have this code change:
```
index 0c11a436..4d6e1601 100644
--- a/server/client.go
+++ b/server/client.go
@@ -1900,25 +1900,24 @@ func (c *client) traceMsg(msg []byte) {
        opts := c.srv.getOpts()
        maxTrace := opts.MaxTracedMsgLen
        headersOnly := opts.TraceHeaders
-       suffix := LEN_CR_LF
 
        // If TraceHeaders is enabled, extract only the header portion of the msg.
        // The suffix is zeroed since the header is only being included.
        if headersOnly {
                msg, _ = c.msgParts(msg)
-               suffix = 0
        }
 
        // Do not emit a log line for empty zero-length payloads.
-       if len(msg) == 0 {
+       l := len(msg) - LEN_CR_LF
+       if l == 0 {
                return
        }
 
-       if maxTrace > 0 && (len(msg)-suffix) > maxTrace {
+       if maxTrace > 0 && l > maxTrace {
                tm := fmt.Sprintf("%q", msg[:maxTrace])
                c.Tracef("<<- MSG_PAYLOAD: [\"%s...\"]", tm[1:maxTrace+1])
        } else {
-               c.Tracef("<<- MSG_PAYLOAD: [%q]", msg[:len(msg)-suffix])
+               c.Tracef("<<- MSG_PAYLOAD: [%q]", msg[:l])
        }
 }
```
which would lead to:
```
[25864] 2025/03/11 10:25:23.679052 [TRC] 127.0.0.1:55287 - cid:5 - <<- [HPUB foo 23 28]
[25864] 2025/03/11 10:25:23.679080 [TRC] 127.0.0.1:55287 - cid:5 - <<- MSG_PAYLOAD: ["NATS/1.0\r\ntest: val\r\n"]
```
If anything, we could add suffix to remove yet the last `\r\n` in case `headersOnly=true`.

kozlovic: We may want to add some protection though and trust that the header is properly formed and has enough characters to remove (say if it were to not contain the last two `\r\n\r\n`). But without the check and if we want to remove the `\r\n` of the last field, code could be updated to this:
```
index 0c11a436..46356b6a 100644
--- a/server/client.go
+++ b/server/client.go
@@ -1906,19 +1906,20 @@ func (c *client) traceMsg(msg []byte) {
        // The suffix is zeroed since the header is only being included.
        if headersOnly {
                msg, _ = c.msgParts(msg)
-               suffix = 0
+               suffix += LEN_CR_LF
        }
 
        // Do not emit a log line for empty zero-length payloads.
-       if len(msg) == 0 {
+       l := len(msg) - suffix
+       if l == 0 {
                return
        }
 
-       if maxTrace > 0 && (len(msg)-suffix) > maxTrace {
+       if maxTrace > 0 && l > maxTrace {
                tm := fmt.Sprintf("%q", msg[:maxTrace])
                c.Tracef("<<- MSG_PAYLOAD: [\"%s...\"]", tm[1:maxTrace+1])
        } else {
-               c.Tracef("<<- MSG_PAYLOAD: [%q]", msg[:len(msg)-suffix])
+               c.Tracef("<<- MSG_PAYLOAD: [%q]", msg[:l])
        }
 }
```
which would produce:
```
[25959] 2025/03/11 10:27:02.971212 [TRC] 127.0.0.1:55296 - cid:5 - <<- [HPUB foo 23 28]
[25959] 2025/03/11 10:27:02.971241 [TRC] 127.0.0.1:55296 - cid:5 - <<- MSG_PAYLOAD: ["NATS/1.0\r\ntest: val"]
```

kozlovic: (just edited the check of `if l == 0` to `if l <= 0` to protect against bad msgs)

server/reload.go
```diff
@@ -139,7 +139,7 @@ func (t *traceOption) Apply(server *Server) {
 	server.Noticef("Reloaded: trace = %v", t.newValue)
 }
 
-// traceOption implements the option interface for the `trace` setting.
+// traceVersboseOption implements the option interface for the `trace_verbose` setting.
```
wallyqs, line 142:
```suggestion
// traceVerboseOption implements the option interface for the `trace_verbose` setting.
```

server/mqtt.go
kozlovic: Actually not even, since MQTT don't have this trailing `\r\n`, this is the reason I had to make a "copy" of the function to make it special for MQTT clients. So I would completely undo the changes in that function.

alexbozhenko:
@bruth Would it be possible to include a reproducible test plan in the PR description? It might be clear to experienced reviewers, but for those of us still ramping up, having explicit commands and expected output would help us understand the behavior change. It would also make it easier to communicate updates to customers and save us some cycles. Thanks!

bruth:
Thanks @kozlovic, will review shortly! Yes @alexbozhenko I will add a test for this, so the expected behavior is clear.

server/client.go
bruth: Understood, I did this intentionally. I will add a comment so its clear.

bruth:
@kozlovic I went with the second approach you suggested. I am going to add a test to exercise this.

server/client.go
```diff
@@ -1897,12 +1897,29 @@ func (c *client) flushSignal() {
...
+	if l <= 0 {
+		return
+	}
+
+	if maxTrace > 0 && l > maxTrace {
 		tm := fmt.Sprintf("%q", msg[:maxTrace])
 		c.Tracef("<<- MSG_PAYLOAD: [\"%s...\"]", tm[1:maxTrace+1])
 	} else {
-		c.Tracef("<<- MSG_PAYLOAD: [%q]", msg[:len(msg)-LEN_CR_LF])
+		c.Tracef("<<- MSG_PAYLOAD: [%q]", msg[:1])
```
kozlovic, line 1922: `l`, not `1`

bruth: Yep, thanks, quickly got that when I ran and wrote the test.

server/client.go
```diff
@@ -1917,9 +1917,9 @@ func (c *client) traceMsg(msg []byte) {
 
 	if maxTrace > 0 && l > maxTrace {
 		tm := fmt.Sprintf("%q", msg[:maxTrace])
-		c.Tracef("<<- MSG_PAYLOAD: [\"%s...\"]", tm[1:maxTrace+1])
+		c.Tracef("<<- MSG_PAYLOAD: [\"%s...\"]", tm[1:len(tm)-1])
```
bruth, line 1920: Changed this since when converting to the quoted form, `\r\n` is escaped and thus four characters.

server/client_test.go
```diff
@@ -1735,6 +1735,74 @@ func TestTraceMsg(t *testing.T) {
 	}
 }
 
+func TestTraceMsgHeadersOnly(t *testing.T) {
```
bruth, line 1738: @alexbozhenko Here is the new test. When `TraceHeaders` is true, it will strip out the payload of the message. You can see the first table test results in the empty string (since it is not logged at all). Where as the third test that includes a header and payload is still logged, but the payload is stripped.

server/client_test.go
```diff
@@ -1735,6 +1735,74 @@ func TestTraceMsg(t *testing.T) {
...
+		Msg             []byte
+		Hdr             int
+		Wanted          string
+		MaxTracedMsgLen int
+	}{
+		{
+			Desc:            "payload only",
+			Msg:             []byte(`test\r\n`),
+			Hdr:             0,
+			Wanted:          "",
```
kozlovic, line 1757: Derek would say to replace `""` with `_EMPTY` (applies to all in this test)

kozlovic (approved):
nit about use of `""` instead of `_EMPTY_`, but otherwise LGTM.

wallyqs:
It is going to be more helpful for investigations if we trace the headers when they are delivered as well with an option, that is the current limitation of tracing (since both payloads and headers are not shown on the receiving side).

```
[42994] 2025/03/13 10:32:32.461243 [TRC] 127.0.0.1:52590 - cid:6 - <<- [SUB _INBOX.Es5igIqrhJdEt9YeaqhPSM.GX7gwt9F  1]
[42994] 2025/03/13 10:32:32.461259 [TRC] 127.0.0.1:52590 - cid:6 - <<- [HPUB foo _INBOX.Es5igIqrhJdEt9YeaqhPSM.GX7gwt9F 22 26]
[42994] 2025/03/13 10:32:32.461272 [TRC] 127.0.0.1:52590 - cid:6 - <<- MSG_PAYLOAD: ["NATS/1.0\r\nfoo: bar\r\n\r\nasdf"]
[42994] 2025/03/13 10:32:32.461289 [TRC] 127.0.0.1:52586 - cid:5 - ->> [HMSG foo 1 _INBOX.Es5igIqrhJdEt9YeaqhPSM.GX7gwt9F 22 26]
```

bruth:
> It is going to be more helpful for investigations if we trace the headers when they are delivered as well with an option

I am not opposed, but to be clear, this is net new? We can do this in a follow-up PR

kozlovic (approved):
LGTM

server/client_test.go
alexbozhenko: thank you, @bruth !

neilalexander:
Please rebase on top of latest main and will queue up for merge, thanks!

server/client.go
```diff
@@ -10,9 +10,7 @@
 // WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 // See the License for the specific language governing permissions and
 // limitations under the License.
-
```
bruth, line 13: I have no idea what happened with this file.. I simply rebased locally. Will fix

PR #6648, opened 2025-03-11
Reduce Travis CI instance size

neilalexander:


neilalexander:
Note to self: `TestJetStreamClusterParallelConsumerCreation` seems to struggle on this instance size.

PR #6668, opened 2025-03-13
[FIXED] Config Reload: Possible panic when adding dedicated route

kozlovic:
When the cluster is configured without any pinned account and the system account is disabled, the first time an account would be added to the `accounts` lister in the `cluster{}` block and the configuration would be reloaded, the remote server would panic.

This would not happen if the system account was not disabled.


wallyqs (approved):

neilalexander (approved):
LGTM, thanks!

PR #6677, opened 2025-03-14
Fixed test `TestJetStreamClusterMetaRecoveryLogic` causing data race

kozlovic:
This was because it was calling `streamAssignment` without proper JS locking.


wallyqs (approved):

neilalexander (approved):
LGTM

PR #6680, opened 2025-03-14
De-flake subject delete marker tests

neilalexander:
These tests could flake because the 2 second pause was right on the boundary between the 1 second TTL and the 1 second max-age limit. Moving it by half a second resolves the flake.


MauriceVanVeen (approved):
LGTM

PR #6682, opened 2025-03-14
(v2.11) [REMOVED] Unused `RemoveSelfReference()` function

kozlovic:
This was introduced long ago by an external contributor but is no longer being used in the project. Since this was more for the server cluster configuration, the risk of breaking user apps is almost nil.

The tests for this API have been removed too.


ripienaar:
Is it not needed to be called anymore? I call it when using server embedded but happy to remove if not needed anymore because it’s handled in another way now?

kozlovic:
> Is it not needed to be called anymore? I call it when using server embedded but happy to remove if not needed anymore because it’s handled in another way now?

@ripienaar It has been removed long ago, see [PR](https://github.com/nats-io/nats-server/pull/578). The server detects routes to self and ignore them. Additional work was done to support hostname resolution that would give several IPs, see [PR](https://github.com/nats-io/nats-server/pull/1590).

ripienaar:
Great happy to remove it then for 2.11

PR #6690, opened 2025-03-18
Fix timing inconsistency between `Nats-TTL` and `MaxAge` timers

neilalexander:
We would only reset the timer based on the newly submitted TTL value if the timer was already running, but that could have been set to `MaxAge` based on no messages being left in the stream if it was configured, delaying the next clean-up. Instead just kick the timer anyway with each new TTL message.


MauriceVanVeen (approved):
LGTM

derekcollison (approved):
LGTM

PR #6692, opened 2025-03-19
Updates to NATS Server Tests GHA pipeline

neilalexander:
Currently this pipeline is disabled, but it makes the following readiness changes:

- Updates the concurrency group to auto-cancel only if not `main` or a release branch;
- Handle builds on pushes and pull requests;
- Use the default GHA runners if no overrides are specified in the repository variables;
- Use the scripts from Travis rather than having yet another place that we need to manage the `go test` commands.


MauriceVanVeen (approved):
LGTM

PR #6693, opened 2025-03-19
Fix TestLongClusterStreamOrphanMsgsAndReplicasDrifting

mprimi:
Test was passing/failing depending on lucky/unlucky timing.

**Issue**:

Each of the subtests of was leaking the goroutine responsible for restarting servers mid-execution.

The leaked goroutine was left running after each sub-test completed. Server configuration files would get cleaned up after test completion, leading to the routine to panic, e.g.:

```
panic: Error processing configuration file: open /tmp/[...]/TestLongClusterStreamOrphanMsgsAndReplicasDriftingR3M2222357531/002/2244942669: no such file or directory
```

The panic would confuse the Go test runtime (which had already moved on to the next test), producing errors such as:
```
=== FAIL: server TestLongClusterStreamOrphanMsgsAndReplicasDrifting (unknown)
```

**Change overview**:

 - Track the restart goroutine using the shared WaitGroup
 - Shut down the goroutine earlier using the global context

N.B. test semantics are slightly altered.
 - Before: restart subroutine *may* be running during stream checks
 - After: restart subroutine is not running during stream checks


wallyqs (approved):
LGTM

MauriceVanVeen (approved):
LGTM

neilalexander (approved):
LGTM

PR #6703, opened 2025-03-20
add OpenSSF scorecard

alexbozhenko:
https://securityscorecards.dev/#how-it-works


neilalexander:
Let's talk about this one internally first with @philpennock, can reopen later if needs be.

PR #6704, opened 2025-03-20
[IMPROVED] Services: Now possible to import same subject from diff accounts

kozlovic:
Previously, if 2 accounts exported the same subject, another account would not be able to import the 2 accounts with the same subject. This would produce an error saying that there was a duplicate.

We are changing the internal structure to store the service imports which is now allowing such services import definition.

See test `TestAccountMultipleServiceImportsWithSameSubjectFromDifferentAccounts` for an example.


derekcollison (approved):
LGTM

PR #6708, opened 2025-03-21
De-flake unit tests

neilalexander:
Also removes the failures from one test that fails purely based on timing info, since it's totally unpredictable, and we already track consumer create times in Gauge.


derekcollison (approved):
LGTM

PR #6730, opened 2025-03-26
[FIXED] JetStream clustering: observer incorrectly elected leader

kozlovic:
In some leafnodes setups with shared system account, the leafnode is supposed to set itself as observer and step down so that only servers in the hub cluster can be elected leaders. A test was flapping but it turns out that there was a race condition that could cause a server that has chosen to be in observer mode and step down to still end-up being elected.

Added a dedicated test that was able to reproduce the issue more easily than the leafnode test.


derekcollison (approved):
LGTM

PR #6731, opened 2025-03-26
[IMPROVED] Config Parsing consistency with authentication timeout option

johnweldon:
Improve the config parsing to make `authorization` -> `timeout` parsing consistent with the `tls_timeout` option

### DIFF
--- a/server/opts.go
+++ b/server/opts.go
@@ -2611,12 +2611,20 @@ func parseLeafAuthorization(v any, errors *[]error) (*authorization, error) {
 			}
 			auth.nkey = nk
 		case "timeout":
-			at := float64(1)
+			at := float64(0)
 			switch mv := mv.(type) {
 			case int64:
 				at = float64(mv)
 			case float64:
 				at = mv
+			case string:
+				d, err := time.ParseDuration(mv)
+				if err != nil {
+					return nil, &configErr{tk, fmt.Sprintf("error parsing leafnode authorization config, 'timeout' %s", err)}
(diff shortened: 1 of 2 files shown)


server/opts.go
```diff
@@ -4131,12 +4131,20 @@ func parseAuthorization(v any, errors *[]error) (*authorization, error) {
 		case "token":
 			auth.token = mv.(string)
 		case "timeout":
-			at := float64(1)
+			at := float64(0)
```
johnweldon, line 4134: Use `0` as default, so that in `setBaselineOptions()` the [call](https://github.com/nats-io/nats-server/blob/8f8fe68d4c7d9ec72fa33181a4b376eb02a095ec/server/opts.go#L5455) to `getDefaultAuthTimeout` will work as expected.

server/opts_test.go
```diff
@@ -3475,3 +3475,117 @@ func TestProcessConfigString(t *testing.T) {
 		})
 	}
 }
+
+func TestAuthorizationTimeoutConfigParsing(t *testing.T) {
+	type testCase struct {
+		name                string
+		config              string
+		expectParsed        float64
+		expectRunning       float64
```
johnweldon, line 3484: This nuance is due to the fact that the config parser produces a certain value, but after the server is started using those values, the running option value could be different.

wallyqs (approved):
LGTM

neilalexander (approved):
LGTM

PR #6740, opened 2025-03-28
CI: Only run race detector on development branches

neilalexander:
This will keep the race detector runs on development branches, but it will no longer run on `main`, release branches or `pull_request` runs.


MauriceVanVeen (approved):
LGTM

PR #6749, opened 2025-03-31
[FIXED] Priority group waiting requests causing tight loop in nextWaiting

piotrpio:
When there are multiple pending requests for a consumer with priority group set, and some of the requests are not receiving (due to priority policy), it is possible to enter a tight loop in `consumer.nextWaiting()` if another pull request is recycled during execution.

### DIFF
--- a/server/consumer.go
+++ b/server/consumer.go
@@ -3596,7 +3596,7 @@ func (o *consumer) nextWaiting(sz int) *waitingRequest {
 		priorityGroup = o.cfg.PriorityGroups[0]
 	}
 
-	lastRequest := o.waiting.tail
+	numCycled := 0
 	for wr := o.waiting.peek(); !o.waiting.isEmpty(); wr = o.waiting.peek() {
 		if wr == nil {
 			break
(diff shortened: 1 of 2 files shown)


server/consumer.go
```diff
@@ -3650,7 +3650,8 @@ func (o *consumer) nextWaiting(sz int) *waitingRequest {
 					// If we have a match, we do nothing here and will deliver the message later down the code path.
 				} else if wr.priorityGroup.Id == _EMPTY_ {
 					o.waiting.cycle()
-					if wr == lastRequest {
+					numCycled++
+					if numCycled >= o.waiting.len() {
```
piotrpio, line 3654: @neilalexander I was wondering if it's possible for `o.waiting` to be modified with new requests during the execution of `nextWaiting`? I don't think so, as I see that everything is done under lock in `loopAndGatherMsgs` but I'd like to make sure comparing to `o.waiting.len()` in here is a valid approach.

neilalexander: I think `o.waiting.len()` should be safe, AFAIK the only places where new requests are added are under the consumer lock.

neilalexander (approved):
LGTM

PR #6754, opened 2025-04-02
FreeBSD: PSE non-CGO for arm64 too

philpennock:
For the pse support for FreeBSD:
 * switch the CGO implementation to require the cgo build tag, rather than per-architecture exemptions
 * rather than have N files, 1 for each arch, to set the constants for that one arch, switch the existing non-CGO file:
   + Be named without an arch constraint (use `_sysctl` instead)
   + Use arch build tags
   + Use an init-time runtime.GOARCH dispatch to set the variables needed.

### DIFF
--- a/server/pse/pse_freebsd_cgo.go
+++ b/server/pse/pse_freebsd_cgo.go
@@ -11,7 +11,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-//go:build !amd64
+//go:build cgo && freebsd
 
 package pse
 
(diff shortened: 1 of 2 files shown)


derekcollison (approved):
LGTM

philpennock:
* f3d3a0a to 6c3f2d7: rebased (as requested by Neil) and squashed
* 6c3f2d7 to 664068b: _no code changes_: updated the comment in pse_freebsd_sysctl.go to more accurately describe the build-tag situation.  The prior text was before the facepalm discover which led to the second commit in the first pass.

PR #6758, opened 2025-04-03
[FIXED] Fix for a bug in recovering ttls without the hashwheel info.

derekcollison:
We had a bug where we were only looking at the first msg since we did not offset buf by index.


neilalexander (approved):
LGTM

PR #6768, opened 2025-04-08
De-flake `TestJetStreamClusterParallelConsumerCreation`

neilalexander:
Using the new JetStream API from the client increases the reliability of this test as it removes the race conditions in the client between the consumer info and the consumer create calls (now fixed by the presence of the consumer actions).


neilalexander:
Going to hold off until nats.go v1.41.1 which to remove the `golang.org/x/text` dependency, see https://github.com/nats-io/nats.go/pull/1849.

MauriceVanVeen (approved):
LGTM

PR #6786, opened 2025-04-10
Improve stream overlap checking with JS API, system subjects

neilalexander:
There are still some cases where stream subjects can break the JetStream API, i.e. creating a stream with `*.>`. This works because it isn't caught by `subjectIsSubsetMatch("*.>", "$JS.>")`. Using `SubjectsCollide` is much more effective as it does not miss where the matching wildcard is a superset of the reserved wildcards.


derekcollison (approved):
LGTM

PR #6791, opened 2025-04-11
Report `GOMAXPROCS` and `GOMEMLIMIT` in `ServerStats`

### LINKED ISSUE #6672: Add GOMAXPROCS and GOMEMLIMIT to ServerStats
### Proposed change

ServerStats shows the memory usage and host CPU cores. This information is useful but is not the full picture of what the server is configured to use. Please add `GOMAXPROCS` and `GOMEMLIMIT` to [ServerStats](https://github.com/nats-io/nats-server/blob/e38b87feccbe00520400a4194a8cacb759747cc4/server/events.go#L362-L379) so it can be used in places like `nats server ls` or with `$SYS.REQ.SERVER.PING`. 

### Use case

Having the GO limits exposed can help with the diagnosis of resource contention issues and be used to identify configuration issues. [GOMAXPROCS is available in varz](https://github.com/nats-io/nats-server/blob/db50019da4e0b7737c3d9aa077f4bacdcd5f9ac6/server/monitor.go#L1217), and would make a valuable addition to ServerStats.

### Contribution

_No response_

neilalexander:
Fixes #6672.

### DIFF
--- a/server/events.go
+++ b/server/events.go
@@ -21,9 +21,11 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"math"
 	"math/rand"
 	"net/http"
 	"runtime"
+	"runtime/debug"
 	"strconv"
 	"strings"
 	"sync"
@@ -376,6 +378,8 @@ type ServerStats struct {
 	Gateways           []*GatewayStat      `json:"gateways,omitempty"`
 	ActiveServers      int                 `json:"active_servers,omitempty"`
 	JetStream          *JetStreamVarz      `json:"jetstream,omitempty"`
+	MemLimit           int64               `json:"gomemlimit,omitempty"`
+	MaxProcs           int                 `json:"gomaxprocs,omitempty"`
 }
 
 // RouteStat holds route statistics.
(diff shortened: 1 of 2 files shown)


server/events.go
```diff
@@ -376,6 +378,8 @@ type ServerStats struct {
 	Gateways           []*GatewayStat      `json:"gateways,omitempty"`
 	ActiveServers      int                 `json:"active_servers,omitempty"`
 	JetStream          *JetStreamVarz      `json:"jetstream,omitempty"`
+	MemLimit           int64               `json:"gomemlimit,omitempty"`
```
ripienaar, line 381: while MaxProcs is in Varz this one isn't, be great to add it to vz also

server/events.go
```diff
@@ -821,6 +825,10 @@ func (s *Server) updateServerUsage(v *ServerStats) {
 	var vss int64
 	pse.ProcUsage(&v.CPU, &v.Mem, &vss)
 	v.Cores = runtime.NumCPU()
+	v.MaxProcs = runtime.GOMAXPROCS(-1)
+	if mm := debug.SetMemoryLimit(-1); mm < math.MaxInt64 {
```
derekcollison, line 829: Does this actually set a limit or just badly named get operation?

ripienaar (approved):
LGTM

server/events.go
neilalexander: Like `GOMAXPROCS`, passing a negative is a no-op. From https://cs.opensource.google/go/go/+/go1.24.2:src/runtime/debug/garbage.go;l=234:
> SetMemoryLimit returns the previously set memory limit. A negative input does not adjust the limit, and allows for retrieval of the currently set memory limit.

derekcollison (approved):
LGTM

PR #6794, opened 2025-04-13
[monitoring] Fix jsz offset bug

bruth:
The core issue was that the account names were not being sorted unless offset was non-zero. However, the first page will always be an offset of zero and thus there is no determinism of the order.

Now the accounts are sorted and the logic is simplified to handle both single account and multiple account queries.

### DIFF
--- a/server/monitor_test.go
+++ b/server/monitor_test.go
@@ -5010,11 +5010,24 @@ func TestMonitorJsz(t *testing.T) {
 	})
 	t.Run("offset-stable", func(t *testing.T) {
 		for _, url := range []string{monUrl1, monUrl2} {
-			info1 := readJsInfo(url + "?accounts=true&offset=1&limit=1")
+			info1 := readJsInfo(url + "?accounts=true&offset=0&limit=1")
 			if len(info1.AccountDetails) != 1 {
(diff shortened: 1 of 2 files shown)


server/monitor_test.go
```diff
@@ -5010,11 +5010,24 @@ func TestMonitorJsz(t *testing.T) {
...
 			if len(info1.AccountDetails) != 1 {
 				t.Fatalf("expected one account to be returned by %s but got %v", url, info1)
 			}
-			info2 := readJsInfo(url + "?accounts=true&offset=1&limit=1")
+			info2 := readJsInfo(url + "?accounts=true&offset=0&limit=1")
+			if len(info2.AccountDetails) != 1 {
+				t.Fatalf("expected one account to be returned by %s but got %v", url, info2)
+			}
+			if info1.AccountDetails[0].Name != info2.AccountDetails[0].Name {
+				t.Fatalf("absent changes, same offset should result in same account but gut: %v %v",
```
wallyqs, line 5022:
```suggestion
				t.Fatalf("absent changes, same offset should result in same account but got: %v %v",
```

bruth:
@neilalexander bumping this for your review on the upcoming patch release.

neilalexander (approved):
LGTM

Issue #6797, opened 2025-04-15
JetStream: Messages >32MB cause corrupt index and malformed/corrupt error, despite server limits allowing it


wartis:
### Observed behavior

When sending a message of exactly 32MB or larger using .jetStream().publish, I get the following server-side error during message lookup:

```
[WRN] Filestore [STR_1] indexCacheBuf corrupt record state: dlen 33554452 slen 12 index 30 rl 33554474 lbuf 33554504
[ERR] Received an error looking up message for consumer '$G > TEST_STREAM > TEST_STREAM_DURABLE': malformed or corrupt message
```

The consumer cannot receive the message, and JetStream behaves as if the message index is corrupted. The issue occurs consistently for payloads ≥ 32MB, even though they're below `max_payload`. 
The same code and message size work correctly in **NATS v2.2**, suggesting a regression in later versions.

### Expected behavior

Messages up to `max_payload` (64MB) should be accepted and correctly stored/delivered by JetStream.

### Server and client version

-   **NATS Server version (failing):** 2.11, 2.10
  
-   **NATS Server version (working):** 2.2.x 

-   **Client:** Java (also reproducible with Go)

-   **Client Library Version:** nats.java 2.21, 2.20, 2.11

### Host environment

**Environment:**

-   **Deployment:** Docker

-   **OS:** Ubuntu 24.04.2 LTS

----------

**JetStream Configuration:**

```hcl
jetstream {
  max_mem_store: 4GB
  max_file_store: 100GB
  store_dir: /data/jetstream
}

```

**Limits:**

```
max_payload: 64MB
max_pending: 64MB

```

### Steps to reproduce

Below is a simplified version of the code I’m using to send random byte arrays as test messages:

```java
public void sendRandomBytesRequest(String subject, int lengthInMb) throws JetStreamApiException, IOException {
    int lengthInBytes = lengthInMb * 1024 * 1024;
    byte[] data = new byte[lengthInBytes];
    new Random().nextBytes(data);

    sendRequest(subject, data);
}

public void sendRequest(String subject, byte[] data) throws JetStreamApiException, IOException {
    NatsMessage message = NatsMessage.builder().subject(subject).data(data).build();
    connection.jetStream().publish(message, createDefaultPublisherOptions(streamName).build());
}

```

**Demo method:**

```java
sender.sendRandomBytesRequest(subject, 10); // works
sender.sendRandomBytesRequest(subject, 20); // works
[...]

PR #6798, opened 2025-04-15
Avoid filestore corruption by rejecting large publishes into JetStream

neilalexander:
We have an effective and seemingly undocumented limit of 32MB for messages published into JetStream as `indexCacheBuf` in the filestore rejects any records that exceed this length. We did not catch this on publish however, and then `indexCacheBuf` would later error when consuming etc.

Related: #6797


derekcollison (approved):
LGTM

PR #6808, opened 2025-04-18
Redact auth tokens in trace logs for enhanced security

### LINKED ISSUE #6666: NATS Server can log secret information on permissions violations
### Observed behavior

When a NATS client has successfully connected to the NATS server and is granted certain permissions, if they attempt to take actions that are outside the scope of their permissions, they will get something like a "Permission Violation" error. At the same time, the NATS server will log this permissions violation and includes the "auth user" that encountered a permissions violation. The [auth user] can include secret information depending on how the client originally connected. 

Based on my understanding of the potential things that can be logged, the Nkey option, Username option, and JWT option likely shouldn't have secret information. The Nkey that a client presents as part of its request to connect (I think) is the public half of the Nkey pair. It also looks like the JWT option explicitly calls out putting `pubKey` in the output. 

Things get dicey with the Token user though.  In my use case, I am using auth callouts and the auth callout service only cares about the information stuffed into the Token option. NATS clients in my environment put their credentials into the Token option. If one of those clients commits a permissions violation, their entire Token (which includes their credentials) are logged by the NATS server.

[auth user]:  https://github.com/nats-io/nats-server/blob/0efea661bac82ec5de6da3b3035b5257947875a0/server/client.go#L6081

### Expected behavior

Expect secret information is not logged by the NATS server. 

We are working around this by now having our clients also specify a username (that is unused by our auth callout service) because that is higher in the switch statement that determines what gets logged. But this is brittle. Any changes to the switch statement inside of the `getAuthUser` method could land us back at where we were originally. 

### Server and client version

- nats-server: `2.10.18`
- nats.go: `v1.37.0`
- nats.net: `v2.5.10`

### Host environment

_No response_

### Steps to reproduce

- Connect to the nats server using the token option
- Violate the permissions that you are granted
- See the NATS server will log your token

aricart:
Updated trace functionality to redact auth_tokens in addition to passwords in logs to prevent sensitive information leakage. Refactored logic to use a generalized `redact` function and added extensive tests to ensure proper redaction in various scenarios.

Fix #6666

### DIFF
--- a/server/client.go
+++ b/server/client.go
@@ -1991,20 +1991,26 @@ func (c *client) processErr(errStr string) {
 
 // Password pattern matcher.
 var passPat = regexp.MustCompile(`"?\s*pass\S*?"?\s*[:=]\s*"?(([^",\r\n}])*)`)
+var tokenPat = regexp.MustCompile(`"?\s*auth_token\S*?"?\s*[:=]\s*"?(([^",\r\n}])*)`)
 
-// removePassFromTrace removes any notion of passwords from trace
+// removeSecretsFromTrace removes any notion of passwords/tokens from trace
 // messages for logging.
-func removePassFromTrace(arg []byte) []byte {
-	if !bytes.Contains(arg, []byte(`pass`)) {
-		return arg
+func removeSecretsFromTrace(arg []byte) []byte {
+	buf := redact("pass", passPat, arg)
+	return redact("auth_token", tokenPat, buf)
+}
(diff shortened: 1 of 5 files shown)


derekcollison (approved):
LGTM

derekcollison:
@wallyqs and @ripienaar could you take a look as well?

ripienaar (approved):
LGTM

PR #6816, opened 2025-04-23
[FIXED] Add total to JSZ and revert behavior

bruth:
In #6794, a limit default was enforced. However there was no indication of size in the response. The new behavior broke clients (Surveyor) that relied on getting everything in one request.

This change reverts the behavior to allow for everything to be returned in one request (fine for small deployments), but it adds a Total field on the response so clients can properly determine if subsequent requests are necessary.

### DIFF
--- a/server/monitor.go
+++ b/server/monitor.go
@@ -2875,20 +2875,19 @@ type MetaClusterInfo struct {
 
 // JSInfo has detailed information on JetStream.
 type JSInfo struct {
-	ID       string          `json:"server_id"`
-	Now      time.Time       `json:"now"`
-	Disabled bool            `json:"disabled,omitempty"`
-	Config   JetStreamConfig `json:"config,omitempty"`
-	Limits   *JSLimitOpts    `json:"limits,omitempty"`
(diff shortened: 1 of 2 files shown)


server/monitor.go
```diff
@@ -2875,20 +2875,19 @@ type MetaClusterInfo struct {
...
+	ID             string           `json:"server_id"`
+	Now            time.Time        `json:"now"`
+	Disabled       bool             `json:"disabled,omitempty"`
+	Config         JetStreamConfig  `json:"config,omitempty"`
+	Limits         *JSLimitOpts     `json:"limits,omitempty"`
+	Streams        int              `json:"streams"`
+	Consumers      int              `json:"consumers"`
+	Messages       uint64           `json:"messages"`
+	Bytes          uint64           `json:"bytes"`
+	Meta           *MetaClusterInfo `json:"meta_cluster,omitempty"`
```
bruth, line 2888: All this is just reordering for formatting alignment. The embedded structure is moved to the top.

neilalexander (approved):
LGTM

PR #6817, opened 2025-04-23
[FIXED] JetStream: message not accepted after expired account is updated

### LINKED ISSUE #6806: jetstream for an expired account doesn't restart when the account is updated
### Observed behavior

If you have an operator mode setup, and an account has jetstream assets these assets will stop when the account expires (this behaviour is correct). However if the account is updated, the server doesn't restart the assets unless restarted (reload has no effect).

### Expected behavior

updating the account in the resolver should re-start the JetStream assets or at the very least a reload of the server should make that work.

### Server and client version

2.11.1
clients latest

### Host environment

_No response_

### Steps to reproduce

[test_js_expires.zip](https://github.com/user-attachments/files/19799224/test_js_expires.zip)

Unzip the test project, will need `nsc` and `nats` cli, and node  installed.
```bash
cd test_js_expires
npm install

# will setup an operator, an account called SYS, and an account called A
# The initial setup will have a normal non-expiring account which will expire later
# This command will start up a server with those assets, and create the stream/consumer
# that are used for the scripts
# the end of the script will print some useful urls, and nats cli commands
./env.sh

# in one terminal
node producer.js

# update the account so that it expires in 1m (you can wait for the producer to bail
# out with bad credentials
./expire.sh

# when dead you can use the `nats server account info A... --json | jq .expired` 
# command that was printed by the script to verify that the JWT is expired

nats server account info A... --json | jq .expired

# rejuvenate the account
./expire.sh
# validate that the JWT is NOT expired
nats server account info A... --json | jq .expired

# restarting the producer fails, jetstream not available (the bug here)
node producer.js


```

kozlovic:
After an account expires, it is expected that an existing or new connection for this account is closed/rejected. However, if the account's expiration time is updated, a new connection should be able to publish into a stream.

This was not the case because on account expiration, all its clients were closed, including internal ones, which would remove the subject interest for the stream.

Resolves #6806

### DIFF
--- a/server/accounts.go
+++ b/server/accounts.go
@@ -3148,7 +3148,9 @@ func (a *Account) expiredTimeout() {
 	// Collect the clients and expire them.
 	cs := a.getClients()
 	for _, c := range cs {
-		c.accountAuthExpired()
+		if !isInternalClient(c.kind) {
+			c.accountAuthExpired()
+		}
 	}
 }
 
--- a/server/jetstream_jwt_test.go
+++ b/server/jetstream_jwt_test.go
@@ -22,6 +22,7 @@ import (
 	"net/http"
 	"net/http/httptest"
 	"strings"
+	"sync"
 	"sync/atomic"
 	"testing"
 	"time"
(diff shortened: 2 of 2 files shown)


server/accounts.go
```diff
@@ -3148,7 +3148,9 @@ func (a *Account) expiredTimeout() {
 	// Collect the clients and expire them.
 	cs := a.getClients()
 	for _, c := range cs {
-		c.accountAuthExpired()
+		if !isInternalClient(c.kind) {
```
kozlovic, line 3151: @derekcollison I am not sure if this should be limited to JETSTREAM only or to all internal clients, or if this is even the right approach...

neilalexander: This probably is the right approach, as I think NRG nodes run with `SYSTEM` client kinds.

derekcollison: Agree should be fine. @neilalexander do we keep type even if in user account?

derekcollison (approved):
LGTM

derekcollison:
> more of a question - this would keep jetstream running, but would bounce the normal clients. If this is the case that means that jetstream is still consuming resources. IMHO if the account is expired, all clients should be stopped (This behaviour may even be an useful tool for some scenarios). The question is what happens when the account is updated - bouncing the server is that part that I think shouldn't be required.

I do not think we should shutdown resources at that point.

server/accounts.go
neilalexander: Yes, it's kind `SYSTEM` either way.

PR #6823, opened 2025-04-23
Fix panic in `monitorConsumer` if meta nil

neilalexander:
The equivalent in `monitorStream` already has this exact check but it never made its way into `monitorConsumer` seemingly. Buildkite runs picked up a panic here.

### DIFF
--- a/server/jetstream_cluster.go
+++ b/server/jetstream_cluster.go
@@ -4732,12 +4732,12 @@ func (o *consumer) raftNode() RaftNode {
 }
 
 func (js *jetStream) monitorConsumer(o *consumer, ca *consumerAssignment) {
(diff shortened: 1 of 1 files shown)


wallyqs (approved):
LGTM

server/jetstream_cluster.go
```diff
@@ -4737,7 +4737,11 @@ func (js *jetStream) monitorConsumer(o *consumer, ca *consumerAssignment) {
 
 	defer o.clearMonitorRunning()
 
-	if n == nil {
+	js.mu.RLock()
```
derekcollison, line 4740: func (js *jetStream) getMetaGroup() RaftNode

derekcollison (approved):
LGTM - Small nit on using function to get meta NRG

PR #6825, opened 2025-04-24
[FIXED] Async tombstones on PurgeEx

derekcollison:
Make sure that we do async flush for multiple tombstones.
Also added in short circuit for PurgeEx when we know we have no more matches for explicit subjects.


neilalexander (approved):
LGTM

MauriceVanVeen (approved):
LGTM

PR #6843, opened 2025-04-29
GSL: Fix subject tree intersection to match literals when followed by additional wildcards

neilalexander:
This is the same fix as #6827 but this time for the generic sublist.


neilalexander:
Will roll this fix into another PR as there's a regression that also needs addressing.

PR #6849, opened 2025-04-29
[FIXED] Monitoring: Issue with Connz filters "cid" and "state=all"

### LINKED ISSUE #6839: When filtering connections by ConnState = all only closed clients are returned
### Observed behavior

When the monitoring endpoint is invoked by connection id and state = all, connection is returned only if closed

### Expected behavior

The connection requested by cid should be returned whether open or closed

### Server and client version

NATS server 2.10
NATS CLI 0.1.1

### Host environment

_No response_

### Steps to reproduce

_No response_

kozlovic:
When filtering on a given connection ID, if the user also filters on the state being "all" and the connection is opened, it would not be returned. It would be if the connection was closed.

Resolves #6839

### DIFF
--- a/server/monitor.go
+++ b/server/monitor.go
@@ -340,7 +340,17 @@ func (s *Server) Connz(opts *ConnzOptions) (*Connz, error) {
 
 	// Search by individual CID.
 	if cid > 0 {
-		if state == ConnClosed || state == ConnAll {
+		// Let's first check if user also selects on ConnOpen or ConnAll
+		// and look for opened connections.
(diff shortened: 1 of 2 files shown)


kozlovic:
@neilalexander I sent you a separate email regarding flappers that prevent me from getting a green build. I think that they are separate issues that should be addressed in a different PR. Thanks!

derekcollison (approved):
LGTM

neilalexander (approved):
LGTM

PR #6852, opened 2025-04-30
Standard exit code when shutdown via signal

mprimi:
State:
 🟧 Code: POC for early feedback 
 🟥 Tests: missing

---

When receiving <signal>, exit with code 128+<signal>.

Pros:
 - Implement de-facto convention
 - Easier integration for any tool/script that expects this convention
 - Easier investigationof "why did server shutdown?" mystery scenarios

Cons:
 - May break some existing integration/tools/script (Hyrum's Law)

Sources:
 - https://tldp.org/LDP/abs/html/exitcodes.html
 - https://en.wikipedia.org/wiki/Exit_status#Shell_and_scripts
 - https://www.baeldung.com/linux/status-codes#fatal-error-signal

---

Behavior without this change:
 - Exit 1 in case of error
 - Exit 0 in case of signal

Target behavior:
 - Exit 1 in case of error
 - Exit 128+signal in case of signal and no shutdown errors
 - Exit 1 in error during shutdown as a result of signal

### DIFF
--- a/main.go
+++ b/main.go
@@ -108,7 +108,8 @@ func main() {
 		fs.Usage,
 		server.PrintTLSHelpAndDie)
 	if err != nil {
-		server.PrintAndDie(fmt.Sprintf("%s: %s", exe, err))
+		fmt.Fprintln(os.Stderr, fmt.Sprintf("%s: %s", exe, err))
+		os.Exit(1)
 	} else if opts.CheckConfig {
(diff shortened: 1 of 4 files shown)


mprimi:
Closing after receiving feedback.

Current behavior is fine as-is. Exit with zero as a result of signal is an appropriate behavior for a long-running service. 
Carrying a bit more info into the exit code does not justify the risk of breakage and churn the change would create.
For example service-managers like systemd or K8s controllers may restart due to "restart on error" policies.

PR #6854, opened 2025-04-30
systemd: switch to Restart=on-failure

philpennock:
I switched the hardened example config file to use `Restart=always` in c78874276 just over a year ago.  That was a mistake.

If the nats-server is asked to shut down, and shuts down cleanly, then the service manager should not restart it.  The server should be restarted if it exits on failure.

https://www.freedesktop.org/software/systemd/man/latest/systemd.service.html#Restart=

### DIFF
--- a/util/nats-server-hardened.service
+++ b/util/nats-server-hardened.service
@@ -19,7 +19,7 @@ ExecStop=/bin/kill -s SIGUSR2  $MAINPID
 User=nats
 Group=nats
 
-Restart=always
+Restart=on-failure
 RestartSec=5
 
 # This should be `lame_duck_duration` + some buffer to finish the shutdown.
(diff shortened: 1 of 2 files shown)


wallyqs (approved):

philpennock:
> @philpennock should we add it to the other unit file too? Seems like something that would be expected by most users. E.g. it should restart in case of panic.

I added it to both unit files in this repo.  Is there another unit file you'd like me to update too?

alexbozhenko:
> > @philpennock should we add it to the other unit file too? Seems like something that would be expected by most users. E.g. it should restart in case of panic.
> 
> I added it to both unit files in this repo. Is there another unit file you'd like me to update too?

Oh, for some reason I thought you updated only the hardened file. Thanks!

PR #6858, opened 2025-05-01
Update const.go

ta2ozg:
disable default inject value:

- gitCommit
- serverVersion

Signed-off-by: Your Name <ozgur@goosey.org>


wallyqs:
Likely this was addressing this issue that happened with Go v1.25 which is now fixed: #7178

PR #6860, opened 2025-05-01
Update ipqueue.go

ta2ozg:
refactoring queue and add defer q.Unlock

Signed-off-by: Your Name <hello@goosey.org>

### DIFF
--- a/server/ipqueue.go
+++ b/server/ipqueue.go
@@ -130,6 +130,8 @@ func (q *ipQueue[T]) push(e T) (int, error) {
 		q.elts = *(q.pool.Get().(*[]T))
 	}
 	q.elts = append(q.elts, e)
+	1++
+	
 	q.Unlock()
 	if l == 0 {
 		select {
@@ -137,7 +139,7 @@ func (q *ipQueue[T]) push(e T) (int, error) {
 		default:
 		}
 	}
-	return l + 1, nil
+	return l, nil
 }
 
 // Returns the whole list of elements currently present in the queue,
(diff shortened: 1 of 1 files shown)


server/ipqueue.go
```diff
@@ -130,14 +130,16 @@ func (q *ipQueue[T]) push(e T) (int, error) {
 		q.elts = *(q.pool.Get().(*[]T))
 	}
 	q.elts = append(q.elts, e)
+	1++
```
neilalexander, line 133: This is a no-op as it's a numeric `1`, not a letter `l`.

server/ipqueue.go
```diff
@@ -177,14 +179,17 @@ func (q *ipQueue[T]) pop() []T {
...
+	
 	l := len(q.elts) - q.pos
 	if l == 0 {
-		q.Unlock()
 		var empty T
 		return empty, false
 	}
 	e := q.elts[q.pos]
-	if l--; l > 0 {
+	1--
```
neilalexander, line 190: Same here.

server/ipqueue.go
```diff
@@ -195,6 +200,9 @@ func (q *ipQueue[T]) popOne() (T, bool) {
 		default:
 		}
 	} else {
+		if q.calc != nil {
+			q.sz -= q.calc(e)
```
neilalexander, line 204: This is not needed as `q.sz` is about to be set to zero anyway.

neilalexander (changes requested):
Please review https://github.com/nats-io/nats-server/blob/main/CONTRIBUTING.md regarding writing a PR title & description for what you are trying to achieve or what problem you are trying to solve, as well as the PR sign-off. Meanwhile some notes below.

server/ipqueue.go
kozlovic: And that would be wrong anyway since we check after that for `if l == 0` to know if we need to signal.

server/ipqueue.go
kozlovic: So basically, these changes can be reduced to the use of `defer`. But although I think that `defer` have become very efficient, I think it should be left the way it was (to make it even more efficient ;-)).

kozlovic (changes requested):
Although we appreciate contributions, I believe that this PR should likely be closed. The only valid change would have been the use of `defer` that we could argue is not needed in such small function.

neilalexander:
Thanks for the PR but I don't think we will continue with this, the `defer` alone isn't enough of an optimisation and could actually be slower than the current code.

PR #6869, opened 2025-05-06
Retire `TestLongClusterStreamOrphanMsgsAndReplicasDrifting`

neilalexander:
This test fails for known reasons and models many patterns that should be better served by client tests or Failground.


MauriceVanVeen (approved):
LGTM

PR #6871, opened 2025-05-06
Optimise `enforceMsgPerSubjectLimit`

neilalexander:
When all of the state was lost and we need to rebuild and enforce a max messages per subject limit, then the `enforceMsgPerSubjectLimit` function could linearly scan multiple times and become very expensive if there are lots of subjects. This refactors the function to use new subject tree intersection logic to more quickly work out if there is any overlap between the subjects needing attention and the subjects in a given block, meaning we only need to scan the store once.

### DIFF
--- a/server/filestore.go
+++ b/server/filestore.go
@@ -4492,11 +4492,18 @@ func (fs *fileStore) enforceMsgPerSubjectLimit(fireCallback bool) {
 	var numMsgs uint64
 
 	// collect all that are not correct.
-	needAttention := make(map[string]*psi)
+	needAttention := stree.NewSubjectTree[uint64]()
+	fblk, lblk := uint32(math.MaxUint32), uint32(0)
 	fs.psim.IterFast(func(subj []byte, psi *psi) bool {
 		numMsgs += psi.total
 		if psi.total > maxMsgsPer {
-			needAttention[string(subj)] = psi
+			needAttention.Insert(subj, psi.total)
+			if psi.fblk < fblk {
+				fblk = psi.fblk
+			}
+			if psi.lblk > lblk {
+				lblk = psi.lblk
+			}
 		}
 		return true
 	})
@@ -4517,13 +4524,25 @@ func (fs *fileStore) enforceMsgPerSubjectLimit(fireCallback bool) {
 		// Rebuild fs state too.
 		fs.rebuildStateLocked(nil)
 		// Need to redo blocks that need attention.
-		needAttention = make(map[string]*psi)
+		needAttention.Empty()
+		fblk, lblk = uint32(math.MaxUint32), uint32(0)
 		fs.psim.IterFast(func(subj []byte, psi *psi) bool {
 			if psi.total > maxMsgsPer {
-				needAttention[string(subj)] = psi
+				needAttention.Insert(subj, psi.total)
+				if psi.fblk < fblk {
+					fblk = psi.fblk
+				}
+				if psi.lblk > lblk {
+					lblk = psi.lblk
+				}
 			}
 			return true
 		})
+	}
+
+	// If nothing to do then stop.
+	if fblk == math.MaxUint32 {
+		return
 	}
 
 	// Collect all the msgBlks we alter.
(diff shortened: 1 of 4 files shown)


server/filestore.go
```diff
@@ -4517,43 +4524,71 @@ func (fs *fileStore) enforceMsgPerSubjectLimit(fireCallback bool) {
...
+		t := time.NewTicker(10 * time.Second)
+		printed := false
+		for {
+			select {
+			case <-t.C:
+				fs.warn("Enforce message per subject limit %d progress: %d/%d blocks, %d subjects remaining", maxMsgsPer, i-fblk, lblk-fblk+1, needAttention.Size())
+				printed = true
+			case <-done:
+				if printed {
+					fs.warn("Enforce message per subject limit %d complete")
```
wallyqs, line 4567: measure how long it took? `took` etc... so that can filter out in the logs like other similar entries

server/filestore.go
```diff
@@ -4517,43 +4524,72 @@ func (fs *fileStore) enforceMsgPerSubjectLimit(fireCallback bool) {
...
+	}()
+	for ; i <= lblk; i++ {
+		mb := fs.bim[i]
+		if mb == nil {
+			continue
+		}
+		mb.mu.Lock()
+		mb.ensurePerSubjectInfoLoaded()
+		mb.mu.Unlock()
+		stree.Intersect(mb.fss, needAttention, func(subj []byte, ss *SimpleState, psptr **psi) {
```
derekcollison, line 4582: Deref of mb.fss without lock.

server/filestore.go
```diff
@@ -4517,43 +4524,72 @@ func (fs *fileStore) enforceMsgPerSubjectLimit(fireCallback bool) {
...
-			mb := fs.bim[i]
-			if mb == nil {
-				continue
+	done := make(chan struct{})
+	defer close(done)
+	i := fblk
+	go func() {
+		t := time.NewTicker(10 * time.Second)
+		started := time.Now()
+		printed := false
```
derekcollison, line 4560: `var printed bool` but in general seems it has no use atm?

server/filestore.go
```diff
@@ -4565,7 +4601,12 @@ func (fs *fileStore) enforceMsgPerSubjectLimit(fireCallback bool) {
 					seq++
 				}
 			}
-		}
+			// If we no longer need to prune this subject then take it out of
```
derekcollison, line 4604: Wording - from first line to second line.

server/filestore.go
```diff
@@ -4517,43 +4524,72 @@ func (fs *fileStore) enforceMsgPerSubjectLimit(fireCallback bool) {
...
+			if ss == nil || psptr == nil || *psptr == nil {
+				return
 			}
-			mb.mu.Unlock()
-			if ss == nil {
-				continue
+			total := (*psptr).total
+			// Update the first/last sequences for the subject if needed.
+			if ss.firstNeedsUpdate || ss.lastNeedsUpdate {
+				mb.recalculateForSubj(bytesToString(subj), ss)
```
derekcollison, line 4589: This expects lock to be held.

server/filestore.go
neilalexander: It's used in that goroutine to determine whether we were running long enough to have printed a status report, we then only print the success message if we did (i.e. the process took longer than 10 seconds).

neilalexander:
Still working on this PR as, having fixed up everything else, `TestNoRaceFileStoreMsgLimitsAndOldRecoverState/MaxMsgsPer` is now unhappy and I think it might have uncovered a bug in recovering with prior state files.

server/filestore.go
```diff
@@ -4517,55 +4524,93 @@ func (fs *fileStore) enforceMsgPerSubjectLimit(fireCallback bool) {
...
+	defer close(done)
+	i := fblk
+	go func() {
+		t := time.NewTicker(10 * time.Second)
+		started := time.Now()
+		var printed bool
+		for {
+			select {
+			case <-t.C:
+				fs.warn("Enforce message per subject limit %d progress: %d/%d blocks, %d subjects remaining", maxMsgsPer, i-fblk, lblk-fblk+1, needAttention.Size())
```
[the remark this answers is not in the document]
neilalexander: Ah yes, this was from an earlier version of the PR. I've updated the log line to say total as suggested, since it isn't safe to remove from `needAttention` inline.

server/filestore.go
```diff
@@ -4517,55 +4524,93 @@ func (fs *fileStore) enforceMsgPerSubjectLimit(fireCallback bool) {
...
+	done := make(chan struct{})
+	defer close(done)
+	i := fblk
+	go func() {
+		t := time.NewTicker(10 * time.Second)
+		started := time.Now()
+		var printed bool
+		for {
+			select {
+			case <-t.C:
```
derekcollison, line 4563: Could we check t one done case and avoid printed bool?

server/filestore.go
```diff
@@ -4517,55 +4524,93 @@ func (fs *fileStore) enforceMsgPerSubjectLimit(fireCallback bool) {
...
-				if err == nil {
-					seq = m.seq + 1
-					if removed, _ := fs.removeMsgViaLimits(m.seq); removed {
-						total--
-						blks[mb] = struct{}{}
-					}
-				} else {
-					// On error just do single increment.
-					seq++
+			first := ss.First
```
derekcollison, line 4601: If first not needed past block could do assignment in for.

derekcollison (commented):
In general LGTM, a few comments.

server/filestore.go
derekcollison: I would remove it myself..

Or if we feel strongly we want to keep it in, try to figure out a cleaner way to not have the printed bool.

neilalexander: OK, have removed it for now. We can always add it back later if we still find pathological cases where it's slow.

derekcollison (approved):
LGTM

PR #6876, opened 2025-05-07
Update to Go 1.24.3, set minimum to 1.23.9

neilalexander:


derekcollison (approved):
LGTM - Do we need to remove travis file at this time?

neilalexander:
I think it is out-of-date with the shape of our test runs as-is and we can always bring it back from git history if needed.

Issue #6881, opened 2025-05-08
WQ Stream Sequence Reset to 0 on abrupt termination


souravagrawal:
### Observed behavior

I encountered an issue where a NATS stream’s first and last sequence numbers were unexpectedly reset to 0 following an abrupt termination of the NATS server. Interestingly, the consumer remained fully caught up with messages and retained its expected state even after the crash, but the stream itself appeared to have been reset.

Before restarting NATS, I inspected the data directory and observed that the only blk file with 0 bytes. Additionally, based on the file timestamps, it was evident that index.db had not been updated recently and appeared stale. 

Below is my stream info before crash
```
  Subjects: test
              Replicas: 1
               Storage: File
Options:
             Retention: WorkQueue
       Acknowledgments: true
        Discard Policy: New
      Duplicate Window: 2m0s
            Direct Get: true
     Allows Msg Delete: true
          Allows Purge: true
        Allows Rollups: false
Limits:
      Maximum Messages: unlimited
   Maximum Per Subject: unlimited
         Maximum Bytes: unlimited
           Maximum Age: unlimited
  Maximum Message Size: unlimited
     Maximum Consumers: unlimited
State:
              Messages: 0
                 Bytes: 0 B
        First Sequence: 100,001
         Last Sequence: 100,000 @ 2025-05-08 14:54:10
      Active Consumers: 1
```

Consumer Info Before Crash
```
Information for Consumer test > test created 2025-05-08T14:54:03+05:30

Configuration:

                    Name: test
               Pull Mode: true
          Deliver Policy: All
              Ack Policy: Explicit
                Ack Wait: 30.00s
           Replay Policy: Instant
         Max Ack Pending: 1,000
       Max Waiting Pulls: 512

State:

  Last Delivered Message: Consumer sequence: 100,000 Stream sequence: 100,000 Last delivery: 2m5s ago
    Acknowledgment Floor: Consumer sequence: 100,000 Stream sequence: 100,000 Last Ack: 2m5s ago
        Outstanding Acks: 0 out of maximum 1,000
    Redelivered Messages: 0
    Unprocessed Messages: 0
           Waiting Pulls: 0 of maximum 512
```

Stream State after crash

```
State:

              Messages: 0
                 Bytes: 0 B
        First Sequence: 0
         Last Sequence: 0
      Active Consumers: 0
```

Consumer State after crash

```
State:

  Last Delivered Message: Consumer sequence: 200,000 Stream sequence: 200,000
[...]

PR #6882, opened 2025-05-08
[FIXED] workqueue reset to 0 when blk file is zero-sized due to unflushed data on crash

souravagrawal:
I encountered an issue where a Workqueue stream’s first and last sequence numbers were unexpectedly reset to 0 following an abrupt termination of the NATS server. Interestingly, the consumer remained fully caught up with messages and retained its expected state even after the crash, but the stream itself appeared to have been reset.

I was able to retrieve a backup of the data after the crash and debug it locally. During analysis, I found that new msgs had not been flushed to disk, resulting in a zero-sized blk file I believe. As a result, during recovery, the stream state remained at zero and the index.db could not be used to reconstruct the state

Resolves : https://github.com/nats-io/nats-server/issues/6881

Signed-off-by: souravagrawal <souravagrawal1111@gmail.com>

### DIFF
--- a/server/filestore.go
+++ b/server/filestore.go
@@ -475,18 +475,20 @@ func newFileStoreWithCreated(fcfg FileStoreConfig, cfg StreamConfig, created tim
 		}
 
 		// Check if our prior state remembers a last sequence past where we can see.
-		if fs.ld != nil && prior.LastSeq > fs.state.LastSeq {
+		if prior.LastSeq > fs.state.LastSeq {
 			fs.state.LastSeq, fs.state.LastTime = prior.LastSeq, prior.LastTime
 			if fs.state.Msgs == 0 {
 				fs.state.FirstSeq = fs.state.LastSeq + 1
 				fs.state.FirstTime = time.Time{}
 			}
-			if _, err := fs.newMsgBlockForWrite(); err == nil {
-				if err = fs.writeTombstone(prior.LastSeq, prior.LastTime.UnixNano()); err != nil {
+			if fs.ld != nil {
+				if _, err := fs.newMsgBlockForWrite(); err == nil {
+					if err = fs.writeTombstone(prior.LastSeq, prior.LastTime.UnixNano()); err != nil {
+						return nil, err
+					}
+				} else {
 					return nil, err
 				}
-			} else {
-				return nil, err
 			}
 		}
 		// Since we recovered here, make sure to kick ourselves to write out our stream state.
--- a/server/filestore_test.go
+++ b/server/filestore_test.go
@@ -9555,3 +9555,33 @@ func TestFileStoreAllLastSeqs(t *testing.T) {
 	require_NoError(t, err)
 	require_True(t, reflect.DeepEqual(seqs, expected))
 }
+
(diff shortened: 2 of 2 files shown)


server/filestore.go
```diff
@@ -489,6 +489,11 @@ func newFileStoreWithCreated(fcfg FileStoreConfig, cfg StreamConfig, created tim
 				return nil, err
 			}
 		}
+		//Use prior state when our stream state could not be recovered from blk files.
+		if fs.state.FirstSeq|fs.state.LastSeq == 0 && prior.FirstSeq > 0 && prior.LastSeq > 0 {
```
neilalexander, line 493: Is this effectively just covered by taking the `if fs.ld != nil` off the above existing condition?

souravagrawal: yes we can remoe the fs.ld != nil, but I believe based on fs.ld != nil a new msgblock has to be created to write tombstones, we can move the fs.ld != nil check before creating msgblock
```
if fs.ld != nil {
   if _, err := fs.newMsgBlockForWrite(); err == nil {
	if err = fs.writeTombstone(prior.LastSeq, prior.LastTime.UnixNano()); err != nil {
		return nil, err
	}
	} else {
	        return nil, err
	}
   }
```

souravagrawal: Hello @neilalexander I have made the changes as you suggested, but I have moved the fs.ld != nil before creating a  new msgblock for tombstones, without it the existing msg block with zero sized will just stay there forever and does not get removed.

neilalexander (approved):
LGTM but would like @derekcollison to take a look too.

derekcollison:
@souravagrawal can you confirm that the machine or VM or container crashed, not just the nats server here? We always flush through to the kernel once a message has been processed and before any publish ack. So the OS would have had to crash in our opinion from yor description above.

souravagrawal:
> @souravagrawal can you confirm that the machine or VM or container crashed, not just the nats server here? We always flush through to the kernel once a message has been processed and before any publish ack. So the OS would have had to crash in our opinion from your description above.

Hello @derekcollison, yes we have observed this issue with VM crash and not only NATS.

derekcollison:
@souravagrawal I am saying that what you describe above can not happen without a kernel crash. For clarity, are you claiming you have seen this with just abnormal termination of the server and the kernel / OS remained running?

souravagrawal:
> @souravagrawal I am saying that what you describe above can not happen without a kernel crash. For clarity, are you claiming you have seen this with just abnormal termination of the server and the kernel / OS remained running?

We've observed this issue multiple times in the event of a VM crash though it's unclear whether the kernel was also affected in those instances, we couldn't initially determine the root cause or reproduce the problem consistently. 

we simulated a kernel panic by issuing the following command in one of the env:
`echo c > /proc/sysrq-trigger`
This reliably reproduced the issue in almost every run. 
To troubleshoot this issue, we examined the NATS data directory before rebooting NATS and found that only a single .blk file was present with a file size of 0 bytes. We restored this data locally to troubleshoot and found that the associated stream state is getting reset to 0.

derekcollison (approved):
LGTM

PR #6883, opened 2025-05-08
TLS: Include subject name and SHA-256 fingerprint in handshake error log lines where possible

neilalexander:
In cases where we have known validation errors on a TLS certificate, include the subject name and the SHA-256 fingerprint of the certificate in the log line so that they can be more easily identified.

Example:
`[ERR] 127.0.0.1:50610 - gid:1 - TLS gateway handshake error: tls: failed to verify certificate: x509: cannot validate certificate for 127.0.0.1 because it doesn't contain any IP SANs (CN=localhost,OU=nats.io,O=Synadia,ST=California,C=US SHA-256: cb11b587f11697bec4b369f7761283f8b536337c538b5a43c07105a906f9eb12)`

### DIFF
--- a/server/client.go
+++ b/server/client.go
@@ -15,8 +15,10 @@ package server
 
 import (
 	"bytes"
+	"crypto/sha256"
 	"crypto/tls"
 	"crypto/x509"
+	"encoding/hex"
 	"encoding/json"
 	"errors"
 	"fmt"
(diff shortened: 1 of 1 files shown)


wallyqs (approved):
LGTM

johnweldon (approved):
LGTM

For another PR, but the request is also to include this detail on the connect and disconnect logs.

PR #6886, opened 2025-05-09
[FIXED] Gateway: TLS configuration reload now applies to implicit remotes

### LINKED ISSUE #6875: Fix Gateways tlsconfig reload
### Proposed change

When reload config, we should update all config in "g.remotes", not only config in "opts.Gateway.Gateways".

Image

https://github.com/nats-io/nats-server/blob/323e97ef773599c92b2ad35fae3b3e9ba5694712/server/gateway.go#L424

### Use case

For example，we config like this, but we have implicit cluster "al-tt-cn-kk", so opts.Gateway.Gateways only contains "aa-ss-kk-4", g.remotes contains "al-tt-cn-kk" and "al-tt-cn-kk".

```
gateway: {
    gateways: [
      {
        "name": "aa-ss-kk-4",
        "urls": [
          "nats://127.0.0.1:7222"
        ]
      }
    ],
    name: al-cn-tt-5,
    port: 7222,
    tls: {
      ca_file: /Users/zego/GolandProjects/go-zego/istioTest/nats/spireserver/bundle.0.pem,
      cert_file: /Users/zego/GolandProjects/go-zego/istioTest/nats/spireserver/svid.0.pem,
      key_file: /Users/zego/GolandProjects/go-zego/istioTest/nats/spireserver/svid.0.key,
      verify: true,
      insecure: true, //不校验
      verify_cert_and_check_known_urls: true
    }
}

```

### Contribution

updateRemotesTLSConfig update follow.

```
// Update remote gateways TLS configurations after a config reload.
func (g *srvGateway) updateRemotesTLSConfig(opts *Options) {
	g.Lock()
	defer g.Unlock()
	for name, cfg := range g.remotes {
		if name == g.name {
			continue
		}
		cfg.Lock()
		var ro *RemoteGatewayOpts
		for _, r := range opts.Gateway.Gateways {
			if r.Name == name {
				ro = r
				break
			}
		}
		// If TLS config is in remote, use that one, otherwise,
		// use the TLS config from the main block.
		if ro != nil && ro.TLSConfig != nil {
			cfg.TLSConfig = ro.TLSConfig.Clone()
		} else if opts.Gateway.TLSConfig != nil {
			cfg.TLSConfig = opts.Gateway.TLSConfig.Clone()
		}
		// Ensure that OCSP callbacks are always setup after a reload if needed.
		mustStaple := opts.OCSPConfig != nil && opts.OCSPConfig.Mode == OCSPModeAlways
		if mustStaple && opts.Gateway.TLSConfig != nil {
			clientCB := opts.Gateway.TLSConfig.GetClientCertificate
			verifyCB := opts.Gateway.TLSConfig.VerifyConnection
			if mustStaple && cfg.TLSConfig != nil {
				if clientCB != nil && cfg.TLSConfig.GetClientCertificate == nil {
					cfg.TLSConfig.GetClientCertificate = clientCB
				}
				if verifyCB != nil && cfg.TLSConfig.VerifyConnection == nil {
					cfg.TLSConfig.VerifyConnection = verifyCB
				}
			}
		}
		cfg.Unlock()
	}
}
```

kozlovic:
When a config reload is issued and a server has implicit remote gateways (that is, gateways that were gossiped and not explicitly configured), the TLS changes were applied only to explicit remotes.

With this change, the changes to the main `tls{}` block will be applied to implicit remotes.

Credit to @song0071000 for the proposed changes in the code.

Resolves #6875

### DIFF
--- a/server/gateway.go
+++ b/server/gateway.go
@@ -424,38 +424,48 @@ func (s *Server) newGateway(opts *Options) error {
 func (g *srvGateway) updateRemotesTLSConfig(opts *Options) {
 	g.Lock()
 	defer g.Unlock()
-
-	for _, ro := range opts.Gateway.Gateways {
-		if ro.Name == g.name {
+	// Instead of going over opts.Gateway.Gateways, which would include only
+	// explicit remotes, we are going to go through g.remotes.
+	for name, cfg := range g.remotes {
+		if name == g.name {
 			continue
 		}
-		if cfg, ok := g.remotes[ro.Name]; ok {
-			cfg.Lock()
-			// If TLS config is in remote, use that one, otherwise,
-			// use the TLS config from the main block.
-			if ro.TLSConfig != nil {
-				cfg.TLSConfig = ro.TLSConfig.Clone()
-			} else if opts.Gateway.TLSConfig != nil {
-				cfg.TLSConfig = opts.Gateway.TLSConfig.Clone()
-			}
-
-			// Ensure that OCSP callbacks are always setup after a reload if needed.
-			mustStaple := opts.OCSPConfig != nil && opts.OCSPConfig.Mode == OCSPModeAlways
-			if mustStaple && opts.Gateway.TLSConfig != nil {
-				clientCB := opts.Gateway.TLSConfig.GetClientCertificate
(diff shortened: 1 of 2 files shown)


derekcollison (approved):
LGTM

PR #6895, opened 2025-05-15
Fix spelling mistake

cjohansen:
We ran into the "message too large" error today, but the server reports it as "message to large". This fixes the spelling mistake.

Signed-off-by: Christian Johansen <christian@cjohansen.no>


derekcollison (approved):
LGTM

PR #6898, opened 2025-05-16
Limit the number of in-memory cached blocks per store

neilalexander:
The cache expiry timer may not be fast enough to expire blocks from the cache when a stream is under either heavy ingest or a high-speed readout, so try to cap the number of cached blocks to a set number. The `evict` IPQ will start to pop entries for eviction once we reach `defaultCacheEvictionThreshold`.

### DIFF
--- a/server/filestore.go
+++ b/server/filestore.go
@@ -205,6 +205,7 @@ type fileStore struct {
 	ttls        *thw.HashWheel
 	sdm         *SDMMeta
 	lpex        time.Time // Last PurgeEx call.
+	evict       *ipQueue[*msgBlock]
 }
 
 // Represents a message store block and its data.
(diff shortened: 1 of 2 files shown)


neilalexander:
Ran this through Gauge and it makes virtually no difference to the JetStream, KV, object store or filestore benchmarks.

server/filestore.go
```diff
@@ -313,6 +314,8 @@ const (
 	maxBufReuse = 2 * 1024 * 1024
 	// default cache buffer expiration
 	defaultCacheBufferExpiration = 10 * time.Second
+	// deault cache size before eviction
```
derekcollison, line 317: Spelling

server/filestore.go
```diff
@@ -7058,6 +7062,14 @@ checkCache:
 	if len(buf) > 0 {
 		mb.cloads++
 		mb.startCacheExpireTimer()
+
+		// If loading this block into the cache caused us to reach the eviction
+		// threshold for cached blocks, evict the oldest one.
+		if n, _ := mb.fs.evict.push(mb); n == defaultCacheEvictionThreshold {
```
derekcollison, line 7068: How do we remove from evict on normal cache timeout? Don't see that code.

derekcollison (commented):
Let's discuss, agree we need to do something here.

neilalexander:
This will be replaced with weak pointers.

PR #6899, opened 2025-05-16
Fix potential deadlock in `MultiLastSeqs`

neilalexander:
`MultiLastSeqs` already held the read lock before calling `AllLastSeqs` which would try to take the read lock again, but this could result in a deadlock if another goroutine is meanwhile trying to take the lock for writing.


MauriceVanVeen (approved):
LGTM

PR #6907, opened 2025-05-19
Add CustomDialer option for LeafNodes

chezgi:
custom dialer is needed when we want to do custom operations like fwmark of connection packets.

Signed-off-by: abbas ali chezgi chezgi@amnesh.ir

### DIFF
--- a/test/leafnode_test.go
+++ b/test/leafnode_test.go
@@ -76,8 +76,8 @@ func runSolicitLeafServerToURL(surl string) (*server.Server, *server.Options) {
 	return RunServer(&o), &o
 }
 
-func runSolicitLeafServerCustomDialer(lso *server.Options,dialer server.CustomDialer) (*server.Server, *server.Options) {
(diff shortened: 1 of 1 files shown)


neilalexander (changes requested):
At the very least this needs tests, but also wondering more about this use-case. This must be an embedded server?

lschmierer:
I like the effort, we have a similar use case.
A few notes though:
- It does not handle the timeout. It might be worthwhile to rewrite the `natsDialTimeout` function to call [`DialContext`](https://pkg.go.dev/net#Dialer.DialContext) with a [`context.WithTimeout`](https://pkg.go.dev/context#WithTimeout) and let a custom `DialContext` function be passed to that.
- If you want to capture all traffic (including DNS) I think you would to also provide your own [resolver `LookupHost` function](https://github.com/nats-io/nats-server/blob/main/server/opts.go#L653). Using DNS host names within a user-space VPN would require this too

PR #6908, opened 2025-05-20
Optimise `firstMatching` by reducing allocations

neilalexander:
This improves `firstMatching` by simplifying the wildcard vs non-wildcard logic and removing a substantial number of allocations by no longer collecting `subs`. This results in quite a notable performance improvement and reduced GC pressure when checking interest state etc.


derekcollison (approved):
LGTM

PR #6912, opened 2025-05-21
Fix possible panic on filestore.go

wallyqs:
Fix to prevent this panic seen in tests:

```
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0xb69322]

goroutine 183 [running]:
github.com/nats-io/nats-server/v2/server.(*fileStore).cleanupOldMeta(0x0)
	server/filestore.go:1984 +0x42
created by github.com/nats-io/nats-server/v2/server.newFileStoreWithCreated.func4 in goroutine 143
	server/filestore.go:519 +0xb1
```

Signed-off-by: Your Name <wally@nats.io>


server/filestore.go
```diff
@@ -503,7 +503,11 @@ func newFileStoreWithCreated(fcfg FileStoreConfig, cfg StreamConfig, created tim
 	// Also make sure we get rid of old idx and fss files on return.
 	// Do this in separate go routine vs inline and at end of processing.
 	defer func() {
-		go fs.cleanupOldMeta()
```
derekcollison, line 506: Probably do check before spinning go routine.

derekcollison: `if fs != nil { go fs.cleanupOldMeta() }`

wallyqs: sounds good, updated to use that instead.

derekcollison (approved):
LGTM

neilalexander (approved):
LGTM

PR #6936, opened 2025-06-02
[ADDED] Protocol: INFO in response to CONNECT contains account info.

kozlovic:
In response to a leafnode or client CONNECT protocol, the server responds
with an INFO that contains the name of the account it is bound to and if
this account is the system account.
    
For client connections, since most of our supported clients send the
CONNECT and a PING and expect a PONG *first*, the server will send
the "Connect Info" protocol only after receiving the first PING and
sending back the first PONG.

### DIFF
--- a/server/client.go
+++ b/server/client.go
@@ -2507,36 +2507,29 @@ func (c *client) processPing() {
 	// If we are here, the CONNECT has been received so we know
 	// if this client supports async INFO or not.
 	var (
-		checkInfoChange bool
+		sendConnectInfo bool
 		srv             = c.srv
 	)
-	// For older clients, just flip the firstPongSent flag if not already
-	// set and we are done.
-	if c.opts.Protocol < ClientProtoInfo || srv == nil {
-		c.flags.setIfNotSet(firstPongSent)
-	} else {
-		// This is a client that supports async INFO protocols.
-		// If this is the first PING (so firstPongSent is not set yet),
-		// we will need to check if there was a change in cluster topology
-		// or we have a different max payload. We will send this first before
-		// pong since most clients do flush after connect call.
-		checkInfoChange = !c.flags.isSet(firstPongSent)
+	// For the first PING (so firstPongSet is false) and for clients
+	// that support async INFO protocols, we will send one with ConnectInfo=true,
+	// the name of the account the client is bound to, and if the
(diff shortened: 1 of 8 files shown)


aricart (approved):
LGTM

kozlovic:
I can't get the PR GA to complete. Keep getting error:
```
=== RUN   TestPSEmulation
    pse_test.go:64: RSSs did not match close enough: 7393280 vs 5140480
--- FAIL: TestPSEmulation (0.01s)
```
which is not even related to this test (since the added test is part of the `test` package, and the failure above is from the `server` package).

kozlovic:
> Should we do this for all accounts for first INFO past successful CONNECT?

@derekcollison We do that for all accounts. We return the name and a boolean indicating if this is the system account or not. Am I misunderstanding your question?

derekcollison:
Sorry I mean for client connections as well.

ripienaar:
+1 for clients also

kozlovic:
@derekcollison @ripienaar Ok, I will see what I can do, but will have to be later today or tomorrow.

kozlovic:
@derekcollison @ripienaar For clients, we have the added difficulty that most of our supported clients send CONNECT+PING and expect PONG in response (or -ERR), anything else will cause a failure. I don't think we can change all clients now, so I am sending the INFO only to clients that support async INFO and after receiving the first PING (and sending back the first PONG).

kozlovic:
Having some issues with LameDuck mode and Websocket clients. Switch to Draft until I can fix it.

kozlovic:
Ok, I think it is ok now. Still the PS emulation test fails, but I am sure this is unrelated.

neilalexander:
Assuming this is intended for 2.12?

ripienaar:
> Assuming this is intended for 2.12?

Soonest would be great but would be up to @neilalexander how he feels about the impact

neilalexander:
BTW if you can rebase on top of latest `main`, the PS emulation test should hopefully be fixed.

kozlovic:
@neilalexander I did rebase, but I am still getting flappers with the "remaining server tests" and after trying several times, I have now the failure with the PS emulation again.

neilalexander (approved):
LGTM

PR #6940, opened 2025-06-03
De-flake `TestPSEmulation`

neilalexander:


neilalexander:
The godocs think that the GC run is handled too here:

> FreeOSMemory forces a garbage collection followed by an attempt to return as much memory to the operating system as possible. (Even if this is not called, the runtime gradually returns memory to the operating system in a background task.)

derekcollison (approved):
LGTM

PR #6941, opened 2025-06-04
Correct copyrights in `.go` files

neilalexander:


MauriceVanVeen (approved):
LGTM

derekcollison (commented):
They should only be updated if they have been changed.

neilalexander:
They were, the script looks at the Git commit dates (ignoring the previous commits where only the copyrights had been updated). A lot of files have been touched this year because of `go fix`, test/tag changes etc, as well as genuine code changes, but did not have the copyrights updated at the same time.

PR #6949, opened 2025-06-06
Include leaf node connections in `connz`

### LINKED ISSUE #6930: Leafnode CID/LID not shown in connection list
### Observed behavior

`nats server request connections` and `/connz`  as of recently (2.11+?) do not show the leafnodes connections anymore.

### Expected behavior

Show leafnode connections with "kind=leafnode" (or similar)

### Server and client version

2.11.3 and later - All operating systems (Windows, Linux, Mac verified)

### Host environment

_No response_

### Steps to reproduce

_No response_

neilalexander:
This worked if you used `nats server report connections` inside an account because we would automatically fill in `Account` in that case, and this worked if you specified `Account` because the account map included both clients and leafs, but the global `s.clients` map didn't so `/connz` or a system account request would omit `s.leafs`.

Fixes #6930.

### DIFF
--- a/server/monitor.go
+++ b/server/monitor.go
@@ -23,6 +23,7 @@ import (
 	"encoding/json"
 	"expvar"
 	"fmt"
+	"maps"
 	"math"
 	"net"
 	"net/http"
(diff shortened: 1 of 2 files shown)


derekcollison (approved):
LGTM

PR #6967, opened 2025-06-11
[ADDED] Add account-level connection stats

bruth:
This adds tracking of msgs and bytes for route, gateway, and leaf connections.

More tests need to be added, but I wanted to get a proposal of the idea prepared for initial review in case there are any glaring issues.

### DIFF
--- a/server/client.go
+++ b/server/client.go
@@ -1446,14 +1446,25 @@ func (c *client) readLoop(pre []byte) {
 		// Updates stats for client and server that were collected
 		// from parsing through the buffer.
 		if c.in.msgs > 0 {
-			atomic.AddInt64(&c.inMsgs, int64(c.in.msgs))
-			atomic.AddInt64(&c.inBytes, int64(c.in.bytes))
+			inMsgs := int64(c.in.msgs)
+			inBytes := int64(c.in.bytes)
+
+			atomic.AddInt64(&c.inMsgs, inMsgs)
+			atomic.AddInt64(&c.inBytes, inBytes)
+
 			if acc != nil {
-				atomic.AddInt64(&acc.inMsgs, int64(c.in.msgs))
-				atomic.AddInt64(&acc.inBytes, int64(c.in.bytes))
+				acc.stats.Lock()
+				acc.stats.inMsgs += inMsgs
+				acc.stats.inBytes += inBytes
+				if c.kind == LEAF {
+					acc.stats.ln.inMsgs += int64(inMsgs)
+					acc.stats.ln.inBytes += int64(inBytes)
+				}
+				acc.stats.Unlock()
 			}
-			atomic.AddInt64(&s.inMsgs, int64(c.in.msgs))
-			atomic.AddInt64(&s.inBytes, int64(c.in.bytes))
+
+			atomic.AddInt64(&s.inMsgs, inMsgs)
+			atomic.AddInt64(&s.inBytes, inBytes)
 		}
 
 		// Signal to writeLoop to flush to socket.
@@ -1806,7 +1817,9 @@ func (c *client) handleWriteTimeout(written, attempted int64, numChunks int) boo
 		c.srv.scStats.leafs.Add(1)
 	}
 	if c.acc != nil {
-		atomic.AddInt64(&c.acc.slowConsumers, 1)
+		c.acc.stats.Lock()
+		c.acc.stats.slowConsumers++
+		c.acc.stats.Unlock()
 	}
 	c.Noticef("Slow Consumer %s: WriteDeadline of %v exceeded with %d chunks of %d total bytes.",
 		scState, c.out.wdl, numChunks, attempted)
@@ -2356,7 +2369,9 @@ func (c *client) queueOutbound(data []byte) {
 		atomic.AddInt64(&c.srv.slowConsumers, 1)
 		c.srv.scStats.clients.Add(1)
 		if c.acc != nil {
-			atomic.AddInt64(&c.acc.slowConsumers, 1)
+			c.acc.stats.Lock()
+			c.acc.stats.slowConsumers++
+			c.acc.stats.Unlock()
 		}
 		c.Noticef("Slow Consumer Detected: MaxPending of %d Exceeded", c.out.mp)
 		c.markConnAsClosed(SlowConsumerPendingBytes)
@@ -4728,6 +4743,8 @@ func (c *client) processMsgResults(acc *Account, r *SublistResult, msg, deliver,
 	// by having an extra size
 	var dlvMsgs int64
 	var dlvExtraSize int64
+	var dlvRouteMsgs int64
+	var dlvLeafMsgs int64
 
 	// We need to know if this is a MQTT producer because they send messages
 	// without CR_LF (we otherwise remove the size of CR_LF from message size).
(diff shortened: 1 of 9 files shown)


bruth:
I will fix the test.. it is an existing one that is doing a *contains string* match for the existing accstatz payload.

server/client.go
```diff
@@ -460,9 +460,11 @@ type readCache struct {
 	rts []routeTarget
 
 	// These are all temporary totals for an invocation of a read in readloop.
-	msgs  int32
-	bytes int32
-	subs  int32
+	msgs    int32
+	bytes   int32
+	subs    int32
+	lfmsgs  int32
```
derekcollison: What does `lf` mean here?

server/client.go
```diff
@@ -4728,6 +4734,8 @@ func (c *client) processMsgResults(acc *Account, r *SublistResult, msg, deliver,
 	// by having an extra size
 	var dlvMsgs int64
```
derekcollison, line 4744: These then dlvClientOrLeafMsgs?

server/client.go
```diff
@@ -4737,15 +4745,32 @@ func (c *client) processMsgResults(acc *Account, r *SublistResult, msg, deliver,
...
+
 		// For non MQTT producers, remove the CR_LF * number of messages
 		if !prodIsMQTT {
 			totalBytes -= dlvMsgs * int64(LEN_CR_LF)
+			routeBytes -= dlvRouteMsgs * int64(LEN_CR_LF)
+			leafBytes -= dlvLeafMsgs * int64(LEN_CR_LF)
 		}
+
 		if acc != nil {
 			atomic.AddInt64(&acc.outMsgs, dlvMsgs)
```
derekcollison: I am wondering now with so many things to update if a normal lock would work better?

server/events.go
```diff
@@ -400,12 +400,20 @@ type GatewayStat struct {
...
 
-// DataStats reports how may msg and bytes. Applicable for both sent and received.
-type DataStats struct {
+type dataStats struct {
 	Msgs  int64 `json:"msgs"`
 	Bytes int64 `json:"bytes"`
 }
 
+// DataStats reports how may msg and bytes. Applicable for both sent and received.
+type DataStats struct {
```
derekcollison, line 409: Can replace first two lines with just `datastats`

server/gateway.go
```diff
@@ -3102,6 +3105,11 @@ func (c *client) processInboundGatewayMsg(msg []byte) {
 		return
 	}
 
+	atomic.AddInt64(&acc.inMsgs, 1)
```
derekcollison: Again might need to look into this a bit more with now 4 atomic operations vs simple mutex (not RW). Might want @neilalexander to weigh in here.

server/monitor_test.go
```diff
@@ -4597,6 +4598,124 @@ func TestMonitorAccountStatzOperatorMode(t *testing.T) {
...
+					require_Equal(t, acc.Received.Msgs, 3)
+				case "n1":
+					// Should have sent 1 message to a client.
+					// Should have received 1 message from n0.
+					require_Equal(t, acc.Sent.Msgs, 1)
+					require_Equal(t, acc.Sent.Routes.Msgs, 0)
+					require_Equal(t, acc.Received.Msgs, 1)
+					require_Equal(t, acc.Received.Routes.Msgs, 1)
+				case "n2":
+					// Should have not recieved anything.
```
derekcollison: fix spelling

server/route.go
```diff
@@ -472,6 +473,11 @@ func (c *client) processInboundRoutedMsg(msg []byte) {
 		return
 	}
 
+	atomic.AddInt64(&acc.inMsgs, 1)
```
derekcollison: Same here.

server/client.go
bruth: Leafnode messages/bytes. I can change references of `lf` to `ln` if that would be more intuitive (as well as add comments).

server/client.go
bruth: Leaf messages are tracked separately. As it exists in this PR, the `dlvMsgs` is still the total sum because that is what existing metrics/reporting relies on. I did not want to break that. So to calculate the pure client messages (non-route, gw, leaf), those counts would be subtracted from this total.

server/client.go
bruth: I put a basic benchmark together and beyond 2 atomics, a mutex is more performant (at least on the M2). 

```
goos: darwin
goarch: arm64
pkg: github.com/nats-io/nats-server/v2/tmp
cpu: Apple M2
BenchmarkAtomicAdd__2-8         129472086                7.870 ns/op
BenchmarkMutexAdd__2-8          158945107                7.969 ns/op
BenchmarkAtomicAdd__4-8         76698957                15.84 ns/op
BenchmarkMutexAdd__4-8          157250071                7.556 ns/op
BenchmarkAtomicAdd__8-8         38381012                31.29 ns/op
BenchmarkMutexAdd__8-8          156376851                7.596 ns/op
```

bruth: Will update to a stats mutex unless @neilalexander has other thoughts.

server/client.go
derekcollison: `ln` I would have recognized.

server/client.go
derekcollison: Make sure it has its own mutex.

derekcollison: And not RW

server/events.go
```diff
@@ -2410,12 +2423,40 @@ func (a *Account) statz() *AccountStat {
 		TotalConns: localConns + leafConns,
 		NumSubs:    a.sl.Count(),
 		Received: DataStats{
-			Msgs:  atomic.LoadInt64(&a.inMsgs),
-			Bytes: atomic.LoadInt64(&a.inBytes),
+			dataStats: dataStats{
```
bruth: Going to update this to use the new lock.

server/jetstream_cluster.go
```diff
@@ -8214,8 +8214,10 @@ func (mset *stream) processClusteredInboundMsg(subject, reply string, hdr, msg [
 		// If we are using the system account for NRG, add in the extra sent msgs and bytes to our account
 		// so that the end user / account owner has visibility.
 		if node.IsSystemAccount() && mset.acc != nil && r > 1 {
-			atomic.AddInt64(&mset.acc.outMsgs, int64(r-1))
-			atomic.AddInt64(&mset.acc.outBytes, int64(len(esm)*(r-1)))
+			mset.acc.smu.Lock()
+			mset.acc.outMsgs += int64(r - 1)
```
bruth: Since this is NRG traffic, should be add this to the route stats as well correct?

server/client.go
```diff
@@ -1446,14 +1446,25 @@ func (c *client) readLoop(pre []byte) {
...
+			inMsgs := int64(c.in.msgs)
+			inBytes := int64(c.in.bytes)
+
+			atomic.AddInt64(&c.inMsgs, inMsgs)
+			atomic.AddInt64(&c.inBytes, inBytes)
+
 			if acc != nil {
-				atomic.AddInt64(&acc.inMsgs, int64(c.in.msgs))
-				atomic.AddInt64(&acc.inBytes, int64(c.in.bytes))
+				acc.smu.Lock()
```
derekcollison: Even though for larger more complex structs that could have multiple locks we use `mu`, for small simple structs you can leave it unamed and just do a lock direct on the struct. So if stats is always non-nil in acc, then you could do `acc.stats.Lock()`

server/gateway.go
```diff
@@ -2732,8 +2732,12 @@ func (c *client) sendMsgToGateways(acc *Account, msg, subject, reply []byte, qgr
 			totalBytes -= dlvMsgs * int64(LEN_CR_LF)
 		}
 		if acc != nil {
-			atomic.AddInt64(&acc.outMsgs, dlvMsgs)
-			atomic.AddInt64(&acc.outBytes, totalBytes)
+			acc.smu.Lock()
```
derekcollison: Maybe acc needs `updateStats(msgs, bytes, rmsgs, rbytes, gwmsgs, gwbytes, lnmsgs, lnbytes)`?

server/client.go
bruth: There are four stats structs on the `acc` now. The existing one is embedded and the new ones, `ln`, `gw`, and `rt`. The choice for the `smu` was for a single lock when any/all of them need to be updated in one shot vs. individual locks. 

The alternative is to refactor this to encapsulate the stats together as one unit (new struct) with a lock.

derekcollison: I think we should do this.

neilalexander: Have refactored the struct shape a bit.

server/gateway.go
neilalexander: I had a go at doing this, but it needs to be split over two functions for inbound vs outbound, and I couldn't come up with a shape that felt good at the callsites. For now I think the current approach is clearer.

derekcollison (approved):
LGTM

PR #6981, opened 2025-06-15
[IMPROVED] Speed up a mirror or source consumer's resync across leafnode connections.

derekcollison:
When a consumer for a source or a mirror is failing to be created, we backoff creation attempts.
If the failures were due to a downed leafnode, meaning the mirror or source is across a leafnode connection, the resync could take more time then desired after the leafnode reconnects.

This improves resync time by hooking into the leafnode's reconnect logic (either via connect or async info).
Once we detect the reconnect, we search for streams that are leaders and are mirrors or sources, and  do not have an active sync consumer.

If we detect this we will reset the consumer backoff and retry with just a small jitter backoff.

### DIFF
--- a/server/jetstream_leafnode_test.go
+++ b/server/jetstream_leafnode_test.go
@@ -1426,3 +1426,249 @@ func TestJetStreamLeafNodeJSClusterMigrateRecoveryWithDelay(t *testing.T) {
 	// long election timer. Now this should work reliably.
 	lnc.waitOnStreamLeader(globalAccountName, "TEST")
 }
+
+// This will test that when a mirror or source construct is setup across a leafnode/domain
+// that it will recover quickly once the LN is re-established regardless
+// of backoff state of the internal consumer create.
+func TestJetStreamLeafNodeAndMirrorResyncAfterConnectionDown(t *testing.T) {
+	tmplA := `
+		listen: -1
+		server_name: tcm
+		jetstream {
+			store_dir: '%s',
+			domain: TCM
+		}
+		accounts {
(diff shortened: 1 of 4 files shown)


server/jetstream_leafnode_test.go
```diff
@@ -1426,3 +1426,249 @@ func TestJetStreamLeafNodeJSClusterMigrateRecoveryWithDelay(t *testing.T) {
...
+	// Now add an additional 500 messages to originals on both sides.
+	for i := 0; i < initMsgs; i++ {
+		// Individual Streams
+		jsA.PublishAsync("foo", []byte("PAYLOAD"))
+		jsB.PublishAsync("bar", []byte("PAYLOAD"))
+		// Bi-directional Sources
+		jsA.PublishAsync("A.foo", []byte("PAYLOAD"))
+		jsB.PublishAsync("B.bar", []byte("PAYLOAD"))
+	}
+	select {
```
neilalexander, line 1601: Minor nit but these can be shortened by using `require_ChanRead()`.

neilalexander (approved):
LGTM

server/jetstream_leafnode_test.go
derekcollison: Those channels are directed.

```
./jetstream_leafnode_test.go:1553:22: cannot use jsA.PublishAsyncComplete() (value of type <-chan struct{}) as chan struct{} value in argument to require_ChanRead
```

neilalexander: Huh, I wonder why we didn't make the helper directed as well as it could be. Oh well, ignore for now.

PR #6986, opened 2025-06-17
NRG: Fix integer overflow on 32-bit platforms

neilalexander:
This PR fixes an integer overflow that would happen only on 32-bit architectures. It also improves the `elen` counting in the Raft code to do the right thing on both 32-bit and 64-bit architectures.

Also adds 32-bit build checks into CI so this kind of thing doesn't sneak in again.


MauriceVanVeen (approved):
LGTM

PR #6990, opened 2025-06-18
Optimise `needAck` with new `SubjectForSeq` store function

neilalexander:
Adds a new `SubjectForSeq` function to the filestore that copies less and performs fewer allocations compared to `LoadMsg`. This is then used inside `needAck` when we don't know what the subject is for the sequence, i.e. when called from `checkStateForInterestStream`.

There are more improvements that need to be made here but for now this will reduce the amount of garbage created by this process, and by extension, reduce GC pressure and CPU usage on large interest-based streams.

### DIFF
--- a/server/filestore.go
+++ b/server/filestore.go
@@ -7399,6 +7399,24 @@ func (mb *msgBlock) msgFromBufEx(buf []byte, sm *StoreMsg, hh hash.Hash64, doCop
 	return sm, nil
 }
 
+// SubjectForSeq will return what the subject is for this sequence if found.
+func (fs *fileStore) SubjectForSeq(seq uint64) (string, error) {
+	fs.mu.RLock()
+	if seq < fs.state.FirstSeq {
+		fs.mu.RUnlock()
+		return _EMPTY_, ErrStoreMsgNotFound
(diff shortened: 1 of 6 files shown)


server/filestore.go
```diff
@@ -7399,6 +7399,22 @@ func (mb *msgBlock) msgFromBufEx(buf []byte, sm *StoreMsg, hh hash.Hash64, doCop
 	return sm, nil
 }
 
+// SubjectForSeq will return what the subject is for this sequence if found.
+func (fs *fileStore) SubjectForSeq(seq uint64) (string, error) {
+	fs.mu.Lock()
```
derekcollison: I think this could be a read lock for fs. And we can release after we get the block.

neilalexander: Can it be a read lock if we have to page the block into the cache as a part of this operation?

derekcollison: The mb will need write lock possible but we do not need it AFAIK in this function from the fs lock.

neilalexander: Done!

derekcollison (approved):
LGTM

PR #6995, opened 2025-06-20
Propagate stream interest policy changes to consumers in all cases

neilalexander:
When switching from limits to interest, we were propagating this update to consumers, but we weren't doing so when switching from interest to limits. This would leave consumers running with `o.retention` set to `InterestPolicy`, continuing to run `checkStateForInterestStream` etc even after the stream itself had moved back to the `LimitsPolicy`.

### DIFF
--- a/server/jetstream_test.go
+++ b/server/jetstream_test.go
@@ -20164,3 +20164,49 @@ func TestJetStreamDirectGetStartTimeSingleMsg(t *testing.T) {
 		})
 	}
 }
+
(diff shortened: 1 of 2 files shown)


derekcollison (changes requested):
LGTM - but need staticcheck fix on test first.

neilalexander:
Yep, just fixed it.

derekcollison (approved):
LGTM

PR #7006, opened 2025-06-25
Report memory usage correctly on Linux with huge pages

neilalexander:
4KB pages were assumed rather than asking the OS.


MauriceVanVeen (approved):
LGTM

PR #7009, opened 2025-06-26
Fix subsz total subs count

piotrpio:
Previously `SUBSZ` reported the number of returned `SubDetails` in `Subsz.Total`, which means that when using pagination if the total number of subs exceeds `Limit` it will cap at that). That does not align with how `Total` is used when using pagination in other endpoints (e.g. in `CONNZ` `Total` contains the full number of connections).


wallyqs (approved):
LGTM

PR #7043, opened 2025-07-07
NRG: Don't compact on catchup until snapshot written

neilalexander:
The `n.wal.Compact()` call before `n.installSnapshot()` is potentially dangerous in a crash scenario as we could blow away the log without having written the snapshot to disk yet. Once the snapshot is written, `n.installSnapshot()` will compact the log itself to the same index (as `lastIndex == n.pindex`).


MauriceVanVeen (approved):
LGTM

PR #7051, opened 2025-07-10
Export `SubjectMatchesFilter`

neilalexander:

### DIFF
--- a/server/sublist.go
+++ b/server/sublist.go
@@ -1434,6 +1434,12 @@ func tokenizeSubjectIntoSlice(tts []string, subject string) []string {
 	return tts
 }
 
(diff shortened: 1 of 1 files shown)


ripienaar:
I imagine this doesn’t change much but are we sure we want to support this wrt semver rules?

server/sublist.go
```diff
@@ -1434,6 +1434,12 @@ func tokenizeSubjectIntoSlice(tts []string, subject string) []string {
 	return tts
 }
 
+// Returns true if the subject matches the provided filter or false
```
wallyqs:
```suggestion
// SubjectMatchesFilter returns true if the subject matches the provided filter or false
```

PR #7054, opened 2025-07-12
(2.12) [ADDED] LeafNodes: `Disabled` option for remotes

kozlovic:
This allow disabling a remote leafnode using configuration reload. If changed from `false` to `true`, a solicited leafnode will be disconnected and will not reconnect.

If the `Disabled` option is changed from `true` to `false`, the leafnode will be solicited again.

Test has been added for various situations, including disabling a leaf that is not currently connected, and checking that we stop trying to (re)connect.


derekcollison (approved):
LGTM

Issue #7037, opened 2025-07-03
Automatic compression selection issues


renato0307:
### Observed behavior

**TL;DR**
We've been facing CPU throttling, high memory consumption, and increasing latency in the hub clusters. 

After we turn compression off in the leafnode connections, all problems disappear.

From our initial analysis, we think a snowball effect occurred: once memory usage reached a certain threshold, the garbage collector was triggered, which in turn increased CPU usage. As the CPU became throttled, RTT started to grow for some connections. This led to an automatic upgrade to a higher compression level, which further increased CPU load and RTT, perpetuating the cycle.

We encountered this issue on our production servers, but were able to reproduce it in our development setup. 

All the data below refers to the development setup.

_Note_: In the development setup, tracing and debugging are enabled. We repeated the same test with `trace` and `debug` set to `false`, without noticing any changes in behavior.

**Observabiltity**

CPU:

![Image](https://github.com/user-attachments/assets/242901a5-1ab9-40fb-bf92-9522ce7e06c1)

Memory:

_Observation_: We stopped the bench and started it again. During the period when the bench was stopped, memory usage never dropped, while CPU usage showed a noticeable decrease.

![Image](https://github.com/user-attachments/assets/8e58be9d-c47b-45bf-8323-8a99e8784055)

Latency:

![Image](https://github.com/user-attachments/assets/22fd8cf3-34b5-487e-b0d8-ab6004614ab9)

Throttling:

![Image](https://github.com/user-attachments/assets/54382f0a-deaa-4545-b0ee-e4801320aae0)

**Profilling**

_Development_: [profiles.zip](https://github.com/user-attachments/files/21040671/profiles.zip)

_Production_: [profiling_prod.zip](https://github.com/user-attachments/files/21041257/profiling_prod.zip)

### Expected behavior

Typical behaviour, the server being able to stand the load.

### Server and client version

```
~ $ nats --version
v0.1.2-0.20250310115758-f4eda5b1b7a3
```

```
~ $ nats-server --version
nats-server: v2.11.4
```

### Host environment

EKS at AWS, nodes running Bottlerocket.

![Image](https://github.com/user-attachments/assets/5c184bed-b3d9-4185-b443-796c9b5dc53e)

```
~ $ cat /etc/os-release
NAME="Alpine Linux"
ID=alpine
VERSION_ID=3.22.0
PRETTY_NAME="Alpine Linux v3.22"
HOME_URL="https://alpinelinux.org/"
[...]

PR #7056, opened 2025-07-12
add sync.Pool for compression writers.

aryehlev:
Should greatly reduce CPU and memory when using compression.
should help resolve:
https://github.com/nats-io/nats-server/issues/7037

Signed-off-by: Aryeh Klein aryehlevklein@gmail.com

### DIFF
--- a/server/compress.go
+++ b/server/compress.go
@@ -1,0 +1,52 @@
+package server
+
+import (
+	"compress/gzip"
+	"io"
+	"sync"
+
+	"github.com/klauspost/compress/s2"
+)
+
+var Gzip gzipPool
+var Snappy snappyPool
+
+type gzipPool struct {
+	writers sync.Pool
+}
+
+type snappyPool struct {
+	writers sync.Pool
+}
+
+func (pool *gzipPool) GetWriter(dst io.Writer) (writer *gzip.Writer) {
+	if w := pool.writers.Get(); w != nil {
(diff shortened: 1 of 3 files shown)


aryehlev:
https://github.com/nats-io/nats-server/issues/7037

neilalexander:
Thanks for the PR, but I am not yet convinced that this is characteristic of what's going on in #7037 and this may also cause us to hold onto and reuse much larger buffers than would be needed for a future connection.

The per-client compressors are already reusing underlying buffers for the lifetime of a connection and I don't really think that it's _creating_ the compressors that is the bulk of the issue, but rather that the latency sensitivity is too sensitive to GC pauses.

Have you done comparative benchmarking for this?

server/client.go
```diff
@@ -1631,8 +1632,9 @@ func (c *client) flushOutbound() bool {
 			}
 		}
 		if err == nil {
-			err = cw.Close()
+			err = c.out.cf(cw)
```
kozlovic, line 1635: In addition to @neilalexander comments, this is wrong and would break things. The writer is reused already (hence the Reset() above). That call would invalidate it causing likely memory corruptions.

aryehklein-rise:
@neilalexander you are right.
i initially thought it was similar issue i had in the past, but then i saw the client compressers are set once, furthermore the library uses sync pool for the exact same purpose. closing this PR.

PR #7066, opened 2025-07-15
[CHANGED] add endpoint names to monitoring overivew page

alexbozhenko:
People refer to monitoring endpoint names by the url. Add it for convenience. 

image

Signed-off-by: Alex Bozhenko <alexbozhenko@gmail.com>

### DIFF
--- a/server/monitor.go
+++ b/server/monitor.go
@@ -1447,33 +1447,33 @@ func (s *Server) HandleRoot(w http.ResponseWriter, r *http.Request) {
 
 	</div>
 	<br/>
-	<a href=.%s>General</a>
-	<a href=.%s>JetStream</a>
-	<a href=.%s>Connections</a>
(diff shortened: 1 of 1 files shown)


neilalexander (approved):
LGTM

derekcollison:
Why do we feel we need this?

derekcollison:
I prefer it left how it is. Trivial to hover over and figure out the url.

alexbozhenko:
@derekcollison The motivation I had in mind is in PR description:
> People refer to monitoring endpoint names by the url. 

When on a call with a customer, we would say "let's check varz", and people would know what we meant. 
We never say, e.g. "let's check "health probe" monitoring endpoint.

derekcollison:
I do not think adding it there does anything except make the UI more cluttered - which I prefer not to do - and you can get the URL trivially by hovering over the link.

PR #7087, opened 2025-07-21
[CHANGED] Make monitoring overview page aesthetically more pleasant

alexbozhenko:
Follow-up to https://github.com/nats-io/nats-server/pull/7066

Still simplifies life of dummies who do not remember all the endpoints by heart, but also makes UI not cluttered by showing endpoint only on hover
image

Signed-off-by: Alex Bozhenko <alexbozhenko@gmail.com>


derekcollison:
We have an action on hover, if you can add it to that it might be ok UI wise IMO.

alexbozhenko:
@derekcollison Updated to show endpoint only on hover. Thanks!

PR #7122, opened 2025-07-30
Add support for illumos

jwntree:
This allow NATS Server to build under the illumos.

Signed-off-by: Jiwon Na jwntree@hotmail.com


neilalexander:
Indeed nice that this PR reports more thoroughly, any chance you can rebase on latest `main` & address the sign-off in the commit message too?

jperkin:
Yeh this approach is better, however it does not currently support the `illumos` build tag. Please can you add it so that this change supports illumos as well as Solaris. Thanks.

PR #7135, opened 2025-08-01
Support illumos and Solaris.

jperkin:
This fixes the build on illumos (and probably Solaris).  While this may be useful for nats-server itself, this is more for other software such as telegraf that depends on this and which is currently unbuildable on illumos due to this problem. Whether you have any interest in supporting these platforms natively, we'd really appreciate this or a similar patch be applied so that other software is not adversely affected.

Previous iterations of this patch supported `sysmem` but used cgo (https://github.com/jperkin/nats-server/commit/7d43c51ba2f00cd1575d1062a6c1546a93f9e6ac), so this newer version simply uses the unsupported routine the same as wasm/zos (which is now combined into a single file).

Many thanks.

Signed-off-by: Jonathan Perkin <jonathan@perkin.org.uk>


neilalexander:
Principally happy with this but the CI is red because the commit message doesn't include the sign-off too. Any chance you can rebase & fix the commit message?

jperkin:
Sorry, yes! I missed that the first time. Pushed.

jwntree:
https://github.com/nats-io/nats-server/pull/7122

Sorry to interrupt, but wouldn’t this be a better approach?
I think you can just take it as is, if you'd like.

Also, I can't understand why you should handle irrelevant zos codes...

jperkin:
Happy for this to be closed in lieu of #7122 once that PR supports illumos. Thanks!

neilalexander:
Have merged #7122, thanks!

PR #7144, opened 2025-08-04
(2.12) Use cipher suites from `crypto/tls`, add `allow_insecure_cipher_suites` option

neilalexander:
This PR makes the following changes:

* We will now automatically be able to configure new ciphers added to `crypto/tls` by name without having to manually update our own configuration maps
* Added a new `allow_insecure_cipher_suites` option to the `tls` block to allow configuring cipher suites that are known to be broken/insecure if needed
* Error when configuring an insecure cipher suite when `allow_insecure_cipher_suites` is not set


derekcollison (approved):
LGTM

alexbozhenko:
Should we also error when tls1.3 ciphers are passed in tls block in the config, since tls1.3 ciphers are not configurable at all?

https://go.dev/blog/tls-cipher-suites
>    Go 1.17, recently released, takes over cipher suite preference ordering for all Go users. While Config.CipherSuites still controls which TLS 1.0–1.2 cipher suites are enabled, it is not used for ordering, and Config.PreferServerCipherSuites is now ignored. Instead, crypto/tls [makes all ordering decisions](https://go.dev/cl/314609), based on the available cipher suites, the local hardware, and the inferred remote hardware capabilities.
https://github.com/golang/go/blame/go1.24.5/src/crypto/tls/cipher_suites.go#L682
https://pkg.go.dev/crypto/tls@go1.24.5#Config.CipherSuites

org-ci-cd:
Seems like the MR is dropping support of TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305.
I'm not sure this was intended ?

neilalexander:
`TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305_SHA256` is the correct name and still supported: https://pkg.go.dev/crypto/tls#TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305_SHA256

PR #7153, opened 2025-08-06
(2.12) [ADDED] Trusted proxies support

kozlovic:
See [ADR](https://github.com/nats-io/nats-architecture-and-design/pull/361)


kozlovic:
@ripienaar @aricart As discussed, setting this as Draft so you can review and check if it is inlined with what is discussed in the ADR. Once that's done, the JWT PR will have to be merged (and released?) and then this PR will be updated to reflect the proper go.mod commit.

server/auth.go
```diff
@@ -65,6 +67,7 @@ type NkeyUser struct {
 	Account                *Account            `json:"account,omitempty"`
 	SigningKey             string              `json:"signing_key,omitempty"`
 	AllowedConnectionTypes map[string]struct{} `json:"connection_types,omitempty"`
+	RequireProxied         bool                `json:"require_proxied,omitempty"`
```
aricart: ProxyRequired?

server/auth.go
```diff
@@ -1103,6 +1136,61 @@ func (s *Server) processClientOrLeafAuthentication(c *client, opts *Options) (au
...
+// from a proxy whose signature can be verified by one of the known
+// trusted key, this function will return `true, true`. If the signature
+// cannot be verified by any, it will return `true, false`.
+// If the connectio is not proxied, or there are no configured trusted
+// proxies, then this function returns `false, false`.
+//
+// Server lock MUST NOT be held on entry since this function will grab
+// the read lock to extract the list of proxy trusted keys. The signature
+// verification process will be done outside of the lock.
+func (s *Server) proxyCheck(c *client, opts *Options) (bool, bool) {
```
aricart, line 1152: I may have this logic wrong:



Server Has Key | Connection Requires Verification | Sig is Provided | Outcome
-- | -- | -- | --
No | No | No | Connection proceeds normally
No | No | Yes | Connection proceeds normally
No | Yes | No | Connection rejected - server cannot verify
No | Yes | Yes | Connection rejected - server cannot verify
Yes | No | No | Connection proceeds normally
Yes | No | Yes | Connection proceeds normally
Yes | Yes | No | Connection rejected - signature required
Yes | Yes | Yes | Connection proceeds if signature validates

For normal connections (non-callout) it works. Now if the do a callout, the callout can then set whether the connection requires proxy verification.

Given that, this mean that all the rejections are errors, so should the proxy check simply be `bool, error`?

aricart: the token for callout could be a jwt that specifies that proxy is required, that will be passed to the callout as part of the options, so then the callout can or not propagate - ie - if the callout token was a jwt with that requirement, possibly that condition then is validated on the get-go and rejected if not.

server/auth.go
ripienaar: Yes this is better also for the other options

kozlovic: Changing in the JWT PR too, I assume.

ripienaar: Ugh, I guess so. Sorry @kozlovic

kozlovic:
@ripienaar @aricart @derekcollison I want to point out something about auth callout. The [doc](https://docs.nats.io/running-a-nats-service/configuration/securing_nats/auth_callout#multiple-accounts) seem to indicate that if there is a locally defined account, it should be empty (no users defined), but if that is not the case, and if we were to have one of these users with "proxy_required", but the auth callout generated JWT were to not specify this property, then the connection would be accepted (that is, even if it did not come from a proxy). See the test below:
```
func TestAuthCalloutProxyRequired(t *testing.T) {
	conf := `
		listen: "127.0.0.1:-1"
		server_name: A
		accounts {
  			AUTH: {
    			users: [ { user: auth, password: auth } ]
  			}
  			APP: {
				users: [ { user: proxy, password: pwd, proxy_required: true } ]
			}
  			SYS: {}
		}
		system_account: SYS

		authorization {
			timeout: 1s
			auth_callout {
				issuer: "ABJHLOVMPA4CI6R5KLNGOB4GSLNIY7IOUPAJC4YFNDLQVIOBYQGUWVLA"
				auth_users: [ auth ]
				account: AUTH
			}
		}
	`
	var invoked atomic.Int32
	handler := func(m *nats.Msg) {
		invoked.Add(1)
		user, si, ci, opts, _ := decodeAuthRequest(t, m.Data)
		require_True(t, si.Name == "A")
		require_True(t, ci.Host == "127.0.0.1")
		if opts.Username == "proxy" {
			var j jwt.UserPermissionLimits
			// If we don't set a ProxyRequired property, the auth callout will cause
			// the connection to be accepted, but should it?
			// j.ProxyRequired = true
			ujwt := createAuthUser(t, user, _EMPTY_, "APP", _EMPTY_, nil, 10*time.Minute, &j)
			m.Respond(serviceResponse(t, user, si.ID, ujwt, _EMPTY_, 0))
		} else {
			m.Respond(nil)
		}
	}
	at := NewAuthTest(t, conf, handler, nats.UserInfo("auth", "auth"))
	defer at.Cleanup()

	// This one should fail because it will require to be proxied and it is not.
	at.RequireConnectError(nats.UserInfo("proxy", "pwd"))

	// Auth callout should have been invoked once.
	require_Equal(t, 1, int(invoked.Load()))
}
```
The user "proxy" is defined in the "APP" account and requires a proxy connection. Since it is not the case, the "first layer" of authentication in the server will deny it because it did not go through a proxy, but right now, the server code invokes (rightfully or not) the auth callout code, which, as seen above, does not have "ProxyRequired" property, and so the connection is accepted. If - of course - I were to uncomment the line `j.ProxyRequired = true`, then the connection would be rejected.

If the account/users configuration above is valid *and* we think that in this case the connection should be rejected, I would need to make some changes to keep track of the "proxyRequired" property (of the user/nkey user/jwt in the first step of authentication) and pass that value to `processClientOrLeafCallout()` along with the current `trustedProxy` boolean. The check then would be something like:
```
		arc, err := decodeResponse(rc, rmsg, racc)
		if err != nil {
			c.authViolation()
			respCh <- titleCase(err.Error())
			return
		}
		if (proxyRequired || arc.ProxyRequired) && !trustedProxy {
			c.authViolation()
			respCh <- titleCase(errProxyRequired.Error())
			return
		}
```
Meaning that if a proxy is required either in the "first step" of authentication or in the auth callout generated JWT and there is not a trusted proxy then fail the connection.

Looking forward to your thoughts on that.

aricart:
The placement account where users go can be populated by call-out or jwts that place the user there.  Is this what you are trying to reconcile?  @kozlovic ping me tomorrow when you are available

ripienaar:
When the user is denied the log is:

```
[2843358] 2025/08/07 08:20:36.670179 [ERR] 127.0.0.1:58594 - cid:5 - authentication error - User "test"
```

Would it be possible to log something specific there to know it's due to no proxy?

Also the `io.nats.server.advisory.v1.client_disconnect` advisory has as Reason `Authentication Failure` be good to get something there also, the CONNZ output also shows the same `Authentication Failure` be good to get something indicating its due to this

Ofcourse we probably should not make any visible different back to the connecting client though, just what we do today wrt these errors

Not added to the ADR yet want to hear it makes sense to you first @kozlovic

kozlovic:
> Not added to the ADR yet want to hear it makes sense to you first @kozlovic

The way the code is, the function where we know the reason is because of not going through a proxy (or that the proxy is not trusted) simply returns a boolean. The caller(s) (higher in the call stack) will then invoke `c.authViolation()` that hardcodes it to the reason you see. I can imagine adding a field in the connection structure that if set will be the auth error to use in that function instead of the default one. Or I bite the bullet and have the whole auth stack return an actual error (to pass to call authViolation()) instead of a boolean.

ripienaar:
> The way the code is, the function where we know the reason is because of not going through a proxy (or that the proxy is not trusted) simply returns a boolean. The caller(s) (higher in the call stack) will then invoke `c.authViolation()` that hardcodes it to the reason you see. I can imagine adding a field in the connection structure that if set will be the auth error to use in that function instead of the default one. Or I bite the bullet and have the whole auth stack return an actual error (to pass to call authViolation()) instead of a boolean.

OK, do you think its feasible to solve or too much atm?

kozlovic:
Changing the calls return value is a no go (too many changes). I guess if you really want a different error for those, we would have to go with the "set auth error" approach in the connection structure unfortunately.

kozlovic:
> The placement account where users go can be populated by call-out or jwts that place the user there. Is this what you are trying to reconcile? @kozlovic ping me tomorrow when you are available

@aricart No, I was just showing (as we discussed offline) that if a user is locally defined with the "proxy_required" property but we end-up invoking auth callout, then if the auth's JWT does NOT set "ProxyRequired=true", then the connection would be accepted. In other words, the auth callout overrides the "local" decision. But, as I mentioned, the documentation seems to indicate that this would not be a "valid" configuration, since it sounds like there should not be any user defined for the accounts locally if one wants to rely on the auth callout.

ripienaar:
> Changing the calls return value is a no go (too many changes). I guess if you really want a different error for those, we would have to go with the "set auth error" approach in the connection structure unfortunately.

lets try that see how you feel about the outcome

aricart:
> > The placement account where users go can be populated by call-out or jwts that place the user there. Is this what you are trying to reconcile? @kozlovic ping me tomorrow when you are available
> 
> @aricart No, I was just showing (as we discussed offline) that if a user is locally defined with the "proxy_required" property but we end-up invoking auth callout, then if the auth's JWT does NOT set "ProxyRequired=true", then the connection would be accepted. In other words, the auth callout overrides the "local" decision. But, as I mentioned, the documentation seems to indicate that this would not be a "valid" configuration, since it sounds like there should not be any user defined for the accounts locally if one wants to rely on the auth callout.

I think this is a mistake in the doc that is there - authcallout is not and all or nothing setup. You can have normal clients that use regular credentials flow to be placed in the account, or users that have a `sentinel` token which then get processed by authcallout.

kozlovic:
> lets try that see how you feel about the outcome

@ripienaar Started with that approach seems ok. Will add/update tests to make sure we get the expected log/monitor/system messages.

kozlovic:
@ripienaar I made the changes. If you could review/test it out, that would be great.

ripienaar:
Really great, PR looks great to me, I'll update the ADR.

Log:

```
[2951823] 2025/08/08 06:51:24.486344 [ERR] [::1]:57314 - cid:7 - proxy connection required - User "test2"
```

Event:

```
{"type":"io.nats.server.advisory.v1.client_disconnect",....:"Proxy Required"}
```

Connz:

Screenshot 2025-08-08 at 09 52 49

server/reload.go
```diff
@@ -2589,3 +2627,34 @@ addLoop:
 
 	return add, remove
 }
+
+func diffProxiesTrustedKeys(old, new []*ProxyConfig) ([]string, []string) {
+	var add []string
+	var del []string
+	// Both "old" and "new" lists should be small...
+	for _, op := range old {
+		var found bool
```
neilalexander: Mostly a nit but this could be `slices.Contains()`, same with below.

server/opts.go
```diff
@@ -5358,6 +5394,94 @@ func parseMQTT(v any, o *Options, errors *[]error, warnings *[]error) error {
...
+			continue
+		}
+		proxy := &ProxyConfig{}
+		for k, v := range tm {
+			tk, v = unwrapValue(v, &lt)
+			switch strings.ToLower(k) {
+			case "key", "public_key":
+				proxy.Key = v.(string)
+				if !nkeys.IsValidPublicKey(proxy.Key) {
+					*errors = append(*errors, &configErr{tk, fmt.Sprintf("invalid proxy key %q", proxy.Key)})
```
neilalexander, line 5465: Should there be a `continue` here?

neilalexander (commented):
Overall looks good, couple minor things.

server/reload.go
kozlovic: But we don't want to compare the whole ProxyConfig, just the Key. Would slices.Contains() do only the compare on the op.Key?

server/opts.go
kozlovic: Good catch, thanks! (was harmless now since we have only 1 field)

server/reload.go
neilalexander: Ah, that'd be `slices.ContainsFunc()` instead then, which may actually end up not being that many less lines I suppose.

kozlovic: Maybe with `slices.ContainsFunc()`, let me try that.

neilalexander: If it ends up gnarly then don't worry about it, but if it does end up being more concise, all the better!

kozlovic: Made the changes!

server/opts.go
kozlovic: Fixed in the pushed commit.

PR #7171, opened 2025-08-14
(2.12) Allow promoting mirrors

neilalexander:
This allows promoting mirrors by removing the mirror configuration in a stream update.

### DIFF
--- a/server/jetstream_cluster.go
+++ b/server/jetstream_cluster.go
@@ -6633,7 +6633,8 @@ func (s *Server) jsClusteredStreamUpdateRequest(ci *ClientInfo, acc *Account, su
 		return
 	}
 	// Check for mirror changes which are not allowed.
-	if !reflect.DeepEqual(newCfg.Mirror, osa.Config.Mirror) {
+	// We will allow removing the mirror config to "promote" the mirror to a normal stream.
+	if newCfg.Mirror != nil && !reflect.DeepEqual(newCfg.Mirror, osa.Config.Mirror) {
 		resp.Error = NewJSStreamMirrorNotUpdatableError()
 		s.sendAPIErrResponse(ci, acc, subject, reply, string(rmsg), s.jsonResponse(&resp))
 		return
--- a/server/jetstream_test.go
+++ b/server/jetstream_test.go
@@ -14404,9 +14404,10 @@ func TestJetStreamMirrorUpdatesNotSupported(t *testing.T) {
 	_, err = js.AddStream(cfg)
 	require_NoError(t, err)
 
+	// Promoting a mirror is supported though.
 	cfg.Mirror = nil
 	_, err = js.UpdateStream(cfg)
-	require_Error(t, err, NewJSStreamMirrorNotUpdatableError())
+	require_NoError(t, err)
 }
 
 func TestJetStreamMirrorFirstSeqNotSupported(t *testing.T) {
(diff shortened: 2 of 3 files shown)


server/jetstream_cluster.go
```diff
@@ -6633,7 +6633,8 @@ func (s *Server) jsClusteredStreamUpdateRequest(ci *ClientInfo, acc *Account, su
 		return
 	}
 	// Check for mirror changes which are not allowed.
-	if !reflect.DeepEqual(newCfg.Mirror, osa.Config.Mirror) {
+	// We will allow removing the mirror config to "promote" the mirror to a normal stream.
```
ripienaar, line 6636: Should we require that subjects be added at the same time?

neilalexander: We could do, not clear that it's strictly necessary though. A stream config with no subjects is a valid config as the server will automatically insert the stream name into the subject list in that case. WDYT?

neilalexander: (Should add that the server only does that if there are no mirrors/sources configured.)

ripienaar: Not a big fan of that name==subject behaviour - but its what we got so at least make sure that happens on promote also when no subject is given?

neilalexander: Yep, have confirmed that it does. It's using all the usual stream update logic and checking.

server/jetstream_cluster.go
```diff
@@ -6633,7 +6633,8 @@ func (s *Server) jsClusteredStreamUpdateRequest(ci *ClientInfo, acc *Account, su
...
+	if newCfg.Mirror != nil && !reflect.DeepEqual(newCfg.Mirror, osa.Config.Mirror) {
```
derekcollison, line 6637: Should we make sure the primary os down here? I would want the mirror to operate like the primary at this point I think.

derekcollison: Should we inherit the subjects from the old primary here?

derekcollison (commented):
Feels like a "step" but worried about ops level semantics. I would like to promote and demote the other at close to the same time.

server/jetstream_cluster.go
ripienaar: We can't assume we will be able to reach the primary stream - this is something that will likely be done during a outage, the primary is dead or so down that its unreachable etc and people now are doing recovery playbooks.

So I dont think we can assume its reachable, ditto about being caught up - thats assuming the primary is reachable.

After this the primary Is basically unusable because it cant again become a mirror of this one.

Lots of moving parts in making this right and I think thats all tooling based moving parts, lots of confirming user intentions, asking if its available or not doing checks where the tooling can maybe even fetching the subjects but handling cases where its not possible etc

neilalexander: Yep, I think our consensus is that the server can't reliably judge what the right thing to do is, after all the mirror origin could be online in our domain, online but in a different domain, offline-ish/flaky, offline never to return. The best we can do is just surface the promotion as a primitive operation in the server and try our best to surface it in the tools in a way that makes sense to operators for different playbooks.

server/jetstream_cluster.go
jnmoyne: For me a stream with no subject and no source (and not being a mirror) is ok: you can still read from it and you can add subject(s) or source(s) later at any time. To me this is a valid stream config. I really do _not_ like automatically using the stream's name as a subject if that is what it is doing!

server/jetstream_cluster.go
jnmoyne: Yes, the main use case for this is IMHO is DR scenarios (specifically 2 sites DR), where we don't have any solution right now. 

IMHO this isn't meant for 'automated' multi-site fault tolerance: you must use a stretch cluster (or multiple clusters with cross sourcing) for that, but for disaster recovery where the administrators makes the decision to make what was the mirror at the DR site into the new primary, the stream being mirrored is going to be unreachable/down at that point. Let them control adding the subjects as part of that DR operation (the subjects could be a little bit different anyways depending on how the applications running at the DR site are configured). Just like it's going to be an administrative operation when they decide to do the reverse when the primary site is back up. They will indeed likely have playbooks for handling disasters that will involve many administrative operations, with NATS being just a small part of it.

server/jetstream_cluster.go
neilalexander: Yeah, I am not a fan particularly of the stream-name-as-subjects behaviour, although I think it's possibly been that way for quite some time.

We can potentially look at changing it if we can gate that behaviour on either the stream create or the API level 2, will have to have a look during the week.

jnmoyne: I did check and indeed it will automatically add the stream's name in the list of subjects if you don't specify a subject at the same time you remove the mirror. 

IMHO it really shouldn't do that (add the stream's name automatically as the subject being listened to), that it did that came to a surprise to me (and I would expect to most), I'm guessing this is a vestigial leftover from the very initial implementation of JS trying to make somewhat transparent on top of Core NATS? Either way: it's not documented anywhere, and in today's JS where streams can have multiple subjects, sources, and subject transforms there's absolutely no relationship (there can not be) between a stream's name and which subjects it may listen on. And anyways it's also inconsistent in that it does that automagic adding the filter when creating the stream only if you don't also specify a source or mirror.

My vote is that we should change and remove this undocumented strange behavior such that you can just remove the mirror from the stream and it doesn't automagically try to add a subject for you if you didn't specify one in the new stream config, or if you create a stream without specifying a subject.

jnmoyne: I should add that if you happen to have another stream already defined that listens on a subject that is the same as the name of the mirroring stream, then when you try to remove the mirror will errors out with 'subject overlap'

server/jetstream_cluster.go
neilalexander: @derekcollison If you're happy with this consensus then we can merge it for experimenting in preview.2 this week.

wallyqs: To promote the mirror you need to delete the original stream?

derekcollison: Might be good to add in same subjects from upstream source if we know them. Could be opt-in and fail if we can't reach it?

neilalexander: I think it'd be better to investigate possibilities like that in a separate PR, I suspect we'd have to issue a stream info to find those and a stream update doesn't feel like the right place to do that.

MauriceVanVeen (approved):
LGTM

PR #7184, opened 2025-08-19
Fix url encoding for ocsp requests

### LINKED ISSUE #7183: URL encoding for OCSP responder requests is broken
### Observed behavior

NATS does not properly encode the requests to an OCSP responder when doing revocation checks against a PKI. The base64 encoded certificate was missing. This can lead to problematic characters in the URL and failed revocation checks.

### Expected behavior

NATS correctly encodes the request in the X.690 format as specified here: https://datatracker.ietf.org/doc/html/rfc6960#appendix-A.1

This is part of the [ADR-38: OCSP Peer Verification](https://github.com/nats-io/nats-architecture-and-design/blob/main/adr/ADR-38.md) feature.

### Server and client version

NATS Version: 2.11.8

### Host environment

NATS is running on Kubernetes in a single cluster with multiple servers. But it's irrelevant for this issue.

### Steps to reproduce

1. Enable OCSP in the config and add the `ocsp_peer` config, like described in the [ADR-38 document](https://github.com/nats-io/nats-architecture-and-design/blob/main/adr/ADR-38.md)
2. Verify the requests on the PKI side. I know this is not easy to replicate, but I attached some screenshots of logs that show the incoming requests in detail.

Image

The two requests at the bottom are from NATS and you can see the `/` in the request URL

azerbe:
Like described in issue #7183 , this PR adds the encoding to the base64 encoded certificates.

Resolves #7183 

Signed-off-by: Alexander Zerbe <alex@aqube.de>

### DIFF
--- a/server/certidp/ocsp_responder.go
+++ b/server/certidp/ocsp_responder.go
@@ -19,6 +19,7 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"net/url"
 	"strings"
 	"time"
 
@@ -55,7 +56,7 @@ func FetchOCSPResponse(link *ChainLink, opts *OCSPPeerConfig, log *Log) ([]byte,
 		return nil, err
 	}
 
-	reqEnc := base64.StdEncoding.EncodeToString(reqDER)
+	reqEnc := encodeOCSPRequest(reqDER)
 
 	responders := *link.OCSPWebEndpoints
 
@@ -68,10 +69,10 @@ func FetchOCSPResponse(link *ChainLink, opts *OCSPPeerConfig, log *Log) ([]byte,
 		Timeout: timeout,
 	}
 	for _, u := range responders {
-		url := u.String()
-		log.Debugf(DbgMakingCARequest, url)
-		url = strings.TrimSuffix(url, "/")
-		raw, err = getRequestBytes(fmt.Sprintf("%s/%s", url, reqEnc), hc)
+		responderURL := u.String()
+		log.Debugf(DbgMakingCARequest, responderURL)
+		responderURL = strings.TrimSuffix(responderURL, "/")
+		raw, err = getRequestBytes(fmt.Sprintf("%s/%s", responderURL, reqEnc), hc)
 		if err == nil {
 			break
 		}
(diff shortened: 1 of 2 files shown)


server/certidp/ocsp_responder.go
```diff
@@ -82,3 +83,10 @@ func FetchOCSPResponse(link *ChainLink, opts *OCSPPeerConfig, log *Log) ([]byte,
 
 	return raw, nil
 }
+
+// encodeOCSPRequest encodes the OCSP request in base64 and URL-encodes it.
+// This is needed to fulfill the OCSP responder's requirements for the request format. (X.690)
+func encodeOCSPRequest(reqDER []byte) string {
+	reqEnc := base64.StdEncoding.EncodeToString(reqDER)
```
neilalexander, line 90: Should this be `base64.RawURLEncoding.EncodeToString(...)` and then you can get rid of `url.QueryEscape`?

azerbe: I was hesitant to remove the padding characters because I don't know how the decoder is implemented. It should be able to determine the correct size by measuring and rounding up, though.

However, using the URLEncoding functions definitely makes sense.

neilalexander: Good point about the padding, in which case `URLEncoding` should be fine indeed.

azerbe: I will test this against our PKI first and report back.

neilalexander: How did it go?

azerbe: Not well. The PKI we use can't handle some of the options, but I'm still waiting for a response.

I also checked back with the RFC 4648 and the URLEncode is not the same like normal base64. 

See https://datatracker.ietf.org/doc/html/rfc4648#section-5 for reference. 

Right now I think that my first approach of using the StdEncoding (without removing padding characters) and then url.Encoding is safer. Will update this PR as soon as I get feedback from my colleagues.

azerbe: OK, I've tested all the combinations against our PKI system.

Only the 'normal' Base64 encoding, including the padding characters, worked.

Therefore, I would stick with my initial proposal to fix the bug and achieve the greatest possible compatibility with external PKI systems.

PR #7189, opened 2025-08-20
(2.12) Add `"leader_since"` to stream and consumer info response

neilalexander:
This reports how long the leader believes it has been leader in stream & consumer info responses, as well as raftz.

### DIFF
--- a/server/jetstream_cluster.go
+++ b/server/jetstream_cluster.go
@@ -8824,9 +8824,10 @@ func (js *jetStream) clusterInfo(rg *raftGroup) *ClusterInfo {
 
 	n := rg.node
 	ci := &ClusterInfo{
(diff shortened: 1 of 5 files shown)


MauriceVanVeen (approved):
LGTM
Does this also register properly when scaling up from R1?

derekcollison (commented):
Should this be a duration instead? Fearful of TZ issues

neilalexander:
> Should this be a duration instead? Fearful of TZ issues

It could be a duration sure, I mostly just followed the pattern of existing fields like `created`.

> Does this also register properly when scaling up from R1?

It's reported from the Raft group directly, so yes.

neilalexander:
We have agreed to keep it as a timestamp (always UTC) so that it is still valid by the time it is received by clients.

PR #7231, opened 2025-08-28
Fix TestNRGLeaderTransfer

sciascid:
The test previously assumed the next AppendEntry after leader stepdown would always contain an EntryLeaderTransfer. In unlucky runs, the first AppendEntry observed would be a simple heartbeat, causing a test failure. To avoid this failure, we no keep receiving messages until a EntryLeaderTransfer is observed, or timeout after a second.

### DIFF
--- a/server/raft_test.go
+++ b/server/raft_test.go
@@ -393,15 +393,30 @@ func TestNRGLeaderTransfer(t *testing.T) {
 
 	require_Equal(t, newLeader.node().ID(), preferredID)
 
-	msg, err := sub.NextMsg(time.Second)
-	require_NoError(t, err)
-	require_NoError(t, sub.Unsubscribe())
+	// Expect to see a EntryLeader message
+	checkFor(t, time.Second, 0, func() (err error) {
+		msg, err := sub.NextMsg(time.Second)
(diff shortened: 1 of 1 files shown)


server/raft_test.go
```diff
@@ -393,15 +393,27 @@ func TestNRGLeaderTransfer(t *testing.T) {
 
 	require_Equal(t, newLeader.node().ID(), preferredID)
 
-	msg, err := sub.NextMsg(time.Second)
-	require_NoError(t, err)
-	require_NoError(t, sub.Unsubscribe())
+	// Expect to see a EntryLeader message
+	msgFound := false
+	deadline := time.Now().Add(time.Second)
+	for time.Now().Before(deadline) {
```
neilalexander: It might be better to use the `checkFor` helper for this, you can just return errors in the closure and it will keep trying until either the deadline is reached or there are no errors returned.

In a loop like this, the `require_` functions will normally result in an early exit, so this may not be doing what you expect.

sciascid: Makes sense, thanks! Should be fixed now.

neilalexander (approved):
LGTM

PR #7235, opened 2025-08-29
Add client stalls to ServerStatsz

wallyqs:
Also adds `omitempty` to StaleConnections

### DIFF
--- a/server/events.go
+++ b/server/events.go
@@ -375,8 +375,9 @@ type ServerStats struct {
 	Received             DataStats             `json:"received"`
 	SlowConsumers        int64                 `json:"slow_consumers"`
 	SlowConsumersStats   *SlowConsumersStats   `json:"slow_consumer_stats,omitempty"`
-	StaleConnections     int64                 `json:"stale_connections"`
+	StaleConnections     int64                 `json:"stale_connections,omitempty"`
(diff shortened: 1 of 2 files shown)


neilalexander (approved):
LGTM

server/events.go
```diff
@@ -375,8 +375,9 @@ type ServerStats struct {
 	Received             DataStats             `json:"received"`
 	SlowConsumers        int64                 `json:"slow_consumers"`
 	SlowConsumersStats   *SlowConsumersStats   `json:"slow_consumer_stats,omitempty"`
-	StaleConnections     int64                 `json:"stale_connections"`
+	StaleConnections     int64                 `json:"stale_connections,omitempty"`
 	StaleConnectionStats *StaleConnectionStats `json:"stale_connection_stats,omitempty"`
+	Stalls               int64                 `json:"stalls,omitempty"`
```
ripienaar, line 380: In vars we have `stale_connections` like here but we have `stalled_clients` there not `stalls` and the variable is `StalledClients`, seems weird to have the discrepancy here

derekcollison: Agree - can someone fix?

wallyqs: right, it should have been `stalled_clients` too

PR #7238, opened 2025-08-31
(2.12) Add `isolate_leafnode_interest` for suppressing east-west leafnode interest propagation

neilalexander:
This allows configuring the `isolate_leafnode_interest` item in the `leafnode` configuration block for the leafnode listener. When enabled, interest from leafnodes is not propagated to other isolated leafnodes, suppressing a potential storm of east-west interest when there are many leafnodes connected to an account.

### DIFF
--- a/server/opts.go
+++ b/server/opts.go
@@ -189,6 +189,10 @@ type LeafNodeOpts struct {
 	// least" test).
 	MinVersion string
 
+	// Isolate subject interest from other leafnode connections, preventing
+	// east-west propagation.
+	IsolateLeafnodeInterest bool `json:"-"`
+
 	// Not exported, for tests.
 	resolver    netResolver
 	dialTimeout time.Duration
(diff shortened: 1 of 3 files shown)


server/opts.go
```diff
@@ -2682,6 +2686,8 @@ func parseLeafNodes(v any, opts *Options, errors *[]error, warnings *[]error) er
 				*errors = append(*errors, err)
 				continue
 			}
+		case "isolate_leafnode_interest":
```
derekcollison: nit: shorter version?

derekcollison (approved):
LGTM - one minor nit 

We also may need to have mixed mode, where some are and some are not - but we can follow up on that one IMO.

ripienaar:
Be good to get this into `LeafInfo` for Leafz etc, or since its a global property maybe into `Leafz`, but seems like something you would want set per connection potentially?

PR #7242, opened 2025-09-01
Add HTTP proxy support for WebSocket leaf node connections

danbailey1000:
## Changes
- Add proxy configuration to `RemoteLeafOpts` in `server/opts.go`
- Implement HTTP CONNECT tunnel establishment in `server/leafnode.go` 
- Support both authenticated and non-authenticated proxy connections

## Use Case
Enables WebSocket leaf node connections in corporate environments where direct connections are blocked by firewalls, but HTTP proxy access is available.

## Configuration Example
```conf
leafnodes {
  remotes [{
    urls: ["ws://remote-server.example.com:443/ws"]
    
    proxy {
      url: "http://corporate-proxy.company.com:8080"
      username: "proxy_user"
      password: "proxy_pass"
      timeout: "30s"
    }
  }]
}

Signed-off-by: Dan Bailey danbailey1000@msn.com

### DIFF
--- a/server/config_check_test.go
+++ b/server/config_check_test.go
@@ -2398,6 +2398,98 @@ func TestConfigCheck(t *testing.T) {
 			errorLine: 9,
 			errorPos:  9,
 		},
+		{
+			name: "leafnode proxy with unsupported scheme",
+			config: `
+				leafnodes {
+					remotes = [
+						{
+							url: "ws://127.0.0.1:7422"
+							proxy {
+								url: "ftp://proxy.example.com:8080"
+							}
+						}
+					]
+				}
+			`,
+			err:       errors.New("proxy URL scheme must be http or https, got: ftp"),
+			errorLine: 6,
+			errorPos:  8,
+		},
+		{
+			name: "leafnode proxy with missing host",
+			config: `
+				leafnodes {
+					remotes = [
+						{
+							url: "ws://127.0.0.1:7422"
+							proxy {
+								url: "http://"
+							}
+						}
+					]
+				}
+			`,
+			err:       errors.New("proxy URL must specify a host"),
+			errorLine: 6,
+			errorPos:  8,
+		},
+		{
+			name: "leafnode proxy with username but no password",
+			config: `
+				leafnodes {
+					remotes = [
+						{
+							url: "ws://127.0.0.1:7422"
+							proxy {
+								url: "http://proxy.example.com:8080"
+								username: "testuser"
+							}
+						}
+					]
+				}
+			`,
+			err:       errors.New("proxy username and password must both be specified or both be empty"),
+			errorLine: 6,
+			errorPos:  8,
+		},
+		{
+			name: "leafnode proxy with password but no username",
+			config: `
+				leafnodes {
+					remotes = [
+						{
+							url: "ws://127.0.0.1:7422"
+							proxy {
+								url: "http://proxy.example.com:8080"
+								password: "testpass"
+							}
+						}
+					]
+				}
+			`,
+			err:       errors.New("proxy username and password must both be specified or both be empty"),
+			errorLine: 6,
+			errorPos:  8,
+		},
+		{
+			name: "leafnode proxy with WSS URL but no TLS config",
+			config: `
+				leafnodes {
+					remotes = [
+						{
+							url: "wss://127.0.0.1:7422"
+							proxy {
+								url: "http://proxy.example.com:8080"
+							}
+						}
+					]
+				}
+			`,
+			err:       errors.New("proxy is configured but remote URL wss://127.0.0.1:7422 requires TLS and no TLS configuration is provided"),
+			errorLine: 6,
+			errorPos:  8,
+		},
 	}
 
 	checkConfig := func(config string) error {
--- a/server/leafnode.go
+++ b/server/leafnode.go
@@ -285,6 +285,11 @@ func validateLeafNode(o *Options) error {
 
 	// If a remote has a websocket scheme, all need to have it.
 	for _, rcfg := range o.LeafNode.Remotes {
+		// Validate proxy configuration
+		if _, err := validateLeafNodeProxyOptions(rcfg); err != nil {
+			return err
+		}
+
 		if len(rcfg.URLs) >= 2 {
 			firstIsWS, ok := isWSURL(rcfg.URLs[0]), true
 			for i := 1; i < len(rcfg.URLs); i++ {
@@ -367,6 +372,60 @@ func validateLeafNodeAuthOptions(o *Options) error {
 		users[u.Username] = struct{}{}
 	}
 	return nil
+}
+
+func validateLeafNodeProxyOptions(remote *RemoteLeafOpts) ([]string, error) {
+	var warnings []string
+
+	if remote.Proxy.URL == _EMPTY_ {
+		return warnings, nil
+	}
+
+	proxyURL, err := url.Parse(remote.Proxy.URL)
+	if err != nil {
+		return warnings, fmt.Errorf("invalid proxy URL: %v", err)
+	}
+
+	if proxyURL.Scheme != "http" && proxyURL.Scheme != "https" {
+		return warnings, fmt.Errorf("proxy URL scheme must be http or https, got: %s", proxyURL.Scheme)
+	}
+
+	if proxyURL.Host == _EMPTY_ {
+		return warnings, fmt.Errorf("proxy URL must specify a host")
+	}
+
+	if remote.Proxy.Timeout < 0 {
+		return warnings, fmt.Errorf("proxy timeout must be >= 0")
+	}
+
+	if (remote.Proxy.Username == _EMPTY_) != (remote.Proxy.Password == _EMPTY_) {
+		return warnings, fmt.Errorf("proxy username and password must both be specified or both be empty")
+	}
+
+	if len(remote.URLs) > 0 {
+		hasWebSocketURL := false
+		hasNonWebSocketURL := false
+
+		for _, remoteURL := range remote.URLs {
+			if remoteURL.Scheme == wsSchemePrefix || remoteURL.Scheme == wsSchemePrefixTLS {
+				hasWebSocketURL = true
+				if (remoteURL.Scheme == wsSchemePrefixTLS) &&
+					remote.TLSConfig == nil && !remote.TLS {
+					return warnings, fmt.Errorf("proxy is configured but remote URL %s requires TLS and no TLS configuration is provided. When using proxy with TLS endpoints, ensure TLS is properly configured for the leafnode remote", remoteURL.String())
+				}
+			} else {
+				hasNonWebSocketURL = true
+			}
+		}
+
+		if !hasWebSocketURL {
+			warnings = append(warnings, "proxy configuration will be ignored: proxy settings only apply to WebSocket connections (ws:// or wss://), but all configured URLs use TCP connections (nats://)")
+		} else if hasNonWebSocketURL {
+			warnings = append(warnings, "proxy configuration will only be used for WebSocket URLs: proxy settings do not apply to TCP connections (nats://)")
+		}
+	}
+
+	return warnings, nil
 }
 
 // Update remote LeafNode TLS configurations after a config reload.
@@ -502,6 +561,67 @@ func (s *Server) setLeafNodeNonExportedOptions() {
 
 const sharedSysAccDelay = 250 * time.Millisecond
 
+// establishHTTPProxyTunnel establishes an HTTP CONNECT tunnel through a proxy server
+func establishHTTPProxyTunnel(proxyURL, targetHost string, timeout time.Duration, username, password string) (net.Conn, error) {
+	proxyAddr, err := url.Parse(proxyURL)
+	if err != nil {
+		// This should not happen since proxy URL is validated during configuration parsing
+		return nil, fmt.Errorf("unexpected proxy URL parse error (URL was pre-validated): %v", err)
+	}
+
+	// Connect to the proxy server
+	conn, err := natsDialTimeout("tcp", proxyAddr.Host, timeout)
+	if err != nil {
+		return nil, fmt.Errorf("failed to connect to proxy: %v", err)
+	}
+
+	// Set deadline for the entire proxy handshake
+	if err := conn.SetDeadline(time.Now().Add(timeout)); err != nil {
+		conn.Close()
+		return nil, fmt.Errorf("failed to set deadline: %v", err)
+	}
+
+	req := &http.Request{
+		Method: http.MethodConnect,
+		URL:    &url.URL{Opaque: targetHost}, // Opaque is required for CONNECT
+		Host:   targetHost,
+		Header: make(http.Header),
+	}
+
+	// Add proxy authentication if provided
+	if username != "" && password != "" {
+		req.Header.Set("Proxy-Authorization", "Basic "+base64.StdEncoding.EncodeToString([]byte(username+":"+password)))
+	}
+
+	if err := req.Write(conn); err != nil {
+		conn.Close()
+		return nil, fmt.Errorf("failed to write CONNECT request: %v", err)
+	}
+
+	resp, err := http.ReadResponse(bufio.NewReader(conn), req)
+	if err != nil {
+		conn.Close()
+		return nil, fmt.Errorf("failed to read proxy response: %v", err)
+	}
+
+	if resp.StatusCode != http.StatusOK {
+		resp.Body.Close()
+		conn.Close()
+		return nil, fmt.Errorf("proxy CONNECT failed: %s", resp.Status)
+	}
+
+	// Close the response body
+	resp.Body.Close()
+
+	// Clear the deadline now that we've finished the proxy handshake
+	if err := conn.SetDeadline(time.Time{}); err != nil {
+		conn.Close()
+		return nil, fmt.Errorf("failed to clear deadline: %v", err)
+	}
+
+	return conn, nil
+}
+
 func (s *Server) connectToRemoteLeafNode(remote *leafNodeCfg, firstConnect bool) {
 	defer s.grWG.Done()
 
@@ -541,6 +661,19 @@ func (s *Server) connectToRemoteLeafNode(remote *leafNodeCfg, firstConnect bool)
 
 	const connErrFmt = "Error trying to connect as leafnode to remote server %q (attempt %v): %v"
 
+	// Capture proxy configuration once before the loop with proper locking
+	remote.RLock()
+	proxyURL := remote.Proxy.URL
+	proxyUsername := remote.Proxy.Username
+	proxyPassword := remote.Proxy.Password
+	proxyTimeout := remote.Proxy.Timeout
+	remote.RUnlock()
+
+	// Set default proxy timeout if not specified
+	if proxyTimeout == 0 {
+		proxyTimeout = dialTimeout
+	}
+
 	attempts := 0
 
 	for s.isRunning() && s.remoteLeafNodeStillValid(remote) {
@@ -557,7 +690,25 @@ func (s *Server) connectToRemoteLeafNode(remote *leafNodeCfg, firstConnect bool)
 				err = ErrLeafNodeDisabled
 			} else {
 				s.Debugf("Trying to connect as leafnode to remote server on %q%s", rURL.Host, ipStr)
-				conn, err = natsDialTimeout("tcp", url, dialTimeout)
+
+				// Check if proxy is configured first, then check if URL supports it
+				if proxyURL != _EMPTY_ && isWSURL(rURL) {
+					// Use proxy for WebSocket connections - use original hostname, resolved IP for connection
+					targetHost := rURL.Host
+					// If URL doesn't include port, add the default port for the scheme
+					if rURL.Port() == _EMPTY_ {
+						defaultPort := "80"
+						if rURL.Scheme == wsSchemePrefixTLS {
+							defaultPort = "443"
+						}
+						targetHost = net.JoinHostPort(rURL.Hostname(), defaultPort)
+					}
+
+					conn, err = establishHTTPProxyTunnel(proxyURL, targetHost, proxyTimeout, proxyUsername, proxyPassword)
+				} else {
+					// Direct connection
+					conn, err = natsDialTimeout("tcp", url, dialTimeout)
+				}
 			}
 		}
 		if err != nil {
@@ -1287,6 +1438,13 @@ func (c *client) processLeafnodeInfo(info *Info) {
 		// otherwise if there is no TLS configuration block for the remote,
 		// the solicit side will not attempt to perform the TLS handshake.
 		if firstINFO && info.TLSRequired {
+			// Check for TLS/proxy configuration mismatch
+			if remote.Proxy.URL != _EMPTY_ && !remote.TLS && remote.TLSConfig == nil {
+				c.mu.Unlock()
+				c.Errorf("TLS configuration mismatch: Hub requires TLS but leafnode remote is not configured for TLS. When using a proxy, ensure TLS leafnode configuration matches the Hub requirements.")
+				c.closeConnection(TLSHandshakeError)
+				return
+			}
 			remote.TLS = true
 		}
 		if _, err := c.leafClientHandshakeIfNeeded(remote, opts); err != nil {
--- a/server/leafnode_proxy_test.go
+++ b/server/leafnode_proxy_test.go
@@ -1,0 +1,885 @@
+// Copyright 2025 The NATS Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package server
+
+import (
+	"errors"
+	"fmt"
+	"net"
+	"net/url"
+	"strings"
+	"testing"
+	"time"
+)
+
+// Basic HTTP proxy for testing
+type testHTTPProxy struct {
+	listener    net.Listener
+	port        int
+	username    string
+	password    string
+	started     bool
+	closeDelay  time.Duration // Delay before closing connections for robustness
+	connections []net.Conn    // Track connections for cleanup
+}
+
+func createTestHTTPProxy(username, password string) *testHTTPProxy {
+	l, err := net.Listen("tcp", "127.0.0.1:0")
+	if err != nil {
+		panic(err)
+	}
+	port := l.Addr().(*net.TCPAddr).Port
+
+	proxy := &testHTTPProxy{
+		listener:    l,
+		port:        port,
+		username:    username,
+		password:    password,
+		closeDelay:  100 * time.Millisecond, // Default delay for test robustness
+		connections: make([]net.Conn, 0),
+	}
+
+	return proxy
+}
+
+func (p *testHTTPProxy) setCloseDelay(delay time.Duration) {
+	p.closeDelay = delay
+}
+
+func (p *testHTTPProxy) start() {
+	if p.started {
+		return
+	}
+	p.started = true
+
+	go func() {
+		for {
+			conn, err := p.listener.Accept()
+			if err != nil {
+				return
+			}
+			p.connections = append(p.connections, conn)
+			go p.handleConnection(conn)
+		}
+	}()
+}
+
+func (p *testHTTPProxy) handleConnection(conn net.Conn) {
+	defer func() {
+		if p.closeDelay > 0 {
+			time.Sleep(p.closeDelay)
+		}
+		conn.Close()
+	}()
+
+	// Set read timeout to prevent hanging on malformed requests
+	conn.SetReadDeadline(time.Now().Add(10 * time.Second))
+
+	// Read the CONNECT request
+	buffer := make([]byte, 4096)
+	n, err := conn.Read(buffer)
+	if err != nil {
+		return
+	}
+
+	request := string(buffer[:n])
+	lines := strings.Split(request, "\r\n")
+
+	if len(lines) == 0 || !strings.HasPrefix(lines[0], "CONNECT ") {
+		conn.Write([]byte("HTTP/1.1 400 Bad Request\r\n\r\n"))
+		return
+	}
+
+	// Check authentication if required
+	if p.username != _EMPTY_ || p.password != _EMPTY_ {
+		authFound := false
+		for _, line := range lines {
+			if strings.HasPrefix(line, "Proxy-Authorization: Basic ") {
+				authFound = true
+				break
+			}
+		}
+		if !authFound {
+			conn.Write([]byte("HTTP/1.1 407 Proxy Authentication Required\r\n\r\n"))
+			return
+		}
+	}
+
+	// Extract target host from CONNECT line
+	parts := strings.Fields(lines[0])
+	if len(parts) < 3 {
+		conn.Write([]byte("HTTP/1.1 400 Bad Request\r\n\r\n"))
+		return
+	}
+
+	targetHost := parts[1]
+
+	// Connect to target with timeout
+	target, err := net.DialTimeout("tcp", targetHost, 5*time.Second)
+	if err != nil {
+		conn.Write([]byte("HTTP/1.1 502 Bad Gateway\r\n\r\n"))
+		return
+	}
+	defer target.Close()
+
+	// Send success response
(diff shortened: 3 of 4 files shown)

### CODE REVIEW

File: server/opts.go [hunk 1]
```diff
@@ -241,6 +241,18 @@ type RemoteLeafOpts struct {
 		NoMasking   bool `json:"-"`
 	}
 
+	// HTTP Proxy configuration for WebSocket connections
```
kozlovic, line 244: If this is only for websocket connections, should then the configuration be moved under the Websocket structure above?

server/leafnode.go
```diff
@@ -553,7 +608,32 @@ func (s *Server) connectToRemoteLeafNode(remote *leafNodeCfg, firstConnect bool)
 				err = ErrLeafNodeDisabled
 			} else {
 				s.Debugf("Trying to connect as leafnode to remote server on %q%s", rURL.Host, ipStr)
-				conn, err = natsDialTimeout("tcp", url, dialTimeout)
+				
+				// Check if this is a WebSocket URL and proxy is configured
+				if isWSURL(rURL) && remote.Proxy.URL != "" {
```
kozlovic: Checking for the remote proxy URL first would avoid the "cost" of checking if the url is WS or not. But also I would capture the `remote.Proxy.URL` before the "for-loop" starts, so we could protect the access to that field with the remote's lock (but I see that there are places where we don't, and I think that is ok since config reload of those properties would not be allowed).

nit: we use `_EMPTY_` instead of the string `""` (just to be consistent).

server/leafnode.go
```diff
@@ -553,7 +608,32 @@ func (s *Server) connectToRemoteLeafNode(remote *leafNodeCfg, firstConnect bool)
...
+					// Use proxy for WebSocket connections
+					targetHost := url
+					if port := rURL.Port(); port != "" {
+						targetHost = net.JoinHostPort(url, port)
+					} else {
+						// Default ports for ws/wss
+						defaultPort := "80"
+						if rURL.Scheme == "wss" {
```
kozlovic: We have a constant for that: `wsSchemePrefixTLS`, but I am not sure we handle the "no port/default" in normal situations, so not sure we need to do it here.

server/leafnode.go
```diff
@@ -498,6 +498,61 @@ func (s *Server) setLeafNodeNonExportedOptions() {
 
 const sharedSysAccDelay = 250 * time.Millisecond
 
+// establishHTTPProxyTunnel establishes an HTTP CONNECT tunnel through a proxy server
+func establishHTTPProxyTunnel(proxyURL, targetHost string, timeout time.Duration, username, password string) (net.Conn, error) {
+	proxyAddr, err := url.Parse(proxyURL)
```
kozlovic, line 566: If the option has been validated (missing in the PR), we should not have an error here.

server/leafnode.go
```diff
@@ -498,6 +498,61 @@ func (s *Server) setLeafNodeNonExportedOptions() {
...
+	}
+
+	// Connect to the proxy server
+	conn, err := natsDialTimeout("tcp", proxyAddr.Host, timeout)
+	if err != nil {
+		return nil, fmt.Errorf("failed to connect to proxy: %v", err)
+	}
+
+	// Send HTTP CONNECT request
+	connectReq := fmt.Sprintf("CONNECT %s HTTP/1.1\r\nHost: %s\r\n", targetHost, targetHost)
```
kozlovic: Probably need to optimize all that with a string buffer, or bytes.Buffer, etc...

server/leafnode.go
```diff
@@ -498,6 +498,61 @@ func (s *Server) setLeafNodeNonExportedOptions() {
...
+	}
+
+	// Read the response
+	response := make([]byte, 1024)
+	if err := conn.SetReadDeadline(time.Now().Add(timeout)); err != nil {
+		conn.Close()
+		return nil, fmt.Errorf("failed to set read deadline: %v", err)
+	}
+
+	n, err := conn.Read(response)
```
kozlovic: Unlikely that the response will be bigger than the `response` buffer, but we could read only partial bytes, do we care or we just fail the connection and will try again?

server/leafnode.go
```diff
@@ -498,6 +498,61 @@ func (s *Server) setLeafNodeNonExportedOptions() {
...
+	if err != nil {
+		conn.Close()
+		return nil, fmt.Errorf("failed to read proxy response: %v", err)
+	}
+
+	// Clear the read deadline
+	conn.SetReadDeadline(time.Time{})
+
+	responseStr := string(response[:n])
+	if !strings.HasPrefix(responseStr, "HTTP/1.1 200") && !strings.HasPrefix(responseStr, "HTTP/1.0 200") {
```
kozlovic: Can use `bytes.HasPrefix` so that we don't need to do string conversion.

server/opts.go
danbailey1000: I have left this outside the websocket structure because I think in future it would make sense to also be able to use the proxy for direct tcp connections also and if we put it in websocket then it would be more difficult to move it later. I have added a warning log if user configures a proxy with a tcp connection warning the proxy configuration will not be used.

server/leafnode.go
danbailey1000: done

server/leafnode.go
danbailey1000: done

server/leafnode.go
danbailey1000: I have added to log message that if there was an error then it would be unexpected

server/leafnode.go
danbailey1000: done

server/leafnode.go
danbailey1000: done

server/leafnode.go
danbailey1000: now considers a partial read possibility (even though unlikely)

danbailey1000:
> Thank you for the contribution. However, this PR cannot be accepted as-is. We would need configuration file parsing, options validation and some sort of test. But more importantly, I have a concern about how this would work. Suppose that you configure the server to have a remote with a non TLS connection, but the proxy uses TLS to communicate with the HUB that requires TLS, then this would not work because the INFO protocol received from the HUB would indicate that TLS is required and the leafnode side of things would not be able to handle that. The same if situation is reversed (leaf wants TLS but proxy connects to HUB without).

I have made changes based on your feedback

- There is now a check that leafnode and Hub tls configuration need to be aligned (both enabled or both disabled) for the proxy to work. A mismatch will raise an error 
- Add a tests for this new proxy support feature

danbailey1000:
@kozlovic I have addressed your feedback ready for another review, Thanks Dan

server/leafnode_proxy_test.go
```diff
@@ -0,0 +1,735 @@
...
+	if p.listener != nil {
+		p.listener.Close()
+	}
+}
+
+func (p *testHTTPProxy) url() string {
+	return fmt.Sprintf("http://127.0.0.1:%d", p.port)
+}
+
+func createTempConfigFile(t *testing.T, content string) string {
```
kozlovic: We don't need that since we already have `createConfFile`

server/leafnode_proxy_test.go
```diff
@@ -0,0 +1,735 @@
...
+	if remote.Proxy.Password != "pass" {
+		t.Errorf("Expected proxy password 'pass', got '%s'", remote.Proxy.Password)
+	}
+	if remote.Proxy.Timeout != 10*time.Second {
+		t.Errorf("Expected proxy timeout 10s, got %v", remote.Proxy.Timeout)
+	}
+}
+
+func TestLeafNodeProxyConfigValidation(t *testing.T) {
+	testCases := []struct {
```
kozlovic: We have a test file called `config_check_test.go` that tries to gather all config check errors. Of course, sometimes we do have tests like this one, but if that can "fit" into the existing one, it would be best.

server/opts.go
```diff
@@ -2993,6 +3035,16 @@ func parseRemoteLeafNodes(v any, errors *[]error, warnings *[]error) ([]*RemoteL
 				}
 			}
 		}
+		// Validate proxy configuration
+		if err, warns := validateLeafNodeProxyOptions(remote); err != nil {
```
kozlovic: You still need to call this from `validateLeafNode()` (which is invoked from `validateOptions` in `NewServer`. The reason is that one could programmatically configure the server and never use a config file (think embedded mode), but we still need to ensure that the provided configuration is correct. Over there, the warnings cannot be used, but at least you will catch the errors.

server/leafnode_proxy_test.go
```diff
@@ -0,0 +1,735 @@
...
+					]
+				}
+			`,
+			expectWarning: false,
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			configFile := createTempConfigFile(t, tc.config)
```
kozlovic: Use `createConfFile()` and there is no need for "defer remove" because `createConfFile()` creates the file in a `t.TempDir()` which gets automatically removed at the end of the test.

server/leafnode_proxy_test.go
```diff
@@ -0,0 +1,735 @@
...
+					}
+				}
+			]
+		}
+	`, hubOpts.Websocket.Port, proxy.url())
+
+	configFile := createConfFile(t, []byte(configContent))
+	defer removeFile(t, configFile)
+
+	spokeOpts, err := ProcessConfigFile(configFile)
```
kozlovic: Instead of process config + run server, use `RunServerWithConfig(configFile)`

server/leafnode_proxy_test.go
```diff
@@ -0,0 +1,735 @@
...
+	if err != nil {
+		t.Fatalf("Error processing config file: %v", err)
+	}
+
+	spoke := RunServer(spokeOpts)
+	defer spoke.Shutdown()
+
+	// Wait for leafnode connection to establish
+	time.Sleep(500 * time.Millisecond)
```
kozlovic: Use `checkLeafConnected(t, spoke)` and `checkLeafConnected(t, hub)`

server/leafnode_proxy_test.go
```diff
@@ -0,0 +1,735 @@
...
+
+	// Verify connection was established
```
kozlovic: These won't be needed.

server/leafnode_proxy_test.go
```diff
@@ -0,0 +1,735 @@
...
+					}
+				}
+			]
+		}
+	`, hubOpts.Websocket.Port, proxy.url())
+
+	configFile := createConfFile(t, []byte(configContent))
+	defer removeFile(t, configFile)
+
+	spokeOpts, err := ProcessConfigFile(configFile)
```
kozlovic: Same comment than before.

server/leafnode_proxy_test.go
```diff
@@ -0,0 +1,735 @@
...
+			expectWarning: false,
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			configFile := createTempConfigFile(t, tc.config)
+			defer os.Remove(configFile)
+
+			// We need to capture warnings, but ProcessConfigFile doesn't return them directly.
```
kozlovic: Actually, since we are under the server package, you can have access to the warnings. See test `TestPingIntervalOld` for more details, but basically, you can cast the returned error like this: `err.(*processConfigErr)` which will then give you access to the `warnings` slice.

server/leafnode_proxy_test.go
```diff
@@ -0,0 +1,735 @@
...
+	hubOpts := DefaultOptions()
+	hubOpts.Websocket.Host = "127.0.0.1"
+	hubOpts.Websocket.Port = -1
+	hubOpts.Websocket.NoTLS = true
+	hubOpts.LeafNode.Host = "127.0.0.1"
+	hubOpts.LeafNode.Port = -1
+	hub := RunServer(hubOpts)
+	defer hub.Shutdown()
+
+	// Wait for server to start and get actual port
```
kozlovic: You don't need that.

server/leafnode_proxy_test.go
```diff
@@ -0,0 +1,735 @@
...
+	time.Sleep(100 * time.Millisecond)
+
+	// Create HTTP proxy
+	proxy := createTestHTTPProxy("", "")
+	proxy.start()
+	defer proxy.stop()
+
+	// Create spoke server with proxy configuration via config file
+	configContent := fmt.Sprintf(`
+		leafnodes {
```
kozlovic, line 367: I would add before:
```
listen: "127.0.0.1:-1"
leafnodes {
...
```

server/leafnode_proxy_test.go
```diff
@@ -0,0 +1,735 @@
...
+	hubOpts.Websocket.Host = "127.0.0.1"
+	hubOpts.Websocket.Port = -1
+	hubOpts.Websocket.NoTLS = true
+	hubOpts.LeafNode.Host = "127.0.0.1"
+	hubOpts.LeafNode.Port = -1
+	hub := RunServer(hubOpts)
+	defer hub.Shutdown()
+
+	// Wait for server to start and get actual port
+	time.Sleep(100 * time.Millisecond)
```
kozlovic: Not needed

server/leafnode_proxy_test.go
```diff
@@ -0,0 +1,735 @@
...
+
+	// Create HTTP proxy with authentication
+	proxy := createTestHTTPProxy("testuser", "testpass")
+	proxy.start()
+	defer proxy.stop()
+
+	// Create spoke server with proxy configuration including auth via config file
+	configContent := fmt.Sprintf(`
+		leafnodes {
```
kozlovic, line 415: Add the listen config.

server/leafnode_proxy_test.go
```diff
@@ -0,0 +1,735 @@
...
+	}
+}
+
+func TestLeafNodeTLSMismatchDetection(t *testing.T) {
+	// This test simulates the TLS mismatch scenario described in the feedback:
+	// - Leafnode configured with proxy but no TLS
+	// - Hub requires TLS
+	// - Connection should fail with appropriate error message
+
+	// Create TLS configuration for hub
```
kozlovic: You could start the hub with a config file:
```
conf := createConfFile(t, []byte(`
    listen: "127.0.0.1:-1"
    leafnodes {
        listen: "127.0.0.1:-1"
        tls: {
             cert_file: "../test/configs/..."
        ..
}`))
hub, hubOpts := RunServerWithConfig(conf)
defer hub.Shutdown()
```

server/leafnode_proxy_test.go
```diff
@@ -0,0 +1,735 @@
...
+	// Create a hub server with TLS required
+	hubOpts := DefaultOptions()
+	hubOpts.LeafNode.Host = "127.0.0.1"
+	hubOpts.LeafNode.Port = -1
+	hubOpts.LeafNode.TLSConfig = tlsConfig
+	hub := RunServer(hubOpts)
+	defer hub.Shutdown()
+
+	// Wait for server to start
+	time.Sleep(100 * time.Millisecond)
```
kozlovic: Not needed

server/leafnode_proxy_test.go
```diff
@@ -0,0 +1,735 @@
...
+
+	// Create HTTP proxy
+	proxy := createTestHTTPProxy("", "")
+	proxy.start()
+	defer proxy.stop()
+
+	// Create spoke server with proxy but no TLS configuration
+	spokeOpts := DefaultOptions()
```
kozlovic: Use config file, will be much easier. See above (and look at other tests) to see how to do that.

Also, add in the leaf nodes configuration:
```
leafnodes {
     reconnect_interval: "50ms"
     remotes [
         ..
     ]
}
```
So that it will try faster to recreate the connection (see below why).

server/leafnode_proxy_test.go
```diff
@@ -0,0 +1,735 @@
...
+			},
+			// Intentionally no TLS configuration
+		},
+	}
+
+	spoke := RunServer(spokeOpts)
+	defer spoke.Shutdown()
+
+	// Wait and verify that connection was NOT established due to TLS mismatch
+	time.Sleep(1 * time.Second)
```
kozlovic: Tests take a long time to run (we have a lot of them) so we try to minimize as much as possible a test duration. You could go down to maybe 250ms here. The first attempt should happen when calling `RunServerWithConfig()` and then, it should try every 50ms, so waiting 250ms should be enough.

server/leafnode.go
```diff
@@ -365,6 +365,69 @@ func validateLeafNodeAuthOptions(o *Options) error {
 	return nil
 }
 
+// validateLeafNodeProxyOptions validates proxy options for remote leafnode configurations.
+func validateLeafNodeProxyOptions(remote *RemoteLeafOpts) (error, []string) {
+	var warnings []string
+
+	if remote.Proxy.URL == "" {
```
kozlovic: Again, if you could use `_EMPTY_` instead of `""` that would be great (for consistency).

server/leafnode.go
```diff
@@ -365,6 +365,69 @@ func validateLeafNodeAuthOptions(o *Options) error {
...
+		return fmt.Errorf("proxy username and password must both be specified or both be empty"), warnings
+	}
+
+	// Check if proxy is configured but all URLs are non-WebSocket
+	if len(remote.URLs) > 0 {
+		hasWebSocketURL := false
+		hasNonWebSocketURL := false
+
+		for _, remoteURL := range remote.URLs {
+			if remoteURL.Scheme == "ws" || remoteURL.Scheme == "wss" {
```
kozlovic: I referred previously to the constants that we have for the ws schemes, use them here.

server/leafnode_proxy_test.go
danbailey1000: done

server/leafnode_proxy_test.go
danbailey1000: moved config related tests into config_check_test

server/opts.go
danbailey1000: done

server/leafnode_proxy_test.go
danbailey1000: done

server/leafnode_proxy_test.go
danbailey1000: done

server/leafnode_proxy_test.go
danbailey1000: done

server/leafnode_proxy_test.go
danbailey1000: done

server/leafnode_proxy_test.go
danbailey1000: done

server/leafnode_proxy_test.go
danbailey1000: done

server/leafnode_proxy_test.go
danbailey1000: done

server/leafnode_proxy_test.go
danbailey1000: done

server/leafnode_proxy_test.go
danbailey1000: done

server/leafnode_proxy_test.go
danbailey1000: done

server/leafnode_proxy_test.go
danbailey1000: done

server/leafnode_proxy_test.go
danbailey1000: done

server/leafnode_proxy_test.go
danbailey1000: done

server/leafnode_proxy_test.go
danbailey1000: done

server/leafnode.go
danbailey1000: done

server/leafnode.go
danbailey1000: done

danbailey1000:
@kozlovic thanks for your review. I have made another round of changes ready for another review. Thanks Dan

server/config_check_test.go
```diff
@@ -2526,3 +2528,170 @@ func TestConfigCheckMultipleErrors(t *testing.T) {
 		}
 	}
 }
+
+// TestConfigCheckLeafNodeProxy tests proxy configuration validation
+func TestConfigCheckLeafNodeProxy(t *testing.T) {
```
kozlovic: Sorry, this is not what I meant, I meant to add to the existing `TestConfigCheck` test. See if you can make it work (check for some of the checks of `leafnodes` in this test to get an idea on how to use it.

server/config_check_test.go
```diff
@@ -2526,3 +2528,170 @@ func TestConfigCheckMultipleErrors(t *testing.T) {
...
+				if err != nil {
+					t.Errorf("Unexpected error: %v", err)
+				}
+			}
+		})
+	}
+}
+
+// TestLeafNodeProxyValidationProgrammatic tests proxy validation when configuring server programmatically
+func TestLeafNodeProxyValidationProgrammatic(t *testing.T) {
```
kozlovic: Move this one back in `leafnode_proxy_test.go` if you don't mind.

server/config_check_test.go
danbailey1000: done

server/config_check_test.go
danbailey1000: done

danbailey1000:
@kozlovic thanks for review. I have made changes suggested ready for review. Thanks Dan

server/leafnode.go
```diff
@@ -365,6 +370,69 @@ func validateLeafNodeAuthOptions(o *Options) error {
 	return nil
 }
 
+// validateLeafNodeProxyOptions validates proxy options for remote leafnode configurations.
+func validateLeafNodeProxyOptions(remote *RemoteLeafOpts) (error, []string) {
```
kozlovic: See report from the linter. Error should be last.

server/leafnode_proxy_test.go
```diff
@@ -0,0 +1,635 @@
...
+			expectWarning: false,
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			configFile := createConfFile(t, []byte(tc.config))
+
+			// Use Options.ProcessConfigFile to capture warnings
+			opts := &Options{}
```
kozlovic: Remove, instead use `opts, err := ProcessConfigFile(configFile)`

server/leafnode_proxy_test.go
```diff
@@ -0,0 +1,635 @@
...
+					errTyped, ok := err.(*processConfigErr)
+					if ok && len(errTyped.Warnings()) > 0 {
+						t.Errorf("Unexpected warnings: %v", errTyped.Warnings())
+					} else if !ok {
+						t.Errorf("Unexpected error: %v", err)
+					}
+				}
+			}
+
+			// Verify configuration was parsed successfully
```
kozlovic: Remove all that since now you handle the errors and warnings.

server/leafnode.go
danbailey1000: done

server/leafnode_proxy_test.go
danbailey1000: done

server/leafnode_proxy_test.go
danbailey1000: done

danbailey1000:
> I have enabled the workflow run and the linter/checks spotted few things. I would recommend that you rebase and squash all your commits into 1 and you need to have a signoff (see https://github.com/nats-io/nats-server/blob/main/CONTRIBUTING.md)

@kozlovic I have made changes and squashed into one signed commit. Thanks Dan

danbailey1000:
@kozlovic I noticed an issue with my squash commit. Now corrected. Thanks Dan

server/leafnode_proxy_test.go
```diff
@@ -0,0 +1,608 @@
...
+	}
+	defer conn.Close()
+
+	// Test that we can communicate through the tunnel
+	conn.Write([]byte("Hello"))
+
+	buffer := make([]byte, 1024)
+	n, err := conn.Read(buffer)
+	if err != nil {
+		t.Fatalf("Failed to read from proxy tunnel: %v", err)
```
kozlovic, line 574: Ok, we have a problem. On my laptop, running this test (and the `TestEstablishHTTPProxyTunnelWithAuth` test), I would often get a failure here. This is likely because the `go func()` above that does the write immediately closes the connection after writing. So I added sync.WaitGroup{} and some go channel to make sure that the `go func()` waits for the read to complete. But the issue now is that the read will wait here for ever (when the test would otherwise fail). So we would need this test to have some deadline set to not have the test suite blocks until it times out. But the real issue is that the `establishHTTPProxyTunnel()` function in leafnode.go is consuming the message "Hello from target server" as part of the HTTP response from the proxy itself. This is not good.

You need to revisit that function to make sure that it consumes ONLY the http response, nothing else.

Could you also name all tests in this file with the same prefix, say "TestLeafNodeHttpProxy" so that we can easily run all the tests? (We could do then something like `go test -race -v -run=TestLeafNodeHttpProxy" ./server -count=10 -vet=off -failfast` and make sure they all pass).

server/leafnode_proxy_test.go
```diff
@@ -0,0 +1,608 @@
...
+	}
+
+	if !strings.Contains(err.Error(), "proxy CONNECT failed") {
+		t.Errorf("Expected proxy authentication error, got: %v", err)
+	}
+}
+
+// TestLeafNodeProxyValidationProgrammatic tests proxy validation when configuring server programmatically
+func TestLeafNodeProxyValidationProgrammatic(t *testing.T) {
+	// Test invalid proxy scheme
```
kozlovic: This test would be a good candidate for the kind of test with `for _, test := range []struct{` where you have a name, a returned option and the expected error, and the body of the for would use the test's option and invoke `validateLeafNode()` and check that we get the expected error. I think that you may want to add a check for empty proxy URL, invalid proxy URL (parse would fail), and invalid timeout (say negative value).

kozlovic:
@danbailey1000 Also, have you tried it with a real http proxy and a NATS Server hub that requires TLS? I wonder if it will even work since most TLS actors will fail if the remote is sending plain text first, which by default NATS Server does (sends the INFO protocol in plain text when getting a TCP connection). We did add the ability to do "TLS first" (see `TLSHandshakeFirst` option in both leafnode and leafnode remote options), so there should be a way to make it work, but I would appreciate if you had a chance to actually run the feature in real life setup and report that it works as expected. Thanks!

wallyqs:
usually when putting a proxy in front of nats it should work by using the TLS first / `tls_handshake_first` from the remote client side making the connection then on the server port in the tls config putting an empty block like this along with the `allow_non_tls` flag so that INFO announces `tls_available` to the clients:

```
tls {}
allow_non_tls: true
```

This setup works for client connections 4222 at least.

server/leafnode_proxy_test.go
danbailey1000: Now changed to use read buffer approach to ensure only http response is consumed

server/leafnode_proxy_test.go
danbailey1000: done

danbailey1000:
> @danbailey1000 Also, have you tried it with a real http proxy and a NATS Server hub that requires TLS? I wonder if it will even work since most TLS actors will fail if the remote is sending plain text first, which by default NATS Server does (sends the INFO protocol in plain text when getting a TCP connection). We did add the ability to do "TLS first" (see `TLSHandshakeFirst` option in both leafnode and leafnode remote options), so there should be a way to make it work, but I would appreciate if you had a chance to actually run the feature in real life setup and report that it works as expected. Thanks!

@kozlovic I have added some tests that uses real nats hub and leaf servers configured via proxy and using TLS and using the tlshandshakefirst and confirmed it is working as expected.

danbailey1000:
@kozlovic @wallyqs I have pushed changes that address the points in your last review. I can squash again in single commit if you want. Thx Dan

danbailey1000:
@kozlovic tests now removed. Thx
- Not sure why the tests did not run for you. I am running on windows machine, and I noticed other many tests do not pass on windows as I guess they are designed to execute on linux machines.
- I am not the admin of the real proxy and NATS servers where we ran into the proxy issues and NATS was not using tls at the moment hence I made a test starting an actual NATS server and leaf and simulated proxy. I think this gives more confidence that these changes work with tls also.

server/leafnode_proxy_test.go
```diff
@@ -0,0 +1,889 @@
...
+	if err != nil {
+		t.Fatalf("Failed to create test server: %v", err)
+	}
+	defer listener.Close()
+
+	targetPort := listener.Addr().(*net.TCPAddr).Port
+	targetHost := fmt.Sprintf("127.0.0.1:%d", targetPort)
+
+	// Use WaitGroup to properly synchronize server response
+	var wg sync.WaitGroup
```
kozlovic: I would like you to make the following changes because, using t.Fatal in a go routine would produce a race (if an error was really happening). The proposed approach solves not doing a t.Fatal/t.Error in a go routine, does the proper synchronization (with timeout) of the go routine (instead of having a `go func() { wg.Wait() close(done) }()`.

[fix.patch](https://github.com/user-attachments/files/22257851/fix.patch)

kozlovic (changes requested):
Please review and apply the proposed fix for the tests. Also, you would need to squash all your commits (and possibly rebase to take advantage of the latest changes in main - if any) because *each* commit needs to be signed off. If I were to trigger the CI tests, it would fail because of that. Thanks again and appreciate your patience.

danbailey1000:
> Please review and apply the proposed fix for the tests. Also, you would need to squash all your commits (and possibly rebase to take advantage of the latest changes in main - if any) because _each_ commit needs to be signed off. If I were to trigger the CI tests, it would fail because of that. Thanks again and appreciate your patience.

@kozlovic Patch now applied (thanks for that :)). Squashed changes into one commit.

neilalexander:
Will need to hold merging for now as we're in a feature freeze ahead of 2.12. Will come back to this afterwards. Thanks!

acormier-maia:
@danbailey1000 I have made some small changes to your PR and tested using a proxy with a regular tcp nats connection which was successful.  AFAIK an HTTP proxy which implements the CONNECT method should not care whether the protocol on top is HTTP or raw TCP (nats). 

PR is up here: https://github.com/danbailey1000/nats-server/pull/1/files

danbailey1000:
> @danbailey1000 I have made some small changes to your PR and tested using a proxy with a regular tcp nats connection which was successful. AFAIK an HTTP proxy which implements the CONNECT method should not care whether the protocol on top is HTTP or raw TCP (nats).
> 
> PR is up here: https://github.com/danbailey1000/nats-server/pull/1/files

@acormier-maia  thanks a lot. Ideally we can also include the tcp support.  This PR has been approved already so I am not sure the best approach now. Maybe wait until this PR is merged, and then we can open another PR to add the tcp support. @kozlovic what would you recommend? thx

kozlovic:
@danbailey1000 Yes, maybe wait for it to be merged and then another PR can be submitted.

neilalexander:
Would you mind rebasing on top of latest `main` please?

danbailey1000:
> Would you mind rebasing on top of latest `main` please?

@neilalexander  now done. Thanks

server/leafnode.go
```diff
@@ -502,6 +561,132 @@ func (s *Server) setLeafNodeNonExportedOptions() {
 
 const sharedSysAccDelay = 250 * time.Millisecond
 
+// establishHTTPProxyTunnel establishes an HTTP CONNECT tunnel through a proxy server
+func establishHTTPProxyTunnel(proxyURL, targetHost string, timeout time.Duration, username, password string) (net.Conn, error) {
```
neilalexander, line 565: I think I would really rather us not build and parse HTTP requests/responses by hand here when the Go standard library has the ability to do this for us, i.e. something like this:
```
	conn, err := natsDialTimeout("tcp", proxyAddr.Host, timeout)
	if err != nil {
		return nil, fmt.Errorf("failed to connect to proxy: %w", err)
	}

	req := &http.Request{
		Method: http.MethodConnect,
		URL:    &url.URL{Opaque: targetHost}, // Opaque is required here
		Host:   targetHost,
		Header: make(http.Header),
	}
 	// do we need auth here?
	req.Header.Set("Proxy-Authorization", "Basic " + ...)

	if err := req.Write(conn); err != nil {
		conn.Close()
		return nil, fmt.Errorf("req.Write: %w", err)
	}

	resp, err := http.ReadResponse(bufio.NewReader(conn), req)
	if err != nil {
		conn.Close()
		return nil, fmt.Errorf("http.ReadResponse: %w", err)
	}
	if resp.StatusCode != http.StatusOK {
		conn.Close()
		return nil, fmt.Errorf("status code: %s", resp.Status)
	}

	// now use "conn"
```
We already have `net/http` imported so this is fine to use.

danbailey1000: @neilalexander I have changed this to now use the standard http library. Thx

neilalexander (approved):
LGTM, thanks!

PR #7244, opened 2025-09-01
壮举：自动注入JWT用户标签作为消息标题

li-zenghui:
## Description

Parse JWT user tags ('key:value') and inject as message headers during auth. Enables client apps to track message senders without manual implementation.

## Features

- Support direct JWT auth and auth callout modes
- Validate headers and prevent injection attacks  
- Add comprehensive unit tests

## Use Case

Chat/gaming apps can automatically identify message senders from JWT tags like 'user-id:12345' -> header 'user-id: 12345'

## Implementation Details

- Parse JWT tags in `key:value` format from UserClaims.Tags
- Inject headers in both `auth.go` (direct JWT) and `auth_callout.go` (callout mode)
- Validate header names per HTTP standards (RFC 7230)
- Prevent CRLF injection attacks
- Only apply to CLIENT connections

JWT tags are automatically injected as message headers:

```go
// 1. Set JWT tags once during user creation (using auth_callout to create JWT)
uc := jwt.NewUserClaims(userPubKey)
uc.Tags.Add("user-id:12345")

// 2. Client connects with JWT - tags automatically become headers
nc, _ := nats.Connect("nats://server:4222", nats.UserCredentials("user.creds"))

// 3. Every message automatically includes user identity in headers
nc.Publish("chat.room1", []byte("Hello world"))
// ↓ Automatically includes headers:
// user-id: 12345

// 4. Receivers get user info from headers (no extra database queries needed)
nc.Subscribe("chat.room1", func(msg *nats.Msg) {
    userID := msg.Header.Get("user-id")      // "12345"
    fmt.Printf("Message from %s : %s\n",  userID, string(msg.Data))
})
```

## Testing

Added unit tests covering:
- Basic functionality and validation
- Security (injection attack prevention)
- Edge cases and error handling
- Different client types

Signed-off-by: li-zenghui <lizenghui0827@163.com>

### DIFF
--- a/server/auth.go
+++ b/server/auth.go
@@ -1033,6 +1033,11 @@ func (s *Server) processClientOrLeafAuthentication(c *client, opts *Options) (au
 		c.nameTag = juc.Name
 		c.mu.Unlock()
 
+		// Inject header from jwt tags
+		if len(juc.Tags) > 0 {
+			c.injectHeader(juc.Tags)
+		}
+
 		// Check if we need to set an auth timer if the user jwt expires.
 		c.setExpiration(juc.Claims(), validFor)
 
--- a/server/auth_callout.go
+++ b/server/auth_callout.go
@@ -301,6 +301,11 @@ func (s *Server) processClientOrLeafCallout(c *client, opts *Options) (authorize
 			c.mu.Unlock()
 		}
 
+		// Inject header from jwt tags
+		if len(arc.Tags) > 0 {
+			c.injectHeader(arc.Tags)
+		}
+
 		// Check if we need to set an auth timer if the user jwt expires.
 		c.setExpiration(arc.Claims(), expiration)
 
--- a/server/client.go
+++ b/server/client.go
@@ -299,6 +299,9 @@ type client struct {
 	nameTag string
 
 	tlsTo *time.Timer
+
+	// Need to inject headers
+	ihdr []byte
 }
 
 type rrTracking struct {
@@ -3393,7 +3396,7 @@ func (c *client) msgHeaderForRouteOrLeaf(subj, reply []byte, rt *routeTarget, ac
 func (c *client) msgHeader(subj, reply []byte, sub *subscription) []byte {
 	// See if we should do headers. We have to have a headers msg and
 	// the client we are going to deliver to needs to support headers as well.
-	hasHeader := c.pa.hdr > 0
+	hasHeader := c.pa.hdr > 0 || len(c.ihdr) > 0
 	canReceiveHeader := sub.client != nil && sub.client.headers
 
 	var mh []byte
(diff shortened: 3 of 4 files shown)


ripienaar:
NATS goes to quite some lengths to avoid coupling publishers and subscribers, it's quite a deep rooted philosophy. So this being hard and an application problem is by design.

The exception is across account boundaries where when importing and exporting the 2 accounts can opt to share information about the messages and you'll then have a new header added that includes user information.

In this PR, what prevents a user without the JWT tags from adding the header `user-id` and just setting it to any value they want?

li-zenghui:
> NATS goes to quite some lengths to avoid coupling publishers and subscribers, it's quite a deep rooted philosophy. So this being hard and an application problem is by design.
> 
> The exception is across account boundaries where when importing and exporting the 2 accounts can opt to share information about the messages and you'll then have a new header added that includes user information.
> 
> In this PR, what prevents a user without the JWT tags from adding the header `user-id` and just setting it to any value they want?

Thank you for your reply. I want to use nats as much as possible in my project. I'll answer your question
1. Users cannot forge their identities because I have discovered the behavior of msg.Header. Get ("key") is consistent with the http header. When multiple values correspond to a key, the first value is obtained by default. Therefore, I placed the nats injection header at the front to ensure that the nats injection header has a higher priority, so that when the user executes msg.Header.Get ("key"), they will only obtain a safe value
2. I don't understand what problems cross-account information sharing can cause. Since the message body can be shared, why can't the message header be shared
3. Regarding what you said, "NATS does its utmost to avoid coupling between publishers and subscribers." I completely agree. However, our application communicates with the server via nats. When a user initiates a request through nats using an application, the server is unable to confirm who sent the request. If it is directly written into the requested data, it may be forged. So I'm wondering why nats can't append some data to each message by default? Of course, only some accounts need it. Communication between services may not require it. So we can control for ourselves whether to attach labels in the auth callout.

ripienaar:
> > NATS goes to quite some lengths to avoid coupling publishers and subscribers, it's quite a deep rooted philosophy. So this being hard and an application problem is by design.
> > The exception is across account boundaries where when importing and exporting the 2 accounts can opt to share information about the messages and you'll then have a new header added that includes user information.
> > In this PR, what prevents a user without the JWT tags from adding the header `user-id` and just setting it to any value they want?
> 
> Thank you for your reply. I want to use nats as much as possible in my project. I'll answer your question
> 
> 1. Users cannot forge their identities because I have discovered the behavior of msg.Header. Get ("key") is consistent with the http header. When multiple values correspond to a key, the first value is obtained by default. Therefore, I placed the nats injection header at the front to ensure that the nats injection header has a higher priority, so that when the user executes msg.Header.Get ("key"), they will only obtain a safe value

But any JWT without these tags won't have this restriction and so could set any header they like - including forge a user id.

> 2. I don't understand what problems cross-account information sharing can cause. Since the message body can be shared, why can't the message header be shared

Headers are always sent across accounts, if you however enable the sharing property you get an ADDITIONAL header set by the server that includes user information of the connection, no1 can forge that header in any way.

> 3. Regarding what you said, "NATS does its utmost to avoid coupling between publishers and subscribers." I completely agree. However, our application communicates with the server via nats. When a user initiates a request through nats using an application, the server is unable to confirm who sent the request. If it is directly written into the requested data, it may be forged. So I'm wondering why nats can't append some data to each message by default? Of course, only some accounts need it. Communication between services may not require it. So we can control for ourselves whether to attach labels in the auth callout.

See above about enabling sharing cross account boundaries.

Application level auth is an application concern.  We dont update TCP to authenticate HTTP requests, we update HTTP. NATS is at a lower level than your applications authentication.

There are other ways to do this also, lets say you wish to publish to `foo.bar.>` you can put in the users JWT that this subject is imported from another account and the server will insert `foo.<user id>.bar.>`, where user id is something of your choosing.  Then every message on those subjects will transparently tagged with your user ID of choices baked into the JWT.

li-zenghui:
Ok. Thank you. I'll think about it again

PR #7255, opened 2025-09-03
[FIXED] LeafNode: Subject propagation with daisy chain and import/export

kozlovic:
The issue can be observed for instance with JetStream and source streams, that would not be populating properly depending on where a server connects to in a given topology. See the new test `TestLeafNodeDaisyChainWithAccountImportExport` for a demonstration of the issue without the fix in place.

### DIFF
--- a/server/client.go
+++ b/server/client.go
@@ -3171,6 +3171,13 @@ func (c *client) addShadowSub(sub *subscription, ime *ime, enact bool) (*subscri
 	// Update our route map here. But only if we are not a leaf node or a hub leafnode.
 	if c.kind != LEAF || c.isHubLeafNode() {
 		c.srv.updateRemoteSubscription(im.acc, &nsub, 1)
+	} else if c.kind == LEAF {
+		// Update all leafnodes that connect to this server. Note that we could have
+		// used the updateLeafNodes() function since when it does invoke updateSmap()
+		// this function already takes care of not sending to a spoke leafnode since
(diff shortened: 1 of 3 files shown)


server/client.go
```diff
@@ -3171,6 +3171,13 @@ func (c *client) addShadowSub(sub *subscription, ime *ime, enact bool) (*subscri
 	// Update our route map here. But only if we are not a leaf node or a hub leafnode.
 	if c.kind != LEAF || c.isHubLeafNode() {
 		c.srv.updateRemoteSubscription(im.acc, &nsub, 1)
+	} else if c.kind == LEAF {
+		// Update all leafnodes that connect to this server. Note that we could have
+		// used the updateLeafNodes() function since when it does invoke updateSmap()
+		// this function already takes care of not sending to a spoke leafnode since
+		// the `nsub` here is already from a spoke leafnode, but to be explicit, we
+		// use this version that updates only leafnodes that connect to this server.
+		im.acc.updateLeafNodesEx(&nsub, 1, true)
```
kozlovic, line 3180: @derekcollison I am pretty sure that - as of now, and with the test case I have added - we would not need a new function, but decided after much thinking, that I would have this version that excludes some type of leafnode connections. Let me know if you want me to remove that and simply use updateLeafNodes().

derekcollison: Agree

derekcollison (approved):
LGTM

PR #7258, opened 2025-09-04
Fix internal JS clients disconnecting on account max connection updates

wallyqs:
Fixes a condition where when running into a account max connections limit, sometimes a number of JetStream streams would have lost interest after getting a `JETSTREAM - maximum account active connections exceeded` error.


derekcollison (approved):
LGTM

Should this just be a helper so we do not have to go through list twice?

derekcollison:
Right now we loop through that list twice - which if big could be costly.

wallyqs:
updated with a helper to avoid looping twice over clients.

derekcollison (approved):
LGTM

PR #7259, opened 2025-09-04
[FIXED] LeafNode: Don't forward shadow sub unsubscribe if spoke leafnode

kozlovic:
This is related to PR #5519 where the propagation of a shadow subscription was suppressed for a subscription coming from a spoke leafnode connection, but the handling of the unsubscribe of such shadow subscription was not fixed, which could result in removing a legitimate interest.

The test `TestLeafNodeDupeDeliveryQueueSubAndPlainSub` that was added in PR #5519 was modified to demonstrate what would have been the issue without the fix in this PR.

### DIFF
--- a/server/client.go
+++ b/server/client.go
@@ -3286,14 +3286,12 @@ func (c *client) unsubscribe(acc *Account, sub *subscription, force, remove bool
 
 	// Check to see if we have shadow subscriptions.
 	var updateRoute bool
-	var updateGWs bool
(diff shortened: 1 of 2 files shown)


kozlovic:
@derekcollison This is the handling of the unsub of a shadow subscription that was not done in PR #5519 that you fixed last year.

derekcollison (approved):
LGTM

PR #7261, opened 2025-09-05
Fix TestNRGNoResetOnAppendEntryResponse

sciascid:
The test proposes a message and expects it to be committed via rg.waitOnTotal(). This was nondeterministic:

- If the leader first received an AppendEntriesResponse from a follower, the proposed message could be committed.
- If the leader first received the AppendEntriesResponse with a higher term (injected by the test), the leader stepped down and the proposed message could not be committed.

The test would sometimes fail depending on the ordering of these events. This fix ensures that the test always follows the second path, where the leader steps down before the proposal can be committed.

### DIFF
--- a/server/raft_helpers_test.go
+++ b/server/raft_helpers_test.go
@@ -100,6 +100,18 @@ func (sg smGroup) unlockAll() {
 	}
 }
 
+// Acquire the lock on all follower nodes.
+func (sg smGroup) lockFollowers() []stateMachine {
+	var locked []stateMachine
+	for _, sm := range sg {
+		if !sm.node().Leader() {
+			locked = append(locked, sm)
+			sm.node().(*raft).Lock()
+		}
+	}
+	return locked[:]
+}
+
 // Create a raft group and place on numMembers servers at random.
 // Filestore based.
 func (c *cluster) createRaftGroup(name string, numMembers int, smf smFactory) smGroup {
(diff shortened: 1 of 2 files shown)


server/raft_helpers_test.go
```diff
@@ -100,6 +100,18 @@ func (sg smGroup) unlockAll() {
 	}
 }
 
+// Acquire the lock on all follower nodes.
+func (sg smGroup) lockFollowers() [](stateMachine) {
```
neilalexander: I think this can just be `[]stateMachine`.

server/raft_test.go
```diff
@@ -936,10 +938,23 @@ func TestNRGNoResetOnAppendEntryResponse(t *testing.T) {
...
+
+	// Temporarily prevent followers to respond
+	locked := rg.lockFollowers()
+
+	// Generate an append entry that the subscriber above can respond to.
 	lsm.proposeDelta(5)
-	rg.waitOnTotal(t, 5)
+
+	// Expect the leader to step down
+	for isLeader == true {
```
neilalexander: Might be better to use `checkFor()` here with a timeout set, otherwise this could potentially stick for the entire timeout of the test if something goes wrong.

sciascid: I'm not sure I understand this comment. require_ChanRead() already takes a timeout.

server/raft_helpers_test.go
sciascid: Fixed

server/raft_test.go
sciascid: ahh nevermind I was looking a the wrong line...

sciascid: Fixed

neilalexander (approved):
LGTM

PR #7263, opened 2025-09-05
Add linux/arm64 to nightly, make it multiplatform

wallyqs:
Makes nightly multi platform and add support for arm64 which we use more in testing lately.

Generates:

* Single
  - synadia/nats-server:nightly-amd64
  - synadia/nats-server:nightly-arm64
  - synadia/nats-server:nightly-20250905-amd64
  - synadia/nats-server:nightly-20250905-arm64

* Multi
  - synadia/nats-server:nightly (both amd64 and arm64)
  - synadia/nats-server:nightly-20250905

Signed-off-by:  Waldemar Quevedo <wally@nats.io>

### DIFF
--- a/.github/actions/nightly-release/action.yaml
+++ b/.github/actions/nightly-release/action.yaml
@@ -26,6 +26,12 @@ runs:
       with:
         go-version: stable
 
+    - name: Set up QEMU
+      uses: docker/setup-qemu-action@v3
+
+    - name: Set up Docker Buildx
+      uses: docker/setup-buildx-action@v3
+
     - name: Build and push Docker images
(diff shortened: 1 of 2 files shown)


neilalexander:
I'm fine with adding the ARM64 builds but I don't see much use in the `-amd64` and `-arm64` tag prefixes, Docker should only pull the right layers for the right architecture regardless.

neilalexander:
I would prefer we get multi-arch images right rather than having new suffixes and breaking an established naming convention.

samuelattwood:
In my view this doesn't break established naming, the multi-arch manifest remains and is how this will typically be pulled, it just also gives a tag to the underlying SHAs that the manifest references. This is a pretty typical approach with GoReleaser.

PR #7276, opened 2025-09-08
Explains the function of the JSLastSequence message header

jnmoyne:
Explains the function of the JSLastSequence message header and add testing it's functionality to the existing tests in all 3 features generating the header


server/stream.go
```diff
@@ -518,14 +518,15 @@ const (
...
+// Headers for republished messages and direct get responses.
 const (
 	JSStream       = "Nats-Stream"
 	JSSequence     = "Nats-Sequence"
 	JSTimeStamp    = "Nats-Time-Stamp"
 	JSSubject      = "Nats-Subject"
-	JSLastSequence = "Nats-Last-Sequence"
-	JSNumPending   = "Nats-Num-Pending"
+	JSLastSequence = "Nats-Last-Sequence" // The sequence number of the precedent message in the batch of responses or of precedent republished message.
+	//"0" means no value (e.g. first message in the batch).
```
derekcollison: Would clean this one up IMO. Meaning try to get on one line or if not, make it above the const.

server/jetstream_test.go
```diff
@@ -13795,14 +13795,21 @@ func TestJetStreamStreamRepublishHeadersOnly(t *testing.T) {
 	}
 
 	checkSubsPending(t, sub, toSend)
-	m, err := sub.NextMsg(time.Second)
-	require_NoError(t, err)
+	last := "0"
```
derekcollison, line 13798: Should we make this a const somewhere in server and use here?

jnmoyne: You mean like `_EMPTY_ `?

Not sure it's worth it IMHO. "0" is used in a few places in the code, but like 95+% in the unit tests. It can mean different things, typically a sequence number but there's some exceptions (e.g. partition number). In this case it's a string as a shortcut/simplification of the code though a sequence number is obviously an int. But happy to do it if you feel like it would be better.

server/stream.go
jnmoyne: Tightened to one line trying to keep it at succinct as possible

server/jetstream_test.go
jnmoyne: Or if you mean something to indicate initial value for last header or start of a batch or republication, yeah it's wouldn't be a bad thing. But I think that the added comment on the const for the field is enough to convey that 0 is the first message. Anyone wanting to know about the const now should know 0 is expected at the start, not sure it's worth a server level constant (would only be used in those 3 tests anyways). 🤷‍♂️

PR #7280, opened 2025-09-09
(2.12) Allow configuring `X25519MLKEM768` curve

### LINKED ISSUE #7279: NATS server won't boot when curve_preferences: [ "X25519Kyber768"]
### Observed behavior

The following config item appears to prevent nats-server from starting.
curve_preferences: [ "X25519Kyber768"] 

### Expected behavior

X25519Kyber768 is supposed to be the default curve as of August 2024 - Go 1.23.

### Server and client version

server: 2.11.8
client: doesn't matter

### Host environment

Doesn't matter.  For example, Windows server.

### Steps to reproduce

add this line to the config file.

curve_preferences: [ "X25519Kyber768"]

neilalexander:
This fixes #7279.

### DIFF
--- a/server/ciphersuites.go
+++ b/server/ciphersuites.go
@@ -42,17 +42,19 @@ func defaultCipherSuites() []uint16 {
 
 // Where we maintain available curve preferences
 var curvePreferenceMap = map[string]tls.CurveID{
-	"X25519":    tls.X25519,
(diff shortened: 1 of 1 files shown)


MauriceVanVeen (approved):
LGTM

PR #7282, opened 2025-09-09
(2.12) Reuse `perAccountCache` structs when increasing generation or pruning

neilalexander:
Previously when pruning the L1 cache, we would have thrown away potentially 1024 objects at once, which would have resulted in work for the GC to do and would also significantly dent the cache quality on each prune.

Now we will only evict the minimum number of entries at a time to try and keep the cache closer to full, and we will reuse `perAccountCache` structs when they are pruned or when they are from an old generation to reduce the number of new allocations.

### DIFF
--- a/server/client.go
+++ b/server/client.go
@@ -508,15 +508,13 @@ func (rcf readCacheFlag) isSet(c readCacheFlag) bool {
 }
 
 const (
-	defaultMaxPerAccountCacheSize   = 8192
-	defaultPrunePerAccountCacheSize = 1024
-	defaultClosedSubsCheckInterval  = 5 * time.Minute
+	defaultMaxPerAccountCacheSize  = 8192
+	defaultClosedSubsCheckInterval = 5 * time.Minute
 )
 
 var (
-	maxPerAccountCacheSize   = defaultMaxPerAccountCacheSize
-	prunePerAccountCacheSize = defaultPrunePerAccountCacheSize
-	closedSubsCheckInterval  = defaultClosedSubsCheckInterval
+	maxPerAccountCacheSize  = defaultMaxPerAccountCacheSize
+	closedSubsCheckInterval = defaultClosedSubsCheckInterval
 )
 
 // perAccountCache is for L1 semantics for inbound messages from a route or gateway to mimic the performance of clients.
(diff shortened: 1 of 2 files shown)


MauriceVanVeen (approved):
LGTM

server/client.go
```diff
@@ -508,15 +508,13 @@ func (rcf readCacheFlag) isSet(c readCacheFlag) bool {
 }
 
 const (
-	defaultMaxPerAccountCacheSize   = 8192
-	defaultPrunePerAccountCacheSize = 1024
```
derekcollison, line 512: We shoudl test this since this was a key design flow that mimic'd the approach Salvatore took in Redis.

server/client.go
```diff
@@ -6035,13 +6033,33 @@ func (c *client) getAccAndResultFromCache() (*Account, *SublistResult) {
...
+				pac = p
+				if len(c.in.pacache) < maxPerAccountCacheSize {
+					break
+				}
+			}
 		}
 
+		// If we can reuse the pac from earlier (i.e. we loaded one but it was an
+		// old generation or we pruned the cache) then do so.
+		if pac == nil {
```
derekcollison, line 6054: Should we consider a formal pool and still take out N per prune?

server/client.go
neilalexander: I did a number of benchmarks with this and saw some moderate improvement in the fanout benchmarks, and the JetStream replicated benchmarks for publishing (concurrent and non-concurrent) saw anywhere between a 15-25% improvement. The rest was all just within the error tolerance +/-.

server/client.go
neilalexander: I did consider a pool and larger batches of frees but it will still likely result in more GC pressure as the pool won't usually hold onto that many entries. It still also doesn't really solve the problem of us having to repopulate 1/8th of the cache each time, possibly with more new allocations, if the pool isn't well populated.

derekcollison: Did you test it or just a hunch?

neilalexander: Not tested a sync pool yet. I can do so.

derekcollison: I want to know if this will cause some use case patterns to degrade..

PR #7283, opened 2025-09-09
(2.12) Fix `no_hdr` requests on direct get by last subject endpoint

neilalexander:
Previously this would not work because we would error if a body was provided. Now we only error if field values are provided that don't make sense for the endpoint.

Co-authored-by: Piotr Piotrowski <piotr@synadia.com>

### DIFF
--- a/server/stream.go
+++ b/server/stream.go
@@ -4932,7 +4932,6 @@ func (mset *stream) processDirectGetLastBySubjectRequest(_ *subscription, c *cli
 	if len(reply) == 0 {
 		return
 	}
-	var minLastSeq uint64
 	hdr, msg := c.msgParts(rmsg)
 	if errorOnRequiredApiLevel(hdr) {
 		hdr := []byte("NATS/1.0 412 Required Api Level\r\n\r\n")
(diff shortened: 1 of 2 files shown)


server/stream.go
```diff
@@ -4932,31 +4932,37 @@ func (mset *stream) processDirectGetLastBySubjectRequest(_ *subscription, c *cli
...
 		}
+	}
 
-		if req.MinLastSeq == 0 {
-			hdr := []byte("NATS/1.0 408 Bad Request\r\n\r\n")
-			mset.outq.send(newJSPubMsg(reply, _EMPTY_, _EMPTY_, hdr, nil, nil, 0))
-			return
-		}
-		minLastSeq = req.MinLastSeq
+	// Error if request fields are provided that don't make sense for this endpoint.
```
piotrpio: Should't we also return an error if `key != req.LastFor`?

neilalexander: We could do, currently we just overwrite it.

piotrpio: Also, any reason for this switch pattern with fallthough here instead of just:
```
	// Error if request fields are provided that don't make sense for this endpoint.
	switch {
	case req.NextFor != _EMPTY_, len(req.MultiLastFor) > 0,
		req.Seq != 0, req.Batch != 0,
		req.StartTime != nil,
		req.UpToSeq != 0, req.UpToTime != nil:

		hdr := []byte("NATS/1.0 408 Bad Request\r\n\r\n")
		mset.outq.send(newJSPubMsg(reply, _EMPTY_, _EMPTY_, hdr, nil, nil, 0))
		return
	}
```

piotrpio: I mean, in this particular case those are the same right?

neilalexander: Mostly just because I like the indentation, if I were just going to do one case like that then I would ordinarily just do an `if`. But either works.

piotrpio: Not feeling strongly about this, mostly I was curious

PR #7290, opened 2025-09-10
Correct and simplify stream assignment check

sciascid:
Remove misleading comment for the loop at the end of isStreamAssigned: "Check if we are the leader of this raftGroup [...]". This code is checking for membership, not leadership of the group.
Instead of fixing the comment, replace the whole loop with a call to isMember.
Same for isConsumerAssigned.


neilalexander (approved):
LGTM

PR #7291, opened 2025-09-10
Fixes handling a batched direct get request where the starting sequence number requested is deleted

jnmoyne:
When processing a direct get batch (even without NextFor) deleted messages are skipped over. This fixes/extends this so the skipping also happens if the requested sequence number happens to be deleted.


MauriceVanVeen (approved):
LGTM

neilalexander (approved):
LGTM

PR #7295, opened 2025-09-10
Filestore fixes

neilalexander:
This PR fixes the following filestore bugs:

- `flushPendingMsgsLocked` could, in some cases, hold onto a buffer after returning it to the pool, which could result in corruption
- `flushPendingMsgsLocked` was not making best use of reused buffers as the head of the buffer was left as "dead space" rather than reclaiming it
- `loadMsgsWithLock` would incorrectly report failing to load the cache as though there is corruption


derekcollison (approved):
LGTM

PR #7300, opened 2025-09-11
Change extended DataStats fields to pointers and export

bruth:
This was introduced in #6967 to support gateway, route, and leaf stats. However the DataStats type is used for client connection stats (e.g. disconnect event) for which the additional fields are not relevant. Having a pointer ensures the omitempty JSON tag is respected and those nil fields are not serialized.

### DIFF
--- a/server/events.go
+++ b/server/events.go
@@ -404,17 +404,17 @@ type GatewayStat struct {
 	NumInbound int       `json:"inbound_connections"`
 }
 
-type dataStats struct {
+type MsgBytes struct {
 	Msgs  int64 `json:"msgs"`
 	Bytes int64 `json:"bytes"`
 }
 
 // DataStats reports how may msg and bytes. Applicable for both sent and received.
 type DataStats struct {
-	dataStats
-	Gateways dataStats `json:"gateways,omitempty"`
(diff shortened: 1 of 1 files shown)


server/events.go
```diff
@@ -412,9 +412,9 @@ type dataStats struct {
 // DataStats reports how may msg and bytes. Applicable for both sent and received.
 type DataStats struct {
 	dataStats
-	Gateways dataStats `json:"gateways,omitempty"`
-	Routes   dataStats `json:"routes,omitempty"`
-	Leafs    dataStats `json:"leafs,omitempty"`
+	Gateways *dataStats `json:"gateways,omitempty"`
```
ripienaar: Can we export `dataStats`? Having unexpected structs in an exported structure is awkward

bruth: Would `MsgBytes` work as a name? (since there is already `DataStats`)

ripienaar: In jsm I went with `DataStatsSet` but will update with whatever you pick here - I am clearly terrible at naming :P

bruth: `DataStats` is quite ambiguous as well, but we need to leave that :) I will do `MsgBytes` for the two field struct of `Msgs` and `Bytes`.

PR #7325, opened 2025-09-17
Flaky test detector

sciascid:
A github workflow that checks if new commits are introduced in a pull request. If so runs them repeatedly to check if the test is flaky.


sciascid:
Commits with dummy tests are there for testing the workflow, should be removed before merging.

PR #1562, opened 2020-08-14
Inhibit Go's default TCP keepalive settings for NATS

philpennock:
Go 1.13 changed the semantics of the tuning parameters for TCP keepalives, including the default value.  This affects all TCP listeners.  The NATS protocol has its own L7 keepalive system (PING/PONG) and the Go defaults are not a good fit for some valid deployment scenarios, while Go doesn't directly expose a working API for tuning these.

Rather than add a configuration knob and pull in another dependency (with portability issues) just disable TCP keepalives for all listeners used for speaking the NATS protocol.

Change the tests so we test the same logic.  Do not change HTTP monitoring or the websocket API.

This addresses the same problem as Issue #1561 but does not resolve all of it, as we're not making this configurable.

/cc @nats-io/core

PR #7329, opened 2025-09-17
[CHANGED] Websocket and MQTT: Inhibit TCP Keep Alive

kozlovic:
In PR #1562, the use of `net.Listen` was replaced with a custom `natsListen` that uses a `net.ListenConfig` where `KeepAlive` has been disabled.

The development of MQTT was happening prior to that PR and when MQTT was merged, it had not been changed to use `natsListen`.

The websocket code was done a bit earlier than PR #1562, and it states that it was not changed for websocket, but I am not sure why. Since websocket is still NATS protocol, NATS has its own PING/PONG that should keep the connection alive. So I am changing the websocket code too.

### DIFF
--- a/server/mqtt.go
+++ b/server/mqtt.go
@@ -514,7 +514,7 @@ func (s *Server) startMQTT() {
 	hp := net.JoinHostPort(o.Host, strconv.Itoa(port))
 	s.mu.Lock()
 	s.mqtt.sessmgr.sessions = make(map[string]*mqttAccountSessionManager)
-	hl, err = net.Listen("tcp", hp)
+	hl, err = natsListen("tcp", hp)
 	s.mqtt.listenerErr = err
 	if err != nil {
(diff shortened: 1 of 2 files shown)


kozlovic:
@philpennock Phil, do you remember in PR #1562, why you specifically did not change the websocket's use of `net.Listen`? Since it is still NATS protocol (where we have ping/pong), I am not sure why it could not be replaced there too (maybe it was because of the tls.Listen vs net.Listen case?).

wallyqs:
I think it makes sense to merge for v2.12 to be consistent here.

wallyqs (approved):
LGTM

neilalexander:
Will merge after release for 2.12.1.

neilalexander:
Actually going to merge now as we need to do an 2.12.0-RC.6 anyway.

PR #7332, opened 2025-09-18
(2.12) Revert Green Tea GC on goreleaser builds

neilalexander:
This reverts #7301 as we may be experiencing some problems with it.

[skip ci]


MauriceVanVeen (approved):
LGTM

Issue #7092, opened 2025-07-22
Go client sdk built-in nats-server unable to read configuration file by passing ConfigFile Options.


yixianOu:
### Observed behavior

```go
package main

import (
	"fmt"
	"net/url"
	"os"
	"os/exec"
	"path/filepath"
	"strings"
	"time"

	"github.com/nats-io/nats-server/v2/server"
	"github.com/nats-io/nats.go"
)

type NATSClient struct {
	url  string
	opts []nats.Option
}

func main() {
	const (
		address          = "0.0.0.0"
		mainPort         = 4222
		clusterPort      = 7422
		leafPort         = 4223
		subjectName      = "foo"
		credsFileAppUser = "~/.local/share/nats/nsc/keys/creds/local/APP/user.creds"
		credsFileSysUser = "~/.local/share/nats/nsc/keys/creds/local/SYS/sys.creds"
	)

	// Step 1: Update operator configuration
	fmt.Println("Updating operator configuration...")
	execCmd := exec.Command("nsc", "edit", "operator", "--require-signing-keys",
		"--account-jwt-server-url", fmt.Sprintf("nats://%s:%d", address, mainPort))
	output, err := execCmd.CombinedOutput()
	if err != nil {
		panic(fmt.Sprintf("Failed to update operator: %v\nOutput: %s", err, string(output)))
	}
	fmt.Printf("Operator update output:\n%s\n", string(output))

	// Step 2: Update APP account
	fmt.Println("Updating APP account configuration...")
	execCmd = exec.Command("nsc", "edit", "account", "APP", "--sk", "generate")
	output, err = execCmd.CombinedOutput()
	if err != nil {
		panic(fmt.Sprintf("Failed to update APP account: %v\nOutput: %s", err, string(output)))
	}
	fmt.Printf("APP account update output:\n%s\n", string(output))

	// Step 3: Generate resolver configuration
	fmt.Println("Generating resolver configuration...")
	execCmd = exec.Command("nsc", "generate", "config", "--nats-resolver", "--sys-account", "SYS")
	output, err = execCmd.CombinedOutput()
	if err != nil {
		panic(fmt.Sprintf("Failed to generate resolver configuration: %v\nOutput: %s", err, string(output)))
	}

	resolverFile := "resolver.conf"
	err = os.WriteFile(resolverFile, output, 0644)
	if err != nil {
		panic(fmt.Sprintf("Failed to write resolver.conf: %v", err))
	}
[...]

PR #7333, opened 2025-09-19
[FIXED] Processes config file during embeded server initialization.

yixianOu:
Resolves https://github.com/nats-io/nats-server/issues/7092

Edited the NewServer function in the server package , server.go file. Using the ProcessConfigFile method of the Options type, to enable the embeded nats-server to read the config file

Prevent configuration files from being processed twice by checking opts.configDigest before calling ProcessConfigFile in NewServer.

Signed-off-by: orician <lzjzxOYX201905@gmail.com>

### DIFF
--- a/server/opts_test.go
+++ b/server/opts_test.go
@@ -3922,7 +3922,7 @@ func TestOptionsProxyRequired(t *testing.T) {
 	checkUsersAndNkeys(o.LeafNode.Users, false, nil)
 }
 
-// TestConfigStructReplacesLoadConfig plagiarises TestLargeMaxPayload but replaces
+// TestConfigStructReplacesLoadConfig plagiarizes TestLargeMaxPayload but replaces
 // LoadConfig call with a config struct.
 func TestConfigStructReplacesLoadConfig(t *testing.T) {
 	// Test 1: Large max_payload error (same as TestLargeMaxPayload first part)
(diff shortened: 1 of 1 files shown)


yixianOu:
This feature was discussed in #7093, but I still believe that if the embedded NATS server could provide the capability to read configuration files, it would make embedding NATS in Go much more user-friendly.

ripienaar:
I am still pro `NewServerFromConfig()` instead of this

rajanisdf:
I will refer to TestLargeMaxPayload to write test cases for this change, and I will also explore whether NewServerFromConfig() can provide a better implementation.

yixianOu:
Thank you for your suggestion. I will refer to TestLargeMaxPayload to write test cases for this change, and I will also explore whether NewServerFromConfig() can provide a better implementation.

yixianOu:
> I am still pro `NewServerFromConfig()` instead of this

This is a good approach considering forward compatibility, but I'd like to first test whether this change might be disruptive.

yixianOu:
> Principally looks good to me, but I think we need a unit test to prove this behaviour. You should be able to plagiarise a good bit of `TestLargeMaxPayload` in `opts_test.go` for example but replace the `LoadConfig` call in the test with a config struct.

Add proper error handling for config file processing failures in `NewServer`
Added 3 new testcases in opts_test.go :
*    `TestConfigStructReplacesLoadConfig`: Validates Options struct replaces `LoadConfig` calls
*    `TestConfigFileProcessingComparison`: Compares `LoadConfig` vs `Options.ProcessConfigFile`(called in `NewServer`)  behavior
*    `TestConfigFileHotReloadWithEmbeddedServer`: Ensures hot reload works with embedded servers

yixianOu:
Should I use `NewServerFromConfig()`  to implement the feature instead of directly modifying `NewServer()`?

PR #7334, opened 2025-09-19
(2.12) Filestore fixes

neilalexander:
This includes the following fixes for the filestore:

* The filestore buffer pools were unbalanced; a buffer retrieved from one that was exactly the pool threshold size would be returned to a different pool. This has been fixed and I've copied the same (working) code from the client pools.
* Removed some code that was a hangover from the partial write caches in `indexCacheBuf` that could potentially grow very large underlying buffers and indexes in the wrong conditions.
* More optimistically reuse the `indexCacheBuf` parameter buffer if possible, since it is going to have originated from the buffer pool anyway via `loadMsgsWithLock`. 
* When flushing pending messages with encryption, recycle the temporary encrypted buffer when done.

### DIFF
--- a/server/filestore_test.go
+++ b/server/filestore_test.go
@@ -2987,6 +2987,12 @@ func TestFileStoreBadConsumerState(t *testing.T) {
 
 func TestFileStoreExpireMsgsOnStart(t *testing.T) {
 	testFileStoreAllPermutations(t, func(t *testing.T, fcfg FileStoreConfig) {
+		if fcfg.Compression != NoCompression || fcfg.Cipher != NoCipher {
+			// TODO(nat): For some reason this test is very flaky in CI when using
+			// encryption or compression but passes fine without.
+			t.SkipNow()
+		}
+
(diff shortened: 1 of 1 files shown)


server/filestore.go
```diff
@@ -944,53 +944,56 @@ func (fs *fileStore) writeStreamMeta() error {
 }
 
 // Pools to recycle the blocks to help with memory pressure.
-var blkPoolBig sync.Pool    // 16MB
-var blkPoolMedium sync.Pool // 8MB
-var blkPoolSmall sync.Pool  // 2MB
+var blkPoolSmall = &sync.Pool{
+	New: func() any {
+		b := [defaultSmallBlockSize]byte{}
```
derekcollison, line 949: This more idiomatic these days vs make()?

neilalexander: Not exactly, the pool in this case is storing `*[N]byte` rather than `[]byte`, since it reduces the number of slice copies on `Get()` and `Put()`.

PR #7357, opened 2025-09-26
Filestore: Use binary search for sequence-from-timestamp lookups

### LINKED ISSUE #7353: JetStream timestamp seek: make `GetSeqFromTime` O(log n) and fix nil-mid deletion edge case
### Summary
- Timestamp-based seeks used by JetStream consumers (e.g., `OptStartTime`) were slow due to linear scans:
  - Linear block selection across `fs.blks`.
  - Linear intra-block scan to find the first message with `ts >= t`.
- We prototyped an improvement that switches both steps to binary search, delivering large speedups (details below).
- While validating, we identified an edge case in the binary search when the probed mid sequence is deleted (nil). The current search may skip the true lower-bound when mid is nil and the correct answer lies to the left.

https://github.com/nats-io/nats-server/pull/7352

### Proposed fix for nil-mid in binary search
When `mid` is deleted (nil), avoid discarding half of the range blindly:
- Probe to the nearest extant message using store-level jumps:
  - Use `LoadNextMsg(_EMPTY_, false, mid, ...)` to jump forward to the next extant seq; ignore if beyond current block (`seq > lseq`).
  - If no forward extant exists, probe backward with `LoadPrevMsg(mid, ...)`; ignore if before the block (`seq < fseq`).
- Update bounds based on the extant message timestamp:
  - If forward extant exists and `ts >= targetTs`: set `result = seq`; `right = seq - 1`.
  - Else if forward extant exists and `ts < targetTs`: `left = seq + 1`.
  - Else if only backward extant exists: compare and update bounds similarly.
  - If neither exists within bounds, shrink toward the other side (segment has no extant messages).

This keeps common-case complexity ~O(log n), with occasional O(1) jumps for deleted mids.

### Compatibility
- No on-disk format changes.
- No API changes.
- Preserves lower-bound semantics: return first sequence with `ts >= target`.
- Existing consumer clamping and stream state logic remain unchanged.

### Additional notes
- The optimization helps filestore; memstore already has a more efficient time lookup.
- Follow-up (optional): consider an in-memory sparse timestamp index per block to reduce probe costs further for repeated seeks. This is not required for correctness and can be scoped separately.

### Server and client version
latest master

### Host environment

_No response_

### Steps to reproduce

_No response_

neilalexander:
This updates the filestore `GetSeqFromTime` function to use a binary search, such that the computational complexity is more predictable. This closes #7352 and fixes #7353 as this is a more idiomatic approach.

Beforehand, the benchmark was very fast for earlier timestamps (as expected with a linear scan) but comparatively glacial for later ones:
```
go test -v ./server -run=XXX -bench=BenchmarkFileStoreGetSeqFromTime
goos: darwin
goarch: arm64
pkg: github.com/nats-io/nats-server/v2/server
cpu: Apple M2 Ultra
BenchmarkFileStoreGetSeqFromTime
BenchmarkFileStoreGetSeqFromTime/Start
BenchmarkFileStoreGetSeqFromTime/Start-24         	18652426	        61.56 ns/op	       0 B/op	       0 allocs/op
BenchmarkFileStoreGetSeqFromTime/Middle
BenchmarkFileStoreGetSeqFromTime/Middle-24        	  114956	     10275 ns/op	       2 B/op	       0 allocs/op
BenchmarkFileStoreGetSeqFromTime/End
BenchmarkFileStoreGetSeqFromTime/End-24           	   60656	     19740 ns/op	       0 B/op	       0 allocs/op
PASS
ok  	github.com/nats-io/nats-server/v2/server	11.673s
```

After the change, this is now more predictable in the entire range:
```
go test -v ./server -run=XXX -bench=BenchmarkFileStoreGetSeqFromTime
goos: darwin
goarch: arm64
pkg: github.com/nats-io/nats-server/v2/server
cpu: Apple M2 Ultra
BenchmarkFileStoreGetSeqFromTime
BenchmarkFileStoreGetSeqFromTime/Start
BenchmarkFileStoreGetSeqFromTime/Start-24         	 9555956	       130.8 ns/op	       0 B/op	       0 allocs/op
BenchmarkFileStoreGetSeqFromTime/Middle
BenchmarkFileStoreGetSeqFromTime/Middle-24        	 9335430	       127.4 ns/op	       0 B/op	       0 allocs/op
BenchmarkFileStoreGetSeqFromTime/End
BenchmarkFileStoreGetSeqFromTime/End-24           	10001110	       118.6 ns/op	       0 B/op	       0 allocs/op
PASS
ok  	github.com/nats-io/nats-server/v2/server	11.296s
```

### DIFF
--- a/server/filestore_test.go
+++ b/server/filestore_test.go
@@ -10534,3 +10534,55 @@ func TestFileStoreCorruptedNonOrderedSequences(t *testing.T) {
 		})
 	}
 }
+
+func BenchmarkFileStoreGetSeqFromTime(b *testing.B) {
+	fs, err := newFileStore(
+		FileStoreConfig{
+			StoreDir:  b.TempDir(),
+			BlockSize: 16,
+		},
+		StreamConfig{
+			Name:     "foo",
+			Subjects: []string{"foo.>"},
+			Storage:  FileStorage,
+		},
+	)
+	require_NoError(b, err)
+	defer fs.Stop()
+
+	for range 4096 {
+		_, _, err := fs.StoreMsg("foo.bar", nil, []byte{1, 2, 3, 4, 5}, 0)
+		require_NoError(b, err)
+	}
+
+	fs.mu.RLock()
+	fs.blks[0].mu.RLock()
+	fs.lmb.mu.RLock()
+	start := time.Unix(0, fs.blks[0].first.ts)
+	middle := time.Unix(0, fs.blks[0].first.ts+(fs.lmb.last.ts-fs.blks[0].first.ts)/2)
+	end := time.Unix(0, fs.lmb.last.ts)
+	fs.blks[0].mu.RUnlock()
+	fs.lmb.mu.RUnlock()
+	fs.mu.RUnlock()
+
+	b.Run("Start", func(b *testing.B) {
+		b.ReportAllocs()
+		for range b.N {
+			fs.GetSeqFromTime(start)
+		}
+	})
+
+	b.Run("Middle", func(b *testing.B) {
+		b.ReportAllocs()
+		for range b.N {
+			fs.GetSeqFromTime(middle)
+		}
+	})
+
+	b.Run("End", func(b *testing.B) {
+		b.ReportAllocs()
+		for range b.N {
+			fs.GetSeqFromTime(end)
+		}
+	})
+}


MauriceVanVeen (approved):
LGTM

server/filestore_test.go
```diff
@@ -10534,3 +10534,56 @@ func TestFileStoreCorruptedNonOrderedSequences(t *testing.T) {
...
+			Name:     "foo",
+			Subjects: []string{"foo.>"},
+			Storage:  FileStorage,
+		},
+	)
+	require_NoError(b, err)
+	defer fs.Stop()
+
+	for range 4096 {
+		_, _, err := fs.StoreMsg("foo.bar", nil, []byte{1, 2, 3, 4, 5}, 0)
```
sciascid, line 10554: You could use StoreMsgRaw()  and pass a timestamp, and avoid the time.Sleep() at the end of the loop.

sciascid (approved):
LGTM

server/filestore_test.go
neilalexander: We don't really need to specify our own timestamps as even the nanosecond precision should be enough (the small block size ensures every message gets its own block), but good spot that I left the sleep in, I will remove it.

MauriceVanVeen (approved):
LGTM

darioalessandro:
Cool 😄 does a nats image is generated upon merging this? Where can I get it?

neilalexander:
If it's a Docker image you'd like to test, I can kick off an early rebuild of the nightly image?

darioalessandro:
Yes, that would be awesome 🙏

neilalexander:
No problem, the updated nightly has been pushed: https://hub.docker.com/r/synadia/nats-server/tags

darioalessandro:
@neilalexander may I have an arm image? all our nats servers have `arm` processors and this would help me to test faster. 

Also would it be possible to cherrypick this fix into `2.11` and `2.12` please?

darioalessandro:
thanks @neilalexander !!!

PR #7385, opened 2025-10-02
Use optional sync when writing TTL and scheduling state in filestore

neilalexander:


MauriceVanVeen (approved):
LGTM

PR #7388, opened 2025-10-02
Always use temporary staging for filestore meta file writes

neilalexander:
This provides better crash-resilience during interrupted filesystem operations, i.e. where writing `meta.inf`/`meta.sum`/`index.db` is interrupted with `O_TRUNC` leaving behind an empty file.

### DIFF
--- a/server/filestore.go
+++ b/server/filestore.go
@@ -11764,29 +11764,47 @@ func (alg StoreCompression) Decompress(buf []byte) ([]byte, error) {
 // sets O_SYNC on the open file if SyncAlways is set. The dios semaphore is
 // handled automatically by this function, so don't wrap calls to it in dios.
 func (fs *fileStore) writeFileWithOptionalSync(name string, data []byte, perm fs.FileMode) error {
-	if fs.fcfg.SyncAlways {
-		return writeFileWithSync(name, data, perm)
-	}
-	<-dios
-	defer func() {
-		dios <- struct{}{}
-	}()
-	return os.WriteFile(name, data, perm)
+	return writeAtomically(name, data, perm, fs.fcfg.SyncAlways)
 }
 
 func writeFileWithSync(name string, data []byte, perm fs.FileMode) error {
+	return writeAtomically(name, data, perm, true)
+}
+
+func writeAtomically(name string, data []byte, perm fs.FileMode, sync bool) error {
+	tmp := name + ".tmp"
+	flags := os.O_CREATE | os.O_WRONLY | os.O_TRUNC
+	if sync {
+		flags = flags | os.O_SYNC
+	}
 	<-dios
 	defer func() {
 		dios <- struct{}{}
 	}()
-	flags := os.O_WRONLY | os.O_CREATE | os.O_TRUNC | os.O_SYNC
-	f, err := os.OpenFile(name, flags, perm)
+	f, err := os.OpenFile(tmp, flags, perm)
 	if err != nil {
 		return err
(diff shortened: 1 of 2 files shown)


server/filestore.go
```diff
@@ -11764,29 +11764,50 @@ func (alg StoreCompression) Decompress(buf []byte) ([]byte, error) {
...
 
 func writeFileWithSync(name string, data []byte, perm fs.FileMode) error {
+	return writeAtomically(name, data, perm, true)
+}
+
+func writeAtomically(name string, data []byte, perm fs.FileMode, sync bool) error {
+	tmp := name + ".tmp"
+	flags := os.O_CREATE | os.O_WRONLY | os.O_TRUNC
+	if sync {
+		flags = flags | os.O_SYNC
```
sciascid, line 11778: I think that the O_SYNC flag is not needed, since you call os.fsync() after each write operation.
And, if writeAtomically() is called with sync=false, you want to skip the fsync() call.

See https://man7.org/linux/man-pages/man2/open.2.html:

```
O_SYNC

Write operations on the file will complete according to the
requirements of synchronized I/O file integrity completion
(by contrast with the synchronized I/O data integrity
completion provided by O_DSYNC.)

By the time [write(2)](https://man7.org/linux/man-pages/man2/write.2.html) (or similar) returns, the output data
and associated file metadata have been transferred to the
underlying hardware (i.e., as though each [write(2)](https://man7.org/linux/man-pages/man2/write.2.html) was
followed by a call to [fsync(2)](https://man7.org/linux/man-pages/man2/fsync.2.html)).  See VERSIONS.
```

neilalexander: I think by contrast it's probably easier to keep `O_SYNC` and remove the manual `.Sync()` call, since with a single big write, they are a practical equivalence. Had not intended to leave `.Sync()` in un-gated like that.

sciascid: Yes, either `O_SYNC` or manual `.Sync()`, agreed

sciascid:
I was going to suggest something similar. In the end I didn't, because I thought that the current approach may work for small files, like Raft's term and vote file for example. As single sector overwrites are typically atomic.

This may be an interesting read: https://www.usenix.org/system/files/conference/osdi14/osdi14-paper-pillai.pdf
See section 2.2.1

server/filestore.go
neilalexander: Fixed

server/filestore.go
```diff
@@ -11764,29 +11764,45 @@ func (alg StoreCompression) Decompress(buf []byte) ([]byte, error) {
...
 	}
-	return f.Close()
+	if err := os.Rename(tmp, name); err != nil {
+		_ = os.Remove(tmp)
+		return err
+	}
+	// To ensure that the file rename was persisted on all filesystems,
+	// try to flush the directory metadata.
+	if d, err := os.Open(filepath.Dir(name)); err == nil {
+		_ = d.Sync()
```
sciascid: Notice, if `sync` is `false`,  this will still do a `.Sync()`. Originally, there were no `.Sync()` calls when calling into `writeFileWithOptionalSync()` and `SyncAlways` disabled.
Is this intended?

neilalexander: This is syncing the directory metadata rather than the file metadata, so that the rename is durable, although this may not be needed if `sync == false`.

neilalexander: I suppose it isn't, I've gated it.

PR #7456, opened 2025-10-21
(2.12) Add opt-in support for the PROXY protocol

bruth:
The use case is to allow for transparent mapping of the source IP to the server for monitoring, auth callout, etc. See https://www.haproxy.org/download/1.8/doc/proxy-protocol.txt

Note, this was largely Claude generated since the spec is well defined.

### DIFF
--- a/server/client_proxyproto.go
+++ b/server/client_proxyproto.go
@@ -71,9 +71,10 @@ const (
 
 var (
 	// Errors
-	errProxyProtoInvalid     = errors.New("invalid PROXY protocol header")
-	errProxyProtoUnsupported = errors.New("unsupported PROXY protocol feature")
-	errProxyProtoTimeout     = errors.New("timeout reading PROXY protocol header")
+	errProxyProtoInvalid      = errors.New("invalid PROXY protocol header")
+	errProxyProtoUnsupported  = errors.New("unsupported PROXY protocol feature")
+	errProxyProtoTimeout      = errors.New("timeout reading PROXY protocol header")
+	errProxyProtoUnrecognized = errors.New("unrecognized PROXY protocol format")
 )
 
 // proxyProtoAddr contains the address information extracted from PROXY protocol header
@@ -118,18 +119,14 @@ func detectProxyProtoVersion(conn net.Conn) (version int, header []byte, err err
 	if _, err = io.ReadFull(conn, header); err != nil {
 		return 0, nil, fmt.Errorf("failed to read protocol version: %w", err)
 	}
-
-	// Check for v1 text format: "PROXY "
-	if string(header) == proxyProtoV1Prefix {
+	switch bytesToString(header) {
+	case proxyProtoV1Prefix:
 		return 1, header, nil
-	}
-
-	// Check for v2 signature (first 6 bytes of 12-byte signature)
-	if string(header) == proxyProtoV2Sig[:6] {
+	case proxyProtoV2Sig[:6]:
 		return 2, header, nil
+	default:
+		return 0, nil, errProxyProtoUnrecognized
 	}
-
-	return 0, nil, fmt.Errorf("unrecognized PROXY protocol format")
 }
 
 // readProxyProtoV1Header parses PROXY protocol v1 text format.
@@ -155,7 +152,7 @@ func readProxyProtoV1Header(conn net.Conn) (*proxyProtoAddr, error) {
 		for i := 0; i < len(line)-1; i++ {
 			if line[i] == '\r' && line[i+1] == '\n' {
 				// Found CRLF - extract just the line portion
-				line = line[:i+2]
+				line = line[:i]
 				goto foundCRLF
 			}
 		}
@@ -165,13 +162,7 @@ func readProxyProtoV1Header(conn net.Conn) (*proxyProtoAddr, error) {
 	return nil, fmt.Errorf("%w: v1 line too long", errProxyProtoInvalid)
 
 foundCRLF:
-	// Line is properly terminated with CRLF
-	if len(line) < 2 {
-		return nil, fmt.Errorf("%w: v1 line too short", errProxyProtoInvalid)
-	}
-
-	// Remove CRLF and parse
-	line = line[:len(line)-2]
+	// Get parts from the protocol
 	parts := strings.Fields(string(line))
 
 	// Validate format
@@ -330,7 +321,7 @@ func parseProxyProtoV2Header(conn net.Conn, header []byte) (*proxyProtoAddr, err
 		// For LOCAL, we should skip the address data if any
 		if addrLen > 0 {
 			// Discard the address data
-			if err := discardBytes(conn, int(addrLen)); err != nil {
+			if _, err := io.CopyN(io.Discard, conn, int64(addrLen)); err != nil {
 				return nil, fmt.Errorf("failed to discard LOCAL command address data: %w", err)
 			}
 		}
@@ -350,7 +341,6 @@ func parseProxyProtoV2Header(conn net.Conn, header []byte) (*proxyProtoAddr, err
 	// Parse address data based on family
 	var addr *proxyProtoAddr
 	var err error
-
 	switch family {
 	case proxyProtoFamilyInet:
 		addr, err = parseIPv4Addr(conn, addrLen)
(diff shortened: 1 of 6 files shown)


server/client_proxyproto_test.go
```diff
@@ -304,14 +304,14 @@ func TestProxyConnRemoteAddr(t *testing.T) {
 func TestClientProxyProtoV2EndToEnd(t *testing.T) {
 	// Start a test server with PROXY protocol enabled
 	opts := DefaultOptions()
-	opts.Port = -1 // Random port
-	opts.ProxyProtocol = true
+	opts.Port = -1
+	opts.ProxiedPort = -1
```
derekcollison: Not a fan of this approach IMO. But I understand why we ended up here.

If we say proxy has to be TLS first enabled, would that resolve?

server/server.go
```diff
@@ -3432,6 +3440,47 @@ func (s *Server) createClientEx(conn net.Conn, inProcess bool) *client {
 		}
 	}
 
+	// Check for proxy protocol if enabled.
+	if !isClosed && !tlsRequired && opts.ProxyProtocol {
+		if len(pre) == 0 {
+			// There has been no pre-read yet, do so so we can work out
+			// if the client is trying to negotiate PROXY.
+			pre = make([]byte, 6)
+			c.nc.SetReadDeadline(time.Now().Add(proxyProtoReadTimeout))
```
kozlovic, line 3449: Looks like readProxyProtoHeader() invoked below is going to do a read. Granted, it would be from tlsMixConn that will already have the pre buffer read, so there won't be an actual io read, but I wonder if this is necessary to do it here.

server/server.go
```diff
@@ -3432,6 +3440,47 @@ func (s *Server) createClientEx(conn net.Conn, inProcess bool) *client {
...
+			n, _ := io.ReadFull(c.nc, pre)
+			c.nc.SetReadDeadline(time.Time{})
+			pre = pre[:n]
+		}
+		conn = &tlsMixConn{conn, bytes.NewBuffer(pre)}
+		addr, err := readProxyProtoHeader(conn)
+		if err != nil && err != errProxyProtoUnrecognized {
```
kozlovic, line 3456: The `errProxyProtoUnrecognized` is returned only if we don't detect a proxy v1 or v2. So you are checking that there was an unexpected error parsing the proxy protocol.

server/server.go
```diff
@@ -3432,6 +3440,47 @@ func (s *Server) createClientEx(conn net.Conn, inProcess bool) *client {
...
+			c.mu.Unlock()
+			s.Warnf("Error reading PROXY protocol header from %s: %v", conn.RemoteAddr(), err)
+			c.closeConnection(MaxConnectionsExceeded)
```
kozlovic: Is that appropriate close reason? Existing ones we have that could fit better: `ProtocolViolation` or `BadClientProtocolVersion` maybe?

server/server.go
```diff
@@ -3432,6 +3440,47 @@ func (s *Server) createClientEx(conn net.Conn, inProcess bool) *client {
...
+			return nil
+		}
```
kozlovic, line 3463: So what if `err == errProxyProtoUnrecognized`? The first 6 bytes of the incoming connection have been read. We don't clear `pre` below if there was an error. So I guess that makes sense to read `pre` in this function as opposed to let `readProxyProtoHeader()` do it?

server/client_proxyproto.go
```diff
@@ -0,0 +1,404 @@
...
+// detectProxyProtoVersion reads the first bytes and determines protocol version.
+// Returns 1 for v1, 2 for v2, or error.
+// The first 6 bytes read are returned so they can be used by the parser.
+func detectProxyProtoVersion(conn net.Conn) (version int, header []byte, err error) {
+	// Read first 6 bytes to check for "PROXY " or v2 signature
+	header = make([]byte, 6)
+	if _, err = io.ReadFull(conn, header); err != nil {
+		return 0, nil, fmt.Errorf("failed to read protocol version: %w", err)
+	}
+	switch string(header) {
```
kozlovic: Could use `bytesToString(header)` helper that we have to avoid men allocation here for the compare.

server/client_proxyproto.go
```diff
@@ -0,0 +1,404 @@
...
+			return nil, fmt.Errorf("failed to read v1 line: %w", err)
+		}
+
+		line = buf[:len(line)+n]
+
+		// Look for CRLF in what we've read so far
+		for i := 0; i < len(line)-1; i++ {
+			if line[i] == '\r' && line[i+1] == '\n' {
+				// Found CRLF - extract just the line portion
+				line = line[:i+2]
```
kozlovic: Since we are going to remove the CR_LF, could we have just stop at `i`?

server/client_proxyproto.go
```diff
@@ -0,0 +1,404 @@
...
+				goto foundCRLF
+			}
+		}
+	}
+
+	// Exceeded max length without finding CRLF
+	return nil, fmt.Errorf("%w: v1 line too long", errProxyProtoInvalid)
+
+foundCRLF:
+	// Line is properly terminated with CRLF
```
kozlovic: We know that we found CRLF otherwise we would have returned above with an error. So the test below and removal of CRLF would not be needed.

server/client_proxyproto.go
```diff
@@ -0,0 +1,404 @@
...
+	if len(line) < 2 {
+		return nil, fmt.Errorf("%w: v1 line too short", errProxyProtoInvalid)
+	}
+
+	// Remove CRLF and parse
+	line = line[:len(line)-2]
+	parts := strings.Fields(string(line))
```
kozlovic, line 166: We could start here.

server/client_proxyproto.go
```diff
@@ -0,0 +1,404 @@
...
+		return nil, err
+	}
+
+	switch version {
+	case 1:
+		// v1 parser expects "PROXY " prefix already consumed
+		return readProxyProtoV1Header(conn)
+	case 2:
+		// Read rest of v2 signature (bytes 6-11, total 6 more bytes)
+		remaining := make([]byte, 6)
```
kozlovic, line 245: Question: If the proto v2 is fixed size. Could we create the buffer for the remaining and do a single ReadFull() instead of breaking it in the next expected bytes require for the "next" check? If it is a valid v2 and the size is fixed, then we should be able to read the full remaining bytes and then work on that buffer for the appropriate remaining checks, etc...

kozlovic (changes requested):

server/server.go
neilalexander: The problem here is that the connection could either have the proxy protocol preamble or not, depending on whether it's a proxied or a direct connection, so we have no choice but to detect that and read it out of the way if so.

server/server.go
neilalexander: It's `!=`, so yes, we're checking that we failed to parse the header at all. Even if we succeed in parsing, it's possible that the proxy preamble will contain a `LOCAL` instead of a proxied IP, at which point `addr == nil` and `err == nil`.

server/server.go
neilalexander: `ProtocolViolation` sounds good to me.

server/server.go
neilalexander: `err == errProxyProtoUnrecognized` could also be a non-proxy-protocol'd connection, so if we clear the pre-read in that case, the next step will not be able to read back from the beginning of the connection, which breaks the protocol. I've added better comments to explain this.

server/client_proxyproto.go
neilalexander: Yep, done.

server/client_proxyproto.go
neilalexander: I don't think it is strictly fixed-size, depending on the address family.

server/client_proxyproto.go
```diff
@@ -0,0 +1,403 @@
...
+				goto foundCRLF
+			}
+		}
+	}
+
+	// Exceeded max length without finding CRLF
+	return nil, fmt.Errorf("%w: v1 line too long", errProxyProtoInvalid)
+
+foundCRLF:
+	// Line is properly terminated with CRLF
```
kozlovic: I would remove that comment and check below (length < 2) since it is not relevant anymore.

server/client_proxyproto.go
```diff
@@ -0,0 +1,403 @@
...
+	if len(line) < 2 {
+		return nil, fmt.Errorf("%w: v1 line too short", errProxyProtoInvalid)
+	}
+
+	// Remove CRLF and parse
```
kozlovic: Change to something like: "// Get the parts from the protocol"

server/client_proxyproto.go
kozlovic: But at least the 16 first bytes of the protocol are fixed: See this section is the referenced protocol spec in the PR's description:
```
So the 16-byte version 2 header can be described this way :

    struct proxy_hdr_v2 {
        uint8_t sig[12];  /* hex 0D 0A 0D 0A 00 0D 0A 51 55 49 54 0A */
        uint8_t ver_cmd;  /* protocol version and command */
        uint8_t fam;      /* protocol family and address */
        uint16_t len;     /* number of following bytes part of the header */
    };

Starting from the 17th byte, addresses are presented in network byte order.
...
```

But it's fine, if it makes the code easier to write/manage, I think it is ok to leave as-is.

server/client_proxyproto.go
neilalexander: Done.

server/client_proxyproto.go
neilalexander: Done.

